PR #137, opened 2014-03-05
event handling in WebdriverJS

christian-bromann:
It should be possible to register callbacks on events that could be influence the tests. For example:

``` javascript
client.on('error', function(err) { ... }); 
client.on('command', function(data) { ... });
client.on('result', function(res) { ... });
client.on('end', function(data) { ... });
```

Also possible could be this syntax:

``` javascript
client.onError(function(err) { ... }); 
client.onCommand(function(data) { ... });
client.onResult(function(res) { ... });
client.onEnd(function(data) { ... });
```

Are there any other events we should support?

Both should be possible to use since the implementation doesn't really cost any extra effort.

It could be also useful to register and listen to own events. A proper use case is when you have two WebdriverJS instances and two different browser (e.g. if you'ld like to test a chat client) then it would be really useful to have own events to tell the other instance to proceed with the test. It enables synchronizations between multiple tests.

### DIFF
--- a/lib/utils/EventHandler.js
+++ b/lib/utils/EventHandler.js
@@ -24,17 +24,40 @@ var EventHandler = function(instance) {
             return true;
         }
 
+        /**
+         * register EventEmitter function to WebdriverJS instance.
+         * The function itself basically just creates a temporary
+         * custom command which gets executed immediately. With
+         * this approach it is possible to chain synchronous logic
+         * within asyncronous workflow.
+         */
         self.instance[funcName] = function() {
             var args = arguments;
 
+            /**
+             * create temporary custom command. In order to keep
+             * the event namespace clean and to avoid overriding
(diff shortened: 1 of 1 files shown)


webmagit:
@christian-bromann here's my usecase that I'm thinking is related to this.

Scenario: A few selectors have changed but need to debug to figure out which page objects need to be updated.

A typical debugging session involves running the script to figure out where the failure is. This is achieved easily by placing a browser.debug() in the onError hook and then correcting the selectors. BUT this involves re-starting the test all over again to get to the next point of failure.

I'd like to be able to correct the selector (and manually perform expected action), exit the debug and let the script continue to look out for the next point of failure (without having to quit and restart the script again)

Appreciate any advice on how wdio can continue onError? (ex: ignore errors on browser.click)

PR #385, opened 2015-01-02
do not throw errors when there is a user error handler

Soarez:
When a `on('error', cb)` is registered, errors shouldn't be thrown, i.e. this program shouldn't crash:

```js
var client = require('webdriverio');

client
  .remote()
  .init()
  .on('error', catcher)
  .url('http://example.com')
  .waitFor('.imaginary', 1000)
  .end();

setTimeout(function() {
  console.log('All ok');
}, 3000);

function catcher(e) {
  console.log('Caught error');
}
```
Or am I looking at this wrong?


christian-bromann:
@Soarez sorry missed to put this into `v2.4.2`. Will do another release tonight including this bugfix.
Cheers

PR #393, opened 2015-01-09
Add ability to get screenshot without writing on disk

MoLLIa:
Sometimes writing screenshot no disk is unnecessary.
Also i've slightly updated tests and docs.

And one more fix for windowHandlePosition test.
I've wrote about it here https://github.com/webdriverio/webdriverio/pull/341#issuecomment-62886775 but looks like you've missed my comment.

Please let me know whether something is wrong.

### DIFF
--- a/lib/commands/saveScreenshot.js
+++ b/lib/commands/saveScreenshot.js
@@ -6,11 +6,16 @@
  * to enhance this command with that functionality.
  *
  * <example>
-    :saveScreenshot.js
-    client.saveScreenshot('snapshot.png');
+ :saveScreenshot.js
+ // receive screenshot as Buffer
+ client.saveScreenshot(function(err, screenshot) { ... });
+ // save screenshot to file and receive as Buffer
+ client.saveScreenshot('snapshot.png', function(err, screenshot) { ... });
+ // save screenshot to file
+ client.saveScreenshot('snapshot.png');
  * </example>
  *
- * @param {String}   fileName    path of generated image (relative to the execution directory)
+ * @param {Function|String}   response    callback or path to generated image (relative to the execution directory)
  *
  * @uses protocol/screenshot
  * @type utility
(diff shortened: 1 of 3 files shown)


lib/commands/saveScreenshot.js
```diff
@@ -31,26 +36,28 @@ module.exports = function saveScreenshot(fileName) {
...
+
+            if (typeof response === 'string') {
+                fs.writeFile(response, screenshot, cb);
+            } else {
+                cb();
+            }
         }
     ], function(err) {
-
-        callback(err, null, response);
```
christian-bromann: we should stay consistent here, meaning that callback should return with error, result and response as parameters here

test/spec/desktop/saveScreenshot.js
```diff
@@ -5,23 +5,30 @@ describe('saveScreenshot', function() {
...
 
         var screenshotPath = path.join(__dirname, '..', '..', '..', 'test.png');
 
         this.client
-            .saveScreenshot(screenshotPath, true)
+            .saveScreenshot(screenshotPath)
             .call(function() {
-
                 fs.exists(screenshotPath, function(fileExists) {
-                    fileExists.should.be.true;
```
christian-bromann: why this change?

test/spec/desktop/saveScreenshot.js
```diff
@@ -5,23 +5,30 @@ describe('saveScreenshot', function() {
...
     before(h.setup());
 
-    var imageSize = {};
-
     it('should take a screenshot and output it on a desired location', function(done) {
 
         var screenshotPath = path.join(__dirname, '..', '..', '..', 'test.png');
 
         this.client
-            .saveScreenshot(screenshotPath, true)
```
christian-bromann: :+1: good catch, was an obsolete parameter

christian-bromann:
@MoLLIa first of all thanks for your contribution. I found couple of things. I still gonna merge this one and will fix them on my own

:+1:

MoLLIa:
You are welcome. Thank you for quick response upon pull request. I am glad to help you.

PR #399, opened 2015-01-23
Fixed promise handler

tommy351:
The current promise handler throws the following error when I trying to chain promise. I found that it requires `getPromise` method, which is not a standard in promise and not implemented by most libraries. This problem can be solved by detecting whether the returned promise supports `getPromise` method.

```
TypeError: Object [object Promise] has no method 'getPromise'
  at .../node_modules/webdriverio/lib/utils/PromiseHandler.js:84:53
  at process._tickDomainCallback (node.js:486:13)
```

### DIFF
--- a/lib/utils/PromiseHandler.js
+++ b/lib/utils/PromiseHandler.js
@@ -85,7 +85,7 @@ module.exports = function PromiseHandler(fnName, fn) {
                  * queue up next "then" if existing
                  */
                 if(returnPromise && returnPromise.then) {
-                    lastThenable.push(returnPromise.getPromise());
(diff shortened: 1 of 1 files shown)


christian-bromann:
Hi @tommy351 .. Thanks for your PR. Could you add an easy reproducible functional test? Or do you have a simple example that reproduces this error?

tommy351:
@christian-bromann Thanks for your fast reply. Here's an example:

``` js
var Q = require('q');

var deferred = Q.defer();
deferred.resolve('success');

wd.init()
  .url('http://localhost:4000')
  .click('a')
  .then(function(){
    return deferred.promise;
  })
  .then(function(result){
    console.log(result);
  });
```

christian-bromann:
Thanks man!

PR #410, opened 2015-02-02
Allow for optional arguments in addCommand

dylancwood:
# Background
The addCommand method allows custom functions to be added to the webdriverio client. When the custom command is called, webdriverio will call the custom function with `this` equal to the client, and it will append a callback to the arguments passed to the custom command.
Example:
```js
var customFunc = function(a) {...};
client.addCommand('customCommand', customFunc);
client.customCommand('foo');
// This will essentially call
// customFunc.call(client, ['foo'].push(nextWaterfall));
```
# Problem
Appending the callback to the arguments passed to the customCommand makes it difficult to gracefully handle optional arguments. For example, if the `customCommand` above were invoked with no arguments, the callback would be the first and only argument. This makes it challenging to use standard techniques to set default values for optional arguments (e.g. `optionalArg = optionalArg || 'default';`).

# Solution
Wrapping the `customFunc` in an argument padding function will cause the number of arguments passed to the customFunc is always the same as the function signature. The library vargs-callback(https://www.npmjs.com/package/vargs-callback) provides an easy way to do this. 

# Example
Place the script below in the root directory of webdriverio as `testOptionalArgs.js`. Run the script with node testOptionalArgs.js (after starting selenium server). 

```js
/**
 * test script to highlight the shortcomings of 
 * simply appending the chaining callback
 * to the list of input args.
 * 
 * Place this script in the webdriverio root directory.
 * Run this script with `node testOptionalArgs.js`.
 */
'use strict';

var webdriverio = require('./index.js');
var client;

var options = {
    desiredCapabilities: {
        browserName: 'chrome'
    }
};

/**
 * example function with optional input params
 * Note that it IS possible to use the `arguments` var 
 * to more safely test for the existance of `text` and `waitTime`.
 * However, even this approach becomes unwieldy and confusing with
 * two or more optional parameters.
 */
var typeExampleText = function(text, waitTime, cb) {
    text = text || 'Example text courtesy of webdriverio';
    waitTime = waitTime || 10000;
    cb = arguments[arguments.length - 1];
    this
        .click('input[name="q"]')
        .keys([text])
        .pause(10000)
        .waitForExist('input[name="q"]')
        .call(cb);
};

client = webdriverio.remote(options);
client.addCommand('typeExampleText',typeExampleText);

client
    .init()
    .url('https://google.com')
    .typeExampleText()//pass no arguments (just use defaults)
    .getValue('input[name="q"]', function(err, res) {
        console.log('You typed: `%s`',res)

    })
    .end();
```
# Test
I added a test :-)

# Discussion
I realize that it is common practice to pass a callback as the last argument to a function. However, in this case, I think that maintaining the number of arguments in the signature makes the addCommand system more intuitive. Since webdriverio is quietly appending the callback, it is very easy for a developer to forget that the callback will be added after the arguments that they pass to the customCommand. Further, handling the possible variations of optional callbacks is greatly simplified with vargs-callback. 

If you're opposed to adding this to the webdriverio library, I understand, but I suggest at least mentioning it in the docs. Let me know if you prefer to go that route, and I will happily add a couple sentences to the docs.

### DIFF
--- a/lib/webdriverio.js
+++ b/lib/webdriverio.js
@@ -4,6 +4,7 @@ var buildPrototype = require('./utils/buildPrototype'),
     EventHandler   = require('./utils/EventHandler'),
     PromiseHandler = require('./utils/PromiseHandler'),
     isMobileHelper = require('./helpers/isMobile');
+    vargsCallback  = require('vargs-callback');
 
 buildPrototype(
     WebdriverIO.prototype,
@@ -74,6 +75,7 @@ function WebdriverIO(options) {
      * to be added to asynchronous chain.
      */
     this.addCommand = function(name, fn) {
+        fn = vargsCallback(fn);
         chainIt.add(this, name, fn);
 
         /**
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
     "request": "~2.34.0",
     "rgb2hex": "^0.1.0",
     "url": "^0.10.1",
+    "vargs-callback": "^0.2.4",
     "wgxpath": "^0.23.0"
   },
   "devDependencies": {
--- a/test/spec/functional/addCommand.js
+++ b/test/spec/functional/addCommand.js
@@ -3,6 +3,7 @@ describe('addCommand', function() {
     before(h.setup());
 
     before(function() {
+        var Url = require('url');
 
         this.client.addCommand('getUrlAndTitle', function(callback) {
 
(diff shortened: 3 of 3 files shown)


fat-troll:
In fact affected me also a lot of times :)

christian-bromann:
@dylancwood thanks for this PR!

To be honest, I am not quite convinced to adapt that change. Here is the reason: in my opinion we shouldn't make this process different then in the nodejs/js world. People are used to the fact the if the function is called with less parameter then expected, the function should/can handle this situation. It takes more words to explain that `addCommand` behaves like `vargs-callback` as just to say "hey, grab the callback with arguments[arguments.length - 1]".

In general I'ld like to keep the API as "nodeiomatic" as possible. I added some [additional notes](https://github.com/webdriverio/webdriver.io/commit/48cb0a7a935a40124cfde6811c4080256b756039) about that topic to make it more clear. Does that make sense @dylancwood ?

dylancwood:
I understand your objection, and agree that keeping things nodeiomatic is best. The added documentation will help others avoid the issues that I had while getting familiar with webdriverio and chainit. Thank you for being so considerate, and I am happy to close this PR.

christian-bromann:
awesome thanks :+1:

PR #420, opened 2015-02-05
Updated waitForExist documentation to be more clear.  Made slight code change

notnarb:
Slight code change because the check to see that 'res' and 'res.value' existed was in the line after res.value.length was accessed.  If operating under the presumption that 'res' and 'res.value' may not exist, the line accessing 'res.value.length' before it would throw an exception.

Note: this is unrelated to issue 419


christian-bromann:
Awesome, thanks!

PR #422, opened 2015-02-07
Issue #419: Add SelectorTimeoutError for waitForVisible and waitForExist

### LINKED ISSUE #419: waitForExist returns (null, false) on a timeout
It seems that most commands will return an error (breaking the chain) when an attempted element is not found, but it seems like waitForExist acts a little differently:
```javascript
if(now - start > ms) {
    return cb(null, false);
}
```
Which allows the chain to continue unless a specific callback is in place to catch a 'false' result value.

e.g.:
```javascript
client()
    .waitForExist('.loadingIcon')
    .click('.loadingIcon')
```
will continue (without failure) to the click event even if the waitForExist loop times out.

---
Perhaps a new error type should be defined and thrown in this instance?
Or maybe the documentation could more clearly reflect that it is not supposed to error on a failure to find an element, and instead return false as a result

notnarb:
Closes #419 

```
  1) Checks to see if passwordz exists:
     Uncaught SelectorTimeoutError
     Problem: waitForExist failed.  Selector: #passwordz failed to find element that was in the DOM

     Callstack:
     -> waitForExist("#passwordz",1000)
```

```
  2) Checks to see if passwordz is visible:
     Uncaught SelectorTimeoutError
     Problem: waitForVisible failed.  Selector: #passwordz failed to find element that was visible

     Callstack:
     -> waitForVisible("#passwordz",1000)
```
Note: this did not implement the SelectorTimeoutError for the other waitFor<X> functions

---
(Sorry for the trailing newline added to the ErrorHandler, I had trouble getting an editor to save the file without it.)

### DIFF
--- a/lib/commands/waitForExist.js
+++ b/lib/commands/waitForExist.js
@@ -67,7 +67,7 @@ module.exports = function waitForExist(selector, ms, reverse) {
                 if(now - start > ms) {
                     var checkType = reverse ? 'not in the DOM' : 'in the DOM';
                     var err = ErrorHandler.SelectorTimeoutError('waitForExist', selector, checkType);
-                    return cb(err, false);
+                    return cb(err);
                 }
 
                 return cb();
--- a/lib/commands/waitForExist.js
+++ b/lib/commands/waitForExist.js
@@ -65,7 +65,9 @@ module.exports = function waitForExist(selector, ms, reverse) {
 
                 var now = new Date().getTime();
                 if(now - start > ms) {
-                    return cb(null, false);
+                    var checkType = reverse ? 'not in the DOM' : 'in the DOM';
+                    var err = ErrorHandler.SelectorTimeoutError('waitForExist', selector, checkType);
+                    return cb(err, false);
                 }
 
                 return cb();
--- a/lib/commands/waitForVisible.js
+++ b/lib/commands/waitForVisible.js
@@ -62,7 +62,11 @@ module.exports = function waitForVisible(selector, ms, reverse) {
             cb();
         }
     ], function(err) {
-
+        // Check if error is because of script timeout
+        if (err && err.isScriptTimeoutError && err.isScriptTimeoutError()) {
+            var checkType = reverse ? 'invisible' : 'visible';
+            err = ErrorHandler.SelectorTimeoutError('waitForVisible', selector, checkType);
+        }
         callback(err, response.selectorExecuteAsync && response.selectorExecuteAsync.executeAsync ? response.selectorExecuteAsync.executeAsync.value : false, response);
 
     });
--- a/lib/utils/ErrorHandler.js
+++ b/lib/utils/ErrorHandler.js
@@ -33,6 +33,7 @@ function ErrorHandler(type, msg) {
         this.message = [];
 
         if(seleniumStack.message && seleniumStack.type && seleniumStack.status) {
+            this.status = seleniumStack.status;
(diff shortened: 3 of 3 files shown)


christian-bromann:
thanks! I will apply that Error for all other waitFor commands to.

lib/commands/waitForExist.js
```diff
@@ -65,7 +65,9 @@ module.exports = function waitForExist(selector, ms, reverse) {
 
                 var now = new Date().getTime();
                 if(now - start > ms) {
-                    return cb(null, false);
```
christian-bromann: I need to redo this again. Think about: if the element wasn't found then it doesn't exist and that's the reason why we are waiting, so we should return true or false:

```js
if(now - start > ms) {
    return cb(null, reverse);
}

**Comment by christian-bromann** [comment]
I've found a more convenient way to make that error messages more expressive. It now returns:

```
RuntimeError: RuntimeError
     (ScriptTimeout:28) A script did not complete before its timeout expired.
     Problem: Timed out waiting for asyncrhonous script result after 206 ms

     Callstack:
     -> executeAsync([Function])
     -> selectorExecuteAsync("#notExists",[Function],false)
     -> waitForVisible("#notExists",200)
```

File: lib/commands/waitForExist.js (hunk 1 above)
Comment (notnarb) [inline lib/commands/waitForExist.js]: I'm not sure I understand.  If you are waiting for an element to not exist (reverse = true), then it would be caught by the (reverse && !isExistent) condition above:
```javascript
  var isExistent = res.value.length > 0;
  if ((!reverse && isExistent) || (reverse && !isExistent)) {
    return cb(null, true);
  }
```
If you are expecting an element to eventually exist, I would think you would want an error to be thrown if it times out.  I'm not sure under what condition this function should time out and be successful.

---
My understanding of this function is that you either are expecting something to exist (reverse=false) or something to not exist (reverse=true) and
* If the condition specified is matched before the timeout, resolve
* If the condition specified is not matched before the timeout, throw an error

christian-bromann: yes, sorry, that makes sense, thanks!
But it should still be:

```js
if(now - start > ms) {
    return cb(null, false);
}
```

I assume that people expect a boolean value to assert against not an error.

notnarb: If you wanted an assertion, I would assume you would want to assert whether or not the command succeeded
```javascript
var assert = require('chai').assert;
client.waitForExist('.textbox', function (error, result) {
    assert.notOk(error)
})
```
just like you would do on any other command that you expect might fail.   

The original issue #419 is that waitForExist doesn't behave like the other waitForX commands because it doesn't break the chain if the element is not found
```javascript
client.waitForExist('.textbox')
      .click('.textbox')
```
To quote you on this issue: 
> You're right, it's not completely clear in our docs. According to similar commands in other Selenium libraries and other commands in this project, all waitFor commands should break the chain if no callback is given and return value would be false. I will mark this as enhancement and will tackle it once I got time. Any PRs are welcome of course.

.

> all waitFor commands should break the chain if no callback is given and return value would be false

And I agree with the idea of not having to provide a custom callback for all waitForX commands that might fail.

christian-bromann: dang, yeah totally right. `waitForExist` definitely should throw if no element was found (or an element was found in reverse mode. With my [recent commit](https://github.com/webdriverio/webdriverio/commit/7bca59b165b24e8fde3b7c40988e679380ede489) this should work. Thanks for clarifying.

notnarb: Great, thanks!

PR #446, opened 2015-03-02
The `chainit` module uses `setImmediate` function.

TheBits:
It is available starting from node 0.10.0.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   ],
   "main": "./index.js",
   "engines": {
-    "node": ">= 0.8.0"
+    "node": ">= 0.10.0"
   },
   "scripts": {
     "test": "node ./test/runner.js",


christian-bromann:
@TheBits can't see `setImmediate` anywhere in the [chainit](https://github.com/vvo/chainit) codebase as well as in the [queue](https://github.com/jessetane/queue/tree/1.0.2) codebase

TheBits:
Did you see `chainit-3.0` branch?
https://github.com/vvo/chainit/blob/93ff39c9e13561aac7c5f6631b8b9d13a44958d9/index.js#L19

christian-bromann:
@TheBits I've seen that there is some effort going on to improve chainit. Once this is done I need to check if this version works with WebdriverIO as it is a critical part of this project. Could be that we still continue to use `v2.1.1`.

PR #460, opened 2015-03-13
Q storing all unhandled rejections leads to major memory leaks when webd...

hippich:
...riverio runs for a long time. This change should stop this behavior.

This is mostly debug behavior - https://github.com/kriskowal/q/wiki/API-Reference#qstopunhandledrejectiontracking


christian-bromann:
Thanks!

PR #466, opened 2015-03-18
Fixed issue with moveToObject

Robinfr:
Fixed an issue with moveToObject where the mouse cursor was not being moved to the center by default (without providing the offset). Previously it was non-compliant with the specs of Webdriver.io(http://webdriver.io/api/action/moveToObject.html) in the API docs, where it says that if no offset is specified the default is the center.

### DIFF
--- a/lib/commands/moveToObject.js
+++ b/lib/commands/moveToObject.js
@@ -98,7 +98,12 @@ module.exports = function moveToObject(selector, xoffset, yoffset) {
             },
             function(res, cb) {
                 response.element = res;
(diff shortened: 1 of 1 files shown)


Robinfr:
Is there any reason this is not being pulled?

christian-bromann:
time @Robinfr ;-)

Robinfr:
I understand you're busy, however if the website mentions 3 business days (http://webdriver.io/contribute.html) and instead it takes half a month.. ;)

PR #483, opened 2015-04-01
Modify description in the docs

Emigre:
[waitForVisible](http://webdriver.io/api/utility/waitForVisible.html)'s description in the docs is slightly misleading. It says that it checks for the visibility and x/y coordinates but it also checks that display is not 'hidden' and opacity is not zero.

I was a little bit confused by it today so I added specificity to the description. I think it's helpful because it's not entirely obvious that it's going to work in that way: [isVisible](http://webdriver.io/api/state/isVisible.html) for example behaves in a different way, considering an element is "visible" even when its opacity is zero.

### DIFF
--- a/lib/commands/waitForVisible.js
+++ b/lib/commands/waitForVisible.js
@@ -5,8 +5,10 @@
  * selector, it returns true (or false if reverse flag is set) if at least one
  * element is visible.
  *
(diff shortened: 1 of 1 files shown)


christian-bromann:
You're right, thanks!

PR #499, opened 2015-04-10
First usable(-ish) working v3.

Robinfr:
Fixed several protocols. Also updated the webdriverio.local example and the selectors test. Also wrote a new example: webdriverio.elements which shows how to use the elements protocol. For now I've added a TODO there since not the whole example is working yet.

### DIFF
--- a/examples/webdriverio.elements.js
+++ b/examples/webdriverio.elements.js
@@ -1,0 +1,39 @@
+var webdriverio = require('../index'),
+    Q = require('q');
+
+var options = {
+    desiredCapabilities: {
+        browserName: 'chrome'
+    }
+};
+
+webdriverio
+    .remote(options)
+    .init()
+    .url('http://www.google.com')
+    .elements('input[type="submit"]')
+    .then(function (elements) {
+        elements = elements.value;
+
+        var promises = [];
+
+        elements.forEach(function (element) {
+            element = element.ELEMENT;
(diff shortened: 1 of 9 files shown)


christian-bromann:
Great work. Thanks! Some thoughts:

1. How about making the example a bit shorter and use chained thens like this:

```js
webdriverio
    .remote(options)
    .init()
    .url('http://www.google.com')
    .elements('input')
    .then(function (elements) {
        var self = this,
            promises = [];

        elements.value.forEach(function (element) {
            promises.push(self.elementIdAttribute(element.ELEMENT, 'type').then(function(res) {
                return res.value;
            }));
        });

        return Q.all(promises);
    }).then(function(inputTypes) {
        console.log(inputTypes);
    })
    .end()
    .then(function(){
        //TODO: get the promise after .end() working
        console.log('Client ended');
    });
```

2. If the query selector is `input` we get a bunch of elements from type `hidden` back but executing the script with `input[type="submit"]` also returns 2 elements, so something weird is going on here

Robinfr:
@christian-bromann no it's perfectly fine. ``input[type="submit"]`` returns the two buttons below the search bar. The other ``input`` items are hidden items.

About the example, yes that looks nice. I would add a few more enters though so the lines don't get too long.

christian-bromann:
Perfect! LGTM

PR #508, opened 2015-04-25
Link to example Gulp + Mocha + selenium-standalone

silvenon:
If you think this could be useful, I've written a post how to integrate WebdriverIO with other tools. Mainly, with selenium-standalone, which allows for completely automatic test running and no dependencies which need to be installed separately (like Selenium server, drivers etc.)

I would also really appreciate a review (I'd give credits at the bottom).

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,8 @@ describe('example page', function() {
 });
 ```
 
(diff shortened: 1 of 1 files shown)


christian-bromann:
I am happy to put that into the readme .. also I will link your article on webdriver.io once we get `v3` published

silvenon:
Great, thanks :+1:

Issue #513, opened 2015-04-30
Need support for positional CSS selectors (:first-child, :first-of-type, :nth-child etc)


vdineva:
Currently WebdriverIO does not support a positional CSS selector which follows an id selector, e.g. `'#someId:first-child'`.

Suppose I need to get to the `#purplebox` div only if it is a first-child of its parent.
In a browser `document.querySelector('#purplebox:first-child')` returns 
1. either the element (if it's a first child of it's parent element) 
```html
<div>
    <div id="purplebox"></div>
</div>
```
OR 
2. null, if #purplebox is not the first child of it's parent element
```html
<div>
    <div class="box black"></div>
    <div id="purplebox"></div>
</div>
````

Currently WebdriverIO throws and uncaughtException for when trying to locate `'#purplebox:first-child'` with the second html structure

Robinfr:
``:first`` is also a jQuery selector actually. You could try using ``first-child`` or ``nth-child``.

christian-bromann:
As I don't think it is a big effort to transform the xPath for those pseudo selectors I guess we can add that for the `v3` milestone

vdineva:
@Robinfr : yes `:first` is not a CSS specification selector but a jQuery syntactic sugar. I've updated the issue. However currently `:first-child` or `nth-child` do not work

christian-bromann:
See #514

PR #514, opened 2015-05-01
Fix for issue #513: Need support for positional CSS selectors (:first-child, :first-of-type, :nth-child etc)

vdineva:
Updated the regular expression for find element strategy

### DIFF
--- a/lib/helpers/find-element-strategy.js
+++ b/lib/helpers/find-element-strategy.js
@@ -20,7 +20,7 @@ module.exports = function findStrategy(args) {
 
     // check value type
     // use id strategy if value starts with # and doesnt contain any other CSS selector-relevant character
-    if (value.indexOf('#') === 0 && value.search(/(\s|>|\.|[|])/) === -1) {
+    if (value.indexOf('#') === 0 && value.search(/(\s|>|\.|\:|[|])/) === -1) {
         using = 'id';
         value = value.slice(1);
 
--- a/test/spec/functional/selectorExecute.js
+++ b/test/spec/functional/selectorExecute.js
@@ -127,4 +127,36 @@ describe('selectorExecute', function() {
             });
     });
 
+    it('should be able to resolve a :first-child css selector', function(done) {
+        this.client
+            .selectorExecute('form input:first-child', function(arr) {
+                    return arr[0].getAttribute('value');
+                }, function(err, res) {
+                    assert.ifError(err);
+                    assert.equal("a", res);
+                    done(err);
+                });
+    });
+
+    it('should be able to resolve a :first-of-type css selector', function(done) {
+        this.client
+            .selectorExecute('.page a:first-of-type', function(arr) {
+                    return arr[0].innerText;
+                }, function(err, res) {
+                    assert.ifError(err);
+                    assert.equal("two", res);
+                    done(err);
+                });
+    });
+
+    it('should be able to resolve a :nth-child() css selector', function(done) {this.client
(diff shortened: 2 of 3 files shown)


christian-bromann:
@vdineva thanks for your contribution.

Looking at your changes makes me feel that you've forgot to push something. The [changes](https://github.com/webdriverio/webdriverio/pull/514/files#diff-f3e68f1ab6fc51142d872b0b32cb899dR23) you made only apply for selectors with an `#` at the beginning. Did you touch any other selector regex without pushing them?

vdineva:
@christian-bromann Yes, I've only updated the regex for selectors that starts with `#`. I noticed that actually most of the time positional selectors work just fine and after some testing I noticed the pattern. Currently a positional element such as `:first-child` fails only after being used after a selector for id. All other cases (positional css selector after an html element e.g. `body :first-child`, or after a class name e.g.`.someClass :first-child`, or on it's own `nth-child(1)`) work fine as they fall though all the `else if` statements and end up at the final `else`, so basically using 'css selector' strategy which later on calls `document.querySelectorAll()`. So basically the only update needed was the regex for selectors with an # at the beginning.

christian-bromann:
@vdineva gotcha! However I don't think it make sense to "fix" that as it wouldn't be an invalid selector since there should only be one single id per element. Therefor something like `#myId:first-Child` isn't valid as there should only be one element with that id.

Is there an use case I've forgot?

vdineva:
@christian-bromann : I might have created some confusion, for which I apologize.

As per Mozilla Developer Network (https://developer.mozilla.org/en-US/docs/Web/CSS/:first-child) : "The :first-child CSS pseudo-class represents any element that is the first child element of its parent."
So `#myId:first-Child` is a valid css selector which will return 
1. either the element (if it's a first child of it's parent element)
```html
<div>
    <div id="myId"></div>
</div>
```
OR
2. null, if #myId is not the first child of it's parent element
```html
<div>
    <div class="box black"></div>
    <div id="myId"></div>
</div>
```

christian-bromann:
@vdineva your right but querying elements using an `id` will always return one element as there is only one element allowed on the page using that id. Meaning: `#myId:first-Child` is unnecessary as `#myId` will do the same trick. Does that make sense? The use of pseudo selectors on elements with an id as prefix doesn't make any sense ;-)

christian-bromann:
With classes it is different. There could be one class applied to multiple elements. The use of pseudo selector to fetch the first/last or nth child makes sense as it makes the query more specific

christian-bromann:
ping @vdineva

vdineva:
@christian-bromann I don't quite agree here. Given the following html snippet, in plain vanilla javasctipt
`document.querySelector('#myId:first-child')` will return the element as it is a first child of it's parent element. (note the `:first-child` selector returns an element that is the first child of its parent )
```html
<div>
    <div id="myId"></div>
</div>
```
With the following html plain javasctipt `document.querySelector('#myId:first-child')` returns `null` as the element with id `myId` is not the first child of its parent
```html
<div>
    <div class="box black"></div>
    <div id="myId"></div>
</div>
```
And since under the hood WebdriverIO uses the javascript API I thought it should mimic it's behavior as closely to it as possible.

I came up with a workaround for my project and am closing this pull request for now.

christian-bromann:
Ahh I see. Well if someone queries `#myId:first-child` he should expect to get an element with an id `#myId` no matter if it is the first or last child as there is only _one_ single element with that id on the page.  WebdriverIO relies on Selenium when it comes to query elements on the page.

Still, thanks for contributing to the project!

Issue #510, opened 2015-04-27
Transfer promiseness takes only the 1st element of result array


silvenon:
```html
<!-- index.html -->

<!-- ... -->
<div class="foo">Item 1</div>
<div class="foo">Item 2</div>
<div class="foo">Item 3</div>
<!-- ... -->
```

This works fine:

```js
// test.js

var assert = require('assert');
var client = require('webdriverio').remote({
  desiredCapabilities: {
    browserName: 'phantomjs'
  }
});

client.init().url('http://localhost:9000/index.html');
client.getText('.foo').then(function(text) {
  assert.deepEqual(text, ['Item 1', 'Item 2', 'Item 3']);
});
client.end();
```

But this compares only the 1st element of the array:

```js
// test.js

var chai = require('chai');
var chaiAsPromised = require('chai-as-promised');
var client = require('webdriverio').remote({
  desiredCapabilities: {
    browserName: 'phantomjs'
  }
});

chai.use(chaiAsPromised);
chai.should();
chaiAsPromised.transferPromiseness = client.transferPromiseness;

client.init().url('http://localhost:9000/index.html');
client.getText('.foo').should.become(['Item 1', 'Item 2', 'Item 3']);
// Uncaught AssertionError: expected 'Item 1' to deeply equal [ 'Item 1', 'Item 2', 'Item 3' ]
client.end();
```

Not sure if this is an issue in WebdriverIO or chai-as-promised, but haven't found anything on Google.

Robinfr:
Yes, you're right. There are massive improvements being done using promises in the v3 branch that will be hopefully be released soon, and this is definitely something that should be a use-case.

christian-bromann:
Todo here:
- make sure example above works (maybe include chai-as-promised to functional test suite)
- implement `transferPromiseness` method
- document it on website

bryan-coursera:
One more thing here to note: on v3, the `client` prototype extends from `null`, which means that when `chai.should()` adds the `should` handler to `Object.prototype`, it doesn't get added to webdriverio clients. 

I've been able to work around this in v3 by changing

```var prototype = Object.create(null)``` to  ```var prototype = Object.create(Object.prototype)```

I'm not sure if this is the best approach, but it might be necessary to continue using `chai.should`

christian-bromann:
good catch @bryan-coursera .. I will fix that

christian-bromann:
Thanks again @bryan-coursera, awesome stuff!!
[...]

PR #522, opened 2015-05-07
Add transferPromiseness support, and write test cases for use of chai-as-promised (#510)

bryan-coursera:
- Adds a `transferPromiseness` function to webdriverio.js for chai-as-promised support
- Extend the webdriverio client from `Object.prototype` to support `chai.should`
- Ådds test cases for chai-as-promised usage

@christian-bromann -- let me know if this looks ok, or if there's anything that I'm missing. thanks!

### DIFF
--- a/test/spec/functional/chaiPromises.js
+++ b/test/spec/functional/chaiPromises.js
@@ -1,0 +1,62 @@
+var Q = require('q');
+
+describe('chai-as-promised', function() {
+
+    /**
+     *  Remove the `should` global on Object.prototype to allow chai.should for these tests,
+     *  and set up chai, chai-as-promised and chai.should properly
+     */
+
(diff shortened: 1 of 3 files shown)


test/spec/functional/chaiPromises.js
```diff
@@ -0,0 +1,62 @@
+var Q = require('q');
```
christian-bromann: Seems that we don't need the `Q` library here, right?

christian-bromann:
@bryan-coursera awesome thanks! One little thing besides that I am happy to merge it!

bryan-coursera:
@christian-bromann good catch -- just removed that require. thanks for the super quick review!

christian-bromann:
Thanks to you Sir!

PR #592, opened 2015-06-30
Update debug example

Sanjo:
Renames debugger to debug


christian-bromann:
oh yes, good catch, thanks!

PR #593, opened 2015-06-30
desiredCapabilities parameter of init is still optional

Sanjo:


christian-bromann:
actually yeah you are right.

PR #596, opened 2015-07-01
Fixes incompatibility with PhantomJS 1.9 Ghostdriver mode (fixes #594)

### LINKED ISSUE #594: Running with PhantomJS Ghostdriver directly no longer works in 3.0.0
Reproduce: https://github.com/Sanjo/phantomjs-webdriver-v3-issue

This works with 2.4.5. If you start the Selenium server additionally, it works. Is this something that you can fix?

Sanjo:
The core problem seems to be that PhantomJS 1.9 has problems with some lowercase headers.
request 2.49.0 is the latest version that works correctly in addition to setting
the Content-Type and Content-Length header with uppercase.

Relevant issues:

* PhantomJS issue: https://github.com/ariya/phantomjs/issues/11421

### DIFF
--- a/lib/utils/RequestHandler.js
+++ b/lib/utils/RequestHandler.js
@@ -7,6 +7,7 @@
 var Q = require('q'),
     url = require('url'),
     request = require('request'),
+    merge = require('deepmerge'),
     errorCodes = require('./errorCodes'),
     ErrorHandler = require('./ErrorHandler'),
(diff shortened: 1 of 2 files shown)


christian-bromann:
Supporting phantomjs 1.9 makes definitely sense even though we need to downgrade request. Thanks for putting this together @Sanjo

christian-bromann:
published `v3.0.1`

PR #598, opened 2015-07-03
Update keys.js

icaaq:
Added link to the different keys as on [addValue](https://github.com/webdriverio/webdriverio/blob/master/lib/commands/addValue.js)


christian-bromann:
Thanks

PR #576, opened 2015-06-14
Grunt

AndreTheHunter:
The big grunt update!
* tests are now far easier to to run by using grunt
* split test/runner.js into setup.js, helper.js and various config in the Gruntfile.js

See the commit logs for more information

Some further work required:
* get `grunt test-mobile` to work
* istanbul code coverage
* Need consistent assertion style. I vote for 'expect' bdd style
* 'use strict' on all modules
* fix ErrorHandler.RuntimeError causing process exit

PR #599, opened 2015-07-03
Grunt

AndreTheHunter:
The big grunt update!
* tests are now far easier to to run by using grunt
* split test/runner.js into setup.js, helper.js and various config in the Gruntfile.js

[See previous pull request](https://github.com/webdriverio/webdriverio/pull/576)

Travis and istanbul are working. See https://travis-ci.org/AndreTheHunter/webdriverio


AndreTheHunter:
[There's a issue](https://github.com/pifantastic/grunt-selenium-server/issues/18) with grunt-selenium-server preventing it from running on Travis. I'm trying to fix it now.

AndreTheHunter:
Fixed the grunt-selenium-server issue. Merge away!

test/tasks/appium.js
```diff
@@ -0,0 +1,41 @@
...
+    function kill(err) {
+        if (err)
+            console.error(err);
+        if (appium)
+            appium.kill('SIGINT');
+    }
+
+    process.on('exit', kill);
+    process.on('SIGINT', kill);
+    process.on('uncaughtException', kill);
```
christian-bromann: How about moving these event listener into the appium task? If they are Appium specific they should rather live there otherwise in the main grunt file

test/helper.js
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+        return function(done) {
+
+            function init() {
+                return WebdriverIO.remote(conf).init();
+            }
+
+            if(options.remoteOptions) {
+                conf = merge(conf, options.remoteOptions);
```
christian-bromann: `merge` is nowhere defined. Let's add `var merge = require('deepmerge');` to the top of the file

christian-bromann:
Just checked your PR and I recognised the following issues:

- the travis build only runs functional tests in each environment
- my intention of separate spec was that I can have specific tests that only get executed in specific environments and also tests that get executed in all environments. Multibrowser tests are a bit special as they have their own test specs and don't share tests between environments. So here is how it should be: in desktop and mobile environments the following tests should get executed: `/test/spec/*.js` and `/test/spec/(mobile|desktop)/*.js`. In functional and multiremote environments only these tests should get executed: `/test/spec/(functional|multiremote)/**/*.js`
- I've seen couple of jshint errors, let's get rid of them too :wink:

test/tasks/appium.js
AndreTheHunter: May as well. Moved and committed.

test/helper.js
AndreTheHunter: Good spot. Done

AndreTheHunter:
Thanks for the feedback and clarification @christian-bromann!
* Getting the travis matrix to run properly seems a bit more involved, but hopefully have it working soon.
* I've been confused on then `/test/spec/*.js` should be included so thank you for clarifying.
* I wasn't running jshint so I'll add it as a grunt watch and test task.

christian-bromann:
@AndreTheHunter looking forward for those changes. Also please update the [docs](https://github.com/webdriverio/webdriverio/blob/master/docs/contribute.md) according to the new grunt setup. Thanks!

christian-bromann:
@AndreTheHunter did you make any progress? Looking forward to merge this

AndreTheHunter:
@christian-bromann I'm having difficulty getting Saucelabs working. Should hopefully have it ready soon

AndreTheHunter:
chromedriver is needed when running the tests locally

christian-bromann:
> chromedriver is needed when running the tests locally

In build all jobs are running on Sauce Labs so there is no need to add Chromedriver as dependency. Tests running locally on Travis are using phantomjs which is totally sufficient for functional tests (selector stuff, chaining tests etc)

AndreTheHunter:
If tests are run locally, they won't be connecting to SauceLabs. How will contributors run those tests? Do we expect everyone to sign up to SauceLabs?

christian-bromann:
> Do we expect everyone to sign up to SauceLabs?

No, of course not.

> How will contributors run those tests?

Now I get your point. Now we explicitly say that the user has to download the standalone server and run it by them self.

I would keep it that way tbh as we can't provide a whole test infrastructure for everyone. It is just too complex and highly depends on the OS of the user. But if adding chromedriver makes it easier to kick of desktop tests then I am fine with having it as devDependency

christian-bromann:
@AndreTheHunter thanks for your contribution. In order to move forward with v4 I had to assign this job to myself. However I was able to copy some stuff from your branch into mine. Thanks!

superseded by https://github.com/webdriverio/webdriverio/compare/v4.0...e2e-into-es2016

PR #612, opened 2015-07-07
it is unclear what values are needed in order to simulate Enter, Down…

yonatan-py:
… and of sort.


christian-bromann:
Is the sentence above:

_You’ll find all supported characters [here](https://code.google.com/p/selenium/wiki/JsonWireProtocol#/session/:sessionId/element/:id/value)_

that unclear?

yonatan-py:
it is not clear that the values one need to input are `Enter`, `Down arrow` and the like and not `U+E007`, `U+E015` and the like.

christian-bromann:
I don't think adding a link to the internal code clarifies it. I've added an additional paragraph. What do you think?

PR #623, opened 2015-07-09
Update doc: update selenium download url

ruanyl:
Selenium 2.45 seems does not work with Firefox 38+ on MacOS. Since people usually copy&paste the command, it might be better to keep it up-to-date.


christian-bromann:
:+1:

Issue #607, opened 2015-07-06
Test runner stops whenever error occurs : "Promise never resolved with an truthy value"


mohdasif2294:
Instead of showing errors on console and continuing pending test cases. The test runner stops abruptly.

AndreTheHunter:
I've noticed this in various tests too. If an error occurs, webdriverio calls process.exit which stops tests from continuing. I think we should remove all calls to process.exit and just throw exceptions.

mohdasif2294:
Yea..right! It happens every time promise on not loaded element triggers (Due to internet problem). Though Assertion errors are properly caught.

christian-bromann:
I see. So the error gets thrown by the [`waitUntil`](https://github.com/webdriverio/webdriverio/blob/master/lib/commands/waitUntil.js#L40) command which gets used in all waitForX commands. So it seems that this behaviour has to do something with the way we handle rejected promises within commands.

aahill50:
I'm not sure if this is exactly the same issue, but it seems to be having the same effect. It seems that errors that occur during part of the chained method calls do not prevent subsequent steps from executing. I was able to replicate this using the following test (mocha):

    it('should do something', function (done) {
            client
                .url('http://www.google.com')
                .click('.not-a-real-element')
                .click('.not-a-real-element-again')
                .call(done)
        });

When I run this test, the first call to ```.click``` throws an error, as I would expect it to. However, it doesn't stop the second ```.click``` call from occurring, which throws another error, which then crashes the entire test suite.

I'm assuming this is not the desired behavior? I would think the first error should prevent the remaining steps from being called.

christian-bromann:
@aahill50 interesting, could you check if a third click gets executed?

aahill50:
@christian-bromann Honestly, it's hard to tell exactly what's going on with more than 2 calls. After playing around with it a bit, I'm actually not 100% sure that additional calls to ```click``` are occurring, because I can't get any output from them in promise handling callbacks. Something strange is happening though, because if you take out the 2nd call to ```.click``` or replace the element with a valid one so no error would be thrown, it works as expected (The test fails, but any remaining tests are still executed normally).

aahill50:
@christian-bromann I think something is going on with an uncaught exception not being handled (maybe through Q?), but I'm having trouble nailing it down.

aahill50:
[...]

PR #624, opened 2015-07-09
Fix issue with throwOnNextTick getting called too early

aahill50:
This seems to fix the issue mentioned in #607. The [throwOnNextTick](https://github.com/webdriverio/webdriverio/blob/master/lib/webdriverio.js#L121) function was getting returned while ```saveScreenshot``` was still processing, and I believe this will prevent that from occurring.

### DIFF
--- a/lib/webdriverio.js
+++ b/lib/webdriverio.js
@@ -114,7 +114,7 @@ module.exports = function WebdriverIO(args, modifier) {
                 'ERROR_' + sanitize.caps(desiredCapabilities) + '_' + new Date().toJSON() + '.png'
             )]);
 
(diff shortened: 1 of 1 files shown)


christian-bromann:
Hi @aahill50 ,
thanks for your effort. I checked out your changes and they seems to not fix the problem. I looked into it and found a different root cause. First of all I made an error when trying to support promises in jasmine. To end a test spec with an error you have to call `done.fail(e)` to do so. I also removed the `process.exit` statement. Here are the changes I've made:

https://github.com/webdriverio/webdriverio/compare/85f6a86563669f5a6d2ea8552a05dc60499f6b06...431072883d0f4585eba981bcdfa78244eee11dfa

Thanks again

PR #635, opened 2015-07-16
Allow to ignore undefined cucumber steps

vslinko:
Things to discuss:

1. Should it be disabled by default like before in wdio or it should be enabled like in original cucumber?

### DIFF
--- a/lib/frameworks/cucumber.js
+++ b/lib/frameworks/cucumber.js
@@ -1,11 +1,12 @@
 var q = require('q'),
     getFilePaths = require('../utils/ConfigParser').getFilePaths;
 
-var CucumberReporter = function(BaseListener, capabilities) {
+var CucumberReporter = function(BaseListener, options) {
     var self = this;
 
     this.listener = BaseListener;
-    this.capabilities = capabilities;
+    this.capabilities = options.capabilities;
+    this.options = options;
     this.failedCount = 0;
 
     Object.keys(CucumberReporter.prototype).forEach(function(fnName) {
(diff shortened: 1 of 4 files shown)


christian-bromann:
@vslinko thanks for the PR!
Could you document the new option here:
https://github.com/webdriverio/webdriverio/blob/master/examples/wdio.conf.js
https://github.com/webdriverio/webdriverio/blob/master/docs/guide/testrunner/configurationfile.md
https://github.com/webdriverio/webdriverio/blob/master/lib/helpers/wdio.conf.ejs

After that I think it is good to go. I guess it should be disabled by default. Tests should fail if a step isn't defined. There are probably other opinions on that though.

lib/frameworks/cucumber.js
```diff
@@ -185,12 +196,17 @@ module.exports.run = function(config, specs, capabilities) {
...
+            reporterOptions.ignoreUndefinedDefinitions =
+                config.cucumberOpts.ignoreUndefinedDefinitions;
+        }
     }
 
     global.cucumber = Cucumber.Cli(execOptions);
     cucumberConf = Cucumber.Cli.Configuration(execOptions);
     runtime = Cucumber.Runtime(cucumberConf);
-    reporter = new CucumberReporter(Cucumber.Listener(), capabilities);
+    reporter = new CucumberReporter(Cucumber.Listener(), capabilities, reporterOptions);
```
christian-bromann: let's simplify this and pass `config` instead of capabilities to the CucumberReporter. Then we can get rid of that if statement too.

lib/frameworks/cucumber.js
```diff
@@ -1,11 +1,12 @@
 var q = require('q'),
     getFilePaths = require('../utils/ConfigParser').getFilePaths;
 
-var CucumberReporter = function(BaseListener, capabilities) {
+var CucumberReporter = function(BaseListener, capabilities, options) {
     var self = this;
 
     this.listener = BaseListener;
     this.capabilities = capabilities;
```
christian-bromann: this should be `this.capabilities = config.capabilities;` then

vslinko:
I added docs.

christian-bromann:
I guess we can remove `It means that wdio will not exit with error code.`
otherwise looks good :+1:

vslinko:
Updated.

christian-bromann:
Sweet! Thx

christian-bromann:
I've seen that even though I had `ignoreUndefinedDefinitions` was set as true the reporter still marked that step as failed. I think this would confuse people so I will report that step as pending then.

vslinko:
I think it should be different status. I suggest `undefined step` as name and yellow as text color.

christian-bromann:

Issue #629, opened 2015-07-13
The client.on('error') triggers up the promise chain.


sean-hill:
When `client.on('error')` is called I get multiple errors because it triggers up the promise chain.

sean-hill:
Just curious when this bug fix would be released :smile: It's kinda affecting my process big time...

christian-bromann:
not sure, I am on it

sean-hill:
No worries, thanks @christian-bromann!

christian-bromann:
This will be fixed with the next release. As you've seen I improved the error handling. I hope this will be fixed with my recent path.

christian-bromann:
Fixed and published with `v3.2.0`

sean-hill:
:+1: THANK YOU!!!

sean-hill:
Hey @christian-bromann, not to be a downer or anything, but with the latest `3.2.0` release, `client.on('error')` does not fire if an element can not be found. The promise chain just stops.

sean-hill:
Hey @christian-bromann I'm still having a little issue with the `client.on('error')` event. Some errors return `e.message` and `e.stack`. For instance the `NoSuchElement` error. Other others return `e.body.value.class` and `e.body.value.message` such as `org.openqa.selenium.NoAlertPresentException`. It'd be nice if there was some consistency with this. Currently I'm having to this:

	var error = e.message || e.body.value.class;
	var stack = e.stack || e.body.value.message;

Thanks!

christian-bromann:
Agree! Created issue #747

PR #661, opened 2015-07-23
Fixed infinite recursion in setViewportSize command

pazone:
Sometimes we can not get desired window size due to screen size limits or etc. In this case setViewportSize was entering infinite recursion because of maxExecutionTime was always 5.

### DIFF
--- a/lib/commands/setViewportSize.js
+++ b/lib/commands/setViewportSize.js
@@ -5,13 +5,13 @@
  * window size will always be bigger since it includes the height of any menu or status bars.
  *
  * <example>
-    :setViewportSize.js
-    client
-        .setViewportSize({
+ :setViewportSize.js
+ client
+ .setViewportSize({
             width: 500,
             height: 500
         })
-        .windowHandleSize(function(err, res) {
+ .windowHandleSize(function(err, res) {
             console.log(res.value); // outputs: "{ width: 500, height: 602 }"
         })
  * </example>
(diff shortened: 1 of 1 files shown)


lib/commands/setViewportSize.js
```diff
@@ -5,13 +5,13 @@
...
-    client
-        .setViewportSize({
+ :setViewportSize.js
+ client
+ .setViewportSize({
             width: 500,
             height: 500
         })
-        .windowHandleSize(function(err, res) {
+ .windowHandleSize(function(err, res) {
```
christian-bromann: can we keep the indention

christian-bromann:
@pazone awesome, thanks for the PR. One minor comment besides that looks good to me. Do yo have an example where setViewportSize is entering in an infinite recursion? Do you think a test would be valuable here?

pazone:
setViewportSize was entering infinite recursion in case when viewport can not be the desired size. Little example: My screen size is 1400x900. And ```setViewportSize({width: 1500, height: 1000})``` will enter in an infinite recursion, because 1400x900 is maximal window size. Look at the old code snippet: 
```javascript
module.exports = function setViewportSize(size) {
//...
var maxExecutionTime = 5; //maxExecutionTime is always === 5
//...
if(--maxExecutionTime && .../*false condition*/) { // maxExecutionTime is decremented, but will be set to 5 when we will enter the function again
    return this.setViewportSize(size); //
}
```
I would like to make unit tests for this command in next PR if it is possible, because I thing, it is still not according to the documentation(no boolean parameter 'type').

sean-hill:
Hey I ran into this also I called `setViewportSize` right after the `init` of the client. Has this been resolved?

christian-bromann:
@sean-hill not yet
@pazone could you fix that indentation thing so I can merge the PR?

pazone:
Of course. I can fix it in next PR of in this PR. You can merge this PR and problem with infinite recursion will be resolved

christian-bromann:
> I can add type parameter it in next PR of in this PR

Don't know what you mean by that :-/
Merging PR and fix indentation change by myself.

Thanks for the PR!

christian-bromann:
I will publish this as soon as I've fixed #629

PR #669, opened 2015-07-28
setViewpoertSize: added type parameter according to documentation

pazone:
According to documentation, ```setViewpoertSize``` should have the boolean parameter ```type```. When ```type``` is set to ```false```, ```setViewpoertSize``` will change window(not viewport) size without indentation and retries.
@christian-bromann did you mean this indentation problem?

### DIFF
--- a/lib/commands/setViewportSize.js
+++ b/lib/commands/setViewportSize.js
@@ -8,12 +8,12 @@
  :setViewportSize.js
  client
  .setViewportSize({
-            width: 500,
-            height: 500
-        })
+    width: 500,
+    height: 500
+ })
  .windowHandleSize(function(err, res) {
-            console.log(res.value); // outputs: "{ width: 500, height: 602 }"
-        })
+    console.log(res.value); // outputs: "{ width: 500, height: 602 }"
+ })
  * </example>
  *
  * @param {Object}   size  window width/height
@@ -24,11 +24,10 @@
  *
  */
 
-
 var getViewportSizeHelper = require('../helpers/_getViewportSize'),
     ErrorHandler = require('../utils/ErrorHandler.js');
 
-module.exports = function setViewportSize(size) {
+module.exports = function setViewportSize(size, type) {
 
     /**
      * parameter check
@@ -37,16 +36,19 @@ module.exports = function setViewportSize(size) {
         throw new ErrorHandler.CommandError('number or type of arguments don\'t agree with setViewportSize command');
     }
 
-    return setSize(this);
+    var shouldIndent = typeof(type) === 'undefined' ? true : type;
+
+    return shouldIndent ? setViewportSize(this) : this.windowHandleSize(size);
 
     /**
      * to set viewport size properly we need to execute the process multiple times
      * since the difference between the inner and outer size changes when browser
      * switch between fullscreen modes or visibility of scrollbar
      */
-    function setSize(client, retryNo) {
+    function setViewportSize(client, retryNo) {
 
         const MAX_TRIES = 5;
+
         if (typeof retryNo === 'undefined') retryNo = 0;
 
         /**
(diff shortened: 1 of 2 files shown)


lib/commands/setViewportSize.js
```diff
@@ -37,16 +36,19 @@ module.exports = function setViewportSize(size) {
         throw new ErrorHandler.CommandError('number or type of arguments don\'t agree with setViewportSize command');
     }
 
-    return setSize(this);
+    var shouldIndent = typeof(type) === 'undefined' ? true : type;
```
christian-bromann: According to our docs the default value of type is false (apply windowHandleSize), so we should switch `true` and `type`.

lib/commands/setViewportSize.js
```diff
@@ -24,11 +24,10 @@
  *
```
christian-bromann: what I meant if indentation is here ^^^ could you also fix this (see line 8 - 14)

test/spec/desktop/viewportSize.js
```diff
@@ -21,4 +21,14 @@ describe('setViewportSize/getViewportSize', function() {
         });
     });
 
+    it('should set window size exactly when parameter \'type\' is false', function() {
+        return this.client.setViewportSize({
+            width: 500,
+            height: 500
+        }, false).windowHandleSize().then(function(res) {
```
christian-bromann: we already have the test above, type should be `true` and we should make sure that the window size was changed not the viewport size

lib/commands/setViewportSize.js
pazone: ok

lib/commands/setViewportSize.js
pazone: As I can see in the doc [here](https://github.com/webdriverio/webdriverio/blob/master/lib/commands/setViewportSize.js#L20)
> param {Boolean}  type  set to `false` to change window size, `true` (default) to change viewport size

Looks like the default value of type is true. Should I edit this doc and change behaviour of type parameter?

christian-bromann: I would like to have `false/undefined` as default so the default action is `windowHandleSize`. If someone set type to `true` it changes the viewport size

test/spec/desktop/viewportSize.js
christian-bromann: scratch that above .. we should test if it changes the window size if no type parameter was assigned and viewport size if type parameter is true

christian-bromann:
Sorry about the confusion about the default value of type

lib/commands/setViewportSize.js
pazone: Ok, it's good idea

christian-bromann:
@pazone I've just seen that the default action was to change the viewport size and __not__ to change the window size. I think we should keep it that way. You're right!

lib/commands/setViewportSize.js
christian-bromann: @pazone please see the last comment (sorry)

pazone:
So, I've done it anyway. The both decisions looks fine. I can remove last commit, if we want to change viewport(not window) size as default.

christian-bromann:
@pazone mhm, what do you thing is best as default value?

pazone:
Actually ```setViewportSize``` is operating with window size and opportunity to change viewport size is looking as an additional feature. But also we have ```windowHandleSize``` to work with browser window. I thing these commands are little bit overloaded by parameters. I hope we will able to find the way to make a little redesign of these commands later. ```setViewportSize``` should change viewport size by default

lib/commands/setViewportSize.js
```diff
@@ -8,12 +8,12 @@
...
-            width: 500,
-            height: 500
-        })
+    width: 500,
+    height: 500
+ })
  .windowHandleSize(function(err, res) {
-            console.log(res.value); // outputs: "{ width: 500, height: 602 }"
-        })
+    console.log(res.value); // outputs: "{ width: 500, height: 602 }"
```
christian-bromann: should be `console.log(res.value); // outputs: "{ width: 500, height: 500 }"` then

christian-bromann:
you know what, I will merge this before confuse you even more, I don't know why I was so confused about the default value. Of course: `setViewportSize` should change the viewport per default. Thanks a lot for that fix!

christian-bromann:

PR #677, opened 2015-07-31
Make it easier for new users to compare reporters

dandv:
If that is too long, we could shorten to

> Which reporter to use? (see http://webdriver.io/guide/testrunner/reporters.html)


christian-bromann:
cool

PR #692, opened 2015-08-05
Use absolute value for window width

amitaibu:
Discussion started in webdriverCSS, until we realized it was a webdriverIO issue -- https://github.com/webdriverio/webdrivercss/issues/98

### DIFF
--- a/lib/protocol/windowHandleSize.js
+++ b/lib/protocol/windowHandleSize.js
@@ -56,8 +56,10 @@ module.exports = function windowHandleSize (windowHandle, size) {
     if(typeof windowHandle === 'object' && windowHandle.width && windowHandle.height) {
         requestOptions.method = 'POST';
         data = {
-            width: windowHandle.width,
-            height: windowHandle.height
+            // The width and height value might return as a negative value, so
+            // we make sure to use its absolute value.
+            width: Math.abs(windowHandle.width),
+            height: Math.abs(windowHandle.height)
         };
     } else if(typeof size === 'object' && size.width && size.height) {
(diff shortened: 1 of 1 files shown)


amitaibu:
@christian-bromann master branch is the `3.x` version. But how can I create a PR for `2.x`?

christian-bromann:
Could you also apply `Math.abs` for the height parameter? After that I would be happy to merge this.

I could create a branch based on the code from the latest 2.x version. But instead of doing this temporary fixes we should put the effort into porting WebdriverCSS to WebdriverIOv3

amitaibu:
> Could you also apply Math.abs for the height parameter?

The reason I didn't do it, is that after many tries I could see only the width as negative. So I prefered not to add unconfirmed code.

> I could create a branch based on the code from the latest 2.x version

That would be handy to have a fix only branch for 2.x

christian-bromann:
> So I prefered not to add unconfirmed code.

well I doesn't hurt to absolute the height as well, you never know :wink: 

> That would be handy to have a fix only branch for 2.x

Let's not fix older versions and concentrate on moving forward with `v3`. This is what most people are waiting for.

amitaibu:
I've added to the height, rebased and pushed. Ready for review:)

christian-bromann:
:+1:

amitaibu:
> Let's not fix older versions and concentrate on moving forward with v3.

Yeah, the only reason I've asked for it is that shoov-webdrivercss now forks webdriverio. I guess I could maintain that fork until v3 lands in webdriverCSS.

PR #695, opened 2015-08-06
Typo fix and list brief difference from Nightwatch and Intern

dandv:


README.md
```diff
@@ -15,6 +15,10 @@ This library is a [webdriver](https://w3c.github.io/webdriver/webdriver-spec.htm
 super easy [Selenium](https://en.wikipedia.org/wiki/Selenium_(software)) tests in your favorite
 BDD/TDD test framework, that will run locally or in the cloud using Sauce Labs, Browsertack or Testingbot.
 
+Unlike integrated testing frameworks (Nightwatch, Intern), WebdriverIO lets you use your favorite test framework,
+such as Cucumber, which the other two don't officially support ([1](https://github.com/nightwatchjs/nightwatch/issues/22#issuecomment-34429738),
+[2](https://github.com/theintern/intern/pull/196)).
```
christian-bromann: I don't think it makes sense to goof on other libraries. They do stuff we don't have and they don't brag about it too

dandv: It's not about bragging; it's about presenting comparison information so users can make the right choice in their situation. For instance, [Intern has a comparison table](https://theintern.github.io/#compare) including QUnit, Mocha, Jasmine, BusterJS, Karma and Nightwatch.

We also have a section higher up in the README about the simpler syntax, which I think is the main point about WebdriverIO. Otherwise, why would a user who's evaluating various WebDriver Node.js clients pick wdio?

Let me see if I can rephrase this in a more neutral manner.

christian-bromann:
:+1: thanks, sound good

PR #697, opened 2015-08-07
There is a bug in the jasmine example

matteonovelli:
The example was taking a screenshot when assertion passes


christian-bromann:
good catch, thanks

PR #699, opened 2015-08-07
waitForXXX bugfix for nonexisting element

### LINKED ISSUE #698: waitForExist, waitForVisible - error isn't thrown when element doesn't exist
When using waitForExist or waitForVisible in 3.x, promise is incorrectly fulfilled with undefined response although element doesn't exist. 

```
var cb, options, webdriverio, client;
webdriverio = require('webdriverio');
options = {desiredCapabilities: {browserName: 'chrome'}};
client = webdriverio.remote(options);

cb = function (success) {
  return function (result) {
    console.log(success, result); // outputs 'success', undefined
    return client.end();
  };
};

client.init()
  .url('http://www.google.com/')
  .waitForExist('div.nonsense')
  .then(cb('success'), cb('fail'));
```

The code above is executed successfully and after timeout, `cb('success')` is called with no parameter - result is undefined.

jaroslav-kubicek:
This proposal should close #698 bug issue. Promise is now rejected if timeout expired and element still doesn't exist.

### DIFF
--- a/lib/commands/waitForExist.js
+++ b/lib/commands/waitForExist.js
@@ -14,8 +14,6 @@
  *
  */
 
-var ErrorHandler = require('../utils/ErrorHandler.js');
-
 module.exports = function waitForExist(selector, ms, reverse) {
 
     /*!
@@ -29,13 +27,7 @@ module.exports = function waitForExist(selector, ms, reverse) {
         reverse = false;
     }
 
-    var start = Date.now();
-
     return this.waitUntil(function() {
-        if (Date.now() > start + ms) {
-            // throw NoSuchElement error if no element was found before timeout
-            throw new ErrorHandler(7);
-        }
         return this.isExisting(selector).then(function(isExisting) {
 
             if(Array.isArray(isExisting)) {
--- a/lib/commands/waitForExist.js
+++ b/lib/commands/waitForExist.js
@@ -14,6 +14,8 @@
  *
  */
 
+var ErrorHandler = require('../utils/ErrorHandler.js');
+
 module.exports = function waitForExist(selector, ms, reverse) {
 
     /*!
@@ -27,7 +29,13 @@ module.exports = function waitForExist(selector, ms, reverse) {
         reverse = false;
     }
 
+    var start = Date.now();
+
     return this.waitUntil(function() {
+        if (Date.now() > start + ms) {
+            // throw NoSuchElement error if no element was found before timeout
+            throw new ErrorHandler(7);
+        }
         return this.isExisting(selector).then(function(isExisting) {
 
             if(Array.isArray(isExisting)) {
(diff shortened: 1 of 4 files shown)


christian-bromann:
I guess the problem here is that we are using catch. The waitUntil command throws if the promise inside doesn't get fulfilled in time. We are catching the error in the catch function to output a better error message. I guess this doesn't get propagated. Could you take a look?

lib/commands/waitForExist.js
```diff
@@ -27,7 +29,13 @@ module.exports = function waitForExist(selector, ms, reverse) {
         reverse = false;
     }
 
+    var start = Date.now();
+
     return this.waitUntil(function() {
+        if (Date.now() > start + ms) {
+            // throw NoSuchElement error if no element was found before timeout
+            throw new ErrorHandler(7);
+        }
```
christian-bromann: waitUntil takes care of this: https://github.com/webdriverio/webdriverio/blob/master/lib/commands/waitUntil.js#L66-L72

lib/commands/waitForVisible.js
```diff
@@ -32,7 +34,13 @@ module.exports = function waitForVisible(selector, ms, reverse) {
         reverse = false;
     }
 
+    var start = Date.now();
+
     return this.waitUntil(function() {
+        if (Date.now() > start + ms) {
+            // throw NoSuchElement error if no element was found before timeout
+            throw new ErrorHandler(7);
+        }
```
christian-bromann: same here

jaroslav-kubicek:
Ok, so I looked deep into it and it's caused by following:

In waitUntil.js, `promise.then(function(res) { ... }` is repeatedly resolved with false and because condition isn't promise, this line is called: `return defer.resolve(self.pause(250).waitUntil(condition, ms, starttime));`. 

After timeout, we are calling this:

```
    if(!timeLeft) {
        return reject(defer);
    }
```

which returns undefined...

jaroslav-kubicek:
It's working when I change it to this:

```
if(!timeLeft) {
        reject(defer);
        return defer.promise;
 }
```

christian-bromann:
:+1: perfect, thanks for looking into this!

Sanjo:
@christian-bromann Can you do a patch release for this (3.1.1) please?

adrai:
+1 but minor not patch, right?
PS. @jaroslav-kubicek well done :beer:

cristian-smocot:
+1

This needs a minor release.
Many people use waitFor... in their scripts.

gary-moon:
+1

This would awesome, since some of my tests were randomly succeeding when they should have been failing.

christian-bromann:

stoeffel:
Awesome work @jaroslav-kubicek :sparkles:

Issue #700, opened 2015-08-07
Improve setViewportSize logic


amitaibu:
Better handle window size, as now we loop for 5 times and hope for the best :)

Continuing [this](https://github.com/webdriverio/webdrivercss/issues/102) issue here.

amitaibu:
We are currently iterating multiple times, trying to reach the size we want. It seems the width is set on the first time, but height isn't (and it may not be set properly). Here's a printout of the process. Note the height isn't set "correctly" in the end (but the question is why do we care about the height)

```
not yet....
{ sessionId: '25610f0d8525c9bf614c2e7853e790de',
  status: 0,
  value: { screenHeight: 635, screenWidth: 960 } }
not yet....
{ sessionId: '25610f0d8525c9bf614c2e7853e790de',
  status: 0,
  value: { screenHeight: 635, screenWidth: 960 } }
not yet....
{ sessionId: '25610f0d8525c9bf614c2e7853e790de',
  status: 0,
  value: { screenHeight: 635, screenWidth: 960 } }
not yet....
{ sessionId: '25610f0d8525c9bf614c2e7853e790de',
  status: 0,
  value: { screenHeight: 635, screenWidth: 960 } }
****************
Result:
{ screenHeight: 635, screenWidth: 960 }
Size:
{ width: 960, height: 707 }
****************
```

How about changing [this](https://github.com/webdriverio/webdriverio/blob/master/lib/commands/setViewportSize.js#L83-L84) part to care only about the width?

```js
if (retryNo < MAX_TRIES && res.value.screenWidth !== size.width) {
```

christian-bromann:
Let's continue the discussion in #701

PR #701, opened 2015-08-07
setViewportSize should repeat action only if width isn't correct

amitaibu:
#700

### DIFF
--- a/lib/commands/setViewportSize.js
+++ b/lib/commands/setViewportSize.js
@@ -78,10 +78,13 @@ module.exports = function setViewportSize(size, type) {
             }).execute(getViewportSizeHelper).then(function (res) {
 
                 /**
-                 * if viewport size not equals desired size, execute process again
+                 * If the viewport's width does not equal the desired width,
+                 * execute the process again.
+                 *
+                 * We care about the width, as it is not always possible to get
+                 * the height correctly.
                  */
-                if (retryNo < MAX_TRIES
-                    && (res.value.screenWidth !== size.width || res.value.screenHeight !== size.height)) {
+                if (retryNo < MAX_TRIES && res.value.screenWidth !== size.width) {
                     return setViewportSize(client, ++retryNo);
                 }
 


amitaibu:
Just to better explain to motivation for this. I have noticed several times that the `width` may be wrong when repeating the `windowHandleSize`

For example, the final result's width was 400, where 640 was requested:

```
Result:
{ screenWidth: 400, screenHeight: 674 }
Size:
{ width: 640, height: 697 }
```

christian-bromann:
@amitaibu I don't quite understand why you've removed the height check, this doesn't really solve the problem, doesn't it?

amitaibu:
> this doesn't really solve the problem, doesn't it?

It actually does. In my cases I've seen the browsers weren't always able to set the `height` value, but always able to set the `width`. So what currently happens with the existing code, is that we try 5 times to set the window size - it may set the height with wrong values, thus fail in the end, as the width somehow gets messed up as-well.

Ever since I've patched this for Shoov I have not received any failing tests.

christian-bromann:
After reading #700 again I know get the problem. Can we do the following: instead of ignoring height could we introduce a flag that allows to ignore height or width? `setViewportSize` should change the height __and__ width of the viewport, if people run into problems because the resolution isn't tall enough the command just keeps resizing until the limit (of 5) is reached and the move forward

amitaibu:
:+1: in vacation now, so I'll change the PR after it :)

christian-bromann:
Nice, have a good time on vacation

christian-bromann:
ping @amitaibu

amitaibu:
Yeah, didn't forget about it - just a bit lazy. I'll tackle it soon :)

philipwalton:
So I just encountered this issue as well. It took me a bit to realize that multiple resizes were happening before I discovered why my tests were flaky.

In my case, I'm resizing the browser and then checking that an action happens before a timeout, but if webdriverio is resizing multiple time, measuring, resizing again, etc., then in many cases the timeout occurrs before the logic runs to check for my action.

* * *
 
Couldn't you measure the difference between the window and the viewport at the original size before resizing, and then assume the browser chrome dimensions aren't going to change?

And I guess at minimum I'd like to see this behavior documented, as it wasn't immediately apparent to me that multiple resizes were happening.

christian-bromann:
> resizing again, etc., then in many cases the timeout occurrs before the logic runs to check for my action.

It reruns the process max 5 times (see [here](https://github.com/webdriverio/webdriverio/blob/master/lib/commands/setViewportSize.js#L30)). We could make this an optional parameter though.

@amitaibu we updated the code base (merged v4 branch). Could you please fetch a fresh copy from the master branch and open another PR? Thanks!

Strongground:
So what happened to this?
Apparently nothing has changed.

christian-bromann:
@Strongground what do you mean by that? According to my last comment it should be fixed by now. If you still see issues feel free to open a new issue.

PR #707, opened 2015-08-11
The last example getText tests wrong string

wcainboundary:
The last example getText tests against the error text instead of the expected text. To match the earlier example, it should be changed from 'wrong status message' to 'Message sent!'.


christian-bromann:
Good catch! :+1:

Issue #590, opened 2015-06-30
set default loggingPrefs for chrome


christian-bromann:
Getting the log buffer in chrome is only possible if you set the right capabilities:

```js
capabilities.loggingPrefs = { "driver": "INFO", "browser": "INFO" }
```

It is useful to set this as default value if browserName is Chrome.
(see also #491)

spatchamatla:
can we also get the browser debug logs as well with the above, when we set **"browser": "DEBUG"?**

christian-bromann:
You mean everything that gets printed via `console.debug`?

spatchamatla:
yes

christian-bromann:
not sure, try it out :wink:

spatchamatla:
got it, i thought, if we can allow which type of logs can be captured using capabilities.loggingPrefs, will be a great addition.

christian-bromann:
well you can enable it in your capabilities already now:

```js
var webdriverio = require('webdriverio');
var client = webdriverio.remote({
    desiredCapabilities: {
        browserName: 'chrome',
        loggingPrefs: { "driver": "INFO", "browser": "INFO" }
    }
});
```
This issue is more about setting those loggingPrefs as default value to make the log command work in chrome.

spatchamatla:
Got it

toddhgardner:
The Selendroid driver that Saucelabs is currently using doesn't appear to support any loggingPrefs being passed. Even with an empty object, it generates this error:

```
io.selendroid.server.model.SelendroidStandaloneDriver createNewTestSession
SEVERE: Error occurred while starting Selendroid session
java.lang.ClassCastException: org.json.JSONObject cannot be cast to org.openqa.selenium.logging.LoggingPreferences
```

Unfortunately, with this change there is no way (AFAIK) to exclude loggingPrefs from being sent, since the deepMerge is smashing the object back onto capabilities. I need a way to exclude it intentionally. Suggestions?

christian-bromann:
@toddhgardner have you tried to pass an empty string or a boolean?

toddhgardner:
@christian-bromann yes, and webdriver.io indeed passes emptystring or false through, neither of which can be case to the LoggingPreferences object. It appears I need a way to intentionally pass nothing.

christian-bromann:
ok, I created an issue (#684) and gonna fix that :wink:

toddhgardner:
[...]

PR #709, opened 2015-08-12
adding explicit check from false loggingPrefs

toddhgardner:
Addressing an issue with Saucelabs Android webdriver not implementing the LoggingPreferences object, passing a JSON object, string, or boolean to it throws an exception.

This allows for a `false` loggingPrefs value to be excluded from sending to Selenium.

https://github.com/webdriverio/webdriverio/issues/590#issuecomment-127332580

### DIFF
--- a/lib/webdriverio.js
+++ b/lib/webdriverio.js
@@ -33,7 +33,8 @@ module.exports = function WebdriverIO(args, modifier) {
         waitforTimeout: 500,
         coloredLogs: true,
         logLevel: 'silent',
-        baseUrl: null
+        baseUrl: null,
+        desiredCapabilities: {}
     }, typeof args !== 'string' ? args : {});
 
     /**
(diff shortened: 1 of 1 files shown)


christian-bromann:
@toddhgardner implemented it in a different way. Explicitly setting it to false to disable loggingPrefs is not really obvious. With my patch loggingPrefs get only set if it is not set by the user.

christian-bromann:

toddhgardner:
@christian-bromann Apologies for my slow reply, I had moved on to some other things.

I totally agree that my solution was non-obvious, however your change doesn't fix the original problem. If *any* value is sent to Saucelabs for loggingPrefs, it throws the error about not being able to cast LoggingPreferences. That includes `null` and `{}`. How do I send no value when a no-value gets the default?

christian-bromann:
I am confused. So with `v3.2.0` WebdriverIO should send no loggingPrefs. Why is this still failing then?

toddhgardner:
How would we get it to send no loggingPrefs? The commit you added earlier had this:

```
if(typeof desiredCapabilities.loggingPrefs === 'undefined') {
     desiredCapabilities.loggingPrefs = {
         browser: 'ALL',
         driver: 'ALL'
     }
}
```

If I understand that correctly, if I do not set a `desiredCapabilities.loggingPrefs`, it will use the default. So how would I intentionally set nothing?

toddhgardner:
@christian-bromann I don't think this is closed. The original problem was that we need to not send the `loggingPrefs` key for Android to work properly. This change does not allow that.

justinwalsh:
@christian-bromann @toddhgardner I agree we need to be able to send no loggingPrefs at all. I would vote for loggingPrefs: false or loggingPrefs:null as the solution. IE 10 doesn't support logging on Browserstack, so I need to be able to disable the default and send nothing instead otherwise the driver throws and exception.

christian-bromann:
ok let's create a new issue, see #865

PR #710, opened 2015-08-12
wdio.conf.ejs: minor correction

angelsanz:
Spec paths are relative to the directory fro which `wdio` was called,
which is not necessarily the location of wdio.conf.js.

On a side note, one might encounter further problems when calling `wdio` from an npm script, as those are run from the project root (you can explicitly `cd` back to where you wanted, in the script). Shall I include a notice for that in the same comment?

### DIFF
--- a/docs/guide/testrunner/configurationfile.md
+++ b/docs/guide/testrunner/configurationfile.md
@@ -44,8 +44,10 @@ exports.config = {
     // ==================
     // Specify Test Files
     // ==================
-    // Define which test specs should run. The pattern is relative to the location of this
-    // file.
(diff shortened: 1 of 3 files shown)


christian-bromann:
Yes, that would be great. Could you also apply these changes here:

https://github.com/webdriverio/webdriverio/blob/master/docs/guide/testrunner/configurationfile.md
https://github.com/webdriverio/webdriverio/blob/master/examples/wdio.conf.js

angelsanz:
Okay, that's done now!

christian-bromann:
Thanks :+1:

angelsanz:
You're welcome :sparkles:

PR #711, opened 2015-08-12
wdio.conf.ejs: framework choices are lowercase

### LINKED ISSUE #691: jasmineNodeOpts is required to be set
If `jasmineNodeOpts` is missing in the config file WebdriverIO throws the following error:
`ERROR: Cannot read property 'expectationResultHandler' of undefined`

__FIX__
Enhance if statement at
https://github.com/webdriverio/webdriverio/blob/master/lib/frameworks/jasmine.js#L94
and check if jasmineNodeOpts is not undefined

/cc @hellis007

angelsanz:
Framework-specific options wouldn't get rendered 
because the check for them was always false.

A possible solution is to use the exact framework names
as declared in https://github.com/webdriverio/webdriverio/blob/master/lib/cli.js#L123

This would fix #691, although in a different manner than proposed there.

### DIFF
--- a/lib/helpers/wdio.conf.ejs
+++ b/lib/helpers/wdio.conf.ejs
@@ -127,7 +127,7 @@ exports.config = {
         outputDir: './'
     },
     <% }
-    if(answers.framework === 'Mocha') { %>
+    if(answers.framework === 'mocha') { %>
     //
     // Options to be passed to Mocha.
     // See the full list at http://mochajs.org/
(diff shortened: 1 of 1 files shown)


christian-bromann:
Merging even though this got obsolete by https://github.com/webdriverio/webdriverio/commit/0227c794f794f65599e8946c7e73b27671b04421. Thanks

PR #752, opened 2015-09-02
Create webdriver.with.appium.chrome

tschuliiia:
How to create a test using an Android device and testing with chrome


christian-bromann:
Thanks :+1:

PR #754, opened 2015-09-02
Update configuration.md

tschuliiia:
seperated examples and explaination

### DIFF
--- a/docs/guide/getstarted/configuration.md
+++ b/docs/guide/getstarted/configuration.md
@@ -38,6 +38,12 @@ tags: ['tag1','tag2'],  // specify some tags (e.g. if you use Sauce Labs)
 name: 'my test'         // set name for test (e.g. if you use Sauce Labs)
 ```
 
+####desiredCapabilities for Appium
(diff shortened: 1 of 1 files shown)


docs/guide/getstarted/configuration.md
```diff
@@ -38,6 +38,12 @@ tags: ['tag1','tag2'],  // specify some tags (e.g. if you use Sauce Labs)
 name: 'my test'         // set name for test (e.g. if you use Sauce Labs)
 ```
 
+####desiredCapabilities for Appium
+If you want to reuse your tests of WebdriverIO together with Appium for real device tests, you just need to exchange the port at the desiredCapabilities and add some options. Android tests are possible with Chrome, iOS tests with Safari, but you need the SafariLauncher to start the Safari. It need to be installed on the device and opened as a native app. It can be downloaded here : https://github.com/budhash/SafariLauncher
+Appium also provides native App testing, but in this case, some App data are needed.
```
christian-bromann: This is not quite true. You can run Appium tests without providing an app

tschuliiia: What do you mean without proving an app? just the browser testing with appium?

christian-bromann: yeah, browser testing on a mobile device though, for example with these capabilities:

```js
caps['browserName'] = 'Safari';
caps['appiumVersion'] = '1.4.10';
caps['deviceName'] = 'iPhone 6 Plus';
caps['deviceOrientation'] = 'portrait';
caps['platformVersion'] = '8.4';
caps['platformName'] = 'iOS';
```

christian-bromann:
I guess we can close this PR, as I don't thing that this paragraph adds much value to the docs.

PR #755, opened 2015-09-02
Update documentation for mocha generator support

TimBeyer:
Explains generator support if using mocha and babel


christian-bromann:
Thaanks!

PR #762, opened 2015-09-06
404 on isExisting callback should be handled and same as element not found

compulim:
If `elements` command return `404 no such element`, it means the element does not exist and `isExisting` should return `false`.

Spec at https://w3c.github.io/webdriver/webdriver-spec.html#handling-errors.

### DIFF
--- a/lib/protocol/elements.js
+++ b/lib/protocol/elements.js
@@ -35,5 +35,11 @@ module.exports = function elements(selector) {
     return this.requestHandler.create(
         requestPath,
         { using: found.using, value: found.value }
-    );
+    ).catch(function (err) {
+        if (err.message === 'no such element') {
+            return [];
+        } else {
+            throw err;
+        }
+    });
 };


lib/commands/isExisting.js
```diff
@@ -45,6 +45,12 @@ module.exports = function isExisting (selector) {
         }
 
         return false;
+    }, function (err) {
+        if (err.message === 'no such element') {
```
christian-bromann: Have you been able to test this? Will this be the exact error message?

christian-bromann:
ping @compulim

compulim:
Sorry for late response.

Yes, this is what I got from `err` object from `RequestHandler`.

``` js
{ [Error: no such element] message: 'no such element', type: 'RuntimeError' }
```

And inside HTTP response body (from Edge).

``` js
{
    "sessionId": "4E829451-4081-422D-B65D-E31DCEE3752D",
    "status": 7,
    "value": {
        "message": "no such element"
    }
}
```

It doesn't follow W3C draft spec either. But under the new [W3C spec](https://w3c.github.io/webdriver/webdriver-spec.html#handling-errors), error will be returned as "JSON error code" in string format, like `no such element`. The new erroneous HTTP response body should looks like this:

``` js
{
    "error": "invalid session id",
    "message": "No active session with id 1234",
    "stacktrace": ""
}
```

In long term, looks like the `RequestHandler` and/or `ErrorHandler` may need to be revised to identify whether the error is from older Selenium spec or newer WebDriver spec.

christian-bromann:
> looks like the RequestHandler and/or ErrorHandler may need to be revised to identify whether the error is from older Selenium spec or newer WebDriver spec

Why not do that now :wink: ?

Usually calling element__s__ returns an empty array if no element was found (spec says: _"Let result be equal to an empty list"_) so the command `elements` actually should never get rejected. Can you confirm that this happens in `isExisting`?

compulim:
Sorry I don't quite get it. `isExisting` is actually using `elements` command. Currently, Edge fail `elements` with 404 instead of an empty array, why would `isExisting` be different?

christian-bromann:
Well that's the point. According to the spec `elements` should return with an empty array instead of returning with 404 (NoSuchElement). I just ran the following snippet:

```js
client
  .init()
  .url('http://google.com')
  .elements('.notExisting').then(console.log)
  .end();
```

And got the following result from Edge:

```txt
RuntimeError: no such element (WARNING: The server did not provide any stacktrace information)
Command duration or timeout: 16 milliseconds
For documentation on this error, please visit: http://seleniumhq.org/exceptions/no_such_element.html
Build info: version: '2.47.1', revision: '411b314', time: '2015-07-30 03:03:16'
System info: host: 'WIN-SB3ER6JQ6ME', ip: '172.20.49.15', os.name: 'Windows 8.1', os.arch: 'x86', os.version: '6.3', java.version: '1.8.0_51'
Driver info: org.openqa.selenium.edge.EdgeDriver
Capabilities [{acceptSslCerts=true, browserVersion=20.10240.16384.0, platformVersion=10, browserName=MicrosoftEdge, takesScreenshot=true, pageLoadStrategy=normal, takesElementScreenshot=true, platformName=windows, platform=ANY}]
Session ID: XXXXXXXX-XXXX-XXXX-XXXX-XXXXE39A2EE3
*** Element info: {Using=css selector, value=.notExisting}
```

Whereas Chrome returned:

```txt
{ status: 0,
  sessionId: '896ef58bf3514bf6a255c5b88ff1dc99',
  value: [] }
```

Therefor I would propose the following: Instead of handling this in `isExisting` I would check the response in the `elements` command. We should make sure that if that command returns with an 404-noSuchElement we should full-fill the command instead with an empty array.

compulim:
I have updated the PR with new commit. Now modify `elements.js` and return `[]` on 404 `no such element`.

To differentiate between new WebDriver and old Selenium responses, I think it should belong to another PR.

christian-bromann:
Cool :+1: thanks

christian-bromann:
Published with `v3.2.5`

PR #763, opened 2015-09-06
Test runner should able to set remote.path by environment variable _PATH

compulim:
This enable test runner to run against Edge.

If `_PATH` is not set (`undefined`), will default to `/wd/hub`, see https://github.com/webdriverio/webdriverio/commit/4eb6585f2ef6fc02ac22489b96960c0c7959203b.

### DIFF
--- a/test/conf/local.js
+++ b/test/conf/local.js
@@ -1,6 +1,7 @@
 var local = {
     host: 'localhost',
     port: process.env._PORT || 4444,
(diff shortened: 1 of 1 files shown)


christian-bromann:
I'm fine with this change but you know that this doesn't change the behavior of the `wdio` test runner. It allows only to run the internal test suite with ms edge.

PR #773, opened 2015-09-09
Fix example .babelrc in the getting started guide

whittle:
It seems that Babel (tested using v5.8.23) expects the .babelrc file to be valid JSON. This patch changes the example .babelrc file content to conform to JSON spec.


christian-bromann:
ohh good catch, thanks :+1:

PR #779, opened 2015-09-11
Distinguish between timeout errors and other errors.

jwalton:
Wondering if you're interested in taking a patch along these lines - same needs to be done for other `waitFor*()` functions if you are.

The case I'm trying to address here is, if you call `waitForVisible()`, and, say, there's an alert showing, then you'll get an "Unexpected alert open" error when `waitForVisible()` [calls `isVisible()`](https://github.com/webdriverio/webdriverio/blob/c02ae25adfd72c5fbec2901a458bd5083c950bc1/lib/commands/waitForVisible.js#L37), but this exception gets eaten and turned into an 'element still not visible...' error, which is a little misleading.

Is this a good way to fix this?  Better ideas?

### DIFF
--- a/test/spec/functional/chaiPromises.js
+++ b/test/spec/functional/chaiPromises.js
@@ -2,16 +2,14 @@ describe('chai-as-promised', function() {
 
     /**
      *  Remove the `should` global on Object.prototype to allow chai.should for these tests,
-     *  and set up chai, chai-as-promised and chai.should properly
+     *  and set up chai and chai.should properly
      */
 
     require('should').noConflict();
 
     var chai = require('chai');
-    var chaiAsPromised = require('chai-as-promised');
 
     chai.Should();
-    chai.use(chaiAsPromised);
 
     before(h.setup());
 
--- a/test/spec/waitFor.js
+++ b/test/spec/waitFor.js
@@ -48,7 +48,7 @@ describe('waitFor',function() {
 
         it('should pass through an error from isExisting()', function() {
             var client = this.client;
-            restore = function() {client.addCommand('isExisting', require('../../lib/commands/isExisting'), true);}
+            var restore = function() {client.addCommand('isExisting', require('../../lib/commands/isExisting'), true);}
             this.client.addCommand('isExisting', function() {throw new Error("My error")}, true);
 
             return expect(this.client.waitForExist('#notExisting', duration))
@@ -151,7 +151,7 @@ describe('waitFor',function() {
 
         it('should pass through an error from isVisible()', function() {
             var client = this.client;
-            restore = function() {client.addCommand('isVisible', require('../../lib/commands/isVisible'), true);}
+            var restore = function() {client.addCommand('isVisible', require('../../lib/commands/isVisible'), true);}
             this.client.addCommand('isVisible', function() {throw new Error("My error")}, true);
 
             return expect(this.client.waitForVisible('#notExisting', duration))
--- a/test/spec/waitFor.js
+++ b/test/spec/waitFor.js
@@ -17,6 +17,16 @@ describe('waitFor',function() {
             return this.client.waitForEnabled('.waitForValueEnabledReverse', duration, true).then(checkTime);
         });
 
+        it('should return with an error if the element never becomes visible', function() {
+            return expect(this.client.waitForEnabled('.waitForValueEnabledReverse', 10))
+            .to.be.rejectedWith('element (.waitForValueEnabledReverse) still not enabled after 10ms');
+        });
+
+        it('should pass through an error from isEnabled()', function() {
+            return expect(this.client.waitForEnabled('#notExisting', duration, true))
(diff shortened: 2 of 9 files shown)


jwalton:
The other approach I was considering was modifying `waitFor()` so the [thrown exception](https://github.com/webdriverio/webdriverio/blob/c02ae25adfd72c5fbec2901a458bd5083c950bc1/lib/commands/waitUntil.js#L40) has some kind of flag indicating this is a timeout, rather than relying on the message string (which seems fragile.)

christian-bromann:
@jwalton thanks for the PR!

How about we check the error message in all waitForXXX commands (e.g. [here](https://github.com/webdriverio/webdriverio/blob/master/lib/commands/waitForText.js#L49)) if it is a timeout (error message is _"Promise never resolved with an truthy value"_ we throw the custom error message otherwise we just throw the propagated error which would be this: 
https://github.com/webdriverio/webdriverio/blob/master/lib/commands/waitUntil.js#L95

jwalton:
Sounds good.  I'll make the changes and see if I can add some tests.

christian-bromann:
Awesome :+1:

jwalton:
Do you have a preferred way for mocking out, say `isVisible()` in a test and forcing it to return an error?

christian-bromann:
We have that test page setup for tests. Just run a simple http server (e.g. `python -m SimpleHTTPServer 8080`) there we have couple of elements to interact with. Just add an element if you think you need it for your test.

jwalton:
Ok, changes and tests.  Lemme know if that's all OK.

test/spec/waitFor.js
```diff
@@ -31,6 +41,22 @@ describe('waitFor',function() {
...
         });
 
+        it('should return with an error if the element never exists', function() {
+            return expect(this.client.waitForExist('#notExisting', 10))
+            .to.be.rejectedWith('element (#notExisting) still not existing after 10ms');
+        });
+
+        it('should pass through an error from isExisting()', function() {
+            var client = this.client;
+            restore = function() {client.addCommand('isExisting', require('../../lib/commands/isExisting'), true);}
```
christian-bromann: missing `var`

test/spec/waitFor.js
```diff
@@ -87,6 +144,22 @@ describe('waitFor',function() {
...
         });
 
+        it('should return with an error if the element never becomes visible', function() {
+            return expect(this.client.waitForVisible('#notExisting', 10))
+            .to.be.rejectedWith('element (#notExisting) still not visible after 10ms');
+        });
+
+        it('should pass through an error from isVisible()', function() {
+            var client = this.client;
+            restore = function() {client.addCommand('isVisible', require('../../lib/commands/isVisible'), true);}
```
christian-bromann: same here `var restore = ...`

test/spec/waitFor.js
jwalton: Sorry.  I mainly write coffee-script.  :P

christian-bromann:
Awesome! Two minor things otherwise LGTM

jwalton:
`var` keywords added.

jwalton:
Hopefully all the tests pass.  I only tried the one I modified.  :D

christian-bromann:
WebdriverIO has the most reliable integration tests ever :laughing:  (not) .. let's see

christian-bromann:
Thanks anyway!

test/spec/functional/chaiPromises.js
```diff
@@ -2,16 +2,14 @@ describe('chai-as-promised', function() {
...
+     *  and set up chai and chai.should properly
      */
 
     require('should').noConflict();
 
     var chai = require('chai');
-    var chaiAsPromised = require('chai-as-promised');
 
     chai.Should();
-    chai.use(chaiAsPromised);
```
christian-bromann: Why did you changed that? Have to revert it. No worries just fyi :wink:

jwalton: Because my tests weren't passing without it.  :P  Maybe just because I was
trying to run one test in isolation?

christian-bromann: What was the error? This should actually not affect your test as it is a functional test.

jwalton: Oh, wait...  Why did I change this...  O_o

I had to change something to I could use .rejectedWith(), but I did that in
the same file as defines 'setup()'.

christian-bromann: I see. Well we have that chaiPromise test to make sure that transferPromiseness works. I guess if we just add the chai stuff in there again we still proper test this function.

jwalton: Oh, this is why I took it out from here, because I added it elsewhere.
(Sorry; I'm in a restaurant doing this on my phone.)

christian-bromann:
published with 3.2.4

PR #786, opened 2015-09-15
fixed xunit report status attribute in the testcase tag

danielgil:
Small fix in the xunit reporter.  When writing the <code>testcase</code> tag, if there were no errors then <code>test.err</code> would be {}, which evaluates to <code>true</code> and ends up marking the status of all tests as 'failed'.

I haven't seen tests for the reporters, if I missed them please point me to them and I'll be glad to add a test.

### DIFF
--- a/lib/reporter/xunit.js
+++ b/lib/reporter/xunit.js
@@ -270,7 +270,7 @@ XUnit.prototype.genTestsuite = function(tests) {
             time: test.time / 1000,
             id: test.id,
             file: test.file,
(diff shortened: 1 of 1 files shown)


christian-bromann:
Thanks, seems that my Jenkins plugins just ignored that. We don't have any reporter tests yet. In v4 each reporter will live it its own NPM module. This will make it easier to write tests for them.

Will update the WebdriverIO package later today

christian-bromann:
published in 3.2.4

PR #788, opened 2015-09-15
Remove undefined argument from after()

silvenon:
I don't know if there was a reason behind this. Also, I added these arguments to generated and example configs.


christian-bromann:
Thanks. I was aware of that and already fixed it for v4. Will get out later today.

silvenon:
I was just thinking, would it be better to, instead of the `undefined`, pass the `res`? Or is there another way to get the `sessionId` from the `after` hook?

christian-bromann:
published in 3.2.4

PR #789, opened 2015-09-15
Expose sessionId and capabilities to the after() hook

silvenon:
Ok, I will need a bit of help here.

1. Was `sessionId` already reachable to `after` somehow?
2. In which order should the arguments go?
3. Is there a way to expose it in both cases?

I needed `sessionId` for updating the job status in my service.

Once this either gets merged and released, or I find out about an existing way to reach `sessionId`, I can add a CrossBrowserTesting `wdio.config.js` to your examples. I finally mastered the crap out of it.

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -164,7 +164,7 @@ process.on('message', function(m) {
                 return global.browser.end();
             }).then(function(res) {
                 if(!config.updateJob || config.host.indexOf('saucelabs') === -1 || !res || !res.sessionId) {
-                    return;
+                    return res.sessionId;
                 }
 
                 /**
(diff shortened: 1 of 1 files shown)


lib/runner.js
```diff
@@ -164,7 +164,9 @@ process.on('message', function(m) {
                 return global.browser.end();
             }).then(function(res) {
                 if(!config.updateJob || config.host.indexOf('saucelabs') === -1 || !res || !res.sessionId) {
-                    return;
+                    return q.fcall(function () {
+                        return res.sessionId;
+                    });
```
christian-bromann: you don't need to call an extra function, just return the value

christian-bromann:
@silvenon thanks for the PR, left one comment

silvenon:
Ok :+1: I wanted to ask if you're happy with the order of arguments so I can add it to the docs. And maybe we could drop `process.pid` as an argument? (I first thought it was session ID, that's why I sent a PR to make that public) Because I don't know why you would need it, so perhaps it's just confusing for the user.

christian-bromann:
@silvenon let's replace the process id with the the capabilities object. This might be more useful I guess. What do you think?

silvenon:
Great idea :smiley: Done. Sorry for the delay.

christian-bromann:
Awesome thanks

PR #797, opened 2015-09-18
Only generate failure tags on error

austek-mdsol:
Only generate failure tags when there are 1+ errors


christian-bromann:
Awesome! Thanks

christian-bromann:
Published with `v3.2.5`

PR #800, opened 2015-09-21
Prevent `RangeError: Invalid array length` for negative indent when tests fail

georgecrawford:


christian-bromann:
Well I actually never considered indents to be negative but it seems to be possible :laughing:

christian-bromann:
Published with `v3.2.5`

PR #809, opened 2015-10-01
Updated xUnit test reporter

sixsixtrample:
Added check if test.err exists and changed test for whether a test has
passed or failed. Tested both a test that should pass and a test that
should fail. Both printed correctly.

### DIFF
--- a/lib/reporter/xunit.js
+++ b/lib/reporter/xunit.js
@@ -98,7 +98,7 @@ function XUnit(options) {
         parentSuite.pending++;
         test.id = ++id;
         test.time = new Date() - tests[test.pid][0].end,
-        test.timestamp = tests[test.pid][0].end = new Date();
+            test.timestamp = tests[test.pid][0].end = new Date();
         test.runnerData = runnerData[test.pid];
         parentSuite.tests.push(test);
     });
@@ -108,7 +108,7 @@ function XUnit(options) {
         parentSuite.passes++;
         test.id = ++id;
         test.time = new Date() - tests[test.pid][0].end,
-        test.timestamp = tests[test.pid][0].end = new Date();
+            test.timestamp = tests[test.pid][0].end = new Date();
         test.runnerData = runnerData[test.pid];
         parentSuite.tests.push(test);
     });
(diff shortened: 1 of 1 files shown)


lib/reporter/xunit.js
```diff
@@ -98,7 +98,7 @@ function XUnit(options) {
         parentSuite.pending++;
         test.id = ++id;
         test.time = new Date() - tests[test.pid][0].end,
```
christian-bromann: could you replace the `,` with an `;` and remove the indentation below?

christian-bromann:
@sixsixtrample  One comment on something that is not related to this but would be great if we can fix this as well. After that it is good to go

lib/reporter/xunit.js
sixsixtrample: Updated.

christian-bromann: @sixsixtrample I still see
```
test.time = new Date() - tests[test.pid][0].end,
```
but it should be
```
test.time = new Date() - tests[test.pid][0].end;
```

lib/reporter/xunit.js
```diff
@@ -108,7 +108,7 @@ function XUnit(options) {
         parentSuite.passes++;
         test.id = ++id;
         test.time = new Date() - tests[test.pid][0].end,
-        test.timestamp = tests[test.pid][0].end = new Date();
+            test.timestamp = tests[test.pid][0].end = new Date();
         test.runnerData = runnerData[test.pid];
```
christian-bromann: please also change it here

lib/reporter/xunit.js
```diff
@@ -118,7 +118,7 @@ function XUnit(options) {
         parentSuite.failures++;
         test.id = ++id;
         test.time = new Date() - tests[test.pid][0].end,
-        test.timestamp = tests[test.pid][0].end = new Date();
+            test.timestamp = tests[test.pid][0].end = new Date();
```
christian-bromann: and here

sixsixtrample:
Ok, I believe they are all fixed.

christian-bromann:
@sixsixtrample please have a look at the _Files Change_ tab. Have you missed to push to Github?

sixsixtrample:
It says I have no uncommitted changes. And I pushed it to my fork. Is there something else I need to do?

christian-bromann:
Thanks!

christian-bromann:
Published with `v3.2.5`

PR #813, opened 2015-10-05
Preserve mocha `this` in testrunner-wrapped generators

pishposh:
Hi! Normally, mocha provides a `this` context to spec and hook functions, which lets you do things like `this.skip()` a test. This p.r. makes this context available to generator functions wrapped for mocha by the testrunner. As a small bonus, I think we’re able to dispense with the adapter between co and mocha’s async `done()` callback, since mocha natively “does the right thing” with promises returned from 0-ary specs/hooks.

Feedback welcome!

### DIFF
--- a/examples/wdio/runner-specs/mocha.test.js
+++ b/examples/wdio/runner-specs/mocha.test.js
@@ -34,4 +34,10 @@ describe('webdriver.io page', function() {
 
     });
 
+    it('should be skippable (pending) from within a generator spec', function* () {
+        yield browser.pause(100);
+        this.skip();
+        throw new Error("this should not be reached");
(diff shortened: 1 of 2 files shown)


christian-bromann:
LGTM! Could you apply the same change to the [wdio-sync](https://github.com/webdriverio/wdio-sync/blob/master/index.js#L43-L59) as well as it will carry that part in v4

pishposh:
Thanks @christian-bromann. Opened a [p.r. for wdio-sync here](https://github.com/webdriverio/wdio-sync/pull/2).

christian-bromann:
Published with `v3.2.5`

PR #815, opened 2015-10-06
Play nice with Babel regenerator

pishposh:
Hi again! This p.r. improves the compatibility of `wdio` with Babel’s regenerator runtime, for both mocha and jasmine frameworks. For mocha, `runInGenerator` in the adapter is replaced with [co-mocha](https://github.com/blakeembrey/co-mocha), which monkey-patches mocha for us, same as `co.wrap()` from yesterday’s patch. Internally, this uses [is-generator](https://www.npmjs.com/package/is-generator), which is able to recognize regenerator-transformed generators that elude the `'function*'` test.

For jasmine, I’ve made the adapter use is-generator directly to detect generators, and I’ve updated detectHarmony to sniff out `global.regeneratorRuntime` explicitly. Maybe we can replace jasmine’s `runInGenerator` with [jasmine-co](https://github.com/gradecam/jasmine-co), like for mocha… what do you think? It looks like it might have to be patched to recognize regenerator generators.

Mocha and jasmine example tests (wdio.mocha.conf.js, wdio.jasmine.conf.js) are updated and working with and without Babel on Node 4 and 0.10 (without --harmony). I moved the generators to separate files so they wouldn’t immediately crash runtimes without generator support (0.10).


package.json
```diff
@@ -60,8 +62,8 @@
     "chai-as-promised": "^5.0.0",
     "coveralls": "~2.11.2",
     "es6-promise": "^2.1.1",
-    "glob": "~5.0.5",
     "istanbul": "^0.3.13",
+    "jasmine": "^2.3.2",
```
christian-bromann: Jasmine should not be a dependecy (Mocha is because we use it to test the framework)

lib/frameworks/mocha.js
```diff
@@ -17,6 +12,11 @@ module.exports.run = function(config, specs, capabilities) {
         requires = config.mochaOpts.require,
         runner;
 
+    /**
+     * enable generators if supported
+     */
+    coMocha(Mocha);
```
christian-bromann: :+1:

christian-bromann: does coMocha detect if generators are supported or not?

christian-bromann: One thing about this is that we now have a mochaCo dependency even though the user doesn't necessarily uses Mocha (same for Jasmine). I am a bit afraid that we introduce unnecessary dependency overhead only to get the code a bit shorter

christian-bromann:
Using `is-generator` to figure out if the function is actually a generator function is the only change I see worth adding. Adding coMocha as dependency only makes sense if people uses Mocha otherwise it is to much of a overhead.
The changes to the example page makes me feel like it got more complex to write tests using yield. Examples should be as simple as possible. I also don't thing that it is a good idea to separate spec files from the actual test

pishposh:
Thanks, that all makes sense. Let me restore runInGenerator and simplify the tests—I think I can do so in a way without breaking runtimes lacking generator support.

PR #816, opened 2015-10-06
Play nice with Babel regenerator, take 2

pishposh:
[(Original p.r. here.)](https://github.com/webdriverio/webdriverio/pull/815)

As suggested. I’ve tested this in Node 4 with and without Babel, and Node 0.10 with Babel, but I’m not sure how best to keep the parser from choking on `function*` when running without generator support. Best I can come up with is to detect generator support within the `describe` block, at runtime, and conditionally require a separate spec file with the generator tests. Maybe we could just split off the generator tests into separate wdio.confs? Thoughts?


christian-bromann:
LGTM! Well if people are using generators they should be aware of the fact that they need a certain version of node to do that. Maybe we could add a sentence to the docs that clarifies this? What do you think?

pishposh:
Yeah, that seems reasonable… I guess the worry is that if even if you aren’t using generators, the examples don’t work if you’re running an older Node (0.12 and below?) unless you explicitly enable --harmony. Maybe some comments inline with the test would be enough to clarify?

christian-bromann:
Well the launcher automatically runs in harmony mode (see [here](https://github.com/webdriverio/webdriverio/blob/master/lib/launcher.js#L65))

pishposh:
Ah, interesting. Looks like 0.10 doesn’t support generators even with --harmony, which explains why these have been erroring for me (and why I needed Babel in the first place). I’ll add a comment to the tests to help anyone else stuck on ≤0.10.

christian-bromann:
:+1:

christian-bromann:
yeah still was on 0.10 as well as I developed v3 but we should move on :wink: There won't be many people still using something lower than 0.12 (I hope)

christian-bromann:
Sweet thanks!

pishposh:
Thanks @christian-bromann!

christian-bromann:
Published with `v3.2.5`

PR #833, opened 2015-10-15
Support for mocha's 'qunit' interface

Flaise:


lib/frameworks/mocha.js
```diff
@@ -18,7 +19,7 @@ module.exports.run = function(config, specs, capabilities) {
         requires = config.mochaOpts.require,
         runner;
 
-    if(typeof config.mochaOpts.ui !== 'string' || !config.mochaOpts.ui.match(/(bdd|tdd)/i)) {
+    if(typeof config.mochaOpts.ui !== 'string') {
```
Flaise: Mocha may have new interfaces added to it in the future. This sort of thing shouldn't be hard-coded.

lib/frameworks/mocha.js
```diff
@@ -46,7 +47,7 @@ module.exports.run = function(config, specs, capabilities) {
     /**
      * enable generators if supported
      */
-    if(hasES6Support) {
+    if(hasES6Support && interfaces[config.mochaOpts.ui]) {
```
Flaise: This prevents the above changes from breaking anything for interfaces you don't support.

lib/frameworks/mocha.js
christian-bromann: if we have it hard coded in the `interfaces` object why not here?

lib/frameworks/mocha.js
christian-bromann: why did you add qunit in the interface at all?

lib/frameworks/mocha.js
Flaise: Because writing it this way makes it so that the worst that happens for interfaces you don't support is that generators won't work. The worst case for the way you wrote it is that it reverts to an entirely different interface.

lib/frameworks/mocha.js
Flaise: Mocha has 5 interfaces. https://mochajs.org/#interfaces One of my projects uses one that you (without this PR) don't support.

christian-bromann:
Ok, looks good. Thanks for the explanations.

PR #840, opened 2015-10-21
Add a little information on the browser variable

FBerthelot:
Because i was lost without this information :cry:


christian-bromann:
Sweet thanks!

PR #847, opened 2015-10-28
Update v3.md

philhaz:
"done" callback is not necessary in the improved example - in fact it will halt the test as it's not called when using promises.


christian-bromann:
good catch

PR #850, opened 2015-11-01
Set value more unicode support

sanoodles:
Makes the setValue function to support more (all?) unicode characters.

The idea of using Array.from(value) instead of value.split('') comes from
https://mathiasbynens.be/notes/javascript-unicode#reversing-strings


christian-bromann:
Nice! Thanks!

PR #852, opened 2015-11-03
Check failure cause in commands using `waitUntil`

### LINKED ISSUE #820: Commands using waitUntil should check failure cause
`waitUntil` will throw an error if the wait times-out (`Promise never resolved with an truthy value`), but also if the command encounters an error (`Promise was fulfilled but got rejected with the following reason: ...`). Commands which use `waitUntil`, such as `waitForVisible`, must check the thrown error to see which reason was encountered. For a time-out, a message like `element (#splash) still  visible after 40000ms` is perfect, but if there was a Selenium error with the looping command, it's vital to print that instead, eg. `waitForVisible failed with error: stale element reference: element is not attached to the page document`.

georgecrawford:
Fixes #820

### DIFF
--- a/lib/commands/waitForEnabled.js
+++ b/lib/commands/waitForEnabled.js
@@ -42,8 +42,11 @@ let waitForEnabled = function (selector, ms, reverse = false) {
             return result !== reverse
         })
     }, ms).catch((e) => {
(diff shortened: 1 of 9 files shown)


christian-bromann:
Sweet! Thanks so much for moving this to v4.0

PR #854, opened 2015-11-04
Corrections and slight enhancements to grammar.

dpm3354:
I believe English readers will gain confidence in this tool if the documentation is easier to read.


christian-bromann:
I totally agree. Thanks for fixing these issues.

dpm3354:
I'd be happy to do more when time lends itself!

PR #857, opened 2015-11-05
Add staleElementRetry helper

### LINKED ISSUE #748: Better protection against stale element reference (feature request)
It would be nice if webdriverio will try to re-fetch the stale element if such error occurs.
Or would it?

For instance I was having some problems testing button with a counter, where waitForCountToIncrease would fail very often because webdriver would get counter reference, but before next check the count updates and re-renders the whole button, and then webdriver check fails.

Natural place would probably be [waitUntil](https://github.com/webdriverio/webdriverio/blob/ce51baa448714e69f19899718236f988417b4b3d/lib/commands/waitUntil.js#L93), with similar check to timeout but using attempts counter, I guess.

Thanks!

georgecrawford:
Fixes #748

### DIFF
--- a/lib/commands/addValue.js
+++ b/lib/commands/addValue.js
@@ -35,6 +35,7 @@
  */
 
 var ErrorHandler = require('../utils/ErrorHandler.js');
+var staleElementRetry = require('../helpers/staleElementRetry');
 
 module.exports = function addValue (selector, value) {
 
@@ -67,6 +68,7 @@ module.exports = function addValue (selector, value) {
 
         return this.unify(elementIdValueCommands);
 
-    });
+    })
+    .catch(staleElementRetry.bind(this, 'isVisible', arguments));
 
 };
--- a/lib/commands/addValue.js
+++ b/lib/commands/addValue.js
@@ -69,6 +69,6 @@ module.exports = function addValue (selector, value) {
         return this.unify(elementIdValueCommands);
 
     })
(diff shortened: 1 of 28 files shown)


christian-bromann:
@georgecrawford great contribution, thanks!
I've seen couple of occurrences where the command name is wrong. Could you fix this?

lib/commands/addValue.js
```diff
@@ -67,6 +68,7 @@ module.exports = function addValue (selector, value) {
 
         return this.unify(elementIdValueCommands);
 
-    });
+    })
+    .catch(staleElementRetry.bind(this, 'isVisible', arguments));
```
christian-bromann: e.g. here, it should be `.catch(staleElementRetry.bind(this, 'addValue', arguments));`

georgecrawford:
Damn, sorry! Seems I got lazy halfway through.

DV3D:
Great work!! hope this gets into a new release quickly :+1:

christian-bromann:
Let's do it. :+1:

georgecrawford:
@christian-bromann We should discuss what to do with this in v4.0. I could apply the same patch, or we could come up with a nicer solution.

christian-bromann:
@georgecrawford for v4 we should use an `afterCommand` hook if possible. Problem will be that if a staleElementError appears we can't just retry the last command with the same parameters (e.g. `elementIdVisible`) we need to query the element again. I guess a solution could be to use the command history and execute the the command that executed an element command.

PR #861, opened 2015-11-06
Correct CSS values in documentation

gitLuigi:
"visibility" and "display" properties.

Also checked:
✓ isVisible.js properties are good.
✓ isVisibleWithinViewport.js properties are good.


georgecrawford:
Thanks @gitLuigi

PR #862, opened 2015-11-08
Pass config object to runner, so that users can pass simple values between processes

georgecrawford:
I found that there were a couple of use cases where it'd be nice to be able to set up a dynamic variable in wdio.conf.js, and have it propagated to the child processes. For example, I'd like each test run to create a timestamped directory in which to save log files, screenshots, etc.

As a user, I'd expect to be able to export the path of my log directory in `onPrepare`, and to be able to use it in `before` and other defined commands. With this PR, I can do this:

wdio.conf.js
```
		onPrepare: function(config, capabilities) {
			config.outputDirs = utils.getOutputDirs();
		},
```

```
		const screenshotFile = `${browser.options.outputDirs.screenshots}/${Date.now()}-${message}.png`;
		browser.saveScreenshot(screenshotFile);
```

### DIFF
--- a/lib/launcher.js
+++ b/lib/launcher.js
@@ -78,7 +78,7 @@ class Launcher {
          * if it is an object run multiremote test
          */
         if (this.isMultiremote()) {
-            return this.startInstance(this.configParser.getSpecs(), caps)
+            return this.startInstance(this.configParser.getSpecs(), caps, config)
         }
 
         /**
@@ -104,7 +104,7 @@ class Launcher {
 
         let exitCode = await new Promise((resolve, reject) => {
             this.resolve = resolve
-            this.runSpecs()
+            this.runSpecs(config)
         })
 
         try {
@@ -118,7 +118,7 @@ class Launcher {
     /**
      * run multiple single remote tests
      */
-    runSpecs () {
+    runSpecs (config) {
         let specsLeft = 0
         let isRunning = false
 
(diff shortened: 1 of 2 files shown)


christian-bromann:
Why not just:
```js
        onBefore: function(config, capabilities) {
            browser.options.outputDirs = utils.getOutputDirs();
        },
```
and
```js
        const screenshotFile = `${browser.options.outputDirs.screenshots}/${Date.now()}-${message}.png`;
        browser.saveScreenshot(screenshotFile);
```
?

georgecrawford:
Because `getOutputDirs` uses a timestamp, and would therefore create a new path every time it's called. I need it to only be called once.

christian-bromann:
The `before` hook only gets called once.

georgecrawford:
Not true. With two capabilities, and:

```
		before: function() {
			console.log('before', browser.options.outputDirs);
		},
```
I get:
```
before { output: './output/2015-11-08 19.45.12',
  screenshots: './output/2015-11-08 19.45.12/screenshots',
  logs: './output/2015-11-08 19.45.12/logs' }

before { output: './output/2015-11-08 19.45.12',
  screenshots: './output/2015-11-08 19.45.12/screenshots',
  logs: './output/2015-11-08 19.45.12/logs' }
```

georgecrawford:
ping @christian-bromann

PR #874, opened 2015-11-12
Added list of options for loglevel

indolering:
Rerunning  `wdio config` just to get the list of logging options is a PITA.


christian-bromann:
:+1:

PR #877, opened 2015-11-13
Added getUrl property

Risto-Stevcev:
Added a getUrl property just like getTitle, to make the url() protocol work with chai promises when retreiving the current url:

    browser.getUrl().should.eventually.be.equal('http://some-url')


christian-bromann:
I guess it's fair to have that kind of command since getTitle also exists. For v4 we are thinking of treating all commands equally.

PR #881, opened 2015-11-17
Update sauce labs job when host set to localhost for use with sauce connect.

### LINKED ISSUE #848: Sauce labs job not updated when host set to localhost because using sauce connect
I am currently working on using webdriver.io to set up some integration tests with sauce labs using sauce connect. I have my host set to localhost and port set to 4445 per the defaults for sauce connect. The tests run properly, but the job is not being updated with the status at the end.

It looks like the issue is with the line below. The code only attempts to update the job when the host includes "saucelabs" which won't happen when hitting localhost with sauce connect.

https://github.com/webdriverio/webdriverio/blob/master/lib/runner.js#L166

Is updateJob used for any of the other services that webdriver.io uses? If not, would it be sufficient to remove the host check here (if so, I'll work on making a pull request). If that isn't a good solution, any suggestions for modifying this to be more flexible and work for the sauce connect localhost case?

juliepagano:
Fixes #848. Sorry it took me so long to get to it!

Per discussion on the issue, I haven't added a test.

When a user is using sauce connect, they may set the host to localhost. The
check for saucelabs in the host name prevented these jobs from updating.

Per discussion on the issue, this throws away errors from the sauce labs api
because they will often be caused by tests that don't use sauce labs.

This commit also modifies this block of code to properly pass along the session
id after calling the sauce labs api. I noticed it wasn't working properly when testing
my fix.

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -163,7 +163,7 @@ process.on('message', function(m) {
                 failures = 1;
                 return global.browser.end();
             }).then(function(res) {
-                if(!config.updateJob || config.host.indexOf('saucelabs') === -1 || !res || !res.sessionId) {
+                if(!config.updateJob || !res || !res.sessionId) {
                     return res.sessionId;
                 }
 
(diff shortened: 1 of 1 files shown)


christian-bromann:
Sweet! Thanks. I will try to get out a new version within the next days.

juliepagano:
Thanks!

PR #896, opened 2015-11-23
More useful BaseReporter for junit reporting

georgecrawford:


lib/utils/ConfigParser.js
```diff
@@ -14,7 +14,7 @@ const DEFAULT_CONFIGS = {
     baseUrl: null,
     waitforTimeout: 1000,
     framework: 'mocha',
-    reporter: 'dot',
+    reporter: [],
```
christian-bromann: if we do this change, we need to remove the dot reporter from the dependencies. I thought it would be nice to have it as default reporter so people don't get confused when they don't see anything in the terminal

lib/utils/BaseReporter.js
```diff
@@ -283,6 +343,51 @@ class BaseReporter extends events.EventEmitter {
         }
     }
 
+    getSpecStats (obj) {
+        if (!this.stats.runner[obj.cid]) {
```
christian-bromann: Instead of the it block , could we use `Object.assign` here?

lib/utils/BaseReporter.js
```diff
@@ -283,6 +343,51 @@ class BaseReporter extends events.EventEmitter {
...
+                start: new Date(),
+                files: obj.specs,
+                suites: {}
+            }
+        }
+        return this.stats.runner[obj.cid].specs[obj.specHash]
+    }
+
+    getSuiteStats (suite) {
+        if (!suite.title) console.log('NO SUITE TITLE', suite, Error().stack)
```
christian-bromann: `console.log` is this a debug leftover?

lib/utils/BaseReporter.js
```diff
@@ -283,6 +343,51 @@ class BaseReporter extends events.EventEmitter {
...
+                start: new Date(),
+                tests: {}
+            }
+        }
+        return specStats.suites[suite.title]
+    }
+
+    getTestStats (test) {
+        const suiteStats = this.getSpecStats(test).suites[test.parent]
+        if (!suiteStats.tests[test.title]) {
```
christian-bromann: same here: please use `Object.assign`

lib/utils/ConfigParser.js
georgecrawford: I'm explicitly falling-back to dot reporter here: https://github.com/webdriverio/webdriverio/pull/896/files#diff-ff5050d3c7f51fca90c4691a4599e852R47

lib/utils/BaseReporter.js
georgecrawford: Hmm. I'm not very familiar with `Object.assign`. Could you show me how to use it in this case?

lib/utils/BaseReporter.js
```diff
@@ -187,6 +224,29 @@ class BaseReporter extends events.EventEmitter {
...
+            } else if (val.length > STRINGLIMIT) {
+                return val.slice(0, STRINGLIMIT) + '... (' + (val.length - STRINGLIMIT) + ' more bytes)'
+            }
+            return val
+        case '[object Array]':
+            return val.map(this.limit.bind(this))
+        case '[object Object]':
+            const copy = {}
+            const keys = Object.keys(val)
+            for (let i = 0, l = keys.length; i < l; i++) {
```
christian-bromann: how about `for (let key of keys) { copy[key] = this.limit(val[key]); }`

lib/utils/BaseReporter.js
```diff
@@ -63,45 +66,76 @@ class BaseReporter extends events.EventEmitter {
...
         this.on('runner:start', (runner) => {
-            if (!this.stats.runner[runner.cid]) {
-                this.stats.runner[runner.cid] = {
-                    start: new Date(),
-                    capabilities: runner.capabilities,
-                    config: runner.config,
-                    tests: []
-                }
-            }
+            this.getSpecStats(runner)
```
christian-bromann: these methods are easier to test if they have the following structure:

```js
this.stats.runner[runner.cid] = this.getSpecStats(runner);
```

lib/utils/ConfigParser.js
christian-bromann: :+1:

lib/utils/BaseReporter.js
christian-bromann: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign
It works basically like a `_.merge` or `_.default``
```js
this.stats.runner[obj.cid] = Object.assign({
    start: new Date(),
    capabilities: obj.capabilities,
    sanitizedCapabilities: sanitize.caps(obj.capabilities),
    config: obj.config,
    specs: {}
}, this.stats.runner[obj.cid]);
```

lib/utils/BaseReporter.js
georgecrawford: Hmm. I understand this would make it simpler to test (presumably you would mock `this.getSpecStats`?), but can't see a way to do it and still keep the abstraction contained within `this.getSpecStats`. Otherwise, for example, `runner:command`, `runner:result`, `test:start`, etc. would all need to assign the result from `this.getTestStats()` to the following nested structure `this.stats.runner[test.cid].specs[test.specHash].suites[test.parent].tests[test.title]`. Is that what you mean?

christian-bromann: @georgecrawford from what I see is that the function sets some default properties to the payload and enhances it with additional data. The confusing part is why does it return something if you actually don't assign it to any object? I am also not a fan of super long object notations like `this.stats.runner[obj.cid].specs[obj.specHash]` , it is super dangerous if one of the values is accidentally undefined. I tried to rewrite the function but didn't found a right solution. Maybe you know better what the actual purpose of it is. Maybe create a helper class to abstract this better where you can initiate an instance with a bunch of data an get a formatted object back to propagate to other reporter. What do you think?

georgecrawford: So, my thinking is that reporters might want either to handle each event as it comes in (in which case, we just forward the payload from the BaseReporter to each reporter), but that some reporters might alternatively only want a summary of everything at the end. The Junit reporter needs to do nothing until the end (it can't, as the XML requires a summary of the test passes & failures, etc). If it collects the data itself based on the events, that's fine until another reporter starts to do exactly the same thing. It seems like a responsibility of the BaseReporter to be able to summarise everything. 

If it's to do this, the BaseReporter needs to be able to associate every event payload with the current spec, suite or test. For example, when `runner:command` is fired, the payload includes enough data (`title`, `parent`, `cid`, `specHash`) to identify where in the large `this.stats` object to store the payload.

The reason `getTestStats()` (and others) are horrible methods is that they have two purposes - in the case of `test:pass` it will return the current test, and we can set `state = 'pass'`, but in the case of `test:start` all we need to do is to create the test record in the right place within `this.stats`, and set `start: new Date()`. Of course, I could tweak it slightly so that `getTestStats()` only returns an object if it exists, then each caller would have to do something like:

```js
        this.on('test:start', (test) => {
            const testStats = this.getTestStats(test) || this.createTestStats(test)
            testStats.state = 'pass'
            this.stats.passes++
            this.propagateEvent('test:pass', test)
        })
```

I suppose memory is also a concern here - keeping so much test output might not work for long-running tests.

I don't like the design I've come up with - I agree with you comments. But perhaps you understand the use-case better now? Let me know your thoughts.

christian-bromann: > It seems like a responsibility of the BaseReporter to be able to summarise everything.

:+1: 

> The reason getTestStats() (and others) are horrible methods is that they have two purposes

Yes!

> I suppose memory is also a concern here - keeping so much test output might not work for long-running tests.

That's a valid point! Though I think that is a different topic and we should not block ourself due to this issue. I have no experiences where the critical point here is. I don't think that we store that much data and I guess we are more constraint of the CPU power (I tried to run ~50 capabilties at once and my computer started lacking).

Performance optimisations is definitely something we should care about but not for this release. I would say even after we improved mobile in WebdriverIO. Maybe [node-cache](https://www.npmjs.com/package/node-cache) could be a solution here. But before we don't have any benchmarks we shouldn't make early assumptions.

Apparently this is more complex then we initially thought. I think we should split this into smaller pieces. That said what do you think about creating classes that only purpose is to take an object of data and format it into the right format so it can be propagated to the other reporters. So maybe have `TestResult` class for the `state = 'pass'` case and a `TestState` for the `test:start` one. It doesn't matter if they only have one or two methods. The goal is to simplify the BaseReporter routines to make the code easy to read.

georgecrawford:
@christian-bromann What do you think of the latest commit? It's simplified things a lot, but would you like to go further?

NB: all the classes are in one file for now - happy to split them out if you like.

christian-bromann:
@georgecrawford will try to review asap

lib/utils/BaseReporter.js
```diff
@@ -40,121 +42,329 @@ const SYMBOLS = {
...
+        this.on('test:start', (test) => {
+            this.stats.testStart(test)
         })
 
-        this.on('test:end', (test) => {
-            this.stats.tests++
+        this.on('hook:start', (hook) => {
+        })
+
+        this.on('hook:end', (hook) => {
```
christian-bromann: Do we need these if nothing happens anyway? Let's remove it.

lib/utils/BaseReporter.js
```diff
@@ -173,6 +383,28 @@ class BaseReporter extends events.EventEmitter {
         return `\u001b[${COLORS[type]}m${str}\u001b[0m`
     }
 
+    limit (val) {
+        switch (Object.prototype.toString.call(val)) {
+        case '[object String]':
+            if (val.length > 100 && val.match(/^([A-Za-z0-9+\/]{4})*([A-Za-z0-9+\/]{4}|[A-Za-z0-9+\/]{3}=|[A-Za-z0-9+\/]{2}==)$/)) {
```
christian-bromann: What's that regex doing?

christian-bromann:
Great work @georgecrawford! One thing though: could you add couple of comments above the class methods and classes so everyone knows what's that function/class is doing and where it is used.

Other than that, lgtm!

lib/runner.js
```diff
@@ -130,11 +158,45 @@ class Runner {
...
+            cid: this.cid,
+            specs: this.specs
         })
         process.exit(failures === 0 ? 0 : 1)
     }
 
+    addTestDetails (payload) {
+        if (this.currentTest) {
+            payload.title = this.currentTest.title
+            payload.parent = this.currentTest.parent
```
christian-bromann: We need to solve this differently. I am sure that Jasmine doesn't provide you title and parent that easily (also in Cucumber this isn't quite applicable). Having that in the main repository seems not correct.

lib/runner.js
```diff
@@ -57,29 +64,49 @@ class Runner {
...
+                event: 'runner:screenshot',
+                cid: m.cid,
+                specs: this.specs,
+                filename: payload.filename,
+                data: payload.data
+            }
+            process.send(this.addTestDetails(details))
+        })
+
+        process.on('test:start', (test) => {
```
christian-bromann: See [comment](https://github.com/webdriverio/webdriverio/pull/896/files#r47439165). Can't we move this to the basereporter? and apply it to the test data there?

lib/runner.js
christian-bromann: After I finished the [cucumber integration](https://github.com/webdriverio/wdio-cucumber-framework) and revisit the other framework integration we actually always send the title and parent of a test spec. So never mind the comment above.

christian-bromann:
@georgecrawford what do you think about the comment above? Would be great to get this merged and finish the junit reporter.

georgecrawford:
This branch has got confused. I've opened https://github.com/webdriverio/webdriverio/pull/991, and will address feedback there.

PR #898, opened 2015-11-23
Fixed the switchTab functionality not working 

### LINKED ISSUE #897: calling switchTab without an window handle is not working
Calling `browser.switchTab()` without a window handle fails to select the first available tab (or window).

Doing this:

    browser.url('http://github.com').newWindow('http://google.com').switchTab().url()

Results in returning `http://google.com` instead of the expected `http://github.com`.

This also results in the `.close()` function not selecting the first available tab as expected, since it relies on the `switchTab` functionality.

wvankuipers:
when calling it without a window handle, fixes #897

please note: I did not get around writing decent tests for this, will add them tomorrow.

### DIFF
--- a/lib/commands/switchTab.js
+++ b/lib/commands/switchTab.js
@@ -23,8 +23,8 @@ module.exports = function switchTab (windowHandle) {
     }
 
     return this.windowHandles().then(function(tabIds) {
-        if(tabIds && tabIds.length) {
-            return this.switchTab(tabIds[0]);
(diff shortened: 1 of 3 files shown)


christian-bromann:
A test would be great! Thanks!

wvankuipers:
@christian-bromann Tests are added and are now passing (and failing on master)

christian-bromann:
Thanks! Need to fix some issues with the build but try to publish this later today

PR #900, opened 2015-11-24
Add request timeout to init call

cyrusmith:
Call to init sometimes hangs forever in my testing environment. Wouldn't it be useful to be able to set request timeout?

### DIFF
--- a/lib/protocol/init.js
+++ b/lib/protocol/init.js
@@ -16,11 +16,15 @@ var ErrorHandler = require('../utils/ErrorHandler.js'),
     packageDotJson = require('../../package.json'),
     merge = require('deepmerge');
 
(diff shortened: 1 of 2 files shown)


christian-bromann:
What's up with your test environment?

christian-bromann:
ping @cyrusmith

christian-bromann:
closing due to inactivity. Don't really see why the init command should have a timeout.

jasonslyvia:
I facing the same issue here, a selenium grid is hung whatsoever, and my test script only exist when mocha timeout. In this case, I'd like to switch to an alternative grid.

But I guess a simple `setTimeout` call to wrap `.init()` should be sufficient.

PR #901, opened 2015-11-24
Move babel-runtime to runtime dependencies

just-boris:
As you can find from the `babel-runtime` name, it should be included in the `dependencies`, not `devDependencies`


christian-bromann:
Good catch :+1:

PR #917, opened 2015-12-02
Added selectByName action and fixed a bug

Risto-Stevcev:
Fixed a bug where `selectByVisibleText` did not properly escape quotes.

The bug is only specific to `selectByVisibleText` because element [attributes](http://www.w3.org/TR/html401/intro/sgmltut.html#h-3.2.2) (such as value or name) cannot have quotes inside them by default:

> The attribute value may only contain letters (a-z and A-Z), digits (0-9), hyphens (ASCII decimal 45), periods (ASCII decimal 46), underscores (ASCII decimal 95), and colons (ASCII decimal 58).

The only way to solve this issue in a satisfactory way is to use the Xpath [concat](http://www.w3.org/TR/xpath/#function-concat) function. A useful reference can be found [here](http://kushalm.com/the-perils-of-xpath-expressions-specifically-escaping-quotes).

Also added `selectByName` so that `option` elements can be selected by their `name` attribute.

### DIFF
--- a/lib/commands/selectByName.js
+++ b/lib/commands/selectByName.js
@@ -1,0 +1,78 @@
+/**
+ *
+ * Select option with a specific name.
+ *
+ * <example>
+    :example.html
+    <select id="selectbox">
+        <option name="someName0" value="someValue0">uno</option>
+        <option name="someName1" value="someValue1">dos</option>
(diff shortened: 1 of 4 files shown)


christian-bromann:
Thanks. Great PR :+1:

PR #918, opened 2015-12-03
emit test:end in Jasmine

jendagroovy:
Hi, the "test:end" event was not emitted when using Jasmine framework


christian-bromann:
:+1: good catch

PR #922, opened 2015-12-04
Cli enhancement

christian-bromann:
As we split up reporter and services from the main package it gets even more confusing when setting up WebdriverIO. The cli config interface should make this setup super easy. Therefor I did some tweaks in order to allow people to choose approved frameworks, services and reporters and give them the opportunity to install them right away. Example:

wdioconfig

### DIFF
--- a/gruntfile.js
+++ b/gruntfile.js
@@ -82,11 +82,15 @@ module.exports = function (grunt) {
             }
         },
         copy: {
-            main: {
+            packagejson: {
                 src: 'package.json',
-                dest: 'build/',
+                dest: 'build/'
             },
-        },
+            ejstemplate: {
+                src: 'lib/helpers/wdio.conf.ejs',
+                dest: 'build/'
+            }
+        }
     })
 
     require('load-grunt-tasks')(grunt)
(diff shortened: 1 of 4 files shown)


georgecrawford:
LGTM - looks like it's time to create the `wdio-selenium-standalone-service` repo. :smile:

georgecrawford:
Which I can't do: `You don’t appear to have permission to create repositories for this organization. Sorry about that.`

christian-bromann:
just upgraded membership ;-)

christian-bromann:
The problem I am facing is that the service don't have enough privileges in order to do stuff like this as they get initialised once the session is already created. We can't initialise them in the launcher because we wouldn't be able to propagate the service object to all runners. I also don't want to introduce another type of plugins which would be something like `wdio-xxx-launcher`. I have to see what's the best solution here.

georgecrawford:
Hmm. Good point. How about services act like `wdio.conf.js` - i.e. they can tap in to the *launcher* (with `onPrepare` and `onComplete` methods, which would be used by `wdio-selenium-standalone-service`) *and/or runners* (like `wdio-sauce-service`) but with the understanding that only JSON will be passed from launcher to runner, and not live service objects.

PR #925, opened 2015-12-07
Fix rendering of session param docs

klamping:
The parameter rendering for [the `session` docs is misformatted](http://webdriver.io/api/protocol/session.html):
image

I haven't tested this locally; just went off of what the JSDoc spec mentions for [default params](http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values).


christian-bromann:
Oh sweet, thanks

PR #928, opened 2015-12-08
Stale element retry

georgecrawford:
@christian-bromann if you are happy with the approach here, I will add in exception handling for the other commands which make use of `elements`, etc.

### DIFF
--- a/lib/commands/waitUntil.js
+++ b/lib/commands/waitUntil.js
@@ -34,7 +34,7 @@
  *
  */
 
-import { CommandError } from '../utils/ErrorHandler'
+import { CommandError, captureCommandCall } from '../utils/ErrorHandler'
 
 const REJECT_MESSAGE = 'Promise never resolved with an truthy value'
 
@@ -77,8 +77,11 @@ let waitUntil = function (condition, ms, starttime = new Date().getTime()) {
             resolve(res)
         }, (err) => {
             clearTimeout(timeout)
-            reject(new CommandError(`Promise was fulfilled but got rejected with the following reason: ${err}`))
+            reject(new CommandError(`Promise was fulfilled but got rejected with the following reason: ${err}`, err))
         })
+    })
+    .catch(err => {
+        throw new CommandError(err.message, err, captureCommandCall(arguments))
     })
 }
 
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -138,9 +138,23 @@ class Runner {
     }
 
     addCommandHooks (config) {
-        config.afterCommand.push((command, args, result, err) => {
-            if (err && err.seleniumStack && err.seleniumStack.type === 'StaleElementReference') {
-                err.wdioExceptionHandler = global.browser[command].bind(global.browser, ...args)
+        config.afterCommand.push((command, args, result, e) => {
+            if (e) {
+                let error = e
+                let commandCall
+                while (error) {
+                    if (!commandCall && error.commandCall) {
+                        commandCall = error.commandCall
+                    }
+                    if (error.seleniumStack && error.seleniumStack.type === 'StaleElementReference') {
+                        if (commandCall) {
+                            global.browser.logger.log(`Caught StaleElementReference in command [${commandCall.name}]. Will re-run with arguments: ${JSON.stringify(commandCall.args)}`)
+                            e.wdioExceptionHandler = global.browser[commandCall.name].bind(global.browser, ...commandCall.args)
+                        }
+                        break
+                    }
+                    error = error.previousError
+                }
             }
         })
     }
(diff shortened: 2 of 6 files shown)


georgecrawford:
Damn, this isn't quite working. The outer exception (e.g. `waitForVisible`) is caught [here](https://github.com/webdriverio/webdriverio/pull/928/files#diff-3b660a335cd0d5e1c680f02c2a2a8c70R175) too late - just *after* the catch [here](https://github.com/webdriverio/wdio-sync/pull/22/files#diff-168726dbe96b3ce427e7fedce31bb0bcR103), so there is no `activeCommand` set on the first exception in the chain handled by the `afterCommand` hook. Can you think of any way I can get that information onto the exception?


example output:
```
waitForVisible throwing now
Setting activeCommand { name: 'waitUntil', args: [ [Function], 40000 ] }
103 'Exception caught now'
Depth 0, error Error: Promise was fulfilled but got rejected with the following reason: Error: stale element reference: element is not attached to the page document, error.activeCommand: undefined
Depth 1, error Error: Promise was fulfilled but got rejected with the following reason: Error: stale element reference: element is not attached to the page document, error.activeCommand: {"name":"waitUntil","args":[null,40000]}
Depth 2, error Error: stale element reference: element is not attached to the page document, error.activeCommand: {"name":"isVisible","args":["#splash"]}
Depth 3, error Error: stale element reference: element is not attached to the page document, error.activeCommand: {"name":"elementIdDisplayed","args":["1"]}
Depth 3, Caught StaleElementReference
Will re-run with arguments: [null,40000]
[22:34:34]:  Caught StaleElementReference in command waitUntil. Will re-run with arguments: [null,40000]
Setting activeCommand { name: 'waitForVisible', args: [ '#splash', 40000, true ] }
```
Note that although `waitForVisible` is the executed command, the re-run is performed with the next command in the exception chain: `Caught StaleElementReference in command waitUntil. Will re-run with arguments: [null,40000]`

georgecrawford:
Also now requires https://github.com/webdriverio/wdio-mocha-framework/pull/11

georgecrawford:
So I originally wanted to automatically add command details (name, args) when an exception was thrown, with a `catch()` here: https://github.com/webdriverio/webdriverio/blob/v4.0/lib/webdriverio.js#L174. It was great, but didn't have a chance to add the details for a command before it was caught in wdio-sync, which is where we run the `afterComand` hook.

So instead it's more of a manual process, and each command has to be instrumented, like https://github.com/webdriverio/webdriverio/pull/928/files#diff-71f548ea1a1d9567f29827805e5bb64e.

I added a new `onCommandException` hook [here](https://github.com/webdriverio/webdriverio/commit/9ede2520a0ab1da55ccbbe89bed15615b1d8f695) so that the `afterCommand` hook wasn't used, as this implies that the command has finished, when the `beforeCommand` hook didn't actually run for a retried inner command. Having mismatch between `beforeCommand` and `afterCommand` is confusing, and breaks the reporting in Allure, for example. This fixes it, as the `onCommandException` hook doesn't fire an event, so reporters don't see it.

lib/commands/waitUntil.js
```diff
@@ -77,9 +77,12 @@ let waitUntil = function (condition, ms, starttime = new Date().getTime()) {
             resolve(res)
         }, (err) => {
             clearTimeout(timeout)
-            reject(new CommandError(`Promise was fulfilled but got rejected with the following reason: ${err}`))
+            reject(new CommandError(`Promise was fulfilled but got rejected with the following reason: ${err}`, err))
```
christian-bromann: This change is obsolete. `reject` only takes one parameter.

lib/runner.js
```diff
@@ -135,6 +137,28 @@ class Runner {
         process.exit(failures === 0 ? 0 : 1)
     }
 
+    addCommandHooks (config) {
+        config.onCommandException.push((command, args, result, e) => {
+            if (e) {
```
christian-bromann: Let's save some indentation and do `if (!e) { return }`

lib/runner.js
```diff
@@ -135,6 +137,28 @@ class Runner {
...
+                let error = e
+                let commandCall
+                while (error) {
+                    if (!commandCall && error.commandCall) {
+                        commandCall = error.commandCall
+                    }
+                    if (error.seleniumStack && error.seleniumStack.type === 'StaleElementReference') {
```
christian-bromann: Let's also save some indentation here

lib/commands/waitUntil.js
georgecrawford: It's the second param to `CommandError`

christian-bromann:
I think we can close this since I got #969 finally working.

PR #932, opened 2015-12-10
Accept custom reporter as constructor

just-boris:
Example with the custom reporter doesn't work in the branch `4.0`.

I am going to use it to test my Allure integration with Webdriver.io, so it is matters for me.

So, I've fixed it.

### DIFF
--- a/lib/launcher.js
+++ b/lib/launcher.js
@@ -47,13 +47,18 @@ class Launcher {
             return reporter
         }
 
-        for (let reporterType of config.reporter) {
-            try {
-                let Reporter = require(`wdio-${reporterType}-reporter`)
-                reporter.add(new Reporter(reporter, config))
-            } catch (e) {
-                throw new Error(`reporter "wdio-${reporterType}-reporter" is not installed. Error: ${e.stack}`)
+        for (let Reporter of config.reporter) {
+            if (typeof Reporter === 'string') {
+                try {
+                    Reporter = require(`wdio-${Reporter}-reporter`)
+                } catch (e) {
+                    throw new Error(`reporter "wdio-${Reporter}-reporter" is not installed. Error: ${e.stack}`)
+                }
             }
+            if (typeof Reporter !== 'function') {
+                throw new Error(`reporter option must be either string or function, but got: ${Reporter}`)
+            }
+            reporter.add(new Reporter(reporter, config))
         }
 
         return reporter


christian-bromann:
Why does the current code not work for you?

just-boris:
This configuration from example doesn't work: https://github.com/webdriverio/webdriverio/blob/master/examples/wdio/wdio.customreporter.conf.js#L39 

It always loads npm-module, but I want to pass function here.

christian-bromann:
How about changing that line to
```js
reporter: [__dirname + ''./reporter/my.custom.reporter'']
```
or something like this

just-boris:
It doesn't work too. Reporter name is always embraced as `wdio-${reporterType}-reporter`

lib/launcher.js
```diff
@@ -47,13 +47,15 @@ class Launcher {
...
-                throw new Error(`reporter "wdio-${reporterType}-reporter" is not installed. Error: ${e.stack}`)
+        for (let Reporter of config.reporter) {
+            if (typeof Reporter === 'string') {
+                try {
+                    Reporter = require(`wdio-${Reporter}-reporter`)
+                } catch (e) {
+                    throw new Error(`reporter "wdio-${Reporter}-reporter" is not installed. Error: ${e.stack}`)
+                }
             }
+            reporter.add(new Reporter(reporter, config))
```
christian-bromann: I don't think this is the right way we should implement this. What if `Reporter` has a spelling issue and is still a string? How about we do an `else if` and check if typeof `reporterType` is function?

just-boris:
Fixed that.
If reporter is not installed or has a spelling issue, the error will be thrown.
If the option has unknown type, error will be thrown too. Checked that on my local setup.

lib/launcher.js
```diff
@@ -47,13 +47,18 @@ class Launcher {
...
+        for (let Reporter of config.reporter) {
+            if (typeof Reporter === 'string') {
+                try {
+                    Reporter = require(`wdio-${Reporter}-reporter`)
+                } catch (e) {
+                    throw new Error(`reporter "wdio-${Reporter}-reporter" is not installed. Error: ${e.stack}`)
+                }
             }
+            if (typeof Reporter !== 'function') {
+                throw new Error(`reporter option must be either string or function, but got: ${Reporter}`)
```
christian-bromann: What do you think about: `...string or function, but got: ${typeof Reporter}')`?

just-boris: `typeof null == 'object'`, so sometimes that message can not be informative.

just-boris: Usually, most values are converted to string as `[object Object]` and I considered that is would be good to explain the error.

christian-bromann: true.

christian-bromann:
thanks @just-boris

PR #936, opened 2015-12-11
Reporter options split into generic and reporter-type-specific

### LINKED ISSUE #935: Change reporterOptions to be per-reporter
I propose that `reporterOptions` is likely to get unwieldy with multiple reporters, as you may want the same key name (e.g. `outputDir`) to be set with different values depending on the reporter. We could continue to support this in a backwards-compatible way, but also support a new format:
```js
reporterOptions: {
    allure: {
        outputDir: 'blah-allure'
    },
    junit: {
        outputDir: 'blah-junit'
    }
}
```

georgecrawford:
Fixes #935


lib/launcher.js
```diff
@@ -47,7 +47,25 @@ class Launcher {
             return reporter
         }
 
+        const baseReporterOptions = config.reporterOptions || {}
+
         for (let reporterType of config.reporter) {
+            config.reporterOptions = {}
```
christian-bromann: I am afraid this will overwrite itself in the reporter will still be the value of `baseReporterOptions` after one event loop cycle since we passing an object per reference. So when the first event occurs the config in the reporter will still be the old value (not sure though). How about we do:

```js
let reporterOptions = Object.assign(config.reporterOptions);
// then delete other reporter options
```

christian-bromann:
How should we handle custom reporter (reporter that are not pushed to NPM see #932 )?

georgecrawford:
I had no idea about custom reporters

georgecrawford:
I addressed custom reporters in https://github.com/webdriverio/webdriverio/commit/2be3002871e00fb72c2958705fb2ad3680ad97e3.

Changes introduced in this PR, which need fixing in the examples:
- custom reporters must export a `reporterName` string property
- reporter options are passed as the third param to the reporter constructor, rather than from browser config object

Happy with this?

lib/launcher.js
```diff
@@ -43,22 +42,51 @@ class Launcher {
...
                 } catch (e) {
-                    throw new Error(`reporter "wdio-${Reporter}-reporter" is not installed. Error: ${e.stack}`)
+                    throw new Error(`reporter "wdio-${reporterName}-reporter" is not installed. Error: ${e.stack}`)
                 }
+                reporters[reporterName] = Reporter
             }
-            if (typeof Reporter !== 'function') {
-                throw new Error(`reporter option must be either an array of strings or functions, but got: ${Reporter}`)
+            if (!Reporter) {
+                throw new Error(`reporter option must be either an array of strings or functions, but got: ${reporterName}`)
```
christian-bromann: please replace "_reporter option_" with "_reporter_ __s__ _option_" ;-)

lib/launcher.js
```diff
@@ -43,22 +42,51 @@ class Launcher {
...
+
+        for (let reporterName in reporters) {
+            const Reporter = reporters[reporterName]
+            const reporterOptions = {}
+            for (let option of Object.keys(config.reporterOptions)) {
+                if (option === reporterName && typeof config.reporterOptions[reporterName] === 'object') {
+                    // Copy over options specifically for this reporter type
+                    for (let reporterOption of Object.keys(config.reporterOptions[reporterName])) {
+                        reporterOptions[reporterOption] = config.reporterOptions[reporterName][reporterOption]
+                    }
```
christian-bromann: Can we use Object.assign here?
```js
reporterOptions = Object.assign(reporterOptions, config.reporterOptions[reporterName])
```

christian-bromann:
Two tiny things, other than that LGTM!

georgecrawford:
OK?

christian-bromann:
Yes

ponmudivn:
@georgecrawford 
Is is possible to have dynamic reporter option based on some cli argument?
plz see my comments in https://github.com/webdriverio/webdriverio/issues/1174#issuecomment-280563717

monolithed:
There's one thing I want to say. The current configuration system does not respect any scoped packages:

**package.json**

```json
{
     "name":  "@foo/wdio-service-bar"
}
```

Of course it's possible to import explicitly:

```js
import fooService from '@foo/wdio-service-bar';

{
    services: [ fooService ]
}
```

Nevertheless, it would be nice to use following scheme:

```js
{
    services: [
       {
           scope: 'foo',
           name: 'bar'
       }
    ]
}
```

✌️

PR #937, opened 2015-12-12
strict array type checking removed

udnisap:
I am using arguments and since arguments are not arrays it fails.


christian-bromann:
You can easily convert your arguments object into an array. I guess this is the proper way to solve the problem.

udnisap:
In those instances it only needs to have `forEach` in the called variables. If the object has it we would be all good right?

christian-bromann:
Probably but it wouldn't be a good pattern to follow. Please transform your arguments object into an actual array and pass it as argument.

PR #939, opened 2015-12-15
Update gettingstarted.md

bwafflemaker:
Changed `quick` to `quickly` since it is being used as an adverb in this instance.


christian-bromann:
:+1:

PR #944, opened 2015-12-16
Updated Appium commands

christian-bromann:
Not sure if this is interesting for you @georgecrawford to review. It is a whole bunch of new and modified appium commands. I went through the whole mjsonwp list and check the driver where these commands are actually implemented. I think it will be nice to have a label next to the command that says where people can use the commands (Android emulator allow a lot more automation that iOS simulator). I created an issue (webdriverio/webdriver.io#37) so I will get remembered when I polish up the website.

Gonna move forward and merge right away. Let me know if you have any questions or feedback.


georgecrawford:
:+1:

PR #947, opened 2015-12-17
Remove docs on optional selector

klamping:
`moveToObject` requires a selector to work, so getting rid of the documentation that says otherwise.


christian-bromann:
:+1:

PR #952, opened 2015-12-18
Support retries in RequestHandler

### LINKED ISSUE #946: Add option to retry failed requests
We write our tests using webdriverio. But unfortunately our selenium grid is not stable enough or we experience network problems. The problem is that webdriverio can't get SessionId for other requests and mocha tests fail after timeout.

Currently `request` library is used to send requests to SG, but it doesn't support retries (or does it?) and it would be very helpful in our case. Would you mind if I make a PR which adds this option (retries for example) and also switches to `got`, `asker` etc?

1999:
I added 3 retries into RequestHandler which fixes #946
Unfortunately I can't run tests on my mac - selenium-standalone seems to be broken.
Probably constants are a bad design here - should've I used them as an option to runner?


lib/utils/RequestHandler.js
```diff
@@ -12,8 +12,57 @@ var Q = require('q'),
     ErrorHandler = require('./ErrorHandler'),
     packageDotJson  = require('../../package.json');
 
+var REQUEST_TIMEOUT = 10000;
+var REQUEST_RETRY_COUNT = 3;
```
christian-bromann: Can you make this user configurable?

christian-bromann:
One comment otherwise LGTM!

lib/utils/RequestHandler.js
1999: I'll take a deeper look on how I can do this. Need 4 more days, sorry :(

1999:
Didn't understand about one comment :(

1999:
Just added configurable options and updated PR. Take a look please.

christian-bromann:
Could you add a functional test too? You could use the nock library to block a random request twice and get it resolve the 3rd time.

1999:
Just tried nock for the first time, seems like everything is working as expected (?)

1999:
:eyes:

christian-bromann:
Looking good, thanks!

christian-bromann:
@1999 your PR seems to break the build? Can you check why?

christian-bromann:
I am afraid I can't port this to the v4.0 branch. It caused failures with Sauce Connect.

1999:
Can I try to port this myself or it's not possible at all?

christian-bromann:
Sure you can create a PR for the v4.0 branch.

1999:
okay, I'll have a closer look at this branch. When do you plan to release it btw?

PR #963, opened 2015-12-22
Feature/issue 962

### LINKED ISSUE #962: WaitUntil produces unexpected results
The `WaitUntil` command produces unexpected results when used multiple in tests.

Example:

     this.waitUntil(function() { [do something that will passe in 1000 ms] }, 1000);

Results in the correct success state.

    this.waitUntil(function() { [do something that will passe in 1000 ms] }, 1000);
    this.waitUntil(function() { [do something that will passe in 2000 ms] }, 2000);
    this.waitUntil(function() { [do something that will passe in 3000 ms] }, 3000);

Results in a incorrect error state.

I have pinpointed the problem to the following lines (78, 79) in the file `lib/commands/waitUntil.js`

     return defer.resolve(self.pause(interval)
                .then(waitUntilPrivate.bind(self, condition, timeout, interval, starttime)));

What will happen is that the `.bind()` function is executed before the self.pause(interval) is finished. This causes the tests to fail.

I will create a pull request that resolves this issue.

wvankuipers:
Fixes #962

### DIFF
--- a/lib/commands/waitUntil.js
+++ b/lib/commands/waitUntil.js
@@ -74,21 +74,21 @@ function waitUntilPrivate(condition, timeout, interval, starttime) {
             if(q.isPromiseAlike(condition)) {
                 defer.reject(new ErrorHandler.CommandError('Promise was fulfilled with a falsy value'));
             }
+
             return defer.resolve(self.pause(interval)
-                .then(waitUntilPrivate.bind(self, condition, timeout, interval, starttime)));
+                .then(function() {
+                    waitUntilPrivate.call(self, condition, timeout, interval, starttime);
+                }));
         }
 
         defer.resolve(res);
-
     }, function(err) {
         clearTimeout(timeoutId);
 
         defer.reject(new ErrorHandler.CommandError('Promise was fulfilled but got rejected with the following reason: ' + err));
-
     });
 
     return defer.promise;
-
 }
 
 module.exports = function waitUntil(condition, timeout, interval) {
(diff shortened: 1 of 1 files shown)


christian-bromann:
Is there a test case we can apply to make sure that this won't happen again?

wvankuipers:
I will look into that, will add it to this PR

christian-bromann:
that would be awesome thanks!

wvankuipers:
While creating the tests I could not come up with a reproducible test case, so I looked into the issue a but further.

Running the `waitForExist` test alone resulted in a correct state, running my complete test case (bases on the cucumber-boilerplate) resulted in a false state. So I started disabling tests and found ou the problem was in the login.feature

    Scenario: login with fake credentials
        Given I open the url "https://github.com/"
        When  I log in to site with username "marketionist" and password "1234"
        Then  I expect that element "#js-flash-container .flash-error" is visible

This executes the next script:

    this.browser
        .timeoutsImplicitWait(5000)
        .click(githubPage.signInButton)
        .setValue(githubPage.loginInput, login)
        .setValue(githubPage.passwordInput, password)
        .click(githubPage.signIn2Button)
        .call(done);

The problem is with the `timeoutsImplicitWait` being used here, this makes the `waitFor` command timeout for some reason. If I do one off the following all goes well:

    this.browser
        .pause(5000)
        .click(githubPage.signInButton)
        .setValue(githubPage.loginInput, login)
        .setValue(githubPage.passwordInput, password)
        .click(githubPage.signIn2Button)
        .call(done);

    this.browser
        .timeoutsImplicitWait(5000)
        .click(githubPage.signInButton)
        .setValue(githubPage.loginInput, login)
        .setValue(githubPage.passwordInput, password)
        .click(githubPage.signIn2Button)
        .timeoutsImplicitWait(250)
        .call(done);

Is this expected behaviour or is there indeed an issue with this case?

christian-bromann:
:+1:

PR #964, opened 2015-12-22
How to contribute documentation changes, a few examples added

fireproofsocks:
1. Added a section to the CONTRIBUTING.md file that attempts to outline how the community can submit updates to the documentation.  (Also applied the recommended style guide to that page).

2. Included examples with sample output for several functions including getCurrentTabId, title, windowHandle, and windowHandles.

Note that currently the `@return` declarations are sometimes inaccurately typed and are not visible anywhere in the generated documentation plages.  This makes it more critical that the examples include sample output.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,9 +1,13 @@
-### We love pull requests. Here's a quick guide:
+# Contributing
+
+## Code
+
+We love pull requests. Here's a quick guide:
 
 1. Fork the repo.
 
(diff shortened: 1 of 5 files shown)


christian-bromann:
@fireproofsocks YES! :tada: 
This is a great step towards better documentation. Thanks so much!
If you are wondering the documentation generator is here: https://github.com/webdriverio/wddoc and yes there is not much documentation there too :-/

Some of your edits will change with v4 but this is ok. I will run over each md file in the final process anyway.

Issue #951, opened 2015-12-18
switchTab(<handle>) and window(<handle>) fail if window is not opened explicitly with newWindow() 


fireproofsocks:
Given a site with 2 pages, main.html and secondary.html, Webdriver seems unable to switch between the two unless the secondary window was opened via the newWindow() command.

The following code demonstrates the problem:

```
// localhost has 2 pages: main.html (parent) and secondary.html
var main;
var secondary;
browser
.url('http://localhost/main.html')
.pause(1000)
.getTitle()
.then(function(title) {
    console.log('On window: ' + title);  // Main
})
.getCurrentTabId().then(function(ret){
    main = ret;
})
.click('span*=Popup') // Open the secondary window via a popup
.pause(1000)
.getTabIds().then(function (handles) {
    secondary = handles[handles.length - 1];
})
.switchTab(secondary)
.pause(1000)
.getTitle()
.then(function(title) {
    console.log('On window: ' + title);  // Expected: Secondary, Actual: Main!
})
.switchTab(main)
.pause(1000)
.getTitle()
.then(function(title) {
    console.log('On window: ' + title); // Main
})
.pause(1000)
.call(done);
});
```

However, if I explicitly open the secondary window via newWindow, the following code works:
```
// localhost has 2 pages: main.html (parent) and secondary.html
// This flow works: open the main window, then open a 2nd window via newWindow(),
// then return to the main window.
var main;
var secondary;
browser
.url('http://localhost/main.html')
.pause(1000)
.getCurrentTabId().then(function(ret){
    main = ret;
})
.getTitle()
.then(function(title) {
    console.log('On window: ' + title); // Main
})
.newWindow('http://testing.app/secondary.html')
.pause(1000)
.getTabIds(function (err, handles) {
    secondary = handles[handles.length - 1];
})
.getTitle()
.then(function(title) {
    console.log('On window: ' + title);  // Secondary (via newWindow())
})
.switchTab(main)
.getTitle()
.then(function(title) {
    console.log('On window: ' + title); // Main
})
.call(done);
```

- node version: 4.2.1
- wdio version: Version 3.3.0
- using selenium-standalone 4.8.0
- browser: chrome, safari, phantomjs

fireproofsocks:
I submitted a PR including failing tests for this issue: https://github.com/webdriverio/webdriverio/pull/965

christian-bromann:
Let's continue our conversation in #965 then.

PR #965, opened 2015-12-22
Added (failing) tests for windows/tabs opened via Javascript

fireproofsocks:
I'm submitting this pull request to share the failing tests I added for demonstrating issue https://github.com/webdriverio/webdriverio/issues/951

If there's an issue with the syntax that can be corrected to make the tests pass, I'd love to see it (and I can add it to the docs).  If this demonstrates a legitimate bug, then please feel free to reference the included tests to reproduce the error.


christian-bromann:
@fireproofsocks thanks for submitting this PR. You've posted the following code:

```js
// localhost has 2 pages: main.html (parent) and secondary.html
var main;
var secondary;
browser
.url('http://localhost/main.html')
.pause(1000)
.getTitle()
.then(function(title) {
    console.log('On window: ' + title);  // Main
})
.getCurrentTabId().then(function(ret){
    main = ret;
})
.click('span*=Popup') // Open the secondary window via a popup
.pause(1000)
.getTabIds().then(function (handles) {
    secondary = handles[handles.length - 1];
})
.switchTab(secondary)
.pause(1000)
.getTitle()
.then(function(title) {
    console.log('On window: ' + title);  // Expected: Secondary, Actual: Main!
})
.switchTab(main)
.pause(1000)
.getTitle()
.then(function(title) {
    console.log('On window: ' + title); // Main
})
.pause(1000)
.call(done);
});
```

I am note sure but if I remember correctly I was also confused by that. I think if you call getTitle on a window you switched to you will still get the old window title of the other tab (not 100% sure though). This seems to be a Selenium issue if turns out to be true. However instead of using getTitle could you try to query an element that only exists on the second page?

fireproofsocks:
Yeah, I'm afraid it might be a Selenium issue... take a look at the tests in that branch for switchTab() -- it relies on the tab id, and it does not change.  In my own project I've tried to select stuff in the popup or populate form fields in there and it fails, but I think it all boils down to the included tests.

christian-bromann:
@fireproofsocks we updated the code base (merged v4 branch). Could you please fetch a fresh copy from the master branch and open a new PR? Thanks!

fireproofsocks:
Errr.... can you walk me through this?  Sorry, my Git shui is not good enough to follow how to do this without completely redoing all my work.

mselvanambi:
Guys, 
I still face the same issue of switching tab not working with  web driver io .
Can someone tel me was there solution already in place for this  or something an issue with the selenium itself ??
Thankyou.

PR #966, opened 2015-12-22
Add doc lines for waitforInterval

mjhm:


christian-bromann:
Thanks. Not sure why I missed that there.

PR #967, opened 2015-12-23
E2e into es2016

christian-bromann:
Rewritten complete test suite of WebdriverIO. The setup is the same as in the other packages with one difference: we run standard mocha tests in dev and coverage mocha tests in ci only. I splitted the suites into desktop, functional, multibrowser, android, ios. I also used the time to improve the documentation of some commands while testing them.
I want to move forward with mobile command optimisations so I again merge this PR right away. If you (@georgecrawford) have any questions or comments let me know. I missed the whole documentation part here but want to tackle this as a whole at the end before release.


georgecrawford:
I'm having a few issues with this:
- I think the `npm test...` tasks need updating
- `npm coverage` is also broken, so Travis shouldn't pass as far as I can see
- `isSeleniumServerRunning = isSeleniumServerRunning === undefined ? true : isSeleniumServerRunning` - is that what you mean? I would expect `grunt test` to run `mochaTest:desktop` and to start the selenium server automatically for me (therefore `isSeleniumServerRunning` should default to `false`). If I'm wrong, could you paste the example CLI commands you're using for different types of testing?
- the tests aren't running correctly for me. Limiting to a single test gives:
```
Running "mochaTest:desktop" (mochaTest) task


  addValue
    1) add a value to existing input value


  0 passing (2m)
  1 failing

  1) addValue add a value to existing input value:
     Error: timeout of 120000ms exceeded. Ensure the done() callback is being called in this test.
```
Any ideas?

georgecrawford:
OK, that last error was my fault - I'd forgotten to start the HTTP server on 8080. Can this be done automatically in grunt? Also - best to add a warning if a process is already running on 8080, which was the problem for me :smile:

georgecrawford:
`stop-selenium-server:dev` does NOT run if tests fail for some reason, leaving the server running, and preventing the next test run from starting until selenium is killed. Can this task be run even if tests fail?

PR #969, opened 2015-12-24
Onerror hook (includes staleElementRetry)

christian-bromann:
please review @georgecrawford 

This is my approach to retry stale element errors. Before the command promise gets resolved I basically resolve it with another promise that executes a set of `onError` handlers. We can now have internal error handlers to interfere on a variety on problems in the Selenium world. I really appreciate your effort you put into solving this issue but if it's okay for you I would like to go with this one as it comes with the following advantages:

- all necessary changes were made webdriverio.js
- no need to attach an error handler in each command or modify error objects
- works also without wdio test runner

Let me know what do you think?

### DIFF
--- a/gruntfile.js
+++ b/gruntfile.js
@@ -87,6 +87,10 @@ module.exports = function (grunt) {
             unit: {
                 src: ['./test/setup-unit.js', 'test/spec/unit/*.js'],
                 options: mochaInstanbulOpts
+            },
+            wdio: {
+                src: ['./test/setup-unit.js', 'test/spec/wdio/*.js'],
+                options: mochaInstanbulOpts
             }
         },
         mochaTest: {
@@ -112,6 +116,10 @@ module.exports = function (grunt) {
             },
             unit: {
                 src: ['./test/setup-unit.js', 'test/spec/unit/*.js'],
+                options: mochaOpts
+            },
+            wdio: {
+                src: ['./test/setup-unit.js', 'test/spec/wdio/*.js'],
                 options: mochaOpts
             }
         },
--- a/lib/helpers/errorHandler.js
+++ b/lib/helpers/errorHandler.js
@@ -3,7 +3,7 @@ export default [
      * stale reference error handler
      */
     function (e) {
-        if (!e.message.match('Element is no longer attached to the DOM')) {
+        if (!e.seleniumStack || e.seleniumStack.type !== 'StaleElementReference') {
             return
         }
 
--- a/lib/helpers/errorHandler.js
+++ b/lib/helpers/errorHandler.js
@@ -2,14 +2,11 @@ export default [
     /**
      * stale reference error handler
      */
-    function (e, command) {
+    function (e) {
         if (!e.message.match('Element is no longer attached to the DOM')) {
             return
         }
-
-        console.log(this.commandList);
-
-        console.log('stale element error handler')
-        return Promise.reject(new Error('buhh'))
+        return Promise.reject(new Error('buhhhh'))
+        return this.getTitle()
     }
 ]
--- a/lib/helpers/errorHandler.js
+++ b/lib/helpers/errorHandler.js
@@ -1,0 +1,15 @@
+export default [
+    /**
+     * stale reference error handler
+     */
+    function (e, command) {
+        if (!e.message.match('Element is no longer attached to the DOM')) {
+            return
+        }
+
+        console.log(this.commandList);
+
+        console.log('stale element error handler')
+        return Promise.reject(new Error('buhh'))
+    }
+]
--- a/lib/helpers/errorHandler.js
+++ b/lib/helpers/errorHandler.js
@@ -35,7 +35,6 @@ export default [
             return
         }
 
-        console.log('repeat', commandToRepeat.name, '(', commandToRepeat.args[0], ')')
         return this[commandToRepeat.name].apply(this, commandToRepeat.args)
     }
 ]
--- a/lib/helpers/errorHandler.js
+++ b/lib/helpers/errorHandler.js
@@ -6,7 +6,36 @@ export default [
         if (!e.message.match('Element is no longer attached to the DOM')) {
             return
         }
-        return Promise.reject(new Error('buhhhh'))
-        return this.getTitle()
+
+        /**
+         * get through command list and find most recent command where an element(s)
+         * command contained the failing json web element
+         */
+        let failingCommand = this.commandList.slice(-1)[0]
+
+        let cnt = this.commandList.length
+        let commandToRepeat
+        for (let command of this.commandList.reverse()) {
+            cnt--
+            if (command.name !== 'element' && command.name !== 'elements') {
+                continue
+            }
+            if (command.name === 'element' && (!command.result[0].value || command.result[0].value.ELEMENT !== failingCommand.args[0])) {
+                continue
+            }
+
+            for (let result of command.result.value) {
+                if (result.ELEMENT === failingCommand.args[0]) {
+                    commandToRepeat = this.commandList[--cnt]
+                }
+            }
+        }
+
+        if (!commandToRepeat) {
+            return
+        }
+
+        console.log('repeat', commandToRepeat.name, '(', commandToRepeat.args[0], ')')
+        return this[commandToRepeat.name].apply(this, commandToRepeat.args)
     }
 ]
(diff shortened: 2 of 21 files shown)


georgecrawford:
@christian-bromann I found that I was getting a different stale element error string (`element is not attached to the page document` rather than `Element is no longer attached to the DOM`), so I think it's safer to match by Selenium error code.

With my test case at https://gist.github.com/georgecrawford/96bc6308a21760a64125, the basic test now passes, but there's more work to do to get a stale element exception to be retried within a custom command. I tried debugging this, but couldn't quite work out how to fix it.

georgecrawford:
OK, I've added some functional tests (the same tests, pretty much, as in https://github.com/webdriverio/webdriverio/pull/857). The first fails genuinely, but I don't know why the onerror work you've done doesn't catch the error and repeat here. The second test is broken, as I'm not sure how to combine `Promise.all()` in a custom command and async/await in the test to ensure the commands are waited for. But when the test is fixed, I'd expect it to also fail until the issue with the first test is addressed.

georgecrawford:
I did a bit more debugging. Test case 1 (`does not throw staleElementReference exception when getting visibility of elements rapidly removed from DOM`), passes for me when I use the `wdio` runner, but not in the functional test as committed in this PR.

Using `wdio`, when a stale element exception is thrown, it hits [this line](https://github.com/webdriverio/webdriverio/blob/onerror-hook/lib/webdriverio.js#L295), and is correctly detected at https://github.com/webdriverio/webdriverio/blob/onerror-hook/lib/webdriverio.js#L118.

When *not* using `wdio`, we hit the [same check](https://github.com/webdriverio/webdriverio/blob/onerror-hook/lib/webdriverio.js#L295), and the following is true:
`onRejected.toString()` is:
```js
function (a){
    return fn.call(that, a);
}
```
and `result.toString()` is:
```js
function () { [native code] }
```
Whatever this native code is (I don't know), it's run [here](https://github.com/webdriverio/webdriverio/blob/onerror-hook/lib/webdriverio.js#L112), and results in an exception being thrown which *isn't caught even by a `try...catch` around the `call()`*. The exception bubbles right up to the mocha test at https://github.com/webdriverio/webdriverio/pull/969/files#diff-8606da0fd6a702523ae78184fc14536fR7, and fails the test there.

I'm lost, but I expect you understand what's going on.

georgecrawford:
More debugging. I'm afraid this onerror hook also doesn't work for a real-world use case. 

My typical use case is to call `browser.waitForVisible('#splash', 10000, true)` - i.e. **wait up to 10 seconds for the splash screen to disappear**. My debug log looks like this:
```
EXECUTING COMMAND waitForVisible [ '#splash', 10000, true ]
EXECUTING COMMAND waitUntil [ [Function], 10000 ]
EXECUTING COMMAND isVisible [ '#splash' ]
EXECUTING COMMAND elements [ '#splash' ]
[13:21:59]:  RESULT		[{"ELEMENT":"0"}]
EXECUTING COMMAND elementIdDisplayed [ '0' ]
[13:21:59]:  RESULT		true
EXECUTING COMMAND pause [ 250 ]

... (many repeats)

EXECUTING COMMAND waitUntil [ [Function], 10000, 1452259319203 ]
EXECUTING COMMAND isVisible [ '#splash' ]
EXECUTING COMMAND elements [ '#splash' ]
[13:22:05]:  RESULT		[{"ELEMENT":"0"}]
EXECUTING COMMAND elementIdDisplayed [ '0' ]
ACTUAL SELENIUM ERROR stale element reference: element is not attached to the page document
errorHandler 6 StaleElementReference
errorHandler 17 failingCommand { name: 'elementIdDisplayed', args: [ '0' ] }
errorHandler 35 commandToRepeat: { name: 'isVisible', args: [ '#splash' ] }
EXECUTING COMMAND isVisible [ '#splash' ]
EXECUTING COMMAND elements [ '#splash' ]
[13:22:08]:  RESULT		[]
commandToRepeat returned result:  false
errorHandler 6 Promise was fulfilled but got rejected with the following reason: Error: stale element reference: element is not attached to the page document
errorHandler 8 returning now
errorHandler 6 Promise was fulfilled but got rejected with the following reason: Error: stale element reference: element is not attached to the page document
errorHandler 8 returning now
```
The runner then just hangs. So there's something about the re-running of `isVisible` which doesn't correctly satisfy `waitUntil`, and therefore `waitForVisible`.

christian-bromann:
Thanks for investigating. I will have to take a look.

christian-bromann:
I was able to reproduce the problem. It seems that if the stale element error happens twice in a row (meaning if the error handler also returns with that error) it doesn't get handled properly. The command resolves but with an undefined value and it fails due to

```
TypeError: Cannot read property 'value' of undefined
    at Object.exports.default (/build/lib/helpers/errorHandler.js:42:61)
```

christian-bromann:
I think I found the problem. The `isVisible` command uses the `unify` helper. That one seems to return a promise where the catch method is set to reject the promise.all helper. If the reject method was passed into the then method we assume that the user handled the error by himself so the onerror hook is not emitted. I will have to see how I can get around that without having an assertion of the promise source code.

christian-bromann:
fyi: if you add a `then(() => true)` after the isVisible command it works :-D

christian-bromann:
@georgecrawford I made some progress and decided to skip this feature for the standalone mode for now. I am struggling to catch the error in the right moment. When isVisible failed due to stale element error it bubbles up the chain [here](https://github.com/webdriverio/webdriverio/blob/onerror-hook/lib/webdriverio.js#L267-L285). So the error appears there when:

- elementIdVisible fails
- WebdriverIO.unify fails
- isVisible fails (last time)

If isVisible fails the onRejected method is `function () { [native code] }` which makes sense because the method comes from the fiber project which is a native node plugin, therefor native code. In standalone mode it's not the case and we get the reject method probably from the native Promise lib from nodejs.

Also I figured that it makes no sense to mix up internal error handler and the onError hook. As you can see the internal error helper has to check the error type to interfere if necessary and the failing command will return the result of that helper. If devs use the onError hook they probably want to take screenshots or whatnot and they would return a successful saveScreenshot command which wouldn't reject the command promise at the end. So I will try to execute the onError hooks somewhere before or after the internal helper runs.

Next thing is that if internal helper method fails twice it throws an error. I am not sure if that is what we want or if the helper should run multiple times which can lead to an endless loop.

Sometimes I still see that the error is reported twice which was already issued a couple of times. Would be good if this get fixed too.

I need a little bit more time to figure this out. It would be a powerful enhancement that is worth waiting for.

georgecrawford:
Thanks for the updates. I do think we should handle multiple stale element errors in a row, but obviously don't want to get into a tight loop. Can we just set an integer constant for now (5 times?), as I'm sure this will handle a very high percentage of these race conditions. We can then easily tweak it at a later stage. We should definitely log somewhere when this exception is encountered.

christian-bromann:
@georgecrawford I finally got it working with a caveat though: we can't enable error handling retries (like stale element retry) for standalone instances. The reasons is that we can't determine if the `onReject` method comes from the user or a framework (e.g. when you return the promise and Mocha checks if it fails and throws if so).

I also enabled services to use this onError hook. In order to test this I implemented a way to use custom services (basically an object of hooks, see https://github.com/webdriverio/webdriverio/commit/9cb6c90ff9473069eec66b85dfca2c65bb77beca)

Error hooks will automatically executed within the fiber context (see [here](https://github.com/webdriverio/webdriverio/blob/onerror-hook/lib/webdriverio.js#L140-L144)). But they respect promises either.

I am really happy to got this working finally. It was a real blocker. Let me know what you think and we can merge this!

christian-bromann:
Oh I forgot to tell how I got it working: my previous problem was that I didn't know when to apply the error handler:

> I am struggling to catch the error in the right moment.

I fixed that by checking the stack trace. It tells me if the command was executed within another command or not. Check out:

```js
client.depth = stack.split('\n').filter((line) => !!line.match(/\/lib\/(commands|protocol)\/(\w+)\.js/)).length
```

That allows me to trigger the error handler in the right moment: when isVisible gets rejected (and not if its subcommand elementIdVisible).

christian-bromann:
@georgecrawford I added integration tests for the wdio runner und ported the stale element retry tests over. This ensures that this feature works. I've seen that if an error handler throws an error it gets picked up by wdio-sync and causes a failure no matter if we resolve or not. I leave that as is for now to move forward and get the branch merged into master. Please review in retrospective. Thanks!

PR #979, opened 2016-01-05
fix the scroll error when using mobile browser

mqli:
When using scroll with out a selector.the callback in Q() object will get the wrong "this",therefor cause a TypeError "this.touchScroll is not a function".
bind this will fix it.


christian-bromann:
good catch!

christian-bromann:
thanks

PR #983, opened 2016-01-06
Return body instead of defer

sbmaxx:
Trying to fix failing tests after #952


christian-bromann:
Thanks for fixing this!

PR #992, opened 2016-01-07
fix waitUntil timeout issue

sbmaxx:
Fixed bug when `waitUntil` was stopped by timeout, but it tried to execute itself after that.


christian-bromann:
Can you add a test for that [here](https://github.com/webdriverio/webdriverio/blob/master/test/spec/functional/waitUntil.js)

sbmaxx:
@christian-bromann added tests and also refactored `waitUntilPrivate` a bit. Check if you like or not.

christian-bromann:
I really like the Timer prototype. Good job :+1:

christian-bromann:
@sbmaxx I can't port this feature into the v4 branch. Tests didn't pass and I think it is because of your changes. One big difference between your attempt and the current implementation is that the timer is waiting until the condition promise has resolved. The problem with that is if the condition for some reasons never resolves the test stops without any notifications.

Could you create a new PR on the v4 branch (or master if already merged) with this issue addressed? Thanks!

sbmaxx:
@christian-bromann ok

PR #1001, opened 2016-01-08
update broken link to saucelabs platform config

kencheeto:
docs.saucelabs.com redirects to wiki.saucelabs.com and the path changed as well


christian-bromann:
Thanks :+1:

PR #1006, opened 2016-01-10
Don't write colon after date in logs

sbmaxx:
We already have date inside square bracket. This should be enough.


christian-bromann:
agree

PR #1012, opened 2016-01-11
Only free up an available runner instance when a runner child process actually exists

georgecrawford:
I found that the launcher assumed that a runner process had exited when the `runner:error` event was received, even if the runner was still alive and sending messages. This resulted in the launcher process ending early, and Node errors as the runners attempted to send messages to a closed process. I believe this small change is enough to fix the issue.


christian-bromann:
Sweet! :+1:

Issue #1018, opened 2016-01-12
Use Selenium's automatic screenshot where possible


georgecrawford:
In some error cases (the example I saw was when throwing an `Error` in the browser context during an `executeAsync` call), Selenium will return a `screen` property on the HTTP response. If the user has opted in to screenshots, we should save this to disk, and record the file path on the error object so it can be referred to. This prevents both webdriverio and the user from having to request another screenshot to be taken when processing the error, which introduces a latency which may well cover up the error condition if it was a quick glitch.

christian-bromann:
Can you check if you get access to the screen property of the failed response somewhere along these lines: https://github.com/webdriverio/webdriverio/blob/master/lib/webdriverio.js#L118-L122

christian-bromann:
I think we can close this since #1021 fixes the issue. Feel free to reopen if not.

PR #1021, opened 2016-01-12
Save Selenium screenshot from command response where available

georgecrawford:
Addresses https://github.com/webdriverio/webdriverio/issues/1018

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -1,4 +1,7 @@
 import merge from 'deepmerge'
+import fs from 'fs'
+import path from 'path'
+import sanitize from './helpers/sanitize'
 
 import ConfigParser from './utils/ConfigParser'
 import { remote, multiremote } from '../'
@@ -176,6 +179,29 @@ class Runner {
         return payload
     }
 
+    saveScreenshot (screenshot) {
+        /**
+         * take screenshot only if screenshotPath is given
+         */
+        if (typeof global.browser.options.screenshotPath !== 'string') {
+            return
+        }
+
+        let screenshotPath = path.join(process.cwd(), global.browser.options.screenshotPath)
+
+        /**
+         * take screenshot only if directory exists
+         */
+        if (!fs.existsSync(screenshotPath)) {
+            throw Error(`screenshotPath "${screenshotPath}" is not writeable`)
+        }
+
+        let filename = path.join(screenshotPath, 'ERROR_' + sanitize.caps(global.browser.desiredCapabilities) + '_' + new Date().toJSON() + '.png')
+        fs.writeFileSync(filename, new Buffer(screenshot, 'base64'))
+
+        global.browser.emit('screenshot', {data: screenshot, filename})
+    }
+
     addCommandHooks (config) {
         config.beforeCommand.push((command, args) => {
             const payload = {
@@ -188,6 +214,10 @@ class Runner {
             process.send(this.addTestDetails(payload))
         })
         config.afterCommand.push((command, args, result, err) => {
+            if (err && err.seleniumStack && err.seleniumStack.screenshot) {
+                this.saveScreenshot(err.seleniumStack.screenshot)
+            }
+
             const payload = {
                 event: 'runner:aftercommand',
                 cid: this.cid,
--- a/lib/utils/ErrorHandler.js
+++ b/lib/utils/ErrorHandler.js
@@ -59,6 +59,7 @@ class ErrorHandler extends Error {
                 }
 
                 this.message = problem
+                this.seleniumStack = seleniumStack
             }
         }
     }
--- a/lib/utils/RequestHandler.js
+++ b/lib/utils/RequestHandler.js
@@ -149,7 +149,8 @@ class RequestHandler {
                         status: body.status,
                         type: ERROR_CODES[body.status] ? ERROR_CODES[body.status].id : 'unknown',
                         message: ERROR_CODES[body.status] ? ERROR_CODES[body.status].message : 'unknown',
-                        orgStatusMessage: body.value ? body.value.message : ''
+                        orgStatusMessage: body.value ? body.value.message : '',
+                        screenshot: body.value && body.value.screen || null
                     })
                 } else if (err.message.indexOf('Nock') > -1) {
                     // for better unit test error output


georgecrawford:
@christian-bromann This is not finished yet - it will only save a screenshot if the error comes through the `afterCommand` hook. Also, there's duplicated logic between these changes and the in-build webdriverio screenshot saving at https://github.com/webdriverio/webdriverio/blob/v4.0/lib/webdriverio.js#L98, so this needs to be abstracted out. But in principle, are you happy with the idea of the runner saving the screenshot and firing an event to say it's done it? The reason not to have it controlled in the same place is that you can't send the huge string buffer up to the launcher using `process.send`, as there's quite a small JSON length limit.

Interestingly, for me, Chromedriver on the Selenium Grid saves a screenshot of the whole OS window, not just the browser, which might potentially be very useful for debugging issues.

christian-bromann:
> The reason not to have it controlled in the same place

What do you mean by that? If possible it would be the best I think if we can have this logic within the webdriverio.js code because then it would also work without using wdio.

georgecrawford:
I'd love to have this in a single place, inside webdriverio, but I couldn't find a way, so I fear we may need to handle it in two separate places. If you call:

```
		browser.executeAsync(function(done) {
			throw Error('George error')
		});
```
...then the error is thrown [here](https://github.com/webdriverio/wdio-sync/blob/master/index.js#L150), and I don't know how to catch it. The only way I could find the `Error` object was in the `afterCommand` callback.

Can you work out how to handle this exception from within the webdrverio.js file? This kind of thing bugs me again and again as I don't really understand what the hell goes on in this file!

To be perfectly honest with you, I don't really have time to implement stuff in webdriverio without using `wdio`, as it's not so familiar with me. I need to concentrate on the parts I'm actually going to use.

christian-bromann:
So where to catch it highly depends whether we are running in wdio mode or if you run the standalone version and call then afterwards or/and have a catch block. I know pretty weird but that's how it is. Let's go through all possibilities:

__1:__ no then called
```js
var browser = require('./build').remote({
    screenshotPath: './',
    desiredCapabilities: {
        browserName: 'phantomjs'
    }
})

browser
    .init()
    .url('http://localhost:8080')
    .executeAsync(function () {
        throw new Error('catch me if you can!')
    })
    .end()
```

In this case you only can catch it [here](https://github.com/webdriverio/webdriverio/blob/v4.0/lib/webdriverio.js#L170-L184).

__2:__ then is called
```js
var browser = require('./build').remote({
    screenshotPath: './',
    desiredCapabilities: {
        browserName: 'phantomjs'
    }
})

browser
    .init()
    .url('http://localhost:8080')
    .executeAsync(function () {
        throw new Error('catch me if you can!')
    }).then(function () {
        // do stuff
    })
    .end()
```
In this case it gets propagated [here](https://github.com/webdriverio/webdriverio/blob/v4.0/lib/webdriverio.js#L94-L118) first and then at the position above.

__3:__ if you use wdio there will be always a then/catch method call after each command due to fiber (wdio-sync). If you have now `screenshotPath` given it already creates a screenshot.

georgecrawford:
Thanks for the first two examples. Could you add one for *3.* please, and a link to the line where I could catch the error inside webdriverio.js? Just so I'm 100% sure.

georgecrawford:
The reason I ask is that with `wdio`, neither [1](https://github.com/webdriverio/webdriverio/blob/v4.0/lib/webdriverio.js#L170-L184) nor [2](https://github.com/webdriverio/webdriverio/blob/v4.0/lib/webdriverio.js#L94-L118) are hit with the exception.

georgecrawford:
I'm now wondering whether a `throw` [here](https://github.com/webdriverio/wdio-sync/blob/master/index.js#L150) is correct?

georgecrawford:
@christian-bromann Another oddity I've just discovered. I happened to have an `afterTest` hook which takes a screenshot on failure. In addition, for the first time, I have set `screenshotPath` to a directory which actually exists. I found that with `wdio`, the automatic screenshot performed by webdriverio.js is not waited for, resulting in this:
```
Demo test
taking screenshot /Users/george/Shell/sandboxes/ft-app-webdriverio/output/all/2016-01-14-16-29-58/screenshots/ERROR_chrome.latest.mac_2016-01-14T16:30:03.137Z.png
  1) should throw an error
taking screenshot /Users/george/Shell/sandboxes/ft-app-webdriverio/output/all/2016-01-14-16-29-58/screenshots/Test failure: An element could not be located on the page using the given search parameters.
screenshot complete /Users/george/Shell/sandboxes/ft-app-webdriverio/output/all/2016-01-14-16-29-58/screenshots/Test failure: An element could not be located on the page using the given search parameters.
  ✓ should pass a test (2ms)
screenshot complete /Users/george/Shell/sandboxes/ft-app-webdriverio/output/all/2016-01-14-16-29-58/screenshots/ERROR_chrome.latest.mac_2016-01-14T16:30:03.137Z.png
```
i.e.:
```
- spec starts
- bad test starts
- error encountered, automatic screenshot starts
- error thrown
- `afterHook` starts and completes screenshot
- bad test completes
- good test starts
- automatic screenshot finally completes, but will not be associated with the correct test
```

christian-bromann:
So you take a screenshot (run a command) in the afterTest hook and it does not wait?

georgecrawford:
Other way round - the afterHook screenshot *does* wait, but the in-built webdriverio one does not. 

________________



George

christian-bromann:
Interesting! I was thinking of dropping that feature because it also applies to v3.x (see #1030). Though it should be possible to delay the execution of the command until the screenshot is taken. It's probably because I don't use the promise that is returned by https://github.com/webdriverio/webdriverio/blob/master/lib/webdriverio.js#L120-L122.

georgecrawford:
OK, happy to drop that feature then. We could suggest that people use an `afterTest` hook instead - or even add one by default. However, my other points in this thread are now even more valid, as I'd like to properly handle screenshots on errors in a single location, and have lots of open questions. Could you add any answers you're able to?

christian-bromann:
@georgecrawford I think we can solve this and the onError hook with #969. This is currently the primary task I am working on. If I return the promise and then throw the error I should be able to reject the actual command. If I can get this working we should go with that solution instead of injecting a new hook.

christian-bromann:
@georgecrawford I ported your path to the webdriverio.js file where it gets handled in standalone mode too

PR #1023, opened 2016-01-13
Don't call logger directly

sbmaxx:
I don't know why `set session id` is so special now. Maybe it shouldn't be?

For example, it log session id on `logLevel = error`


christian-bromann:
@sbmaxx we updated the code base (merged v4 branch). Could you please fetch a fresh copy from the master branch and open a new PR? Thanks!

PR #1032, opened 2016-01-15
Update selectors.md to have text that makes sense

jonathan-fielding:
Hey 

I updated the docs for selectors so this line made a bit more sense

Regards

Jonathan


christian-bromann:
Awesome thanks!

PR #1039, opened 2016-01-19
Various tweaks to onerror & staleElementreference

georgecrawford:


test/fixtures/specs/stale.spec.js
```diff
@@ -23,4 +23,24 @@ describe('staleElementRetry', () => {
         browser.url(conf.testPage.staleTest)
         browser.staleMe(100)
     })
+
+    it('does not throw staleElementReference exception when waiting for element to become invisible but which is removed from DOM in a custom command', () => {
+
+        browser.addCommand('waitForInvisibleInParallel', (iterations = 20) => {
+            const promises = []
+            for (let i = 1; i <= iterations; i++) {
+                promises.push(browser.waitForVisible('.staleElementContainer2 .stale-element-container-row-' + i, 10000, true))
```
christian-bromann: `browser.waitForVisible()` does not return a promise here. You have to add async as function name to run in a promise context

georgecrawford:
Still fails with `Promise was fulfilled but got rejected with the following reason: Error: Element is no longer attached to the DOM`

christian-bromann:
I am wondering if this is a legit use case people are facing. I think we should use nock here to trigger a stale element retry and see what happens. I see that this is hard to reproduce otherwise.

christian-bromann:
I will merge it as a start to test waitForXXX commands for stale element errors

georgecrawford:
It's a legit use-case - it's my *primary* use-case! I run into this every day, with a 'page loading' overlay which often throws a staleElementReference exception in my `waitForVisible(..., ..., false)` when removed from the DOM.

Any ideas why this test fails?

georgecrawford:
It's to do with `this.unify` in `isVisible`, right? I think we somehow need to run the `internalErrorHandler` here.

PR #1042, opened 2016-01-21
Add failing test for complex waitForVisible()

georgecrawford:
This test, I think covers the issue I'm seeing in my own test suite.

Debug output:
```
isVisible call count: 1
elementIdDisplayed call count: 1
elementIdDisplayed count: 1, res [object Object]
isVisible count: 1, res true
isVisible call count: 2
elementIdDisplayed call count: 2
elementIdDisplayed error count: 2, e: Element is no longer attached to the DOM
isVisible error count: 2, e: Element is no longer attached to the DOM
caught StaleElementReference
StaleElementReference. Will retry isVisible with .someSelector
isVisible call count: 3
elementIdDisplayed call count: 3
elementIdDisplayed count: 3, res [object Object]
isVisible count: 3, res true
isVisible call count: 4
.someSelector still visible after 1 second. Will wait a bit longer.
isVisible call count: 5
elementIdDisplayed call count: 4
elementIdDisplayed call count: 5
elementIdDisplayed count: 4, res [object Object]
elementIdDisplayed error count: 5, e: Element is no longer attached to the DOM
isVisible count: 4, res true
isVisible error count: 5, e: Element is no longer attached to the DOM
caught StaleElementReference
```
... at this point, there's no output for a few seconds - no commands are executed. Then:
```․

0 passing (10.50s)
1 failing

1) correctly retries inside waitForVisible:
element (.someSelector) still  visible after 5000ms
```

So the retry after the second `StaleElementReference` catch is not working as expected.

I also hit some issues with Nock which I don't understand. I get the output above only if I run the single test in isolation (`correctly retries inside waitForVisible`). If I allow `catches errors if an inner command fails` to run as well, then `correctly retries inside waitForVisible` doesn't seem to call commands as expected at all. Not sure how to fix this.

### DIFF
--- a/test/fixtures/specs/stale.spec.js
+++ b/test/fixtures/specs/stale.spec.js
@@ -1,6 +1,33 @@
 import conf from '../../conf/index.js'
 import nock from 'nock'
 
+let sessionId
+let scope
+
+function goodElementRequest (times = 1) {
+    scope.post(`/wd/hub/session/${sessionId}/elements`).times(times).delayConnection(100).reply(200, {
+        status: 0,
+        value: [{ ELEMENT: '0' }]
+    })
+}
+
+function staleElementError (times = 1) {
+    scope.get(`/wd/hub/session/${sessionId}/element/0/displayed`).times(times).delayConnection(100).reply(500, {
+        status: 10,
+        type: 'StaleElementReference',
+        value: {
+            message: 'Element is no longer attached to the DOM'
+        }
+    })
+}
+
+function isDisplayed (times = 1, value) {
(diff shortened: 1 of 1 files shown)


christian-bromann:
@georgecrawford I merged this into master and fixed the issue.

- One problem was that the waitUntil condition gets executed even though the timeout is expired, fix: https://github.com/webdriverio/webdriverio/commit/6882800a9c8a8675be3c8ae130b77b4ecd5a4d7d
- also the test was not quite correct because the elements endpoint isn't it as often as originally expected, fix: https://github.com/webdriverio/webdriverio/commit/054f7ef59a1073c99f4601289bac936b947d4bb2

PR #1044, opened 2016-01-21
Update moveToObject.js

DV3D:
Guess I stumbled upon a bug. This is my first github bugfix attempt :)

So I'm not entirely sure this works, because I don't use mobile testing yet, but this seems more like what the description is saying:
Default: move to center of the element
When offset is set: move to left top of the element + offset

Can someone verify + test this?

Thanks!
Daan


christian-bromann:
Your fix makes sense to me. The build is kinda broken right now but we will merge the new v4 branch soon into master. Thanks! And congrats to your first Github bugfix attempt!

PR #1048, opened 2016-01-23
Port of #992

sbmaxx:
In #992 
> @sbmaxx I can't port this feature into the v4 branch. Tests didn't pass and I think it is because of your changes. One big difference between your attempt and the current implementation is that the timer is waiting until the condition promise has resolved. The problem with that is if the condition for some reasons never resolves the test stops without any notifications.

> Could you create a new PR on the v4 branch (or master if already merged) with this issue addressed? Thanks!

I run many tests on master/timer branch, and sometime `phantomjs` hangs. Chrome works fine. I think that `mocha`'s big timeout can be the issue. 

Also, i'm not sure about `await` in tests. Why just not return promise always?

### DIFF
--- a/lib/commands/waitForEnabled.js
+++ b/lib/commands/waitForEnabled.js
@@ -14,7 +14,7 @@
  *
  */
 
-import { CommandError } from '../utils/ErrorHandler'
+import { CommandError, isTimeoutError } from '../utils/ErrorHandler'
 
 let waitForEnabled = function (selector, ms, reverse = false) {
     /*!
@@ -42,7 +42,7 @@ let waitForEnabled = function (selector, ms, reverse = false) {
             return result !== reverse
         })
     }, ms).catch((e) => {
-        if (e.message === 'Promise never resolved with an truthy value') {
+        if (isTimeoutError(e)) {
             let isReversed = reverse ? '' : 'not'
             throw new CommandError(`element (${selector}) still ${isReversed} enabled after ${ms}ms`)
         }
--- a/lib/commands/waitForExist.js
+++ b/lib/commands/waitForExist.js
@@ -14,7 +14,7 @@
  *
  */
 
-import { CommandError } from '../utils/ErrorHandler'
+import { CommandError, isTimeoutError } from '../utils/ErrorHandler'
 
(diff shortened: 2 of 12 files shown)


sbmaxx:
@christian-bromann take a look please

christian-bromann:
@sbmaxx could rebase against current master?

christian-bromann:
can you check? One of your unit test is failing:

```
  58 passing (5s)
  1 failing
  1) timer ticks should not tick after timeout:
      AssertionError: expected 5 to equal 6
      + expected - actual
      -5
      +6
      
      at Context.<anonymous> (test/spec/unit/timer.js:37:41)
```

sbmaxx:
@christian-bromann should be OK now

christian-bromann:
Awesome thanks!

christian-bromann:
@sbmaxx I running into problems using the new implementation of waitUntil .. the timeout gets reach way to early.

sbmaxx:
@christian-bromann can you show an example?

christian-bromann:
@sbmaxx never mind I was wrong, I faced a different timeout :wink:

kevinsawicki:
Did this change the value of `this` inside a `waitUntil` callback?

I'm seeing a different `this` value when upgrading from `3.2.6` to `4.0.3`. Just wondering if that was intentional or not.

christian-bromann:
@kevinsawicki `this` should refer to `browser`, so you can do `this.click(...)...`. What is the value of this for you?

kevinsawicki:
@christian-bromann  inside a `waitUntil` callback on 4.0.3 I'm not seeing a `this` value of something like:

```
{ _delay: 250,
  _timeout: 5000,
  _fn: [Function],
  _leading: true,
  _start: 1456941856301,
  _ticks: 0 }
```

christian-bromann:
Gotcha. This is the wrong scope then. I will fix this.

kevinsawicki:
> Gotcha. This is the wrong scope then. I will fix this. Thanks for reporting!

@christian-bromann would it be helpful if I opened a pull request for fixing the `this` issue?

PR #1060, opened 2016-01-27
Add typescript support

MrHayato:
Added another filter parameter to the ConfigParser to allow .ts files
Added section to configuration docs on setting up TypeScript support
Modified ConfigParser test to check for .ts files

### DIFF
--- a/test/spec/unit/configparser.js
+++ b/test/spec/unit/configparser.js
@@ -9,7 +9,8 @@ describe('ConfigParser', () => {
         configParser.addConfigFile(FIXTURES_PATH + '/exclude.conf.js')
         let specs = configParser.getSpecs()
         specs.should.not.include(__filename)
-        specs.should.include(__dirname + '/call.js')
+        specs.should.include(path.resolve(__dirname + '/call.js'))
+        specs.should.include(path.resolve(__dirname + '/typescript.ts'))
     })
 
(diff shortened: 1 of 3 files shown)


test/spec/unit/configparser.js
```diff
@@ -9,7 +9,8 @@ describe('ConfigParser', () => {
         configParser.addConfigFile(FIXTURES_PATH + '/exclude.conf.js')
         let specs = configParser.getSpecs()
         specs.should.not.include(__filename)
-        specs.should.include(__dirname + '/call.js')
+        specs.should.include(path.resolve(__dirname + '/call.js'))
+        specs.should.include(path.resolve(__dirname + '/typescript.ts'))
```
christian-bromann: Could we move this file into the fixtures directory and modify the exclude.conf.js to also point to lets say all ts files in that fixture directory?

MrHayato: Would you be opposed to an exclude3.conf.js that handles this? I can create a new test that uses exclude3 and revert this one back to the original.

christian-bromann: nope, not at all

MrHayato: Actually see my next commit and see if that works for you

christian-bromann:
Sweet thanks!

PR #1075, opened 2016-02-07
docs(test.js): update example to use getTitle.then

StrangeWind:
I'm brand new to webdriver.io, and it took a few minutes to figure out that the use of `.title()` in the example code provided was causing my test.js script to silently fail. My research showed that `.getTitle().then()` is the correct approach.


StrangeWind:
My apologies, I neglected to add () after .getTitle - closing request un-pulled.

PR #1076, opened 2016-02-07
docs(test.js): update to use .getTitle().then()

StrangeWind:
This particular example was failing silently on node 4.2; I updated to bring it in line with the homepage example with use of `.getTitle().then()` instead of `.title()`


christian-bromann:
thanks :+1:

PR #1078, opened 2016-02-08
Fix contributor's name

sbmaxx:


christian-bromann:
This is generated automatically so it could be that it gets added again when doing a new release.

PR #1080, opened 2016-02-08
dont call logger directly

sbmaxx:
port of #1023


christian-bromann:
An info method makes sense. Thanks!

PR #1082, opened 2016-02-09
Bringing back selectByName

Risto-Stevcev:
I noticed that `selectByName` got lost somehow. I updated the original code into ES6. I also did this:

```js
        let normalized = `[normalize-space(@name) = "${name.trim()}"]`
```

Template strings accept expressions within the `${}`, which differs from how the other `selectBy*` commands do it. I hope that's ok.

Props for switching to ES6 :sunglasses:

### DIFF
--- a/lib/commands/selectByName.js
+++ b/lib/commands/selectByName.js
@@ -1,0 +1,71 @@
+/**
+ *
+ * Select option with a specific name.
+ *
+ * <example>
+    :example.html
+    <select id="selectbox">
+        <option name="someName0" value="someValue0">uno</option>
+        <option name="someName1" value="someValue1">dos</option>
+        <option name="someName2" value="someValue2">tres</option>
+        <option name="someName3" value="someValue3">cuatro</option>
+        <option name="someName4" value="someValue4">cinco</option>
+        <option name="someName5" value="someValue5">seis</option>
+    </select>
+
+    :selectByName.js
+    client
+        .getValue('#selectbox').then(function(value) {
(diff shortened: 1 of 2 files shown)


christian-bromann:
@Risto-Stevcev thanks for the PR

`selectByValue` and `selectByName` seem to be pretty similar and would introduce code duplication. Could we instead of adding a new `selectBy*` just generalise this? I would say it makes more sense renaming `selectByValue` to `selectByAttribute` and let the user decide which attribute he wants to use. What do you think?

Risto-Stevcev:
@christian-bromann I agree, it's essentially the same code: `selectByAttribute`. Also made your build fail bc I forgot to switch `function(res)` to `(res) =>` (my bad).

Risto-Stevcev:
It would also make the API more flexible, since there are really popular libraries that add their own attributes (ie angular), and I'm sure a lot of people would love to use those selectors.

alexgorbatchev:
I'm currently getting this error, it doesn't seem like `selectByName` is in the code, even though it's in the API docs. Was it removed?

```
/.../index.js:17
  .selectByName('#foo', 'bar')
   ^

TypeError: client.init(...).url(...).getText(...).then(...).selectByName is not a function
```

Risto-Stevcev:
@alexgorbatchev Yeah, by accident. @christian-bromann did a hard push when he merged v4 to master because the branches deviated too much to be practical enough to merge. Some commits got lost, including the one where I added selectByName.

PR #1083, opened 2016-02-09
Grid protocol

georgecrawford:
Seems to work nicely:

```js
		const nodeDetails = browser.getGridNodeDetails();
		console.log('nodeDetails', nodeDetails);
```
gives:
```
{ request:
   { class: 'org.openqa.grid.common.RegistrationRequest',
     id: null,
     name: null,
     description: null,
     configuration:
      { port: 5557,
        nodeConfig: '/Applications/gridconfiguration/configs/safariLatestNode.json',
        servlets: [],
        host: '192.168.2.1',
        cleanUpCycle: 2000,
        browserTimeout: 120000,
        hubHost: 'example.com',
        registerCycle: 10000,
        'Dwebdriver.chrome.driver=/Applications/gridconfiguration/drivers/mac/chromedriver_2_21_mac32': '',
        capabilityMatcher: 'org.openqa.grid.internal.utils.DefaultCapabilityMatcher',
        newSessionWaitTimeout: -1,
        url: 'http://example.com:5557',
        remoteHost: 'http://example.com:5557',
        register: true,
        id: 'MacMiniA10',
        throwOnCapabilityNotPresent: true,
        nodePolling: 2000,
        proxy: 'org.openqa.grid.selenium.proxy.DefaultRemoteProxy',
        maxSession: 4,
        role: 'node',
        jettyMaxThreads: -1,
        nodeTimeout: 120,
        hubPort: 80,
        timeout: 90000 },
     capabilities:
      [ { seleniumProtocol: 'WebDriver',
          platform: 'MAC',
          firefoxVersion: '42',
          browserName: 'firefox',
          maxInstances: 2,
          version: 'latest' },
        { seleniumProtocol: 'WebDriver',
          platform: 'MAC',
          browserName: 'chrome',
          maxInstances: 8,
          chromeVersion: '46',
          version: 'latest' },
        { seleniumProtocol: 'WebDriver',
          platform: 'MAC',
          browserName: 'safari',
          maxInstances: 1,
          version: 'latest' } ] },
  session: '33c2d04d-6bc1-424e-ba6f-63c400114554',
  internalKey: 'd710b167-6f29-4097-a15c-7f7bbdb73edb',
  inactivityTime: 78,
  proxyId: 'MacMiniA10' }
```

Easy to log this info, using Allure reporter for example:
screenshot 2016-02-09 13 56 47

### DIFF
--- a/lib/commands/getGridNodeDetails.js
+++ b/lib/commands/getGridNodeDetails.js
@@ -1,0 +1,44 @@
+/**
+ *
+ * Get the details of the Selenium Grid node running a session (only when running on a Grid)
+ *
+ * <example>
+    :grid.js
+    browser.getGridNodeDetails().then(function(details) {
+        console.log(details);
+
+        // {
+        //     success: true,
+        //     msg: "proxy found !",
+        //     id: "MacMiniA10",
+        //     request: {
+        //         ...
+        //         configuration: {
+        //             ...
+        //         },
+        //         capabilities: [
+        //             {
+        //                 ...
+        //             }
+        //         ]
+        //     }
+        // }
+    })
+ * </example>
+ *
+ * @uses protocol/gridTestSession, protocol/gridProxyDetails
+ * @type grid
+ */
+import { CommandError } from '../utils/ErrorHandler'
+
+let getGridNodeDetails = function (selector, longClick = false) {
+    if (!this.isGrid) {
+        throw new CommandError('getGridNodeDetails command is only supported when running on Selenium Grid')
+    }
+
+    return this.gridTestSession().then((session) =>
+        this.gridProxyDetails(session.proxyId)
+    )
+}
+
+export default getGridNodeDetails
--- a/lib/commands/getGridNodeDetails.js
+++ b/lib/commands/getGridNodeDetails.js
@@ -31,7 +31,7 @@
  */
 import { CommandError } from '../utils/ErrorHandler'
 
-let getGridNodeDetails = function (selector, longClick = false) {
+let getGridNodeDetails = function () {
     if (!this.isGrid) {
         throw new CommandError('getGridNodeDetails command is only supported when running on Selenium Grid')
     }
--- a/lib/commands/getGridNodeDetails.js
+++ b/lib/commands/getGridNodeDetails.js
@@ -45,11 +45,7 @@ let getGridNodeDetails = function () {
             delete details.success
             delete details.id
 
-            for (let key of Object.keys(session)) {
-                details[key] = session[key]
-            }
-
-            return details
+            return Object.assign(details, session)
         })
     )
 }
--- a/lib/commands/getGridNodeDetails.js
+++ b/lib/commands/getGridNodeDetails.js
@@ -37,7 +37,20 @@ let getGridNodeDetails = function (selector, longClick = false) {
     }
 
     return this.gridTestSession().then((session) =>
-        this.gridProxyDetails(session.proxyId)
+        this.gridProxyDetails(session.proxyId).then((details) => {
+            delete session.msg
+            delete session.success
+
+            delete details.msg
+            delete details.success
+            delete details.id
+
+            for (let key of Object.keys(session)) {
+                details[key] = session[key]
+            }
+
+            return details
+        })
     )
 }
 
--- a/lib/commands/getGridNodeDetails.js
+++ b/lib/commands/getGridNodeDetails.js
@@ -1,6 +1,6 @@
 /**
  *
- * Get the details of the Selenium Grid node running a session (only when running on a Grid)
+ * Get the details of the Selenium Grid node running a session
  *
  * <example>
     :grid.js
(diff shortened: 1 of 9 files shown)


lib/commands/getGridNodeDetails.js
```diff
@@ -0,0 +1,44 @@
...
+        // }
+    })
+ * </example>
+ *
+ * @uses protocol/gridTestSession, protocol/gridProxyDetails
+ * @type grid
+ */
+import { CommandError } from '../utils/ErrorHandler'
+
+let getGridNodeDetails = function (selector, longClick = false) {
```
christian-bromann: I guess the params are obsolete

lib/protocol/gridProxyDetails.js
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    return this.requestHandler.create({
+        path: `/api/proxy?id=${proxyId}`,
+        method: 'GET',
+        requiresSession: false,
+        gridCommand: true
+    })
+    .catch(e => {
+        console.log(e.stack)
```
christian-bromann: Why do we print this out instead of propagating it back to the user?

lib/protocol/gridTestSession.js
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    return this.requestHandler.create({
+        path: `/api/testsession?session=${sessionId}`,
+        method: 'GET',
+        requiresSession: false,
+        gridCommand: true
+    })
+    .catch(e => {
+        console.log(e.stack)
```
christian-bromann: Same question applies here

lib/commands/getGridNodeDetails.js
georgecrawford: Thanks, good spot.

lib/protocol/gridProxyDetails.js
georgecrawford: Debugging. Removed.

lib/protocol/gridTestSession.js
georgecrawford: Fixed.

christian-bromann:
@georgecrawford can you push the fixes :wink:

georgecrawford:
Heh, sorry - I'm half asleep today.

lib/commands/getGridNodeDetails.js
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    return this.gridTestSession().then((session) =>
+        this.gridProxyDetails(session.proxyId).then((details) => {
+            delete session.msg
+            delete session.success
+
+            delete details.msg
+            delete details.success
+            delete details.id
```
christian-bromann: Do we need all these delete statements?

lib/commands/getGridNodeDetails.js
```diff
@@ -0,0 +1,57 @@
...
+
+            for (let key of Object.keys(session)) {
+                details[key] = session[key]
+            }
```
christian-bromann: Let's use `Object.assign` here to be consistent with the rest of the code.

lib/commands/getGridNodeDetails.js
```diff
@@ -0,0 +1,57 @@
...
+ */
+import { CommandError } from '../utils/ErrorHandler'
+
+let getGridNodeDetails = function () {
+    if (!this.isGrid) {
+        throw new CommandError('getGridNodeDetails command is only supported when running on Selenium Grid')
+    }
+
+    return this.gridTestSession().then((session) =>
+        this.gridProxyDetails(session.proxyId).then((details) => {
```
christian-bromann: I think we miss a return here too

christian-bromann:
@georgecrawford ok I found two last things I would change, otherwise LGTM

lib/commands/getGridNodeDetails.js
georgecrawford: There's an implicit `return` because of fat arrow syntax

lib/commands/getGridNodeDetails.js
georgecrawford: Done.

lib/commands/getGridNodeDetails.js
christian-bromann: oh right, I missed that

georgecrawford:
@christian-bromann Could you advise on which files I should update to ensure the `gridPath` config var is well documented?

christian-bromann:
@georgecrawford good point, as an option for the standalone mode it should be definitely added here: https://github.com/webdriverio/webdriverio/blob/master/docs/guide/getstarted/configuration.md#configuration. Since it is a specific use case, would you mind to create a new page [here](https://github.com/webdriverio/webdriverio/tree/master/docs/guide/usage) that explains real quick how to use this add one in order to grab information from the grid?

georgecrawford:
Sure, will do. Bear in mind that I'm currently adding only one param, `gridPath`, and we may need to add more such as `gridHost` or `gridPort` - I don't really know what the possibilities are for setup. But we can add those when people raise issues.

christian-bromann:
@georgecrawford this is still a bit confusing: why do we have to set `gridPath`? I think is options is obsolete! I would rather just do the request no matter what and if the server response with 500 we return a custom error message saying that you don't run a Selenium grid. What do you think?

georgecrawford:
Oh, that would probably work. It's just that I don't know whether the grid API URL is something that could change. I'm happy to keep it simpler if you like.

christian-bromann:
Yeah, let's say for now that the grid url doesn't change. Let's implement in a way that it's easy to also add the other api commands.

georgecrawford:
@christian-bromann OK?

christian-bromann:
Looks good 👌🏻

PR #1092, opened 2016-02-11
setImmediateValue always throws an exception

agauthiez:
The assertion to throw the `ProtocolError` was wrong.

### DIFF
--- a/lib/protocol/setImmediateValue.js
+++ b/lib/protocol/setImmediateValue.js
@@ -15,7 +15,7 @@
 import { ProtocolError } from '../utils/ErrorHandler'
 
 let setImmediateValue = function (id, value) {
-    if (typeof id !== 'string' || typeof id !== 'number') {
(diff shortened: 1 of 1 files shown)


agauthiez:
By the way I just noticed that the top documentation block includes `@for android` but I get a 501 ("Not yet implemented" `RuntimeError`) from Appium 1.4.16 when using the Android driver. It works with iOS though. Shouldn't it be changed?

christian-bromann:
Thanks, good catch!

agauthiez:
By the way I just noticed that the top documentation block includes `@for android`. Using Appium 1.4.16, it returns a 501 (a "Not yet implemented" RuntimeError) with the Android driver while it works on iOS. Shouldn't it be changed?

christian-bromann:
Oh interesting, yeah could you do another PR then? I probably switched that up then.

PR #1094, opened 2016-02-11
Ensure boolean flags are actually boolean

georgecrawford:
I was seeing results like this, which I assume is not desired:
```js
isMobile: true
isIOS: null
isAndroid: [ 'Android', index: 0, input: 'Android' ]
```


christian-bromann:
Good point :+1:

PR #1095, opened 2016-02-12
Wrap stats logging in try...catch but print errors, so runner doesn't crash

georgecrawford:
This doesn't address cases where a `Unrecognised suite [] for runner` error is encountered, which still need fixing, but at least won't let the runner crash in such cases.


christian-bromann:
I don't think this is a good idea. Instead of suppressing these errors we should try to fix them in the first place. The reporter really don't have much logic in it and if they are broken than something serious is broken.

georgecrawford:
OK, agreed. We just need to find some failing test cases.

PR #1099, opened 2016-02-12
Change setImmediateValue @for to iOS

agauthiez:
As pointed out on #1092 thread, `setImmediateValue` top documentation block includes @for android. Using Appium 1.4.16, it returns a 501 (a "Not yet implemented" RuntimeError) with the Android driver but it works on iOS.


christian-bromann:
Thanks!

Issue #1101, opened 2016-02-12
update isVisible documentation on array return value


bevanhunt:
https://github.com/webdriverio/webdriverio/blob/3d7201416c97850c24d5ea12fea26284fe7c4960/test/spec/isVisible.js

http://webdriver.io/api/state/isVisible.html

IsVisible if it matches multiple elements will return an array. This is missing from the docs. Please add. Cheers.

christian-bromann:
Got added. This applies to all commands btw.

PR #1103, opened 2016-02-14
Update isVisible.js

patthiel:
Added note per #1101


christian-bromann:
Thanks! This applies to all commands btw

patthiel:
I was thinking that might be the case, but when i tried waitForExist on a selector that exists multiple times throughout the page it only returned true/false for the first found selector. I guess that'd make sense though, within the context of that command.

PR #1107, opened 2016-02-15
update broken google code links to github wiki

psimyn:
Can point to their [new docs](https://seleniumhq.github.io/docs/) when these pages get added, but for now the links all redirected to project root on github


christian-bromann:
Thanks!

PR #991, opened 2016-01-07
Improved BaseReporter (replaces #896)

georgecrawford:
Reworking of https://github.com/webdriverio/webdriverio/pull/896, including:

- Add command hooks to enable reporting of `beforeCommand`, `afterCommand` and `screenshot` (cd8887f)
- Rework base reporter to be more useful, and allow for concurrent specs (0be45bf)
- Report screenshot file paths (0be45bf)
- Report `app` when sanitizing capabilities (2a85761)
- Enable `runner:command` and `runner:result` to be associated with a test (7057bdf)
- Expose upstream errors when failing to load wdio framework or plugin (2c02405)

PR #1108, opened 2016-02-16
Remove app path from sanitize.caps() so it fixes #1104

### LINKED ISSUE #1104: Logger does not sanitize app path
When providing the [`app` desired capability](https://wiki.saucelabs.com/display/DOCS/Test+Configuration+Options#TestConfigurationOptions-ApplicationPath), the logger fails because it tries to open a log file containing the app path:
```
Error: ENOENT: no such file or directory, open '/.../log/google.android./my/super/long/path/to/app.apk.2016-02-15T10:30:06.742.log'
```
I think we have to options here: do not include the app name in file name or escape the path properly.

agauthiez:
Quickfix for #1104 as the app name is polluting the log file name. Unless you run parallel tests on different applications (Is that common?), it will not lead to file name conflicts.

### DIFF
--- a/lib/helpers/sanitize.js
+++ b/lib/helpers/sanitize.js
@@ -29,7 +29,7 @@ let caps = function (caps) {
      * mobile caps
      */
     if (caps.deviceName) {
-        result = [sanitizeString(caps.deviceName), sanitizeString(caps.platformName), sanitizeString(caps.platformVersion)]
+        result = [sanitizeString(caps.deviceName), sanitizeString(caps.platformName), sanitizeString(caps.platformVersion), sanitizeString(caps.app)]
     } else {
         result = [sanitizeString(caps.browserName), sanitizeString(caps.version), sanitizeString(caps.platform), sanitizeString(caps.app)]
     }
--- a/lib/helpers/sanitize.js
+++ b/lib/helpers/sanitize.js
@@ -12,6 +12,7 @@ let sanitizeString = function (str) {
     }
 
     return String(str)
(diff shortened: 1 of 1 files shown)


christian-bromann:
@georgecrawford what do you think? It's kinda related to #991

agauthiez:
If we want to keep the app name, we can either only keep the file name with `caps.app.replace(/^.*[\\\/]/, '')` or replace all slashes with something. I would prefer the first solution, the second one can lead to crazy long log file names.

georgecrawford:
I think we should keep the app name, but we should certainly escape it. I'd do `caps.app.replace(/^.*\/([^\/]+)\/?$/, '$1')` to get the last segment of the path (or the whole value if it doesn't look like a path). Does that work?

agauthiez:
Yes, it works. I did the change.

lib/helpers/sanitize.js
```diff
@@ -28,9 +29,9 @@ let caps = function (caps) {
      * mobile caps
      */
     if (caps.deviceName) {
-        result = [sanitizeString(caps.deviceName), sanitizeString(caps.platformName), sanitizeString(caps.platformVersion), sanitizeString(caps.app)]
+        result = [sanitizeString(caps.deviceName), sanitizeString(caps.platformName), sanitizeString(caps.platformVersion)]
```
georgecrawford: Can you now reinstate `caps.app` here?

agauthiez:
My bad. Done.

christian-bromann:
LGTM! Thanks @agauthiez

PR #1117, opened 2016-02-18
Timer: clear main timeout on end

ilgooz:
the function registered to nodejs's event queue with setTimeout
is never terminated. therefore nodejs process never ends
even after the promise is resolved and it'll wait untill the callback
function get executed. this a fix to this issue.

### DIFF
--- a/lib/utils/Timer.js
+++ b/lib/utils/Timer.js
@@ -32,7 +32,7 @@ class Timer {
             this._timeoutId = setTimeout(this.tick.bind(this), this._delay)
         }
 
-        setTimeout(() => {
+        this._mainTimeoutId = setTimeout(() => {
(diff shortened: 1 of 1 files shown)


christian-bromann:
Good catch! :+1: 
__cc__ @sbmaxx

sbmaxx:
Really good catch, @ilgooz 

But i think naming should be `this._rejectTimeout`.

P.S. > it can be also done with `timer.unref`:
```js
setTimeout(() => {
    this._reject('timeout')
    this.stop()
}, this._timeout).unref()
```

ilgooz:
@christian-bromann @sbmaxx thank you! I think <code>this._rejectTimeout</code> is a good name for me too.

PR #1141, opened 2016-03-02
Adjust validation checks for location object

michaeljones:
Hi,

It seems to me that the validation checks here needed to be tweaked as they would fail given a value of 0 which is a valid value.

I am not familiar with the project and so have only grepped for tests for this `setGeoLocation` and haven't found any to update. I can try to dig deeper and add one if desirable.

Kind regards,
Michael

### DIFF
--- a/lib/commands/setGeoLocation.js
+++ b/lib/commands/setGeoLocation.js
@@ -15,7 +15,10 @@ let setGeoLocation = function (location) {
     /*!
      * parameter check
      */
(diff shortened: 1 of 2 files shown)


christian-bromann:
> I am not familiar with the project and so have only grepped for tests for this setGeoLocation and haven't found any to update. I can try to dig deeper and add one if desirable.

that's fine we haven't much coverage for protocol commands .. maybe soon

christian-bromann:
LGTM!

michaeljones:
Thanks!

PR #1143, opened 2016-03-03
Add more special selector characters to CSS detection.

brendandahl:
'[' and ']' weren't escaped properly and there were several CSS special characters missing.

Ref:
https://www.w3.org/TR/css3-selectors/

Aside 1: I can't get all unit tests to run locally. It seems to only run one of the spec files (sanitize.js). I had to manually run just the selectors.js file.

Aside 2: I personally would prefer if matching always defaulted to css selectors and if I wanted to use a special syntax I'd explicitly call another function or pass in an argument saying that.

### DIFF
--- a/lib/helpers/findElementStrategy.js
+++ b/lib/helpers/findElementStrategy.js
@@ -25,7 +25,7 @@ let findStrategy = function (...args) {
 
     // check value type
     // use id strategy if value starts with # and doesnt contain any other CSS selector-relevant character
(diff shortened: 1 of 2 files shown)


brendandahl:
As for 1, looking in the logs you seem to have the same problem on travis.
https://travis-ci.org/webdriverio/webdriverio/jobs/113280729#L692

christian-bromann:
> As for 1, looking in the logs you seem to have the same problem on travis.

I fixed this issue.

Indeed the id selector was selected even though it should be css selector. Good catch!

PR #1148, opened 2016-03-04
Update webdriverio.externalAddCommandHelper.js

beauwalls1:
The arguments and call() aren't needed.


christian-bromann:
Nice thanks!

PR #1161, opened 2016-03-10
Fix a bug in windowHandleSize without explicit window handle

ipoddubny:
The constructed URL contained ``window/[object Object]/size`` instead of ``window/current/size``.
Selenium server seems to ignore it and works anyway, but PhantomJS does not.

### DIFF
--- a/test/spec/desktop/windowHandleSize.js
+++ b/test/spec/desktop/windowHandleSize.js
@@ -2,7 +2,19 @@ import conf from '../../conf/index'
 
 describe('windowHandleSize', () => {
     it('should change window size of current window if no handle is given', async function () {
-        await this.client.windowHandleSize({ width: 500, height: 500 })
+        const requestHandler = this.client.requestHandler
+        const origHandlerCreate = requestHandler.create
+
+        this.client.requestHandler.create = function (options, data) {
+            expect(options.path).to.match(/\/window\/current\/size$/)
+            return origHandlerCreate.call(requestHandler, ...arguments)
+        }
+
+        try {
+            await this.client.windowHandleSize({ width: 500, height: 500 })
+        } finally {
+            requestHandler.create = origHandlerCreate
+        }
 
         const windowSize = await this.client.windowHandleSize()
         windowSize.value.width.should.be.equal(500)
--- a/lib/protocol/windowHandleSize.js
+++ b/lib/protocol/windowHandleSize.js
@@ -6,12 +6,12 @@
     :windowHandleSize.js
     // get the size of
     // a specified window
-    client.windowHandleSize('dc30381e-e2f3-9444-8bf3-12cc44e8372a', function(err,res) { .. });
+    client.windowHandleSize('dc30381e-e2f3-9444-8bf3-12cc44e8372a');
 
     // the current window
-    client.windowHandleSize(function(err,res) { ... });
+    client.windowHandleSize();
 
-    // change the position of
+    // change the size of
     // a specified window
     client.windowHandleSize('dc30381e-e2f3-9444-8bf3-12cc44e8372a', {width: 800, height: 600});
 
@@ -34,6 +34,10 @@ import { ProtocolError } from '../utils/ErrorHandler'
 let windowHandleSize = function (windowHandle = 'current', size) {
     let data = {}
 
+    if (typeof windowHandle === 'object') {
+        [windowHandle, size] = ['current', windowHandle]
+    }
+
     /*!
      * protocol options
      */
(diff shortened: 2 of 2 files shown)


christian-bromann:
@ipoddubny how do you use `windowHandleSize`? It seems that you had the size object as first parameter.

ipoddubny:
Yes, the size object is the first parameter, and it is ok according to the comments in the file.

Here's where it's called:
 - https://github.com/webdriverio/webdriverio/blob/master/lib/commands/setViewportSize.js#L44
 - https://github.com/webdriverio/webdriverio/blob/master/lib/commands/setViewportSize.js#L67

The bug was introduced in the commit c882d6a089:
```diff
-    var requestOptions = {
-        path: '/session/:sessionId/window/:handle/size'.replace(/:handle/, typeof windowHandle === 'string' ? windowHandle : 'current'),
+    let requestOptions = {
+        path: `/session/:sessionId/window/${handle}/size`,
         method: 'GET'
-    };
```

christian-bromann:
I see, thanks for pointing that out. Could you add a simple integration test to make sure this doesn't happen again?

ipoddubny:
I've added a test, not sure if it's simple enough but it works.

test/spec/desktop/windowHandleSize.js
```diff
@@ -2,7 +2,19 @@ import conf from '../../conf/index'
 
 describe('windowHandleSize', () => {
     it('should change window size of current window if no handle is given', async function () {
-        await this.client.windowHandleSize({ width: 500, height: 500 })
```
christian-bromann: I am confused, why does this test has passed in the past? I just ran it on the current master branch and it works fine. So what's the actual problem then?

christian-bromann:
@ipoddubny I don't get the bug this PR is solving. According to the windowHandleSize test, having the size object as first parameter seems to work fine. Could you paste an example that reproduces the problem?

test/spec/desktop/windowHandleSize.js
ipoddubny: It does not pass on master branch on my machine. And it cannot pass, since I've added a check that the URL ends with "/window/current/size", and in the current master branch it actually sends requests to "/window/[object Object]/size". You can see this in the Selenium server log, for example.

ipoddubny:
When you call windowHandleSize with an object, you get for example ``windowHandle = { width: 123, height: 123}, size = undefined``.
And then...
```js
    let requestOptions = {
        path: `/session/:sessionId/window/${windowHandle}/size`,
        method: 'GET'
    }
```
...you get "window/[object Object]/size" in requestOptions.path.

christian-bromann:
You're right! I turned on the WebdriverIO logs and got:

```

[15:24:32]  COMMAND     POST     "/wd/hub/session"
[15:24:32]  DATA                {"desiredCapabilities":{"javascriptEnabled":true,"locationContextEnabled":true,"handlesAlerts":true,"rotatable":true,"browserName":"chrome","chromeOptions":{"args":["--disable-popup-blocking"]},"loggingPrefs":{"browser":"ALL","driver":"ALL"},"requestOrigins":{"url":"http://webdriver.io","version":"4.0.3","name":"webdriverio"}}}
[15:24:34]  INFO        SET SESSION ID 41c10979-6b27-45e5-97d5-f379d26a5a0f
[15:24:34]  RESULT              {"applicationCacheEnabled":false,"rotatable":false,"chrome":{"userDataDir":"/var/folders/q4/gbz7qjl95m90bpmcj7l_3bqm0000gn/T/.org.chromium.Chromium.WVEI69"},"takesHeapSnapshot":true,"databaseEnabled":false,"handlesAlerts":true,"version":"48.0.2564.116","platform":"MAC","browserConnectionEnabled":false,"nativeEvents":true,"acceptSslCerts":true,"webdriver.remote.sessionid":"41c10979-6b27-45e5-97d5-f379d26a5a0f","locationContextEnabled":true,"webStorageEnabled":true,"browserName":"chrome","takesScreenshot":true,"javascriptEnabled":true,"cssSelectorsEnabled":true}
  windowHandleSize boob
[15:24:34]  COMMAND     POST     "/wd/hub/session/41c10979-6b27-45e5-97d5-f379d26a5a0f/url"
[15:24:34]  DATA                {"url":"http://127.0.0.1:8080/index.html"}
[15:24:36]  RESULT              null
[15:24:36]  COMMAND     POST     "/wd/hub/session/41c10979-6b27-45e5-97d5-f379d26a5a0f/window/[object%20Object]/size"
[15:24:36]  DATA                {"width":500,"height":500}
[15:24:36]  RESULT              null
[15:24:36]  COMMAND     GET      "/wd/hub/session/41c10979-6b27-45e5-97d5-f379d26a5a0f/window/current/size"
[15:24:36]  DATA                {}
[15:24:36]  RESULT              {"width":500,"hCode":500,"class":"org.openqa.selenium.Dimension","height":500}
    ✓ should change window size of current window if no handle is given (372ms)
```

So it seems that the chromedriver is treating unknown windows to be the current window because the browser window actually changes the size properly.

Good catch @ipoddubny :+1:

PR #1173, opened 2016-03-17
You need to enable co-mocha or all your tests will silently pass

graingert:


christian-bromann:
@graingert there is no support for generators planned in v4. All commands are running synchronous anyway.

graingert:
@christian-bromann this is a documentation update

christian-bromann:
I know but since there is absolutely no need to use co-mocha in v4 there shouldn't be any documentation around it

christian-bromann:
Please note that some parts of our docs aren't still updated :-/

PR #1175, opened 2016-03-17
don't escape multiple reporters

graingert:


graingert:
This causes reporters to be equal to "wdio-dot&#x27;,&#x27;junit-reporter"

christian-bromann:
:+1:

PR #1176, opened 2016-03-17
Fix some grammer in sync config default comment-docs.

graingert:


christian-bromann:
Could you also apply this change to `/examples/wdio.conf.js`?

christian-bromann:
Thanks!

PR #1197, opened 2016-03-26
Run proper service hook according the argument

just-boris:
Now always is called `onPrepare` hook, even if `onComplete` was meant


christian-bromann:
Good catch!

PR #1201, opened 2016-03-29
Update install command to junit-reporter

dcypherthis:
The doc was still referencing the dot-reporter instead of the junit-reporter. Likely a copy-and-paste error.


christian-bromann:
:+1:

PR #1203, opened 2016-03-29
Fix spacing in `waitForExist` if `reverse` is set

ChrisOelmueller:
The previous version would return `"still  existing"` with two blanks.


christian-bromann:
:+1: thanks

Issue #1205, opened 2016-03-30
spec and xunit reporters do not work


gravitypersists:
wdio.conf.js [here](https://gist.github.com/gravitypersists/210e208c5df23d67b8026bc1b792990e)

using `v4.0.5`

`npm install wdio-spec-reporter` is reporting a missing package

```
/Users/michael/workspace/acceptance-tests/node_modules/webdriverio/build/lib/launcher.js:127
                        throw _iteratorError;
                        ^

Error: reporter "wdio-spec-reporter" is not installed. Error: Error: Cannot find module 'wdio-spec-reporter'
    at Function.Module._resolveFilename (module.js:326:15)
    at Function.Module._load (module.js:277:25)
    at Module.require (module.js:354:17)
    at require (internal/module.js:12:17)
    at Launcher.initReporters (/Users/michael/workspace/acceptance-tests/node_modules/webdriverio/build/lib/launcher.js:102:40)
    at new Launcher (/Users/michael/workspace/acceptance-tests/node_modules/webdriverio/build/lib/launcher.js:43:31)
    at Object.<anonymous> (/Users/michael/workspace/acceptance-tests/node_modules/webdriverio/build/lib/cli.js:354:20)
    at Module._compile (module.js:410:26)
    at Object.Module._extensions..js (module.js:417:10)
    at Module.load (module.js:344:32)
    at Launcher.initReporters (/Users/michael/workspace/acceptance-tests/node_modules/webdriverio/build/lib/launcher.js:104:35)
    at new Launcher (/Users/michael/workspace/acceptance-tests/node_modules/webdriverio/build/lib/launcher.js:43:31)
    at Object.<anonymous> (/Users/michael/workspace/acceptance-tests/node_modules/webdriverio/build/lib/cli.js:354:20)
    at Module._compile (module.js:410:26)
    at Object.Module._extensions..js (module.js:417:10)
    at Module.load (module.js:344:32)
    at Function.Module._load (module.js:301:12)
    at Module.require (module.js:354:17)
    at require (internal/module.js:12:17)
    at Object.<anonymous> (/Users/michael/workspace/acceptance-tests/node_modules/webdriverio/bin/wdio:5:1)
npm ERR! Test failed.  See above for more details
```

marcodejongh:
@gravitypersists  This is a duplicate
https://github.com/webdriverio/webdriverio/issues/1129
Try searching next time

gravitypersists:
[...]

PR #1213, opened 2016-03-31
update wdio conf files to indicate dot is the only reporter supported…

gravitypersists:
see https://github.com/webdriverio/webdriverio/issues/1205


christian-bromann:
Thanks a ton! :+1:

PR #1230, opened 2016-04-07
manually fetches screenshot if no data was provided

sullenor:
simple solution for https://github.com/webdriverio/webdriverio/issues/1229

PR #1249, opened 2016-04-14
Bugfix: take screenshots on failure

### LINKED ISSUE #1229: provide screenshot data to the event
Hi,

I found out that sometimes webdriver doesn't provide enougth information to the allure reporter which may result in errors.

I made a small research and found out if webdriver tries to make a screenshot manually, it emits screenshot event with empty data:

https://github.com/webdriverio/webdriverio/blob/master/lib/webdriverio.js#L218
https://github.com/webdriverio/webdriverio/blob/master/lib/webdriverio.js#L221-L225

which is collected by baseReporter:

https://github.com/webdriverio/webdriverio/blob/master/lib/runner.js#L91-L98
https://github.com/webdriverio/webdriverio/blob/master/lib/utils/BaseReporter.js#L82-L84

So, in the end we will get something like this:

```javascript
{
  event: 'runner:screenshot',
  specs: ['xx'],
  filename: 'name_of_the_file',
  data: undefined,
}
```

And exception occurs somewhere here after: https://github.com/webdriverio/wdio-allure-reporter/blob/master/lib/reporter.js#L162 (see the issue: https://github.com/webdriverio/wdio-allure-reporter/issues/7).

I managed to build a small example which reproduces it: https://github.com/sullenor/wdio-bugreport/tree/reporter

gronke:
Screenshots that are taken automatically on test failure should emit the `runner:screenshot` event before `test:fail`, so that reporters can group them according to the failed test.

This fix is related to https://github.com/webdriverio/wdio-allure-reporter/pull/8 and #1230 

- fixes #1229

### DIFF
--- a/lib/webdriverio.js
+++ b/lib/webdriverio.js
@@ -173,8 +173,6 @@ let WebdriverIO = function (args, modifier) {
         if (!options.isWDIO && typeof onRejected === 'function') {
             delete err.screenshot
             return onRejected(err)
-        } else if (typeof onRejected === 'function') {
-            onRejected(err)
         }
 
         if (!this) {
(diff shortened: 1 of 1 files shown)


christian-bromann:
I came up with an own solution that is a bit cleaner in my opinion. But I applied your suggested changes.

PR #1300, opened 2016-05-12
Adding clarification to 'first step' guide

ShelbyZ:
## Proposed changes

Running through the initial setup guide presented a problem that I did not have FireFox installed on my test system.  It seems like there were two options either installing FireFox or finding a web driver.  This felt like a step that could be mentioned in the initial documentation as it would effect some users of WebdriverIO.

## Types of changes

## Checklist

- [x ] I have read the [CONTRIBUTING](/CONTRIBUTING.md) doc

Guide expects user has FireFox installed which may not be correct in all cases and present the user with error information in their command prompt running selenium.  While that error information is useful enough to point out that selenium cannot find the binary path, it does create a disconnect between new users of WebdriverIO.

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -56,7 +56,11 @@ webdriverio
     .end();
 ```
 
-** 6. Run your test file**
+** 6. Install browser and appropriate driver**
+
+In the example above FireFox is specified as the desired browser, but may not be currently installed on the testing system. The choice is to either install FireFox or download [Third Party Drivers](http://docs.seleniumhq.org/download/#thirdPartyDrivers) and place them in the same directory as the selenium standalone server.
+
+** 7. Run your test file**
 ```sh
 $ node test.js
 ```


docs/guide.md
```diff
@@ -56,7 +56,11 @@ webdriverio
     .end();
 ```
 
-** 6. Run your test file**
+** 6. Install browser and appropriate driver**
+
+In the example above FireFox is specified as the desired browser, but may not be currently installed on the testing system. The choice is to either install FireFox or download [Third Party Drivers](http://docs.seleniumhq.org/download/#thirdPartyDrivers) and place them in the same directory as the selenium standalone server.
```
christian-bromann: > and place them in the same directory as the selenium standalone server.

What does this mean? Just put the binary in the same directory as your jar file? I don't thin that is how it works

ShelbyZ: Well there are two options, they could install the drivers in a globally available path location or local to the selenium jar file.  The only documentation I could quickly find that references adding the path to drivers was in FireFox's documentation around their Marionette executable - [Setting up the Marionette executable](https://developer.mozilla.org/en-US/docs/Mozilla/QA/Marionette/WebDriver#Setting_up_the_Marionette_executable).

christian-bromann: But you say: _"[...] to either install Firefox or install Third Party Drivers [...]"_ which is not quite right. If you install Firefox at the default path you should be find by running a selenium standalone server. However if you only want to run a driver you need to start the driver server. I just don't get what you mean by this sentence.

ShelbyZ: I suppose there are two ways of looking at it either selenium standalone server can load the web driver or the driver can be run standalone.  If the goal of the guide is to have users run with FireFox then they are required to have it installed.  In the other case, assuming they want to utilize WebdriverIO, they should download a web driver and have selenium standalone server take care of running it.

christian-bromann: Yes, so if you want to run a Firefox test you should have Firefox installed on your system.

christian-bromann:
Sorry but I don't think this is a good addition to the docs, it rather complicates things. Thanks though!

Issue #1303, opened 2016-05-13
Be more flexible when using elementIds vs. selectors


lutzj82:
I'm using setImmediateValue to automate an iOS app.  The function works, but I'd like to see it use a selector instead of having to pass elementId.  Ideally, it should support the synchronous model as well.

Thanks!

example: 
`let element = browser.element('#elId');
element.setImmediateValue('blah');
`

dcypherthis:
@lutzj82 You don't' have to define element first. when calling `setImmediateValue()` you can pass 2 arguments, the first is the element you are targeting, and the second one is the value you want to set.  

Example:

`browser.setImmediateValue('elementId#', 'blah');

You can see this in the docs here: http://webdriver.io/api/mobile/setImmediateValue.html

agauthiez:
@dcypherthis Indeed it takes 2 arguments but the documentation is unclear. According to the source, it takes a

christian-bromann:
@agauthiez sorry I had to close the PR because it wasn't really a solution of the actual problem. So I had this in mind for quite some time now: allow using elementIds when selector are required + allow to use selectors when elementIds are required (mostly when protocol commands are executed). What do you think?

We need to find a solution that works across all action/protocol commands

agauthiez:
@christian-bromann: No problem. I understand your point and I agree. Does it mean you would consider deprecating a bunch of protocol commands? For instance, if `click` handles element ids we could drop `elementIdClick`. :smiley:

christian-bromann:
> Does it mean you would consider deprecating a bunch of protocol commands?

This is a good question. Don't know yet.

christian-bromann:
This issue is obsolete with update to v5. Closing!

PR #1310, opened 2016-05-17
Selector handling for setImmediateValue

agauthiez:
## Proposed changes

`setImmediateValue` only takes an element id as first parameter, thus an id has to be retrieved before using this command. As suggested in #1303, I implemented a handler for selectors so we can call the method directly, e.g. `setImmediateValue('~myInput', 'foo')`.

## Types of changes

## Checklist

## Further comments

- As my implementation involves performing 2 protocol commands call, should we move `lib/protocol/setImmediateValue.js` to `lib/commands/setImmediateValue.js`?
- I rewrote the documentation so it is clearer what kind of arguments can be passed
- I did not added tests for it (yet?)

### DIFF
--- a/lib/protocol/setImmediateValue.js
+++ b/lib/protocol/setImmediateValue.js
@@ -3,8 +3,11 @@
  * Set immediate value in app.
  *
  * <example>
-    :setImmediateValueSync.js
-    browser.setImmediateValue(el, 'foo')
+    :setImmediateValueId.js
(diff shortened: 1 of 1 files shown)


christian-bromann:
This would fix the issue only for this command but there are other protocol commands that face similar issues.

christian-bromann:
Let's continue discussion at #1303

PR #1311, opened 2016-05-17
Export Launcher

monolithed:
## Proposed changes

I have to launch my tests in a custom grunt plugin. And I found a [way](https://github.com/webdriverio/grunt-webdriver/blob/master/src/webdriver.js#L9):

```js
import path from 'path'
import resolve from 'resolve'

const Launcher = require(path.join(path.dirname(resolve.sync('webdriverio')), 'lib/launcher'));
new Launcher(...);
```

But this code is ugly, so I think we should to export the Launcher.

```js
import WebdriverIO from 'webdriverio';

new WebdriverIO.Launcher(...);
```

## Types of changes

@christian-bromann, what do you think?

### DIFF
--- a/index.js
+++ b/index.js
@@ -19,6 +19,7 @@ import WebdriverIO from './lib/webdriverio'
 import Multibrowser from './lib/multibrowser'
 import ErrorHandler from './lib/utils/ErrorHandler'
 import getImplementedCommands from './lib/helpers/getImplementedCommands'
+import Launcher from './lib/launcher'
 import pkg from './package.json'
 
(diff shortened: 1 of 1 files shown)


christian-bromann:
Yeah, I see the value in exporting it. 👍

monolithed:
@christian-bromann, I would appreciate if you could publish to npm :)

christian-bromann:
There will be a release later today

monolithed:
@christian-bromann, what about release? 🙂

christian-bromann:
just released a new version

monolithed:
Thanks!

PR #1317, opened 2016-05-19
add beforeSession hook

sebastiendavid:
## Proposed changes

Hi, I propose to add a hook executed before webdriver session creation. For example, I need it to set dynamically names of browserstack sessions. It can't be done during `onPrepare` hook when there are multiple specs or multiple capabilities, and a browserstack session name cannot be edited after its creation.
## Types of changes
## Checklist
## Further comments

About the runner: not sure about how to test it... at least tests passed (test:wdio). Tell me if I forgot something.


christian-bromann:
Hi,
what kind of names do you assign to the browser stack session?

sebastiendavid:
For example I want to set dynamically the name of a capability before each specs:
```
  {
      browserName: 'chrome',
      browser: 'Chrome',
      browser_version: '50',
      os: 'Windows',
      os_version: '10',
      resolution: '1280x800',
      project: pkg.name,
      build: pkg.version,
      name: 'SOME NAME'
  }
```
Name which can be displayed by the browserstack automate instead of the session id.
screen shot 2016-05-19 at 22 16 03
This name has to be post when the session is created (POST on http://hub.browserstack.com/wd/hub/session)

christian-bromann:
Browserstacks API allows to do that with a request also after the test right? Also you can change the capabilities in the onPrepare hook too.

sebastiendavid:
I tried to do that, but Browserstack API just allow to edit the session status (not the name), and the onPrepare hook is executed once and not before each session.

christian-bromann:
So you only want to set the name for your capability? Why do you need to do that dynamically before the session starts?

sebastiendavid:
To have in Browserstack a list of well named sessions and not a big list of session ID or with the same name.

christian-bromann:
So and why do you need this exactly before the session starts?

sebastiendavid:
Ok, let's consider this conf.wdio.js:
```
{
  ...,
  specs: [
    './tests/scenarios/noop.scenario.js',
    './tests/scenarios/noop2.scenario.js',
  ],
  capabilities: [
    {
      browserName: 'chrome',
      browser: 'Chrome',
      browser_version: '50',
      os: 'Windows',
      os_version: '10',
      resolution: '1280x800',
      project: pkg.name,
      build: pkg.version,
      name: 'some name'
    }
  ],
  ...
} 
```
Tests are launched with wdio cli, I added some logs in hooks:
```
> onPrepare
> before
> before
> done!
```
Results in browserstack:
screen shot 2016-05-20 at 17 53 16
As you can see, the name is the same for each spec, if you don't set a name: you'll have the session IDs.
Now if I set a name during onPrepare hook, only the first spec will display this name, the second one will have a session id or the name setted in the capability from wdio.conf.js.

So why do I need this before the session starts ?
Because I want to have nice names in Browserstack. If you have a better solution to be well integrated with Browserstack: tell me please...

christian-bromann:
> Because I want to have nice names in Browserstack.

I agree, but what difference does it make to do that in the `onPrepare` or `beforeSession` hook? I would ask Browserstack to enhance their API in order to allow setting session names after the test has run.

sebastiendavid:
> what difference does it make to do that in the onPrepare or beforeSession hook

`onPrepare` is executed once, `beforeSession` before each test execution as `before` hook, but the difference is that `beforeSession` is executed before `browser.init()`.

> I would ask Browserstack to enhance their API

The issue here is not Browserstack API, it's definitly wdio cli which have a lot of nice hooks but not the one I need.

Thank you for your time, I'll find another solution. Regards.

avkrash:
Let me put my two cents.
I think this hook is a good idea. At the moment I need to add extensions for some tests, but other should run without it. And I have no idea how to implement it without separate testruns.
More hooks more  possibilities.

christian-bromann:
@avkrash good point indeed. After reviewing this PR again, it seems to be a valid addition after all. Sorry for not seeing this earlier @sebastiendavid .

avkrash:
Thanks for fast answer @christian-bromann.
Glad you endorse this idea.

christian-bromann:
see #1745

PR #1325, opened 2016-05-23
Don't set a default browserName when isMobile

uniqueid001:
## Proposed changes

I want to test an Android app using appium.  In `desiredCapaiblities`, I set `browserName=''` and left `app` undefined.  appium rejected the session due to unsupported browser `'firefox'`.  I consciously left `app` undefined because I want to run my test on the already-installed app rather than installing a new version.  This PR removes the test for truthiness of `desiredCapabilities.app`, instead utilizing the existing `mobileDetector()` function to identify when a mobile app is being tested so that `browserName` can be left as an empty string or undefined.

## Types of changes

Test `isMoble` rather than `desiredCapabilities.app` when determining to
set a default value for `desiredCapabilities.browserName`.  This allows
allows `desiredCapabilities.app` to be blank, which instructs appium to
skip the upload and installation of an app, instead using the
already-installed copy.

### DIFF
--- a/lib/webdriverio.js
+++ b/lib/webdriverio.js
@@ -89,14 +89,14 @@ let WebdriverIO = function (args, modifier) {
         rotatable: true
     }, options.desiredCapabilities || {})
 
+    let { isMobile, isIOS, isAndroid } = mobileDetector(desiredCapabilities)
+
     /**
      * if no caps are specified fall back to firefox
      */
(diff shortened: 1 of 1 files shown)


christian-bromann:
👍🏻

uniqueid001:
The problem is that mobileDetector() returns `isMobile===true` when `caps.browserName` is falsey.  This is incorrect.

Issue #459, opened 2015-03-13
Using with node-inspector


thedeeno:
I'm trying to attach node-inspector to my test run and drop a debugger in one of my steps. I can successfully attach the inspector, but I can't drive the browser one execution has stopped.

If I hit the debugger in the following example and type `client.url('http://hello.com')` I get a `WebdriverIO` promise back, but my selenium-standalone never receives get message.

    debugger;
    // try to use client from node inspector here
    client.url('http://google.com').call(done);

What am I doing wrong?

christian-bromann:
@thedeeno could you paste in some Selenium / WebdriverIO logs?

zerbons:
@christian-bromann it seems to not execute until you exit the debugger. Any ideas on how to achieve a REPL interface for webdriverio?

christian-bromann:
@zerbons WebdriverIO doesn't have a REPL interface yet. `v3` will introduce a method called `debug` that allows you to stop the queue and fiddle around in the browser. However I would like to see support for that in the future

stoffeastrom:
@christian-bromann 

As I reported on gitter [here](https://gitter.im/webdriverio/webdriverio?at=55e6a90ce55df6ee79bf29cf). It's possible but maybe not the easiest way. I think it would be really nice to have a way of running the capabilities in the same process. This would make the debugging experience so much better. We could introduce a `debug` argument to the `cli`. This would just run them in the same process instead of forking child processes. Does this make sense?

marthinus-engelbrecht:
@stoffeastrom Thanks for the tip :+1: , got it working for me. @christian-bromann would really be awesome to see a more integrated solution tough.

georgecrawford:
Just adding my :+1: - a REPL to send webdriverio commands to the browser would be _amazingly_ useful.

christian-bromann:
Debugging ability just got published with v4.0.9. For info see http://webdriver.io/guide/testrunner/debugging.html.

@georgecrawford I believe a repl functionality is a different topic but definitely nice to have.

kurtharriger:
For those interested in a repl.  I hacked this together https://github.com/kurtharriger/wdio-repl as an alternative to the debugger.  Supports es6 and await which is useful for writing tests and experimenting.

christian-bromann:
@kurtharriger 👏  this is so awesome.

Viky-zhang:
[...]

PR #1345, opened 2016-05-26
Debugging w/ node inspector

kurtharriger:
## Proposed changes

This enables forking child processes with a different debugger port for each node process to enable the use of node-debugger. It is loosely based on the gitter code referenced in this issue https://github.com/webdriverio/webdriverio/issues/459.

 The config.debug flag is used so that it is not necessary debug the launcher process to debug child processes. I also choose to use `--debug` only and not `--debug-brk` as it appears that adding debugger in my code was sufficient to pause until debugger was attached and you can just leave the debug flag on if desired.

Port is based on capability index so first child process would be on 5859 and second child process on 5860.  When using this it is often useful to set config.maxInstances to 1 as you're unlikely to try and debug them all at the same time. I thought about doing this automatically when debug option was specified, but decided to handle that when in my own config generation, such as:

```
debug: !!process.env.DEBUG,
maxInstances: process.env.DEBUG ? 1 : 100,
```

Then run:
```node-inspector --debug-port 5859```

It also often helps to use `--no-preload` to load faster when adding `debugger` in the test itself.

I was also pleasantly surprised that when sync is true that async commands actually behave synchronously in the node-inspector console which makes for a pretty nice repl as well.

If you want to add `--debug-brk`, explicitly specify a different port, or add other options you can do so in using `execArgv` config option.

## Types of changes

## Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,7 +25,7 @@ to know that you have a clean state.
 
   or install it with npm package [selenium-standalone](https://github.com/vvo/selenium-standalone)
   ```sh
-  $ npm install -g selenium-standalone
+  $ npm install -g selenium-standalone phantomjs
   $ selenium-standalone install
   $ selenium-standalone start
   ```
--- a/docs/guide/getstarted/configuration.md
+++ b/docs/guide/getstarted/configuration.md
@@ -109,6 +109,18 @@ Default interval for all waitForXXX commands.
 Type: `Number`<br>
 Default: *250*
 
+## debug
+Enables node debugging
+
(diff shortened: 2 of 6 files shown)


christian-bromann:
@kurtharriger this just made my day! Thank you so much! Code looks good but could you please write a simple manual how to debug a test in the docs? Maybe under the testrunner section?

kurtharriger:
Added additional documentation to the guide

christian-bromann:
Awesome thanks so much

christian-bromann:
published and deployed

PR #1352, opened 2016-05-28
`browserName === undefined` should not indicate mobile

uniqueid001:
## Proposed changes

Fixes an issue revealed by PR #1325:  when `caps.browserName` is left undefined
the mobileDetector() function returned `isMobile===true`.  This PR updates the `mobileDetector()` function so that a falsey `browserName` is not considered to indicate driving a mobile app.

## Types of changes

Fixes an issue revealed by PR #1325:  when browserName is left undefined
the mobileDetector function returned `isMobile===true`.  This affects
tests that do not specify any capabilities
(eg `var client = require('webdriverio').remote(); client.init()...`)

### DIFF
--- a/lib/helpers/mobileDetector.js
+++ b/lib/helpers/mobileDetector.js
@@ -12,7 +12,7 @@ let mobileDetector = function (caps) {
             (typeof caps.deviceName !== 'undefined') ||
             // Check browserName for specific values
             (caps.browserName === '' ||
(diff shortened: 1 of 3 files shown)


uniqueid001:
I would like to create some tests for the mobileDetector() function.  I can't run them now because my system at home has Ubuntu 15.10 which has nodejs v0.10.25.  I'll update this PR when and if I can create some tests for it.

uniqueid001:
I added unit tests for the mobileDetector() and the remote() functions.

christian-bromann:
❤️  this!

PR #1361, opened 2016-06-05
[FEATURE] Improve reporting of failed tests

kevinmcdonnell:
## Proposed changes

Include the parent name when reporting failed tests. Currently if `it` blocks in tests have the the same name it is not possible to determine which test failed

## Types of changes

## Checklist

- [X] I have read the [CONTRIBUTING](/CONTRIBUTING.md) doc

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

### DIFF
--- a/lib/utils/BaseReporter.js
+++ b/lib/utils/BaseReporter.js
@@ -217,7 +217,8 @@ class BaseReporter extends events.EventEmitter {
                       this.color('error message', '%s') +
                       this.color('bright yellow', '%s') +
                       this.color('error stack', '\n%s\n')
-            console.log(fmt, (i + 1), test.title, test.err.message, test.runningBrowser, test.err.stack)
+            var title = typeof test.parent !== 'undefined' ? test.parent + ' ' + test.title : test.title
+            console.log(fmt, (i + 1), title, test.err.message, test.runningBrowser, test.err.stack)
         })
     }
 


christian-bromann:
@kevinmcdonnell can we be 100% that `test.parent` will always contain a value?

kevinmcdonnell:
I could put in a check for `test.parent` and then prepend

An `it` blocks need to be nested in `describe` blocks so it seems likely that parent would always exist.

christian-bromann:
> An it blocks need to be nested in describe blocks so it seems likely that parent would always exist

That is true but doesn't necessarily apply for frameworks like Cucumber. Let's just be sure and wrap it in an if statement

lib/utils/BaseReporter.js
```diff
@@ -217,7 +217,7 @@ class BaseReporter extends events.EventEmitter {
                       this.color('error message', '%s') +
                       this.color('bright yellow', '%s') +
                       this.color('error stack', '\n%s\n')
-            console.log(fmt, (i + 1), test.title, test.err.message, test.runningBrowser, test.err.stack)
+            console.log(fmt, (i + 1), test.parent + ': ' + test.title, test.err.message, test.runningBrowser, test.err.stack)
```
monolithed:
```js
this.color('error title', '%s) %s:\n\t') +
```

Please add indentation

monolithed:
```js
this.color('error title', '%s) %s:\n\t') +
```

Please, don't forget about indentation

kevinmcdonnell: What do you mean?

monolithed: I mean that you should care about block levels.:

Actual:

```
describe
title
```

Expected:

```
describe
   title
```

reporter-spec-fail

kevinmcdonnell:
@monolithed your screenshot shows indentation which I believe is part of the spec reporter. This fix is not for a particular reporter.

Currently, your example would report without the describe context.

```
0) should return -1 when the value is not present
```

After this fix it would include the describe
```
0) #indexOf() should return -1 when the value is not present
```

I've tested this fix with and without parent being defined, though I'm not sure how to handle nested describes

christian-bromann:
Thanks @kevinmcdonnell 👍

PR #1401, opened 2016-06-21
Expose the process number to the child processes

kurtharriger:
## Proposed changes

This change provides the child process a unique incrementing instance number.

## Types of changes

## Checklist

## Further comments

Our tests startup mock services which need a unique port number, we have been randomly generating this, but we are running into few issues with this.  We would perhaps like this to be more deterministic and allocate within a given port range for each service.  Unfortunately there doesn't seem to be a way to determine how much we should offset from the start port.  

Additionally, due to parallelism log output from child processes can be interleaved which makes it difficult to correlate messages from test runner.  We can use process.pid for this but a sequential test runner name might be easier for human reader to correlate than a pid. 

I attached this value to the browser.desiredCapabilities object as this is available from within the tests setup and didn't want to add another global, but open to other suggestions on how to expose this to child runner plugins/service/setup methods. 

I did see that cid was recently changed to be 1a, 1b, etc... but the the cid doesn't appear to be available to plugins, hooks, etc... so not sure if there is a way to get it.  But the partitioning here by capability isn't desired and we would need to convert alpha value back to numeric offset and leave gaps in address range to account for unknown number runners per capability.  Also not sure how this behaves if there are more than 26 processes? and we think we will probably exceed this as we scale up.

### DIFF
--- a/lib/launcher.js
+++ b/lib/launcher.js
@@ -21,6 +21,7 @@ class Launcher {
         this.processes = []
         this.schedule = []
         this.rid = []
+        this.processesStarted = 0
     }
 
     /**
@@ -249,7 +250,7 @@ class Launcher {
         let config = this.configParser.getConfig()
         let debug = caps.debug || config.debug
         let rid = this.getRunnerId(cid)
-        let processNumber = this.getNumberOfRunningInstances() + 1
+        let processNumber = this.processesStarted + 1
 
         // process.debugPort defaults to 5858 and is set even when process
         // is not being debugged.
@@ -290,6 +291,8 @@ class Launcher {
             specs,
             isMultiremote: this.isMultiremote()
         })
+
+        this.processesStarted++
     }
 
     /**
--- a/lib/launcher.js
+++ b/lib/launcher.js
@@ -249,11 +249,12 @@ class Launcher {
         let config = this.configParser.getConfig()
         let debug = caps.debug || config.debug
         let rid = this.getRunnerId(cid)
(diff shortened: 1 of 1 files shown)


lib/launcher.js
```diff
@@ -255,11 +255,12 @@ class Launcher {
         let config = this.configParser.getConfig()
         let debug = caps.debug || config.debug
         let rid = this.getRunnerId(cid)
+        let processNumber = this.getNumberOfRunningInstances() + 1
```
christian-bromann: not sure if that method returns an unique process number

christian-bromann:
> Also not sure how this behaves if there are more than 26 processes? and we think we will probably exceed this as we scale up.

That would only happen if a single capability would need to test more than 26 tests and even than it would just continue with `0{` which is fine I think

lib/launcher.js
kurtharriger: You're right.  I thought I could just use the number of running instances instead of adding state, but I didn't think about what would happen after a process terminates and the count is decremented.  So if maxInstances is 10 the 11th process would have id 10 and 10 may still be running.

kurtharriger:
Yeah I guess that is fine, but I don't think cid is available outside the runner class?

From a hook we do something like this:
new Drymock({port: randomPort})

We are running into address in use errors even though we are using a port scanner and we would like to do something more like this instead of random ports:
new Drymock({port: 27000+browser.desiredCapabilities.processNumber})

christian-bromann:
I assume this is ready to get merged.

PR #1403, opened 2016-06-22
Don't resume stdin if not paused

timdp:
May seem redundant, but it keeps Node from hanging (at least) on Windows for some reason.

I inspected `process._getActiveHandles()` after a `Launcher#run()` and found out that there was a `ReadStream` still open where `isTTY=true`. Thus, I looked for `stdin` in the code and found this line. Took a wild guess by adding the `if` and it worked.

Tested with gulp-webdriver 2.0.1 + Node 4.4.4 + Windows 10 x64.

### DIFF
--- a/lib/launcher.js
+++ b/lib/launcher.js
@@ -158,7 +158,9 @@ class Launcher {
         /**
          * make sure the program will not close instantly
          */
-        process.stdin.resume()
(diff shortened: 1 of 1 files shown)


christian-bromann:
LGTM, nice catch

PR #1417, opened 2016-06-29
Update waitForExists docs

amilner42:
## Proposed changes

Added 4 words to clarify documentation

Under the situation that the element is never discovered, instead of false being returned, a error is thrown, thought it would be worth adding a line in the docs to explain this (tripped me up).


christian-bromann:
LGTM 👍

PR #1423, opened 2016-06-30
Put an example to documentation

cagdastulek:
## Proposed changes

There is no example explaining how to use a WebElementJSONObject response as a parameter to `frame` function. I am a newbie, so maybe I am not doing it right but this example code worked for me fine. I am definitely open to suggestions for a better approach.

## Types of changes

## Checklist

### DIFF
--- a/lib/protocol/frame.js
+++ b/lib/protocol/frame.js
@@ -3,7 +3,7 @@
  * the server should switch to the page's default content.
  * <example>
     :frameExample.js
-    # Using `element` to find an iframe and providing it to `frame` method
+    // Using `element` to find an iframe and providing it to `frame` method
     browser.waitForExist('iframe[name="my_iframe"]');
     var my_frame = browser.element('iframe[name="my_iframe"]').value;
     browser.frame(my_frame);
--- a/lib/protocol/frame.js
+++ b/lib/protocol/frame.js
@@ -2,7 +2,8 @@
  * Change focus to another frame on the page. If the frame id is null,
  * the server should switch to the page's default content.
  * <example>
-    :Using `element` to find an iframe and providing it to `frame`
+    :frameExample.js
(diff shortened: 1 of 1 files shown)


lib/protocol/frame.js
```diff
@@ -1,6 +1,12 @@
 /**
  * Change focus to another frame on the page. If the frame id is null,
  * the server should switch to the page's default content.
+ * <example>
+    :Using `element` to find an iframe and providing it to `frame`
```
christian-bromann: this line is suppose to be the file name. To stay consistent with the other commands let's change that to be an file name and make the sentence to a comment

christian-bromann:
One comment otherwise LGTM

lib/protocol/frame.js
cagdastulek: What should be the value of filename? I copied the format from the following link but I didn't understand what the filename stands for (couldn't find it anywhere). See https://github.com/webdriverio/webdriverio/edit/master/lib/commands/waitForExist.js

christian-bromann: @cagdastulek just something like `frameExample.js`

cagdastulek:
@christian-bromann Just did the suggested change. Thanks for the help.

lib/protocol/frame.js
```diff
@@ -1,6 +1,13 @@
 /**
  * Change focus to another frame on the page. If the frame id is null,
  * the server should switch to the page's default content.
+ * <example>
+    :frameExample.js
+    # Using `element` to find an iframe and providing it to `frame` method
```
christian-bromann: This is not a valid JavaScript comment 😉

cagdastulek: Life is hard. Curse of working on multiple languages 😅

cagdastulek:
Done 😄

PR #1427, opened 2016-07-01
Added waitForUrl command

monolithed:
## Proposed changes

`waitForUrl` command waits for the current page url to match the provided value.
It's useful when you have a redirection and need to wait for the final url
 * to be loaded, before running you're next command.

## Types of changes

## Checklist

### DIFF
--- a/lib/commands/waitForUrl.js
+++ b/lib/commands/waitForUrl.js
@@ -1,0 +1,57 @@
+/**
+ *
+ * Waits for the current page url to match the provided value.
+ * It's useful when you have a redirection and need to wait for the final url
+ * to be loaded, before running you're next command.
+ *
+ * <example>
+    :waitForUrlExample.js
+    it('should detect the expected url', function () {
+        browser.waitForValue('https://example.com');
+    });
+ * </example>
+ *
+ * @alias browser.waitForUrl
+ * @param {Number=}  ms       time in ms (default: 500)
+ * @param {Boolean=} reverse  if true it waits for the opposite (default: false)
+ * @returns {Boolean}
+ * @uses util/waitUntil, property/getUrl
+ * @type utility
+ *
+ */
+
+import { WaitUntilTimeoutError, isTimeoutError } from '../utils/ErrorHandler'
+
+let waitForUrl = function (url, ms, reverse) {
+    /**
+     * we can't use default values for function parameter here because this would
+     * break the ability to chain the command with an element if reverse is used, like
+     *
+     * ```js
+     * var elem = browser.element('#elem');
+     * elem.waitForXXX(10000, true);
+     * ```
+     */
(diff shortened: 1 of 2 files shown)


patthiel:
Hah, we have our own version of this command but it's 10x more hacky. 👍

christian-bromann:
@monolithed I am not sure if this is worth adding to the API. First of it turns out to be not useful if you run tests in different environments where your test pages are something like `staging.*.com`, `dev.`.com` etc. So it makes more sense to do this:

```js
browser.waitUntil(() => browser.getUrl().includes('/somepage'), 'some page never loaded`);
```

Also usually Selenium automatically waits until a page is fully loaded. It also considers redirects.

monolithed:
>  I am not sure if this is worth adding to the API. 

Hm. I've a lot of test cases with the following steps:

1. Open some page
2. Click on "sign up" link
3. Wait "http://....." link

It has been [implemented](http://docs.casperjs.org/en/latest/modules/casper.html?highlight=waitForUrl#waitforurl) in the CasperJS (I know that's not so good argument though)

> Also usually Selenium automatically waits until a page is fully loaded. It also considers redirects.

There'a some restrictions

1. Click handler like `event.preventDefault()` or `return false` may prevent clicking
2. The required link may be absent (or changed)

> First of it turns out to be not useful if you run tests in different environments where your test pages are something like staging.*.com, dev..com` etc.

It can be solved with predicate (boolean-valued function)

```js
let waitForUrl = function (value, ms, reverse) {
       return this.waitUntil(() => {
             let url = this.getUrl();
             url = typeof value === 'function' ? value(url) : value;

             return url && url === value && !reverse
       })
}, ms);
```

```js
waitForUrl(value => value.includes('/some_page'));
waitForUrl('/some_page');
```

:)

christian-bromann:
> waitForUrl(value => value.includes('/some_page'));

I don't see a big difference between that and `browser.waitUntil(() => browser.getUrl().includes('/somepage'))`. 

> Click handler like event.preventDefault() or return false may prevent clicking

That is why it makes even more sense to wait for a certain element on the expected page.

I am just not a big fan of adding more commands to the API because we already have tons of commands. Adding another one that barely adds any functionality doesn't make sense to me. Sorry. You will be fine by adding a custom command for your test suite.

monolithed:
> That is why it makes even more sense to wait for a certain element on the expected page.

It does't make sense to test the third-party pages because it takes more time and introduces some additional complexity. We have about 8 k e2e tests and `waitForUrl` is one the popular commands for us.

> You will be fine by adding a custom command for your test suite.

We have already one :)

christian-bromann:
> It does't make sense to test the third-party pages because it takes more time and introduces some additional complexity.

What do you mean by that?

monolithed:
img-2016-07-04-01-20-04

We don't care about what will happen on the next page.

PR #1445, opened 2016-07-11
Fix session reload

prshreshtha:
## Proposed changes

When renaming `refresh` to `reload`, some changes were left behind causing it to not work as expected.

## Types of changes

_See Further comments_

## Checklist

## Further comments

The only breaking part of this change would be the `onRefresh` -> `onReload` hook. I doubt there would be too many complaints though since `refresh` means something else entirely in the context of webdriverio!


christian-bromann:
Thanks! 👍

PR #1451, opened 2016-07-12
Fixed travis ci node version

amilajack:
This will test against the most recent version of node 4

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ notifications:
 language: node_js
 
 node_js:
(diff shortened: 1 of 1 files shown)


christian-bromann:
looks good, I didn't have any strong feelings about only testing on 4.4.3, so fine to merge it.

amilajack:
What do you think about testing against node 6?

christian-bromann:
Don't see any need for that yet. v4 is an LTS version and we should be fine with that for the next month.

PR #1458, opened 2016-07-15
Added osx support to travis ci

amilajack:


christian-bromann:
Can you please explain your PR?

amilajack:
This runs tests against osx as well as linux. This is good for finding platform specific bugs related to node-gpy or other platform-specific issues.

christian-bromann:
Why should we test on multiple platforms? The only dependency that might be affected is `wdio-sync` which depends on `fibers`. That project is well tested though. Given the fact we would run twice as much tests on an already unstable build I don't think this PR adds any value to the project.

PR #1461, opened 2016-07-17
Await error screenshot before promise rejection

just-boris:
## Proposed changes

This change is intended to fix https://github.com/webdriverio/wdio-allure-reporter/issues/24
After that, screenshot command will be part of test workflow and will be finished before the test framework will get the failure.

Also, I am aware of https://github.com/webdriverio/webdriverio/pull/1441, which contains similar changes, but I think that this is not necessary to introduce new configuration option.

## Types of changes

*Note*. According to currently existing  tests, there are nobody relied on current behavior. So, I consider this as bugfix to resolve a mess between `test:fail` and `runner:command` events.

## Checklist

### DIFF
--- a/test/spec/unit/reject.js
+++ b/test/spec/unit/reject.js
@@ -1,0 +1,61 @@
+import { remote } from '../../../index'
+import { tmpdir } from 'os'
+
+describe('rejection', () => {
+    before(async function() {
+        this.client = remote({
+            screenshotPath: tmpdir()
+        })
+
+        const createSession = global.mock('post', '/session')
+        await this.client.init()
+        createSession.isDone().should.be.true
+    })
+
+    it('should not take screenshot if rejection handler is provided', function () {
+        mock('post', '/session/123ABC/element', { status: 7 })
+
+        let screenshot
+        this.client.on('screenshot', data => {
+            screenshot = data
+        })
+
+        return this.client.click('.missing')
+            .catch(e => {
+                expect(screenshot).to.be.undefined
+                expect(e.message).to.match(/An element could not be located/)
+            })
+    })
+
+    it('should take screenshot if there is no error handler', function () {
+        mock('post', '/session/123ABC/element', { status: 7 })
+        mock('get', '/session/123ABC/screenshot', { status: 0, value: '' })
+
+        let screenshot
+        this.client.on('screenshot', data => {
+            screenshot = data
+        })
+
+        return this.client.click('.missing')
+            .then(() => { throw new Error('Unexpected success') })
+            .catch(e => {
+                expect(screenshot).to.be.defined
+            })
+    })
+
+    it('should prevent infinite loop when screenhot attempt has failed', function () {
+        mock('post', '/session/123ABC/element', { status: 7 })
+        mock('get', '/session/123ABC/screenshot', { status: 6 })
+
+        let screenshot
+        this.client.on('screenshot', data => {
+            screenshot = data
+        })
+
+        return this.client.click('.missing')
+            .then(() => { throw new Error('Unexpected success') })
+            .catch(e => {
+                expect(screenshot).to.be.undefined
+            })
+    })
+})
--- a/lib/webdriverio.js
+++ b/lib/webdriverio.js
@@ -173,23 +173,32 @@ let WebdriverIO = function (args, modifier) {
         if (!options.isWDIO && typeof onRejected === 'function') {
             delete err.screenshot
             return onRejected(err)
-        } else if (typeof onRejected === 'function') {
-            onRejected(err)
-        }
-
-        if (!this) {
-            throw err
-        } else if (this.depth !== 0) {
-            return this.promise
         }
 
         /**
-         * take screenshot only if screenshotPath is given
+         * take screenshot only if screenshotPath is given and we are on top of promise chain
          */
-        if (typeof options.screenshotPath !== 'string' || err.shotTaken) {
-            return throwException(err, stacktrace)
+        if (!this || this.depth !== 0 || typeof options.screenshotPath !== 'string' || err.shotTaken) {
+            if (typeof onRejected === 'function') {
+                onRejected(err)
+            }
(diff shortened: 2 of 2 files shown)


j0tunn:
I think of probable infinite recursion in case of `saveScreenshot` fail

just-boris:
Added on more test. Couldn't properly test integration with custom reporters, so I linked my version through `npm link webdriverio` and tested them together.

christian-bromann:
@j0tunn @just-boris it seems that both of your PRs are doing the same thing, right? Or are there any differences?

just-boris:
I see two important distinctions:

* @j0tunn introduces new option `screenshotOnReject` which enforces screenshot for rejection. By default it is only taken in case when there is no `onReject` function provided, or in `wdio` environment. My PR, keeps this logic as before.
* That option also triggers screenshot attachment for every error, even though there was no error before.

Necessary for me change regarding `onReject` comes there aside, as far as it necessary to implement their feature.

For me it looks like that feature can be easily implemented on top of this changes.

j0tunn:
@just-boris There was a recursion problem in my code on `screenshot` method fail. Code fail -> try to get screenshot -> fail -> try to get screenshot ... Take a look at this test: https://github.com/j0tunn/webdriverio/blob/d7b1050c07341b16577682caa6b924751ee6c068/test/spec/unit/remote.js#L96

> For me it looks like that feature can be easily implemented on top of this changes.

Tell me how it will look like? Really I don't want to save image to FS, subscribe to `screenshot` event, than read that image, than pass it to reporter where it will be saved to the FS again.

@christian-bromann no, our PRs about different things, this one is about fixing rejection order.
My PR is about attaching screenshot to the error.

just-boris:
@j0tunn look at [this place in your fork](https://github.com/j0tunn/webdriverio/blob/feature/screenshot.on.reject/lib/webdriverio.js#L204). You have also changed order of `onReject` call.

Will look at your test and will try to reproduce your recursion example.

j0tunn:
> You have also changed order of onReject call

Yep, it was necessary for correct work of my code.

christian-bromann:
@just-boris after looking into the PRs it looks like as it would make more sense that we integrate your PR into @j0tunn ones, what do you think?

just-boris:
Well, how we can do that? Changes should go after each other, but there are definitely conflicts.

--
By the way, I tried to get infinite loop, but I couldn't even when I tried to use `replyWithError` in nock.

just-boris:
@j0tunn, see #1478. I figured out why you got the infinite recursion. The reason is because your test is so artificial and uses `sinon` instead of `nock`.

@christian-bromann As far as the infinite recursion was possible before, but nobody suffers from it right now, I vote for merge this. Then @j0tunn will can rebase his branch with this changes

christian-bromann:
I would agree as your changes are smaller.

@j0tunn any rejections?

j0tunn:
no problem

just-boris:
@christian-bromann ping. 
Is there anything, that I should do before merge?

christian-bromann:
No looks good, thanks 👍

just-boris:
Hello @christian-bromann , 
After merge this, the build on master started failing. That happened because all my changes was based on Babel5, but now master has Babel6. It turns out, that it makes sense here.

My guess it that `this.depth` is always 0 now, because Babel changed something in stack traces

just-boris:
Here two examples: 

Webdriver.io v4.2.1 (latest version on Babel 5)
```
Error
    at Object.commandGroup.(anonymous function) [as element] (/Users/just-boris/coding/webdriverio/build/lib/webdriverio.js:606:25)
    at Object.click (/Users/just-boris/coding/webdriverio/build/lib/commands/click.js:44:17)
    at Object.exec (/Users/just-boris/coding/webdriverio/build/lib/helpers/safeExecute.js:28:24)
    at Object.resolve (/Users/just-boris/coding/webdriverio/build/lib/webdriverio.js:171:29)
    at lastPromise.then.resolve.call.depth (/Users/just-boris/coding/webdriverio/build/lib/webdriverio.js:381:32)
    at _fulfilled (/Users/just-boris/coding/webdriverio/node_modules/q/q.js:820:54)
```

Latest master (with Babel 6)
```
Error
    at Object.commandGroup.(anonymous function) [as element] (webdriverio.js:524:25)
    at Object.click (click.js:36:17)
    at Object.exec (safeExecute.js:19:24)
    at Object.resolve (webdriverio.js:111:29)
    at lastPromise.then.resolve.call.depth (webdriverio.js:324:32)
    at _fulfilled (/Users/just-boris/coding/webdriverio/node_modules/q/q.js:834:54)
```

As you can see, the path information is missing, so we are unable to detect current depth.

Currently I don't know how to fix this, any help will be appreciated

just-boris:
Found a related babel issue: https://github.com/babel/babel/pull/3523

Let's wait for fix

christian-bromann:
@just-boris excellent work! Thanks

test/spec/unit/reject.js
```diff
@@ -0,0 +1,61 @@
...
+        this.client = remote({
+            screenshotPath: tmpdir()
+        })
+
+        const createSession = global.mock('post', '/session')
+        await this.client.init()
+        createSession.isDone().should.be.true
+    })
+
+    it('should not take screenshot if rejection handler is provided', function () {
```
j0tunn: @just-boris 
these tests are now failing in master. Can you take a look?

just-boris: I know.
This is because of Babel issue: https://github.com/babel/babel/pull/3608. (We are parsing stack traces, but they are broken now). You can ping that guys if you want faster resolution.

@christian-bromann  we can disable these test, while Babel feature is open. It affects only `babel-register`, so users will not notice anything

christian-bromann: sure

just-boris: Babel `v6.14.0` with the fix is out.

PR #1464, opened 2016-07-17
Fix wdio conf template

andriilazebnyi:
## Proposed changes

Generating wdio.conf.js using wdio cli produces broken wdio.conf.js if 2 (or more) services are selected.

As the result, **services** section looks like:

>     // Test runner services
    // Services take over a specific job you don't want to take care of. They enhance
    // your test setup with almost no effort. Unlike plugins, they don't add new
    // commands. Instead, they hook themselves up into the test process.
    services: ['firefox-profile&#39;,&#39;selenium-standalone'],

This happens because quotes are being escaped.

## Types of changes

## Checklist


christian-bromann:
Thanks!

PR #1542, opened 2016-08-17
Add files list to package.json for limiting data getting to npm

paazmaya:
## Proposed changes

Add files property to `package.json`, https://docs.npmjs.com/files/package.json#files, to limit the amount of data published to npm.

## Types of changes

## Checklist

- [ ] I have read the [CONTRIBUTING](/CONTRIBUTING.md) doc

## Further comments

Before adding the `files` property, `npm pack` created `tgz` of the size: 802 KB

After adding  the `files` property, `npm pack` created `tgz` of the size: 202 KB

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,14 @@
     "test-multibrowser": "grunt test:multibrowser",
     "test-snyk": "snyk test"
   },
+  "files": [
+    "bin",
+    "build",
+    "lib",
+    "index.js",
+    "package.json",
+    "README.md"
+  ],
   "dependencies": {
     "archiver": "1.0.0",
     "babel-runtime": "^6.9.2",
--- a/package.json
+++ b/package.json
@@ -35,8 +35,6 @@
   "files": [
     "bin",
     "build",
-    "lib",
-    "index.js",
     "package.json",
     "README.md"
   ],


paazmaya:
FYI, I came across this issue while running automated testing on another project, which failed due to:

```
warning: Could not stat path 'node_modules/webdriverio/test/site/platforms/android/CordovaLib/build/intermediates/classes/debug/org/apache/cordova/NativeToJsMessageQueue$OnlineEventsBridgeMode$OnlineEventsBridgeModeDelegate.class': Filename too long
```

package.json
```diff
@@ -32,6 +32,14 @@
     "test-multibrowser": "grunt test:multibrowser",
     "test-snyk": "snyk test"
   },
+  "files": [
+    "bin",
+    "build",
+    "lib",
+    "index.js",
```
christian-bromann: I believe we don't need `lib` and `index.js` too.

just-boris: including `lib` is nice because I beleive that ES6-code can be natively in future.

just-boris:
:tada: for that change.

I was a bit suprised when realized that webdriver.io package is 10Mb big

package.json
christian-bromann: However it includes a lot of not used files and would be only important if we would enable source maps

paazmaya: It might be then ok to remove `lib` since its contents is being build and available in `build` directory.

paazmaya: ...and since `main` property points to `./build/index.js`, it would be safe to remove `index.js` from the list too.

paazmaya:
After removing `lib` and `index.js` from the `files` list, the resulting npm package `tgz` file has the size of 11 KB.

christian-bromann:
Thanks so much!

paazmaya:
Please release it so you can have some 🍇

christian-bromann:
Your wish is my command. Now give me all your 🍇 🍇 🍇 🍇  and 🍻 🍻 🍻 🍻

(fyi I was wondering for a while why the npm publish command took so long, I blamed it on NPM but was totally wrong with that)

christian-bromann:
unpublished version due to #1544

paazmaya:
Oh no!

christian-bromann:
@paazmaya can you see what's wrong?

christian-bromann:
Apparently it didn't include all files inside the build directory:

screen shot 2016-08-17 at 18 33 39

paazmaya:
... looking at it

paazmaya:
Confirmed, for some reason `lib` does not end up in the package, while it is under `build` at the time of package creation..

paazmaya:
From https://docs.npmjs.com/files/package.json#files

> If you name a folder in the array, then it will also include the files inside that folder.

Perhaps the wording here reveals that folders are not included.

christian-bromann:
It suppose to behave the same as gitignore but it doesn't.

paazmaya:
Made no different to include `build/lib` to the `files`.

paazmaya:
Removing `.npmignore` made no difference.

Issue #1565, opened 2016-08-30
WDIO test runner unable to find cucumber step definitions


gsccheng:
When executing the WDIO test runner with 
```
$ ./node_modules/.bin/wdio wdio.conf.js
```
a new Chrome window (as I have configured) pops up briefly and I get:
```
ERROR: ENOENT: no such file or directory, lstat '/Users/me/path/to/app/features/step_definitions/**'
```
Meanwhile, I have:
1) Installed the Cucumber framework: `wdio-cucumber-framework`
2) Set `wdio.conf.js` file to `framework: 'cucumber'`
3) Am using the default wdio generated cucumberOpts, except for `require` where I used `step_definitions` instead of `step-definitions`: 
```
    cucumberOpts: {
        require: ['./features/step_definitions/**/*.js'],        // <string[]> (file/dir) require files before executing features
        backtrace: false,   // <boolean> show full backtrace for errors
        compiler: [],       // <string[]> ("extension:module") require files with the given EXTENSION after requiring MODULE (repeatable)
        dryRun: false,      // <boolean> invoke formatters without executing steps
        failFast: false,    // <boolean> abort the run on first failure
        format: ['pretty'], // <string[]> (type[:path]) specify the output format, optionally supply PATH to redirect formatter output (repeatable)
        colors: true,       // <boolean> disable colors in formatter output
        snippets: true,     // <boolean> hide step definition snippets for pending steps
        source: true,       // <boolean> hide source uris
        profile: [],        // <string[]> (name) specify the profile to use
        strict: false,      // <boolean> fail if there are any undefined or pending steps
        tags: [],           // <string[]> (expression) only execute the features or scenarios with tags matching the expression
        timeout: 20000,     // <number> timeout for step definitions
        ignoreUndefinedDefinitions: false, // <boolean> Enable this config to treat undefined definitions as warnings.
    },
```

If I were to run `$ cucumber.js` from the command line it would read my file: `features/step_definitions/auth/login.js`

Yet, the WDIO test runner doesn't appear to be able to do this, and from the error message, it appears the search path is truncated since I don't see a `/*.js`

What can I do here to get cucumber to read my step definitions using the WDIO test runner? Thanks.

gsccheng:
Thanks, it works now. I just realized the comments also say it expects a directory: `files/dir`

cheungkayip:
[...]

PR #1572, opened 2016-09-01
Fix: default location for cucumber step definitions

hungrydavid:
## Proposed changes

As mentioned in [#1565](https://github.com/webdriverio/webdriverio/issues/1565) the standard pattern for cucumber step definitions results in a not found error, even when they exist. By just specifying the directory rather than the globbing pattern the suite runs as expected.

## Types of changes

## Checklist

## Further comments

I couldn't find associated tests for this, but the entire suite passes with the change. So I'm guessing this part isn't covered? Please just let me know whether either that is fine as is, or where relevant tests should live. Thanks.

The generated './features/step_definitions/**/*.js' throws a  not found error, just the directory path is needed ie './features/step_definitions/'

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -200,7 +200,7 @@ WDIO Configuration Helper
         type: 'input',
         name: 'stepDefinitions',
         message: 'Where are your step definitions located?',
-        default: './features/step-definitions/**/*.js',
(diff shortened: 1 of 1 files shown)


christian-bromann:
👍  thanks!

Issue #1488, opened 2016-07-25
In lib/helpers/errorHandler.js tries access object like array for result of element command.


CTaylor1118:
## The problem

The staleElementError error handler tries access the result of the element command as if it were an array. Since the element command just returns an object, this causes an error and can't handle a staleElementError when element is used in a command instead of elements (say for the click command).

## Environment

* WebdriverIO version: 4.2.1
* Node.js version: 0.12.7
* [Standalone mode or wdio testrunner](http://webdriver.io/guide/getstarted/modes.html): Standalone

mnaughto:
I'm running into the same problem. If I add a few additional checks in that code, the stale element error just comes through. How soon might a pull request be accepted if I made one?

ebbe-brandstrup:
Same problem.

Indeed, the ```staleElementError``` is making the mistake of using the result of the WebdriverIO [ELEMENT](http://webdriver.io/api/protocol/element.html) action as an array instead of a JSON object (like the docs state it returns) - just like @CTaylor1118 said. 

I have to revert to an older version of WebdriverIO for the time being. I hope I'll have time to create a PR with a fix shortly though.

### Environment
- WebdriverIO version 4.2.16
- Node.js version 4.5.0
- Standalone mode

CTaylor1118:
I have a fix for this in my fork as well as fixes for a few other bugs. I had to include the build files as we are using node 0.12.7. Eventually I will get around to opening a pull request for the changes after I remove the build files.

ebbe-brandstrup:
I created a PR with a fix now and I included a test that verifies that the error handler doesn't crash when an 'element' command is encountered anymore.

@CTaylor1118, I'd really appreciate some feedback if you also made a test to verify the fix and found some better way to exercise the bug.

christian-bromann:
Thanks @ebbe-brandstrup for the PR and thanks @CTaylor1118 for reporting
https://github.com/webdriverio/webdriverio/pull/1606

PR #1606, opened 2016-09-22
Fixed crash in the 'stale element reference' error handler

ebbe-brandstrup:
Fix for issue #1488.

## Proposed changes

Fixed crash in the 'stale element reference' error handler (lib/helpers/errorHandler.js) that occurred when an 'element' command was in the command list.

The crash occurred because it was assuming that a result value for 'element' commands was always present and then that the result was an array - when in face a result for that type of command is always a WebElement JSON object.

Added test that exercises such a scenario.

**Note**: I am in doubt whether the test I added is a good way to exercise the bug. It was difficult to build up a command list that contained an 'element' (not 'element**s**') command in a natural way and have the error condition trigger in the error handler. Feedback or comments appreciated if this approach isn't good.

## Types of changes

## Checklist

### DIFF
--- a/lib/helpers/errorHandler.js
+++ b/lib/helpers/errorHandler.js
@@ -20,14 +20,20 @@ export default [
             if (command.name !== 'element' && command.name !== 'elements') {
                 continue
             }
-            if (command.name === 'element' && (!command.result[0].value || command.result[0].value.ELEMENT !== failingCommand.args[0])) {
+
+            if (command.name === 'element' && (!command.result || !command.result.value || command.result.value.ELEMENT !== failingCommand.args[0])) {
                 continue
             }
 
-            for (let result of command.result.value) {
-                if (result.ELEMENT === failingCommand.args[0]) {
-                    commandToRepeat = this.commandList[i - 1]
-                    break
+            if (command.result) {
+                // Ensure an array to simply the below logic
+                const results = Array.isArray(command.result.value) ? command.result.value : command.result.value !== null ? [command.result.value] : []
+
+                for (let result of results) {
+                    if (result.ELEMENT === failingCommand.args[0]) {
+                        commandToRepeat = this.commandList[i - 1]
+                        break
+                    }
                 }
             }
 
--- a/lib/helpers/errorHandler.js
+++ b/lib/helpers/errorHandler.js
@@ -17,23 +17,25 @@ export default [
         for (let i = this.commandList.length - 1; i >= 0; --i) {
             const command = this.commandList[i]
 
+            if (!command.result || !command.result.value) {
+                continue
+            }
+
             if (command.name !== 'element' && command.name !== 'elements') {
                 continue
             }
 
-            if (command.name === 'element' && (!command.result || !command.result.value || command.result.value.ELEMENT !== failingCommand.args[0])) {
+            if (command.name === 'element' && (command.result.value.ELEMENT !== failingCommand.args[0])) {
                 continue
             }
 
-            if (command.result) {
-                // Ensure an array to simply the below logic
-                const results = Array.isArray(command.result.value) ? command.result.value : command.result.value !== null ? [command.result.value] : []
+            // Ensure an array when evaluating the result, so the logic is the same for 'element' and 'elements' commands
+            const results = Array.isArray(command.result.value) ? command.result.value : command.result.value !== null ? [command.result.value] : []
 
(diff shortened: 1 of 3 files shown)


ebbe-brandstrup:
I was surprised to see the CI test run fail. The failing test looks odd to me, and running ```grunt test:wdio``` locally passes without issue.

Any way to re-trigger the build or point me to the problem?

christian-bromann:
@ebbe-brandstrup thanks for contributing. Can you check if you can unquarantine this test with the fix: https://github.com/webdriverio/webdriverio/blob/master/test/spec/functional/staleElementRetry.js#L7-L17

lib/helpers/errorHandler.js
```diff
@@ -20,14 +20,20 @@ export default [
...
                 continue
             }
 
-            for (let result of command.result.value) {
-                if (result.ELEMENT === failingCommand.args[0]) {
-                    commandToRepeat = this.commandList[i - 1]
-                    break
+            if (command.result) {
+                // Ensure an array to simply the below logic
+                const results = Array.isArray(command.result.value) ? command.result.value : command.result.value !== null ? [command.result.value] : []
```
CTaylor1118: @ebbe-brandstrup My solution here was a little different. Your solution is basically throwing out the possibility of retrying on any command that uses element rather than elements. My solution was not to ensure an array but simple check if we had an the result of an element command and if it was, check to see if that result was the elements we were looking for and set commandToRepeat there then break. You can see mine at https://github.com/CTaylor1118/webdriverio/blob/master/lib/helpers/errorHandler.js.

CTaylor1118:
There is another issue that could be breaking that test this is the incorrect context being passed into the error handler when in standalone mode.

lib/helpers/errorHandler.js
ebbe-brandstrup: Thanks for the feedback!

As far as I can tell, the logic here will still accept a match from an 'element' command. In the case of an 'element' command with a result, the ```results``` variable will be an array with just that one result in it. Basically, the array part just means I don't need an ```if``` for the 'element' case. The amount of simplification that accomplishes is very, very little though... I would be fine just using an ```if``` like you did too.

ebbe-brandstrup:
@CTaylor1118, the problem of an incorrect context being passed into the error handler in standalone mode... would that be something for a separate PR?

ebbe-brandstrup:
I have enabled the quarantined test @christian-bromann. It runs fine in this branch... but actually also in master.

I did a bit of cleanup in the stale element reference handler also. 

I still cannot reproduce the failing test locally (it passes everytime) and I'd need some assistance on how to approach debugging why it fails on your Travis CI.

Basically, I can't do much more with this PR currently.

CTaylor1118:
@ebbe-brandstrup The incorrect context being passed in probably needs a different pull request. There were some other issues as well, such as the command being retried hanging because the previous command was not finished (because it is waiting on the result of the command being retried) so it would just sit deadlocked. I implemented a solution for this but it was a quick fix and might not fit into how the maintainers here would want the problem solved.

christian-bromann:
@ebbe-brandstrup thanks! Will look into the fix asap

ebbe-brandstrup:
@CTaylor1118 Are your other fixes in a state where you could make a PR?

@christian-bromann Anything further I can do? The error message from the failing test makes no sense to me because as far as I can find in the webdriverio source, no place is appending anything to the error message that for some reason ends up with ``` (undefined)``` at the end of it - also, when run locally with clean npm install and git clone, everything is green.

```expected 'An element could not be located on the page using the given search parameters ("undefined").' to equal 'An element could not be located on the page using the given search parameters.'```

CTaylor1118:
@ebbe-brandstrup The fix I made works but I am not really satisfied by the way I solved the problem. Basically I had to place a boolean to indicate when there was a failure when retrying, otherwise it didn't bubble up correctly.

christian-bromann:
@ebbe-brandstrup nope, it's all good. The error message comes from previous (not released) changes. Will have to look into it but haven't had a chance yet

ebbe-brandstrup:
@CTaylor1118 Maybe just make the PR and @christian-bromann or other contributors might have a good suggestion or find it a good enough solution? I would at least appreciate the fixes you made a lot :)

christian-bromann:
Thanks again, will fix the failing test and release this fix asap

Issue #1559, opened 2016-08-23
Uncaught RuntimeError: Underlying driver does not implement advanced user interactions yet.


kdaimiel:
I need to test my mobile app in different browsers and I am trying to use touchFlick protocol in a non mobile device then I get this error.

## Environment

* WebdriverIO version: 4.2.7
* ChromeDriver version: 2.22.397932

christian-bromann:
What error?

kdaimiel:
Here is the error:

     Uncaught RuntimeError: Underlying driver does not implement advanced user interactions yet.
      at elementIdDisplayed("0") - isVisible.js:71:55

christian-bromann:
touchFlick doesn't work on desktop, mobile only!

kdaimiel:
That's not truth, I have a laptop with touchable screen and as I see it ChromeDriver supports touchFlick, so why do you throw the following exception with an action like swipe which implement touchFlick protocol?

 Uncaught CommandError: swipe command is not supported on non mobile platforms

christian-bromann:
> ChromeDriver supports touchFlick

Where is this documented?

> why do you throw the following exception

Because I never thought that a driver would support it for desktop.

kdaimiel:
Hi, I could see it on https://sites.google.com/a/chromium.org/chromedriver/downloads. It seems to be an experimental feature from ChromeDriver 2.16

ChromeDriver 2.16

Changes:
Experimental support for Touch Actions
Fixed a bug when navigating back or forwards, where ChromeDriver sometimes would not wait for the page to finish loading
Fixed a bug when navigating or reloading a page, where ChromeDriver would sometimes send commands to the wrong <iframe>

christian-bromann:
> Experimental support for Touch Actions

I am not sure that this is meant for desktop. You can run Chrome on Android or iOS too where this function might be useful. Try to comment out the if statement in the touchflick command and see if chromedriver supports it.

kdaimiel:
Yes, I have commented the if statement and doesn't work with selenium-standanlone server, it seems that they don't support this kind of interactions yet, but It works in SauceLabs and with a selenium parched by me. Thanks

PR #1633, opened 2016-10-05
Allowing touch actions on Chrome

kdaimiel:
This new feature is related to #1559

## Proposed changes

It seems that ChromeDriver supports touch actions so I have added this functionality to be able to run this kind of actions in a browser with Selenium. This has come because I had to test hybrid apps with a touch screen laptop.

## Types of changes

## Checklist

## Further comments

I have exposed the same problem to Selenium and they have already accepted. It will be able in next releases.
https://github.com/SeleniumHQ/selenium/pull/2694

### DIFF
--- a/lib/commands/hold.js
+++ b/lib/commands/hold.js
@@ -10,16 +10,9 @@
  *
  */
 
-import { CommandError, RuntimeError } from '../utils/ErrorHandler'
+import { RuntimeError } from '../utils/ErrorHandler'
 
 let hold = function (selector) {
-    /*!
-     * compatibility check
-     */
-    if (!this.isMobile && !this.isChrome) {
-        throw new CommandError('hold command is only supported for mobile platforms')
-    }
-
     return this.element(selector).then((res) => {
         /**
          * check if element was found and throw error if not
--- a/lib/commands/hold.js
+++ b/lib/commands/hold.js
@@ -16,7 +16,7 @@ let hold = function (selector) {
     /*!
      * compatibility check
      */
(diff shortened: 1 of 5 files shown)


christian-bromann:
I don't think it makes sense to add additional `isChrome` flags. I would rather then just remove all if statements that prevent capabilities from using touch actions

kdaimiel:
Yes, it has more sense to me but I did't want to affect the checking of other browser. I have already removed the statements.

christian-bromann:
👍

psylence303:
Hi, am I getting it wrong or right now I can use it only if I use selenium jar thing directly? If I use 'wdio-selenium-standalone-service' package then I still getting 'Underlying driver does not implement advanced user interactions yet.' when trying to call touchFlick.

christian-bromann:
We haven't published the new version yet.

psylence303:
@christian-bromann any plans on releasing this? as far as I see, wasn't included in the 4.3.0 :(

christian-bromann:
@psylence303 it was included 😉

kdaimiel:
@psylence303 as I see, this is included in webdriverio, but 'wdio-selenium-standalone-service' uses selenium-standalone npm package and this is still using selenium-sever v2.53.1 as a default version. This is solved in selenium-server 3.0 so we have to wait for next releases until they update to Selenium3 or you can change config manually.

psylence303:
@christian-bromann @kdaimiel thank you for the information!

PR #1681, opened 2016-11-02
screenshotOnReject: Ability to specify connection retry options

DudaGod:
## Proposed changes

This PR adds ability to specify `connectionRetryTimeout` and `connectionRetryCount` on request to take screenshot after test is failed. This changes is intended to reduce waiting time. 
When test failed and Selenium trying to take screenshot quite often ie11 waiting for a response for 90000ms * 3 = 4.5 minutes (at worst). It's too long.

## Types of changes

## Checklist

### DIFF
--- a/docs/guide/getstarted/configuration.md
+++ b/docs/guide/getstarted/configuration.md
@@ -122,12 +122,26 @@ Type: `String`|`null`<br>
 Default: *null*
 
 ### screenshotOnReject
-Attaches a screenshot of a current page to the error if the Selenium driver crashes
+Attaches a screenshot of a current page to the error if the Selenium driver crashes.
+Can be specified as object to set the timeout and count of retries on the attempt to take screenshot.
 
-Type: `Boolean`<br>
+Type: `Boolean`|`Object`<br>
 Default: *false*
 
 **Note**: Attaching screenshot to the error uses extra time to get screenshot and extra memory to store it. So for the sake of performance it is disabled by default.
+
+**Example:**
+
+```js
+// take screenshot on reject
+screenshotOnReject: true
+
+// take screenshot on reject and set some options
+screenshotOnReject: {
+    connectionRetryTimeout: 30000,
+    connectionRetryCount: 0
+}
+```
 
 ### waitforTimeout
 Default timeout for all waitForXXX commands.
--- a/lib/webdriverio.js
+++ b/lib/webdriverio.js
@@ -212,6 +212,10 @@ let WebdriverIO = function (args, modifier) {
         const client = unit()
         const failDate = new Date()
 
+        if (typeof options.screenshotOnReject === 'object') {
+            client.requestHandler = createRequestHandlerDecorator(options.screenshotOnReject)
+        }
+
         // don't take a new screenshot if we already got one from Selenium
         const getScreenshot = typeof err.screenshot === 'string'
             ? () => err.screenshot
@@ -228,6 +232,14 @@ let WebdriverIO = function (args, modifier) {
                     client.emit('screenshot', { data: screenshot, filename })
                 }
             })
+    }
+
+    function createRequestHandlerDecorator (opts) {
+        return Object.create(requestHandler, {
+            defaultOptions: {
+                value: Object.assign({}, requestHandler.defaultOptions, opts)
+            }
+        })
     }
 
     function saveScreenshotSync (screenshot, date) {
--- a/test/spec/unit/remote.js
+++ b/test/spec/unit/remote.js
@@ -157,5 +157,66 @@ describe('remote method', () => {
 
             return assert.isRejected(client.getUrl(), /some-error/)
         })
+
+        describe('connection retry options', () => {
+            var mkRequest_ = (client) => {
+                var requestOpts
+
+                RequestHandler.prototype.create.restore()
+                sandbox.stub(RequestHandler.prototype, 'create', function () {
+                    requestOpts = this.defaultOptions
+                })
+
+                return client.getUrl()
+                    .catch(() => requestOpts)
+            }
+
+            it('should take screenshot with default connection retry options', () => {
+                var client = remote({
+                    connectionRetryTimeout: 100500,
+                    connectionRetryCount: 10,
+                    screenshotOnReject: true
+                })
+
+                return mkRequest_(client)
+                    .then((requestOpts) => {
+                        assert.equal(requestOpts.connectionRetryTimeout, 100500)
+                        assert.equal(requestOpts.connectionRetryCount, 10)
+                    })
+            })
+
+            it('should take screenshot with specified connection retry options', () => {
+                var client = remote({
+                    connectionRetryTimeout: 100500,
+                    connectionRetryCount: 10,
+                    screenshotOnReject: {
+                        connectionRetryTimeout: 5000,
+                        connectionRetryCount: 0
+                    }
+                })
+
+                return mkRequest_(client)
+                    .then((requestOpts) => {
+                        assert.equal(requestOpts.connectionRetryTimeout, 5000)
+                        assert.equal(requestOpts.connectionRetryCount, 0)
+                    })
+            })
+
+            it('should use default connection retry option for not specified options', () => {
+                var client = remote({
+                    connectionRetryTimeout: 100500,
+                    connectionRetryCount: 10,
+                    screenshotOnReject: {
+                        connectionRetryCount: 0
+                    }
+                })
+
+                return mkRequest_(client)
+                    .then((requestOpts) => {
+                        assert.equal(requestOpts.connectionRetryTimeout, 100500)
+                        assert.equal(requestOpts.connectionRetryCount, 0)
+                    })
+            })
+        })
     })
 })


tormozz48:
🆗

docs/guide/getstarted/configuration.md
```diff
@@ -110,13 +110,21 @@ Type: `String`|`null`<br>
...
+Type: `Boolean`|`Object`<br>
 Default: *false*
 
 **Note**: Attaching screenshot to the error uses extra time to get screenshot and extra memory to store it. So for the sake of performance it is disabled by default.
 
+**Example:**
+
+```js
+connectionRetryTimeout: 30000 // timeout on the connection to the Selenium server
+connectionRetryCount: 0       // count of retries to connect to the Selenium server
```
j0tunn: You probably should add options descriptions in doc instead of explaining them in example. If you want to add an example, it should include `screenshotOnReject` option:
```js
// take screenshot on reject
screenshotOnReject: true;

// take screenshot on reject and set some options
screenshotOnReject: {
    connectionRetryTimeout: 30000, // timeout on the connection to the Selenium server
    connectionRetryCount: 0       // count of retries to connect to the Selenium server
}
```

lib/utils/RequestHandler.js
```diff
@@ -214,7 +214,7 @@ class RequestHandler {
                     return reject(new RuntimeError(error))
                 }
 
-                if (retryCount === totalRetryCount) {
+                if (retryCount >= totalRetryCount) {
```
j0tunn: are there any tests for this case. This looks like bugfix in fact

lib/utils/RequestHandler.js
```diff
@@ -238,6 +238,14 @@ class RequestHandler {
             })
         })
     }
+
+    extendDefaultOptions (opts) {
```
j0tunn: `updateOpts` or `setOpts` maybe

lib/utils/RequestHandler.js
```diff
@@ -238,6 +238,14 @@ class RequestHandler {
...
+        const shouldExtend = Object.keys(opts).some((key) => opts[key] !== this.defaultOptions[key])
```
j0tunn: why do we need to check this stuff? Why not just update `this.defatultOptions` unconditionally?

lib/webdriverio.js
```diff
@@ -64,7 +64,7 @@ let WebdriverIO = function (args, modifier) {
     }
 
     let logger = new Logger(options, eventHandler)
-    let requestHandler = new RequestHandler(options, eventHandler, logger)
+    let requestHandler = new RequestHandler(Object.assign({}, options), eventHandler, logger)
```
j0tunn: You should clone `options` inside `RequestHandler` constructor

lib/utils/RequestHandler.js
```diff
@@ -238,6 +238,14 @@ class RequestHandler {
...
+
+        if (shouldExtend) {
+            this.defaultOptions = Object.assign(this.defaultOptions, opts)
```
j0tunn: In fact `defaultOptions` are not *default* for the RequestHandler, they are just `options`. I think you should rename it

lib/webdriverio.js
j0tunn: One more think: right now `RequestHandler` modifies passed options. There can be some side effects after you start cloning them. You should check this

lib/webdriverio.js
```diff
@@ -208,14 +208,34 @@ let WebdriverIO = function (args, modifier) {
...
 
         // don't take a new screenshot if we already got one from Selenium
         const getScreenshot = typeof err.screenshot === 'string'
             ? () => err.screenshot
-            : () => rawCommands.screenshot.call(client).then((res) => res.value)
+            : () => {
+                setConnectionRetryOpts(client, options.screenshotOnReject)
+
+                return rawCommands.screenshot.call(client).then((res) => {
+                    setConnectionRetryOpts(client, {})
```
j0tunn: on `screenshot` command fail this wouldn't happen

lib/webdriverio.js
DudaGod: I've checked, no side effects will be

test/spec/unit/remote.js
```diff
@@ -157,5 +158,76 @@ describe('remote method', () => {
...
+                connectionRetryCount: 0
+            }
+
+            RequestHandler.prototype.extendDefaultOptions.restore()
+
+            var client = remote({screenshotOnReject})
+
+            return client.getUrl()
+                .catch(() => {
+                    console.log(client.options)
```
DudaGod: delete this

test/spec/unit/remote.js
```diff
@@ -157,5 +158,76 @@ describe('remote method', () => {
...
+                connectionRetryTimeout: 30000,
+                connectionRetryCount: 0
+            }
+            var client = remote({screenshotOnReject})
+
+            return client.getUrl()
+                .catch(() => assert.calledWith(RequestHandler.prototype.extendDefaultOptions, screenshotOnReject))
+        })
+
+        it('setting connection options should not affect on browser options', () => {
```
j0tunn: `should not affect on` -> `should not modify`

lib/webdriverio.js
j0tunn: also there should be a test for this case

lib/utils/RequestHandler.js
j0tunn: lets split this PR:
- bugfix for retryCount
- new feature for screenshotOnReject

lib/utils/RequestHandler.js
```diff
@@ -16,7 +16,10 @@ class RequestHandler {
         this.gridApiStartPath = '/grid/api'
         this.eventHandler = eventHandler
         this.logger = logger
-        this.defaultOptions = options
+        this.options = options
```
j0tunn: `this.options` are shared between `webdriverio` instance and `requestHandler` so I suggest to do not touch `RequestHandler` at all. Lets set connection retry options in `webdriverio` instance.

lib/webdriverio.js
```diff
@@ -208,14 +208,33 @@ let WebdriverIO = function (args, modifier) {
...
     function takeScreenshot (err) {
         const client = unit()
         const failDate = new Date()
 
         // don't take a new screenshot if we already got one from Selenium
         const getScreenshot = typeof err.screenshot === 'string'
             ? () => err.screenshot
-            : () => rawCommands.screenshot.call(client).then((res) => res.value)
+            : () => {
+                setConnectionRetryOpts(client, options.screenshotOnReject)
```
j0tunn: Lets return previous connection retry options values from `setConnectionRetryOpts`, store them inside some variable and then set them back. Like
```js
const prevRetryOpts = setConnectionRetryOpts(client, options.screenshotOnReject)
...
    .finally(() => setConnectionRetryOpts(client, prevRetryOpts))
```

lib/utils/RequestHandler.js
j0tunn: of course there should be a test for this behaviour

test/spec/unit/remote.js
```diff
@@ -157,5 +158,104 @@ describe('remote method', () => {
...
+                        connectionRetryCount: 0
+                    }
+                })
+
+                var defaultConnectionRetryTimeout = client.options.connectionRetryTimeout
+                var defaultConnectionRetryCount = client.options.connectionRetryCount
+
+                var takeScreenshot = RequestHandler.prototype.create.withArgs('/session/:sessionId/screenshot')
+
+                if (message === 'failed') {
```
j0tunn: thats a bad practice to join tests and then add invariants in them. You should split them into separate test cases. There will be some duplication, but thats acceptable for tests and also you can use some helper to decrease duplication

test/spec/unit/remote.js
```diff
@@ -157,5 +158,104 @@ describe('remote method', () => {
...
+            it(`should reset connection retry options to default if takeScreenshot is ${message}`, () => {
+                var client = remote({
+                    screenshotOnReject: {
+                        connectionRetryTimeout: 30000,
+                        connectionRetryCount: 0
+                    }
+                })
+
+                var defaultConnectionRetryTimeout = client.options.connectionRetryTimeout
+                var defaultConnectionRetryCount = client.options.connectionRetryCount
```
j0tunn: In fact we can't be sure that default values are not the 3000 and 0. Lets set those options to some values and then check that values

test/spec/unit/remote.js
```diff
@@ -157,5 +158,104 @@ describe('remote method', () => {
...
+                return client.getUrl()
+                    .catch(() => {
+                        var {connectionRetryTimeout, connectionRetryCount} = client.requestHandler.connectionOpts
+                        assert.strictEqual(connectionRetryTimeout, defaultConnectionRetryTimeout)
+                        assert.strictEqual(connectionRetryCount, defaultConnectionRetryCount)
+                    })
+            })
+        })
+
+        it('should set specified connection opts before takeScreenshot and reset to default after', () => {
```
j0tunn: I think this test case is redundant

lib/protocol/screenshot.js
```diff
@@ -10,8 +10,12 @@
  *
  */
 
-let screenshot = function () {
-    return this.requestHandler.create('/session/:sessionId/screenshot')
+let screenshot = function (opts) {
```
DudaGod: Yeah, sorry it wasn't a good solution.

I've changed the code, can you look again please?

lib/webdriverio.js
```diff
@@ -230,6 +234,14 @@ let WebdriverIO = function (args, modifier) {
             })
     }
 
+    function mkRequestHanlderDecorator (opts) {
```
christian-bromann: What does this decorator suppose to do here?

j0tunn: It creates `RequestHandler` instance with updated `defaultOptions` leaving original `requestHandler` unchanged.

DudaGod: I don't understand, what you trying to say that - "Also there is a wording the name"?
Rename "mk" - to "make" or "create" ?

christian-bromann: No `mkRequestHandlerDecorator` rather `mkRequestHanlderDecorator` and yes `makeRequestHandlerDecorator` would be even better.

lib/webdriverio.js
```diff
@@ -230,6 +234,14 @@ let WebdriverIO = function (args, modifier) {
             })
     }
 
+    function createRequestHanlderDecorator (opts) {
```
christian-bromann: I am concerned about this solution because once you take a screenshot the requestHandler is decorated and will used for the next requests even though they aren't screenshot requests

christian-bromann:
@DudaGod how about you check the url path in the request handler and skip the retry mechanism if the command is a screenshot command. I would even say we only need to retry for init commands so when the session is created.

j0tunn:
> how about you check the url path in the request handler and skip the retry mechanism if the command is a screenshot command. I would even say we only need to retry for init commands so when the session is created.

That would be a breaking change while current solution is a tiny enhancement. It gives us flexible configuration.
Besides not only init commands should be waited and retried. Screenshot command is a heavy operation and it can be called not only on fail. Execute some js can be long operation. Some POST requests, etc.

lib/webdriverio.js
DudaGod: No, if you look more attentively you will see that it is decorate only screenshot request, because in takeScreenshot method created a new client instance.
Moreover I've checked it a lot of times and I can say with confidence that it decorates only screenshot request  on fail.

christian-bromann:
@DudaGod so you patch the request handler when a screenshot is taken but doesn't this also patch all upcoming requests?

DudaGod:
@christian-bromann no, it doesn't patch all upcoming requests, because in `takeScreenshot` function creates new webdriverIO instance. You can see it on this line - https://github.com/webdriverio/webdriverio/pull/1681/files#diff-3b660a335cd0d5e1c680f02c2a2a8c70L212 (line - 212).

So it affects only on screenshot requests.

j0tunn:
I'll explain a bit more detailed:
```js
// inside takeScreenshot()
const client = unit() // create new client

// inside unit()
let client = Object.create(prototype)
...
client.requestHandler = requestHandler // yes, here requestHandler is shared
...
return client

// inside takeScreenshot()
// shared requestHandler replaced with new one
client.requestHandler = createRequestHandlerDecorator(options.screenshotOnReject)

// inside createRequestHandlerDecorator
return Object.create(requestHandler, {
    defaultOptions: {
        value: Object.assign({}, requestHandler.defaultOptions, opts) // original requestHandler is not modified
    }
})

// inside takeScreenshot()
// call screenshot command on new client instance
rawCommands.screenshot.call(client).then((res) => res.value)
```

j0tunn:
LGTM

eGavr:
@christian-bromann 🆙

christian-bromann:
Thanks @DudaGod

DudaGod:
@christian-bromann  Can you tell please when should we expect new version?

eGavr:
@christian-bromann 

+1

DudaGod:
@christian-bromann 
+2

christian-bromann:
I am on vacation and will release a new version next week

DudaGod:
@christian-bromann Can you tell when you will release a new version?

DudaGod:
@christian-bromann sorry for disturb you again, but we really need this new version...

christian-bromann:
@DudaGod sorry but there is some stuff (like #1717) I want to get in first before releasing the next version

PR #1682, opened 2016-11-03
Throw on service launcher error

silvenon:
I had a service launcher which was silently failing. Now it properly fails, like the service itself does. This means every service will need to include a `launcher.js` with at least an empty default export, I hope that's ok.

This change helped me solve the [wdio-static-server-service problem](https://github.com/LeadPages/wdio-static-server-service/pull/1).

### DIFF
--- a/lib/launcher.js
+++ b/lib/launcher.js
@@ -439,7 +439,11 @@ End selenium sessions properly ...
 
             try {
                 service = require(`wdio-${serviceName}-service/launcher`)
-            } catch (e) {}
+            } catch (e) {
+                if (!e.message.match(`Cannot find module 'wdio-${serviceName}-service/launcher'`)) {
+                    throw new Error(`Couldn't initialise launcher from service "${serviceName}".\n${e.stack}`)
+                }
+            }
 
             if (service && (typeof service.onPrepare === 'function' || typeof service.onPrepare === 'function')) {
                 launchServices.push(service)
--- a/package.json
+++ b/package.json
@@ -92,6 +92,7 @@
     "istanbul": "^0.3.13",
     "load-grunt-tasks": "^3.2.0",
     "mocha": "^2.3.4",
+    "mock-require": "^2.0.0",
     "nock": "8.0.0",
     "saucelabs": "^1.0.1",
     "sinon": "^1.17.2",
--- a/test/fixtures/services/wdio-launcher-failure-service/index.js
+++ b/test/fixtures/services/wdio-launcher-failure-service/index.js
@@ -1,0 +1,1 @@
+module.exports = {}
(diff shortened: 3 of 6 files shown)


lib/runner.js
```diff
@@ -331,7 +331,7 @@ class Runner {
             try {
                 service = require(`wdio-${serviceName}-service`)
             } catch (e) {
-                if (!e.message.match(`Cannot find module '${serviceName}'`)) {
+                if (!e.message.match(`Cannot find module '${serviceName}'`).length) {
```
christian-bromann: Why length? If the message doesn't match it returns null, try `"foo".match(/bar/)`

silvenon: Oops, I thought it returns an empty array 😅 fixing…

silvenon:
Ok, I removed that commit.

christian-bromann:
@silvenon I believe the reason why I left an empty catch there was that some services don't have launcher and in this case this require should be totally ignored

silvenon:
Now WebdriverIO will throw only if the launcher exists, but fails.

christian-bromann:
> Now WebdriverIO will throw only if the launcher exists, but fails.

What I meant was that all services have a launcher.js in their root directory. But given a service would not need a launcher this file would be missing and the error thrown. I wanted to have the launcher optional because not all services need to hook into the onPrepare/onComplete hook.

silvenon:
@christian-bromann I understand, that's why I modified the code to allow services not to have a launcher. But if they do have a launcher, which fails, an error is thrown. Is that the desired behavior?

christian-bromann:
> Is that the desired behavior?

Yes. I guess we need to modify the if statement and only throw an error if the message is different than "module not found"

silvenon:
I thought that's what the code was doing? "In case the error is **not** thrown because the launcher wasn't found, display the error stack."

christian-bromann:
Oh right. I haven't seen that.
Since this is a critical change it would be great to add a unit test for it, what do you think?

silvenon:
I'm not sure how to test for this. The tests I added now are failing, I'm not sure why `process.env.NODE_PATH` doesn't behave as I expected, i.e. why it doesn't set `test/fixtures/services` as the load path for modules. Both tests can't find the `launcher.js` for those services, so I'm positive that's where they fail.

Any suggestions?

christian-bromann:
I don't think you can modify the NODE_PATH in runtime. Might be a good idea to use https://github.com/boblauer/mock-require

silvenon:
mock-require is cool, but it has a bug where it executes the module before the actual `require` happens, so it's throwing too early. The fix boblauer/mock-require#10 is not yet merged, so I installed `mock-require` from GitHub, which isn't ideal.

I commented there to send a notification to the author, so this PR should wait a couple of days more, maybe the fix will be merged.

silvenon:
Yay, the author of mock-require published the fix! Now this PR is ready. 🎉

silvenon:
I'm not sure if this test failure is my fault. Could you re-run it?

PR #1703, opened 2016-11-12
fixed misleading names in waitForXYZ commands

lukas-linhart:
## Proposed changes

Only a documentation change, fixing some misleading function and file names in most of the waitForXYZ commands. Should not break anything :)


christian-bromann:
Oh dang .. good catch, thanks!

PR #1706, opened 2016-11-14
Command/configure http

eGavr:
## Proposed changes

This pull request adds new command `configureHttp`. This feature is useful when you need to change http request options in runtime. For example, I set options `connectionRetryCount` and `connectionRetryTimeout` when I create instance of `wdio`:

```js
const client = webdriverio.remote({
    // ... lots of other options
    connectionRetryCount: 3,
    connectionRetryTimeout: 90000
})

client
    .init()
    .url('https://google.com')
    .end()
```

According to the code above, for all requests (`init`, `url`, `end`) http timeout will be 90000 ms and the amount of retries will be 3, but in tests there is often no need to wait for some requests for 90000 ms with 3 retries, for example, for closing of a session (`end` method), so it seems to be great to have the ability to configure http request options in runtime:

```js
const client = webdriverio.remote({
    // ... lots of other options
    connectionRetryCount: 3,
    connectionRetryTimeout: 90000
})

client
    .init()
    .url('https://google.com')
    .configureHttp({ connectionRetryCount: 1, connectionRetryTimeout: 5000 })
    .end() // timeout on closing of a session will be 5000 ms with 1 retry
```

## Types of changes

## Checklist

### DIFF
--- a/lib/commands/configureHttp.js
+++ b/lib/commands/configureHttp.js
@@ -1,0 +1,25 @@
+/**
+ * Set http request options
+ *
+ * <example>
+    :configureHttp.js
+    client
+        .init()
+        .url('http://google.com')
+        // ... other commands
+        .configureHttp({ // configuring of next requests
+            connectionRetryCount: 1,
+            connectionRetryTimeout: 5000
+        })
+        .end() // timeout on closing of a session will be 5000 ms with 1 retry
+ * </example>
+ *
+ * @param {Object} opts request options (`{protocol: string, hostname: string, port: string, connectionRetryTimeout: number, connectionRetryCount: number}`)
+ * @type utility
+ */
+
+let configureHttp = function (opts) {
+    Object.assign(this.requestHandler.defaultOptions, opts)
+}
+
+export default configureHttp
--- a/test/spec/unit/configureHttp.js
+++ b/test/spec/unit/configureHttp.js
@@ -1,0 +1,61 @@
+import { remote } from '../../../index.js'
+import RequestHandler from '../../../lib/utils/RequestHandler'
+import q from 'q'
+
+describe('configureHttp command', () => {
+    const sandbox = sinon.sandbox.create()
+
+    const stubRequestResult = () => q({body: {sessionId: '', value: ''}})
+
+    const assertRequestUri = (key, value) => {
+        const requestOpts = RequestHandler.prototype.request.lastCall.args[0]
+
+        assert.equal(requestOpts.uri[key], value)
+    }
+
+    beforeEach(() => sandbox.stub(RequestHandler.prototype, 'request').returns(stubRequestResult()))
+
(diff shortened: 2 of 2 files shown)


christian-bromann:
> but in tests there is often no need to wait for some requests for 90000 ms with 3 retries, for example, for closing of a session (end method)

I agree. However WebdriverIO won't wait when the response came in. So I don't see a need for that command.

eGavr:
But according to the code it will do https://github.com/webdriverio/webdriverio/blob/master/lib/protocol/session.js#L80-L90

Besides, it is not the only example of the usage of this method.

For an instance, some browser sessions start much faster than another (chrome starts faster than ie), so using `configureHttp` method we can set session request timeout for different browsers.

Consider we have `connectionRetryTimeout` 2 minutes. 
As it usually goes, if chrome does not start in 1 minute it will never start (something went wrong in selenuim) and it is better to retry than to waste time and wait one more minute, but if ie (it is really slow browser ) does not start in 1 minute it is better to keep waiting (it usually starts in 1 or 2 minutes). According to this information, I'd like to set session request timeout for chrome – 1 minute, for ie – 2 minutes.

christian-bromann:
> For an instance, some browser sessions start much faster than another (chrome starts faster than ie), so using configureHttp method we can set session request timeout for different browsers.

Why would you want to do that? If you have a retry timeout set at 90s and Chrome starts in like 10 seconds, WebdriverIO will continue right after these 10s.

If your grid needs more than one minute to start a browser you probably should improve the grid response time instead of increase the timeout of the selenium driver.

Additionally this command does nothing but changing the options. All commands do interact somehow with the Selenium protocol. I think this should stay a configuration thing and nothing that should be changed interactively. We already have too many commands and adding a command to change a certain option would blow up the WebdriverIO API even more.

eGavr:
What about `end` method ? I did not find the logic which does not wait for response.

> If your grid needs more than one minute to start a browser you probably should improve the grid response time instead of increase the timeout of the selenium driver

Some browsers can not be improved by default :) You are probably lucky not to run tests in ie8 :)

It is really useful to set timeouts for session start and session end which differs with timeouts on other requests, it helps to run tests faster.

[wd](https://github.com/admc/wd/blob/master/doc/api.md) supports this method :)

j0tunn:
@eGavr why do you need this command in webdriverio core? It can be easily added by `addCommand` on session initialization.

christian-bromann:
@j0tunn can you elaborate how you would do that?

j0tunn:
```js
client.addCommand('configureHttp', function (opts) {
    Object.assign(this.requestHandler.defaultOptions, opts);
});
```

christian-bromann:
thanks @j0tunn 

@eGavr please see @j0tunn workaround. I think doing this is better as adding another (non action / config only) command.

eGavr:
@j0tunn It is already in our tool :)

PR #1714, opened 2016-11-17
Improved doc on value property

wvankuipers:
## Proposed changes

Documentation update

## Types of changes

- [ ] Documentation

## Checklist

## Further comments

None

The `value` property is not optional an can also be an array. Fixed this in the documentation.


christian-bromann:
Yup, good catch 👍

PR #1724, opened 2016-11-22
Wait for visible within viewport

wvankuipers:
## Proposed changes

I added a command to wait for an element to be visible within the viewport

## Types of changes

## Checklist

## Further comments

I created this command so I don't have to "fake" it by doing a timeout before checking if a element exists within the current viewpoint. This new command is useful for checking "sticky" menu's or "popover" elements on a page that appear after a certain timeout.

### DIFF
--- a/lib/commands/waitForVisibleWithinViewport.js
+++ b/lib/commands/waitForVisibleWithinViewport.js
@@ -1,0 +1,103 @@
+/**
+ *
+ * Wait for an element (selected by css selector) for the provided amount of
+ * milliseconds to be (in)visible within the viewport. If multiple elements get
+ * queryied by given selector, it returns true (or false if reverse flag is set)
+ * if at least one element is visible.
+ *
+ * <example>
+    :index.html
+    <div id="elem" style="position:absolute; top: -100px; height: 50px;">
+        Hello World!
+    </div>
+    <script type="text/javascript">
+        setTimeout(function () {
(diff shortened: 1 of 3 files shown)


christian-bromann:
Hey @wvankuipers ,
I am not sure about this addition. `waitForXXX` commands should only cover most used scenarios. I already declined waitForAlert PRs because you usually don't use that command that often and it is easy replaceable with

```js
browser.waitUntil(() => browser.isVisibleWithinViewport("..."))
```

I don't want to introduce new commands that aren't really necessary. Our API is blown up already. What do you think?

wvankuipers:
This is true, you can use the above to wait for an element to be visible within the viewport, but how do we wait for a element to leave the viewport?

    browser.waitUntil(() => (!browser.isVisibleWithinViewport(btnSelector)));

Results in an error:
    
    { [Error: An element could not be located on the page using the given search parameters (".elem").]
  type: 'NoSuchElement',
  message: 'An element could not be located on the page using the given search parameters (".elem").'

christian-bromann:
> but how do we wait for a element to leave the viewport?

just do

```js
browser.waitUntil(() => !browser.isVisibleWithinViewport("..."))
```

The error just says that it can't find the element (".elem")

wvankuipers:
Hmm it seems to be a issue with my test script 😒 

Lets close this PR as I agree that it can be replaced by the command you provided!

PR #1729, opened 2016-11-24
Add reporter sanitization.

### LINKED ISSUE #1728: --reporters command line argument does not work
## The problem

The --reporter command line argument does not work.  It always defaults to the dot reporter.  This is because the launcher's initReporters() function requires config.reporters to be an array, and when coming from the CLI interface it is a string.

scottfalkingham:
## Proposed changes

Add the missing reporter sanitization which prevented the --reporters command line argument from working.  Fixes #1728.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -312,6 +312,13 @@ $ wdio wdio.conf.js
 }
 
 /**
+ * sanitize reporters
+ */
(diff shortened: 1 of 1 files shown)


christian-bromann:
Awesome thanks!

Issue #1591, opened 2016-09-14
Intermittent Reporting of Failure in Mocha suiteSetup


fijijavis:
## The problem

When a failure occurs in a Mocha suiteSetup intermittently the failure is not bubbling up properly.  
## Environment
- WebdriverIO version: 4.2.11
- Node.js version: 4.4.7
- [Standalone mode or wdio testrunner](http://webdriver.io/guide/getstarted/modes.html): wdio
- if wdio testrunner, running synchronous or asynchronous tests: sync
- Additional wdio packages used (if applicable): package.json is attached
## Details

When a Mocha suiteSetup fails I am seeing intermittent behavior with reporting of the failure.  In some instances all I see the full results and others all I am seeing is part of the results.

Example Full Result:

```
0 passing (12.00s)                                                                                                                                    
1 failing                                                                                                                                             

1) sample suite "before all" hook:                                                                                                                    
element (#thisshouldfail) still not visible after 5000ms                                                                                              
running chrome on ANY                                                                                                                                 
Error: element (#thisshouldfail) still not visible after 5000ms                                                                                       
    at Context.<anonymous> (tests/sample.test.js:13:17)                                                                                               
    at elements("#thisshouldfail") - isVisible.js:54:17                                                                                               
    at isVisible("#thisshouldfail") - waitForVisible.js:37:22                    `

Example Partial Result
`0 passing (12.00s)`
```

I'm thinking this is an issue with waiting for screen prints to be captured. When the screenshotPath in my conf file is a "real" path the percentage of time when I don't get a full report is over 90%.  However if I set the screenshotPath to something like "/dev/null" the percentage of time when I don't get a full report drops to like 30%.
## Link to Selenium/WebdriverIO logs

Create a [gist](https://gist.github.com) which is a paste of your _full_ Selenium/WebdriverIO logs, and link them here. Do _not_ paste your full Appium logs here, as it will make this issue very long and hard to read! If you are reporting a bug, _always_ include Selenium/WebdriverIO logs!
## Code To Reproduce Issue [ Good To Have ]

Attached package.json, conf.js and spec.js

[Archive.zip](https://github.com/webdriverio/webdriverio/files/472219/Archive.zip)

msegado:
I can reproduce this in a regular test too, not just in a suiteSetup (formula seems to be timeouts + errorShots = no reporting; can you change the issue title to reflect?). Here are my test file and config:

```js
/* file: test.js */
[...]

PR #1730, opened 2016-11-24
Fixed 1591, when a screenshot is created the results are now displayed

wvankuipers:
## Proposed changes

Fixed issue #1591 If a screenshot is generated for a failed test the results are now being displayed. (as expected)

## Types of changes

## Checklist

## Further comments

Small fix, the fail handle prevented the error from bubbling up to the point of displaying the results.

### DIFF
--- a/lib/webdriverio.js
+++ b/lib/webdriverio.js
@@ -200,7 +200,7 @@ let WebdriverIO = function (args, modifier) {
         err.shotTaken = true
         return takeScreenshot(err)
             .catch((e) => logger.log('\tFailed to take screenshot on reject:', e))
-            .then(() => fail(err, onRejected))
+            .then(() => { throw err })
     }
 
     function insideCommand (command, unit) {


christian-bromann:
@wvankuipers thanks for jumping in on this! Can you provide a reproducible example, so I can check out the fix?

wvankuipers:
I can re-create it using the cucumber-boilerplate project with the current feature test:

`Feature: Test
    Scenario: Screenshot vs Error
       Given I open the url "https://google.com"
       Then  I wait on element "#none-exising-element" for 10ms to be visible
       When  I click on the element "#none-exsting-element"`

lib/webdriverio.js
```diff
@@ -200,7 +200,7 @@ let WebdriverIO = function (args, modifier) {
         err.shotTaken = true
         return takeScreenshot(err)
             .catch((e) => logger.log('\tFailed to take screenshot on reject:', e))
-            .then(() => fail(err, onRejected))
+            .then(() => { throw err })
```
christian-bromann: I am not sure if this is a proper solution for the problem. `fail` does some stuff and can't just be ignored.

wvankuipers: The trow will bubble up and trigger the `reject` (line [#L173](https://github.com/wvankuipers/webdriverio/blob/61b43ed8f5cb868e41251bcdff54a432136b46f5/lib/webdriverio.js#L173)) again, triggering [#L190](https://github.com/wvankuipers/webdriverio/blob/61b43ed8f5cb868e41251bcdff54a432136b46f5/lib/webdriverio.js#L190) and [#L181](https://github.com/wvankuipers/webdriverio/blob/61b43ed8f5cb868e41251bcdff54a432136b46f5/lib/webdriverio.js#L181).

christian-bromann: @wvankuipers are you sure this happens in all cases?

wvankuipers: @christian-bromann as far as I can see it happens in all cases, I'm not 100% sure, but maybe we can add test scenario's for all the cases we can think of?

christian-bromann: > but maybe we can add test scenario's for all the cases we can think of?

That would be great. This is a super critical part of the whole project.

PR #1800, opened 2017-01-05
Clarify that the config option is called "capabilities" 

unformatt:
In Getting Started docs, clarify that the config option is called "capabilities" not "desiredCapabilities"


christian-bromann:
@unformatt thanks for the PR. The docs are correct here. The option that defines the capabilities when running WebdriverIO in standalone mode is actually `desiredCapabilities`. When running the wdio testrunner it is `capabilities` (see also [here](http://webdriver.io/guide/testrunner/configurationfile.html)).

PR #1839, opened 2017-01-24
Add missing start <a> element.

graingert:
doc fix

### DIFF
--- a/docs/guide/usage/selectors.md
+++ b/docs/guide/usage/selectors.md
@@ -65,7 +65,7 @@ console.log(browser.getText('h1*=Welcome')); // outputs: "Welcome to my Page"
 The same works for ids and class names:
 
 ```html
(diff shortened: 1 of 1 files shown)


christian-bromann:
This actually is not a fix. The example should demonstrate how to fetch an element by any text not just by link text

graingert:
@christian-bromann that is what the example shows.

christian-bromann:
The example above shows for tag name + text, the modified example should show with classname or id + text

graingert:
@christian-bromann it still shows classname or id + text

christian-bromann:
Yes so I don't understand why we need an extra link around it?

christian-bromann:
we can replace the div with h1 tho

christian-bromann:
@graingert ahh no I get it. Fixed the issue.

PR #1845, opened 2017-01-26
Add jsdoc to test runner configuration hooks

lewisf:
Some of them were obvious based on the name of the parameter but others
were not as obvious. I added it to all the hooks I'm familar with.


christian-bromann:
@lewisf thanks so much

PR #1864, opened 2017-02-09
Removed 2nd SIGINT

monolithed:
## Proposed changes

The 2nd SIGINT is not an evident way to interrupt test running and not usable with `maxInstances` at all.

Less pain, please.

### Before

img-2017-02-09-14-10-52

### After

img-2017-02-09-14-10-23

## Types of changes

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -12,7 +12,6 @@ class Runner {
         this.sigintWasCalled = false
         this.hasSessionID = false
         this.failures = 0
-        this.forceKillingProcess = false
         this.isRunning = false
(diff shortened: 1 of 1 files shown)


christian-bromann:
@monolithed the flow here is that the first ctrl + c stops all running selenium sessions of the testrunner. If the server isn't responding the second ctrl + c just kills the process all together leaving the session running though.

What is your path fixing here?

monolithed:
@christian-bromann, maybe you are right and it would be fixed by another way.... 
The problem is that forked child processes are not closed correctly when parent was killed.

PR #1893, opened 2017-02-24
Add the ability to send your own events for runner's reporters

KOPTE3:
## Proposed changes

With these changes, we can send custom messages from the runner process to the parent process, and to understand in the reporters process of what it was this message.

## Types of changes

## Checklist

### Reviewers:
@christian-bromann

### DIFF
--- a/lib/launcher.js
+++ b/lib/launcher.js
@@ -351,14 +351,14 @@ class Launcher {
 
     /**
      * emit event from child process to reporter
-     * @param  {String} rid Runner ID
+     * @param  {String} cid
      * @param  {Object} m event object
      */
-    messageHandler (rid, m) {
+    messageHandler (cid, m) {
         this.hasStartedAnyProcess = true
 
         if (!m.cid) {
-            m.cid = rid
+            m.cid = cid
         }
 
         if (m.event === 'runner:error') {
--- a/lib/launcher.js
+++ b/lib/launcher.js
@@ -319,7 +319,7 @@ class Launcher {
         this.processes.push(childProcess)
 
         childProcess
-            .on('message', this.messageHandler.bind(this))
+            .on('message', this.messageHandler.bind(this, rid))
             .on('exit', this.endHandler.bind(this, rid))
 
         childProcess.send({
@@ -351,10 +351,15 @@ class Launcher {
 
     /**
      * emit event from child process to reporter
-     * @param  {Object} m  event object
+     * @param  {String} rid Runner ID
+     * @param  {Object} m event object
      */
-    messageHandler (m) {
+    messageHandler (rid, m) {
         this.hasStartedAnyProcess = true
+
+        if (!m.cid) {
+            m.cid = rid
+        }
 
         if (m.event === 'runner:error') {
             this.reporters.handleEvent('error', m)
@@ -365,7 +370,7 @@ class Launcher {
 
     /**
      * Close test runner process once all child processes have exited
-     * @param  {Number} cid  Capabilities ID
+     * @param  {Number} rid  Capabilities ID
      * @param  {Number} childProcessExitCode  exit code of child process
      */
     endHandler (rid, childProcessExitCode) {
--- a/lib/launcher.js
+++ b/lib/launcher.js
@@ -289,7 +289,7 @@ class Launcher {
     startInstance (specs, caps, cid, server) {
         let config = this.configParser.getConfig()
         let debug = caps.debug || config.debug
-        let rid = this.getRunnerId(cid)
+	    cid = this.getRunnerId(cid)
         let processNumber = this.processesStarted + 1
 
         // process.debugPort defaults to 5858 and is set even when process
@@ -319,11 +319,11 @@ class Launcher {
         this.processes.push(childProcess)
 
         childProcess
-            .on('message', this.messageHandler.bind(this, rid))
-            .on('exit', this.endHandler.bind(this, rid))
+            .on('message', this.messageHandler.bind(this, cid))
+            .on('exit', this.endHandler.bind(this, cid))
 
         childProcess.send({
-            cid: rid,
+            cid,
             command: 'run',
             configFile: this.configFile,
             argv: this.argv,
@@ -370,17 +370,17 @@ class Launcher {
 
     /**
      * Close test runner process once all child processes have exited
-     * @param  {Number} rid  Capabilities ID
+     * @param  {Number} cid  Capabilities ID
      * @param  {Number} childProcessExitCode  exit code of child process
      */
-    endHandler (rid, childProcessExitCode) {
+    endHandler (cid, childProcessExitCode) {
         this.exitCode = this.exitCode || childProcessExitCode
         this.runnerFailed += this.exitCode !== 0 ? 1 : 0
 
         // Update schedule now this process has ended
         if (!this.isMultiremote()) {
             // get cid (capability id) from rid (runner id)
-            let cid = parseInt(rid, 10)
+            cid = parseInt(cid, 10)
 
             this.schedule[cid].availableInstances++
             this.schedule[cid].runningInstances--


monolithed:
@christian-bromann, could you accept this PR?

christian-bromann:
> With these changes, we can send custom messages from the runner process to the parent process, and to understand in the reporters process of what it was this message.

@KOPTE3 can you explain your change and use case in a little bit more depth? Thanks!

@monolithed I was on vacation and haven't had the time to look into it

monolithed:
@christian-bromann 
> can you explain your change and use case in a little bit more depth? Thanks!

I can tell our story with Allure...

**The common usage**:

```js
describe(() => {
   it('test', () => {
        allure.createStep('step 1', () => {
            browser.url('/login');
        });
   })
});
```

That's ok, but not so comfortable for us.

**Our usage**:

**step.ts**

```ts
...
global.step = function (target, propertyName, descriptor) {
	...
	descriptor.value = self.createStep(`${target.constructor.name}#${propertyName}`, descriptor.value);
	...
}
...
```

**page.ts**

```ts
class Page {
     open (path, query) {
            ...
            browser.url(path);
            ...
     }
}

export default Page;
```

**steps/login.ts**

```ts
class Login extends Page {
     @step
     open (path, query) {
            super.open(path, query);
     }
}

export default Login;
```

**test.ts**

```js
import Login from './login';

describe(() => {
   it('test', () => {
            Login.open();
   })
});
```

So, we need the `cid` parameter to execute our code in the the same context

KOPTE3:
Hi, @christian-bromann

I will try to explain why these changes are needed.


### Now webdriverio architecture (simplistically) consists of two components:

- one Launcher
- some Runners

1. The Launcher initialized reporters and fork some runners. Then it subscribes to the event 'message':

   ```javascript
   // launcher.js
   childProcess
               .on('message', this.messageHandler.bind(this, cid))
               .on('exit', this.endHandler.bind(this, rid))
   ```

2. In the processes, we forked, we send messages using `process.send()` (see the `runner.js` file).
3. Launcher process receives them, parses, and sends them to reporters. In the reporters, we are processing these messages and generate reports.

Each message that we send from child processes (in the `runner.js` file), contains a field `cid`. We determine the value of this field only once when the runner process starts and it no longer changes. So I want to bind this value to each child (runner) process in the parent (launcher) process.

This will allow us to send custom messages from the child (runner) processes and handle them into reporters.

It is now possible to write custom reporters. To do this, we need to register an event handler on several events which get triggered during the test. With my changes we will be able to create custom test-events and handle it in the custom reporters.

### For example

__We have a test:__

```javascript
describe('Our awesome test', () => {
    it('test super feature', () => {
        process.send({
            event: 'add-feature-to-report',
            feature: 'super-feature'
        });
        
        assert.equal([] + {}, '[object Object]', 'Wat?');
    })
});
```

__And in reporter we register event handler:__

```javascript
this.on('add-feature-to-report', ({event, cid, feature}) => {
    console.log(`Process ${cid}. We tests feature: ${feature}`);
});
```

__Output with my changes:__

```
Process 0-0. We tests feature: super-feature
```

__Without my changes:__

```
Process undefined. We tests feature: super-feature
```

Since the the driver can run multiple tests in parallel, the second case is bad. We can not determine the mapping between features done and tests.

### Why do I need it?

I am developing a reporter for [Allure Framework](https://github.com/allure-framework/allure-mocha). And I need the ability to create your own test-events, such as in the example above.

christian-bromann:
> Output with my changes:
> 
> Process 0-0. We tests feature: super-feature

@KOPTE3 @monolithed This is not quite true. The `cid` value is just a number. The id that has the value `0-0`, `0-1` and so on is the rid. Are you sure you only want the cid?

KOPTE3:
> This is not quite true. The cid value is just a number. The id that has the value 0-0, 0-1 and so on is the rid. Are you sure you only want the cid?

Of course, @christian-bromann! You are absolutely right
I'm just confused in variable notation. In [this place](https://github.com/webdriverio/webdriverio/blob/master/lib/launcher.js#L326) variable is passed under a different name. So in reporters cid is actually rid.

GeoffreyBooth:
I’m very interested in this as well. I’m trying to configure webdriver.io (in sync mode) with Cucumber and using wdio-allure-reporter, but out of the box the Allure reports don’t include any attachments (not even screenshots) for failing tests. I want to attach browser console logs and server logs. I have a hook that gets those logs as strings, but I don’t see how to add them as attachments to Allure reports. @KOPTE3 do you care to share what you have so far? Is there some way to add an attachment to an Allure result without modifying webdriver.io or wdio-allure-reporter?

lib/launcher.js
```diff
@@ -351,11 +351,16 @@ class Launcher {
...
-     * @param  {Object} m  event object
+     * @param  {String} rid Runner ID
+     * @param  {Object} m event object
      */
-    messageHandler (m) {
+    messageHandler (rid, m) {
         this.hasStartedAnyProcess = true
 
+        if (!m.cid) {
+            m.cid = rid
```
christian-bromann: let's not confuse `cid` and `rid` here

KOPTE3: I corrected the remark

lib/launcher.js
```diff
@@ -319,7 +319,7 @@ class Launcher {
         this.processes.push(childProcess)
 
         childProcess
-            .on('message', this.messageHandler.bind(this))
+            .on('message', this.messageHandler.bind(this, rid))
```
christian-bromann: you are passing `rid` here ...

lib/launcher.js
```diff
@@ -351,11 +351,16 @@ class Launcher {
 
     /**
      * emit event from child process to reporter
-     * @param  {Object} m  event object
+     * @param  {String} cid
+     * @param  {Object} m event object
      */
-    messageHandler (m) {
+    messageHandler (cid, m) {
```
christian-bromann: but here it is `cid` again

KOPTE3: Yes. This is correct, because [here](https://github.com/webdriverio/webdriverio/blob/master/lib/launcher.js#L326) we pass variable `rid` as variable `cid`. And [here](https://github.com/webdriverio/webdriverio/blob/master/lib/runner.js#L441) we subscribe on messages and [here](https://github.com/webdriverio/webdriverio/blob/master/lib/runner.js#L21) we save this `cid` (which is actually `rid`).

KOPTE3: And [here](https://github.com/webdriverio/webdriverio/blob/master/lib/runner.js#L69) we send messages with `cid`.

KOPTE3: I think this is a little confusion in the names of variables. But the consequences of this confusion will be difficult to correct. Because in reporters everywhere is used `cid` (which is actually `rid`)...
I'm not mistaken?

christian-bromann: Maybe we should get rid of `rid` in `startInstance` and `endHandler`

GeoffreyBooth: Might I suggest perhaps a different PR would be a good place to standardize on either `rid` or `cid` (probably `cid`)?

christian-bromann: @KOPTE3 usually yes but the changes are so small that I would be fine to have it in this PR but happy to merge 2 PRs as well 😉

KOPTE3: > Maybe we should get rid of `rid` in `startInstance` and `endHandler`

How do I do this, if in [`startInstance`](https://github.com/webdriverio/webdriverio/blob/master/lib/launcher.js#L292) we use both of these variables?

KOPTE3:
Are there still some requirements for this PR?

christian-bromann:
> Are there still some requirements for this PR?

time for maintainer  😉

monolithed:
@christian-bromann, these changes are not so critical, but they are very useful for many people who wants to make a custom reporter. We have been waiting for this release for 20 days. So, there is  a reasonable question we'd like to know what are the doubts?

christian-bromann:
> So, there is a reasonable question we'd like to know what are the doubts?

No doubts but I don't have the time to merge and release it yet. I hope I will get on it this weekend

PR #1905, opened 2017-03-09
Update Saucelabs example to use .then()

jeresig:
## Proposed changes

The example was still using the old callback-style, when it should be using `.then()` (as written it didn't output anything when run).

## Types of changes

## Checklist


christian-bromann:
Thanks for the PR. Any chance we could update the [testobject example](https://github.com/webdriverio/webdriverio/blob/master/examples/cloudservices/webdriverio.testingbot.js#L21) too? Thanks so much!

PR #1920, opened 2017-03-12
Added clarification to waitforTimeout option

broder:
## Proposed changes

The lowercase `f` in `waitforTimeout` was the source of some confusion for me when setting up so I've added some further documentation which highlights this. I'm not sure if this is the best way to inform users or if you think this is necessary at all!

## Types of changes

## Checklist

### DIFF
--- a/docs/guide/getstarted/configuration.md
+++ b/docs/guide/getstarted/configuration.md
@@ -150,7 +150,7 @@ screenshotOnReject: {
 ```
 
 ### waitforTimeout
-Default timeout for all waitForXXX commands.
+Default timeout for all waitForXXX commands. Note the lowercase `f`.
 
 Type: `Number`<br>
 Default: *500*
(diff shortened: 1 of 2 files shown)


broder:
The property is given in all of these files but I still think it would be good to explicitly highlight the fact that it's not camel case.

christian-bromann:
With given I meant that it is already in the code (e.g. of the config file) so that a user intentionally would need to rename the property to fall into the lowercase trap. Therefor I don't see a need for that there.

broder:
Fair enough - I've updated the PR accordingly

monolithed:
> The lowercase f in waitforTimeout was the source of some confusion

I'm absolutely agree. I've already added [a fix](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/14945/files#diff-edb3c2fb25a2ea3c4a46bc3ce3edbe84R424) to `@types/webdriverio`, but we sill no way to have the config file written in TypeScript (

Issue #113, opened 2013-12-04
Run with chromedriver


brianmhunt:
I am trying to get `webdriverjs` to work with chromedriver.

The README seems to indicate that a local driver should simply work, however I found no mention of chromedriver in the source.

I would think it would seem to be relatively straightforward to get this working, requiring:
1. Starting (and exiting) chromedriver on demand;
2. Connecting to `localhost:9515` and changing the `startPath` in `lib/utils/RequestHandler.js` to `/` (or, more specifically, making it an option) or add `--url-base=/wd/hub` to the chromedriver options.

Or at least, that's what I have gleaned from `selenium-webdriver` and [a comment on stackoverflow](http://stackoverflow.com/a/18394150/19212).

Then again, I don't quite get chromedriver. It's just a bit surprising that it appears to not be supported by webdriverjs.

Please note that I am unable to use java in our test environment.

dristic:
As a heads up, if you get the `unhandled request` error for all endpoints when hitting chrome driver, you need to start chrome driver on windows without the first slash:

`./chromedriver.exe --url-base=wd/hub --port=PORT`

Glavin001:
I have found the following works:

```
npm install webdriverio chromedriver
```

```javascript
const webdriverio = require('webdriverio');
const chromedriver = require('chromedriver');

const port = 9515;
const args = [
  '--url-base=wd/hub',
  `--port=${port}`
];
chromedriver.start(args);

const options = {
  port,
  desiredCapabilities: {
    browserName: 'chrome'
  }
};
webdriverio
  .remote(options)
  .init()
  .on('error', (e) => console.error(e))
  .once('end', () => {
    console.log('end');
    chromedriver.stop();
  })
  .url('http://www.google.com')
  .getTitle().then(function (title) {
    console.log('Title was: ' + title);
  })
  .end();
```

Hope that helps someone else!

kolya-ay:
@Glavin001, it seems that `once('end', () => chromedriver.stop())` stops chromedriver too early for correct 'chrome' process shutdown (e.g. `chromium` instance is still in memory). How do you handle this? Or maybe it's only my concern..

christian-bromann:
` process.nextTick` is the way to handle this 😉

thebarty:
Guys, I lost some time on this. @Glavin001 Thanks to your code I finally got it working!!!

Any chance of adding a link to this topic in the readme "ChromeDriver"-Paragraph?

[...]

PR #1933, opened 2017-03-17
Add ability to run geckodriver without selenium

APshenkin:
## Proposed changes

As you know there is ability to run tests without selenium at all. You just need to start driver and send requests to it. Selenium is just proxy, that unifies some difference between drivers and send responses to client in Selenium JSONWire format. But there is other implementation of Selenium, that don't unifies difference but they provide isolation for browsers. Here is an example: https://github.com/aandryashin/selenoid

In example above you can manage containers/browsers isolation and it's take only one role. The problem with difference should be resolved by client.

Now there is only one difference - geckodriver. It uses W3C standart and has other format of responses (don't has status code, elements return in other format etc)
example:
```
Selenium format:
{"value":[{"ELEMENT":"a815dba6-e075-964c-8aa6-2983cb893e7a”}]}
W3C geckodriver format:
{"value":[{"element-6066-11e4-a52e-4f735466cecf":"a815dba6-e075-964c-8aa6-2983cb893e7a”}]}
```

This pull request give you ability to use selenoid or geckodriver directly. There is new option `useW3C`. When it's `true` and browser is Firefox you can use geckodriver directly or use tools like selenoid.

This PR is not guarantied that everything will works great. I've just added an adhoc and hasn't much time to test it. So if you will be able to continue this theme, it would be great.

P.S.
Several thoughts why it's needed:
1. Exclude JVM from test process. If I develop tests on JS, I want to decrease amount of other frameworks, that I should use.
2. Have any alternative for hubs
3. Ability to use browser driver directly
3. Selenoid provide smaller docker containers for browsers and they starts faster than selenium containers. It's very nice for grid dockerization.

## Types of changes

## Checklist

### DIFF
--- a/lib/commands/addValue.js
+++ b/lib/commands/addValue.js
@@ -56,7 +56,7 @@ let addValue = function (selector, value) {
         let elementIdValueCommands = []
 
         for (let elem of res.value) {
-            elementIdValueCommands.push(self.elementIdValue(elem.ELEMENT, value))
+            elementIdValueCommands.push(self.elementIdValue(elem[Object.keys(elem)[0]], value))
         }
 
         return this.unify(elementIdValueCommands)
--- a/lib/protocol/alertAccept.js
+++ b/lib/protocol/alertAccept.js
@@ -26,7 +26,7 @@ let alertAccept = function () {
         method: 'POST'
     }
 
-    return this.requestHandler.create(requestOptions, {dummy: "dummy"})
+    return this.requestHandler.create(requestOptions, {dummy: 'dummy'})
 }
 
 export default alertAccept
--- a/lib/protocol/alertAccept.js
+++ b/lib/protocol/alertAccept.js
@@ -26,7 +26,7 @@ let alertAccept = function () {
         method: 'POST'
     }
 
-    return this.requestHandler.create(requestOptions)
+    return this.requestHandler.create(requestOptions, {dummy: "dummy"})
 }
 
 export default alertAccept
--- a/lib/protocol/elementIdAttribute.js
+++ b/lib/protocol/elementIdAttribute.js
@@ -25,7 +25,7 @@ let elementIdAttribute = function (id, attributeName) {
         command = '/property/'
     }
 
-    return this.requestHandler.create(`/session/:sessionId/element/${id}/attribute/${attributeName}`)
+    return this.requestHandler.create(`/session/:sessionId/element/${id}${command}${attributeName}`)
 }
 
 export default elementIdAttribute
--- a/lib/protocol/elementIdAttribute.js
+++ b/lib/protocol/elementIdAttribute.js
@@ -19,6 +19,12 @@ let elementIdAttribute = function (id, attributeName) {
         throw new ProtocolError('number or type of arguments don\'t agree with elementIdAttribute protocol command')
     }
 
+    var command = '/attribute/'
+
+    if (this.options.useW3C && this.desiredCapabilities.browserName === 'firefox' && attributeName === 'value') {
+        command = '/property/'
+    }
+
     return this.requestHandler.create(`/session/:sessionId/element/${id}/attribute/${attributeName}`)
 }
 
--- a/lib/protocol/window.js
+++ b/lib/protocol/window.js
@@ -31,8 +31,10 @@ let window = function (windowHandle) {
     }
 
     if (typeof windowHandle === 'string') {
-        data = { name: windowHandle }
-        requestOptions.method = 'POST'
+        if (this.options.useW3C && this.desiredCapabilities.browserName === 'firefox') {
+            data = { handle: windowHandle }
+        } else data = { name: windowHandle };
+        requestOptions.method = 'POST';
     }
 
     return this.requestHandler.create(requestOptions, data)
--- a/lib/protocol/window.js
+++ b/lib/protocol/window.js
@@ -33,8 +33,8 @@ let window = function (windowHandle) {
     if (typeof windowHandle === 'string') {
         if (this.options.useW3C && this.desiredCapabilities.browserName === 'firefox') {
             data = { handle: windowHandle }
-        } else data = { name: windowHandle };
-        requestOptions.method = 'POST';
+        } else data = { name: windowHandle }
+        requestOptions.method = 'POST'
     }
 
     return this.requestHandler.create(requestOptions, data)
--- a/lib/protocol/windowHandleMaximize.js
+++ b/lib/protocol/windowHandleMaximize.js
@@ -1,6 +1,6 @@
 /**
  *
- * Maximize the specified window if not already maximized. If the :windowHandle URL parameter is "current",
+ * Maximize the specified window if not already maximized. If the :windowHandle URL parameter is 'current',
  * the currently active window will be maximized.
  *
  * @param {String=} windowHandle window to maximize (if parameter is falsy the currently active window will be maximized)
@@ -11,7 +11,6 @@
  */
 
 let windowHandleMaximize = function (windowHandle = 'current') {
-
     var path = `/session/:sessionId/window/${windowHandle}/maximize`
 
     if (this.options.useW3C && this.desiredCapabilities.browserName === 'firefox') {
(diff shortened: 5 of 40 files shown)


lib/commands/addValue.js
```diff
@@ -56,7 +56,7 @@ let addValue = function (selector, value) {
         let elementIdValueCommands = []
 
         for (let elem of res.value) {
-            elementIdValueCommands.push(self.elementIdValue(elem.ELEMENT, value))
+            elementIdValueCommands.push(self.elementIdValue(elem[Object.keys(elem)[0]], value))
```
christian-bromann: Can we unify this operation? I've seen it couple of times in this PR and when the behavior changes we would need to go to each of these files. I think we can check the element value once before we make the request in the requestHandler.

lib/protocol/alertAccept.js
```diff
@@ -26,7 +26,7 @@ let alertAccept = function () {
         method: 'POST'
     }
 
-    return this.requestHandler.create(requestOptions)
+    return this.requestHandler.create(requestOptions, {dummy: 'dummy'})
```
christian-bromann: First of all why can't we just pass in an empty object (`{}`). Also as I said before, instead of changing each file, let's handle the payload one time in the requestHandler create function

lib/protocol/elementIdAttribute.js
```diff
@@ -19,7 +19,13 @@ let elementIdAttribute = function (id, attributeName) {
         throw new ProtocolError('number or type of arguments don\'t agree with elementIdAttribute protocol command')
     }
 
-    return this.requestHandler.create(`/session/:sessionId/element/${id}/attribute/${attributeName}`)
+    var command = '/attribute/'
+
+    if (this.options.useW3C && this.desiredCapabilities.browserName === 'firefox' && attributeName === 'value') {
```
christian-bromann: Why limiting it to firefox only? If this is the new protocol we should make it for all browser. If someone has useW3C enabled we should assume the driver has the webdriver protocol implemented

lib/protocol/window.js
```diff
@@ -31,7 +31,9 @@ let window = function (windowHandle) {
     }
 
     if (typeof windowHandle === 'string') {
-        data = { name: windowHandle }
+        if (this.options.useW3C && this.desiredCapabilities.browserName === 'firefox') {
```
christian-bromann: same here

lib/protocol/windowHandleMaximize.js
```diff
@@ -11,10 +11,16 @@
  */
 
 let windowHandleMaximize = function (windowHandle = 'current') {
+    var path = `/session/:sessionId/window/${windowHandle}/maximize`
+
+    if (this.options.useW3C && this.desiredCapabilities.browserName === 'firefox') {
```
christian-bromann: and here

lib/utils/RequestHandler.js
```diff
@@ -135,9 +136,16 @@ class RequestHandler {
             if (this.sessionID === null && requestOptions.requiresSession !== false) {
                 this.sessionID = body.sessionId
 
+                var options = body.value
+
+                if (this.defaultOptions.useW3C && this.defaultOptions.desiredCapabilities.browserName === 'firefox') {
```
christian-bromann: and here

lib/utils/RequestHandler.js
```diff
@@ -13,6 +13,7 @@ class RequestHandler {
     constructor (options, eventHandler, logger) {
         this.sessionID = null
         this.startPath = options.path === '/' ? '' : options.path || '/wd/hub'
+        this.selenoid = options.selenoid
```
christian-bromann: what does `options.selenoid` contain? Can we make this more general?

lib/utils/RequestHandler.js
```diff
@@ -175,8 +183,11 @@ class RequestHandler {
                 /**
                  * Resolve with a healthy response
                  */
-                if (!err && body && body.status === 0) {
-                    return resolve({body, response})
+
+                if (!err && body) {
```
christian-bromann: and here

lib/commands/addValue.js
APshenkin: Ok, will think how to do that

lib/protocol/alertAccept.js
APshenkin: When passing empty object to geckodriver directly, it returns error, that body data is invalid or missing. Looks like request library don't send data, if it's empty

lib/protocol/elementIdAttribute.js
APshenkin: The problem is, that only geckodriver for Firefox support now W3C protocol. Chrome, IE, Safari and Edge use old Selenium JSONWire protocol at this moment.
If I want to use direct drivers without selenium I want to setup this in config one time and do not change depending on the browser. Maybe other name for this option will resolve this misunderstanding

lib/protocol/alertAccept.js
christian-bromann: > Looks like request library don't send data, if it's empty

That can be changed. I would rather like to see `{}` than that.

lib/utils/RequestHandler.js
APshenkin: Sorry, obsolete option. Will remove it

lib/protocol/elementIdAttribute.js
christian-bromann: @APshenkin ok, let's at least have this check within a helper script so we only need to change the check in one file

lib/commands/addValue.js
APshenkin: I can propose to send `elem` object to requestHandler and check the value in it, is it OK?

christian-bromann: > is it OK?

sounds good to me

APshenkin: Checked the code again and found, that we add elementId in path in protocol helpers. Looks like we shouldn't move this logic to request handler. I propose to create new helper in utils that will return elementId. and use it in each command helper

monolithed:
> The Microsoft Edge implementation of WebDriver supports commands from both the W3C WebDriver Specification and the JSON Wire Protocol

https://docs.microsoft.com/en-us/microsoft-edge/webdriver-commands

> ChromeDriver is an implementation of the WebDriver standard

https://github.com/scheib/chromium/blob/458d956db161377610486b7c82a95fc485f60b9b/chrome/test/chromedriver/README.txt#L4

https://github.com/scheib/chromium/commit/de6591ecb593950240b0c8ec607c0808d7215bd0

So, it seems IE and Chrome have already implemented the current standard protocol

christian-bromann:
I am currently working on a patch that will provide full support with latest Webdriver spec.

sherlock1982:
Is it already possible to run geckodriver standalone in 4.8.0 ?

christian-bromann:
Yes!

sherlock1982:
Could you please provide a sample? Like [here](https://github.com/webdriverio/webdriverio/issues/113).
There's no --url-base in geckodriver.

zeljkofilipin:
@sherlock1982 could you please provide more information on what you did? I can not find it in the documentation.

sherlock1982:
You can launch geckodriver directly on your host without selenium hub or node. Than in Webdriver you need a configuration similar to this:

```
{
    host: 'localhost',
    desiredCapabilities: {
        browserName: 'firefox'
    },
    path: '/'
}
```

The same for chromedriver:

```
{
    host: 'localhost',
    desiredCapabilities: {
        browserName: 'chrome'
    },
}
```

Please note that 'path' you can specify in chromedriver commandline parameters but firefox should be only '/'.

Also note that chrome works almost like selenium. But firefox strictly follows standard. Which causes:
#2392 and #2394 

I made a patch to fix last one:

```
patch-package
--- a/node_modules/webdriverio/build/lib/utils/RequestHandler.js
+++ b/node_modules/webdriverio/build/lib/utils/RequestHandler.js
@@ -237,8 +237,8 @@ var RequestHandler = function () {
                     /**
                      * Resolve with a healthy response
                      */
-                    if (!err && body && (body.status === 0 || body.value && !body.status && !(body.value.error || body.value.stackTrace))) {
-                        return resolve({ body: body, response: response });
+                    if (!err && body && (body.status === 0 || (body.value === null && !body.status) || (body.value !== undefined && !body.status && !(body.value.error || body.value.stackTrace)))) {
+                        return resolve({body, response})
                     }
 
                     if (fullRequestOptions.gridCommand) {


```

christian-bromann:
> Also note that chrome works almost like selenium. But firefox strictly follows standard. Which causes:

This is fixed and will be released any second

PR #1935, opened 2017-03-17
Update snyk config to allow Travis master and PR builds again

timbru31:
## Proposed changes

This PR fixes two issues with snyk
- runs snyk only when no PR is created (i.e. master, branches from contributors - since auth token is a secret in Travis and could be exposed)
- ignores vulnerable shelljs dependency for 30 days, so master runs the testsuite again  
(side note: this happens due to npm@2 peerDependency resolving)

## Types of changes

- [x] Change to build pipeline

## Checklist

## Further comments

The CONTRIBUTING doc links to a GitHub page, **not** the one in the project root. I'm fine with both, since OSS is <3

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,10 @@ before_script:
   - cd test/site/www && bower install && cd ../../../
   - phantomjs --webdriver=4444 2>&1 >/dev/null &
 
-script: "sleep 5 && npm run test-snyk && npm run test"
+script:
+    - sleep 5
+    - 'if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then npm run test-snyk; fi'
+    - npm run test
 
 addons:
   sauce_connect: true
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -10,7 +10,10 @@ before_script:
   - cd test/site/www && bower install && cd ../../../
   - phantomjs --webdriver=4444 2>&1 >/dev/null &
 
-script: "sleep 5 && npm run test-snyk && npm run test"
+script:
+    - sleep 5
+    - 'if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then bash npm run test-snyk; fi'
```
timbru31: this needs to be changed to == false, but I want to trigger it and see if things are working like it should at a PR

timbru31:
The `wdio` env job seems just to be failing, but not due to the snyk error anymore:
https://travis-ci.org/webdriverio/webdriverio/jobs/212206576

PR #1946, opened 2017-03-20
feat: make url act like url.resolve

realzoberg:
## Proposed changes

I suggest to change browser.url interaction with baseUrl. It will be more clear and useful if it acts like 
url.resolve.

For example

basePath = https://base_path/baz
pass /foo/bar => https://base_path/foo/bar
pass foo/bar => https://base_path/baz/foo/bar
pass https://foo.bar => https://foo.bar

## Types of changes

## Checklist

## Further comments

Maybe I should create alternative url command and mark existing url as deprecated? Or add optional behaviour to url with this functionality?

Have to mention that I need this functionality in my project.

### DIFF
--- a/lib/protocol/url.js
+++ b/lib/protocol/url.js
@@ -1,3 +1,5 @@
+const nodeUrl = require('url')
+
 /**
  *
  * Protocol binding to load or get the URL of the browser.
@@ -23,8 +25,8 @@ let url = function (uri) {
 
     if (typeof uri === 'string') {
         data.url = uri
-        if (typeof this.options.baseUrl === 'string' && /^(\/|\?)/.test(data.url)) {
-            data.url = this.options.baseUrl + data.url
+        if (typeof this.options.baseUrl === 'string') {
+            data.url = nodeUrl.resolve(this.options.baseUrl, data.url)
         }
     }
 
--- a/test/site/www/base-url/index.html
+++ b/test/site/www/base-url/index.html
@@ -1,0 +1,15 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>WebdriverIO Testpage</title>
+    <meta name="apple-mobile-web-app-status-bar-style" content="black-translucent">
+    <meta name="viewport" content="user-scalable=no, width=device-width, initial-scale=1, maximum-scale=1">
+</head>
+<body>
+
+    <header>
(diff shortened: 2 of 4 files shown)


realzoberg:
@christian-bromann I need baseUrl with pathname for my tests. Should I create one more configuration with custom baseUrl for my test?

christian-bromann:
@realzoberg not sure what you mean?

realzoberg:
@christian-bromann we setup baseUrl for tests here
https://github.com/webdriverio/webdriverio/blob/master/test/setup.js#L9
https://github.com/webdriverio/webdriverio/blob/master/test/conf/defaults.js#L2

And here we path baseUrl for tests https://github.com/webdriverio/webdriverio/blob/master/test/setup.js#L39

Looks like I have to create custom setup.js if I wanna run my tests with different baseUrl?

christian-bromann:
> Looks like I have to create custom setup.js if I wanna run my tests with different baseUrl?

can you try to set `client.baseUrl` directly before you run the command?

realzoberg:
@christian-bromann That really works! Thank you!

test/site/www/base-url/index.html
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
```
christian-bromann: For what do you need this page?

realzoberg: @christian-bromann I retrieve this page in this test case https://github.com/webdriverio/webdriverio/pull/1946/files#diff-f445c8ec5b775aaf329a644cb15a09b8L8

lib/protocol/url.js
```diff
@@ -1,3 +1,5 @@
+const nodeUrl = require('url')
```
christian-bromann: @realzoberg let's use import here:

```js
import url from 'url'
```

christian-bromann:
ping @realzoberg .. please update on top of current master. We updated our test infrastructure and enabled PRs builds.

mattc41190:
💥 not a minor change lol 💥

jankcat:
If @realzoberg doesn't take this work on in the next bit, I will redo it with the requested changes. One other change that should probably be made: documenting the behaviour [here](http://webdriver.io/api/protocol/url.html).

PR #1949, opened 2017-03-22
Add missing unit tests for Sanitize.caps, Sanitize.args and Sanitize.css

timbru31:
## Proposed changes

This PR adds three missing unit tests for the Sanitize helper and uses template strings instead of string concatenation

## Types of changes

- [x] Missing tests

## Checklist

### DIFF
--- a/lib/helpers/sanitize.js
+++ b/lib/helpers/sanitize.js
@@ -23,6 +23,10 @@ let sanitizeString = function (str) {
  * @param {Object} caps  Selenium capabilities
  */
 let caps = function (caps) {
+    if (!caps) {
+        return ''
+    }
+
     let result
 
     /**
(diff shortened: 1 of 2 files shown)


lib/helpers/sanitize.js
```diff
@@ -23,6 +23,10 @@ let sanitizeString = function (str) {
  * @param {Object} caps  Selenium capabilities
  */
 let caps = function (caps) {
+    if (!caps) {
```
timbru31: without input validation it would fail in line 35 (since it assumes caps is an object)

lib/helpers/sanitize.js
```diff
@@ -43,11 +47,15 @@ let caps = function (caps) {
  * @param  {Array} args arguments object
  */
 let args = function (args) {
+    if (!args || !Array.isArray(args)) {
+        return ''
+    }
+
     return args.map((arg) => {
```
timbru31: without input validation it would fail since it directly invokes `.map`

timbru31:
@christian-bromann a friendly reminder ;)

PR #1959, opened 2017-03-30
exitCode doesn't reset after failed tests

WillBrock:
## Proposed changes

When a test fails it is setting the exitCode to 1 but it never sets it back to 0. This results in the exitCode to be 1 when the next test is ran and internally is increasing runnerFailed by 1 after each test even if it passes.

When the bail config option is set to say 10 and a test fails only 9 more tests will run and then bail no matter if they pass or fail.

## Types of changes

## Checklist

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -15,7 +15,7 @@ _Put an `x` in the boxes that apply_
 
 _Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._
 
-- [ ] I have read the [CONTRIBUTING](/CONTRIBUTING.md) doc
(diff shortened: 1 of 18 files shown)


monolithed:
Thanks for the PR!

christian-bromann:
@WillBrock can you please elaborate on what this change will do and why it is needed? If one test fails it should exit with 1 no matter if following test are passing.

monolithed:
Oh, I misunderstood the problem, so the same question.

WillBrock:
Inside the endHandler method this.runnerFailed is set to 0 or 1 based on the exitCode. When a single test fails exitCode is set to 1 resulting in runnerFailed to be true and increase by 1. Since exitCode is never reset to 0 runnerFailed keeps increasing by one after each test no matter a pass or fail.

runnerFailed is used to keep track of how many tests have failed before exiting based on the "bail" number that is set in the config file.

If the bail config option is set to 10 and you have 50 tests and the 6th test fails then only 16 tests will be ran because runnerFailed increases by 1 after the first failed test and the attempts will hit the bail limit.

christian-bromann:
@WillBrock please rebase on top of current master. I updated the projects test infrastructure.

christian-bromann:
I don't think resetting exit code is a good idea. Found a better approach. Thanks for finding this bug.

PR #1977, opened 2017-04-11
Query parameters

hborawski:
## Proposed changes

This change allows a user to append query parameters based on an object populated in the default options when creating the client. This is useful if requests need to be proxied through a specialized tunnel for instance.

## Types of changes

## Checklist

## Further comments

This may not fit a use case for many users, but it is something that we use at scale. We have been using our own fork of this at the company where I work to use this functionality, but I would prefer to be able to point at the main repository and not have to pull in updates manually. This change does not break the API, so hopefully we can work out something to get this integrated.

### DIFF
--- a/docs/guide/getstarted/configuration.md
+++ b/docs/guide/getstarted/configuration.md
@@ -162,7 +162,7 @@ Type: `Number`<br>
 Default: *500*
 
 ### queryParams
-A key-value store of query parameters to be added to every request send by webdriverio.
+A key-value store of query parameters to be added to every selenium request.
 Type: `Object`<br>
 Default: None
 
@@ -173,7 +173,8 @@ queryParams: {
   specialKey: 'd2ViZHJpdmVyaW8='
 }
 
-//Every request will end with "?specialKey=d2ViZHJpdmVyaW8="
+// Selenium request would look like:
+// http://127.0.0.1:4444/v1/session/a4ef025c69524902b77af5339017fd44/window/current/size?specialKey=d2ViZHJpdmVyaW8%3D
 ```
 
 ## debug
--- a/docs/guide/getstarted/configuration.md
+++ b/docs/guide/getstarted/configuration.md
@@ -161,7 +161,23 @@ Default interval for all waitForXXX commands.
 Type: `Number`<br>
 Default: *500*
 
+### queryParams
+A key-value store of query parameters to be added to every request send by webdriverio.
+Type: `Object`<br>
+Default: None
+
+**Example:**
+
+```js
+queryParams: {
+  specialKey: 'd2ViZHJpdmVyaW8='
+}
+
+//Every request will end with "?specialKey=d2ViZHJpdmVyaW8="
+```
+
 ## debug
+
 Enables node debugging
 
 Type: `Boolean`<br>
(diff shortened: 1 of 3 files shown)


christian-bromann:
@hborawski can you explain the reason why you guys need to attach query parameters to each Selenium request?

hborawski:
We have an internal platform that proxies our requests to SauceLabs and this platform requires that we append an apikey as a query parameter to every command sent by webdriverio (or webdriver for teams that use java). I am not entirely sure why it is done this way, but in an effort to push webdriverio as our standard for UI E2E testing and move away from the java based webdriver approach, we need to be able to use this platform for the teams that rely on it.

docs/guide/getstarted/configuration.md
```diff
@@ -161,7 +161,23 @@ Default interval for all waitForXXX commands.
 Type: `Number`<br>
 Default: *500*
 
+### queryParams
+A key-value store of query parameters to be added to every request send by webdriverio.
```
christian-bromann: Can we clarify that the query is added to the Selenium request?

docs/guide/getstarted/configuration.md
```diff
@@ -161,7 +161,23 @@ Default interval for all waitForXXX commands.
...
+Default: None
+
+**Example:**
+
+```js
+queryParams: {
+  specialKey: 'd2ViZHJpdmVyaW8='
+}
+
+//Every request will end with "?specialKey=d2ViZHJpdmVyaW8="
```
christian-bromann: Let's have a selenium request example here to clarify things better

lib/utils/RequestHandler.js
```diff
@@ -59,11 +59,20 @@ class RequestHandler {
             throw new RuntimeError(101)
         }
 
+        // Build query string based on queryParams in defaultOptions
+        var queryString = ''
+        if (this.defaultOptions.queryParams !== undefined) {
```
christian-bromann: Instead of checking for being undefined let's do a typecheck to make sure queryParams is an object.

lib/utils/RequestHandler.js
```diff
@@ -59,11 +59,20 @@ class RequestHandler {
...
+            var opts = this.defaultOptions
+            queryString = '?' + Object.keys(this.defaultOptions.queryParams).map(function (key) {
+                return key + '=' + encodeURIComponent(opts.queryParams[key])
```
christian-bromann: let's do `${key}=${encodeURIComponent(opts.queryParams[key])}`

lib/utils/RequestHandler.js
```diff
@@ -59,11 +59,20 @@ class RequestHandler {
...
+            }).join('&')
+        }
+
         newOptions.uri = url.parse(
             this.defaultOptions.protocol + '://' +
             this.defaultOptions.hostname + ':' + this.defaultOptions.port +
             (requestOptions.gridCommand ? this.gridApiStartPath : this.startPath) +
-            requestOptions.path.replace(':sessionId', this.sessionID || ''))
+            requestOptions.path.replace(':sessionId', this.sessionID || '') + queryString)
```
christian-bromann: Instead of adding the querystring to the uri let's use the `qs` parameter

lib/utils/RequestHandler.js
christian-bromann: actually there is no need to parse of the object, you can just apply the object via `qs` param

Issue #1978, opened 2017-04-11
Potential Memory Leak in Test Runner


jankcat:
## The problem

I am trying to run multiple mocha specs with several tests in each spec file, across 2-3 browsers on saucelabs. It seems that the parent node process is not releasing memory when the child process closes:

First set of child processes:
![screen shot 2017-04-11 at 1 16 50 pm](https://cloud.githubusercontent.com/assets/7744158/24927002/b90b5fce-1eba-11e7-9703-08ef208617a7.png)

Gaining memory on the parent as we send commands to the grid:
![screen shot 2017-04-11 at 1 18 11 pm](https://cloud.githubusercontent.com/assets/7744158/24927047/d2c1eb68-1eba-11e7-8cc9-ad5cbc846c3f.png)

After two of the children have completed, new ones start:
![screen shot 2017-04-11 at 1 21 10 pm](https://cloud.githubusercontent.com/assets/7744158/24927076/ef10192a-1eba-11e7-9f8e-22d02b76c26d.png)

I have built an example here: https://github.com/jankcat/wdio-example-leak
Run the ./run.sh file after setting your sauce creds in main.config.js

## Environment

* WebdriverIO version: webdriverio@4.6.2
* Node.js version: 7.7.2
* WDIO Test Runner
* running synchronous tests
* Additional wdio packages used:
  * "wdio-mocha-framework": "^0.5.9"
  * "wdio-sauce-service": "^0.3.1"
  * "wdio-screenshot": "^0.5.2"

## Link to Selenium/WebdriverIO logs

https://gist.github.com/jankcat/86758abb32156f63e8e402b308d41071

## Code To Reproduce Issue [ Good To Have ]

I have built an example here: https://github.com/jankcat/wdio-example-leak
Run the ./run.sh file after setting your sauce creds in main.config.js

jankcat:
Also, here is the error I get in terminal once the memory hits 1.5gb on the parent process:
```
<--- Last few GCs --->

[50353:0x103801600]   952483 ms: Mark-sweep 1397.7 (1443.5) -> 1397.7 (1443.5) MB, 18.4 / 0.0 ms  allocation failure GC in old space requested
[50353:0x103801600]   952505 ms: Mark-sweep 1397.7 (1443.5) -> 1397.7 (1411.5) MB, 22.2 / 0.0 ms  last resort 
[...]

PR #1993, opened 2017-04-19
Remove screenshot data from ReporterStats payload

jankcat:
## Proposed changes

I removed the raw screenshot data in the RunnerStats.output() function, so that it is not stored in the parent process's memory.

Addresses #1978.

## Types of changes

## Checklist

I did not see any current UTs on this function or class, let me know if I missed something...

## Further comments

I chose to remove the data from the payload specifically in the ReporterSpecs class (instead of in the process listener/runner or BaseReporter), so that any custom reporters that utilize this data have the opportunity to use the raw screenshot data before it is scrubbed.

### DIFF
--- a/lib/utils/ReporterStats.js
+++ b/lib/utils/ReporterStats.js
@@ -245,6 +245,10 @@ class ReporterStats extends RunnableStats {
 
     output (type, runner) {
         runner.time = new Date()
+        // Remove the screenshot data to reduce RAM usage on the parent process
+        if (type === 'screenshot') {
+            runner.data = null
+        }
         if (ReporterStats.getIdentifier(runner) && runner.parent) {
             this.getTestStats(runner).output.push({
                 type,


jankcat:
Awesome, thanks!

christian-bromann:
@jankcat as mentioned in the gitter chat please add a comment explaining why we remove the screenshot there

jankcat:
@christian-bromann I have added a comment now

lib/utils/ReporterStats.js
```diff
@@ -245,6 +245,10 @@ class ReporterStats extends RunnableStats {
 
     output (type, runner) {
         runner.time = new Date()
+        // As mentioned in Issue #1978, remove the screenshot data to reduce RAM usage on the parent process
```
christian-bromann: no need to mention the issue number, git allows us to figure that out 😉

jankcat:
Fair enough, I have removed that part of the comment 😛 , thanks for the tip

PR #2001, opened 2017-04-20
Add missing before/after Cucumber functions

ryantmer:
These are present elsewhere, but appear to have been missed in these two files.

## Proposed changes

Filling in some missing fields in example and helper files.

## Types of changes

## Checklist

## Further comments

(None)

### DIFF
--- a/lib/helpers/wdio.conf.ejs
+++ b/lib/helpers/wdio.conf.ejs
@@ -295,6 +295,42 @@ exports.config = {
     // afterSuite: function (suite) {
     // },
     /**
+     * Runs before a Cucumber feature
+     * @param {Object} feature feature details
+     */
+    // beforeFeature: function (feature) {
+    // },
+    /**
+     * Runs after a Cucumber feature
+     * @param {Object} feature feature details
+     */
+    // afterFeature: function (feature) {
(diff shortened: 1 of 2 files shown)


ryantmer:
Looks like Sauce had some trouble connecting - I'd imagine because this is coming from a fork?

lib/helpers/wdio.conf.ejs
```diff
@@ -295,6 +295,42 @@ exports.config = {
     // afterSuite: function (suite) {
     // },
     /**
+     * Runs before a Cucumber feature
```
christian-bromann: Can we check if the framework is Cucumber and only include these when it is (and also not include `beforeTest` and other Mocha/Jasmine like hooks

ryantmer: Just to clarify, from poking around it seems that the Mocha/Jasmine ones are before/after Suite, Hook, and Test, is that correct?

christian-bromann: Yes

ryantmer:
Updated, tried to match the ordering that was already there (which seems to be ~test-chronological).

Issue #1788, opened 2016-12-22
Handle alerts


monolithed:
## The problem

There are no any exceptions for the following commands: `alertAccept`, `alertDismiss`, `alertText`.

```js
exports.config = {
	mochaOpts: { timeout: 60 * (60 * 1000) },
};
```

```js
browser.url('https://mail.ru');
browser.alertAccept();
```

This test will be finished in an hour. At least, this situation looks like a big problem for me and other people who don't know about that.

In accordance with the WebDriver protocol these errors [should be handled](https://www.w3.org/TR/webdriver/#handling-errors)

|Error Code|HTTP Status|JSON Error Code| Description|
|-----------|-------------|-----------------|------------|
| no such alert | 400 | no such alert | An attempt was made to operate on a modal dialog when one was not open|

## The log

```
22:29:05.461 INFO - Executing: [accept alert])
22:29:11.169 WARN - Exception: timeout: cannot determine loading status
from timeout: Timed out receiving message from renderer: -0.006
  (Session info: chrome=55.0.2883.95)
  (Driver info: chromedriver=2.25.426935 (820a95b0b81d33e42712f9198c215f703412e1a1),platform=Mac OS X 10.12.2 x86_64) (WARNING: The server did not provide any stacktrace information)
Command duration or timeout: 15.01 seconds
Build info: version: '2.53.1', revision: 'a36b8b1', time: '2016-06-30 17:37:03'
System info: host: 'aabashkin.local', ip: '192.168.1.43', os.name: 'Mac OS X', os.arch: 'x86_64', os.version: '10.12.2', java.version: '1.8.0_102'
Driver info: org.openqa.selenium.chrome.ChromeDriver
[...]

PR #2016, opened 2017-04-25
changed 'returns' to 'return', added throws to alerts docs

jankcat:
## Proposed changes

- Updated instances of `@returns` to instead be `@return` in the jsdocs, as the wddoc generator only supports the `@return` tag. 
- Updated the contribution doc to reflect the change mentioned above.
- Added the `@throws` tag to the `alertDismiss`, `alertAccept`, and `alertText` protocol functions, as per #1788.

## Types of changes

## Checklist

## Further comments

These changes will make no visual difference to the API documentations without https://github.com/webdriverio/wddoc/pull/3

### DIFF
--- a/lib/commands/getViewportSize.js
+++ b/lib/commands/getViewportSize.js
@@ -21,7 +21,7 @@
  *
  * @alias browser.getViewportSize
  * @param {String} property  if "width" or "height" is set it returns only that property
- * @returns {Object}  viewport width and height of the browser
+ * @return {Object}  viewport width and height of the browser
  * @uses protocol/execute
  * @type window
  *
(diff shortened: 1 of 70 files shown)


lib/commands/getViewportSize.js
```diff
@@ -21,7 +21,7 @@
  *
  * @alias browser.getViewportSize
  * @param {String} property  if "width" or "height" is set it returns only that property
- * @returns {Object}  viewport width and height of the browser
+ * @return {Object}  viewport width and height of the browser
```
christian-bromann: actually all of these getXXX command should have `{Type|Type[]}` as return type but this is a different story

jankcat: For sure. There's probably quite a bit of doc sanitization to be done.

PR #2019, opened 2017-04-26
Update launcher.js

RayWasHere:
## Proposed changes

Added config to run specs in order provided (array in config). This is more useful for mobile apps where you must log, rather than have a before for all specs just to sign in.

## Types of changes

## Checklist

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

### DIFF
--- a/lib/launcher.js
+++ b/lib/launcher.js
@@ -257,23 +257,14 @@ class Launcher {
             if (schedulableCaps.length === 0) {
                 break
             }
-            
-            if (config.runSpecInOrder) {
-                this.startInstance(
-                    [schedulableCaps[0].specs[0]],
-                    schedulableCaps[0].caps,
-                    schedulableCaps[0].cid,
-                    schedulableCaps[0].seleniumServer
-                );
-                schedulableCaps[0].specs.shift();
-            } else {
-                this.startInstance(
-                    [schedulableCaps[0].specs.pop()],
-                    schedulableCaps[0].caps,
-                    schedulableCaps[0].cid,
-                    schedulableCaps[0].seleniumServer
-                )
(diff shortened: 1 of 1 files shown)


lib/launcher.js
```diff
@@ -257,13 +257,23 @@ class Launcher {
...
             }
-
-            this.startInstance(
-                [schedulableCaps[0].specs.pop()],
-                schedulableCaps[0].caps,
-                schedulableCaps[0].cid,
-                schedulableCaps[0].seleniumServer
-            )
+            
+            if (config.runSpecInOrder) {
```
christian-bromann: Instead of adding a new property (which if you want to add one has to be documented) make this a default behavior

lib/launcher.js
```diff
@@ -257,13 +257,23 @@ class Launcher {
...
+                this.startInstance(
+                    [schedulableCaps[0].specs[0]],
```
christian-bromann: why not using shift right here?

lib/launcher.js
RayWasHere: Making this default behaviour could potentially affect others' tests. I'll add to config doc.

christian-bromann: > Making this default behaviour could potentially affect others' tests. I'll add to config doc.

No, let's make this default behavior. I remember other people getting confused by that. Next version will be a minor version bump again (to v4.8.0) so doing this kind of change is ok

christian-bromann: Adding new configs makes setting up WebdriverIO just more complicated. We should be careful what we add to it.

PR #2029, opened 2017-04-27
Change way selector is retrieved in hasFocus

### LINKED ISSUE #2018: hasFocus can't be chained
## The problem

Certain commands do not work when chained from an element request.  (e.g. `hasFocus()`). When invoked as a chained command, they throw the exception 'Cannot read property 'selector' of undefined'. This limits the usefulness of chaining and, in particular, the Pageobject approach.

## Environment

* WebdriverIO version: 4.7.1
* Node.js version: 6.8.0
* [Standalone mode or wdio testrunner](http://webdriver.io/guide/getstarted/modes.html): wdio test runner
* if wdio testrunner, running synchronous or asynchronous tests: synchronous
* Additional wdio packages used (if applicable): ['sauce', 'selenium-standalone', 'static-server']
* Using Mocha test

## Details

The issue appears to be related to the handling of missing parameters.  This line here is the one that throws the exception:  https://github.com/webdriverio/webdriverio/blob/master/lib/commands/hasFocus.js#L30

The `value` of the last promise is `undefined` (and there's no check against that).  Switching to use `this.lastResult.selector` seems to work OK, though I don't know enough about the code base to suggest this change globally.

https://gist.github.com/webOS101/5058e80082b13e2d071d0e6446cd05c3

```
		browser.url('http://webdriver.io/');
		$('#docsearch').hasFocus();
```

webOS101:
## Proposed changes

This resolves #2018 by changing the method of retrieving the previous element(s).

## Types of changes

## Checklist

## Further comments

Could not find any way to reproduce the issue with the current test suite so no test is attached.

### DIFF
--- a/lib/commands/hasFocus.js
+++ b/lib/commands/hasFocus.js
@@ -1,6 +1,7 @@
 /**
  *
- * Return true or false if the selected DOM-element currently has focus.
+ * Return true or false if the selected DOM-element currently has focus. If the selector matches
+ * multiple elements, it will return true if one of the elements has focus.
  *
  * <example>
     :index.html
@@ -19,9 +20,9 @@
  * </example>
  *
  * @alias browser.hasFocus
- * @param {String} selector   select active element
- * @return {Boolean}         true if element has focus
- * @uses protocol/execute
+ * @param {String} selector   selector for element(s) to test for focus
+ * @return {Boolean}         true if one of the matching elements has focus
+ * @uses protocol/execute protocol/elements
  * @type state
  *
  */
@@ -39,19 +40,15 @@ let hasFocus = function (selector) {
 
         return res.value
     }).then((elements) => {
-        return this.execute(function (selectorArray) {
+        return this.execute(function (elemArray) {
             var focused = document.activeElement
             if (!focused) {
                 return false
             }
 
-            var instanceOfSelectorHasFocus = false
-            for (var i = 0; i < selectorArray.length; i++) {
-                if (focused === selectorArray[i]) {
-                    instanceOfSelectorHasFocus = true
-                }
-            }
-            return instanceOfSelectorHasFocus
+            return elemArray
+                .filter((elem) => focused === elem)
+                .length > 0
         }, elements)
     })
 
--- a/lib/commands/hasFocus.js
+++ b/lib/commands/hasFocus.js
@@ -26,23 +26,34 @@
  *
  */
 
+import { RuntimeError } from '../utils/ErrorHandler'
+
 let hasFocus = function (selector) {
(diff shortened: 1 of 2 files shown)


lib/commands/hasFocus.js
```diff
@@ -26,23 +26,34 @@
...
+            var focused = document.activeElement
+            if (!focused) {
+                return false
             }
-        }
-        return instanceOfSelectorHasFocus
-    }, finalSelector)
+
+            var instanceOfSelectorHasFocus = false
+            for (var i = 0; i < selectorArray.length; i++) {
```
christian-bromann: Shorter:
```js
return selectorArray
    .filter((elem) => focused === elem)
    .filter((hasFocus) => Boolean(hasFocus)
    .length > 0
```

christian-bromann: Also let's rename `selectorArray` to `elemArray` or something along these lines.

lib/commands/hasFocus.js
```diff
@@ -26,23 +26,34 @@
...
 
+import { RuntimeError } from '../utils/ErrorHandler'
+
 let hasFocus = function (selector) {
-    const finalSelector = selector === null ? this.lastPromise.inspect().value.selector : selector
-    let result = this.execute(function (selector) {
-        var focused = document.activeElement
-        if (!focused) {
-            return false
+    let result = this.elements(selector).then((res) => {
```
christian-bromann: So this command basically checks if one of the matching elements has focus (so it returns always a boolean value). Let's make sure to note that in the `@return` annotation

christian-bromann: Let's also let's add `protocol/elements` to the `@uses` annotation

test/fixtures/specs/element.spec.js
```diff
@@ -12,6 +12,11 @@ describe('element as first class citizen', () => {
         expect(header.getLocation('y')).to.be.equal(20)
     })
 
+    it('should be able to use hasFocus on element', () => {
+        let login = browser.element('[name=login]')
+        expect(login.hasFocus()).to.be.true
```
christian-bromann: The test page doesn't have a element like that afaik

webOS101: Interesting...  I got that from using `execute` to get the name of `document.activeElement`!

webOS101: Also, the test passes. :P

lib/commands/hasFocus.js
webOS101: Can do!  Not sure the Boolean filter step is necessary...

Issue #1701, opened 2016-11-11
isVisible() is not returning a result when the element is not visible


krobertson92:
## The problem

isVisible() is not returning a result when the element is not visible when you are operating on browser.element.
This also causes allure report generation to fail.

## Environment

* WebdriverIO version: 4.4.0
* Node.js version: v6.9.1
* [Standalone mode or wdio testrunner](http://webdriver.io/guide/getstarted/modes.html): wdio testrunner
* if wdio testrunner, running synchronous or asynchronous tests: synchronous
* Additional wdio packages used (if applicable): wdio-mocha-framework, wdio-allure-reporter

## Details

```
var elem = browser.element('//div[@id="search-layer-selector"]/div[2]')
if(!elem.isVisible()){
  //do something
}
```
**In the case where elem IS visible (works as expected)**: 
With verbose logging turned on, the above results in a COMMAND line followed by a DATA line followed by a RESULT line:
```
[14:16:39]  COMMAND	POST 	 "/wd/hub/session/e28a716e-ae88-4ea8-a1f2-961d073bc289/element"
[14:16:39]  DATA		{"using":"xpath","value":"//div[@id=\"search-layer-selector\"]/div[2]"}
[14:16:39]  RESULT		{"ELEMENT":"4"}
[14:16:39]  COMMAND	GET 	 "/wd/hub/session/e28a716e-ae88-4ea8-a1f2-961d073bc289/element/4/displayed"
[14:16:39]  DATA		{}
[14:16:39]  RESULT		true
```

**In the case where elem is NOT visible (not as expected- missing RESULT)**: 
With verbose logging turned on, the above results in a COMMAND line followed by a DATA line but no 
RESULT line:
```
[14:20:18]  COMMAND	POST 	 "/wd/hub/session/4ca5c595-9791-45ff-a8c2-ad34ac9aa16a/element"
[14:20:18]  DATA		{"using":"xpath","value":"//div[@id=\"search-layer-selector\"]/div[2]"}
[14:20:18]  COMMAND	POST 	 "/wd/hub/session/4ca5c595-9791-45ff-a8c2-ad34ac9aa16a/elements"
[14:20:18]  DATA		{"using":"xpath","value":"//div[@id=\"search-layer-selector\"]/div[2]"}
[14:20:18]  RESULT		[]
```

If I do this instead, it works correctly:
```
if(!browser.isVisible('//div[@id="search-layer-selector"]/div[2]')){
    // do something
}
```

[...]

PR #2043, opened 2017-05-03
Remove additional screenshot data params from ReporterStats

jankcat:
## Proposed changes

I have still been seeing memory pile up with large test suites that have many commands and screenshots. Not to the extent of #1978, but still getting close to the node process memory cap.

I have added a config option (`suppressReporterStats`), which defaults to false for backwards compatibility. If the option is explicitly set to `true`, the ReporterStats.output() function will still store the command `type`, but not the command `payload` in its internals. This has allowed me to run tests with many screenshots (purely visual validation type tests) and use 10% of the RAM previously used (meaning a 90% reduction).

## Types of changes

## Checklist

## Further comments

This is a sort of part two to #1993. The solution provided in #1993 greatly reduces the memory usage, but this PR allows for an even greater reduction for those who need it.

I have checked each supported reporter for backward compatibility and they seem to work as expected.

### DIFF
--- a/lib/utils/ReporterStats.js
+++ b/lib/utils/ReporterStats.js
@@ -248,6 +248,10 @@ class ReporterStats extends RunnableStats {
         // Remove the screenshot data to reduce RAM usage on the parent process
         if (type === 'screenshot') {
             runner.data = null
+        } else if (type === 'result' && runner.requestOptions.uri.path.includes('screenshot')) {
+            runner.body = null
+        } else if (type === 'aftercommand' && runner.command === 'saveDocumentScreenshot') {
+            runner.result = null;
         }
         if (ReporterStats.getIdentifier(runner) && runner.parent) {
             this.getTestStats(runner).output.push({
--- a/lib/utils/ReporterStats.js
+++ b/lib/utils/ReporterStats.js
@@ -245,26 +245,27 @@ class ReporterStats extends RunnableStats {
 
     output (type, runner) {
         runner.time = new Date()
+        const storedRunner = merge({}, runner, { clone: true })
         // Remove the screenshot data to reduce RAM usage on the parent process
         const knownScreenshotCommands = ['saveDocumentScreenshot', 'saveViewportScreenshot', 'saveElementScreenshot',
             'checkViewport', 'checkDocument', 'checkElement']
         if (type === 'screenshot') {
-            runner.data = null
+            storedRunner.data = null
         } else if (type === 'result' && runner.requestOptions.uri.path.includes('screenshot')) {
-            runner.body = null
+            storedRunner.body = null
         } else if (type === 'aftercommand' && knownScreenshotCommands.includes(runner.command)) {
-            runner.result = null
+            storedRunner.result = null
         }
         if (ReporterStats.getIdentifier(runner) && runner.parent) {
             this.getTestStats(runner).output.push({
                 type,
-                payload: runner
+                payload: storedRunner
             })
         } else {
             // Log commands, results and screenshots executed outside of a test
             this.getSpecStats(runner).output.push({
                 type,
-                payload: runner
+                payload: storedRunner
             })
         }
     }
--- a/lib/utils/ReporterStats.js
+++ b/lib/utils/ReporterStats.js
@@ -246,11 +246,13 @@ class ReporterStats extends RunnableStats {
     output (type, runner) {
         runner.time = new Date()
         // Remove the screenshot data to reduce RAM usage on the parent process
+        const knownScreenshotCommands = ['saveDocumentScreenshot', 'saveViewportScreenshot', 'saveElementScreenshot',
+            'checkViewport', 'checkDocument', 'checkElement']
         if (type === 'screenshot') {
             runner.data = null
         } else if (type === 'result' && runner.requestOptions.uri.path.includes('screenshot')) {
             runner.body = null
-        } else if (type === 'aftercommand' && runner.command === 'saveDocumentScreenshot') {
+        } else if (type === 'aftercommand' && knownScreenshotCommands.includes(runner.command)) {
             runner.result = null
         }
         if (ReporterStats.getIdentifier(runner) && runner.parent) {
--- a/lib/utils/ReporterStats.js
+++ b/lib/utils/ReporterStats.js
@@ -253,14 +253,14 @@ class ReporterStats extends RunnableStats {
             this.getTestStats(runner).output.push({
                 type,
                 // Suppress the level of detail if we need to conserve RAM
-                payload: (config.suppressReporterStats) ? null : runner,
+                payload: (config.suppressReporterStats) ? null : runner
             })
         } else {
             // Log commands, results and screenshots executed outside of a test
             this.getSpecStats(runner).output.push({
                 type,
                 // Suppress the level of detail if we need to conserve RAM
-                payload: (config.suppressReporterStats) ? null : runner,
+                payload: (config.suppressReporterStats) ? null : runner
             })
         }
     }
--- a/lib/utils/ReporterStats.js
+++ b/lib/utils/ReporterStats.js
@@ -78,10 +78,9 @@ class HookStats extends RunnableStats {
 }
 
 class ReporterStats extends RunnableStats {
-    constructor (config) {
+    constructor () {
         super('base')
         this.runners = {}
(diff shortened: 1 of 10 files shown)


jankcat:
[Here on the native-screenshot branch](https://github.com/jankcat/wdio-example-leak/tree/native-screenshot), you can see the issue. It is linked to my fork/PR'd webdriver, so you will need to link/build the fork to experience the fix.

Memory with default options, after 20ish tests have passed:
![https://github.com/jankcat/wdio-example-leak/raw/native-screenshot/docs/not%20supressed/Memory%20Growing.png]

Memory with suppression option, after 20ish tests have passed:
![https://github.com/jankcat/wdio-example-leak/raw/native-screenshot/docs/suppressed/After%20the%20Same%20number%20of%20tests.png]

christian-bromann:
@jankcat instead of suppressing these, can we just remove the whole stats object after the test has finished? I worried about how the reported work when we do that.

jankcat:
The info hits the reporters before it gets to the ReporterStats class, as it goes Reporters > BaseReporter Class > ReporterStats class, so it shouldn't affect the way reporters work. 

I could look at removing it post-testcase, but any higher level (post-suite, for example) would still cause these issues on larger test suites.

christian-bromann:
@jankcat I think we keep this info around so you can create a report in afterTest or when the Selenium session ends. So I was thinking what would happen if we delete the stats somewhere [here](https://github.com/webdriverio/webdriverio/blob/master/lib/utils/BaseReporter.js#L147)?

jankcat:
@christian-bromann when is the `end` event you linked to fired? I think the problem is that the screenshot data is still growing somewhere, even though the `screenshot` event type has the screenshot data trimmed out.

I will investigate this further to see if I am correct and then loop back

jankcat:
@christian-bromann I have looked into this some more:

- I removed the screenshot data from the `result` event (when the command was a `screenshot` command), as it existed there as well, and still see the memory grow (not as fast, but still at unsustainable numbers) 
- I can't seem to find other instances where this screenshot data exists.

I like the idea of being able to kill the data instead of storing it, especially when using wdio mainly for its driving/running capabilities or where the user doesn't need detailed reporting beyond the capabilities of the wdio-*-reporter system.

jankcat:
Wait, I figured it out. I watched every event and its size, and have found what I think are all the raw screenshot data parameters. I will update momentarily.

jankcat:
@christian-bromann I have found the remaining screenshot data params, and removed them. I reverted my other changes, making this a much simpler fix.

A few notes: 
- The screenshots were also logged in the result/response from the webdriver
- The screenshots were also logged in `aftercommand`, which, in some cases (like the reproducible example above) was using an additional 27~mb of RAM per page!
- Other events/details are still logged, but based on my estimation, it would take 200k+ other events to get anywhere near the nodeJS default RAM cap.

lib/utils/ReporterStats.js
```diff
@@ -248,6 +248,10 @@ class ReporterStats extends RunnableStats {
         // Remove the screenshot data to reduce RAM usage on the parent process
         if (type === 'screenshot') {
             runner.data = null
+        } else if (type === 'result' && runner.requestOptions.uri.path.includes('screenshot')) {
+            runner.body = null
+        } else if (type === 'aftercommand' && runner.command === 'saveDocumentScreenshot') {
```
christian-bromann: This is not a WebdriverIO screenshot. Any chance we can restrict this to API commands? Because in case that `wdio-screenshot` renames their API commands this problem will reoccur.

christian-bromann:
@jankcat this is awesome

nice job

lib/utils/ReporterStats.js
christian-bromann: Also when we remove the data at this point is there any chance that reporter like `wdio-allure-reporter` are getting restricted (e.g. not being able to use that screenshots anymore)?

christian-bromann:
Just got in a new comment by @dhoulker:

> I'm using the webdriver visual regression service, and have an issue very similar to the one reported (and fixed) here: webdriverio/webdriverio#1993 However this fix doesn't solve the problem when taking screenshots using the visual regression service. The base64 image string still stays in the body of the reporterstats object and causes webdriver to crash. Anyone else had this issue?

dhoulker:
Thanks for Looping me in Christian, i've had exactly the issue above and wasn't sure how to approach it. I'm using wdio-visual-regression-service which in turn uses wdio-screenshot. 

As soon as i run tests at any scale it falls over because of the memory leak. 

`saveScreeshot.js`(where the issue has been fixed) does this:

`this.emit('screenshot', {data: res.value, filename})`

However wdio-screenshot does not, instead we end up with the `runner:result` event being fired. 

The `body.value` then contains the image as a base64 string.

I was looking to fix this in wdio-screenshot, but haven't been able to so far.

christian-bromann:
> I was looking to fix this in wdio-screenshot, but haven't been able to so far.

@dhoulker I think it has to get fixed here

jankcat:
### Regarding Reporters:
I don't believe this should affect the reporters. But, if this does surface as an issue with reporters: 
1. I could do a deep clone on the runner object and make all changes to that instead (in case reporters are referencing the entire object instead of pulling out what they need). 
2. We could revisit the idea of a config option, like a  `logScreenshots` flag that defaults to `true`?

### Regarding Brittleness/Completeness
I have been pondering how we can remove all the unwanted screenshot data, while maintaining stability, as @christian-bromann has a good point in that `wdio-screenshot` could decide to change their API and it is out of our control:
1. We could leave it as is, hope they do not change the commands or address it when they do, and add the other two `wdio-screenshot` commands (`saveElementScreenshot`, `saveViewportScreenshot`). This will make it work now, is the fastest for today, but is less stable long term because the aforementioned reasons.
2. We could change it to remove the `runner.result` of any command with `screenshot` in the command name. This is also very little work, is less likely to break with API changes, and would catch other plugins and custom commands. The fact that it would catch custom commands might be an issue for some though, and so it has unforeseen consequences.
3. We could look at making the `ReporterStats` logging of `runner.result` on Custom Commands configurable, but based on the [information currently received](https://gist.github.com/jankcat/25cb3fd60a24739eb1af76e6adf3e2e4) in the `ReportersStats` `output` function, this would mean a few chunks of work:
    - Since commands can return whatever they want in the `runner.result` parameter, add a documented and known parameter (lets say `doNotStore` or something). If this parameter is present and `true`, `ReporterStats` nulls out the `runner.result` before storing. If the parameter is not present or `true`, nothing changes for backward compatibility.
    - After that is implemented, we can PR `wdio-screenshot`. For backward compatibility, we can add a config option to `wdio-screenshot`. If the config option is set to `true`, then the output of wdio-screenshot comes back as an `object` with that `doNotStore` parameter included. If the config option is not `true` then nothing changes for backwards compatibility. 
    - Since everything defaults to current behaviour, I think this solution would be fully backwards compatible, and it would leave the responsibility of not destroying RAM with screenshots on plugin developers

Thoughts?

jankcat:
@dhoulker do you see the screenshot data in the `aftercommand` event from `wdio-visual-regression-service` as well?

jankcat:
@christian-bromann unless you have any comments, I think this is good to go.

jankcat:
@christian-bromann when do you think you'll have time to look at this?

jankcat:
@christian-bromann Can you review this? I don't like having to use a local npm link or a forked install to use webdriverio on my visual test suites 😜

jankcat:
Any time to review? @christian-bromann

michal-boruczkowski:
Nice work guys!

I spend whole day trying to track this issue... I think im experiencing bit different aspect of it but I'm sure this PR fixes it. I added comment here:
https://github.com/webdriverio/webdriverio/issues/1701#issuecomment-320237026

jankcat:
Hey @michal-boruczkowski, since I was dependant on this PR and it still hasn't been reviewed (*cough* @christian-bromann *cough*), I have been using [my own fork](https://github.com/jankcat/webdriverio). The downsides are 1) it doesn't have any commits since my commits in it, and 2) it requires building after doing `npm i` as the compiled files are not in the repo. 

I think @christian-bromann has been busy doing his Masters or something tho, so thats fair.

michal-boruczkowski:
Thanks! I think I'll use it for now

PR #2055, opened 2017-05-09
Reverting my previous edit

mikeapr4:
Although the `--debug-brk` switch works, due to multiple forked processes, there may be multiple paused processes. As the V8 debug URL isn't outputted for these processes, they remain invisibly paused, without a way to resume execution. This switch cannot unfortunately be used. Apologies

### DIFF
--- a/docs/guide/testrunner/debugging.md
+++ b/docs/guide/testrunner/debugging.md
@@ -95,10 +95,7 @@ To start debugging, open the following URL in Chrome:
 ```
 You'll want to open that url, which will attach the debugger.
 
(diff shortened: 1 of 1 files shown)


christian-bromann:
@mikeapr4 even if you set maxInstances to 1? I think this is somehow a requirement for using this right?

mikeapr4:
Yeah, I've got maxInstances as 1. In fact, it's not to do with further spawned processes, because with `--debug-brk` off, the same inspect URL works later during the tests. Maybe something times out somewhere along the way.

PR #2056, opened 2017-05-09
Added option to disable deprecation warnings

jankcat:
## Proposed changes

I added a config option to enable/disable deprecation warnings. The config option is `deprecationWarnings`, and is set to `true` by default. This means that by default, deprecation warnings will be displayed. Setting `deprecationWarnings` to `false` will cause deprecation warnings to not be displayed. Just like the simulations.

## Types of changes

## Checklist

## Further comments

To recap... `deprecationWarnings` option...

### DIFF
--- a/docs/guide/testrunner/gettingstarted.md
+++ b/docs/guide/testrunner/gettingstarted.md
@@ -32,7 +32,6 @@ Options:
   --watch               watch specs for changes
   --logLevel, -l        level of logging verbosity (default: silent)
   --coloredLogs, -c     if true enables colors for log output (default: true)
-  --deprecationWarnings if true use of deprecated commands will output a warning (default: true)
   --bail                stop test runner after specific amount of tests have failed (default: 0 - don't bail)
   --screenshotPath, -s  saves a screenshot to a given path if a command fails
   --baseUrl, -b         shorten url command calls by setting a base url
--- a/docs/guide/testrunner/gettingstarted.md
+++ b/docs/guide/testrunner/gettingstarted.md
@@ -32,6 +32,7 @@ Options:
   --watch               watch specs for changes
   --logLevel, -l        level of logging verbosity (default: silent)
   --coloredLogs, -c     if true enables colors for log output (default: true)
+  --deprecationWarnings if true use of deprecated commands will output a warning (default: true)
   --bail                stop test runner after specific amount of tests have failed (default: 0 - don't bail)
   --screenshotPath, -s  saves a screenshot to a given path if a command fails
   --baseUrl, -b         shorten url command calls by setting a base url
--- a/examples/wdio/wdio.mocha.conf.js
+++ b/examples/wdio/wdio.mocha.conf.js
@@ -56,5 +56,4 @@ exports.config = {
     onComplete: function() {
         console.log('that\'s it');
     }
-    
 };
--- a/examples/wdio/wdio.mocha.conf.js
+++ b/examples/wdio/wdio.mocha.conf.js
@@ -33,7 +33,6 @@ exports.config = {
      */
     logLevel: 'silent',
     coloredLogs: true,
-    deprecationWarnings: true,
     screenshotPath: 'shots',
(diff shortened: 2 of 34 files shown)


examples/wdio/wdio.mocha.conf.js
```diff
@@ -56,5 +57,5 @@ exports.config = {
     onComplete: function() {
         console.log('that\'s it');
     }
-
-};
\ No newline at end of file
+
```
christian-bromann: I guess this is not needed 😉

docs/guide/testrunner/gettingstarted.md
```diff
@@ -32,6 +32,7 @@ Options:
   --watch               watch specs for changes
   --logLevel, -l        level of logging verbosity (default: silent)
   --coloredLogs, -c     if true enables colors for log output (default: true)
+  --deprecationWarnings if true use of deprecated commands will output a warning (default: true)
```
christian-bromann: if you want to add it here you need to add support for it in `cli.js`. If you ask me .. not needed

lib/helpers/deprecationWarning.js
```diff
@@ -1,4 +1,5 @@
 export default function deprecate (commandName) {
+    if (!global.browser.options.deprecationWarnings) return
```
christian-bromann: This wouldn't work in standalone mode. Let's call all `deprecate` functions with `this` context and access that option in that way.

christian-bromann: or apply the option as argument

docs/guide/testrunner/gettingstarted.md
jankcat: Good call. Will fix.

lib/helpers/deprecationWarning.js
jankcat: Oh shoot. Thanks for the tip. Will make it an argument then.

examples/wdio/wdio.mocha.conf.js
jankcat: Haha WebStorm automatically makes sure files end in a newline... I can take the line out after-the-fact on github if you want

jankcat: Oh wait, `insert_final_newline = true` is part of the `.editorconfig`, so this is intentional I suppose.

jankcat:
@christian-bromann requested changes made

jankcat:
To show it works in standalone:
The [moveTo.js](https://github.com/webdriverio/webdriverio/blob/master/test/spec/unit/moveTo.js) unit test uses a deprecated function in standalone mode.
screen shot 2017-05-09 at 11 20 44 am

When I update this line:
```
const client = remote({})
```
To this:
```
const client = remote({deprecationWarnings: false})
```
The UT output looks as expected:
screen shot 2017-05-09 at 11 19 02 am

webOS101:
Just food for thought -- would it make sense to pass an array of the methods to suppress instead, so that new deprecation warnings don't get stifled unexpectedly?

examples/wdio/wdio.mocha.conf.js
christian-bromann: I was talking about the unnecessary line (59). Let's remove this

lib/protocol/context.js
```diff
@@ -25,6 +25,6 @@ export default function context (id) {
         data.name = id
     }
 
-    deprecate('context')
+    deprecate('context', this.options)
```
christian-bromann: Can we remove the deprecate warning here? This is an Appium command and won't be removed anytime soon.

lib/protocol/contexts.js
```diff
@@ -12,6 +12,6 @@
 import deprecate from '../helpers/deprecationWarning'
 
 export default function contexts () {
-    deprecate('contexts')
+    deprecate('contexts', this.options)
```
christian-bromann: same here

jankcat:
@christian-bromann I have made the requested changes.

PR #2089, opened 2017-05-18
Update webdriverio.multiremote.chat.js

snackycracky:
- the socket.io url is outdated. 
- also the this.timeout = 9999 is not supported anymore by mocha

### DIFF
--- a/examples/multiremote/webdriverio.multiremote.chat.js
+++ b/examples/multiremote/webdriverio.multiremote.chat.js
@@ -21,11 +21,10 @@ describe('multiremote example', function() {
         return matrix.init().url('https://socketio-chat.now.sh/');
     });
 
-    it('should login the browser', function(done) {
+    it('should login the browser', function() {
         browserA.setValue('.usernameInput', 'Browser A');
         browserB.setValue('.usernameInput', 'Browser B');
-        matrix.sync().call(done);
-        return Promise.resolve();
+        return matrix.sync();
     });
 
     it('should submit the login form', function() {
@@ -40,7 +39,7 @@ describe('multiremote example', function() {
 
     it('should read the message in browserB', function() {
         return browserB
-            .pause(1000)
+            .pause(200)
             .getText('.messageBody*=My name is').then(function(message) {
                 var name = message.slice(11);
                 return browserB.setValue('.inputMessage', 'Hello ' + name + '! How are you today?').keys('Enter');
--- a/examples/multiremote/webdriverio.multiremote.chat.js
+++ b/examples/multiremote/webdriverio.multiremote.chat.js
@@ -25,6 +25,7 @@ describe('multiremote example', function() {
         browserA.setValue('.usernameInput', 'Browser A');
         browserB.setValue('.usernameInput', 'Browser B');
         matrix.sync().call(done);
+        return Promise.resolve();
     });
 
     it('should submit the login form', function() {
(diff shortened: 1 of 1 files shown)


examples/multiremote/webdriverio.multiremote.chat.js
```diff
@@ -17,16 +17,14 @@ var WebdriverIO = require('../../build'),
...
-
     it('should open chat application', function() {
-        return matrix.init().url('http://chat.socket.io/');
+        return matrix.init().url('https://socketio-chat.now.sh/');
     });
 
     it('should login the browser', function(done) {
         browserA.setValue('.usernameInput', 'Browser A');
         browserB.setValue('.usernameInput', 'Browser B');
-        return matrix.sync().call(done);
```
snackycracky: this was throwing an error. `Either return or call done`

examples/multiremote/webdriverio.multiremote.chat.js
```diff
@@ -17,16 +17,14 @@ var WebdriverIO = require('../../build'),
 
 describe('multiremote example', function() {
 
-    this.timeout = 99999999;
-
     it('should open chat application', function() {
-        return matrix.init().url('http://chat.socket.io/');
+        return matrix.init().url('https://socketio-chat.now.sh/');
```
snackycracky: this is the chat-iframe on the official socket.io chat-demo site.

examples/multiremote/webdriverio.multiremote.chat.js
christian-bromann: @snackycracky let's rather return the promise

snackycracky: the only thing that would work is:
```
... 
 matrix.sync().call(done);
 return Promise.resolve();
```
it's a little cleaner

examples/multiremote/webdriverio.multiremote.chat.js
```diff
@@ -17,16 +17,15 @@ var WebdriverIO = require('../../build'),
...
     });
 
     it('should login the browser', function(done) {
         browserA.setValue('.usernameInput', 'Browser A');
         browserB.setValue('.usernameInput', 'Browser B');
-        return matrix.sync().call(done);
+        matrix.sync().call(done);
+        return Promise.resolve();
```
snackycracky: here is the return statement.

snackycracky:
As I said, that’s not working, it just halts, and won’t ever continue. I am not sure why.

christian-bromann:
@snackycracky this is a bug then, let's remove `matrix.sync()` then and merge this PR

snackycracky:
It's not working, I tried multiple things. Just 
```
matrix.sync().call(done); // without return
```
or 
```
matrix.sync().call(done);
return Promise.resolve()
```
is actually passing the tests...

Completelt removing `matric.sync();`
or doing like this is not working:
```
return Promise.resolve().then(() => {
            return browserA.setValue('.usernameInput', 'Browser A'); 
        }).then(() => {
            return browserB.setValue('.usernameInput', 'Browser B'); 
        }).then(done)
 
```

christian-bromann:
Then there is a bug in there. Let's remove that change then and create a different issue for this.

snackycracky:
Should I remove 'return Promise.resolve()'? Or what change do you mean ? I would rather keep the change in favor for working examples and create an issue. Its nice when you clone a repo and the included examples are actually running.

christian-bromann:
> Should I remove 'return Promise.resolve()'?

Yes!

> I would rather keep the change in favor for working examples and create an issue.

To make it work, just remove the `.call(done)` and the `done` parameter in the function head. The `Promise.resolve()` is just confusing adds no functionality.

snackycracky:
ok done. Works. :shipit:

christian-bromann:
@snackycracky so this works how it is right now? If so there is no bug then 😃

PR #2098, opened 2017-05-23
Accept a list of specs to run piped from stdin

sha1sum:
## Proposed changes

This does not alter the previous operation of the CLI `wdio` utility, but instead adds the ability to pass a list of spec filenames to be run by piping them via stdin (from a utility such as `grep` or `find` in *nix-based systems).

I did not see any issues linked to this type of behavior, but if I happened to miss one, please forgive me.

## Types of changes

This simply adds detection of piped input from `stdin` and processes the input as if it were a list of specs to be run if present. Documentation has also been updated.

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works &mdash; _**NOTE:** This is somewhat hard to do with the CLI, but I've definitely tested the CLI portion._

## Further comments

Piping lists of files to tools is quite familiar and standard for most OS X and other *nix-based shell users. Given that `wdio` does not pre-process files and that running multiple `wdio` commands with a single `--spec` argument via `xargs` or the like is a little silly, I figured this would be a nice addition.


sha1sum:
Note that this build is failing due to an integration that I cannot fix and therefore cannot correct.

christian-bromann:
> Note that this build is failing due to an integration that I cannot fix and therefore cannot correct.

Noticed. Sorry about that

mkj28:
In this context - is it expected that you can pass comma-separated list to --spec?
wdio --spec=src/spec/test1.js,src/spec/test2.js etc.

PR #2137, opened 2017-06-09
Temporary improve promises speed by using bluebird-q

seth2810:
## Proposed changes

As a [benchmarks](http://bluebirdjs.com/docs/benchmarks.html) shows, bluebird module is much faster than current q. But we can't move quickly to it, because [we need a deferred objects](https://github.com/webdriverio/webdriverio/blob/master/lib/webdriverio.js#L315). But, as a temporary solution, we can use drop-in replacement module named [bluebird-q](https://github.com/petkaantonov/bluebird-q), that provide same API as q, but use bluebird power under the hood.

## Types of changes

Replace all occurrences of q to bluebird-q.

## Checklist

## Further comments

### DIFF
--- a/lib/helpers/errorHandler.js
+++ b/lib/helpers/errorHandler.js
@@ -1,4 +1,4 @@
-import q from 'q'
+import q from 'bluebird-q'
 
 export default [
     /**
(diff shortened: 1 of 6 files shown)


christian-bromann:
@seth2810 do you have any performance metrics proofing that using Bluebird here makes tests faster?

PR #2138, opened 2017-06-11
Documentation - pageobjects.md Fix ES6 Page Object

mossnz:
Fixing ES6 page.js error on exported module

## Proposed changes

The ES6 sample page objects provided in the documentation `pageobjects.md` do not work in Node 8.0.0

The export of `page.js` will cause an error when extended in the sample `login.js` of
`TypeError: Class extends value #<Page> is not a constructor or null`

## Types of changes

## Checklist

### DIFF
--- a/docs/guide/testrunner/pageobjects.md
+++ b/docs/guide/testrunner/pageobjects.md
@@ -44,7 +44,7 @@ class Page {
 	}
 
 }
-module.exports = new Page();
+module.exports = Page;
 ```
 
 We will always export an instance of a page object and never create that instance in the test. Since we are writing end to end tests we always see the page as a stateless construct the same way as each http request is a stateless construct. Sure, the browser can carry session information and therefore can display different pages based on different sessions, but this shouldn't be reflected within a page object. These state changes should emerge from your actual tests.


christian-bromann:
You are right, this is indeed not correct. Can we also change the non es6 example in [line 30](https://github.com/webdriverio/webdriverio/pull/2138/files#diff-1ff639a6e3fd0c1a405d827d4e5fb2d1R30)? Thanks!

mossnz:
@christian-bromann 

If we change [line 30](https://github.com/webdriverio/webdriverio/pull/2138/files#diff-1ff639a6e3fd0c1a405d827d4e5fb2d1R30) in the non ES6 example,

The non ES6 `login` page will still load without an error, but when attempting to use [line 71](https://github.com/webdriverio/webdriverio/pull/2138/files#diff-1ff639a6e3fd0c1a405d827d4e5fb2d1L71), `login.open()` an error `TypeError: Cannot read property 'call' of undefined` will occur.

It seems that the `page.js` exports need to be different for differences in ES6 classes and Object.create

christian-bromann:
@mossnz is it possible that Node v8 behaves different than lower versions?

ianrenauld:
I agree with the change for the ES6 version and line 30 did work for me when I was using non-ES6, but then again I never tried it on Node v8.

One more thing that I feel we should change on this document is to add `LoginPage.flash.waitForVisible();` right after both `LoginPage.submit();` lines in the login.spec.js file.  Otherwise for me the example did not work (the "flash" message takes a little while to show up after submitting the form).

It's always best to have actual working examples so beginners get going on the right foot!

christian-bromann:
thanks @mossnz for the PR. I added some additional changes to it (including the suggestion of @ianrenauld).

Issue #2105, opened 2017-05-25
Unable to use appium locators with id


APshenkin:
## The problem

Need to modify regex for id validation. Now if there are `: . /` symbols in locator that starts with `#` WebdriverIO doesn't understand that it's `ID` locator
https://github.com/webdriverio/webdriverio/blob/master/lib/helpers/findElementStrategy.js#L41
Now:
^#-?[_a-zA-Z]+[_a-zA-Z0-9-]*$
Should be:
^#-?[_a-zA-Z]+[_a-zA-Z0-9.:/-]*$
## Environment

* WebdriverIO version: 4.7
* Node.js version: 6.9.5
* [Standalone mode or wdio testrunner](http://webdriver.io/guide/getstarted/modes.html): Standalone

## Code To Reproduce Issue [ Good To Have ]
here is locator to test regex `#io.selendroid.testapp:id/buttonStartWebview`

christian-bromann:
fixed by #2156

PR #2156, opened 2017-06-23
change regex for id

APshenkin:
https://github.com/webdriverio/webdriverio/issues/2105

### DIFF
--- a/docs/guide/services/iedriver.md
+++ b/docs/guide/services/iedriver.md
@@ -1,0 +1,79 @@
+name: iedriver
+category: services
+tags: guide
+index: 12
+title: WebdriverIO - IEDriver Service
+---
+
+WDIO IEDriver Service
+================================
+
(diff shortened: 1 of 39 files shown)


christian-bromann:
Can you please add a unit test for this?

APshenkin:
@christian-bromann I added test for this change. After this change you are not be able to locate pseudo class elements via css selector, but I'm not sure that Webdriver can locate pseudo class elements itself https://stackoverflow.com/a/37355292

DavertMik:
This change is important for Appium tests. This simplifies locators from

```
//android.widget.TextView[@resource-id="io.selendroid.testapp:id/label_email_data"]
```
to 

```
#io.selendroid.testapp:id/label_email_data
```

APshenkin:
Now in this PR https://github.com/webdriverio/webdriverio/pull/2161

PR #2187, opened 2017-07-10
Use local npm dependencies rather than the global ones

olivierbeaulieu:
## Proposed changes

When attempting to build webdriverio locally, you'll get a bunch of errors because of potentially missing global packages (run-s, run-p, etc.).

They are being referenced globally in the package.json, even though they are installed as local dependencies in node_modules, they're not used.

This PR changes that by using the locally installed dependencies rather than the global ones. This means that anyone can make a build with a simple `npm install && npm run build`, with no errors related to missing dependencies.

## Types of changes

## 
Checklist


christian-bromann:
This is no necessary. NPM automatically looks into local node_modules folder first.

PR #2229, opened 2017-08-07
handle a case when error is created with details

just-boris:
## Proposed changes

There is [a line](https://github.com/webdriverio/webdriverio/blob/master/lib/utils/ErrorHandler.js#L89) that creates and ErrorHandler with 3 arguments. Currently this case is not handled, so we are receiving an empty Error object at the end.

## Types of changes

## Checklist

## Further comments

Here is a code to reproduce the issue

```js
const webdriverio = require('webdriverio');
const browser = webdriverio.remote();

browser
    .init()
    .url('https://google.com')
    .setValue('[name="q"]', undefined) // I give invalid value and expect to see an error message, that `setValue` has for me
    .catch(error => console.log(error)) // log error message
    .end();
```

With the latest webdriver.io version you will see an empty `Error` without any message. After that change you will see `Error: number or type of arguments don't agree with setValue command` as it was expected.

### DIFF
--- a/lib/utils/ErrorHandler.js
+++ b/lib/utils/ErrorHandler.js
@@ -17,7 +17,8 @@ class ErrorHandler extends Error {
             if (msg === 7 && details) {
                 this.message = `${this.message.slice(0, -1)} ("${details}").`
             }
-        } else if (arguments.length === 2) {
+        } else if (arguments.length > 1) {
+            this.details = details
             this.message = msg
             this.type = type
         } else if (arguments.length === 1) {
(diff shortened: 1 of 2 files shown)


just-boris:
> Is there any chance to write a unit test for this?

Yes, it is possible. Will update the PR later. So far I just have submitted this via Github online editor in order not to forget about this issue :)

just-boris:
Done.

Don't know, why some tests have failed. Is it because of my change?

PR #2237, opened 2017-08-08
show related imports of gulp plugin code example

NicholasAaronBrady:
minor, but for some reason it took me a couple minutes before I face-palmed

## Proposed changes

The code sample used in the gulp plugin does not explicitly note where the `webdriver` variable comes from which can lead to some confusion. It is relatively obvious, but it doesn't hurt to be explicit in the docs.

I deleted the rest of the automated form because it isn't really applicable for such a small docs change. I don't think this is a completely necessary change, but if you agree it might make the docs more clear, here is a PR!

### DIFF
--- a/docs/guide/plugins/gulp-webdriver.md
+++ b/docs/guide/plugins/gulp-webdriver.md
@@ -21,6 +21,8 @@ npm install gulp-webdriver --save-dev
 You can run WebdriverIO locally running this simple task:
 
 ```js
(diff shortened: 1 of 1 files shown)


NicholasAaronBrady:
ah, also, the link "Improve this doc" is broken on that page: http://webdriver.io/guide/plugins/gulp-webdriver.html

NicholasAaronBrady:
not at all :)

PR #2247, opened 2017-08-18
return readable errors in debug mode

lgandecki:
## Proposed changes

Before:
> browser.waitForExist(".we")
Error

Now:

> browser.waitForExist('.we')
Thrown: Error: element (".we") still not existing after 1000ms

Makes for a much easier debugging experience - otherwise it's hard to guess what's wrong. 
I also removed the stack from the message - I can't think of a situation where it would be helpful, but it surely messes up the debug. Another option would be to decrease the number of stack lines displayed.

## Types of changes

## Checklist

I don't think those internals are tested

### DIFF
--- a/lib/commands/debug.js
+++ b/lib/commands/debug.js
@@ -91,7 +91,9 @@ let debug = function (commandTimeout = 5000, enableStdout, enableLogging) {
         }, (e) => {
             commandIsRunning = false
             clearTimeout(timeout)
-            return callback(e)
+            const commandError = new Error(e.message)
+            delete commandError.stack
+            return callback(commandError)
         })
     }
 


lgandecki:
One more thing - I'm testing on node 8, the proper behavior might have worked on older versions, not sure if repl API's changed.

christian-bromann:
> One more thing - I'm testing on node 8, the proper behavior might have worked on older versions, not sure if repl API's changed.

@lgandecki so this is working fine for me on node v6. Can you ensure that the behavior is consistent across the recent node versions?

lgandecki:
Sure

$ node -v 
v6.1.0

> browser.waitForExist('ssdfds')
Error: element ("ssdfds") still not existing after 1000ms

Without my change (but still on node 6.1.0) it gave me the same exact message.

christian-bromann:
> Without my change (but still on node 6.1.0) it gave me the same exact message.

I was curious if the behavior is equal with your change.

lgandecki:
hmm... Not sure I understand, the behavior of node 6 doesn't change with my change.
The behavior of node 8 without my change is:
Error,
with my change on node 8 it is the same as with/without my change on node 6

PR #2257, opened 2017-08-25
Update transferpromises.md

tdickers:
Provide an example of how to use `executeAsync` with Chai promises.

## Proposed changes

Documentation update for execute + Promises

## Types of changes

Documentation

## Checklist

### DIFF
--- a/docs/guide/usage/transferpromises.md
+++ b/docs/guide/usage/transferpromises.md
@@ -35,6 +35,15 @@ describe('my app', function() {
             .isVisible('#status_message').should.eventually.be.true
             .getText('#status_message').should.eventually.be.equal('Message sent!');
     });
+    
+    it('should return an async function call', function() {
+        return browser.executeAsync(function(done) {
+          setTimeout(function() {
+            done(true);
+          }, 5000);
+        })
+      ).should.eventually.have.property('value', true);
+    });
 });
 ```
 


tdickers:
I'm not sure what you mean about it not working; the following example fails, and a second argument is shown at http://chaijs.com/api/bdd/#method_property.

The problem I was trying to address was some initial confusion I had using `executeAsync` with Chai promises, since the value isn't returned directory.

```
"use strict";
const chai = require("chai"),
  chaiAsPromised = require("chai-as-promised"),
  expect = chai.expect;

chai.use(chaiAsPromised);
chai.Should();

let browser = require("webdriverio").remote({
  desiredCapabilities: {
    browserName: "firefox"
  }
});

describe("Load page using wdio with sync disabled and execute js using promises", function() {
  this.timeout(120000);

  before(() => {
    chaiAsPromised.transferPromiseness = browser.transferPromiseness;
    return browser.init();
  });

  it("Should load page and invoke async", function() {
    return browser
      .url("https://www.google.com")
      .executeAsync(function(done) {
        setTimeout(function() {
          done(true);
        }, 5000);
      })
      .should.eventually.have.property("value", false);
  });
});
```

christian-bromann:
What I meant was that execAsync always has a `value` property according to the JSONWireProtocol. So this example is a little bit simple (like `Promise([]).should.eventually.have.property("value")`)

PR #2200, opened 2017-07-18
Fix "debug" in new version node.js

hermanho:
## Proposed changes

[//]: # (Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue.)

Bug fixing 
https://github.com/webdriverio/webdriverio/issues/1640#issuecomment-292419434
https://github.com/webdriverio/webdriverio/issues/2194

References:
https://nodejs.org/en/docs/guides/debugging-getting-started/#legacy-debugger
https://code.visualstudio.com/docs/nodejs/nodejs-debugging#_supported-nodelike-runtimes

## Types of changes

[//]: # (What types of changes does your code introduce to WebdriverIO?)
[//]: # (_Put an `x` in the boxes that apply_)

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

[//]: # (_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._)

- [ ] I have read the [CONTRIBUTING](https://github.com/webdriverio/webdriverio/blob/master/CONTRIBUTING.md) doc
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

## Further comments

[//]: # (If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...)

Issue #2194, opened 2017-07-13
Not possible to use `debug: true` in WDIO config


NicolasSiver:
## The problem

It is not possible to use `debug:true` flag in the configuration.

## Environment

* WebdriverIO version: `4.8.0`
* Node.js version: `8.1.4`
* [Standalone mode or wdio testrunner](http://webdriver.io/guide/getstarted/modes.html): testrunner
* if wdio testrunner, running synchronous or asynchronous tests: sync
* Additional wdio packages used (if applicable): `selenium-standalone` _service_ via Safari WebDriver.

## Details

Error: 

> (node:41307) [DEP0062] DeprecationWarning: `node --debug` and `node --debug-brk` are invalid. Please use `node --inspect` or `node --inspect-brk` instead.

christian-bromann:
Please use a lower node version in the meantime.

christian-bromann:
This will be fixed with #2200. Let's continue this convo there.

Issue #1640, opened 2016-10-11
Debugging with node-inspector fails


mdotasia:
## The problem

I'm trying to get debugging with node-inspector work as described here: http://webdriver.io/guide/testrunner/debugging.html. As expected I get message `Debugger listening on [::]:5859`. Then I run

```
node-inspector --debug-port 5859 --no-preload
```

and open

```
http://127.0.0.1:8080/?port=5859
```

in Chrome and see this error in command-line:

```
Node Inspector v0.12.8
Visit http://127.0.0.1:8080/?port=5859 to start debugging.
/usr/local/lib/node_modules/node-inspector/lib/InjectorClient.js:111
      cb(error, NM[0].ref);
                     ^

TypeError: Cannot read property 'ref' of undefined
    at InjectorClient.<anonymous> (/usr/local/lib/node_modules/node-inspector/lib/InjectorClient.js:111:22)
    at /usr/local/lib/node_modules/node-inspector/lib/DebuggerClient.js:121:7
    at Object.value (/usr/local/lib/node_modules/node-inspector/lib/callback.js:23:20)
    at Debugger._processResponse (/usr/local/lib/node_modules/node-inspector/lib/debugger.js:95:21)
    at Protocol.execute (_debugger.js:121:14)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:188:7)
    at readableAddChunk (_stream_readable.js:176:18)
    at Socket.Readable.push (_stream_readable.js:134:10)
    at TCP.onread (net.js:543:20)
```

Uninstall and reinstall node-inspector did not help.
## Environment
- WebdriverIO version: 4.2.16
- Node.js version: 6.7.0
- [Standalone mode or wdio testrunner](http://webdriver.io/guide/getstarted/modes.html): wdio synchronous tests with mocha
- Additional wdio packages used (if applicable): selenium-standalone
- Chrome: Version 53.0.2785.143 (64-bit)
- OS: macOS 10.12
## Link to Selenium/WebdriverIO logs

Logging webdriverio in `error` mode yields empty log.

mdotasia:
This one seems to be the cause: https://github.com/node-inspector/node-inspector/issues/905.

mdotasia:
Wondering if debugging can be rewritten using `node --inspect` instead of `node-inspector`, see https://medium.com/@paul_irish/debugging-node-js-nightlies-with-chrome-devtools-7c4a1b95ae27#.2abd03s4e.

[...]

Issue #2337, opened 2017-10-05
runner.err.message.indexOf


chrisadubois:
## The problem

Briefly describe the issue you are experiencing (or the feature you want to see added to WebdriverIO). Tell us what you were trying to do and what happened instead. Remember, this is _not_ a place to ask questions. For that, join the Gitter chat room ([![Gitter](https://badges.gitter.im/Join Chat.svg)](https://gitter.im/webdriverio/webdriverio?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge))!

## Environment

* WebdriverIO version: latest
* Node.js version: 4.8.3
* [Standalone mode or wdio testrunner](http://webdriver.io/guide/getstarted/modes.html):
* if wdio testrunner, running synchronous or asynchronous tests: either/or
* Additional wdio packages used (if applicable): N/A

## Details

Describe in more detail the problem you have been experiencing, if necessary.

In our CI build, we consistently see random failures due to an exception thrown from 
`runner.err.message.indexOf` because `runner.err.message` is undefined.  There is no particular rhyme or reason to the failures besides that there is something wrong, but we believe they are due to random internal failures, for example including a firefox profile when it does not belong in the wdio.conf.js file;  however, it is difficult to debug these kinds of failures because this exception gets thrown and there is no stack available.  I'm very new to wdio, but it seems to me that the try from the try/catch block from `utils/ReporterStats` -> `testFail` could be moved up and the stack could be printed from the catch to stop these types of errors, or at the very least print the logs.  I'm also not sure how the error is occurring because the code in wdio very clearly checks `if (runner.err && runner.err.message && runner.err.message.indexOf(message) > -1)` Perhaps I am misinformed, doing something wrong, or the like, but it's not like our tests always fail either, they pass at least 80% of the time, these failures seem to pop up randomly.
See suggestion gist below.

## Link to Selenium/WebdriverIO logs N/A

Create a [gist](https://gist.github.com) which is a paste of your _full_ Selenium/WebdriverIO logs, and link them here. Do _not_ paste your full Appium logs here, as it will make this issue very long and hard to read! If you are reporting a bug, _always_ include Selenium/WebdriverIO logs! N/A

## Code To Reproduce Issue [ Good To Have ] N/A

Please remember that, with sample code; it's easier to reproduce bug and much faster to fix it. N/A

Replacement Code:
https://gist.github.com/chrisadubois/af892cb47e4c63797318c77536a7dfc5

[...]

PR #2350, opened 2017-10-16
Fix debug in Node 7+

mltsy:
## Proposed changes

Fixes the bug caused by the `--debug` flag being replaced with `--inspect` in Node 7+

This is a re-submission of an abandoned PR, with requested changes: https://github.com/webdriverio/webdriverio/pull/2200

Related Issues: #1640 , #2194

## Types of changes

## Checklist

## Further comments

I wasn't able to run tests because when I got to `bower install` I got a `command not found` message.  I tried `npm install bower` and that didn't help.  Any hints?

Also, I'm not sure I understand how these changes are supposed to work, now that I've looked at them.  I would think we'd just want to use `--inspect` if the node version is 7+ and use `--debug` otherwise, but this seems more complex than that... or just harder to understand perhaps?  (Sorry about this - I'm the the odd position of submitting a PR someone else wrote!) :wink:


mltsy:
Oh boy... I dont' know if we'll be able to get @hermanho's attention to sign the CLA.  If not, I can remove that commit, and just re-create the same functionality, which... I might do differently anyway, depending on the response to my questions under "Further Comments".

NicolasSiver:
Hello guys.
Where are we with these changes?

codiophile:
> Also, I'm not sure I understand how these changes are supposed to work, now that I've looked at them. I would think we'd just want to use --inspect if the node version is 7+ and use --debug otherwise, but this seems more complex than that... or just harder to understand perhaps? (Sorry about this - I'm the the odd position of submitting a PR someone else wrote!)

I'm as confused as you. I just changed `--debug` to `--inspect` and everything works fine. No clue what this code is good for.

I opened a pull request with this change https://github.com/webdriverio/webdriverio/pull/2377 before I saw this one. It doesn't check node version, but maybe it should? In what version was `--inspect` introduced? Maybe it's supported far back enough, that we don't need that check?

mltsy:
Yeah, it looks like that goes back to at least Node 6.3, which is before the current LTS release (6.11) so I would think that would be fine at this point. (Does wdio already require Node 6.3+ ?)  @christian-bromann - what do you think of closing this and merging #2337?

codiophile:
@mltsy You got the number wrong for the merge request. It's #2377 :)

mltsy:
Oops!  Thanks, fixed :)

mltsy:
Oh!  I actually thought #2377 looked more straight-forward and solved the problem (I was suggesting merging that one and closing this one).  I actually don't understand how this one is supposed to work, but if you think this does what we need, then great! :)

codiophile:
I don't mind whichever PR you merge, as long as it works, but I have a couple of concerns about this one.

My first concern is that sending both `--debug` and `--inspect` to a node version that supports both, e.g. node 6, might cause a conflict. Also, sending `--debug` to a version in which it is deprecated, will cause node to print the deprecation warning. This might be harmless, but not nice and not necessary.

My second concern is that the debugging is now based on the arguments node was started with, rather than whether the debug flag is set. I think this makes sense, but it makes the debug flag redundant, which hasn't been addressed. I think we should deprecate it and simply use this strategy rather than a mix of both.

I might make a PR later to show you what I mean.

christian-bromann:
> Oh! I actually thought #2377 looked more straight-forward and solved the problem

@mltsy @codiophile is it backwards compatible to node v4 tho?

> I think we should deprecate it and simply use this strategy rather than a mix of both.

I agree. @mltsy can you address this? I would be happy to remove one config property from the wdio.conf.js

codiophile:
I didn't make the PR yet, because I got stuck on building it, so that I could test it. Hence #2401. I think it would have made things clear.

> @mltsy @codiophile is it backwards compatible to node v4 tho?

My idea is that we don't have to worry about the node version if we just pass along the flags that we get from `process.execArgv`, i.e. if the parent is started with `--debug`, the children will be started with `--debug` and if the parent is started with `--inspect`, the children will be started with `--inspect`. That way we leave it to the user to choose which flag to use, which I think is the best option considering that node 6 supports both. 

To answer your question, `--inspect` was experimental in node 6 and I don't think it was available in node 4, but I haven't checked that.

As soon as I get things working I would be happy to make the PR.

christian-bromann:
> To answer your question, --inspect was experimental in node 6 and I don't think it was available in node 4, but I haven't checked that.

With the next version we still support node v4. Once Node v10 is out (probably May next year) I will kill support for Node v4.

> As soon as I get things working I would be happy to make the PR.

Awesome, thanks!

christian-bromann:
Closing in favor of #2418

PR #2404, opened 2017-11-06
Async await elements fix

lgandecki:
## Proposed changes

```javascript
  const notExistingElement = browser.element(".doesntExist")
  console.log("!!!!!!!!!!!!", await notExistingElement.isVisible());
```

Without the change:

```
Failures:

1) Scenario: Visit Google - features/webdriver.feature:8
   Step: Then I see the title of "Google" - features/webdriver.feature:10
   Step Definition: node_modules/cucumber/lib/index.js:176
   Message:
     Error: An element could not be located on the page using the given search parameters (".doesntExist").
         at new CommandError (/Users/lukaszgandecki/projects/chimp/test-project/node_modules/chimp/node_modules/webdriverio/build/lib/utils/ErrorHandler.js:131:12)
         at Object.elements (/Users/lukaszgandecki/projects/chimp/test-project/node_modules/chimp/node_modules/webdriverio/build/lib/protocol/elements.js:61:23)

```

With the change:
```
!!!!!!!!!!!! false
```
Similarly for waitForVisible, waitForExist, etc. Basically, current behavior doesn't let me use pageObjects pattern with elements that I have to wait for, or are intersted in checking of their visibility.

This fixes that. Unfortunately, I found another problem:

with:

```javascript
  const notExistingElement = browser.element(".doesntExist")
  console.log("!!!!!!!!!!!!", await notExistingElement.isVisible());
  console.log("!!!!!!!!!!!!", await notExistingElement.isVisible());
```

So, when trying to operate on this twice, I get:

```
!!!!!!!!!!!! false
  ✖ Then I see the title of "Google"

Failures:

1) Scenario: Visit Google - features/webdriver.feature:8
   Step: Then I see the title of "Google" - features/webdriver.feature:10
   Step Definition: node_modules/cucumber/lib/index.js:176
   Message:
     Error: selector needs to be typeof `string`
```

Same happens for existing elements, even without my change :-(

```javascript
  const existingElement = browser.element("a")
  console.log("!!!!!!!!!!!!", await existingElement.isVisible());
  console.log("!!!!!!!!!!!!", await existingElement.isVisible());
```
```
  ✔ When I visit "http://www.google.com"
!!!!!!!!!!!! true
  ✖ Then I see the title of "Google"

Failures:

1) Scenario: Visit Google - features/webdriver.feature:8
   Step: Then I see the title of "Google" - features/webdriver.feature:10
   Step Definition: node_modules/cucumber/lib/index.js:176
   Message:
     Error: selector needs to be typeof `string`
```

So.. the fix already makes it a bit better (and allows me to start rewriting my tests from chimp 1.0 to chimp 2.0 (which is now a very thin wrapper around the webdriverio library, allowing user to use any test runner), but not perfect. I will look into that, but meanwhile, having this in the codebase would be already great.

As for tests: would love some direction as to where to test things like this. At this moment async/await doesn't get too much love in terms of testing. I couldn't find a documentation on starting up the wdio tests. npm run test:wdio is failing for me. I could possibly put the tests there.

I would imagine a couple simple tests like this:
```javascript
  const existingElement = browser.element("a")
  expect(await existingElement.isVisible()).toEqual(true)

  const nonExistingElement = browser.element(".doesntExist")
  expect(await nonExistingElement.isVisible()).toEqual(false)

// doesnt throw when accessing the same page object twice
  const nonExistingElement = browser.element(".doesntExist")
  expect(await nonExistingElement.isVisible()).toEqual(false)
  expect(await nonExistingElement.isVisible()).toEqual(false)
```
Please let me know how we could push forward with those changes :)

## Types of changes

## Checklist


christian-bromann:
@lgandecki your example has an issue: if you run without wdio testrunner you need to await all commands not only `isVisible`:

```js
const notExistingElement = await browser.element(".doesntExist")
console.log("!!!!!!!!!!!!", await notExistingElement.isVisible());
```

`browser.element` is not different than `element.isVisible`.

lgandecki:
that's actually not completely correct. There is something wrong here.
If i do this:

```javascript
  const existingElement = await browser.element(".doesntExist")
  console.log("!!!!!!!!!!!!", await existingElement.isVisible());
```
I get TypeError: existingElement.isVisible is not a function , because what I get from await browser.element is a non-element, on which I can't do any actions.

When I run browser.element() on existing elements it works properly.
so:

```javascript
const existingElement = await browser.element("a")
console.log("!!!!!!!!!!!!", await existingElement.isVisible());
```

this works with and without my change. Basically, it doesn't try to find the document, until you call a specific action on it.

lgandecki:
So, maybe you are right, and I SHOULD be doing await before browser.element, but then I should be able to run things like isVisible, waitForExist, etc, even if the element doesn't exist at the moment when I call element(). The downside of this, is that we unnecessary look for the element, before we actually need it ( I might want to create a page object filled with a bunch of browser.elements() that are not validated till I need them)

simple example of what we do:

```javascript
export default class AssessmentList extends BasePageObject {

  root = '#tblBenchmarkAssessments';
  selectors = {
    optionsButton: '.page-header>.dropdown',
    newAssessment: '.addNewAssessment',
  };

  get instance(): WdioElement {
    return this.getElement(this.root);
  }

  get optionsButton(): WdioElement {
    return this.getElement(this.selectors.optionsButton);
  }

  get newAssessment(): WdioElement {
    return this.getElement(this.selectors.newAssessment);
  }
  async openNewAssessmentEditor() {
    await this.optionsButton.waitForVisible();
    await this.optionsButton.click();
    await this.newAssessment.waitForVisible();
    await this.newAssessment.click();
  }
```

christian-bromann:
@lgandecki the nice syntax you have in WDIO e.g.
```js
const elem = $("#elem")
elem.click();
```
You don't get when using standalone mode. There is a lot of magic happening in the [wdio-sync](https://github.com/webdriverio/wdio-sync/blob/master/index.js#L304) package that we __can't__ move over to the standalone version.

> but then I should be able to run things like isVisible, waitForExist, etc, even if the element doesn't exist at the moment when I call element()

Probably but this requires to enhance the element result.

lgandecki:
Hmm... so are you basically saying webdriverio is not compatible with async/await, and shouldn't be used that way? I don't expect to get the wdio syntax (basically, the same we have in chimp 1.0). I want to get rid of as much complexity as possible, that's why I'm looking into async / await. It's not as nice, but it's simpler. looks like we are very close here. 

Anyway - What's wrong with my tweak? At least when I have page objects as I showed above, the tweak makes it work. Do you see a downside to the change I proposed?

christian-bromann:
> Hmm... so are you basically saying webdriverio is not compatible with async/await

It is but the syntax between sync mode and async mode is different. In standalone mode the syntax is not that enhanced.

> Do you see a downside to the change I proposed?

Yes, it doesn't not solve your initial problem. Doing stuff like
```js
var elem = await browser.element("foo");
await foo.click()
```
doesn't work because the element result is an object that doesn't know anything about the WebdriverIO API. We have to enhance the element result for that. However if you do:
```js
await browser.element("foo").click()
```
it will work because the element result is transferred to the click command.

lgandecki:
Just checked this.

```javascript
await browser.element(".doesntExist").isExisting());
```

works with my change, without it, it throws:

     Error: An element could not be located on the page using the given search parameters (".doesntExist").

so, I'm not sure if the statement:

> await browser.element("foo").click()
> it will work because the element result is transferred to the click command.
> 

is fully correct. It will be, with my tweak :) . Without it, it breakes for elements that don't exist (yet)

samhatoum:
>  We have to enhance the element result for that.
It sounds like this is what we need in order to get a consistent API between standalone and WDIO mode. Perhaps this is an enhancement for the v5 changes. How can we help do this?

samhatoum:
The [Element-as-first-citizen](http://webdriver.io/api.html#Element-as-first-citizen docs) state that "The same works in standalone mode". 

I think the options here are:

  1. The API needs to actually be "The same works in standalone mode", in which case my comment above about the enhancement would resolve this issue
  2. The docs need to reflect that the API between Standalone and WDIO are inconsistent and that users can't use the PageObject pattern in the same way. across both versions.
c) Something I've not thought of yet :)

I'm personally for what you suggested @christian-bromann, which is to enhance WebdriverIO and provide users with a consistent API. Can you please reopen this issue to discuss this until a viable resolution is in place?

samhatoum:
Thanks for reopening @christian-bromann. Can we help get these enhancements in v5?

christian-bromann:
> Can we help get these enhancements in v5?

Sure, with a PR 😉

samhatoum:
haha, would love some guidance on what you'd like to see and the approach you want us to take on adding the enhanced element to standalone mode

lgandecki:
I get the behavior I want when - with addition to the change from this PR - I also remove the

  delete existingElement.isExecuted

This isExecuted is used only in one place with a comment:
"commands executed inside commands don't have to wait on any promise"

Could you elaborate a bit? Is this neccessary? is there a different way of achieving this, that would also allow us to do things like:

```javascript
const element = browser.element(".something")
if (await element.isVisible()) {
   await element.click() 
}
```

currently since the element gets marked as executed, I can't perform any more actions on it..

I think this might be a cleaner, less breaking way to go forward then to enhance results of element call.. especially since:
1) it will do unnecessary check for the element, even if I might not need it yet. ( for example two calls to check if isVisible() )
2) it will not work when the element doesn't exist at the moment of the selection, but might later. e.g. 
```javascript
const element = await browser.element(".notExisting")
await element.waitForExist() 
```
will not work (without even more changes) because at the moment of the .element("") call it will return an object with NoSuchElement. We could make it so it returns noSuchElement and still allow me to preform actions on it, based strictly on selector, but, really, is
```javascript
await (await browser.element("a")).isVisible(); 
```
nicer syntax than
```javascript
await browser.element("a").isVisible(); 
```
? ;-)

webOS101:
Is this fix similar to #2513?  I think my fix addresses this same issue.

PR #2418, opened 2017-11-08
Enable support for both --inspect and --debug

codiophile:
## Proposed changes

As promised in #2350 here's the pull request for enabling debugging for newer versions of node.

## Types of changes

## Checklist

## Further comments

I didn't have the chance to fully test this yet and I didn't update the documentation. The logic is put directly in `startInstance`, but could be broken out into its own function, which would make it unit testable. #2350 was done in a similar fashion, so I'm not sure unit tests are considered important for this functionality.

### DIFF
--- a/lib/launcher.js
+++ b/lib/launcher.js
@@ -295,14 +295,27 @@ class Launcher {
      */
     startInstance (specs, caps, cid, server) {
         let config = this.configParser.getConfig()
-        let debug = caps.debug || config.debug
         cid = this.getRunnerId(cid)
         let processNumber = this.processesStarted + 1
 
         // process.debugPort defaults to 5858 and is set even when process
         // is not being debugged.
-        let debugArgs = (debug)
-            ? [`--debug=${(process.debugPort + processNumber)}`] : []
+        let debugArgs = []
+        let debugType
+        let debugHost = ''
+        let debugPort = process.debugPort
+        for (let i in process.execArgv) {
+            let [, type, host] = process.execArgv[i].match('--(debug|inspect)(?:-brk)?(?:=(.*):)?')
+            if (type) {
+                debugType = type
+            }
+            if (host) {
+                debugHost = `${host}:`
+            }
+        }
+        if (debugType) {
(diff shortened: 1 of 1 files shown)


christian-bromann:
How is this different from #2350?

codiophile:
> How is this different from #2350?

It addresses all the concerns I had with #2350 as per my [comment](https://github.com/webdriverio/webdriverio/pull/2350#issuecomment-341683970) and adds a few additional improvements. Here's the list:

1. The children inherit the debug flag from its parent, rather than adding it if `debug` is set in the configuration.

2.  The children will use `--inspect` in case the parent is using `--inspect` and `--debug` if the parent is using `--debug`. This means that we do not have to worry about the node version, since the user will start the parent with the appropriate flag for the version of node he or she is using.

3. It passes on the host from the parent to the children. I'm not sure how used this feature is, but it was simple enough to add it.

4. Under no circumstances will it run the children with both `--debug` and `--inspect`, which is something #2350 would do.

codiophile:
@christian-bromann I've been using this quite successfully for a while now. I would appreciate feedback on what's missing to get this merged. I can see that the tests are failing, but I don't really see why `sauce-connect-launcher` would be affected by this change. Maybe you could help me figure out if it is a side effect of this change?

In addition to this change, the `debug` option in the config should be properly deprecated. Alternatively we could keep it and hide this functionality behind that option, but I'm not sure why we would want that. I haven't done that yet, because I'd like to know if this will get merged before investing more time in it.

christian-bromann:
> what's missing to get this merged

My lack of time 😉 

> I can see that the tests are failing

Because travis is not setup properly to run tests on SauceLabs when you propose PRs. I haven't got around to fix this.

> In addition to this change, the debug option in the config should be properly deprecated.

Once we stop supporting node v5 and lower we can remove it completely.

mltsy:
Yay! :confetti_ball: Thanks for fixing this :+1:

mltsy:
It might be a good time to switch these sections around to make Node 6+ the main documented strategy, and link to a special section for "Node versions < 6.3": http://webdriver.io/guide/testrunner/debugging.html#Node-Inspector

PR #2428, opened 2017-11-09
Don't clone on merge

mhemesath:
## Proposed changes

In 4.9 deepmerge was updated to version 2 which contains a non-passive change to always clone properties on merge. This seems to break passing in services as instances.  This change passes `{ clone: false }` to every relevant `merge()` call making it behave consistently with previous versions.

## Types of changes

## Checklist

### DIFF
--- a/lib/utils/ConfigParser.js
+++ b/lib/utils/ConfigParser.js
@@ -11,7 +11,7 @@ const HOOKS = [
     'beforeFeature', 'beforeScenario', 'beforeStep', 'afterFeature',
     'afterScenario', 'afterStep', 'onError', 'onReload'
 ]
-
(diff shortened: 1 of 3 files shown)


lib/utils/ConfigParser.js
```diff
@@ -98,7 +98,7 @@ class ConfigParser {
             /**
              * clone the original config
              */
-            var fileConfig = merge(require(filePath).config, {})
+            var fileConfig = merge(require(filePath).config, {}, { clone: false })
```
christian-bromann: Instead of passing the same parameters into the merge call everytime, how about making a constant like `MERGE_OPTIONS` and just pass these in?

Issue #2447, opened 2017-11-16
Callback to allow retry when session creation fails


codiophile:
## The problem

I described the problem in detail in #2431. In summary, we are testing on mobile devices that are not always in a usable state. When this happens the session creation fails and the scenarios we were supposed to run in that session do not execute. We'd like to have the opportunity to reboot the device and retry session creation when this happens, so that we can get some reliable results.

## Details

I still think #2431 is worthy of consideration, but since we need this urgently I added a hook to allow retries when session creation fails. This is the alternative solution I presented in #2431. It's a lot less intrusive and a lot less effort to implement. We are currently running it, both to verify it and to evaluate its usefulness. I will shortly open a PR with this change.

christian-bromann:
> I will shortly open a PR with this change.

Thanks for the info. Let's close this and have further discussions in your PR

PR #2448, opened 2017-11-16
Add session failure hook to allow retries

codiophile:
## Proposed changes

This implements the feature suggested in #2447. It's a callback to allow retries when session creation fails.

## Types of changes

## Checklist

## Further comments

As mentioned in #2447, we are already using this, since we urgently needed to improve our run rate. It's too early to evaluate the impact of this change, since we don't have enough runs yet, but it has allowed us to successfully recover when the session creation failed.

I haven't added any tests or any documentation yet, because I figured you might have opinions on the name of the hook, parameters, etc. and it's better to add the documentation when the implementation has been agreed upon.

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -162,7 +162,23 @@ class Runner {
         global.$$ = (selector) => global.browser.elements(selector).value
 
         try {
-            let res = await global.browser.init()
+            let retries = 0
+            let retry = true
+            let res
+            while (retry) {
+                try {
+                    res = await global.browser.init()
+                    retry = false
+                } catch (e) {
+                    let shouldRetry = await this.runHook('sessionFailure', config, this.caps, this.specs, retries, e)
+                    retry = shouldRetry.some(x => x) // If any hook returned true, we retry.
+                    if (!retry) {
+                        throw e
+                    }
+                }
+                retries++
+            }
+
             global.browser.sessionId = res.sessionId
             this.haltSIGINT = false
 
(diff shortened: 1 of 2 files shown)


christian-bromann:
> As mentioned in #2447, we are already using this, since we urgently needed to improve our run rate. It's too early to evaluate the impact of this change, since we don't have enough runs yet, but it has allowed us to successfully recover when the session creation failed.

How about using the beforeSession hook to check if devices are available? This allows you to block the session creation until you can ensure that you can initialise one. This is a patch to the wrong place of the problem.

codiophile:
> How about using the beforeSession hook to check if devices are available? This allows you to block the session creation until you can ensure that you can initialise one. This is a patch to the wrong place of the problem.

To be fair, there is an API to check if the device is ready, but I do believe that session creation will fail occasionally, even when the device state is reported as ready. Retrying session creation seems like basic functionality that will benefit a lot of people. Are you sure there's no way to get webdriverio to support this?

christian-bromann:
> but I do believe that session creation will fail occasionally

We already have a command retry mechanism, why we can't integrate it into this?

> Are you sure there's no way to get webdriverio to support this?

As I said, I think this should not be fixed in the WebDriver client and rather be addressed in the device pool you are maintaining. I am currently don't see how this could help other people in other use cases. With the increasing number of commands and functionality of WebdriverIO I became careful of what needs to be added to the core without creating to much confusion to people or providing ways to follow antipatterns.

codiophile:
> We already have a command retry mechanism, why we can't integrate it into this?

That sounds interesting. How do I use it?

christian-bromann:
See RequestHandler.js

PR #2463, opened 2017-11-21
fix: 'request' in RequestHandler should be fulfilled if body.status is 0

rostik404:
## Proposed changes

In some cases error field in body is not empty, but status is `0`. `isSuccessfulResponse` should return `true` if `body.status` is 0.

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/lib/helpers/utilities.js
+++ b/lib/helpers/utilities.js
@@ -12,6 +12,14 @@ export function isSuccessfulResponse (body) {
     }
 
     /**
+     *
+     * if it has a status property, it should be 0
+     */
+    if (body.status === 0) {
+        return true
+    }
+
+    /**
      * if it has a status property, it should be 0
      * (just here to stay backwards compatible to the jsonwire protocol)
      */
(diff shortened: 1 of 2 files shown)


lib/helpers/utilities.js
```diff
@@ -11,6 +11,14 @@ export function isSuccessfulResponse (body) {
         return false
     }
 
+    /**
+     *
+     if it has a status property, it should be 0
```
sipayRT: `* if it has a status property, it should be 0`

christian-bromann:
When does this happen? Can this be reported rather in the operadriver project? I feel uncomfortable to implement something that is against the spec.

rostik404:
This is not about the operadriver. We can call some jquery function in `execute` and get instance of jquery in a body, which includes error field: `error: 'function ( data, fn ) {\n\t\treturn arguments.length > 0 ?\n\t\t\tthis.on( name, null, data, fn ) :\n\t\t\tthis.trigger( name );\n\t}'`

PR #2477, opened 2017-11-26
process.stdin.isTTY is false when running commands from a Node script

WillBrock:
## Proposed changes

Adding !process.stdout.isTTY which will allow launch() to be ran when called from a node script. Node's child_process spawn is treated as a pipe so the original process.stdin.isTTY is false and launch() would never get called.

Doing something like the following will never launch with just process.stdin.isTTY.

```
const { spawn } = require(`child_process`);

spawn(`./node_modules/.bin/wdio wdio.config.js --spec ./some/test/spec.js`);
```

## Types of changes

- [ x] Bugfix (non-breaking change which fixes an issue)

## Checklist

## Further comments

If there is a better solution let me know.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -424,8 +424,9 @@ const launch = () => {
  * if stdin.isTTY, then no piped input is present and launcher should be
  * called immediately, otherwise piped input is processed, expecting
  * a list of files to process.
+ * stdin.isTTY is false when command is from nodes spawn since it's treated as a pipe
  */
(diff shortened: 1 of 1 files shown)


WillBrock:
@christian-bromann I'm not exactly sure what you mean by that.

christian-bromann:
@WillBrock see http://webdriver.io/guide/testrunner/gettingstarted.html#Run-the-test-runner-programmatically

PR #2493, opened 2017-11-30
add elementIdProperty W3C command and use it in getValue

sherlock1982:
## Proposed changes

DOM properties and HTML attributes are not the same. Currently `browser.getValue` command uses W3C protocol command `/attribute` which is wrong because getValue should return current value of the DOM property and not a value of HTML attribute. 

In geckodriver `/attribute` returns null while `/property` works fine.
In chromedriver `/attribute` somehow mashes up this terms. Currently `/property` is not implemented in chromedriver but they have an open issue to implement it.

The change is to add `/property` command and use it in getValue but fallback to `/attribute` command in case `/property` is not supported.

I run test:desktop with both chrome and firefox.

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/lib/commands/getValue.js
+++ b/lib/commands/getValue.js
@@ -18,7 +18,7 @@
  * @alias browser.getValue
  * @param   {String} selector input, textarea, or select element
  * @return {String}          requested input value
- * @uses protocol/elements, protocol/elementIdAttribute
+ * @uses protocol/elements, protocol/elementIdProperty
  * @type property
  *
  */
@@ -34,12 +34,12 @@ let getValue = function (selector) {
             throw new CommandError(7, selector || this.lastResult.selector)
         }
 
-        let elementIdAttributeCommands = []
+        let elementIdPropertyCommands = []
         for (let elem of res.value) {
-            elementIdAttributeCommands.push(this.elementIdAttribute(elem.ELEMENT, 'value'))
+            elementIdPropertyCommands.push(this.elementIdProperty(elem.ELEMENT, 'value'))
         }
 
-        return this.unify(elementIdAttributeCommands, {
+        return this.unify(elementIdPropertyCommands, {
             extractValue: true
         })
     })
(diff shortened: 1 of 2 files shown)


lib/commands/getValue.js
```diff
@@ -34,12 +34,12 @@ let getValue = function (selector) {
             throw new CommandError(7, selector || this.lastResult.selector)
         }
 
-        let elementIdAttributeCommands = []
+        let elementIdPropertyCommands = []
         for (let elem of res.value) {
-            elementIdAttributeCommands.push(this.elementIdAttribute(elem.ELEMENT, 'value'))
```
christian-bromann: Can we have a fallback here that (like [here](https://github.com/webdriverio/webdriverio/blob/master/lib/protocol/keys.js#L48)) to ensure that this still works for driver that haven't implemented the `elementIdProperty` command?

j0tunn:
@sherlock1982 Hi. Have you tried it in ChromeMobile and appium?
After this change I'm instantly getting
```
✘ Error: The URL '/wd/hub/session/45a2998f-f6c2-4d8e-a787-431e2feeee0a/element/0.5258714092291512-4/property/value' did not map to a valid resource
    at elementIdProperty("0.5258714092291512-4", "value") - getValue.js:35:54
```
while with `webdriverio@4.9.11` it works fine.

`appium@1.7.1`, `ChromeMobile@63`

j0tunn:
Thats a message from appium's [404 handler](https://github.com/appium/appium-base-driver/blob/v2.19.0/lib/express/middleware.js#L66-L72). Looks like it knows nothing about `/property` request

sherlock1982:
Nope I haven't checked. Can you ask them to implement this command? Because if command is not known it should go to fallback but the message should be different. That's why it doesn't even try it.

christian-bromann:
Added that edge case and will release once I get a passing build

j0tunn:
@christian-bromann Thanks!

christian-bromann:
@j0tunn released as v4.10.1

Issue #2510, opened 2017-12-12
Chaining with isExisting gives inconsistent results


webOS101:
## The problem

Calling `isExisting()` on a non-existent element when chaining can produce false positives as the chain is disregarded.

## Environment

* WebdriverIO version: 4.9.11 (wdio 1.0.3)
* Node.js version: v8.1.4
* [Standalone mode or wdio testrunner](http://webdriver.io/guide/getstarted/modes.html): both
* if wdio testrunner, running synchronous or asynchronous tests: sync
* Additional wdio packages used (if applicable): n/a

## Details

HTML source
```html
<html>
	<body>
		<div id="first">
			<div class="class">one</div>
			<div class="class">two</div>
		</div>
		<div id="second">
			 Empty
		</div>
	</body>
</html>
```
REPL output:
```
> $('#second').$('.class').isExisting()
true
> $('#second .class').isExisting()
false
```

Expect both commands to return the same result.

The problem is that once the selector fails, it re-executes with `elements()` at the root scope, returning the array of matching divs.

## Link to Selenium/WebdriverIO logs

Selenium log: https://gist.github.com/webOS101/c55ec4a097dfbdcc2fa49b5e9f82c802

## Code To Reproduce Issue [ Good To Have ]

See above.

treyturner:
I tried patching these changes into 4.13.1, but it didn't resolve my issue. I'm still getting false positives for `.isVisible()` and `.isExisting()` due to the root scope issue. Very much looking forward to a fix. This is purported to be fixed in the v5 beta according to #2571, but I tried `5.0.0-beta14` and the problem persists.

webOS101:
That's sad to hear, on both parts. :(  We haven't updated to 5.0 beta yet so I haven't tested any of my failure cases there.

christian-bromann:
Let's get this fixed then before we release it.

treyturner:
I'm having a hard time figuring out if this is the issue I'm suffering, or if it's something else entirely. In `5.0.0-beta.2`:
```
> // Here's code for a dropdown. When the anchor is clicked, a <ul> appears in div.options.
> apps.settings.window.content.general.locationFormat.dropdown.container.getHTML()
'<td>
    <div class="drop-down-menu undefined">
        <a class="btn selection ">Decimal Degrees</a>
        <div class="options">
        </div>
    </div>
</td>'

> // Check to see if the list exists (it doesn't)
[...]

PR #2513, opened 2017-12-14
Add failing test case and fix for #2510

webOS101:
## Proposed changes

Allow chaining from a failed selector to work correctly.  In other words, if any part of a chain of selectors fails, allow the failure to propagate instead of starting the selectors over again at the root level.

## Types of changes

## Checklist

## Further comments

There is one thing that I'm a little concerned about, and that's in `elements.js`, where I'm modifying the previous result.  I could create a new result object and copy out some of the fields from the previous failure.  If it's OK to modify the old value, then it can go as-is.

Also, there a failing tests for me when run locally, but the tests are also failing on `master`.  The total number of failures does not change with this PR.


webOS101:
This PR has been open without comments since December.  Is there something more needed to push this through?

kamranayub:
Just ran into this because we are testing a React app which re-renders elements when updated and realized this bug is manifesting due to element IDs being lost during multiple user interactions.

It would be awesome if this could be prioritized and merged.

PR #2520, opened 2017-12-15
Add support for filtering specs

webOS101:
## Proposed changes

This change modifies the `--spec` command line option to treat any of the specified strings as a regex to apply against the set of specs defined in the configuration file. This makes it possible to easily execute a subset of specs without specifying the full path.  This should be fully backwards compatible .  For example:

```sh
$ wdio wdio.conf.js --spec dialog
```

This will execute any spec file whose name contains the word 'dialog'.

## Types of changes

## Checklist

## Further comments

One (possible?) downside to this change is that when `--spec` is supplied on the command line, the runner will glob the files twice.  The second time, it should only be globbing against actual filenames so this doesn't seem like a bad trade-off. It is possible this could be done through a different command line option, but it seems a natural fit for the existing `--spec` option.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -94,7 +94,7 @@ optimist
     .describe('reporters', 'reporters to print out the results on stdout')
     .alias('reporters', 'r')
     .describe('suite', 'overwrites the specs attribute and runs the defined suite')
-    .describe('spec', 'specifies spec file(s) or filter(s) for specs piped from stdin')
+    .describe('spec', 'specifies spec file(s) to run or filter(s) for specs defined in the specs attribute')
     .describe('cucumberOpts.*', 'Cucumber options, see the full list options at https://github.com/webdriverio/wdio-cucumber-framework#cucumberopts-options')
     .describe('jasmineOpts.*', 'Jasmine options, see the full list options at https://github.com/webdriverio/wdio-jasmine-framework#jasminenodeopts-options')
     .describe('mochaOpts.*', 'Mocha options, see the full list options at http://mochajs.org')
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -94,7 +94,7 @@ optimist
     .describe('reporters', 'reporters to print out the results on stdout')
     .alias('reporters', 'r')
     .describe('suite', 'overwrites the specs attribute and runs the defined suite')
-    .describe('spec', 'run only a certain spec file - overrides specs piped from stdin')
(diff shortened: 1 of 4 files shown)


webOS101:
The test for filtering is a little weak.  It would be nicer if there were a more full set to filter against but I didn't update the config used in that test suite.  If desired, I can make more changes to that test so it won't accidentally pass if the filtering doesn't apply.

lib/cli.js
```diff
@@ -94,7 +94,7 @@ optimist
     .describe('reporters', 'reporters to print out the results on stdout')
     .alias('reporters', 'r')
     .describe('suite', 'overwrites the specs attribute and runs the defined suite')
-    .describe('spec', 'run only a certain spec file - overrides specs piped from stdin')
+    .describe('spec', 'specifies spec file(s) or filter(s) for specs piped from stdin')
```
christian-bromann: I think this wording change makes it a little bit confusing. It looks like you only can pipe in the specs.

lib/utils/ConfigParser.js
```diff
@@ -152,23 +152,32 @@ class ConfigParser {
...
 
         /**
-         * run single spec file only, regardless of multiple-spec specification
+         * run only specified spec files, regardless of multiple-spec specification
+         * If spec is a file on disk, it is set as the current spec, if it is not, it
+         * is treated as a string match filter for the multiple-spec specification.
          */
         if (typeof object.spec === 'string') {
-            const specs = []
+            const specs = new Set()
```
christian-bromann: What is the advantages if using Set here instead of just an array?

lib/cli.js
webOS101: I can try to spiff it up some.  Is there a limit to how long you want that to be?

lib/utils/ConfigParser.js
webOS101: It was to prevent the same file from being added twice.  Could de-dupe after, unless something does that later and it's unnecessary.

lib/cli.js
webOS101: I tried to reword it slightly and removed the references to stdin, but maybe that's not entirely what you wanted.  I'm trying to convey that you can either specify the exact spec files you want to run or else you want to filter any spec files that have been defined elsewhere.  Suggestions appreciated.

webOS101:
Are there any changes you want on this ticket to get it to close?

Issue #2484, opened 2017-11-28
Retries on a per-specfile basis


bennieswart:
We are running tests with wdio, which very nicely creates a new browser instance for each spec file. We have rigged this so that a new server+db is created for each browser that starts up, allowing tests to run in complete isolation. This all works very nicely.

Sometimes tests have false negatives, however, and in these cases we want to rerun the test, again with a new server+db. This would 'just work' if wdio had some way to do a specfile level retry.

I'm aware that describe-level and it-level (mocha) retries are possible, but since a new browser instance is created on a per-specfile basis, it would also be really handy to have retries on a per-specfile basis.

Is this possible?

christian-bromann:
@bennieswart as described in http://webdriver.io/guide/testrunner/retry.html#Rerun-suites-in-MochaJS please use `this.retries(4)`
Join our  [![Gitter](https://badges.gitter.im/webdriverio/webdriverio.svg)](https://gitter.im/webdriverio/webdriverio?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge) for these kind of questions. This is not necessary an issue.

bennieswart:
@christian-bromann maybe I was not clear enough: I'm aware of the describe-level and it-level retries, and as I understand, this is what you are describing.

These kinds of retries will only rerun the test code, without launching a new browser instance and causing (in our case) a new server+db to be created. The problem here is that the first try might leave the db in an unintended state, causing all future retries to fail immediately. This is why a new db is required for each retry.

As far as I can tell this is not possible (or at least not documented - I've not looked into the code yet).
Do you not think that retries on this level may be useful? Especially in situations like the one I just described.

christian-bromann:
Ah, so you want per file retries instead of per test/describe. No this is not possible but can get implemented. PRs welcome.

bennieswart:
Great. I just wanted to run this by you to see if it makes sense. I intend to make a PR, but with the upcoming holidays it might be a while before I get to it.

kokujin:
Any news on this?

bennieswart:
@kokujin funny you should ask right now... I've literally just gotten back to this today.

Should have a PR ready before the end of the week. It works on the principle I've outlined earlier, in that when `retries > 0`, messages are buffered. This is the only downside I see, but at least it only comes when `specFileRetries` is set to a non-default value. Otherwise everything works as per normal.

kokujin:
[...]

PR #2556, opened 2018-01-18
Add retries on a per-specfile basis

bennieswart:
## Proposed changes

Add retries on a per-specfile basis, as discussed in https://github.com/webdriverio/webdriverio/issues/2484.

Previously only test-level and suite-level retries were available, which
are fine in most cases. In any test which involves state, such as on a
server or in a db, the state may be left invalid once the test fails the
first time. Any subsequent retries may have no chance of passing due to
the invalid state they would have to start with.

A new browser instance is created for each specfile, which makes this an
ideal place to hook and setup any other states (server, db). Retries on
this level would mean that the whole setup process could simply be
repeated as if it were for a new specfile, with the results of the
failing specfile (that is to be retried) being discarded.

## Types of changes

## Checklist

## Further comments

Documentation kindly provided by @pdirt and pending in #2614.
Will integrate docs with this PR if so requested. Otherwise, just merge both.

### DIFF
--- a/lib/launcher.js
+++ b/lib/launcher.js
@@ -334,11 +334,6 @@ class Launcher {
         // If an arg appears multiple times the last occurrence is used
         let execArgv = [ ...defaultArgs, ...debugArgs, ...capExecArgs ]
 
-        let childProcess = child.fork(path.join(__dirname, '/runner.js'), process.argv.slice(2), {
-            cwd: process.cwd(),
-            execArgv
-        })
-
         /**
          * increase framework timeout to 10m when debugging is enabled
          */
@@ -360,24 +355,50 @@ class Launcher {
             this.argv[frameworkOptsKey][frameworkTimeoutKey] = 10 * 60 * 1000
         }
 
-        this.processes.push(childProcess)
+        let childProcess
+        let retries = this.configParser.getConfig().specFileRetries || 0
+        let spawn = () => {
+            let messages = []
+            let idx = this.processes.indexOf(childProcess)
 
-        childProcess
-            .on('message', this.messageHandler.bind(this, cid))
-            .on('exit', this.endHandler.bind(this, cid))
+            childProcess = child.fork(path.join(__dirname, '/runner.js'), process.argv.slice(2), {
+                cwd: process.cwd(),
+                execArgv
+            })
 
-        childProcess.send({
-            cid,
-            command: 'run',
-            configFile: this.configFile,
-            argv: this.argv,
-            caps,
-            processNumber,
-            specs,
-            server,
-            isMultiremote: this.isMultiremote()
-        })
+            this.processes.splice(idx === -1 ? this.processes.length : idx, 1, childProcess)
 
+            childProcess.on('message', (...args) => {
+                if (retries === 0) {
+                    this.messageHandler(cid, ...args)
+                } else {
+                    messages.push(args)
+                }
+            })
+            childProcess.on('exit', (exitCode, ...args) => {
+                if (exitCode === 0 || retries === 0) {
+                    messages.forEach(message => this.messageHandler(cid, ...message))
+                    this.endHandler(cid, exitCode, ...args)
+                } else {
+                    retries--
+                    spawn()
+                }
+            })
+
+            childProcess.send({
+                cid,
+                command: 'run',
+                configFile: this.configFile,
+                argv: this.argv,
+                caps,
+                processNumber,
+                specs,
+                server,
+                isMultiremote: this.isMultiremote()
+            })
+        }
+
+        spawn()
         this.processesStarted++
     }
 
--- a/lib/launcher.js
+++ b/lib/launcher.js
@@ -368,17 +368,18 @@ class Launcher {
 
             this.processes.splice(idx === -1 ? this.processes.length : idx, 1, childProcess)
 
-            childProcess.on('message', (...args) => {
+            childProcess.on('message', message => {
+                message._timestamp = new Date()
                 if (retries === 0) {
-                    this.messageHandler(cid, ...args)
+                    this.messageHandler(cid, message)
                 } else {
-                    messages.push(args)
+                    messages.push(message)
                 }
             })
-            childProcess.on('exit', (exitCode, ...args) => {
+            childProcess.on('exit', exitCode => {
                 if (exitCode === 0 || retries === 0) {
-                    messages.forEach(message => this.messageHandler(cid, ...message))
-                    this.endHandler(cid, exitCode, ...args)
+                    messages.forEach(message => this.messageHandler(cid, message))
+                    this.endHandler(cid, exitCode)
                 } else {
                     retries--
                     spawn()
--- a/lib/launcher.js
+++ b/lib/launcher.js
@@ -339,6 +339,27 @@ class Launcher {
             execArgv
         })
 
+        /**
+         * increase framework timeout to 10m when debugging is enabled
+         */
+        for (const argv of execArgv) {
+            if (argv !== '--inspect') {
+                continue
+            }
+
+            const frameworkOptsKey = config.framework === 'mocha'
+                ? 'mochaOpts' : config.framework === 'jasmine'
+                    ? 'jasmineNodeOpts' : 'cucumberOpts'
+            const frameworkTimeoutKey = config.framework === 'jasmine'
+                ? 'defaultTimeoutInterval' : 'timeout'
+
+            if (!this.argv[frameworkOptsKey]) {
+                this.argv[frameworkOptsKey] = {}
+            }
+
+            this.argv[frameworkOptsKey][frameworkTimeoutKey] = 10 * 60 * 1000
+        }
+
         this.processes.push(childProcess)
 
         childProcess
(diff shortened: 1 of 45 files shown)


bennieswart:
We've been running this patch for a couple of days now and the retries seem to be working nicely. I just noticed however that the reported duration of tests have dropped down to tens of millis. Only tests that have passed on their final run are correctly reporting duration. This correlates with the messages being buffered. I suspect start/end timestamps are assigned by the message handler as apposed to being contained within the message itself. Will look into this a bit later - should be an easy fix.

avkrash:
@bennieswart can you add @christian-bromann as reviewer, maybe it will hasten this PR

pmerwin:
Awesome feature!  Thanks for working on it!

bennieswart:
@christian-bromann I was waiting for your approval on the approach used before I went further. Will address those two points as soon as I get some free time.

pmerwin:
@bennieswart  I have time to help, and would love to.

bennieswart:
@pmerwin by all means go ahead!

pmerwin:
@bennieswart 
Ok, for docs, we want to add to:
http://webdriver.io/guide/testrunner/retry.html ( I assume)
and a mocha test in tests for the test.

This will be my first time contributing, so I am not certain of what to do?  Can you point me in the direction of how I edit:

http://webdriver.io/guide/testrunner/retry.html

bellstrand:
@pmerwin You can edit the docs in this repo. The file you're after is located at "[/docs/guide/testrunner/retry.md](https://github.com/webdriverio/webdriverio/blob/master/docs/guide/testrunner/retry.md)".

pmerwin:
@bellstrand  ok will do!

pdirt:
@bellstrand @bennieswart 

Here is the PR for the docs for this PR.  Please let me know if I have the info correct.

https://github.com/webdriverio/webdriverio/pull/2614

bennieswart:
@christian-bromann I have added tests as requested.
It was somewhat tricky to determine whether or not the specfile-level retries are working, since each retry is a new process and hence state cannot be preserved. You're welcome to suggest a better approach than the current one if required.

pdirt:
@bennieswart Is there any indication that a spec file failed and was retried in the logs?

lib/utils/ReporterStats.js
```diff
@@ -3,14 +3,14 @@ import crypto from 'crypto'
 import sanitize from '../helpers/sanitize'
 
 class RunnableStats {
-    constructor (type) {
+    constructor (type, start) {
```
christian-bromann: Can you elaborate why these changes are necessary?

test/fixtures/specs/retries.spec.js
```diff
@@ -0,0 +1,29 @@
+console.log('Retry file loaded.')
+
+describe('retry', function () {
+    this.retries(2)
```
christian-bromann: Why are you including Mocha timeouts here?

test/spec/wdio/retries.js
```diff
@@ -0,0 +1,42 @@
+import { spawn } from 'child_process'
+import path from 'path'
+import Launcher from '../../../build/lib/launcher'
+
+const STANDALONE_FLAG = '--run-launcher-standalone'
+
+if (process.argv.includes(STANDALONE_FLAG)) {
```
christian-bromann: What is the standalone doing?

lib/utils/ReporterStats.js
bennieswart: This is explained in a [previous comment](https://github.com/webdriverio/webdriverio/pull/2556#issuecomment-359717793) and has to do with incorrect timestamps of buffered messages.
> We've been running this patch for a couple of days now and the retries seem to be working nicely. I just noticed however that the reported duration of tests have dropped down to tens of millis. Only tests that have passed on their final run are correctly reporting duration. This correlates with the messages being buffered. I suspect start/end timestamps are assigned by the message handler as apposed to being contained within the message itself. Will look into this a bit later - should be an easy fix.

test/fixtures/specs/retries.spec.js
bennieswart: I couldn't find any tests that tested the existing retry mechanisms, so I added some for those as well. I can remove them if you don't think they're necessary. I thought it couldn't hurt to have tests for retries on all three levels (test, suite, file).

christian-bromann: Got it but this should not be tested together with the spec retry. Having separate tests for it would make things easier I guess

test/spec/wdio/retries.js
bennieswart: I'll address your overall comment here as well. The temporary file approach you suggest was something I considered as opposed to spawning a separate process for the retry-tests and checking the output. The reason I went with the separate process is that it doesn't require modifying the filesystem, whereas the temporary file obviously does. It seemed that, since the tests already spawn child processes, the separate process approach would be guaranteed to always work, whereas there could be situations in which modifying the filesystem would not (readonly access, for whatever reason).

Here, a separate process is spawned that runs the retry-tests and checks the output of that process to determine whether or not this test should pass or fail.
Depending on whether or not the flag is set, this file either *spawns* the new process, or *is* the new process. This could just as easily have been done using two different files, but keeping everything together made sense.

If you prefer the temporary file approach, I'll change it, no problem.

test/fixtures/specs/retries.spec.js
bennieswart: Alright. Should I move them directly into `test/spec/wdio/retries.js` (since testing test-level and suite-level retries are quite trivial), or do you want them as a separate spec in the fixtures? Or something else?

bennieswart:
@christian-bromann what are your thoughts on my previous comments? I'll make those changes once we're certain about the direction this is going.

christian-bromann:
@bennieswart sorry for the late response, I will look into this asap

lib/launcher.js
```diff
@@ -360,24 +355,51 @@ class Launcher {
...
+                cwd: process.cwd(),
+                execArgv
+            })
+
+            this.processes.splice(idx === -1 ? this.processes.length : idx, 1, childProcess)
+
+            childProcess.on('message', message => {
+                message._timestamp = new Date()
+                if (retries === 0) {
+                    this.messageHandler(cid, message)
```
christian-bromann: Just got an idea from the Gitter channel, let's add the retry count to the message object so we can propagate the info to the reporters.

bennieswart: The way things currently work will only ever send messages to the reporters when `retries === 0`.
When `retries > 0` and the specfile fails, buffered messages are discarded, and the reporters would be unaware that the failing testrun ever existed in the first place.

The reason for this is so that existing code continues to work. I suspect that if we suddenly report failing tests to reporters (even if they are retried soon afterwards) then a lot of things in the field will probably break.

christian-bromann: > then a lot of things in the field will probably break.

Good point, let's leave stuff as is then

bennieswart: I'm all for the idea of reporting the failures and retries to reporters, as this would allow for more insight into what happened. I just suspect things would become slightly more involved then than they currently are.
Mocha's suite-level and test-level retries are transparent (to my knowledge). Perhaps the specfile-level retries should be too.
Your call, ultimately.

bennieswart:
@christian-bromann please provide feedback on the previous discussions so I know which direction to take the changes you have requested.
If you could list the outstanding changes along with your preference on each that would be great!

dannsam:
@christian-bromann @bennieswart can I be of any help with this PR? Are there any outstanding comments that need to be addressed ?

YmerejRedienhcs:
@bennieswart @christian-bromann any movement on this?  I seems like it would be very useful.

caccialdo:
This is definitely a useful feature for us too. For now, we've forked this branch and use it as a private package but it would be great to see it officially part of the runner 👍

bennieswart:
@christian-bromann are we still moving forward with this branch? I see the commit history has gone wacky (I haven't changed anything), but I can easily fix it if you still plan on moving forward.

christian-bromann:
@bennieswart we updated the code basis to v5. Would you mind applying the patch to the new code basis?

bennieswart:
@christian-bromann with regards to the discussion here https://github.com/webdriverio/webdriverio/pull/2556#discussion_r172349719, do you want notify reporters of the spec failure even though they are going to be retried? Or do you want to keep the failure transparent as is currently being done.

As mentioned previously, reporting failures (that will be retried) will probably break some existing things, but conversely the additional detail in the report may be useful. If ever there was a time to make such a (possibly) breaking change, I guess a major version bump is that time. Your call.

christian-bromann:
> Or do you want to keep the failure transparent as is currently being done.

I think we should keep it transparent. It would be good if we would pass the retry count to reporters so they can use it to display how many retries the test took.

bennieswart:
@christian-bromann I've had a look at the v5 codebase and it seems that an implementation for v5 would look quite different than the v4 implementation. Would it be possible to still merge this into v4 and then proceed to port it to v5 also? I'm guessing many people will be staying on v4 for quite some time to come until v5 has stabilized sufficiently (including me).

I've updated the source and destination branches of this PR, but the commit history is still broken. Perhaps it will also update when the PR is reopened. Otherwise, I'll just open a new one. That's if you're good with still getting this into v4.

christian-bromann:
> I'll just open a new one.

Please open a new RR. Happy to merge it into v4.

YmerejRedienhcs:
Thank you to both of you for this.

PR #2558, opened 2018-01-19
- fix a crash in errorHandler when commandList is undefined

willem-delbare:
It's possible for the errorHandler to crash under certain conditions. 

(Stacktrace for error)
TypeError
Cannot read property 'slice' of undefined
/var/app/current/node_modules/webdriverio/build/lib/helpers/errorHandler.js in Object.exports.default at line 30:42
/var/app/current/node_modules/webdriverio/build/lib/webdriverio.js in null.<anonymous> at line 199:27
Array.map
/var/app/current/node_modules/webdriverio/build/lib/webdriverio.js in Object.resolve at line 198:77
/var/app/current/node_modules/webdriverio/build/lib/webdriverio.js in null.<anonymous> at line 490:32
/var/app/current/node_modules/webdriverio/node_modules/q/q.js in _rejected at line 864:24
/var/app/current/node_modules/webdriverio/node_modules/q/q.js in null.<anonymous> at line 890:30
...


christian-bromann:
@willem-delbare `this.commandList` should always be an array. I think the problem here is a wrong scope. Do you have a reproducible example? Your fix isn't really solving the actually issue.

willem-delbare:
We think it occurs when the selenium server is under heavy load. This can happen when you use commands like webdriverio's isVisible in the wrong way. 

Eg isVisible('#container a')

If #container contains about 100 anchor tags, it will now spam the selenium server with commands. A user might expect it to only query one element at most. Maybe this function should be amended or log a warning in such cases. It's unlikely to even receive a response for all 100 queries (some will timeout).

I'll post an update when we have a minimal sample to reproduce.

christian-bromann:
ping @willem-delbare

Issue #2574, opened 2018-01-31
Cannot getValue on a textarea tag with Appium on iOS


ianrenauld:
## The problem

Running the following command:

```
const comment = $('#product-comment-textarea').getValue;
```

works fine with WebdriverIO 4.9.9, but returns a "Method has not yet been implemented" with WebdriverIO 4.10.1.  The idea is to get the content of a <textarea> element.

## Environment

* WebdriverIO version: 4.10.1
* Node.js version: 6.7.0
* wdio testrunner, running synchronous
* Appium: 1.7.2

## Details

Some context: testing an hybrid app using Appium. The above command works fine for Android on all recent versions of Webdriver.io but does not work for iOS (using XCUITest) on 4.10.1 (works fine on 4.9.9)

## Link to Selenium/WebdriverIO logs

Here is the log for WebdriverIO 4.9.9:

```
[09:56:31]  COMMAND     POST     "/wd/hub/session/113800f5-ed40-41b6-aa7d-f7b0ea474193/element"
[09:56:31]  DATA                {"using":"id","value":"product-comment-textarea"}
[09:56:31]  RESULT              {"ELEMENT":"5562"}
[09:56:31]  COMMAND     GET      "/wd/hub/session/113800f5-ed40-41b6-aa7d-f7b0ea474193/element/5562/attribute/value"
[09:56:31]  DATA                {}
[09:56:31]  RESULT              "20180131095532 comment"
```

And here is the log for WebdriverIO 4.10.1:

```
[09:59:13]  COMMAND     POST     "/wd/hub/session/147b25cc-91a2-4195-a3f3-524febd0b4e5/element"
[09:59:13]  DATA                {"using":"id","value":"product-comment-textarea"}
[09:59:13]  RESULT              {"ELEMENT":"5556"}
[09:59:13]  COMMAND     GET      "/wd/hub/session/147b25cc-91a2-4195-a3f3-524febd0b4e5/element/5556/property/value"
[09:59:13]  DATA                {}
```

Is it possible to introduce a workaround in WebdriverIO to use the old GET command, or should we complain to Appium to fix this?

christian-bromann:
[...]

PR #2580, opened 2018-02-01
Add another UnknownCommand indicator

ianrenauld:
Error message "Method has not yet been implemented" was added, has seen with Appium 1.7.2 while running on iOS

## Proposed changes

# To solve issue #2574 I added the message "Method has not yet been implemented" as a message indicating that this is an unknown command

## Types of changes

## Checklist

## Further comments
I wanted to add unit test for my change but i could not figure out how to run the mobile tests (which I assume I'd need, since I can only reproduce this issue in a specific case with Appium on iOS).  Anyway, I tried the change with my existing test that was erroring out and it passed with flying colors. For a one line change, I hope this is sufficient.


ianrenauld:
Added unit tests similar to the existing ones

PR #2600, opened 2018-02-14
Update getHTML.js - update return information

rickschubert:
I propose to change the documentation for `getHTML()` as it was unclear to me that it returns an array of strings when there are several elements with this selector. I only found this out by trial and error and could have saved myself a lot of time if I would have known that it returns an array (I thought it would only return the string of the first element found with this selector).


christian-bromann:
Closing due to inactivity

PR #2617, opened 2018-02-26
Revert "Add process.stdout.isTTY"

### LINKED ISSUE #2609: Cannot pipe test specs on stdin unless stdout is a tty?
## The problem

You can't provide a list of specs on `stdin` unless `stdout` is a TTY.

## Environment

* WebdriverIO version: v4.10.2
* Node.js version: v8.9.3
* wdio testrunner, running synchronous tests

## Details

I've been pulling my hair out for about a week trying to figure out this problem, and finally managed to track it down today.  I don't know if this is by design or not, but it's had me going insane.

I finally managed this morning to narrow the problem down to this simple reproduction.

This works fine, and runs all the tests in the test directory:
`echo 'test/*.js' | wdio ./wdio.conf.js`

This, however, runs no tests at all:
`echo 'test/*.js' | wdio ./wdio.conf.js > log.txt`

I finally tracked this issue down to this code from [lib/cli.js](https://github.com/webdriverio/webdriverio/blob/d48df91e29ef90b603a326b8653a7619bcba03c7/lib/cli.js#L437), which led me to #2477.  It's not clear to me what that PR was intended to fix, but it seems to me that the current problem is much worse than the problem that this change was supposed to fix.

```
/**
 * if stdin.isTTY, then no piped input is present and launcher should be
 * called immediately, otherwise piped input is processed, expecting
 * a list of files to process.
 * stdin.isTTY is false when command is from nodes spawn since it's treated as a pipe
 */
if (process.stdin.isTTY || !process.stdout.isTTY) {
    launch()
} else {
```

If this really is the intended behavior, and you can't provide specs on stdin unless stdout is a TTY, then hopefully anyone else who runs into this problem will at least be able to find this ticket and not have to fight with it for days like I did..

`echo 'test/*.js' | wdio ./wdio.conf.js > log.txt`

jasonk:
## Proposed changes

This reverts the changes from PR #2477, which was described as "Adding !process.stdout.isTTY which will allow launch() to be ran when called from a node script. Node's child_process spawn is treated as a pipe so the original process.stdin.isTTY is false and launch() would never get called."

The reason I'm reverting it is that the original behavior was correct, since the `wdio` runner is documented as taking a list of spec files on stdin when stdin is not a TTY, it's completely expected that it would "block" when you run it in a way where it's stdin is not a TTY.  The current behavior however, is extremely broken, it's now impossible to pass spec files on stdin unless `process.stdout` is a TTY.

Here are some of the ways the original problem could be fixed in the calling code without making this change:
* [Run the testrunner programmatically instead of using spawn](http://webdriver.io/guide/testrunner/gettingstarted.html#Run-the-test-runner-programmatically)
* Ignore stdin when calling `spawn`: `stdio : [ 'ignore', null, null ]`
* Provide empty input when calling `spawn`: `input : ''`

This reverts commit 209fad10858a3b0c95226a8ea113f34495cc6455.

Resolves #2609.

## Types of changes

Note that the original change in #2477 was a breaking change (although it was marked as a bugfix), but this PR is simply restoring the documented behavior.

## Checklist

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -432,9 +432,8 @@ const launch = () => {
  * if stdin.isTTY, then no piped input is present and launcher should be
  * called immediately, otherwise piped input is processed, expecting
  * a list of files to process.
- * stdin.isTTY is false when command is from nodes spawn since it's treated as a pipe
  */
-if (process.stdin.isTTY || !process.stdout.isTTY) {
+if (process.stdin.isTTY) {
     launch()
 } else {
     let stdinData = ''


christian-bromann:
@WillBrock can you comment on that?

WillBrock:
I think we should just revert my original change. I'm not sure if other languages like PHP or python will have the same issue or not though.

PR #2823, opened 2018-07-26
Add docs for return values from checkViewport

ediblecode:
The existing docs only showed usage (parameters) of the visual regression
service methods, but didn't say anything about their return values. These
return values can be useful when asserting e.g.: `expect(result.isExactSameImage, errorMessage).to.be.true;` so should ideally be documented.

## Proposed changes

Documents the return values from `checkViewport`

## Types of changes

Documentations

## Checklist

- [n/a] I have added tests that prove my fix is effective or that my feature works

## Further comments

n/a


christian-bromann:
@ediblecode can you make sure to make a pull request for the service too?

ediblecode:
Yes, I will do. There's no site for that right, just as part of the readme?

christian-bromann:
> There's no site for that right, just as part of the readme?

Exactly

docs/guide/services/visual-regression.md
```diff
@@ -144,6 +144,35 @@ available:
...
+            misMatchTolerance: 0.01
+        })
+    }
+  };
+  ```
+
+* **isSameDimensions**  `Boolean` <br>
+  True if the *taken* and *reference* images have the same dimensions; otherwise false.
+
+* **isExactSameImage**  `Boolean` <br>
```
emilyrohrbough: This is great! Our project uses these values. I would make note that `isExactSameImage` can return `true`, even if `isSameDimensions` is false. We saw this happen when a element (for example a div) grew in size, but the initial screenshot's content and the new (larger) screenshot matched 100%.

ediblecode: Thanks @emilyrohrbough I'll add this in

christian-bromann:
@ediblecode could you please make a new PR against the v4 branch? Thanks

ediblecode:
@christian-bromann sure, will do when I get chance - probably week after next

PR #2841, opened 2018-08-13
bump rgb2hex dependency to 0.1.9

radiovisual:
I still get the `isn't a valid rgb or rgba color` error when using a fresh install of webdriverio, because it is still using version `0.1.4` but since this error was fixed upstream in version `0.1.9` of rgb2hex, I figure it would be safe to use this as the minimal default for this dependency.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "webdriverio",
-  "version": "4.12.0",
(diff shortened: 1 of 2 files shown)


radiovisual:
@christian-bromann I updated the `package-lock.json`. 👍

PR #2884, opened 2018-08-28
Adding exclude spec functionality to cli arguments

PippoRaimondi:
## Proposed changes

Adding the capability to exclude spec(s) file(s) from command line, not a very big piece of work but can be handy... I've also tested the change locally (did some exploratory testing as well and it's ok)

## Types of changes

## Checklist

- [x ] I have added tests that prove my fix is effective or that my feature works
- [x ] I have added necessary documentation (if appropriate)

## Further comments


docs/guide/testrunner/organizesuite.md
```diff
@@ -137,6 +137,27 @@ $ grep -r -l --include "*.js" "myText" | wdio wdio.conf.js
 
 _**Note:** This will_ not _override the `--spec` flag for running a single spec._
 
+## Exclude Selected Tests
+
+When needed, if you need to exclude particular spec file(s) from a run, you can use the `--exclude` parameter (Mocha, Jasmine). For example if you want to exclude your login
```
christian-bromann: Why only Mocha and Jasmine? In Cucumber with `--spec` you can select feature files, why not be able to exclude them?

lib/utils/ConfigParser.js
```diff
@@ -272,6 +253,42 @@ class ConfigParser {
...
+     * options from cli argument
+     *
+     * @param  {String} filterObj  list of files in a string from
+     * cli argument
+     * @return {Boolean} exclude  true or false
+     */
+    setFilePathToFilterOptions (filterObj, exclude = false) {
+        const FILTER_TYPE = exclude ? 'exclude' : 'specs'
+
+        if (typeof filterObj === 'string') {
```
christian-bromann: Let's do
```js
if (typeof filterObj !== 'string') {
  return
}
// ...
```
and save an identation

lib/utils/ConfigParser.js
```diff
@@ -272,6 +253,42 @@ class ConfigParser {
...
+            const objsList = ConfigParser.getFilePaths(this._config[FILTER_TYPE])
+
+            for (let obj of objList) {
+                if (fs.existsSync(obj) && fs.lstatSync(obj).isFile()) {
+                    objs.add(path.resolve(process.cwd(), obj))
+                    continue
+                }
+                objsList.forEach(file => {
+                    if (file.match(obj)) {
+                        objs.add(file)
```
christian-bromann: in case of exclude=true shouldn't we not add the spec file here?

lib/utils/ConfigParser.js
```diff
@@ -157,29 +159,8 @@ class ConfigParser {
...
-            }
-
-            if (specs.size === 0) {
-                throw new Error(`spec file ${object.spec} not found`)
-            }
-
-            this._config.specs = [...specs]
-        }
+        this.setFilePathToFilterOptions(object.spec)
+        this.setFilePathToFilterOptions(object.exclude, true)
```
christian-bromann: Can we change the structure of this fn to
```js
this._config.specs = this.setFilePathToFilterOptions(object.spec)
this._config. exclude = this.setFilePathToFilterOptions(object. exclude)
```
this way it makes more sense and is better to test.

docs/guide/testrunner/organizesuite.md
PippoRaimondi: typo, will fix...

lib/utils/ConfigParser.js
PippoRaimondi: we want to add it since that's how it will decide what files to exclude from the run

christian-bromann: right

lib/utils/ConfigParser.js
PippoRaimondi: we have to move the typeof condition up to the merge implementation to go down this path... I will push it so you see how it looks... I have mixed feelings about it...

PippoRaimondi: what's your thoughts? if it looks ok I will go down a similar route for the other PR...

lib/utils/ConfigParser.js
```diff
@@ -272,6 +256,44 @@ class ConfigParser {
...
+     * @param  {String} filterObj  list of files in a string from
+     * cli argument
+     * @return {Boolean} exclude  true or false
+     */
+    setFilePathToFilterOptions (filterObj, exclude = false) {
+        const FILTER_TYPE = exclude ? 'exclude' : 'specs'
+
+        const objs = new Set()
+        const objList = filterObj.split(/,/g)
+        const objsList = ConfigParser.getFilePaths(this._config[FILTER_TYPE])
```
christian-bromann: Instead of having to set `exclude=true/false` let's pass in `ConfigParser.getFilePaths(this._config[FILTER_TYPE])` as second parameter which makes the function more open.

lib/utils/ConfigParser.js
```diff
@@ -272,6 +256,44 @@ class ConfigParser {
...
+
+        for (let obj of objList) {
+            if (fs.existsSync(obj) && fs.lstatSync(obj).isFile()) {
+                objs.add(path.resolve(process.cwd(), obj))
+                continue
+            }
+            objsList.forEach(file => {
```
christian-bromann: How about do:
```js
paramFile = objsList.find(file => file.match(obj))
if (paramFile) {
  objs.add(paramFile)
}
```

lib/utils/ConfigParser.js
```diff
@@ -272,6 +256,44 @@ class ConfigParser {
...
+                if (file.match(obj)) {
+                    objs.add(file)
+                }
+            })
+        }
+
+        if (objs.size === 0) {
```
christian-bromann: I don't think we need this anymore given the changed structure of the method.

PippoRaimondi: it is still possible to get into this function with a non-existing file hence this condition is still necessary.

lib/utils/ConfigParser.js
PippoRaimondi: I like it, fixed.

lib/utils/ConfigParser.js
PippoRaimondi: ok, I was a bit bothered with all the exclude = true thing... I am happier with this approach.

lib/utils/ConfigParser.js
```diff
@@ -272,6 +258,42 @@ class ConfigParser {
...
+            throw new Error(`File ${filterObj} not found`)
+        }
+
+        return objs
+    }
+
+    isTypeOfString (obj) {
```
christian-bromann: this method is so simple that I don't think we need to make a class function out of it

lib/utils/ConfigParser.js
```diff
@@ -272,6 +258,42 @@ class ConfigParser {
         return this._capabilities
     }
 
+    /**
+     * sets config attribute with file paths from filtering
+     * options from cli argument
+     *
+     * @param  {String} filterObj  list of files in a string from
+     * cli argument
+     * @return {Boolean} exclude  true or false
```
christian-bromann: this needs to get updated

lib/utils/ConfigParser.js
```diff
@@ -272,6 +258,42 @@ class ConfigParser {
...
+     */
+    setFilePathToFilterOptions (filterObj, config) {
+        const objs = new Set()
+        const objList = filterObj.split(/,/g)
```
christian-bromann: instead of `objList` and `objsList` could we give more meaningful names?

lib/utils/ConfigParser.js
PippoRaimondi: yeah, a bit overhead... removed.

lib/utils/ConfigParser.js
PippoRaimondi: whoooops, good catch

lib/utils/ConfigParser.js
PippoRaimondi: changed, hopefully it will make more sense now.

PR #2922, opened 2018-09-01
Terminals without color support

WillBrock:
## Proposed changes

Terminals without color support are printing out ansi characters. 

1) FORCE_COLOR expects a value to be 0 for no color support but was getting `undefined`.

https://github.com/chalk/supports-color/blob/master/index.js#L19

2) In `wdio-local-runner/src/index.js` i'm not sure exactly why that is always forcing color output so I might be missing the reasoning behind it. 

Screenshot from Jenkins:

ansi-support

## Types of changes

## Checklist

### DIFF
--- a/packages/wdio-cli/src/launcher.js
+++ b/packages/wdio-cli/src/launcher.js
@@ -23,7 +23,7 @@ class Launcher {
         }
 
         this.interface = new CLInterface(config, specs)
-        config.runnerEnv.FORCE_COLOR = this.interface.hasAnsiSupport
+        config.runnerEnv.FORCE_COLOR = Number(this.interface.hasAnsiSupport)
 
         const Runner = initialisePlugin(config.runner, 'runner')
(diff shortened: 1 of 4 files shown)


packages/wdio-cli/src/launcher.js
```diff
@@ -23,7 +23,7 @@ class Launcher {
         }
 
         this.interface = new CLInterface(config, specs)
-        config.runnerEnv.FORCE_COLOR = this.interface.hasAnsiSupport
+        config.runnerEnv.FORCE_COLOR = Number(this.interface.hasAnsiSupport)
```
christian-bromann: `this.interface.hasAnsiSupport` is a boolean, why do we cast it to integer?

WillBrock: The supports color package specifically looks for a 0 and not false on FORCE_COLOR.

https://github.com/chalk/supports-color/blob/master/index.js#L19

PR #2871, opened 2018-09-12
Allow cli input when running debug command

christian-bromann:
_From @christian-bromann on June 20, 2018 8:39_

## The problem

Our cli interface is not allowing to do any inputs if wdio testrunner is running.

_Copied from original issue: webdriverio/v5#75_


christian-bromann:
The only thing missing here:
- unit tests
- fix error when attaching worker to DevTools app

christian-bromann:
> * fix error when attaching worker to DevTools app

I won't be able to fix this due to some weird behavior in DevTools frontend. See https://github.com/ChromeDevTools/devtools-frontend/issues/103. I believe this is the reason why Fibers get stuck as well.

PR #2963, opened 2018-09-22
Fetch element by JS function

christian-bromann:
## The problem

Sometimes given selector strategies won't work and you need to use some JS to fetch an element. You can already use that via `browser.execute` in v4 today. However it won't convert that element into an element object.

## Proposal

Allow to pass in a function to the `$` and `$$` commands which will fetch an element and wrap it into an element object.


christian-bromann:
@WillBrock addressed both issues. Anything left?

PR #2965, opened 2018-09-23
Add support for find by attribute and content

gztomas:
## Proposed changes

Add support for selecting elements by attribute as well as content.
Before this change it was only possible to select some element by a given tag name, id or class and a certain content, like `div#main*=Hello`.
After this change it will be also possible to select by attribute, like `div[data-test-name="main"]*=Hello`

### Motivation
In our project we are using `data-test-name` attributes for marking elements that will be consumed by our tests. Since we are using [styled components](https://github.com/styled-components) which already generate random class names for elements, we avoid polluting the class attribute by using this specific attribute for tests element targeting.

## Types of changes

## Checklist

## Further comments

I have taken this as an opportunity for removing some repeated code, so now all the id class or attribute + content cases are handled by a single regular expression.

### DIFF
--- a/packages/webdriverio/src/utils.js
+++ b/packages/webdriverio/src/utils.js
@@ -43,12 +43,12 @@ export const findStrategy = function (value, isW3C) {
                value.indexOf('*/') === 0) {
         using = 'xpath'
 
-    // use link text strategy if value startes with =
+    // use link text strategy if value starts with =
     } else if (value.indexOf('=') === 0) {
         using = 'link text'
         value = value.slice(1)
 
-    // use partial link text strategy if value startes with *=
+    // use partial link text strategy if value starts with *=
     } else if (value.indexOf('*=') === 0) {
         using = 'partial link text'
         value = value.slice(2)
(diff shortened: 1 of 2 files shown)


gztomas:
@christian-bromann please let me know if you need anything from my side to move this forward.

packages/webdriverio/src/utils.js
```diff
@@ -86,50 +86,54 @@ export const findStrategy = function (value, isW3C) {
...
+            /^([a-z0-9]*)/,
+            // optional . or # + class or id
+            /(?:(\.|#)(-?[_a-zA-Z]+[_a-zA-Z0-9-]*))?/,
+            // optional [attribute-name="attribute-value"]
+            /(?:\[(-?[_a-zA-Z]+[_a-zA-Z0-9-]*)(?:=(?:"|')([a-zA-z0-9\-_. ]+)(?:"|'))?\])?/,
+            // *=query or =query
+            /(\*)?=(.+)$/,
+        ].map(rx => rx.source).join('')))
+
+        if (match) {
```
christian-bromann: If match is `null` then `using` would be `css selector` and value something odd not useful. I would propose to throw an error if no proper using/value pair was found.

gztomas: @christian-bromann please consider that is an inherited behavior, not changing with this PR.

None of these conditions matching would be equivalent to `match` being `null` in the proposed changes:
https://github.com/webdriverio/webdriverio/blob/5a94587926998abed8b78418573b67df712061fe/packages/webdriverio/src/utils.js#L98
https://github.com/webdriverio/webdriverio/blob/5a94587926998abed8b78418573b67df712061fe/packages/webdriverio/src/utils.js#L106
https://github.com/webdriverio/webdriverio/blob/5a94587926998abed8b78418573b67df712061fe/packages/webdriverio/src/utils.js#L118

I think that throwing when `match` is `null` would break the CSS selector use case for `findStrategy` - e.g.`findStrategy('#id')`. Unit tests confirmed locally that thought for me. What do you think?

christian-bromann: makes sense

gztomas: Thanks! Is there any chance of having this for `v4`?

christian-bromann: You need to make a separate PR to the v4 branch

PR #2995, opened 2018-10-10
Update babel.md

wswebcreation:
Change to correct hook implementation for jasmine / CucumberJS


wswebcreation:
@christian-bromann 

I would agree on that, this PR is just for fixing the docs for V4, because using the `before`-hook isn't working (Or I'm doing something wrong).

Maybe just use Babel 7 and `@babel/register` in the docs of V5

christian-bromann:
> I would agree on that, this PR is just for fixing the docs for V4

Then this PR should be point to the v4 branch

> Maybe just use Babel 7 and `@babel/register` in the docs of V5

That is the plan

wswebcreation:
I'll create a new PR for both branches

PR #2999, opened 2018-10-11
chore(templates): add git templates

wswebcreation:
## Proposed changes

This PR adds GitHub templates like in the image below. It will help the developers/maintainers of WebdriverIO by stearing people to the right template to provide all the needed info related to there bug/documentation/feature request/questions

image

The above will be shown when a user clicks on the new issue button
image


wswebcreation:
We talked about it a few weeks ago and it was easy to implement

PR #3026, opened 2018-10-28
Add browser question to questionnaire

### LINKED ISSUE #3025: Add 'browsers' to config wizard options
I usually run my tests using Chrome, due it having really great support via Chromedriver. That means every time I create a new project, I need to open the wdio.conf.js file after creation and change 'firefox' to 'chrome'.

I think it'd be useful to allow for browsers to be selected when going through the configuration wizard. You can select multiple browsers, similar to how services are done.

screen shot 2018-10-26 at 5 09 28 pm

haven't thought of any alternatives, aside from not making the change.

**Additional context**
What browsers should we include in the options by default? Chrome and Firefox makes sense to me. Maybe IE/Edge? Safari has poor webdriver support in my experience, so maybe leave that out?

If this sounds good, I'd like to take a stab at writing the code to make it happen. Seems like I could copy what's being done for services/reporters pretty easily.

aberonni:
## Proposed changes

This is an attempt to close #3025 - adding browser options to `wdio config`.

## Types of changes

## Checklist

## Further comments

I would've like to test this but can't figure out how to do so. Not sure whether the ejs syntax is correct.

### DIFF
--- a/packages/wdio-cli/src/config.js
+++ b/packages/wdio-cli/src/config.js
@@ -39,6 +39,11 @@ export const SUPPORTED_RUNNERS = [
     ' lambda - https://www.npmjs.com/package/wdio-lambda-runner'
 ]
 
+export const SUPPORTED_BROWSERS = [
+    'firefox',
+    'chrome'
+]
+
 const LOG_LEVELS = ['trace', 'debug', 'info', 'warn', 'error']
 
 export const USAGE = `
@@ -51,7 +56,7 @@ Usage: wdio repl <browserName>
 config file defaults to wdio.conf.js
 The [options] object will override values from the config file.
 An optional list of spec files can be piped to wdio that will override configured specs.
-Same applies to the exclude option. It can take a list of specs to exclude for a given run 
+Same applies to the exclude option. It can take a list of specs to exclude for a given run
 and it also overrides the exclude key from the config file.`
 
 export const CONFIG_HELPER_INTRO = `
@@ -148,6 +153,12 @@ export const CLI_PARAMS = [{
 }]
 
 export const QUESTIONNAIRE = [{
+    type: 'checkbox',
+    name: 'browsers',
+    message: 'On which browsers will you be testing?',
+    choices: SUPPORTED_BROWSERS,
+    default: SUPPORTED_BROWSERS
+}, {
     type: 'list',
     name: 'runner',
     message: 'Where should your tests be launched',
--- a/packages/wdio-cli/src/templates/wdio.conf.tpl.ejs
+++ b/packages/wdio-cli/src/templates/wdio.conf.tpl.ejs
@@ -73,14 +73,14 @@ exports.config = {
     // Sauce Labs platform configurator - a great tool to configure your capabilities:
     // https://docs.saucelabs.com/reference/platforms-configurator
     //
-    capabilities: [{
+    capabilities: [<% for(var i=0; i < service.browsers.length; i++) { %>{
         // maxInstances can get overwritten per capability. So if you have an in-house Selenium
-        // grid with only 5 firefox instances available you can make sure that not more than
+        // grid with only 5 <%= services.browsers[i] %> instances available you can make sure that not more than
         // 5 instances get started at a time.
         maxInstances: 5,
         //
-        browserName: 'firefox'
-    }],
+        browserName: '<%= services.browsers[i] %>'
+    },<% } %>],
     //
     // ===================
     // Test Configurations


klamping:
FYI, here's the list of possible browsers based on [the Selenium wiki](https://github.com/SeleniumHQ/selenium/wiki/DesiredCapabilities#used-by-the-selenium-server-for-browser-selection):
`android, chrome, firefox, htmlunit, internet explorer, iPhone, iPad, opera, safari`

Of those, I'd suggest just adding 'internet explorer' to the list of choices.

packages/wdio-cli/src/config.js
```diff
@@ -39,6 +39,11 @@ export const SUPPORTED_RUNNERS = [
     ' lambda - https://www.npmjs.com/package/wdio-lambda-runner'
 ]
 
+export const SUPPORTED_BROWSERS = [
```
christian-bromann: We should not limit this to two browser. Edge as well as Safari Technology Preview have excellent implementation of latest WebDriver spec. Also why not include mobile?

packages/wdio-cli/src/config.js
```diff
@@ -148,6 +153,12 @@ export const CLI_PARAMS = [{
 }]
 
 export const QUESTIONNAIRE = [{
+    type: 'checkbox',
+    name: 'browsers',
+    message: 'On which browsers will you be testing?',
```
christian-bromann: This question should not come first. Depending on where the tests are being run (local or on a cloud) we can already make assumptions to what to ask. For example: if running tests locally we don't need to ask for platform and version however when running tests on SauceLabs we definitely want to do that.

packages/wdio-cli/src/config.js
klamping: For why not to include mobile, my thought is that mobile requires extra setup through Appium, which may not be obvious to newcomers who see mobile and think "sure, why not test on it if it's that easy".

klamping:
My main thought on this is to help newcomers/basic projects get set up a little easier. Sure, we could add the extras to it, but I think it's more helpful to just keep it simple and let folks know the basic browsers they can test on. 

This is similar to how services/reporters are configured. There are a lot of options that one can dive in to, but the wizard just helps you get the basics set up. I really think just setting up the browser name should be all the wizard takes care of.

packages/wdio-cli/src/config.js
christian-bromann: All environments require extra steps, see my response https://github.com/webdriverio/webdriverio/pull/3026#issuecomment-434001469

klamping:
Just to play devil's advocate, the wizard is already assuming folks have Firefox downloaded and will be using something to download geckodriver. Adding in Chrome as an option doesn't really assume more.

christian-bromann:
You are right. So the solution should be to set Safari as default if the user is running a Mac OS and Edge if running on Windows.

erwinheitzman:
This makes it seem like it installs the browsers for you while it doesn't, the only thing it does is adding it to the capabilities.
If we aren't installing anything and we are only adding the browserName to the capabilities then what is the actual benefit of having this other then it might cause confusion?
I like your idea @klamping I just think it wil confuse people when they add a browser in the setup and it won't run since it's not installed/set up (which is what the cli is for).
Looking forward to a reply :)

klamping:
These are all really good points. 

Here are some options I see:
- Leave config as is (w/o the change). It's worked well enough so far :)
- Add question after asking about services to install. If 'selenium-standalone' is chosen, then we can (i think) assume they'll have the browsers installed for them. If 'chromedriver' is chosen, then set the browser to be 'chrome'.
- Ask the question, then validate the responses match the browsers they have installed. This is probably the most liable to break, as having the browser installed doesn't mean they have their setup correct.

christian-bromann:
I would like to ensure that the simplest setup works (Edge on Win and Safari on Mac) because we can assume that both browser and driver are setup correctly. So if the user marks local testing as answer we just set one capability based on his OS so that he can immediately run a test. For additional browser we should have good documentation on what is the simples way to add them.

aberonni:
Just to chip in and add my $0.02

> If we aren't installing anything and we are only adding the browserName to the capabilities then what is the actual benefit of having this other then it might cause confusion?

I wanted to add this because I usually just work with chrome and chromedriver, so I would select chromedriver as a service during the questions, but then it felt weird that I had to go and change the browser from firefox to chrome.

> Add question after asking about services to install. If 'selenium-standalone' is chosen, then we can (i think) assume they'll have the browsers installed for them. If 'chromedriver' is chosen, then set the browser to be 'chrome'.

This is one of the options that @klamping proposes above, and the one I think would make the most sense. I'd be happy to try and implement this if we all agree that this would be useful. We should probably build this on top of @christian-bromann 's suggestion:

> I would like to ensure that the simplest setup works (Edge on Win and Safari on Mac) because we can assume that both browser and driver are setup correctly.

klamping:
@christian-bromann my only concern with switching to safari is webdriver support. You mentioned that the newer safari offers much better support; any idea when that will be released as an official Safari update?

christian-bromann:
> any idea when that will be released as an official Safari update?

You can already by setting `'technologyPreview': True`

christian-bromann:
Closing as I think we agreed that this is not the solution for the indicated problem.

PR #2857, opened 2018-11-02
Implement watch task

christian-bromann:
_From @christian-bromann on August 24, 2018 8:47_

## The problem

Not sure if this is really important but in v4 we had a watch task which was not ported to v5. Let's reevaluate the importance of it and implement it if we see it to be useful.

_Copied from original issue: webdriverio/v5#115_


christian-bromann:
I will move forward with this. If there are any concerns in hindsight please let me know.

PR #3032, opened 2018-11-05
(chore): scope wdio packages

christian-bromann:
## Proposed changes

Scope wdio packages maintained in this repo. Biggest reason for this is not be able to differentiate between "official" maintained packages by the wdio projects committers and 3rd party contributors. Also with the v5 release this is a good opportunity to do this. It seems to be adopted by other big projects too, e.g. BabelJS or Jest.

## Types of changes

## Checklist

## Further comments


scripts/link.js
```diff
@@ -8,9 +8,17 @@ const shell = require('shelljs')
...
 const packages = shell.ls(packagesDir)
 shell.mkdir(nodeModulesDir)
+shell.mkdir(nodeModulesWDIODir)
 packages.forEach(
-    (pkg) => shell.ln('-s', path.join(packagesDir, pkg), path.join(nodeModulesDir, pkg))
+    (pkg) => shell.ln(
+        '-s',
+        path.join(packagesDir, pkg),
+        pkg.startsWith('wdio-')
+            ? path.join(nodeModulesWDIODir, pkg.replace('wdio-', ''))
```
webOS101: Is there a reason to leave the `wdio-` prefix on the directories or would this be a good time to clean them up by removing that prefix? It's certainly handy in this case, for figuring out where things live, but it seems extra noise in the file naming.

christian-bromann: No particular reason. I see the prefix in Babel and Jest too. I think it helps to differentiate which packages will be under the `@wdio` NPM org and which are packages without org.

PR #3033, opened 2018-11-05
Dependency Updates

christian-bromann:
## Proposed changes

- Upgrade to Babel 7
- minimise dependency requirements for `webdriver` package so that it can be compiled with Webpack and run in a browser.

## Types of changes

## Checklist


packages/webdriverio/src/commands/element/$.js
```diff
@@ -69,5 +68,6 @@ export default async function $ (selector) {
         return client
     }, prototype)
 
+    console.log(123, wrapCommand);
```
caoxu2000: was this for debug? should this be removed?

PR #3035, opened 2018-11-05
Increase default test timeout for all frameworks to 60 seconds

klamping:
## Proposed changes

The default timeout of 10 seconds for Mocha/Jasmine (20 for Cucumber) is almost always too short for the majority of WebdriverIO tests. 

Even the most basic of test suites can easily take over 10 seconds just loading the page.

This change updates the configuration template settings for the test framework, so they all allow for 60 seconds of test execution before timing out. This should work for the majority of test cases.

This also resolves an issue with error messaging in the following scenario (v5):
- I run a test that tries to find a non-existent element on the page
- The element command waits for 10 seconds for the element to appear
- The element is never found, but before that happens, the test framework timeout triggers, resulting in the following error:
image

The expected (more useful) error would be the "element not found" error that occurs when WebdriverIO triggers the error:
image

By increasing the default test framework timeout, we can ensure that the default `waitForTimeout` setting is triggered in such cases. 

Note: It'd be a good idea to update all the various places the `wdio.conf.js` file is shown in the documentation to reflect this change. I wanted to run it by the contributors first to get their thoughts before spending the time hunting down where these settings are documented.

## Types of changes

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works: **n/a I think**
- [ ] I have added necessary documentation (if appropriate) **see comments above**

## Further comments

### DIFF
--- a/packages/wdio-cli/src/templates/wdio.conf.tpl.ejs
+++ b/packages/wdio-cli/src/templates/wdio.conf.tpl.ejs
@@ -169,7 +169,8 @@ exports.config = {
     // Options to be passed to Mocha.
     // See the full list at http://mochajs.org/
     mochaOpts: {
-        ui: 'bdd'
+        ui: 'bdd',
+        timeout: 60000
     },<% }
     if(answers.framework === 'jasmine') { %>
     //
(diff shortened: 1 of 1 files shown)


BorisOsipov:
Agree with @klamping - 60 sec is reasonable value for the average default timeout.

PR #3052, opened 2018-11-19
Use exact package versions

Bamieh:
Using the carrot `^` for linking packages will cause unexpected bugs when trying to use older patch versions.

Currently I am trying to use version `webdriverio-beta.3` but the internal dependencies started using `beta.5` automatically because of the carrot, leading to unexpected errors.

I also updated `peerDependencies` where needed to accept all the variations of version `5` instead of only `5.0.*`.

### DIFF
--- a/packages/wdio-allure-reporter/package.json
+++ b/packages/wdio-allure-reporter/package.json
@@ -31,7 +31,7 @@
     "url": "https://github.com/webdriverio/webdriverio/issues"
   },
   "dependencies": {
-    "@wdio/reporter": "^5.0.0-beta.5",
+    "@wdio/reporter": "5.0.0-beta.5",
     "allure-js-commons": "^1.3.2"
   },
   "devDependencies": {
--- a/packages/wdio-applitools-service/package.json
+++ b/packages/wdio-applitools-service/package.json
@@ -35,7 +35,7 @@
   },
   "dependencies": {
     "@applitools/eyes.webdriverio": "^2.0.0",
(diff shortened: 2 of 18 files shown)


christian-bromann:
@Bamieh can you please provide a description for the reason of this change?

Bamieh:
@christian-bromann I was just updating the PR description as I noticed I have submitted it without a description.

Mainly using the carrot will cause previous versions o the beta to use the latest one internally, which might not be compatible with each other.

Currently I am installing `beta.3` but the dependency tree is showing `beta.5` in the internal dependencies as it is indicated by the carrot `^`.

Bamieh:
@christian-bromann 	that is a different issue :) I simply have not upgrade to the new beta version however `npm install` on the CI caused this hassle.

I'm getting this error:

> You can’t reassign a plugin after appling another plugin

although it was perfectly working fine perviously.

Bamieh:
From experience working with multiple monorepo projects, it is better to use the exact versions internally, to prevent packages using different versions internally than expected. Lerna automatically upgrades the version when publishing regardless.

When `beta.3` was published, it was tested against all `beta.3` versions; Nothing guarantees that it will work with future versions (even with patch / custom versions) so we should not link them internally with carrots `^` to not break the code of those who did not upgrade.

christian-bromann:
@Bamieh I have no objections pinning to an exact version. Would you mind to squash all commits into one commit?

Bamieh:
@christian-bromann 	done 👍

PR #2856, opened 2018-11-30
Debugging not working

christian-bromann:
_From @christian-bromann on August 24, 2018 9:18_

## The problem

Enable to use the debug utilities it should be as simple as passing `--inspect` to the wdio command and then use the `debugger` statement. There still seem to be some issues with connecting the runner process correctly to the DevTools app (Node.js v10). This should be fixed before release.

Eventual docs should be updated if something essential is changing.

_Copied from original issue: webdriverio/v5#116_


christian-bromann:
missing: unit tests

PR #3085, opened 2018-11-30
Allow to pick dc to run tests in

christian-bromann:
## Proposed changes

SauceLabs is expanding their services to other regions in the world. WebdriverIO wants to allow user to easily pick between DCs. 

## Types of changes

This change allows user to define the WebDriver backend region from sauce via the `region` property, e.g. 

```js
    user: process.env.SAUCE_USERNAME,
    key: process.env.SAUCE_ACCESS_KEY,
    region: 'eu',
```

Will run the test against https://ondemand.eu-central-1.saucelabs.com:443/wd/hub/. Any additional regions will be supported, e.g.

```js
    user: process.env.SAUCE_USERNAME,
    key: process.env.SAUCE_ACCESS_KEY,
    region: 'us-east-1',
```

Will run the test against https://ondemand.us-east-1.saucelabs.com:443/wd/hub/

## Checklist

### DIFF
--- a/examples/wdio.conf.js
+++ b/examples/wdio.conf.js
@@ -27,8 +27,10 @@ exports.config = {
     key:  'xxxxxxxxxxxxxxxx-xxxxxx-xxxxx-xxxxxxxxx',
     //
     // If you run your tests on SauceLabs you can specify the region you want to run your tests
-    // in via the `region` property.
-    region: 'eu', // for eu-central-1
+    // in via the `region` property. Available short handles for regions are:
+    // us: us-west-1 (default)
+    // eu: eu-central-1
+    region: 'us',
     //
     // ==================
     // Specify Test Files
--- a/examples/wdio.conf.js
+++ b/examples/wdio.conf.js
@@ -26,6 +26,10 @@ exports.config = {
     user: 'webdriverio',
     key:  'xxxxxxxxxxxxxxxx-xxxxxx-xxxxx-xxxxxxxxx',
     //
+    // If you run your tests on SauceLabs you can specify the region you want to run your tests
+    // in via the `region` property.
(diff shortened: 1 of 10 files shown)


examples/wdio.conf.js
```diff
@@ -26,6 +26,10 @@ exports.config = {
     user: 'webdriverio',
     key:  'xxxxxxxxxxxxxxxx-xxxxxx-xxxxx-xxxxxxxxx',
     //
+    // If you run your tests on SauceLabs you can specify the region you want to run your tests
+    // in via the `region` property.
+    region: 'eu', // for eu-central-1
```
wswebcreation: For now there's only one region, but this is not clear in the docs. By providing a string I would, as a user, assume I can also use `us/usa` and so on. What's the chance that there will be a new region before a new major release? 
Wouldn't a `boolean` be better here?

christian-bromann: > Wouldn't a `boolean` be better here?

I would say no because I would have to change the code again. The reason I introduced short regions is because some people aren't familiar with region names.

christian-bromann: I can also just remove the short version and have people to put in the full region name

wswebcreation: I would suggest to just add this in the WDIO-conf / documentation

```js
    // If you run your tests on SauceLabs you can specify the region you want to run your tests
    // in via the `region` property.
    // Regions: eu|us|..
    region: 'eu',
```

Like you have in the `logLevel`-docs

```js
    // Level of logging verbosity: trace | debug | info | warn | error
    logLevel: 'info',
```

wswebcreation:
> @wswebcreation I made it so you can pick `us` or `eu` and added some description to it

Thanks, makes it more clear 👍

PR #2867, opened 2018-12-04
Port touchAction command

christian-bromann:
See implementation details here:
https://github.com/webdriverio/webdriverio/blob/v4/lib/commands/touchAction.js

closing webdriverio/v5#6

### DIFF
--- a/packages/webdriverio/src/commands/element/touchAction.js
+++ b/packages/webdriverio/src/commands/element/touchAction.js
@@ -1,0 +1,145 @@
+/**
+ *
+ * The Touch Action API provides the basis of all gestures that can be automated in Appium.
+ * It is currently only available to native apps and can not be used to interact with webapps.
+ * At its core is the ability to chain together _ad hoc_ individual actions, which will then be
+ * applied to an element in the application on the device. The basic actions that can be used are:
+ *
+ * - press (pass selector or (x,y) or both)
+ * - longPress (pass selector or (x,y) or both)
+ * - tap (pass selector or (x,y) or both)
+ * - moveTo (pass selector or (x,y) or both)
+ * - wait (pass ms (as milliseconds))
+ * - release (no arguments)
+ *
+ * If you use the touchAction command with a selector you don't need to pass the selector to each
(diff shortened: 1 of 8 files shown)


packages/webdriverio/src/commands/element/touchAction.js
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * The Touch Action API provides the basis of all gestures that can be automated in Appium.
+ * It is currently only available to native apps and can not be used to interact with webapps.
+ * At its core is the ability to chain together _ad hoc_ individual actions, which will then be
+ * applied to an element in the application on the device. The basic actions that can be used are:
+ *
+ * - press (pass selector or (x,y) or both)
+ * - longPress (pass selector or (x,y) or both)
+ * - tap (pass selector or (x,y) or both)
+ * - moveTo (pass selector or (x,y) or both)
```
wswebcreation: Since Appium `1.8.0` `moveTo` commands for iOS and Android need to have absolute coordinates see [here](https://github.com/appium/appium/releases/tag/v1.8.0) and [here](https://gist.github.com/dpgraham/8b8584fd725d71561bde9e1ec863b237).

This means the `x|y` coordinate needs to be a `x|y` coordinate on the screen, not relative from the element.
To make this more generic for all screens it's advised to work with the screensize with for example `browser.windowHandleSize().value` and calculate the position you want to go based on that, see for example the [`Gestures.js`](https://github.com/webdriverio/appium-boilerplate/blob/master/tests/helpers/Gestures.js)-helper in the `appium-boilerplate`

packages/webdriverio/src/commands/element/touchAction.js
```diff
@@ -0,0 +1,62 @@
...
+        // simple touch action using selector and x y variables
+        // tap location is 30px right and 20px down relative from the center of the element
+        screen.touchAction({
+            action: 'tap', x: 30, y:20
+        })
+
+        // multi action on an element (drag&drop)
+        screen.touchAction([
+            'press',
+            { action: 'moveTo', x: 200, y: 0 },
```
wswebcreation: See `moveTo` remark

PR #2073, opened 2018-12-04
addCommand does not work for multiremote with wdio testrunner

christian-bromann:
## The problem

Custom commands, registered via the addCommand on the global multi *browser* instance in wdio testrunner, does not work for the respective browser instances.  Seems like the command is not registered on each single browser instance.

But it's working fine in standalone mode.

The following code fails in wdio-testrunner in multiremote mode:

```js
browserA = browser.select("browserA");
browserB = browser.select("browserB");

browser.addCommand("myCustomCommand", async function () {
  await this.pause(500);
});

browser.myCustomCommand() // works, executes custom commands on all browsers

browserA.myCustomCommand() // does not work -> browserA.myCustomCommand is not a function
browserB.myCustomCommand() // does not work -> browserB.myCustomCommand is not a function
```

I created a sample repo to reproduce (see below).

## Environment

* WebdriverIO version: 4.8.0
* Node.js version: 6.9.5
* [Standalone mode or wdio testrunner](http://webdriver.io/guide/getstarted/modes.html): wdio testrunner async

```
1) multiremote custom command called on browserA should work:
browserA.myCustomCommand is not a function
```
## Code To Reproduce Issue 
https://github.com/zinserjan/webdriverio-multiremote-experiments


christian-bromann:
There was indeed an issue not being able to register custom command on a multiremote instance. E.g.

```js
    matrix.addCommand('go', async function (url) {
        await this.url(url + '#custom')
        await this.pause(4000)
        return this.getCurrentUrl()
    })
```

Would still not attach the `go` function to the multiremote instance. This was not a problem when using standard WebdriverIO.

The behavior described by @zinserjan in the initial issue still exists though. After some testing I realised that this would be not a nice experiences because the return value of a multiremote command call is always an array of results. So you would never be able to use the same logic with a multiremote instance than with a standard one, e.g.

```js
browser.addCommand('myCustomCommand', async function (param) {
    const commandResult = await this.execute(() => 'foobar')
    return { param, commandResult }
})
```

`commandResult` would be `['foobar', 'foobar']` in multiremote but in normal just `foobar`. Therefor the custom command function can't be used in both modes.

I think this is ok and an expected behavior.

PR #3113, opened 2018-12-10
make sure jsonwp and w3c caps are formatted correctly (fix #3108)

### LINKED ISSUE #3108: Apply W3C capabilities correctly for non W3C compliant drivers
We define capabilities for the W3C and non W3C case, however if the customer has set W3C capabilities (with `alwaysMatch` etc) these caps won't work with non W3C compliant drivers. We should tweak https://github.com/webdriverio/webdriverio/blob/master/packages/webdriver/src/index.js#L23-L24 to make smart decisions on how we apply the caps

jlipps:
## Proposed changes

An initial stab at fixing #3108. I have not looked into tests or anything, just wanted to see if this is a good approach, @christian-bromann.

## Types of changes

This could potentially be a breaking change, but shouldn't be unless users are relying on weird things.

### DIFF
--- a/packages/webdriver/src/index.js
+++ b/packages/webdriver/src/index.js
@@ -25,7 +25,7 @@ export default class WebDriver {
         if (params.capabilities && params.capabilities.alwaysMatch) {
             // user passed in w3c-style caps (multi-layer object)
             w3cCaps = params.capabilities;
-            jsonwpCaps = params.capabilities.alwaysMatch);
+            jsonwpCaps = params.capabilities.alwaysMatch;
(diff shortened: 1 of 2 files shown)


jlipps:
@christian-bromann i'm getting stuck with adding a test for the new case. i haven't used jest before so maybe that's why. can you take a look at what i'm doing in the most recent commit and advise a way to have it actually work?

christian-bromann:
@jlipps I created a new PR where all tests seem to pass, wanna take a look: #3119 ?

PR #3115, opened 2018-12-10
fix logging target determination

jlipps:
we might wind up in a webpack environment where 'process'
is defined but fake, in which case we can't rely on
'process.release' existing

### DIFF
--- a/packages/wdio-logger/src/index.js
+++ b/packages/wdio-logger/src/index.js
@@ -1,6 +1,6 @@
 /**
  * environment check to allow to use this package in a web context
  */
-export default (typeof process !== 'undefined' && process.release.name === 'node')
+export default (typeof process !== 'undefined' &&  && typeof process.release !== 'undefined' && process.release.name === 'node')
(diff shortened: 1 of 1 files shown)


packages/webdriver/src/index.js
```diff
@@ -14,7 +14,12 @@ import AppiumProtocol from '../protocol/appium.json'
 export default class WebDriver {
     static async newSession (options = {}, modifier, proto = {}, commandWrapper) {
         const params = validateConfig(DEFAULTS, options)
-        logger.setLevel('webdriver', params.logLevel)
+
+        if (logger.setLevel) {
```
christian-bromann: Let's move this to the `@wdio/logger` package so that whoever consumes this expects a `setLevel` to be available if though the web version of it is a noop

jlipps: 👍

jlipps:
ok @christian-bromann I think this is good to go

Issue #3097, opened 2018-12-04
Add the ability to refetch an entire element chain to v5


abjerstedt:
**Is your feature request related to a problem? Please describe.**
I would like to add the ability to v5 for wdio to refetch an element via refetching the whole chain. 
Currently Given:
$('a').$('b').$('c')
If 'a' or 'b' fall stale, the system will try to fetch 'c', but it will never be successful as it is scoped to an element that is no longer in the DOM (and therefore the rest of the chain is stale).

**Describe the solution you'd like**
I will be adding something similar to the following code.
```
export default function refetch () {
    let currentElement = this;

    // Generate selector array
    let selectors = [];
    do {
        selectors.push(currentElement.selector);
        currentElement = currentElement.parent || {};
    } while (currentElement.elementId);
    // Iterator is currently set to Browser
    selectors.reverse();

    // Chain elements through the array
    return selectors.reduce( (elementPromise, selector) =>
        elementPromise.then((element) => element.$(selector)), Promise.resolve(currentElement));
}
```

abjerstedt:
@christian-bromann The question is whether i add this as an element command, and have the middleware call it (so it can be adhoc invoked by the wdio consumer); or just to the middleware. 
I propose the first.

christian-bromann:
> The question is whether i add this as an element command, and have the middleware call it (so it can be adhoc invoked by the wdio consumer); or just to the middleware.

I would suggest to do it only in the middleware because the wdio user doesn't know that an element might has become stale and why to ask the user to call an extra command if we can hide this internally.

PR #3118, opened 2018-12-10
wedbdriverio & webdriver: initial stale element handler work

abjerstedt:
## Proposed changes

I am adding a handler to refetch an element in the event of a stale element, and then retry the failed command.

This resolves https://github.com/webdriverio/webdriverio/issues/3097

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -11,7 +11,7 @@ module.exports = {
         '@babel/plugin-proposal-export-default-from',
         '@babel/plugin-proposal-class-properties',
         '@babel/plugin-proposal-optional-catch-binding'
-],
+    ],
     env: {
         development: {
             sourceMaps: 'inline',
--- a/babel.config.js
+++ b/babel.config.js
@@ -9,8 +9,9 @@ module.exports = {
     plugins: [
         '@babel/plugin-proposal-function-bind',
         '@babel/plugin-proposal-export-default-from',
-        '@babel/plugin-proposal-class-properties'
-    ],
+        '@babel/plugin-proposal-class-properties',
+        '@babel/plugin-proposal-optional-catch-binding'
+],
     env: {
         development: {
             sourceMaps: 'inline',
--- a/packages/webdriver/src/request.js
+++ b/packages/webdriver/src/request.js
@@ -94,6 +94,17 @@ export default class WebDriverRequest extends EventEmitter {
             }
 
             /**
+             *  stop retrying as this will never be successful.
+             *  we will handle this at the elementErrorHandler
+             */
+
+            if(error.message.includes("stale element reference")) {
+                log.warn(`Request encountered a stale element - terminating request`);
+                this.emit('response', { error })
+                return reject(error)
+            }
+
+            /**
              * stop retrying if totalRetryCount was exceeded or there is no reason to
              * retry, e.g. if sessionId is invalid
              */
--- a/packages/webdriver/tests/request.test.js
+++ b/packages/webdriver/tests/request.test.js
@@ -72,6 +72,24 @@ describe('webdriver request', () => {
             expect(req.emit).toBeCalledWith('response', { result: expectedResponse })
         })
 
+        it('should short circuit if request throws a stale element exception', async () => {
+            const req = new WebDriverRequest('POST', 'session/:sessionId/element')
+            req.emit = jest.fn()
+
+            const opts = Object.assign(req.defaultOptions, { 
+                uri: { path: '/wd/hub/session/foobar-123/element/some-sub-sub-elem-231/click' }, body: { foo: 'bar' } })
+
+            expect(req._request(opts)).rejects.toEqual(
+                new Error('stale element reference: element is not attached to the page document'))
+            expect(req.emit.mock.calls).toHaveLength(1)
+            expect(warn.mock.calls).toHaveLength(1)
+            expect(warn.mock.calls).toEqual([['Request encountered a stale element - terminating request']])
+
+            request.retryCnt = 0
+            warn.mockClear();
+            request.mockClear()
+        })
+
         it('should retry requests but still fail', async () => {
             const req = new WebDriverRequest('POST', '/session')
             req.emit = jest.fn()
--- a/packages/webdriverio/src/middlewares.js
+++ b/packages/webdriverio/src/middlewares.js
@@ -22,39 +22,21 @@ export const elementErrorHandler = (fn) => (commandName, commandFn) => {
                 `that wasn't found, waiting for it...`
             )
 
-            return await fn(commandName, () => {
-                /**
-                 * create new promise so we can apply a custom error message in cases waitForExist fails
-                 */
-                return new Promise((resolve, reject) => this.waitForExist().then(resolve, reject)).then(
-                    /**
-                     * if waitForExist was successful requery element and assign elementId to the scope
-                     */
-                    () => {
-                        return this.parent.$(this.selector).then(async (elem) => {
-                            this.elementId = elem.elementId
-                            try {
-                                return await fn(commandName, commandFn).apply(this, args)
-                            } catch(error) {
-                                if (error.message.includes("stale element reference")) {
-                                    const element = await refetchElement(this);
-                                    this.elementId = element.elementId;
-                                    this.parent = element.parent;
+            /**
+             * create new promise so we can apply a custom error message in cases waitForExist fails
+             */
+            try {
+                await this.waitForExist()
+            } catch (error) {
+                throw new Error(
+                    `Can't call ${commandName} on element with selector "${this.selector}" because element wasn't found`)
+            }
 
-                                    return await fn(commandName, commandFn).apply(this, args)
-                                }
-                                throw error;
-                            }
-                        })
-                    },
-                    /**
-                     * if waitForExist failes throw custom error
-                     */
-                    () => {
-                        throw new Error(`Can't call ${commandName} on element with selector "${this.selector}" because element wasn't found`)
-                    }
-                )
-            }).apply(this)
+            /**
+             * if waitForExist was successful requery element and assign elementId to the scope
+             */
+            const element = await this.parent.$(this.selector);
+            this.elementId = element.elementId
         }
 
         try {
--- a/packages/webdriverio/src/middlewares.js
+++ b/packages/webdriverio/src/middlewares.js
@@ -1,4 +1,5 @@
 import logger from '@wdio/logger'
+import refetchElement from './utils/refetchElement'
 
 const log = logger('webdriverio')
 
@@ -9,7 +10,7 @@ const log = logger('webdriverio')
  * @param  {Function} fn  commandWrap from wdio-sync package (or shim if not running in sync)
  */
 export const elementErrorHandler = (fn) => (commandName, commandFn) => {
-    return function (...args) {
+    return async function (...args) {
         /**
          * wait on element if:
          *  - elementId couldn't be fetched in the first place
@@ -21,7 +22,7 @@ export const elementErrorHandler = (fn) => (commandName, commandFn) => {
                 `that wasn't found, waiting for it...`
             )
 
-            return fn(commandName, () => {
+            return await fn(commandName, () => {
                 /**
                  * create new promise so we can apply a custom error message in cases waitForExist fails
                  */
@@ -30,9 +31,20 @@ export const elementErrorHandler = (fn) => (commandName, commandFn) => {
                      * if waitForExist was successful requery element and assign elementId to the scope
                      */
                     () => {
-                        return this.parent.$(this.selector).then((elem) => {
+                        return this.parent.$(this.selector).then(async (elem) => {
                             this.elementId = elem.elementId
-                            return fn(commandName, commandFn).apply(this, args)
+                            try {
+                                return await fn(commandName, commandFn).apply(this, args)
+                            } catch(error) {
+                                if (error.message.includes("stale element reference")) {
+                                    const element = await refetchElement(this);
+                                    this.elementId = element.elementId;
+                                    this.parent = element.parent;
+
+                                    return await fn(commandName, commandFn).apply(this, args)
+                                }
+                                throw error;
+                            }
                         })
                     },
                     /**
@@ -45,7 +57,18 @@ export const elementErrorHandler = (fn) => (commandName, commandFn) => {
             }).apply(this)
         }
 
-        return fn(commandName, commandFn).apply(this, args)
+        try {
+            return await fn(commandName, commandFn).apply(this, args);
+        } catch(error) {
+            if (error.message.includes("stale element reference")) {
+                const element = await refetchElement(this);
+                this.elementId = element.elementId;
+                this.parent = element.parent;
+
+                return await fn(commandName, commandFn).apply(this, args)
+            }
+            throw error;
+        }
     }
 }
 
--- a/packages/webdriverio/src/middlewares.js
+++ b/packages/webdriverio/src/middlewares.js
@@ -27,7 +27,7 @@ export const elementErrorHandler = (fn) => (commandName, commandFn) => {
              */
             try {
                 await this.waitForExist()
-            } catch (error) {
+            } catch {
                 throw new Error(
                     `Can't call ${commandName} on element with selector "${this.selector}" because element wasn't found`)
             }
--- a/packages/webdriverio/src/utils/refetchElement.js
+++ b/packages/webdriverio/src/utils/refetchElement.js
@@ -1,0 +1,16 @@
+export default async function refetchElement (currentElement) {
+    let selectors = [];
+
+    //Crawl back to the browser object, and cache all selectors
+    while(currentElement.elementId && currentElement.parent) {
+        selectors.push(currentElement.selector);
+        currentElement = currentElement.parent;
+    }
+    selectors.reverse();
+
+    // Beginning with the browser object, rechain
+    return selectors.reduce(async (elementPromise, selector) => {
+        const resolvedElement = await elementPromise;
+        return resolvedElement.$(selector);
+    }, Promise.resolve(currentElement));
+}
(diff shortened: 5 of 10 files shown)


abjerstedt:
I haven't added any tests yet, I want to make sure the proposed logic/ placement makes sense before i write the tests @christian-bromann

packages/webdriver/src/request.js
```diff
@@ -93,6 +93,15 @@ export default class WebDriverRequest extends EventEmitter {
                 return resolve(body)
             }
 
+            /**
+             *  stop retrying as this will never be successful.
+             *  we will handle this at the elementErrorHandler
+             */
+            if(error.message.includes("stale element reference")) {
```
erwinheitzman: Don't we have reference to exceptions like the StaleElementReferenceException somewhere?
Would be nice to have one place for reference and be able to do: instance of StaleElementReferenceException
Though I can imagine it being outside of the scope of this pull request, I think it's worth to mention ;)

packages/webdriverio/src/middlewares.js
```diff
@@ -45,8 +45,33 @@ export const elementErrorHandler = (fn) => (commandName, commandFn) => {
...
+                let selectors = [];
+                do {
+                    selectors.push(currentElement.selector);
+                    currentElement = currentElement.parent || {};
+                } while (currentElement.elementId);
+                // Iterator is currently set to Browser
+                selectors.reverse();
+
+                // Chain elements through the array
+                const newElement =  selectors.reduce((elementPromise, selector) =>
```
erwinheitzman: elementPromise is of the same type as the selector here, so isn't this simply a string to locate the element?

packages/webdriver/src/request.js
christian-bromann: This was in [v4](https://github.com/webdriverio-boneyard/v4/blob/master/lib/helpers/constants.js#L33-L73). I was thinking of eventually creating an module to handle and reference all kinds of WebDriver related errors but this is out of scope of this PR.

packages/webdriverio/src/middlewares.js
```diff
@@ -45,8 +45,33 @@ export const elementErrorHandler = (fn) => (commandName, commandFn) => {
             }).apply(this)
         }
 
-        return fn(commandName, commandFn).apply(this, args)
+        try {
+            return fn(commandName, commandFn).apply(this, args)
```
christian-bromann: I think this command returns a promise and wouldn't throw on a stale element error

packages/webdriverio/src/middlewares.js
```diff
@@ -45,8 +45,33 @@ export const elementErrorHandler = (fn) => (commandName, commandFn) => {
...
+        } catch (error ) {
+            if (error.message.includes(`stale element reference`)) {
+                let currentElement = this;
+
+                // Generate selector array
+                let selectors = [];
+                do {
```
christian-bromann: let's do a normal while loop here. Even though there is nothing wrong with do/while loops but it seems odd

packages/webdriverio/src/middlewares.js
abjerstedt: CURRENT (not this code) behavior is that the request retries will throw a warning twice, and then rejects with the exception. I'll try a .catch() and see if it works.

packages/webdriverio/src/middlewares.js
abjerstedt: Switched - however, i moved this off to its own command.

packages/webdriverio/src/middlewares.js
abjerstedt: Updated to no longer use promises.

packages/webdriver/tests/request.test.js
```diff
@@ -72,6 +72,17 @@ describe('webdriver request', () => {
...
 
+        it('should short circuit if request throws a stale element exception', () => {
+            const req = new WebDriverRequest('', '')
+            req.emit = jest.fn()
+
+            const opts = Object.assign(req.defaultOptions, { uri: { path: '/wd/hub/failing' } })
+            expect(req._request(opts, 2)).rejects.toEqual(new Error('Error: Could not send request'))
+
+            const
+
```
christian-bromann: All three lines above can be removed

packages/webdriverio/src/commands/element/refetch.js
```diff
@@ -0,0 +1,23 @@
...
+    //Crawl back to the browser object, and cache all selectors
+    while(currentElement.elementId && currentElement.parent) {
+        selectors.push(currentElement.selector);
+        currentElement = currentElement.parent;
+    }
+    selectors.reverse();
+
+    // Beginning with the browser object, rechain
+    const element = selectors.reduce((element, selector) =>
+        element.$(selector).waitForExist(), currentElement);
```
christian-bromann: You can't call async commands in a sync environment

packages/webdriverio/src/commands/element/refetch.js
```diff
@@ -0,0 +1,23 @@
+export default function refetch () {
```
christian-bromann: I would not make a new command for this. This needs to happen automatically once an element request fails.

abjerstedt: I can move it from a command to a webdriverio specific function, but i need it accessible to other areas besides the middleware. Are you alright with me moving it to something like /src/refetch.js and importing it to the specific files?

abjerstedt:
Still a WIP.

packages/webdriver/tests/request.test.js
abjerstedt: Sorry - this was still a WIP... i got called into a meeting before i was able to put a comment on it.

packages/webdriverio/src/middlewares.js
abjerstedt: Ok - after extensive testing, this needs to be a try/catch, as it does NOT return a promise.

packages/webdriverio/src/commands/element/refetch.js
abjerstedt: Ok moved to utils; and only invoked inside the middleware.

packages/webdriverio/src/commands/element/refetch.js
abjerstedt: Fixed.

packages/webdriver/tests/request.test.js
abjerstedt: I think this test is done.

packages/webdriverio/src/middlewares.js
```diff
@@ -32,7 +33,19 @@ export const elementErrorHandler = (fn) => (commandName, commandFn) => {
                     () => {
                         return this.parent.$(this.selector).then((elem) => {
                             this.elementId = elem.elementId
-                            return fn(commandName, commandFn).apply(this, args)
+                            try{
+                                return fn(commandName, commandFn).apply(this, args)
+                            } catch(error) {
+                                if (error.message.includes("stale element reference")) {
+                                    refetchElement.call(this).then(element => {
```
christian-bromann: `return` is missing, otherwise we would continue without having the stale element refetched

christian-bromann: for better unit-testability I would call it `refetchElement(this)` so in the test you can just insert the mock as param

packages/webdriverio/src/middlewares.js
christian-bromann: what does it return?

packages/webdriverio/src/utils/refetchElement.js
```diff
@@ -0,0 +1,27 @@
+export default function refetchElement () {
+    let currentElement = this;
+
+    return new Promise ( async () => {
```
christian-bromann: why this way? Just do `export default async function refetchElement () {` and you can use `await` right away

packages/webdriverio/src/utils/refetchElement.js
```diff
@@ -0,0 +1,27 @@
...
+            selectors.push(currentElement.selector);
+            currentElement = currentElement.parent;
+        }
+        selectors.reverse();
+
+        // Beginning with the browser object, rechain
+        const element = await selectors.reduce(async (elementPromise, selector) => {
+            const resolvedElement = await elementPromise;
+            const newElement = await resolvedElement.$(selector);
+            await newElement.waitForExist();
```
christian-bromann: `waitForExist` should not be necessary here, if you fetch the element the line before and it is not there, it would throw an error

packages/webdriverio/src/middlewares.js
```diff
@@ -32,7 +33,19 @@ export const elementErrorHandler = (fn) => (commandName, commandFn) => {
                     () => {
                         return this.parent.$(this.selector).then((elem) => {
                             this.elementId = elem.elementId
-                            return fn(commandName, commandFn).apply(this, args)
+                            try{
+                                return fn(commandName, commandFn).apply(this, args)
+                            } catch(error) {
+                                if (error.message.includes("stale element reference")) {
+                                    refetchElement(this).then(element => {
```
christian-bromann: still missing a `return`

christian-bromann: once the `refetchElement` method finished we can continue with the original `fn(commandName, commandFn)` but we definitely wait until all element were refetched

packages/webdriverio/src/middlewares.js
```diff
@@ -45,7 +58,19 @@ export const elementErrorHandler = (fn) => (commandName, commandFn) => {
             }).apply(this)
         }
 
-        return fn(commandName, commandFn).apply(this, args)
+        try {
+            return fn(commandName, commandFn).apply(this, args)
+        } catch(error) {
+            if (error.message.includes("stale element reference")) {
+                refetchElement(this).then(element => {
```
christian-bromann: same as above

packages/webdriverio/src/middlewares.js
abjerstedt: Why does this need a return? Theres a .then() off of the returned value of the function that then updates the parent and element id of the this context. We then return the function call itself with the updated context.

packages/webdriverio/src/middlewares.js
abjerstedt: We aren't returning the updated element.. we are returning the result of the command with the updated context.

packages/webdriverio/src/utils/refetchElement.js
abjerstedt: Updated.

packages/webdriverio/src/utils/refetchElement.js
abjerstedt: Removed.

packages/webdriverio/src/middlewares.js
abjerstedt: Updated

packages/webdriverio/src/middlewares.js
abjerstedt: I think the problem is indentation on the return...

abjerstedt: Done. I tried the async path, but it just turned into a rabbit hole.

packages/webdriverio/src/middlewares.js
abjerstedt: Done.

abjerstedt:
@christian-bromann @erwinheitzman Okay - other than if i need to add a UT specifically to hit the edge case of a stale element after a not found element, this is finally ready for actual review.

packages/webdriverio/src/middlewares.js
```diff
@@ -21,31 +22,35 @@ export const elementErrorHandler = (fn) => (commandName, commandFn) => {
...
-                        throw new Error(`Can't call ${commandName} on element with selector "${this.selector}" because element wasn't found`)
-                    }
-                )
-            }).apply(this)
+            /**
+             * create new promise so we can apply a custom error message in cases waitForExist fails
+             */
+            try {
+                await this.waitForExist()
+            } catch (error) {
```
CrispusDH: `(error)` is redundant if you don't use it

packages/webdriverio/src/utils/refetchElement.js
```diff
@@ -0,0 +1,16 @@
+export default async function refetchElement (currentElement) {
```
CrispusDH: Why do you use `default` keyword? I mean is it not a [bad practice](https://basarat.gitbooks.io/typescript/docs/tips/defaultIsBad.html)?

packages/webdriverio/src/middlewares.js
abjerstedt: Its not redundant when its necessary.  You HAVE to catch with a variable, even when you don't intend to use it.

abjerstedt: You have to catch an error with a variable even if you don't intend to use it.

packages/webdriverio/src/utils/refetchElement.js
abjerstedt: Why would it be bad practice?

CrispusDH: I added one link above. And there are several the same articles

abjerstedt: @christian-bromann You make the call. It seems the majority of the v5 code base uses export default, and this would deviate from it.

packages/webdriverio/src/middlewares.js
christian-bromann: I would say, let's add https://babeljs.io/docs/en/babel-plugin-proposal-optional-catch-binding and avoid the variable

packages/webdriverio/src/utils/refetchElement.js
christian-bromann: I am fine having it this way .. this file contains one utility function and only one so it is fine imo to directly export it

packages/webdriverio/src/middlewares.js
```diff
@@ -21,31 +22,35 @@ export const elementErrorHandler = (fn) => (commandName, commandFn) => {
...
         }
 
-        return fn(commandName, commandFn).apply(this, args)
+        try {
+            return await fn(commandName, commandFn).apply(this, args);
+        } catch(error) {
+            if (error.message.includes("stale element reference")) {
+                const element = await refetchElement(this);
+                this.elementId = element.elementId;
+                this.parent = element.parent;
```
christian-bromann: you are only updating the `elementId` of the current element and its parent but what happens if you have a higher level chain like `$('foo').$('bar').$('loo')`?

packages/webdriverio/tests/__mocks__/request.js
```diff
@@ -130,6 +130,60 @@ const requestMock = jest.fn().mockImplementation((params, cb) => {
...
+            body: {}
+        }, {});
+    }
+
+
+    /**
+     * Simulate a stale element
+     */
+
+    if (params.uri.path === `/wd/hub/session/${sessionId}/element/${genericSubSubElementId}/click`) {
```
christian-bromann: This if will never be reached because there is the same condition in the if in line 137 which will return

packages/webdriverio/src/middlewares.js
abjerstedt: I'm blowing away the entire ancestry, and that would be handled by the this.parent = element.parent.

packages/webdriverio/tests/__mocks__/request.js
abjerstedt: Good catch. Looks like i had a bad merge on my final rebase. Will fix.

abjerstedt: Fixed.

packages/webdriverio/src/middlewares.js
abjerstedt: Done.

packages/webdriverio/tests/middleware.test.js
```diff
@@ -0,0 +1,61 @@
+import logger from '@wdio/logger'
+import { remote } from '../src'
+import request from 'request'
+
+const { warn  } = logger()
+const waitForExist = require('../src/commands/element/waitForExist')
```
christian-bromann: I think we can use `jest` mocking capabilities here:
```js
jest.mock('../src/commands/element/waitForExist', () => .mockImplementation ( ( ) => { throw new Error(`Promise was rejected with the following reason`)})
```

packages/webdriverio/tests/middleware.test.js
```diff
@@ -0,0 +1,61 @@
...
+            .rejects.toThrow(`Can't call click on element with selector "#foo" because element wasn't found`)
+    });
+
+    it('should succesfully click on an element that falls stale after being refound', async () => {
+        waitForExist.default =
+            jest.fn().mockImplementation (( ) => { return true });
+
+        const elem = await browser.$('#foo')
+        const subElem = await elem.$('#subfoo')
+        const subSubElem = await subElem.$('#subsubfoo');
```
christian-bromann: this should fail the lint test because the project doesn't do `;`

CrispusDH: It might be better to add linter checks to CI for avoiding spending time on it during review

christian-bromann: @CrispusDH yes, I was assuming that eslint is already configured this way but was wrong about that. So @abjerstedt please ignore this, I will make a separate PR with a lint rule patch and fixing all semicolon appearances

abjerstedt: @CrispusDH There are. We were discussing this in the gitter channel, and it seems the rule isn't strictly enforced. Christian said we can ignore this comment for now, and enforce in a follow-up pr.

packages/webdriverio/tests/middleware.test.js
abjerstedt: Done.

babel.config.js
```diff
@@ -9,8 +9,9 @@ module.exports = {
     plugins: [
         '@babel/plugin-proposal-function-bind',
         '@babel/plugin-proposal-export-default-from',
-        '@babel/plugin-proposal-class-properties'
-    ],
+        '@babel/plugin-proposal-class-properties',
+        '@babel/plugin-proposal-optional-catch-binding'
+],
```
christian-bromann: indentation issue

abjerstedt: done.

PR #3152, opened 2018-12-19
chore: update examples in changelog

wswebcreation:
Updated the examples to be more clear

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,9 +40,9 @@ This version comes with a variety of technical changes that might affect the fun
     // ...
   ```
   ```js
-  // v5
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -42,8 +42,9 @@ This version comes with a variety of technical changes that might affect the fun
   ```js
   // v5
   await browser.url('https://duckduckgo.com/')
-  const elem = await browser.$('#search_form_input_homepage')
+  const elem = await $('#search_form_input_homepage')
```
christian-bromann: Can we revert this change? The example script here shows the standalone version. We could keep it like this but then we should remove all `await`s

PR #3156, opened 2018-12-19
wdio-cli: Fix output of reporters/services when generating config file

WillBrock:
## Proposed changes

When generating the wdio.conf.js file from that startup questions it's putting them in like below which does not work when it tries to require the package:

```reporters : ['@wdio/spec']```

It should be listed as the actual package name like:

```reporters : ['@wdio/spec-reporter']```

## Types of changes

## Checklist

### DIFF
--- a/packages/wdio-cli/src/templates/wdio.conf.tpl.ejs
+++ b/packages/wdio-cli/src/templates/wdio.conf.tpl.ejs
@@ -141,9 +141,7 @@ exports.config = {
     // your test setup with almost no effort. Unlike plugins, they don't add new
     // commands. Instead, they hook themselves up into the test process.
     <% if(answers.services.length) {
-    %>services: ['<%- answers.services.map(function(service) {
-        return service.slice(0, -8);
-    }).join("','") %>'],
+    %>services: ['<%- answers.services.join("','") %>'],
     <% } else {
     %>// services: [],<% } %>//
     // Framework you want to run your specs with.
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by christian-bromann** [comment]
@WillBrock actually we should change it the other way around. In v5 you can still apply `spec` if you want to use the `@wdio/spec-reporter`. If it can't find the "officially maintained" package it will check the legacy package name for all community collaborators `wdio-spec-reporters`. This has the advantage if someone stops maintaining a popular package we can just take it over by moving it into our org.

Would you mind changing it in a way that it puts the actual reporter/service name into the template?

File: packages/wdio-cli/src/templates/wdio.conf.tpl.ejs [hunk 1]
```diff
@@ -141,7 +141,11 @@ exports.config = {
     // your test setup with almost no effort. Unlike plugins, they don't add new
     // commands. Instead, they hook themselves up into the test process.
     <% if(answers.services.length) {
-    %>services: ['<%- answers.services.join("','") %>'],
+    %>services: ['<%- answers.services.map(function(service) {
+        const offset = service.startsWith('@wdio') ? 6 : 5;
+
+        return service.slice(0, -8).substr(offset);
```
christian-bromann: just do `return service.slice(service.startsWith('@wdio') ? 6 : 5, -8).substr(offset)`

packages/wdio-cli/src/templates/wdio.conf.tpl.ejs
```diff
@@ -156,9 +160,13 @@ exports.config = {
     // The only one supported by default is 'dot'
     // see also: http://webdriver.io/docs/dot-reporter.html
     <% if(answers.reporters.length) {
-    %>reporters: ['<%- answers.reporters.join("','") %>'],
+    %>reporters: ['<%- answers.reporters.map(function(reporter) {
+        const offset = reporter.startsWith('@wdio') ? 6 : 5;
+
+        return reporter.slice(0, -9).substr(offset);
```
christian-bromann: same

packages/wdio-cli/src/templates/wdio.conf.tpl.ejs
WillBrock: Updated.

packages/wdio-cli/src/templates/wdio.conf.tpl.ejs
WillBrock: Updated.

PR #2874, opened 2018-12-24
Create SSL cert for website

christian-bromann:
## The problem

Our website (even though just a static page) is not on https. Since it is free we should enable it.


christian-bromann:
I will go ahead and merge as this contains no functional changes.

PR #3174, opened 2018-12-25
webdriver: Change getTitle to getCurrentTitle

WillBrock:
## Proposed changes

In the change log and blog post it has ```title``` and ```getTitle``` changing to ```getCurrentTitle``` which I think makes sense since we now have ```getCurrentUrl``` but the protocol json files were still referencing ```getTtitle```.

This updates the json files along with other references of ```getTitle```.

## Types of changes

## Checklist


WillBrock:
Ok, I think I like that better. I'll open a new pr for that.

PR #2862, opened 2018-08-30
Native Support for Typings

christian-bromann:
_From @christian-bromann on July 10, 2018 12:31_

## The problem

Many users are using WebdriverIO with typescript. We should ensure that these people have a first level support for the new object structure and commands.

Support should be added for:
- webdriver
- webdriverio
- any service that adds commands

_Copied from original issue: webdriverio/v5#86_

PR #3218, opened 2018-12-31
Webdriver and WebdriverIO Typing

StephenABoyd:
## Proposed changes

Adding Typescript type definitions for v5, as well as an automatic generation of WebdriverIO definitions. Also, updating the JS docs to match the current parameters.

## Types of changes

## Checklist

## Further comments

So this is my first time creating type definitions, and I'm sure there are some commands that need to be added. This is building off of @christian-bromann PR [2862](https://github.com/webdriverio/webdriverio/pull/2862)

I have placed definition files in the ```webdriver``` and ```webdriverio``` packages, and set the ```types``` to point to them, so they will be automatically imported with the repo.

I have placed the definition files in my own project to test, so I know that what is there currently works.


StephenABoyd:
As I'm implementing this temporarily into my project I'm noticing a few things missing, or that automatic generation would be hard to catch. Many of the return types in the JS docs need to be set to ```any``` or add another parameter with what values could be returned in the object that gets returned. 

For example:
```
$(`//div').getLocation();
```

Would return an object with an x and y coordinate. That should be placed in the type definition as well.

StephenABoyd:
@christian-bromann To get these type definitions perfect, there is a lot of manual work. Some methods have to be created twice to handle the different parameters that can be passed in, then returned. 

For example: 
```getLocation```, if it takes the parameter of ```x``` or ```y``` then it will only return a number, otherwise it returns an object with ```{x:number,y:number}```. Ideally we would have two methods for this that handle which return Typescript should expect.

I'm trying to come up with a good way to auto generate this, and it seems like it may be a lot of manual cases in the generation script.

christian-bromann:
@StephenABoyd after giving them a second review, I am not sure if it makes sense to commit the generated type definitions. All we need is to have them within the NPM package, no? In this case it is only important to build them before the release.

StephenABoyd:
@christian-bromann Yeah that will work. I'll remove those

PR #3239, opened 2019-01-03
Unable pass null value to switchToFrame command for WebDriver

martomo:
## Proposed changes

* Resolves issue with `switchToFrame` command not accepting `null` as valid value.

## Types of changes

## Checklist

## Further comments

Issue was introduced with #3216, due to the fact `null` is actually properly validated (matched to `object` before).

Double checked the W3C standard specification and `switchToFrame` is the only other command which accepts `null` as value, next to `setWindowRect`.


packages/webdriver/protocol/appium.json
```diff
@@ -1047,7 +1047,7 @@
       "ref": "http://appium.io/docs/en/commands/interactions/touch/multi-touch-perform/",
       "parameters": [{
         "name": "actions",
-        "type": "object[]",
+        "type": "(array|object)[]",
```
christian-bromann: I don't think array is correct here. What would `array[]` mean?

packages/webdriver/src/utils.js
```diff
@@ -106,6 +106,16 @@ export function isValidParameter (arg, expectedType) {
     return true
 }
 
+/**
+ * get type of command argument
+ */
+export function getArgumentType (arg) {
+    if (Array.isArray(arg)) {
+        return 'array'
```
christian-bromann: `array` is not a type

packages/webdriver/tests/utils.test.js
```diff
@@ -44,11 +46,23 @@ describe('utils', () => {
...
         expect(isValidParameter(['1', false], '(number|string|object)[]')).toBe(false)
     })
 
+    it('getArgumentType', () => {
+        expect(getArgumentType(1)).toBe('number')
+        expect(getArgumentType(1.2)).toBe('number')
+        expect(getArgumentType(null)).toBe('null')
+        expect(getArgumentType('text')).toBe('string')
+        expect(getArgumentType({})).toBe('object')
+        expect(getArgumentType([])).toBe('array')
```
christian-bromann: Array of what?

packages/webdriver/protocol/appium.json
martomo: As explained in the commit message:
> Reason for supporting both is to address inconsistencies between documentation on
Appium (http://appium.io/docs/en/commands/interactions/touch/multi-touch-perform/)
which shows array of objects as JS example, when in tests the expected result is
an array of arrays. This way both options are possible.

packages/webdriver/src/utils.js
martomo: Do we want to allow both `[]` or `{}` when expected is supposed to be `object`?

packages/webdriver/tests/utils.test.js
martomo: I agree, this is somewhat contradicting with the `[]` suffixed definitions.

PR #3244, opened 2019-01-03
webdriverio: isDisplayed fix to return false on non existing elements

abjerstedt:
## Proposed changes

Currently isDisplayed() is being handled by the middleware, and forcing the command to waitForExist (1st problem for negative tests), as well as throwing an error if not found (2nd problem for negative tests). isDisplayed() will now properly immediately check and return false if not found or it doesn't exist.

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/packages/wdio-webdriver-mock-service/src/index.js
+++ b/packages/wdio-webdriver-mock-service/src/index.js
@@ -80,7 +80,7 @@ export default class WebdriverMockService {
         this.command.findElement().twice().reply(200, { value: elem2Response })
 
         this.command.elementClick(ELEMENT_ID).once().reply(200, { value: null })
-        this.command.elementClick(ELEMENT_ID).times(1).reply(500, { value: {
+        this.command.elementClick(ELEMENT_ID).times(4).reply(500, { value: {
             error: 'stale element reference error',
             message: 'stale element reference error'
         } })
--- a/packages/wdio-webdriver-mock-service/src/index.js
+++ b/packages/wdio-webdriver-mock-service/src/index.js
@@ -33,6 +33,7 @@ export default class WebdriverMockService {
          * register request interceptors for specific scenarios
          */
         global.browser.addCommand('waitForElementScenario', ::this.waitForElementScenario)
+        global.browser.addCommand('isDisplayedScenario', ::this.isDisplayedScenario)
         global.browser.addCommand('staleElementRefetchScenario', ::this.staleElementRefetchScenario)
     }
 
@@ -47,6 +48,17 @@ export default class WebdriverMockService {
         this.command.findElements().reply(200, { value: [elemResponse] })
         this.command.elementClick(ELEMENT_ID).once().reply(200, { value: null })
     }
+
+    isDisplayedScenario() {
+        this.nockReset()
+
+        const elemResponse = { 'element-6066-11e4-a52e-4f735466cecf': ELEMENT_ID }
+
+        this.command.findElement().times(2).reply(404, NO_SUCH_ELEMENT)
+        this.command.findElement().times(2).reply(200, { value: elemResponse })
+        this.command.isElementDisplayed(ELEMENT_ID).once().reply(200, { value: true })
+    }
+
 
     staleElementRefetchScenario () {
         this.nockReset()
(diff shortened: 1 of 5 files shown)


packages/wdio-webdriver-mock-service/src/index.js
```diff
@@ -58,7 +58,7 @@ export default class WebdriverMockService {
         this.command.findElement().twice().reply(200, { value: elem2Response })
 
         this.command.elementClick(ELEMENT_ID).once().reply(200, { value: null })
-        this.command.elementClick(ELEMENT_ID).times(4).reply(500, { value: {
+        this.command.elementClick(ELEMENT_ID).times(1).reply(500, { value: {
```
abjerstedt: It only fails one time cause, the request handler doesn't retry on stale elements.

packages/webdriverio/src/commands/element/isDisplayed.js
```diff
@@ -41,5 +41,5 @@
  */
 
 export default function isDisplayed() {
-    return this.isElementDisplayed(this.elementId)
+    return this.elementId ? this.isElementDisplayed(this.elementId) : false
```
abjerstedt: if called on "No Such Element" it will immediately return false.

tests/smoke.test.js
```diff
@@ -8,7 +8,8 @@ describe('smoke test', () => {
     it('should wait for elements if not found immediately', () => {
         browser.waitForElementScenario()
         const elem = $('elem')
-        assert.equal(elem.isDisplayed(), true)
+        //Element will be found
+        elem.click()
```
abjerstedt: If the click doesn't throw up... we are good!

packages/webdriverio/src/commands/element/isDisplayed.js
christian-bromann: This information could be outdated. I think it would make more sense to refetch the element at least once to see if it appeared in the time between it was fetched and the command called.

tests/smoke.test.js
christian-bromann: We should have an additional test for this use case.

packages/webdriverio/src/commands/element/isDisplayed.js
abjerstedt: Is this for people who cache the element prior to referencing it? In a lazyloading scenario it should be within milliseconds (depending on network latency) between the fetch and the check.

abjerstedt: done

tests/smoke.test.js
abjerstedt: I think done?

abjerstedt: Or did you mean you want me to write a test specifically for displayed?

packages/webdriverio/src/commands/element/isDisplayed.js
```diff
@@ -41,5 +41,8 @@
  */
 
 export default function isDisplayed() {
-    return this.isElementDisplayed(this.elementId)
+    return this.parent.$(this.selector).then((elem) => {
```
WillBrock: I think we can use async / await here.

christian-bromann: I agree

christian-bromann: Also I would propose to try/catch the isDisplayed command and return false if the command fails due to the error that the element does not exist

tests/smoke.test.js
```diff
@@ -8,7 +8,8 @@ describe('smoke test', () => {
     it('should wait for elements if not found immediately', () => {
         browser.waitForElementScenario()
         const elem = $('elem')
-        assert.equal(elem.isDisplayed(), true)
+        //Element will be found
+        assert.doesNotThrow(() => elem.click())
```
christian-bromann: This is now a complete different test, we should have two test cases for them

packages/webdriverio/src/commands/element/isDisplayed.js
abjerstedt: Done.

tests/smoke.test.js
abjerstedt: Ok. I added a negative test to ensure that isDisplayed returns false on a non existent element.

packages/wdio-webdriver-mock-service/src/index.js
christian-bromann: @abjerstedt doesn't matter, fact is that if you would do the same request over and over again, you would always get stale element exception

abjerstedt: Reverted.

PR #3246, opened 2019-01-04
Updating docs for wdio-junit-reporter

TuHuynhVan:
Updating doc, user need to install ```@wdio/reporter``` as well

## Proposed changes

Need to update one more line to guide user to install ```@wdio/repoter``` package to avoid error ```Error: Cannot find module '@wdio/reporter'```


christian-bromann:
We should add `@wdio/reporter` to the dependencies instead.

PR #3250, opened 2019-01-04
Update other commands which allow/expect 'null' as value.

martomo:
## Proposed changes

* Resolves issue with `switchToFrame` command not accepting `null` as valid value for JsonWireProtocol.
* Allow defining of `null` value for `moveToElement` (JsonWireProtocol) command and `background` (Appium) command.

## Types of changes

## Checklist

## Further comments

This should be it hopefully, in #3239 I completely forgot to check other protocols for supported values.

### DIFF
--- a/packages/webdriver/protocol/appium.json
+++ b/packages/webdriver/protocol/appium.json
@@ -748,7 +748,7 @@
       "ref": "http://appium.io/docs/en/commands/device/app/background-app/",
       "parameters": [{
         "name": "duration",
-        "type": "number",
+        "type": "(number|null)",
         "description": "timeout to restore app, if 'null' app won't be restored",
         "default": "null"
       }],
--- a/packages/webdriver/protocol/jsonwp.json
+++ b/packages/webdriver/protocol/jsonwp.json
@@ -219,7 +219,7 @@
       "ref": "https://github.com/SeleniumHQ/selenium/wiki/JsonWireProtocol#sessionsessionidframe",
       "parameters": [{
         "name": "id",
-        "type": "(string|number|object)",
+        "type": "(string|number|object|null)",
         "description": "Identifier for the frame to change focus to",
         "required": true
       }]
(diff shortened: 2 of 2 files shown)


packages/webdriver/protocol/appium.json
```diff
@@ -748,7 +748,7 @@
       "ref": "http://appium.io/docs/en/commands/device/app/background-app/",
       "parameters": [{
         "name": "duration",
-        "type": "number",
+        "type": "(number|null)",
```
christian-bromann: Where is this documented that it also accepts `null`?

packages/webdriver/protocol/jsonwp.json
```diff
@@ -219,7 +219,7 @@
       "ref": "https://github.com/SeleniumHQ/selenium/wiki/JsonWireProtocol#sessionsessionidframe",
       "parameters": [{
         "name": "id",
-        "type": "(string|number|object)",
+        "type": "(string|number|object|null)",
```
christian-bromann: What happen if you apply `null` here?

packages/webdriver/protocol/jsonwp.json
```diff
@@ -874,7 +874,7 @@
       "ref": "https://github.com/SeleniumHQ/selenium/wiki/JsonWireProtocol#sessionsessionidmoveto",
       "parameters": [{
         "name": "element",
-        "type": "string",
+        "type": "(string|null)",
```
christian-bromann: same question.

packages/webdriver/protocol/appium.json
martomo: See parameter description below:
> timeout to restore app, if 'null' app won't be restored

packages/webdriver/protocol/jsonwp.json
martomo: See command description above:
> Change focus to another frame on the page. If the frame `id` is `null`, the server should switch to the page's default content.

packages/webdriver/protocol/jsonwp.json
martomo: See parameter description below:
> opaque ID assigned to the element to move to, as described in the WebElement JSON Object, if not specified or is null, the offset is relative to current position of the mouse

PR #3271, opened 2019-01-07
Fix instance prototype and enable custom commands for elements

christian-bromann:
 - **WebdriverIO version:** 5.2.2
 - **Mode:** Standalone mode
 - **Node.js version:** 10.14.2
 - **NPM version:** 6.4.1
 - **Platform name and version:** Appium 1.10 and Android emulator

**Config of WebdriverIO**
Running Cucumber over standalone Wdio connected to Appium that talks to an Amdroid emulator 

**Describe the bug**
driver.getGeoLocation is not a function
driver.setGeoLocation is not a function

**To Reproduce**
```
async function crash(){
let a = await wdio.remote({
        host : "localhost",
        port : 4723,
        capabilities : {
            appiumVersion : '1.10.0',
            platformName: "Android",
            platformVersion: opts.platformVersion,
            deviceName: "Android Emulator",
            app: opts.app,
            automationName: "UiAutomator2",
            newCommandTimeout : 600
        }
    });
return a.getGeoLocation();
}
```

could be related to webdriverio/webdriverio#3120

### DIFF
--- a/packages/webdriver/src/index.js
+++ b/packages/webdriver/src/index.js
@@ -54,7 +54,10 @@ export default class WebDriver {
         params.capabilities = response.value.capabilities || response.value
         params.isW3C = isW3CSession(params.capabilities)
 
-        const prototype = Object.assign(getPrototype(params.isW3C, isChromiumSession(params.capabilities)), proto)
+        const isMobile = Boolean(params.capabilities.deviceName || params.capabilities.platformVersion)
+        const prototype = Object.assign(
+            getPrototype(params.isW3C, isChromiumSession(params.capabilities), isMobile),
+            proto)
         const monad = webdriverMonad(params, modifier, prototype)
         return monad(response.value.sessionId || response.sessionId, commandWrapper)
     }
--- a/packages/webdriver/src/utils.js
+++ b/packages/webdriver/src/utils.js
@@ -117,13 +117,29 @@ export function getArgumentType (arg) {
 /**
  * creates the base prototype for the webdriver monad
  */
-export function getPrototype (isW3C, isChrome) {
+export function getPrototype (isW3C, isChrome, isMobile) {
     const prototype = {}
     const ProtocolCommands = merge(
-        isW3C ? WebDriverProtocol : JsonWProtocol,
-        MJsonWProtocol,
-        AppiumProtocol,
-        isChrome ? ChromiumProtocol : {}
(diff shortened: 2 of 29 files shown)


packages/webdriver/src/utils.js
```diff
@@ -117,13 +117,29 @@ export function getArgumentType (arg) {
 /**
  * creates the base prototype for the webdriver monad
  */
-export function getPrototype (isW3C, isChrome) {
+export function getPrototype (isW3C, isChrome, isMobile) {
```
martomo: @christian-bromann Would it maybe make more sense to pass `capabilities` as an argument and determine `isMobile` and `isChrome` from within the getter?

Gilad-WT:
@christian-bromann are you planing to finish this or do you want me to try helping?

christian-bromann:
> are you planing to finish this or do you want me to try helping?

I will finish this as it requires some more changes to the core.

Gilad-WT:
@christian-bromann Ok Thanks

packages/webdriver/src/utils.js
christian-bromann: @martomo yeah, if we need to add one more `is*` as method param we have to rewrite it.

packages/webdriver/src/index.js
```diff
@@ -54,7 +54,10 @@ export default class WebDriver {
         params.capabilities = response.value.capabilities || response.value
         params.isW3C = isW3CSession(params.capabilities)
 
-        const prototype = Object.assign(getPrototype(params.isW3C, isChromiumSession(params.capabilities)), proto)
+        const isMobile = Boolean(params.capabilities.deviceName || params.capabilities.platformVersion)
```
martomo: @christian-bromann It makes sense to check for `deviceName` as this is for mobile only, however `platformVersion` is also returned by browsers on desktop, for example in Firefox.
https://github.com/webdriverio/webdriverio/blob/2b8dacc430ff4fa2d7c401e555846c31cfe979ab/packages/webdriver/tests/__fixtures__/geckodriver.response.json#L18

Couldn't we re-use the same check as for Appium to distinguish between mobile/desktop?
https://github.com/webdriverio/webdriverio/blob/2b8dacc430ff4fa2d7c401e555846c31cfe979ab/packages/webdriver/src/utils.js#L198

Issue #3102, opened 2018-12-05
logLevel in version 5 doesn't filter console logging


thesunny:
**Environment (please complete the following information):**
 - **WebdriverIO version:** 5.0.0-beta.12
 - **Mode:** Standalone
 - **If WDIO Testrunner, running sync/async:** async
 - **Node.js version: 8.11.1
 - **NPM version:** yarn 1.7.0
 - **Browser name and version:** Chrome 70
 - **Platform name and version:** OSX 10.14
 - **Additional wdio packages used (if applicable):** none

**Config of WebdriverIO**
```js
const config = {
  logLevel: "error",
  capabilities: {
    browserName: "chrome",
    chromeOptions: {
      args: ["--window-size-1024,768"],
    },
  },
  baseUrl: "http://192.168.150.234:3000",
}
```

**Describe the bug**
Set the logLevel to `error` but the driver is displaying all log levels to the console (e.g. DEBUG and INFO)

**To Reproduce**

```js
const browser = remote(config)
browser.url('https://www.google.com') // anything should do it
```

**Expected behavior**
Expect to run the commands without logging to the console

**Log**
Nothing specially relevant in the logs except that they are displaying at all when they shouldn't.

**Additional context**

> Please use the correct markdown

christian-bromann:
@thesunny I couldn't reproduce the issue. The only log content I see when switching to error was:

> 2018-12-05T23:54:36.172Z DEBUG wdio-config: wdio-sync found, running tests synchronous

which is a known "issue". Can you provide a sample script and your output?

thesunny:
@christian-bromann 

Sorry, the bug was more complicated than I thought and I needed to do more to reproduce it.

Here's a breakdown:

- If you create a new `remote` session, the `logLevel` works fine.
- If you create a `remote` session and then you `attach` to that session with the `remote`s config within the same running instance of nodejs, everything works fine
- If you create a `remote` session, then you save the config. In a separate nodejs instance, you `attach` to that session, then the `logLevel` doesn't work.

In order to reproduce, I had to create two separate programs which you should run with `babel-node`.

**First program**
```js
import { remote } from "webdriverio"

const config = {
  logLevel: "error",
  capabilities: {
    browserName: "chrome",
    chromeOptions: {
      args: ["--window-size-1024,768"],
    },
  },
}

async function app() {
  const browser = await remote(config)
  await browser.url("https://www.google.com")
  console.log({
    sessionId: browser.sessionId,
[...]

PR #3288, opened 2019-01-09
Validate before attaching session to webdriver (using multiremote)

### LINKED ISSUE #3257: [v5]Not able to run Multiremote with wdio test runner

 - **WebdriverIO version:** [5.2.0]
 - **Mode:** [WDIO Testrunner]
 - **Node.js version:** [8.11.4]
 - **NPM version:** [6.5.0]
 - **Browser name and version:** [Chrome  71]
 - **Platform name and version:** [Mac]
 - **Additional wdio packages used (if applicable):** [@wdio/selenium-standalone service]

**Config of WebdriverIO**
```
exports.config = {
    runner: 'local',
    //
    specs: [
        './qa/selenium-tests/soak/videoSoakLive.spec.js'
    ],
    //
    exclude: [
        // 'path/to/excluded/files'
    ],
    //
    maxInstances: 2,
    //
    capabilities: {
        maxInstances: 2,
        myChromeBrowser: {
            desiredCapabilities: {
                browserName: 'chrome'
            }
        },
        myFirefoxBrowser: {
            desiredCapabilities: {
                browserName: 'firefox'
            }
        }
    },
    //
    logLevel: 'trace',
    //
    deprecationWarnings: true,
    //
    bail: 0,
    //
    baseUrl: 'google.com/',
    //
    waitforTimeout: 10000,
    //
    connectionRetryTimeout: 90000,
    //
    connectionRetryCount: 3,
    //
    services: ['selenium-standalone'],
    //
    framework: 'mocha',
    //
    reporters: ['dot'],
    //
    mochaOpts: {
        ui: 'bdd',
        timeout: 60000
    },
}

```

**Describe the bug**
Not able to run Multiremote through wdio test runner. It throws error: 
`Error: Illegal key values seen in w3c capabilities: [browserA, browserB]`

**To Reproduce**
It is reproducible with official examples
1. Pull the wedriverio repo: https://github.com/webdriverio/webdriverio.git
2. `cd ./webdriverio`
3. `npm install`
4. `npm run setup`
5. Run the test: `npm run test:multiremote `

**Expected behavior**
2 browser have to be started

**Log**
Worker Error:
Error: Illegal key values seen in w3c capabilities: [browserA, browserB]
    at Request._callback /webdriverio/packages/webdriver/src/request.js:95:27)
    at Request.self.callback (/webdriverio/packages/webdriver/node_modules/request/request.js:185:22)
    at emitTwo (events.js:126:13)
    at Request.emit (events.js:214:7)
    at Request.<anonymous> (/webdriverio/packages/webdriver/node_modules/request/request.js:1161:10)
    at emitOne (events.js:116:13)
    at Request.emit (events.js:211:7)
    at IncomingMessage.<anonymous> (/webdriverio/packages/webdriver/node_modules/request/request.js:1083:12)
    at Object.onceWrapper (events.js:313:30)
    at emitNone (events.js:111:20)

**Additional context**

martomo:
## Proposed changes

* Validation of configuration when attaching multiremote to webdriver.
* Properly detect whether desired capabilities contains multiple browser configurations in wdio-runner.
* Set protocol and environment flags accordingly for multiremote session.
* Updated wdio example for multiremote configuration and standalone example in documentation.

Fixes #3257, #3342.

## Types of changes

## Checklist

## Further comments

Prior to validation it would return one or more of the following errors, depending on the configuration.

```
2019-01-08T22:01:30.142Z INFO webdriver: [GET] undefined://undefined/:undefinedundefined/session/1E8E4428-33B7-4C3B-B61D-E035BA14916B/url
2019-01-08T22:01:30.146Z ERROR webdriver: Request failed due to Error: Error: Invalid protocol: undefined:
```

```
2019-01-08T22:02:31.036Z INFO webdriver: [GET] http://undefined/:undefinedundefined/session/1E8E4428-33B7-4C3B-B61D-E035BA14916B/url
2019-01-08T22:02:31.057Z ERROR webdriver: Request failed due to Error: Error: getaddrinfo ENOTFOUND undefined undefined:80
```

```
2019-01-08T22:03:17.263Z INFO webdriver: [GET] http://0.0.0.0/:undefinedundefined/session/1E8E4428-33B7-4C3B-B61D-E035BA14916B/url
2019-01-08T22:03:17.274Z ERROR webdriver: Request failed due to Error: Error: connect ECONNREFUSED 0.0.0.0:80
```

Also with changes made in #3102, it would throw the following error when `logLevel` is not specified, this is now handled by validation.

```
(node:4136) UnhandledPromiseRejectionWarning: log.setLevel() called with invalid level: undefined
(node:4136) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:4136) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

### DIFF
--- a/examples/wdio/multiremote/mocha.test.js
+++ b/examples/wdio/multiremote/mocha.test.js
@@ -5,26 +5,32 @@ describe('multiremote example', () => {
 
     it('should login the browser', () => {
         const nameInput = $('.usernameInput')
-        nameInput.addValue('Browser A')
+        nameInput.addValue('Browser')
         browser.keys('Enter')
     })
 
     it('should post something in browserA', () => {
         const msgElemBrowserA = browserA.$('.inputMessage')
+
         msgElemBrowserA.addValue('Hey Whats up!')
         browser.pause(1000)
         browserA.keys('Enter')
 
         msgElemBrowserA.addValue('My name is Edgar')
+        browser.pause(1000)
         browserA.keys('Enter')
         browser.pause(200)
     })
 
     it('should read the message in browserB', () => {
-        const msgElemBrowserB = browserA.$('.inputMessage')
+        const msgElemBrowserB = browserB.$('.inputMessage')
+
         const chatLineBrowserB = browserB.$('.messageBody*=My name is')
+        chatLineBrowserB.waitForExist(5000)
+
         const message = chatLineBrowserB.getText()
-        const name = message.slice(11)
+        const name = message.split(' ').pop()
+
         msgElemBrowserB.addValue(`Hello ${name}! How are you today?`)
         browserB.keys('Enter')
         browser.pause(5000)
--- a/examples/wdio/multiremote/mocha.test.js
+++ b/examples/wdio/multiremote/mocha.test.js
@@ -4,8 +4,10 @@ describe('multiremote example', () => {
     })
 
     it('should login the browser', () => {
-        const nameInput = $('.usernameInput')
-        nameInput.addValue('Browser')
+        const browserANameInput = browserA.$('.usernameInput')
+        const browserBNameInput = browserB.$('.usernameInput')
+        browserANameInput.addValue('BrowserA')
+        browserBNameInput.addValue('BrowserB')
         browser.keys('Enter')
     })
 
--- a/packages/wdio-runner/src/utils.js
+++ b/packages/wdio-runner/src/utils.js
@@ -82,9 +82,24 @@ export function sanitizeCaps (caps) {
 }
 
 /**
- * initialise browser instance depending whether remote or multiremote is requested
+ * checks whether capabilities contains multiple browser configurations
+ * @param  {Object}  capabilities  desired session capabilities
+ * @return {Boolean}               true if multiple browsers configured
  */
-export async function initialiseInstance (config, capabilities, isMultiremote) {
+export function hasMultiremoteOptions (capabilities) {
+    return Object.keys(capabilities).length > 0 &&
+           Object.values(capabilities).every(browserConfig => {
+               return browserConfig.hasOwnProperty('capabilities')
+           })
+}
+
+/**
+ * initialise browser instance depending whether remote or multiremote is requested
+ * @param  {Object}  config        configuration of sessions
+ * @param  {Object}  capabilities  desired session capabilities
+ * @return {Promise}               resolves with browser object
+ */
+export async function initialiseInstance (config, capabilities) {
     /**
      * check if config has sessionId and attach it to a running session if so
      */
@@ -96,7 +111,7 @@ export async function initialiseInstance (config, capabilities, isMultiremote) {
         })
     }
 
-    if (!isMultiremote) {
+    if (!hasMultiremoteOptions(capabilities)) {
         log.debug('init remote session')
         config.capabilities = sanitizeCaps(capabilities)
         return remote(config)
--- a/packages/wdio-runner/src/utils.js
+++ b/packages/wdio-runner/src/utils.js
@@ -86,7 +86,7 @@ export function sanitizeCaps (caps) {
  * @param  {Object}  capabilities  desired session capabilities
  * @return {Boolean}               true if multiple browsers configured
  */
-export function hasMultiremoteOptions (capabilities) {
+export function isMultiremoteConfig (capabilities) {
     return (
         Object.keys(capabilities).length > 0 &&
         Object.values(capabilities).every(
@@ -113,7 +113,7 @@ export async function initialiseInstance (config, capabilities) {
         })
     }
 
-    if (!hasMultiremoteOptions(capabilities)) {
+    if (!isMultiremoteConfig(capabilities)) {
         log.debug('init remote session')
         config.capabilities = sanitizeCaps(capabilities)
         return remote(config)
--- a/packages/wdio-runner/src/utils.js
+++ b/packages/wdio-runner/src/utils.js
@@ -87,10 +87,12 @@ export function sanitizeCaps (caps) {
  * @return {Boolean}               true if multiple browsers configured
  */
 export function hasMultiremoteOptions (capabilities) {
-    return Object.keys(capabilities).length > 0 &&
-           Object.values(capabilities).every(browserConfig => {
-               return browserConfig.hasOwnProperty('capabilities')
-           })
+    return (
+        Object.keys(capabilities).length > 0 &&
+        Object.values(capabilities).every(
+            (browserConfig) => browserConfig.hasOwnProperty('capabilities')
+        )
+    )
 }
 
 /**
--- a/packages/webdriver/src/constants.js
+++ b/packages/webdriver/src/constants.js
@@ -90,3 +90,17 @@ export const DEFAULTS = {
         type: 'string'
     }
 }
+
+export const SESSION_DEFAULTS = Object.assign({}, DEFAULTS, {
+    // Capabilities are optional when attaching session,
+    // due to the fact drivers not always return capabilities
+    // in the initial (JsonWireProtocol) session response.
+    capabilities: {
+        type: 'object',
+        required: false
+    },
+    sessionId: {
+        type: 'string',
+        required: true
+    }
+})
--- a/packages/webdriver/src/index.js
+++ b/packages/webdriver/src/index.js
@@ -4,7 +4,7 @@ import { validateConfig } from '@wdio/config'
 
 import webdriverMonad from './monad'
 import WebDriverRequest from './request'
-import { DEFAULTS } from './constants'
+import { DEFAULTS, SESSION_DEFAULTS } from './constants'
 import { getPrototype, environmentDetector } from './utils'
 
 import WebDriverProtocol from '../protocol/webdriver.json'
@@ -78,17 +78,12 @@ export default class WebDriver {
      * allows user to attach to existing sessions
      */
     static attachToSession (options = {}, modifier, userPrototype = {}, commandWrapper) {
-        if (typeof options.sessionId !== 'string') {
-            throw new Error('sessionId is required to attach to existing session')
-        }
+        const params = validateConfig(SESSION_DEFAULTS, options)
+        logger.setLevel('webdriver', params.logLevel)
 
-        logger.setLevel('webdriver', options.logLevel)
-
-        options.capabilities = options.capabilities || {}
-        options.isW3C = options.isW3C || true
-        const prototype = Object.assign(getPrototype(options.isW3C), userPrototype)
-        const monad = webdriverMonad(options, modifier, prototype)
-        return monad(options.sessionId, commandWrapper)
+        const prototype = Object.assign(getPrototype(params.isW3C), userPrototype)
+        const monad = webdriverMonad(params, modifier, prototype)
+        return monad(params.sessionId, commandWrapper)
     }
 
     static get WebDriver () {
--- a/packages/webdriver/src/index.js
+++ b/packages/webdriver/src/index.js
@@ -1,5 +1,4 @@
 import logger from '@wdio/logger'
-import merge from 'lodash.merge'
 import { validateConfig } from '@wdio/config'
 
 import webdriverMonad from './monad'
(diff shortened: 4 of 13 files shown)


examples/wdio/multiremote/mocha.test.js
```diff
@@ -5,26 +5,32 @@ describe('multiremote example', () => {
 
     it('should login the browser', () => {
         const nameInput = $('.usernameInput')
-        nameInput.addValue('Browser A')
+        nameInput.addValue('Browser')
```
christian-bromann: The idea here is that one browser is logging in with "Browser A" and the other with "Browser B" let's keep this even though it wasn't quite right before.

packages/wdio-runner/src/utils.js
```diff
@@ -73,10 +73,25 @@ export function sanitizeCaps (caps) {
...
+ * checks whether capabilities contains multiple browser configurations
+ * @param  {Object}  capabilities  desired session capabilities
+ * @return {Boolean}               true if multiple browsers configured
+ */
+export function hasMultiremoteOptions (capabilities) {
+    return Object.keys(capabilities).length > 0 &&
+           Object.values(capabilities).every(browserConfig => {
+               return browserConfig.hasOwnProperty('capabilities')
+           })
+}
```
christian-bromann: How about this formatting:
```js
export function hasMultiremoteOptions (capabilities) {
    return (
        Object.keys(capabilities).length > 0 &&
        Object.values(capabilities).every(
            (browserConfig) => browserConfig.hasOwnProperty('capabilities')
        )
    )
}
```

packages/webdriver/src/constants.js
```diff
@@ -90,3 +90,21 @@ export const DEFAULTS = {
         type: 'string'
     }
 }
+
+export const SESSION_DEFAULTS = Object.assign({}, DEFAULTS, {
+    // Capabilities are optional when attaching session,
+    // due to the fact drivers not always return capabilities
+    // in the initial (JsonWireProtocol) session response.
```
christian-bromann: In which case does this apply? Drivers should always return the session capabilities.

packages/webdriver/src/index.js
```diff
@@ -66,17 +66,12 @@ export default class WebDriver {
      * allows user to attach to existing sessions
      */
     static attachToSession (options = {}, modifier, proto = {}, commandWrapper) {
-        if (typeof options.sessionId !== 'string') {
-            throw new Error('sessionId is required to attach to existing session')
-        }
-
-        logger.setLevel('webdriver', options.logLevel)
+        const params = validateConfig(SESSION_DEFAULTS, options)
```
christian-bromann: Can we get this working without validating the session again? It should not be required if the session was valid before.

packages/webdriver/src/constants.js
martomo: In this case I'm referring to chromedriver, which has "capabilities" stored as `value`.
https://github.com/webdriverio/webdriverio/blob/684551612eb0ff5b97f8a861d53e91e683bbe987/packages/webdriver/tests/__fixtures__/chromedriver.response.json#L3-L4

Whereas other drivers (e.g. geckodriver) return both `sessionId` and `capabilities` in the same object.
https://github.com/webdriverio/webdriverio/blob/684551612eb0ff5b97f8a861d53e91e683bbe987/packages/webdriver/tests/__fixtures__/geckodriver.response.json#L2-L4

So an example snippet for attaching a session would be as followed:
```js
const request = require('request')

async function initSession(browserName) {
    const capabilities = {
        browserName: browserName
    }
    const postData = {
        capabilities: {
            alwaysMatch: capabilities,
            firstMatch: [{}]
        },
        desiredCapabilities: capabilities
    };

    return new Promise(function (resolve, reject) {
        request({
            url: 'http://localhost:4444/wd/hub/session',
            method: 'POST',
            json: postData
        }, function (error, res, body) {
            if (!error && res.statusCode == 200) {
                resolve(body);
            } else {
                reject(error);
            }
        });
    });
}


(async () => {
    const response = await initSession('firefox')
    const session = await attach(response.value)
    console.log(await session.getUrl())
})().catch((e) => console.log(e))
```

packages/webdriver/src/index.js
martomo: Unless there is a way to distinguish between prior validated sessions and "raw" session responses which are passed to create new webdriver(io) instances (e.g. the snippet I've commented above).
Alternative would be not to expose `attach` or `WebDriver.attachToSession`.
Otherwise default values should be specified, this is not much different from piping default values like it was done before.

packages/wdio-runner/src/utils.js
martomo: Works for me!

examples/wdio/multiremote/mocha.test.js
christian-bromann: it is not about being generic. The use case for multibrowser is to simulate multiple "actors". So in this case e.g. player A and player B

martomo: @christian-bromann If you want I can change the label back to 'Browser A', for me it's all the same.

packages/webdriver/src/constants.js
martomo: Or in the case of internally calling the code, which simply does not have `capabilities` available (on top level).
For example when using the `MultiRemote` class which is a wrapper for multiple instances with different capabilities.

https://github.com/webdriverio/webdriverio/blob/5750a9ba4050ddf9a3b310734700d606782432e9/packages/webdriverio/src/index.js#L59-L62

packages/webdriver/src/index.js
martomo: As mentioned in the comment above, there are cases where the sessions might be valid before, but wrapped in another class (e.g. `MultiRemote`), which simply does not have these "default" values at their disposal.

This way we always have the required values when using an external session or a wrapper for multiple sessions.

packages/wdio-runner/src/utils.js
```diff
@@ -73,10 +73,27 @@ export function sanitizeCaps (caps) {
     }, {})
 }
 
+/**
+ * checks whether capabilities contains multiple browser configurations
+ * @param  {Object}  capabilities  desired session capabilities
+ * @return {Boolean}               true if multiple browsers configured
+ */
+export function hasMultiremoteOptions (capabilities) {
```
christian-bromann: How about calling this `isMultiremoteSession`?

packages/webdriver/src/index.js
christian-bromann: Ultimately I would love to have such an interface:

```js
const driver = await remote({ ... })
await driver.url('https://google.com')

await newDriver = attach(driver)
console.log(await newDriver.getUrl()) // outputs "Google"
```

martomo: As far I'm aware **webdriver** package is not aware of the drivers/browsers constructed by the (multi)remote functions in the **webdriverio** package.

Probably in a separate pull request the webdriverio `attach` function could be refactored to accept a driver/browser instance which is constructed by `remote` or `multiremote`.
Even then, you still would preferably validate the (required) configuration here in `WebDriver.attachToSession`.

packages/wdio-runner/src/utils.js
christian-bromann: how about `isMultiremoteCapability` then?

packages/webdriver/src/index.js
christian-bromann: Let's make changes to both methods then. One, in Webdriver so that we can do:
```js
const client = await WebDriver.newSession({
  path: '/',
  capabilities: { browserName: 'firefox' }
})
await client.navigateTo('https://www.google.com/ncr')

const client2 = await WebDriver.attachToSession(client)
console.log(await client2.getUrl()) // outputs "Google"
```
and make changes to the webdriverio attach method to be able to do the above example.

packages/wdio-runner/src/utils.js
martomo: I renamed it to `isMultiremoteConfig`, considering having capabilities both as argument and function name declaration is somewhat repetitive.

packages/webdriver/src/index.js
martomo: I'm not sure whether I understand what you are trying to suggest in the previous comment, isn't `client` and `client2` the exact same?
There is no need to create `client2`, as you already have `client`. Both are single instances pointing to the same browser session, I think you are confusing attaching of sessions with adding of instances to `MultiRemoteDriver`.

martomo: It's about attaching a (initial) session response from chrome/gecko/safari driver to WebDriver instance. Not attaching existing WebDriver instance to WebDriver instance, as you suggested in the previous `client`/`client2` example.

christian-bromann: why would you prefer to create a session manually instead of using the `remote` method?

martomo: Isn't that the purpose of `attachToSession`?

christian-bromann: @martomo no, it is to attach to an existing webdriver session .. but there is no reason to initiate the session manually when you can use the `remote` method.

christian-bromann:
Sorry I need to close this. Your PR helped me a lot though to figure out what is wrong with multiremote. I created a PR with a fix in #3348. Some of the changes here don't actually fix the issue and I wouldn't approve to merge them in.

martomo:
@christian-bromann I don't understand why this needed to be cancelled, I asked for feedback over the last couple days without response.

Especially considering it still provided some additional updates/fixes of issues reported #3257 in #3342, which are not handled in #3348.

Is it because I also included some other fixes of a previous cancelled pull request (#3291)?
Did I not properly write the unit tests to increase coverage of the codebase?

For future reference, it would be great to get more feedback than just "some of the changes don't actually fix", that way some people might be encouraged to try again next time when contributing.

tlabeeuw:
@christian-bromann Just tried #3348 in Webdriver.io 5.3.3, and it seems to have fixed the selenium case but Appium is still not working with a capability set that was working for 4x:
```
[0-0] 2019-01-16T19:31:32.031Z INFO webdriver: DATA { capabilities:
   { alwaysMatch:
      { appiumVersion: '1.10.0',
        platformName: 'iOS',
        platformVersion: '11.2',
        deviceName: 'iPad Air 2',
        automationName: 'XCUITest',
        app: '../ambulnz-emt.app.zip',
        useNewWDA: false,
        processArguments: '{"args": ["-Configuration", "bx-1881"]}' },
     firstMatch: [ {} ] },
  desiredCapabilities:
   { appiumVersion: '1.10.0',
     platformName: 'iOS',
     platformVersion: '11.2',
     deviceName: 'iPad Air 2',
     automationName: 'XCUITest',
     app: '../ambulnz-emt.app.zip',
     useNewWDA: false,
     processArguments: '{"args": ["-Configuration", "bx-1881"]}' } }
[0-0] 2019-01-16T19:31:31.977Z WARN webdriver: Request failed due to Illegal key values seen in w3c capabilities: [app, appiumVersion, automationName, deviceName, platformVersion, processArguments, useNewWDA]
```

christian-bromann:
> Especially considering it still provided some additional updates/fixes of issues reported in #3257 and #3342, which are not handled in #3348.

- #3257 is fixed, I was able to run a simple example with multiremote and wdio testrunner (in fact we have smoke tests for that now)
- #3342 also fixed

What specific updates/fixes are you talking about? The only thing I can see are the changes to the repo example which I would love to get merged for sure.

> Is it because I also included some other fixes of a previous cancelled pull request (#3291)?

This PR contained fixes for things that should rather be proposed with a different pull request. With that it is simpler for us to understand what the fix exactly does (like: one PR for one small problem). Additionally it included changes to things I would not agree on. For example: the utility `isMultiremoteConfig`. Even though capabilities is not a valid capability, just determine whether or not the session is a multiremote session by checking for a `capability` key is not good enough. Also the changes to the monad and how the multiremote prototype is generate were not really necessary.

Again, the PR was really useful for me to find the right knobs that allow to fix the original issue. I really appreciated it. I hope this doesn't discourage you from making more PRs.

PR #3290, opened 2019-01-09
Do mobile detection in webdriver package

christian-bromann:
## Proposed changes

Make mobile detection within the webdriver package since we need it to determine if the mobile protocols should get applied to the prototype or not

## Types of changes

Moved some code around.

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

[//]: # (_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._)

- [x] I have read the [CONTRIBUTING](https://github.com/webdriverio/webdriverio/blob/master/CONTRIBUTING.md) doc
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)

## Further comments

[//]: # (If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...)

PR #3291, opened 2019-01-09
Unify logic for mobile device detection in webdriver(io)

martomo:
## Proposed changes

* Define `isAndroid`, `isIOS`, `isMobile` flags in webdriver along with the `isW3C` flag instead of in webdriverio, having a unified approach for mobile checks.
* Check whether values passed into `attachToSession` are boolean instead of setting flags always to `true`, as `false` will pipe through to fallback/default value.
* Attach protocol to multiremote session by using `__propertiesObject__` reference.

## Types of changes

## Checklist


abjerstedt:
@martomo looks like you'll need to up UT coverage

christian-bromann:
@martomo I took some ideas from your PR here and put it into #3290

PR #3311, opened 2019-01-11
Fixes type generation for URL variables

ablok:
## Proposed changes

This PR fixes the webdriver type generation. When there were more URL variables than `sessionId` or `elementId` they would not end up in the typings as input parameters.

## Types of changes

## Checklist

## Further comments

Since the variables are in the url, they are always of type `string` and can not be optional.

### DIFF
--- a/scripts/type-generation/webdriver-generate-typings.js
+++ b/scripts/type-generation/webdriver-generate-typings.js
@@ -19,11 +19,15 @@ for (const [protocolName, definition] of Object.entries(PROTOCOLS)) {
 
     for (const [, methods] of Object.entries(definition)) {
         for (const [, description] of Object.entries(methods)) {
-            const { command, parameters = [], returns } = description
+            const { command, parameters = [], variables = [], returns } = description
+            const vars = variables
+                .filter((v) => v.name != "sessionId" && v.name != "elementId")
+                .map((v) => `${v.name}: string`)
             const params = parameters.map((p) => `${p.name}${p.required === false ? '?' : ''}: ${p.type.toLowerCase()}`)
+            const paramsAndVars = params.concat(vars);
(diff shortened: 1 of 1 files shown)


christian-bromann:
@StephenABoyd do you mind taking a look?

ablok:
Actually, just thought of something. Would variables come before parameters in the input parameters? Or the other way around? Or does one exclude the other?

christian-bromann:
> Would variables come before parameters in the input parameters?

Yes, here is an example:

```json
  "/session/:sessionId/element/:elementId/element": {
    "POST": {
      "command": "findElementFromElement",
      "description": "The Find Element From Element command is used to find an element from a web element in the current browsing context that can be used for future commands.",
      "ref": "https://w3c.github.io/webdriver/#dfn-find-element-from-element",
      "variables": [{
        "name": "elementId",
        "description": "the id of an element returned in a previous call to Find Element(s)"
      }],
      "parameters": [{
        "name": "using",
        "type": "string",
        "description": "a valid element location strategy",
        "required": true
      }, {
        "name": "value",
        "type": "string",
        "description": "the actual selector that will be used to find an element",
        "required": true
      }],
      "returns": {
        "type": "String",
        "name": "element",
        "description": "A JSON representation of an element object."
      }
    }
  },
```
If you want to call this command you would need to do this:
```js
// packages⁩/⁨webdriverio⁩/src/commands/element⁩/selectByAttribute.js
const normalized = `[normalize-space(@${attribute.trim()}) = "${value.trim()}"]`
    const optionElement = await this.findElementFromElement(this.elementId, 'xpath', `./option${normalized}|./optgroup/option${normalized}`)
```

ablok:
Switched the order so that variables will be the first input parameters.

PR #3317, opened 2019-01-14
Update GettingStarted.md

yogibaba2:
In getting started doc 'Generate Configuration File' with wdio, please update in to 'http://' instead 'http://localhost' as when running a spec url is getting added to 'http://localhost' and failing and making the beginners confuse.


christian-bromann:
`http://` is not a valid base url. I am not sure where your confusion comes from but this won't solve it.

yogibaba2:
Thanks for the quick response, i got the purpose of that. I wiil suggest that please add (your application base url) instead (just press enter).

Again thanks
A nice framework indeed.

PR #3319, opened 2019-01-14
Add missing typings

ablok:
## Proposed changes

- Adds type `DesiredCapabilities[]` to the typings and documentation
- Adds property `execArgv` to the typings and documentation
- Adds typings and documentation for the various hooks

## Types of changes

## Checklist

## Further comments

Please review thoroughly since I did some assumptions.
I added the missing `execArgv` property but have 3 questions:
- Did not add a default to a constant file since the default is null (I presume). Is that ok? I apparently also did not add a default for the `services` property that I added in a previous MR.
- I made it optional. Is that ok? Looking at the other properties, I am unclear on when it should be optional or not.
- I put it in the `webdriver.tpl.d.ts` file and not in the `webdriverio.tpl.d.ts` file. Is that ok?

### DIFF
--- a/docs/Options.md
+++ b/docs/Options.md
@@ -85,12 +85,6 @@ Count of request retries to the Selenium server
 Type: `Number`<br>
 Default: `3`
 
-### execArgv
-Node arguments to specify when launching child processes
-
-Type: `String[]`
-Default: `null`
-
 ---
 
 ## WDIO Options
@@ -182,6 +176,12 @@ reporters: [
     }]
 ]
 ```
+
+### execArgv
+Node arguments to specify when launching child processes
+
+Type: `String[]`
+Default: `null`
 
 ## Hooks
 
--- a/docs/Options.md
+++ b/docs/Options.md
@@ -45,7 +45,7 @@ Default: `null`
 Defines the capabilities you want to run in your WebDriver session. Check out the [WebDriver Protocol](https://w3c.github.io/webdriver/#capabilities) for more details. If you run an older driver that doesn't support WebDriver you need to apply the [JSONWireProtocol capabilities](https://github.com/SeleniumHQ/selenium/wiki/DesiredCapabilities) to successfuly run a session. Also a useful utility is the Sauce Labs [Automated Test Configurator](https://wiki.saucelabs.com/display/DOCS/Platform+Configurator#/)
 that helps you to create this object by clicking together your desired capabilities.
 
-Type: `Object`<br>
+Type: `Object`|`Object[]`<br>
 Default: `{ browserName: 'firefox' }`<br>
 
 **Example:**
--- a/docs/Options.md
+++ b/docs/Options.md
@@ -44,9 +44,10 @@ Default: `null`
 ### capabilities
 Defines the capabilities you want to run in your WebDriver session. Check out the [WebDriver Protocol](https://w3c.github.io/webdriver/#capabilities) for more details. If you run an older driver that doesn't support WebDriver you need to apply the [JSONWireProtocol capabilities](https://github.com/SeleniumHQ/selenium/wiki/DesiredCapabilities) to successfuly run a session. Also a useful utility is the Sauce Labs [Automated Test Configurator](https://wiki.saucelabs.com/display/DOCS/Platform+Configurator#/)
 that helps you to create this object by clicking together your desired capabilities.
+Note: `Object[]` is only used when using the wdio testrunner.
 
 Type: `Object`|`Object[]`<br>
-Default: `{ browserName: 'firefox' }`<br>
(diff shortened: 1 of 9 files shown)


examples/wdio.conf.js
```diff
@@ -301,6 +301,19 @@ exports.config = {
      */
     onComplete: function (exitCode, config, capabilities, results) {
     },
+    /**
+    * Gets executed when an error happens, good place to take a screenshot.
+    * @param {String} error message
+    */
+    onError: function(message) {
```
christian-bromann: This hook was not ported and should be removed.

ablok:
Can you please tell me:
-How do I determines if a property can be optional in the config file?
-What determines if there needs to be a default specified in the constants file.
-How do I determine if a property in the config (like `services` and `execArgv`) is part of webdriver or webdriverio?

docs/Options.md
```diff
@@ -45,7 +45,7 @@ Default: `null`
 Defines the capabilities you want to run in your WebDriver session. Check out the [WebDriver Protocol](https://w3c.github.io/webdriver/#capabilities) for more details. If you run an older driver that doesn't support WebDriver you need to apply the [JSONWireProtocol capabilities](https://github.com/SeleniumHQ/selenium/wiki/DesiredCapabilities) to successfuly run a session. Also a useful utility is the Sauce Labs [Automated Test Configurator](https://wiki.saucelabs.com/display/DOCS/Platform+Configurator#/)
 that helps you to create this object by clicking together your desired capabilities.
 
-Type: `Object`<br>
+Type: `Object`|`Object[]`<br>
```
christian-bromann: This is not quite right .. it should be just `Object` because it describes the webdriverio remote options. It should be moved to `WDIO Options`

ablok: I am not quite sure I understand. According to the documentation it should be an array of DesiredCapabilities. https://webdriver.io/docs/configurationfile.html
Can you elaborate.

christian-bromann: `desiredCapabilities` is not an option anymore. The option section you have changed belongs to `webdriverio` package options. Only if you run webdriverio with the wdio testrunner the `capability` should be an array.

christian-bromann: options for webdriver: https://github.com/webdriverio/webdriverio/blob/master/packages/webdriver/src/constants.js#L3-L92
options for webdriverio (as standalone package): same as webdriver
options for wdio: https://github.com/webdriverio/webdriverio/blob/master/packages/webdriverio/src/constants.js#L31

ablok: services option is missing from both ;)
But I guess it belongs to webdriverio, so I will add it there.

ablok: @christian-bromann And what about `execArgv`, in which one does that belong?

ablok: @christian-bromann Pushed some changes, hope this is better.

packages/webdriverio/src/constants.js
```diff
@@ -146,6 +153,20 @@ export const WDIO_DEFAULTS = {
             return true
         }
     },
+    /**
+     * set of WDIO services to use
+     */
+    services: {
+        type: 'string[]',
```
christian-bromann: this can also be an object when you provide an not published service, e.g.

```js
services: [new MyCustomService()]
```

christian-bromann: actually, see second example here: https://webdriver.io/docs/customservices.html

docs/Options.md
```diff
@@ -44,9 +44,10 @@ Default: `null`
 ### capabilities
 Defines the capabilities you want to run in your WebDriver session. Check out the [WebDriver Protocol](https://w3c.github.io/webdriver/#capabilities) for more details. If you run an older driver that doesn't support WebDriver you need to apply the [JSONWireProtocol capabilities](https://github.com/SeleniumHQ/selenium/wiki/DesiredCapabilities) to successfuly run a session. Also a useful utility is the Sauce Labs [Automated Test Configurator](https://wiki.saucelabs.com/display/DOCS/Platform+Configurator#/)
 that helps you to create this object by clicking together your desired capabilities.
+Note: `Object[]` is only used when using the wdio testrunner.
 
-Type: `Object`<br>
-Default: `{ browserName: 'firefox' }`<br>
+Type: `Object`|`Object[]`<br>
```
christian-bromann: As I mentioned, we should move this to WDIO options and keep `Object` as type. Under WDIO options we can put capabilities and mention to have it set as Object (for multiremote) or Object[]

christian-bromann: to clarify, 
__WebDriver Options__ has a capabilities option that has type `Object`
__WDIO Options__ has also a capabilities option from type `Object|Object[]` , first when run as multiremote

ablok:
I just pushed a fix :)

christian-bromann:
@ablok see #3329

PR #3381, opened 2019-01-22
Retries per specfile

bennieswart:
## Proposed changes

Add retries on a per-specfile basis, as discussed in #2484, and PR'd pre-v5 to v4 in #2556, now residing in https://github.com/webdriverio-boneyard/v4/pull/1.

Previously only test-level and suite-level retries were available, which
are fine in most cases. In any test which involves state, such as on a
server or in a db, the state may be left invalid once the test fails the
first time. Any subsequent retries may have no chance of passing due to
the invalid state they would have to start with.

A new browser instance is created for each specfile, which makes this an
ideal place to hook and setup any other states (server, db). Retries on
this level would mean that the whole setup process could simply be
repeated as if it were for a new specfile, with the results of the
failing specfile (that is to be retried) being discarded.

## Types of changes

## Checklist

## Further comments

The above referenced issues serve as required background as some previous discussion has already taken place.
This PR contains all (or most?) of the tests required to validate this feature. I started implementing in @wdio/cli, but then realized that even though retries worked, the reporters were getting notified multiple times. As per https://github.com/webdriverio/webdriverio/pull/2556#discussion_r172349719, a decision must be made on how to handle this. Possible options include:
1. Handle retries transparently and only notify reporters once a run succeeds, or during the final run. This way, reporters never receive a message that could possible be invalidated by a retry.
2. Notify reporters of all events, and add a retry event.
3. Like (1), but add a retry count to the reported event, possibly runner:end.

I'm still finding my way about the codebase, so if any committers feel inspired to just take this code and shove it into the right place, please do so. Alternatively, any guidance as to the proper location would also help a lot.

### DIFF
--- a/docs/Options.md
+++ b/docs/Options.md
@@ -140,6 +140,12 @@ If you only want to run your tests until a specific amount of tests have failed 
 Type: `Number`<br>
 Default: `0` (don't bail, run all tests)
 
+### specFileRetries
+The number of times to retry an entire specfile when it fails as a whole
+
+Type: `Number`<br>
+Default: `0`
+
 ### waitforTimeout
 Default timeout for all waitForXXX commands. Note the lowercase `f`. This timeout __only__ affects commands starting with waitForXXX and their default wait time. To increase the timeout of the test please see the framework docs.
 
--- a/packages/wdio-cli/src/interface.js
+++ b/packages/wdio-cli/src/interface.js
@@ -34,9 +34,12 @@ export default class WDIOCLInterface extends EventEmitter {
         this.clockTimer = 0
         this.jobs = new Map()
         this.start = Date.now()
+        // The relationship between totalWorkerCnt and these counters are as follows:
+        //   totalWorkerCnt - retries = finished = passed + failed
         this.result = {
             finished: 0,
             passed: 0,
+            retries: 0,
             failed: 0
         }
         this.messages = {
@@ -63,12 +66,18 @@ export default class WDIOCLInterface extends EventEmitter {
     /**
      * clear job from interface
      */
-    clearJob ({ cid, passed }) {
+    clearJob ({ cid, passed, retries }) {
         this.jobs.delete(cid)
-        this.result.finished++
+        const retry = !passed && retries > 0
+        if (!retry) {
+            this.result.finished++
+        }
 
         if (passed) {
             this.result.passed++
+        } else if (retry) {
+            this.totalWorkerCnt++
+            this.result.retries++
         } else {
             this.result.failed++
         }
@@ -117,8 +126,8 @@ export default class WDIOCLInterface extends EventEmitter {
     }
 
     updateView (wasJobCleared) {
-        const totalJobs = this.totalWorkerCnt
-        const pendingJobs = this.totalWorkerCnt - this.jobs.size - this.result.finished
+        const totalJobs = this.totalWorkerCnt - this.result.retries
+        const pendingJobs = totalJobs - this.jobs.size - this.result.finished
         const runningJobs = this.jobs.size
         const isFinished = runningJobs === 0
 
@@ -153,6 +162,7 @@ export default class WDIOCLInterface extends EventEmitter {
                 `${clockSpinnerSymbol} ` +
                 `${this.jobs.size} running, ` +
                 `${this.result.passed} passed, ` +
+                (this.result.retries ? `${this.result.retries} retries, ` : '') +
                 `${this.result.failed} failed, ` +
                 `${totalJobs} total ` +
                 `(${Math.round((this.result.finished / totalJobs) * 100)}% completed)`)
@@ -251,10 +261,11 @@ export default class WDIOCLInterface extends EventEmitter {
     }
 
     printSummary() {
-        const totalJobs = this.totalWorkerCnt
+        const totalJobs = this.totalWorkerCnt - this.result.retries
 
         return this.interface.log(
             'Test Suites:\t', chalk.green(this.result.passed, 'passed') + ', ' +
+            (this.result.retries ? chalk.yellow(this.result.retries, 'retries') + ', ' : '') +
             (this.result.failed ? chalk.red(this.result.failed, 'failed') + ', ' : '') +
             totalJobs, 'total',
             `(${totalJobs ? Math.round((this.result.finished / totalJobs) * 100) : 0}% completed)`
--- a/packages/wdio-cli/src/launcher.js
+++ b/packages/wdio-cli/src/launcher.js
@@ -333,13 +333,12 @@ class Launcher {
         const passed = exitCode === 0
 
         if (!passed && retries > 0) {
-            this.interface.totalWorkerCnt++
             this.schedule[parseInt(cid)].specs.push({ files: specs, retries: retries - 1, rid: cid })
         } else {
             this.exitCode = this.exitCode || exitCode
             this.runnerFailed += !passed ? 1 : 0
         }
-        this.interface.emit('job:end', { cid, passed })
+        this.interface.emit('job:end', { cid, passed, retries })
 
         // Update schedule now this process has ended
         if (!this.isMultiremote) {
--- a/packages/wdio-cli/src/launcher.js
+++ b/packages/wdio-cli/src/launcher.js
@@ -124,7 +124,7 @@ class Launcher {
             this.schedule.push({
                 cid: cid++,
                 caps: capabilities,
-                specs: this.configParser.getSpecs(capabilities.specs, capabilities.exclude),
+                specs: this.configParser.getSpecs(capabilities.specs, capabilities.exclude).map(s => ({ files: [s], retries: config.specFileRetries })),
                 availableInstances: capabilities.maxInstances || config.maxInstancesPerCapability,
                 runningInstances: 0,
                 seleniumServer: { hostname: config.hostname, port: config.port, protocol: config.protocol }
@@ -208,11 +208,14 @@ class Launcher {
                 break
             }
 
+            let specs = schedulableCaps[0].specs.shift()
             this.startInstance(
-                [schedulableCaps[0].specs.shift()],
+                specs.files,
                 schedulableCaps[0].caps,
                 schedulableCaps[0].cid,
-                schedulableCaps[0].seleniumServer
+                schedulableCaps[0].seleniumServer,
+                specs.rid,
+                specs.retries
             )
             schedulableCaps[0].availableInstances--
             schedulableCaps[0].runningInstances++
@@ -241,10 +244,14 @@ class Launcher {
      * Start instance in a child process.
      * @param  {Array} specs  Specs to run
      * @param  {Number} cid  Capabilities ID
+     * @param  {String} rid  Runner ID override
+     * @param  {Number} retries  Number of retries remaining
      */
-    startInstance (specs, caps, cid, server) {
+    startInstance (specs, caps, cid, server, rid, retries) {
         let config = this.configParser.getConfig()
-        cid = this.getRunnerId(cid)
+        // Retried tests receive the cid of the failing test as rid
+        // so they can run with the same cid of the failing test.
+        cid = rid || this.getRunnerId(cid)
         let processNumber = this.runnerStarted + 1
 
         // process.debugPort defaults to 5858 and is set even when process
@@ -292,7 +299,8 @@ class Launcher {
             caps,
             specs,
             server,
-            execArgv
+            execArgv,
+            retries
         })
         worker.on('message', ::this.interface.onMessage)
         worker.on('error', ::this.interface.onMessage)
@@ -318,11 +326,19 @@ class Launcher {
      * Close test runner process once all child processes have exited
      * @param  {Number} cid       Capabilities ID
      * @param  {Number} exitCode  exit code of child process
+     * @param  {Array} specs      Specs that were run
+     * @param  {Number} retries   Number or retries remaining
      */
-    endHandler ({ cid, exitCode }) {
+    endHandler ({ cid, exitCode, specs, retries }) {
         const passed = exitCode === 0
-        this.exitCode = this.exitCode || exitCode
-        this.runnerFailed += !passed ? 1 : 0
+
+        if (!passed && retries > 0) {
+            this.interface.totalWorkerCnt++
+            this.schedule[parseInt(cid)].specs.push({ files: specs, retries: retries - 1, rid: cid })
+        } else {
+            this.exitCode = this.exitCode || exitCode
+            this.runnerFailed += !passed ? 1 : 0
+        }
         this.interface.emit('job:end', { cid, passed })
 
         // Update schedule now this process has ended
--- a/packages/wdio-cli/src/templates/wdio.conf.tpl.ejs
+++ b/packages/wdio-cli/src/templates/wdio.conf.tpl.ejs
@@ -151,6 +151,9 @@ exports.config = {
     // before running any tests.
     framework: '<%= answers.framework %>',
     //
+    // The number of times to retry the entire specfile when it fails as a whole
+    // specFileRetries: 1,
+    //
     // Test reporter for stdout.
     // The only one supported by default is 'dot'
     // see also: https://webdriver.io/docs/dot-reporter.html
--- a/packages/wdio-local-runner/src/worker.js
+++ b/packages/wdio-local-runner/src/worker.js
@@ -23,9 +23,10 @@ export default class WorkerInstance extends EventEmitter {
      * @param  {object}   caps        capability object
      * @param  {string[]} specs       list of paths to test files to run in this worker
      * @param  {object}   server      configuration details about automation backend this session is using
+     * @param  {number}   retries     number of retries remaining
      * @param  {object}   execArgv    execution arguments for the test run
      */
-    constructor (config, { cid, configFile, caps, specs, server, execArgv }) {
+    constructor (config, { cid, configFile, caps, specs, server, execArgv, retries }) {
         super()
         this.cid = cid
         this.config = config
@@ -34,6 +35,7 @@ export default class WorkerInstance extends EventEmitter {
         this.specs = specs
         this.server = server || {}
         this.execArgv = execArgv
+        this.retries = retries
         this.isBusy = false
     }
 
@@ -128,7 +130,7 @@ export default class WorkerInstance extends EventEmitter {
     }
 
     _handleExit (exitCode) {
-        const { cid, childProcess } = this
+        const { cid, childProcess, specs, retries } = this
 
         /**
          * delete process of worker
@@ -137,7 +139,7 @@ export default class WorkerInstance extends EventEmitter {
         this.isBusy = false
 
         log.debug(`Runner ${cid} finished with exit code ${exitCode}`)
-        this.emit('exit', { cid, exitCode })
+        this.emit('exit', { cid, exitCode, specs, retries })
         childProcess.kill('SIGTERM')
     }
 
@@ -148,7 +150,7 @@ export default class WorkerInstance extends EventEmitter {
      * @return null
      */
     postMessage (command, argv) {
-        const { cid, configFile, caps, specs, server, isBusy } = this
+        const { cid, configFile, caps, specs, server, retries, isBusy } = this
 
         if (isBusy && command !== 'endSession') {
             return log.info(`worker with cid ${cid} already busy and can't take new commands`)
@@ -162,7 +164,7 @@ export default class WorkerInstance extends EventEmitter {
             this.childProcess = this.startProcess()
         }
 
-        this.childProcess.send({ cid, command, configFile, argv, caps, specs, server })
+        this.childProcess.send({ cid, command, configFile, argv, caps, specs, server, retries })
         this.isBusy = true
     }
 }
--- a/packages/wdio-reporter/src/index.js
+++ b/packages/wdio-reporter/src/index.js
@@ -157,6 +157,7 @@ export default class WDIOReporter extends EventEmitter {
         this.on('runner:end',  /* istanbul ignore next */ (runner) => {
             rootSuite.complete()
             this.runnerStat.failures = runner.failures
+            this.runnerStat.retries = runner.retries
             this.runnerStat.complete()
             this.onRunnerEnd(this.runnerStat)
         })
--- a/packages/wdio-reporter/src/stats/runner.js
+++ b/packages/wdio-reporter/src/stats/runner.js
@@ -15,5 +15,6 @@ export default class RunnerStats extends RunnableStats {
         this.specs = runner.specs
         this.sessionId = runner.sessionId
         this.isMultiremote = runner.isMultiremote
+        this.retry = runner.retry
     }
 }
--- a/packages/wdio-runner/src/index.js
+++ b/packages/wdio-runner/src/index.js
@@ -27,9 +27,10 @@ export default class Runner extends EventEmitter {
      * @param  {Object}    caps           capabilities to run session with
      * @param  {String}    configFile     path to config file to get config from
      * @param  {Object}    server         modified WebDriver target
+     * @param  {Number}    retries        number of retries remaining
      * @return {Promise}                  resolves in number of failures for testrun
      */
-    async run ({ cid, argv, specs, caps, configFile, server }) {
+    async run ({ cid, argv, specs, caps, configFile, server, retries }) {
         this.cid = cid
         this.specs = specs
         this.caps = caps
@@ -102,7 +103,8 @@ export default class Runner extends EventEmitter {
                     caps[browserName] = browser[browserName].capabilities
                     return caps
                 }, {})
-                : browser.options.capabilities
+                : browser.options.capabilities,
+            retry: (this.config.specFileRetries || 0) - (retries || 0)
         })
 
         /**
@@ -140,7 +142,8 @@ export default class Runner extends EventEmitter {
 
         this.reporter.emit('runner:end', {
             failures,
-            cid: this.cid
+            cid: this.cid,
+            retries
         })
 
         return this._shutdown(failures)
--- a/tests/smoke.runner.js
+++ b/tests/smoke.runner.js
@@ -43,6 +43,51 @@ import { SERVICE_LOGS, LAUNCHER_LOGS } from './helpers/fixtures'
     )
 
     /**
+     * specfile-level retries
+     */
+    let retry_failed = false
+    try {
+        await launch(
+            path.resolve(__dirname, 'helpers', 'config.js'),
+            {
+                specs: [path.resolve(__dirname, 'mocha', 'retry_and_fail.js')],
+                specFileRetries: 1
+            })
+    } catch (e) {
+        retry_failed = true
+    }
+    if (!retry_failed) {
+        throw Error('Expected retries to fail but they passed')
+    }
+
+    let retry_filename = path.join(__dirname, '.retry_succeeded')
+    let logfiles = ['wdio-0-0.log', 'wdio-0-1.log'].map(f => path.join(__dirname, f))
+    let rmfiles = [retry_filename, ...logfiles]
+    rmfiles.forEach(filename => {
+        if (fs.existsSync(filename)) {
+            fs.unlink(filename, err => {
+                if (err) {
+                    throw Error(`Unable to delete ${filename}`)
+                }
+            })
+        }
+    })
+    await launch(
+        path.resolve(__dirname, 'helpers', 'config.js'),
+        {
+            specs: [path.resolve(__dirname, 'mocha', 'retry_and_pass.js')],
+            outputDir: path.dirname(logfiles[0]),
+            specFileRetries: 1,
+            retry_filename
+        })
+    if (!fs.existsSync(logfiles[0])) {
+        throw Error(`Expected ${logfiles[0]} to exist but it does not`)
+    }
+    if (fs.existsSync(logfiles[1])) {
+        throw Error(`Expected ${logfiles[1]} to not exist but it does`)
+    }
+
+    /**
      * for some reason the process get stuck therefor exit it
      */
     process.exit(0)
--- a/docs/ConfigurationFile.md
+++ b/docs/ConfigurationFile.md
@@ -160,6 +160,9 @@ exports.config = {
     // Make sure you have the wdio adapter package for the specific framework installed before running any tests.
     framework: 'mocha',
     //
+    // The number of times to retry the entire specfile when it fails as a whole
+    specFileRetries: 1,
+    //
     // Test reporter for stdout.
     // The only one supported by default is 'dot'
     // see also: https://webdriver.io/docs/dot-reporter.html and click on "Reporters" in left column
--- a/docs/Retry.md
+++ b/docs/Retry.md
@@ -76,3 +76,15 @@ module.exports = function () {
 ```
 
 Reruns can only be defined in your step definitions file and not in your feature file.
+
+## Add retries on a per-specfile basis
+Previously only test-level and suite-level retries were available, which are fine in most cases. In any test which involves state, such as on a server or in a db, the state may be left invalid once the test fails the first time. Any subsequent retries may have no chance of passing due to the invalid state they would have to start with.
+A new browser instance is created for each specfile, which makes this an ideal place to hook and setup any other states (server, db). Retries on this level mean that the whole setup process will simply be repeated as if it were for a new specfile.
+```js
+module.exports = function () {
+    /**
+     * The number of times to retry the entire specfile when it fails as a whole
+     */
+    specFileRetries: 1
+}
+```
--- a/examples/wdio.conf.js
+++ b/examples/wdio.conf.js
@@ -130,6 +130,9 @@ exports.config = {
     // installed before running any tests.
     framework: 'mocha',
     //
+    // The number of times to retry the entire specfile when it fails as a whole
+    specFileRetries: 1,
+    //
     // Test reporter for stdout.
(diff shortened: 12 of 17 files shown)


packages/wdio-cli/src/launcher.js
```diff
@@ -269,20 +269,43 @@ class Launcher {
...
-            caps,
-            specs,
-            server,
-            execArgv
-        })
-        worker.on('message', ::this.interface.onMessage)
-        worker.on('error', ::this.interface.onMessage)
-        worker.on('exit', ::this.endHandler)
+        let retries = config.specFileRetries || 0
+        const spawn = () => {
```
christian-bromann: `spawn` and `buffer` need to be moved out as utilities otherwise it is impossible to unit tests these.

bennieswart: @christian-bromann thanks for the feedback. As mentioned in the PR description, having the code here doesn't stop the events from every run of the same spec being sent to the reporters. Someone with more knowledge of the codebase (like yourself) should make a recommendation on the best place to put this, and also what  approach (1, 2 or 3) should be taken regarding the buffering of events.

abjerstedt: Tbh... you probably should still report the failed test to the reporters either way. If a test is flakey, you I would want to explicitly know about it.

bennieswart: I agree @abjerstedt. At least in an ideal world where we could quickly patch all reporters to handle this correctly, this reporting every single detail would certainly be beneficial. A reporter can then simply discard what it doesn't need.
If we were to take this approach, where everything is sent to the reporters, it would cause existing reporters to report duplicates for retried specfiles, with all the first ones failed, except (possibly/hopefully) the last one. When viewed as a whole, however, this could cause the testrun as a whole to fail, since the reporter has been told of the failures. If only exit code is checked for success, this could be more easily worked around. We can also emit some kind of event that would link a failed run to its retry, which reporters can then be updated to take into account.

With this approach, I suspect, most of the code here can be reverted, and we can simply instantiate a brand new runner with the same details for a retry, and emit an event to link it to the failed run (or add a field to the runner:start event of the retry, linking back to the old run).

Thoughts?

christian-bromann: I like this idea. Ultimately the reporter shouldn't log anything until the test has finished which only should happen if there is no retry set or retry is set and retried that amount of time. Since we are in the early adaption face we can  do these "breaking" changes because I haven't seen a reporter that did the work to adapt for v5.

> Possible options include:

I like number 2. Again the reporter should collect the data and should reason about it when the test (worker) is done.

> I'm still finding my way about the codebase

How was it so far? I would be curious to know if the setup is understandable and the code good to read.

Thanks!

abjerstedt: wdio-cli is where the exit code is thrown. It seems we can grab the test before it hits that exit code and reset everything

bennieswart: @christian-bromann personally I like the partitioning of modules. It organizes nicely. It does seem like there is some coupling left here and there, but this can be improved over time. I can sympathize with wanting to get the partitioning done asap and leaving smaller coupling issues for later - it must have been a herculean task.
What I would definitely recommend, which I imagine would be very beneficial for newcomers, is some sort of top-level overview of how everything fits together. For example, how the cli, the runners, the adapters, and the reporters etc all fit together. I imagine a well thought-out flow chart graphic would serve well here.

As for the desired approach here, I also think 2 is more desirable. I'll see if I can implement it in the following way:
- Revert most of the code here. Leave existing event reporting as-is. Don't buffer anything.
- When a runner fails, and it has retries left, then launch a new runner with the same configuration. Any events fired by the failed runner are left untouched, as well as any events fired by the new runner. This will cause reporters to receive duplicates for that runner.
- Link the failed runner to the new runner. Linking back from the new runner with a property in the `runner:start` event, something like `retryFor: faliedRunnerInstance`. I also think adding a property to the `runner:end` event of the failed runner can be useful to indicate that it will be retried, something like `retryPending: true`. Reporters can be updated to account for retries using these properties.

Does this sound more or less correct?

christian-bromann: > Does this sound more or less correct?

Yes. We would need to make additional changes in the reporters as separate PR with base to this one so that we can merge it all together eventually.

bennieswart:
@christian-bromann @abjerstedt you can have another look. Updates made as discussed.

packages/wdio-cli/src/launcher.js
```diff
@@ -9,6 +9,14 @@ import { getLauncher, runServiceHook } from './utils'
 
 const log = logger('wdio-cli:Launcher')
 
+class Specs {
```
christian-bromann: No need to create a class with no logic. Just pass along a plain object.

packages/wdio-cli/src/launcher.js
```diff
@@ -276,15 +284,22 @@ class Launcher {
...
             server,
-            execArgv
+            execArgv,
+            retry: { for: specs.retryFor, remain: specs.retryRemain }
         })
         worker.on('message', ::this.interface.onMessage)
         worker.on('error', ::this.interface.onMessage)
-        worker.on('exit', ::this.endHandler)
+        worker.on('exit', (...args) => {
+            if (args[0].exitCode && specs.retryRemain > 0) {
```
christian-bromann: Let's move this into the `endHandler` too

packages/wdio-reporter/src/stats/runner.js
```diff
@@ -14,5 +14,6 @@ export default class RunnerStats extends RunnableStats {
         this.config = runner.config
         this.specs = runner.specs
         this.isMultiremote = runner.isMultiremote
+        this.retryFor = runner.retryFor
```
christian-bromann: Where is this property used?

bennieswart:
@christian-bromann thanks for the feedback!

Regarding the properties:
`retryRemain` is an integer, the number of retries remaining for the spec. This is used internally for the retry logic and is also given to reporters, since it is useful info for a reporter to have.
`retryFor` is a string, the cid of the runner that failed, for which this run is a retry. This is solely for use in reporters, to allow them to link retry runners to the original failing runners.
`retry` is simply an object grouping the above two properties together when passing them along internally. We can drop this property in favor of passing the above two separately.

Regarding reporters:
Any existing reporters will start receiving the `retryRemain` and `retryFor` properties. Existing reporters that do not consider these properties yet will report multiple runs for the same spec, but all except the last retry will have failed. This will probably cause existing reporters to fail. See https://github.com/webdriverio/webdriverio/pull/3381#discussion_r249895572 for a previous explanation.
The simplest solution is probably just to ensure wdio returns 0 when failed runs are retried which then succeed. The 0 exitcode will indicate success even though some runs may have failed.

packages/wdio-reporter/src/stats/runner.js
christian-bromann: I don't think we need this because `this.cid` already contains the same value

packages/wdio-local-runner/src/worker.js
```diff
@@ -25,7 +25,7 @@ export default class WorkerInstance extends EventEmitter {
      * @param  {object}   server      configuration details about automation backend this session is using
      * @param  {object}   execArgv    execution arguments for the test run
      */
-    constructor (config, { cid, configFile, caps, specs, server, execArgv }) {
+    constructor (config, { cid, configFile, caps, specs, server, execArgv, retry }) {
```
christian-bromann: Please add the parameter to the description above

packages/wdio-reporter/src/index.js
```diff
@@ -148,6 +148,7 @@ export default class WDIOReporter extends EventEmitter {
         this.on('runner:end',  /* istanbul ignore next */ (runner) => {
             rootSuite.complete()
             this.runnerStat.failures = runner.failures
+            this.runnerStat.retryRemain = runner.retryRemain
```
christian-bromann: without needing `retryFor` we can just call this `retry`

bennieswart:
@christian-bromann perhaps I was unclear about the purpose of `retryFor`. Suppose there is only one spec, which fails the first time and succeeds the second time. The reporter for the first run would have
`{ cid: '0-0', retryRemain: 1, retryFor: undefined }`
and the reporter for the second run would have
`{ cid: '0-1', retryRemain: 0, retryFor: '0-0' }`.
Within a report `cid` and `retryFor` will never contain the same value.

With multiple runners one cannot simply assume the retry was for the previous cid, therefor it must be explicitly kept.

christian-bromann:
> Within a report `cid` and `retryFor` will never contain the same value.

Why do we want to change the CID? It stands for `capability id` and if we rerun the test we run it with the same capabilities.

bennieswart:
I was under the impression that `cid` would always be unique per runner. Is this (1) wrong, (2) currently the case but a coincidence, or (3) always the case?

What do reporters use to uniquely distinguish between testruns? If they use `cid`, won't they get confused if all of a sudden there are multiple runs for the same `cid`? If not `cid`, then we should just add some other value to `retryFor` to point back to the failed run. What value would be appropriate if not the `cid`?

christian-bromann:
> I was under the impression that `cid` would always be unique per runner. Is this (1) wrong, (2) currently the case but a coincidence, or (3) always the case?

Answer: 2

Since we currently have no retry logic on runner basis every cid is unique per runner.

> What do reporters use to uniquely distinguish between testruns? If they use `cid`, won't they get confused if all of a sudden there are multiple runs for the same `cid`?

They use the CID. The reporter logic lives within the the worker process, so they can't be any confusion because the workers live independently from each other.

Let's say we have 2 capabilities and 3 spec files. The wdio testrunner would run it as:
- 0-0 Chrome + SpecA
- 0-1 Chrome + SpecB
- 1-0 Firefox + SpecA
- 1-1 Firefox + SpecB
- 2-0 Safari + SpecA
- 2-1 Safari + SpecB

It makes sense to keep the CID the same so we know which tests were retried. So for example if Firefox and SpecB is flaky the junit reporter would need to create the following files:

- wdio-0-0-junit-reporter.log (first file)
- wdio-0-0-1-junit-reporter.log (second file after first retry)

christian-bromann:
sorry, pressed shift+enter and accidentally closed the PR

bennieswart:
@christian-bromann thanks for the clarification. In that case reporters can group by `cid` to get retried runs, and sort by start time or `retry` to get the order, which makes `retryFor` unnecessary. I'll ping you again when changes have been made.

bennieswart:
@christian-bromann changes made as requested. You can have another look.

packages/wdio-cli/src/launcher.js
```diff
@@ -99,7 +99,7 @@ class Launcher {
         if (this.isMultiremote) {
             return new Promise((resolve) => {
                 this.resolve = resolve
-                this.startInstance(this.configParser.getSpecs(), caps, 0)
+                this.startInstance({ files: this.configParser.getSpecs() }, caps, 0)
```
christian-bromann: What is the advantage of passing in an object instead of an array?

tests/specfile_retries.conf.js
```diff
@@ -0,0 +1,21 @@
+const path = require('path')
```
christian-bromann: This file is not necessary imo .. just add a new testrun in `smoke.runner.js`

tests/__fixtures__/specfile_retries.spec.js
```diff
@@ -0,0 +1,33 @@
...
+    it('should execute three times due to two suite-level retries', function () {
+        console.log('Test for two retries loaded.')
+
+        throw Error('Deliberate error.')
+    })
+
+    it('should run for at least 500ms to test reported duration', function () {
+        browser.pause(500)
+    })
+})
```
christian-bromann: This is testing mocha based retries. The smoke test should create a file and fail in the first run and check if that created file exist and pass in the second. Also we should move this to `/tests/mocha/retry.js`

packages/wdio-runner/src/index.js
```diff
@@ -139,7 +140,8 @@ export default class Runner extends EventEmitter {
 
         this.reporter.emit('runner:end', {
             failures,
-            cid: this.cid
+            cid: this.cid,
+            retries: failures && retries || 0
```
christian-bromann:
```suggestion
            retries: failures ? retries : 0
```

christian-bromann: just easier to read

bennieswart: Agreed. The ugliness was an artifact from a previous edit. On second thought though, I don't think we should truncate the retries to 0 just because the test passed. It is probably more correct to always pass the correct retry count.
Which do you prefer: (1) `failures ? retries : 0`, (2) `retries || 0`, (3) `retries`.
Personally I would go for (3) as it gives the most information: if you didn't specify retries to begin with, with would be `undefined`. (2) ensures the field is always a number, replacing `undefined` with `0`.

packages/wdio-cli/src/launcher.js
bennieswart: The list of specs to run per capability is stored in `this.schedule[cid]`. Previously, there were no retries, and runner ids were generated sequentially. Now, each list of specs could possibly be accompanied by a retry count as well as a runner id override (to ensure the retry for '0-0' runs as '0-0' again). Changing `this.schedule[cid] = specs` to `this.schedule[cid] = { files: specs, cid: ..., retries: ... }` allows tracking of the additional fields nicely.

What we could do is to make `rid` and `retries` separate parameters to `startInstance` and keep `specs` as an array. Inside `startInstance`, the line `cid = this.getRunnerId(cid)` should then be changed to `cid = rid || this.getRunnerId(cid)`. Something along those lines. I do agree that is cleaner.

bennieswart:
@christian-bromann no worries, I totally agree that this should be done right and I don't mind going through a few extra iterations to achieve that. I agree with your comments and I've added some suggestions that I think are even better. I'll go ahead and implement those suggestions so that you can see how those changes would look. We can always revert if you prefer the earlier versions.

tests/__fixtures__/specfile_retries.spec.js
bennieswart: I'm not exactly sure how you want the tests restructured.

Here is a list of all the things that should be tested:
1. Test that mocha test-level and suite-level retries work as expected. This is technically not necessary since the folks at mocha will take care of this, however, it does make sense to just have a sanity check for the interaction of test-level retries and suite-level retries with specfile-level retries.
2. Test that specfile-level retries cause specs to be rerun until they succeed or run out of retries.
3. Test that retried specs that ultimately fail also cause the testrun as a whole to fail, and that retried specs that ultimately succeed do not cause the testrun as a whole to fail.
4. Test that reporters receive the correct `retries` property.
5. Test that reporters receive the correct duration of retried and succeeded runs.
6. Test that retries run with the correct runner id.
This is what I can think of now, feel free to add more.

To test this, we need to run a (mocha) test that exhibits all the possible behaviours. Call this Test A. It must be run at least twice, once to retry and fail, and once to retry and succeed. 

Then, the outcomes of the runs of Test A must be analyzed to see if the correct actions were taken and if the correct results were reported. Call this Test B.

This is the reason for having the two separate test files. Furthermore, to record the `_duration` and `retries` properties from Test A, a reporter must be attached which allows us to record those properties. This is the reason for the custom reporter. Furthermore, it seems like one cannot override the `reporters` argument for `launch` in the smoke test with a class reference as it gets lost during serialisation, which is the reason for the separate config file.

Do you agree with the concept of two separate tests? If not a custom reporter, how would one record the `_duration` and `retries` properties of runs for Test A? And if a custom reporter, it seems like the separate config file is necessary so the reporter class can be defined in the config file itself.

This is a mouthful... I hope I clarified properly. If I'm completely off, please clarify, otherwise, please suggest on the way forward.

christian-bromann: thanks for your explanation. Here are my thoughts:

- I don't think we need to test 1. at all - as you said this is handled by the mocha devs and we should trust them
- I agree we you that we should have a smoke test were the test passes after retry and where it still fails after retrying, I suggest to create two test files in `/tests/mocha`, one that has a simple test and always fails and another where it checks if a e.g. "retryMe" file exist in the same dir, if not than create that file and fail if yes than pass, then we create two launcher tests in `/tests/smoke.runner.js` where we run 2 tests using these files (re 2. and 3.)
- We should not check if reporters receive the retry or duration property, that should be captured on the unit test level (re 4. and 5.)
- I think it would be cool to check with the retry/passing smoke test if 2 wdio  logs were created (this will also indirect check if the runner id was passed in correctly)

To give some clarification of how the smoke tests are structured: the idea behind them is to run wdio testrunner tests that check specific things e2e (no unit tests / jest tests in there). It really should test things on a functional level (no details). All spec files for these testruner tests should be located in `/tests/mocha`, the testruns should be programmatically triggered in `/tests/smoke.runner.js`. The main config is located in `/tests/helpers/config.js` so that the smoke runner can always use that and overwrites specific properties based on the test scenario (e.g. spec files or reporters).

If you have further questions feel free to PM me on Gitter.

christian-bromann:
@bennieswart mind updating the branch?

tests/smoke.runner.js
```diff
@@ -42,6 +42,59 @@ import { SERVICE_LOGS, LAUNCHER_LOGS } from './helpers/fixtures'
...
+            path.resolve(__dirname, 'helpers', 'config.js'),
+            {
+                specs: [path.resolve(__dirname, 'mocha', 'retry_and_pass.js')],
+                outputDir: path.dirname(logfiles[0]),
+                specFileRetries: 1,
+                retry_filename
+            })
+        retry_passed = true
+    } catch (e) {
+    }
```
christian-bromann: No reason to wrap both of the launch calls into try/catch

packages/wdio-cli/src/launcher.js
```diff
@@ -292,11 +297,12 @@ class Launcher {
...
             specs,
             server,
-            execArgv
+            execArgv,
+            retries
         })
         worker.on('message', ::this.interface.onMessage)
         worker.on('error', ::this.interface.onMessage)
-        worker.on('exit', ::this.endHandler)
+        worker.on('exit', ({ cid, exitCode }) => this.endHandler({ cid, exitCode, specs, retries }))
```
christian-bromann: The worker has access to `specs` and `retries` and can pass them along in the exit event so that we can keep `worker.on('exit', ::this.endHandler)` here

packages/wdio-cli/src/launcher.js
```diff
@@ -241,10 +244,12 @@ class Launcher {
...
      * @param  {Array} specs  Specs to run
      * @param  {Number} cid  Capabilities ID
+     * @param  {Number} rid  Runner ID override
+     * @param  {Number} retries  Number of retries remaining
      */
-    startInstance (specs, caps, cid, server) {
+    startInstance (specs, caps, cid, server, rid, retries) {
         let config = this.configParser.getConfig()
-        cid = this.getRunnerId(cid)
+        cid = rid || this.getRunnerId(cid)
```
christian-bromann: Can we add a comment saying that in order to retry tests with the same `cid` we pass the original `cid` along as `rid` to be able to identify tests better

bennieswart:
@christian-bromann changes made. As for your question, this is the output of the smoke test:
```
"spec" Reporter:
------------------------------------------------------------------
[chrome #0-0] Spec: /home/staple/git/webdriverio/tests/mocha/retry_and_pass.js
[chrome #0-0] Running: chrome
[chrome #0-0]
[chrome #0-0] fail on first run and succeed on second
[chrome #0-0]    ✖ pass if /home/staple/git/webdriverio/tests/.retry_succeeded exists, otherwise create it and fail
[chrome #0-0]
[chrome #0-0] 1 failing (499ms)
[chrome #0-0]
[chrome #0-0] 1) fail on first run and succeed on second pass if /home/staple/git/webdriverio/tests/.retry_succeeded exists, otherwise create it and fail
[chrome #0-0] Deliberate error.
[chrome #0-0] Error: Deliberate error.
[chrome #0-0]     at Context.<anonymous> (/home/staple/git/webdriverio/tests/mocha/retry_and_pass.js:12:13)
[chrome #0-0]     at apply (/home/staple/git/webdriverio/packages/wdio-sync/src/index.js:28:28)
[chrome #0-0]     at Context.executeSync (/home/staple/git/webdriverio/packages/wdio-sync/src/index.js:26:12)
[chrome #0-0]     at call (/home/staple/git/webdriverio/packages/wdio-sync/src/index.js:161:33)
------------------------------------------------------------------
[chrome #0-0] Spec: /home/staple/git/webdriverio/tests/mocha/retry_and_pass.js
[chrome #0-0] Running: chrome
[chrome #0-0]
[chrome #0-0] fail on first run and succeed on second
[chrome #0-0]    ✓ pass if /home/staple/git/webdriverio/tests/.retry_succeeded exists, otherwise create it and fail
[chrome #0-0]
[chrome #0-0] 1 passing (363ms)


Test Suites:	 1 passed, 1 failed, 2 total (100% completed)
Time:		 🕑  2.25s

Smoke test successful
```

christian-bromann:
I wonder if we should incorporate retries in reporters as part of this PR or an upcoming one because I feel like this could be confusing to people seeing more test executed as normal. How about:

```js
Test Suites:	 1 passed, 1 retries, 1 total (100% completed)
```

bennieswart:
@christian-bromann have a look at the changes made to achieve the suggestion in your previous comment in the latest commit. I've also cleaned up the previous commits. The output of the smoke test is now:
```
"spec" Reporter:
------------------------------------------------------------------
[chrome #0-0] Spec: /home/staple/git/webdriverio/tests/mocha/retry_and_pass.js
[chrome #0-0] Running: chrome
[chrome #0-0]
[chrome #0-0] fail on first run and succeed on second
[chrome #0-0]    ✖ pass if /home/staple/git/webdriverio/tests/.retry_succeeded exists, otherwise create it and fail
[chrome #0-0]
[chrome #0-0] 1 failing (475ms)
[chrome #0-0]
[chrome #0-0] 1) fail on first run and succeed on second pass if /home/staple/git/webdriverio/tests/.retry_succeeded exists, otherwise create it and fail
[chrome #0-0] Deliberate error.
[chrome #0-0] Error: Deliberate error.
[chrome #0-0]     at Context.<anonymous> (/home/staple/git/webdriverio/tests/mocha/retry_and_pass.js:12:13)
[chrome #0-0]     at apply (/home/staple/git/webdriverio/packages/wdio-sync/src/index.js:28:28)
[chrome #0-0]     at Context.executeSync (/home/staple/git/webdriverio/packages/wdio-sync/src/index.js:26:12)
[chrome #0-0]     at call (/home/staple/git/webdriverio/packages/wdio-sync/src/index.js:161:33)
------------------------------------------------------------------
[chrome #0-0] Spec: /home/staple/git/webdriverio/tests/mocha/retry_and_pass.js
[chrome #0-0] Running: chrome
[chrome #0-0]
[chrome #0-0] fail on first run and succeed on second
[chrome #0-0]    ✓ pass if /home/staple/git/webdriverio/tests/.retry_succeeded exists, otherwise create it and fail
[chrome #0-0]
[chrome #0-0] 1 passing (352ms)


Test Suites:	 1 passed, 1 retries, 1 total (100% completed)
Time:		 🕑  2.15s

Smoke test successful
```

packages/wdio-cli/src/interface.js
```diff
@@ -153,6 +161,7 @@ export default class WDIOCLInterface extends EventEmitter {
                 `${clockSpinnerSymbol} ` +
                 `${this.jobs.size} running, ` +
                 `${this.result.passed} passed, ` +
+                `${this.result.retries} retries, ` +
```
christian-bromann: Let's only display this if there are retries.

packages/wdio-cli/src/interface.js
```diff
@@ -251,10 +260,11 @@ export default class WDIOCLInterface extends EventEmitter {
     }
 
     printSummary() {
-        const totalJobs = this.totalWorkerCnt
+        const totalJobs = this.totalWorkerCnt - this.result.retries
 
         return this.interface.log(
             'Test Suites:\t', chalk.green(this.result.passed, 'passed') + ', ' +
+            (this.result.retries ? chalk.yellow(this.result.retries, 'retries') + ', ' : '') +
```
christian-bromann: same here

bennieswart:
@christian-bromann changes made as requested. I've incorporated them into the last commit as it's pretty small. As for the other two :+1:, should we just wait for those members to react or should I actively pick arbitrary members and follow up with them?

packages/wdio-cli/src/interface.js
```diff
@@ -34,9 +34,11 @@ export default class WDIOCLInterface extends EventEmitter {
         this.clockTimer = 0
         this.jobs = new Map()
         this.start = Date.now()
+        // totalWorkerCnt - retries = finished = passed + failed
```
erwinheitzman: should this stay commented out?

packages/wdio-cli/src/interface.js
```diff
@@ -63,12 +65,18 @@ export default class WDIOCLInterface extends EventEmitter {
...
+        const retry = !passed && retries > 0
+        if (!retry) {
+            this.result.finished++
+        }
 
         if (passed) {
             this.result.passed++
+        } else if (retry) {
+            this.totalWorkerCnt++
+            this.result.retries++
```
erwinheitzman: shouldn't the amount of retries left, go down instead of increase?

packages/wdio-reporter/src/index.js
```diff
@@ -155,6 +155,7 @@ export default class WDIOReporter extends EventEmitter {
         this.on('runner:end',  /* istanbul ignore next */ (runner) => {
             rootSuite.complete()
             this.runnerStat.failures = runner.failures
+            this.runnerStat.retries = runner.retries
```
BorisOsipov: 1) Add this into doc https://github.com/webdriverio/webdriverio/tree/master/packages/wdio-reporter#runner-events
2) Is it possible to add info about retries to `runner:start` event? In fact I want to have info - is current runner doing spec retry or not. here this event emited
`https://github.com/webdriverio/webdriverio/blob/6c55c8afe2be480c18784ba42189ebc31d0585eb/packages/wdio-runner/src/index.js#L94`. 
It is usefull for some powerfull reporters like allure\reportportal.

packages/wdio-cli/src/interface.js
bennieswart: This is actually a comment - I'll clarify a bit better.

packages/wdio-cli/src/interface.js
bennieswart: This is the amount of times a retry has been run, not the amount of retries remaining. If the above comment was clearer, this probably would have been too :)

packages/wdio-reporter/src/index.js
bennieswart: 1. Will do.
2. Have a look at https://github.com/webdriverio/webdriverio/pull/3381#issuecomment-465943670 and the roughly 6 comments preceding it. When a new runner starts, if its runner id matches that of an earlier runner, you know it's a retry.

BorisOsipov: 1. Thank you!
2. It is nice that `cid` is the same. But we can't save info which cid's already executed(and processed by reporter) because on every `runner.run()` we create new instance of reporter see
https://github.com/webdriverio/webdriverio/blob/91b7863149d925dfea2fa9fb3d825e1e90743a3d/packages/wdio-runner/src/index.js#L61
So it will be nice to have such info from runner start event.

Mayby I miss something in runner\report relationships @christian-bromann ?

bennieswart:
@BorisOsipov I added the `retries` property to the reporter event documentation. Still waiting on @christian-bromann for comments on the start event.

@erwinheitzman I updated the comment to be more clear.

Also fixed the tests.

docs/Options.md
```diff
@@ -140,6 +140,12 @@ If you only want to run your tests until a specific amount of tests have failed
 Type: `Number`<br>
 Default: `0` (don't bail, run all tests)
 
+### specFileRetries
+The number of times to retry an entire specfile when it fails as a whole
+
+Type: `Number`<br>
+Default: *0*
```
WillBrock: Can you update the ```Default: *0*``` to ```Default: `0` ```

packages/wdio-cli/src/templates/wdio.conf.tpl.ejs
```diff
@@ -168,7 +168,10 @@ exports.config = {
     mochaOpts: {
         ui: 'bdd',
         timeout: 60000
-    },<% }
+    },
+    // The number of times to retry the entire specfile when it fails as a whole
+    // specFileRetries: 1,
+    //<% }
```
WillBrock: I think this line should be un-commented. ```//<% }```

bennieswart:
@WillBrock changes made and integrated in the documentation commit.

bennieswart:
Cool, travis is happy now. Do the codecov checks need to pass as well?

christian-bromann:
> Do the codecov checks need to pass as well?

Yes, can you check where the code coverage is missing?

bennieswart:
All checks happy. Just have a look in the test commit if you're happy with those unit tests.

packages/wdio-reporter/src/index.js
BorisOsipov: Summon @christian-bromann again

christian-bromann: @BorisOsipov is right, `runner.run()` creates a new worker instance and therefor creates a new process where the reporter is fresh initialized with no info about any previous rerun. Let's add the retry number to the `runner:start`

christian-bromann:
@bennieswart see https://github.com/webdriverio/webdriverio/pull/3381#discussion_r262390517

bennieswart:
@BorisOsipov @christian-bromann changes made:
````diff
diff --git a/packages/wdio-reporter/README.md b/packages/wdio-reporter/README.md
index 1c6dddca..86a13717 100644
--- a/packages/wdio-reporter/README.md
+++ b/packages/wdio-reporter/README.md
@@ -313,7 +313,8 @@ RunnerStats {
      timeouts: { implicit: 0, pageLoad: 300000, script: 30000 } },
   sanitizedCapabilities: 'firefox.59_0.darwin',
   config: [Object],
-  specs: [ '/path/to/project/test/my.test.js' ] }
+  specs: [ '/path/to/project/test/my.test.js' ] },
+  retry: 0
 ```
 
 ##### onRunnerEnd
diff --git a/packages/wdio-reporter/src/index.js b/packages/wdio-reporter/src/index.js
index b4e4f295..1f8b4cf5 100644
--- a/packages/wdio-reporter/src/index.js
+++ b/packages/wdio-reporter/src/index.js
@@ -54,6 +54,7 @@ export default class WDIOReporter extends EventEmitter {
         this.on('runner:start',  /* istanbul ignore next */ (runner) => {
             rootSuite.cid = runner.cid
             this.runnerStat = new RunnerStats(runner)
+            this.runnerStat.retry = runner.retry
             this.onRunnerStart(this.runnerStat)
         })
 
diff --git a/packages/wdio-runner/src/index.js b/packages/wdio-runner/src/index.js
index 18fe9655..9e988585 100644
--- a/packages/wdio-runner/src/index.js
+++ b/packages/wdio-runner/src/index.js
@@ -103,7 +103,8 @@ export default class Runner extends EventEmitter {
                     caps[browserName] = browser[browserName].capabilities
                     return caps
                 }, {})
-                : browser.options.capabilities
+                : browser.options.capabilities,
+            retry: (this.config.specFileRetries || 0) - (retries || 0)
         })
 
         /**
````

File: packages/wdio-reporter/src/index.js [hunk 21]
```diff
@@ -54,6 +54,7 @@ export default class WDIOReporter extends EventEmitter {
         this.on('runner:start',  /* istanbul ignore next */ (runner) => {
             rootSuite.cid = runner.cid
             this.runnerStat = new RunnerStats(runner)
+            this.runnerStat.retry = runner.retry
```
christian-bromann: Can we set the property in the `RunnerStats` class?

bennieswart:
@christian-bromann changes made:
````diff
diff --git a/packages/wdio-reporter/src/index.js b/packages/wdio-reporter/src/index.js
index 1f8b4cf5..b4e4f295 100644
--- a/packages/wdio-reporter/src/index.js
+++ b/packages/wdio-reporter/src/index.js
@@ -54,7 +54,6 @@ export default class WDIOReporter extends EventEmitter {
         this.on('runner:start',  /* istanbul ignore next */ (runner) => {
             rootSuite.cid = runner.cid
             this.runnerStat = new RunnerStats(runner)
-            this.runnerStat.retry = runner.retry
             this.onRunnerStart(this.runnerStat)
         })
 
diff --git a/packages/wdio-reporter/src/stats/runner.js b/packages/wdio-reporter/src/stats/runner.js
index ce3e0a48..9a467dae 100644
--- a/packages/wdio-reporter/src/stats/runner.js
+++ b/packages/wdio-reporter/src/stats/runner.js
@@ -15,5 +15,6 @@ export default class RunnerStats extends RunnableStats {
         this.specs = runner.specs
         this.sessionId = runner.sessionId
         this.isMultiremote = runner.isMultiremote
+        this.retry = runner.retry
     }
 }
````

christian-bromann:
Thanks @bennieswart 🙏  for pushing this out. I hope the long review phase wasn't too exhausting but this is a critical change that needs to be reviewed carefully.

asos-domhoare:
Respect to you guys. Been following the progress of this keenly. Great work...great feature. Thank you.

bennieswart:
Sweet! Thanks @christian-bromann for the guidance on this.

harpet06:
Seconded, big props @bennieswart ! 

This is a whole lot more elegant than the solution I was cooking up!

PR #3427, opened 2019-01-25
change logger imports/exports for optimized webpack bundling

jlipps:
Currently, while the logger package works when bundled within webpack, it does so at the cost of requiring many node dependencies, which we obviously want to avoid in a web context so that our bundle size is not outrageous.

This change is more or less a hack to this end.

### DIFF
--- a/packages/wdio-logger/src/index.js
+++ b/packages/wdio-logger/src/index.js
@@ -1,6 +1,21 @@
 /**
  * environment check to allow to use this package in a web context
  */
-export default (typeof process !== 'undefined' && typeof process.release !== 'undefined' && process.release.name === 'node')
-    ? require('./node').default
-    : require('./web').default
+
+// By default, import the web code using a literal require, so that in webpack
+// contexts, it will always be bundled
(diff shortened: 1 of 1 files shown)


packages/wdio-logger/src/index.js
```diff
@@ -1,6 +1,21 @@
...
+// By default, import the web code using a literal require, so that in webpack
+// contexts, it will always be bundled
+let mode = require('./web').default
+
+// Then, if we're in a Node.js context, require the node version of this module
+// using a variable, so that it will _not_ be included in a bundle, either
+// during compilation or execution
+if (typeof process !== 'undefined' && typeof process.release !== 'undefined' && process.release.name === 'node') {
+    const nodeMode = './node'
+    mode = require(nodeMode)
```
christian-bromann:
```suggestion
    mode = require(nodeMode).default

Comment (christian-bromann) [reply packages/wdio-logger/src/index.js]: yeah you should be able to do

```sh
$ lerna run --scope @wdio/logger build
```

PR #3425, opened 2019-01-25
Consider Appium as non W3C

mgrybyk:
## Proposed changes

Consider Appium as non w3c

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/webdriverio/webdriverio/blob/master/CONTRIBUTING.md) doc
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

## Further comments

fix #3424

PR #3347, opened 2019-01-15
Fix for #3344

nami-varthakavi:
## Proposed changes

[//]: # (Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue.)

## Types of changes

[//]: # (What types of changes does your code introduce to WebdriverIO?)
[//]: # (_Put an `x` in the boxes that apply_)

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

[//]: # (_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._)

- [ ] I have read the [CONTRIBUTING](https://github.com/webdriverio/webdriverio/blob/master/CONTRIBUTING.md) doc
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

## Further comments

[//]: # (If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...)

PR #3428, opened 2019-01-25
Temporary workaround for addValue and setValue

mgrybyk:
## Proposed changes

temporary make `/session/:sessionId/element/:elementId/value` to send both value and text parameters.

## Types of changes

workaround of Appium defect

## Checklist

## Further comments

I've performed following tests locally:

- firefox // isW3C true
- chrome // isW3C false
- android real device // isW3C true


mgrybyk:
@christian-bromann @jlipps @nami-varthakavi here is another workaround that is not braking things now.

nami-varthakavi:
> @christian-bromann @jlipps @nami-varthakavi here is another workaround that is not braking things now.

How can elementSendKeys take three arguments?

packages/webdriverio/src/commands/element/addValue.js
```diff
@@ -32,5 +32,5 @@ export default function addValue (value) {
         return this.elementSendKeys(this.elementId, transformToCharString(value))
     }
 
-    return this.elementSendKeys(this.elementId, value)
+    return this.elementSendKeys(this.elementId, value, transformToCharString(value))
```
nami-varthakavi: how come elementSendKeys take three arguments?

mgrybyk: @nami-varthakavi I've added one more parameter to "/session/:sessionId/element/:elementId/value" in https://github.com/webdriverio/webdriverio/pull/3428/files#diff-e4d79ef8dfe79517f19a6960718bfa10

nami-varthakavi: I would say might as well have another command with ```{
        "name": "value",
        "type": "string[]",
        "description": "The sequence of keys to type. An array must be provided.",
        "required": true
      }]	      }]```

mgrybyk: We can choose between jsonwp and webdriver command in this case.
Jsonwp scenario is unchanged, I've only changed webdriver's command to isolate workaround as much as possible.
Change doesn't affect w3c usage of value command but adds compatibility with jsonwp value command

mgrybyk: it is not possible to have two commands for same endpoint and I don't want to add such logic for temporary workaround. Also, I don't want to override W3C command with JSONWP command even for Appium, however if you insist, I can change ProtocolCommands setup and override W3C "/session/:sessionId/element/:elementId/value" with JSONWP "/session/:sessionId/element/:elementId/value".

nami-varthakavi: I understand this is a workaround but I would probably not change the basic structure of existing command. Rather go by the above approach i mentioned and make it is only applicable to JSONWP. we don't complicate things and understand clearly what is the purpose of both commands.

mgrybyk: I'm not changing command, I'm extending it.

mgrybyk: Appium uses W3C, any changes to JSONWP won't make any effect

nami-varthakavi: I might be ok with `however if you insist, I can change ProtocolCommands setup and override W3C "/session/:sessionId/element/:elementId/value" with JSONWP "/session/:sessionId/element/:elementId/value".` but I would @christian-bromann should give a final say on this.

mgrybyk: my first PR was overriding W3C with JSONWP commands for Appium.

mgrybyk: https://github.com/webdriverio/webdriverio/pull/3425

mgrybyk: > I might be ok with `however if you insist, I can change ProtocolCommands setup and override W3C "/session/:sessionId/element/:elementId/value" with JSONWP "/session/:sessionId/element/:elementId/value".` but I would say @christian-bromann should give a final say on this.

It would be really ugly solution. I had to change ProtocolCommands logic and add condition to addValue:
```
export default function addValue (value) {
    if (!this.isW3C || this.isMobile) { // added  || this.isMobile
        return this.elementSendKeys(this.elementId, transformToCharString(value))
    }

    return this.elementSendKeys(this.elementId, value, transformToCharString(value))
}
```

nami-varthakavi: I added this solution before. It was reverted back :(

nami-varthakavi: https://github.com/webdriverio/webdriverio/pull/3347

mgrybyk: @nami-varthakavi your solution causing failure for me:
```
Error: Malformed type for "text" parameter of command elementSendKeys
Expected: string
Actual: object
```

mgrybyk: which is absolutely obvious because **parameter type check** validates parameters against W3C command that accept string, and you are sending an array.

mgrybyk:
@nami-varthakavi may you provide an example or explain in a different way because I'm not sure if I understand what you mean

packages/webdriverio/src/commands/element/addValue.js
mgrybyk: I'd better say that architecture is quite solid so that we can apply hacky workarounds easily, which is good.

nami-varthakavi:
@mgrybyk Forget about my previous comment. I think I found another issue with setValue command in a multi-remote situation.

mgrybyk:
Not sure, I'd better rise another ticket for this

mgrybyk:
@christian-bromann may we apply this workaround?

nami-varthakavi:
> Not sure, I'd better rise another ticket for this and we can discuss it there because it is not directly related to current PR

Yeah I will.

mgrybyk:
@christian-bromann done.
may you please take a look at #3443 before?

maybe I'll have to add `isSauce` as well.
```
    // Workaround https://github.com/appium/appium/issues/12085
    if (this.isMobile || this.isSauce) {
        return this.elementSendKeys(this.elementId, value, transformToCharString(value))
    }
```

ericzered:
Hello, sorry to edit this old subject but it looks like I have a similar issue even I've got the last webdriverio package version.
When I run a test in Multiremote, launching Chrome Desktop and Android App, I have the error : 
```
Malformed type for "text" parameter of command elementSendKeys
Expected: string
Actual: object
```
for the setValue.
If I run only Chrome Desktop on Multiremote, it's ok. It only happen when Chrome is launch with mobile/appium browser.
So reading the commit, I've made a test and I changed this line : 
https://github.com/webdriverio/webdriverio/blob/master/packages/webdriverio/src/commands/element/addValue.js#L32
to put this line instead : 
https://github.com/webdriverio/webdriverio/blob/master/packages/webdriverio/src/commands/element/addValue.js#L37
so it return always the 'appium workaround' it's ok for Chrome Desktop.... but it's not ok now when I run only appium.
So is they're a way to have setValue always working ?
Thanks a lot

PR #3467, opened 2019-01-31
set logLevel per logger and disable driver logger

### LINKED ISSUE #3440: Be able to configure / disable particular logger
There is no way to enable/disable or configure specific logger.

Example: it is not possible to disable one of or all driver logs.

it should be possible to do something like:
```
    // Set specific log levels per logger. Use 'silent' level to disable logger
    logLevels: {
        webdriver: 'info',
        'wdio-applitools-service': 'info'
    },

    // If outputDir is provided WebdriverIO can capture driver session logs
    // it is possible to configure which logTypes to include/exclude.
    driverLog: {
        // include: ['logcat'], // only include certain types of logs
        exclude: ['bugreport', 'server']
    }
```
be able to enable / disable particular logger at least somehow

mgrybyk:
## Proposed changes

ability to set log level per logger and disable driver logger
```
  outputDir: __dirname,
  logLevel: 'info',
  
  logLevels: {
    webdriver: 'warn',
    webdriverio: 'debug',
  },

  // extremely helpful with Appium if outputDir is provided
  capabilities: [{
      deviceName: 'Pixel 3',
      excludeDriverLogs: ['bugreport', 'server'] // '*' to exclude all logs
  }]
```

## Types of changes

## Checklist

## Further comments

resolve #3440

### DIFF
--- a/docs/ConfigurationFile.md
+++ b/docs/ConfigurationFile.md
@@ -105,7 +105,11 @@ exports.config = {
         "moz:firefoxOptions": {
           // flag to activate Firefox headless mode (see https://github.com/mozilla/geckodriver/blob/master/README.md#firefox-capabilities for more details about moz:firefoxOptions)
           // args: ['-headless']
-        }
+        },
+        // If outputDir is provided WebdriverIO can capture driver session logs
+        // it is possible to configure which logTypes to exclude.
+        // excludeDriverLogs: ['*'], // pass '*' to exclude all driver session logs
+        excludeDriverLogs: ['bugreport', 'server'],
     }],
     //
     // Additional list of node arguments to use when starting child processes
@@ -118,6 +122,16 @@ exports.config = {
     //
     // Level of logging verbosity: trace | debug | info | warn | error
     logLevel: 'info',
+    //
+    // Set specific log levels per logger
+    // use 'silent' level to disable logger
+    logLevels: {
+        webdriver: 'info',
+        'wdio-applitools-service': 'info'
+    },
+    //
+    // Set directory to store all logs into
+    outputDir: __dirname,
     //
     // If you only want to run your tests until a specific amount of tests have failed use
     // bail (default is 0 - don't bail, run all tests).
--- a/packages/wdio-cli/src/templates/wdio.conf.tpl.ejs
+++ b/packages/wdio-cli/src/templates/wdio.conf.tpl.ejs
@@ -84,7 +84,11 @@ exports.config = {
         // 5 instances get started at a time.
         maxInstances: 5,
         //
-        browserName: 'firefox'
+        browserName: 'firefox',
+        // If outputDir is provided WebdriverIO can capture driver session logs
+        // it is possible to configure which logTypes to include/exclude.
+        // excludeDriverLogs: ['*'], // pass '*' to exclude all driver session logs
+        // excludeDriverLogs: ['bugreport', 'server'],
     }],
     //
     // ===================
@@ -94,6 +98,20 @@ exports.config = {
     //
     // Level of logging verbosity: trace | debug | info | warn | error
     logLevel: '<%= answers.logLevel %>',
+    //
+    // Set specific log levels per logger
+    // loggers:
+    // - webdriver, webdriverio
+    // - wdio-applitools-service, wdio-browserstack-service, wdio-devtools-service, wdio-sauce-service
+    // - wdio-mocha-framework, wdio-jasmine-framework
+    // - wdio-local-runner, wdio-lambda-runner
+    // - wdio-sumologic-reporter
+    // - wdio-cli, wdio-config, wdio-sync, wdio-utils
+    // Level of logging verbosity: trace | debug | info | warn | error | silent
+    // logLevels: {
+        // webdriver: 'info',
+        // 'wdio-applitools-service': 'info'
+    // },
     //
     // If you only want to run your tests until a specific amount of tests have failed use
     // bail (default is 0 - don't bail, run all tests).
--- a/packages/wdio-logger/src/node.js
+++ b/packages/wdio-logger/src/node.js
@@ -41,6 +41,7 @@ const SERIALIZERS = [{
 }]
 
 const loggers = {}
+let logLevelsConfig = {}
 const logCache = new Set()
 let logFile
 
@@ -86,7 +87,7 @@ prefix.apply(log, {
     template: '%t %l %n:',
     timestampFormatter: (date) => chalk.gray(date.toISOString()),
     levelFormatter: (level) => chalk[COLORS[level]](level.toUpperCase()),
-    nameFormatter: (name) => chalk.whiteBright(name || 'global')
+    nameFormatter: (name) => chalk.whiteBright(name)
 })
 
 export default function getLogger (name) {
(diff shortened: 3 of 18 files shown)


packages/wdio-runner/tests/reporter.test.js
```diff
@@ -159,7 +159,7 @@ describe('BaseReporter', () => {
             reporterSyncTimeout: 100
         })
 
-        setTimeout(() => (reporter.reporters[0].inSync = true), 110)
+        setTimeout(() => (reporter.reporters[0].inSync = true), 112)
```
mgrybyk: this test is often failing for me and even sometimes in CI.
Adding 1ms increases stability by 90%, had no failures with 2ms added

xlenz:
@christian-bromann Good morning, will you be able to take a look, please?

klamping:
Just wanted to say this is a neat idea and looks pretty well implemented.

docs/ConfigurationFile.md
```diff
@@ -119,6 +119,23 @@ exports.config = {
...
+    // If outputDir is provided WebdriverIO can capture driver session logs
+    // it is possible to configure which logTypes to include/exclude.
+    driverLog: {
+        // include: ['logcat'], // only include certain types of logs
+        exclude: ['bugreport', 'server']
+    },
+    //
+    // Set directory to store all logs into
+    outputDir: __dirname,
+    //
```
christian-bromann: Can we add this (completely commented out though) to https://github.com/webdriverio/webdriverio/blob/master/packages/wdio-cli/src/templates/wdio.conf.tpl.ejs ?

docs/ConfigurationFile.md
```diff
@@ -119,6 +119,23 @@ exports.config = {
     // Level of logging verbosity: trace | debug | info | warn | error
     logLevel: 'info',
     //
+    // Set specific log levels per logger
+    // use 'silent' level to disable logger
```
christian-bromann: I think we need more documentation on which logger are available. I am not sure if the logger name is always the package name. This can confuse people.

docs/ConfigurationFile.md
```diff
@@ -119,6 +119,23 @@ exports.config = {
...
+    logLevels: {
+        webdriver: 'info',
+        'wdio-applitools-service': 'info'
+    },
+    //
+    // If outputDir is provided WebdriverIO can capture driver session logs
+    // it is possible to configure which logTypes to include/exclude.
+    driverLog: {
```
christian-bromann: This is problematic, if you run tests with different browser and different driver supporting different log types we run into issues. This needs to live on the capability level.

mgrybyk:
> logLevels option is fine but we need to make sure that if you pass in webdriver that it also accounts for log names like webdriver:requestModule

I'm not sure here, isn't webdriver:requestModule a different logger? If yes maybe we should be able to configure all of them separately?

christian-bromann:
> I'm not sure here, isn't webdriver:requestModule a different logger?

I think in order to differentiate loggers I used a format similar to `<package-name>:<subtype>`. That needs to be check tho.

> If yes maybe we should be able to configure all of them separately?

No because how are users suppose to know the different loggers we use in our codebase. I think if we simplify this on a per package basis it is already configuration enough.

docs/ConfigurationFile.md
mgrybyk: There are following loggers

- webdriver
- webdriverio

services:
- wdio-applitools-service
- wdio-browserstack-service
- wdio-devtools-service
- wdio-sauce-service

test frameworks
- wdio-mocha-framework
- wdio-jasmine-framework

runners
- wdio-local-runner
- wdio-lambda-runner

reporters
- wdio-sumologic-reporter

@wdio utils
- wdio-cli
- wdio-config
- wdio-sync
- wdio-utils

---

do you think we should keep it somewhere in docs? Users may configure particular logger simply depending on output.

mgrybyk: some packages, like allure-reporter, reusing existing logger instances, ex wdio-runner.

mgrybyk: @christian-bromann updated everything except of this

mgrybyk: not sure if it is a proper place for loggers list
https://github.com/webdriverio/webdriverio/pull/3467/files#diff-bad5e7d995f0493ee7a0c93b3090b0b3R107

mgrybyk:
Fighting with codecov/path, unable to figure out what am I missing

docs/ConfigurationFile.md
```diff
@@ -106,6 +106,14 @@ exports.config = {
           // flag to activate Firefox headless mode (see https://github.com/mozilla/geckodriver/blob/master/README.md#firefox-capabilities for more details about moz:firefoxOptions)
           // args: ['-headless']
         }
+    }, {
+        // If outputDir is provided WebdriverIO can capture driver session logs
+        // it is possible to configure which logTypes to exclude.
+        // excludeDriverLogs: ['*'], // pass '*' to exclude all driver session logs
+        excludeDriverLogs: ['bugreport', 'server'],
```
christian-bromann: Can you just incorporate this into the 2 capabilities above?

docs/ConfigurationFile.md
christian-bromann: What about loggers like this https://github.com/webdriverio/webdriverio/blob/00bdd480d34757277d8b9f4f62f504311671a82e/packages/wdio-devtools-service/src/commands.js#L11 , I am sure we will have more specific loggers, how do we manage these?

packages/wdio-cli/src/templates/wdio.conf.tpl.ejs
```diff
@@ -85,6 +85,14 @@ exports.config = {
...
         //
         browserName: 'firefox'
+    // }, {
+        // If outputDir is provided WebdriverIO can capture driver session logs
+        // it is possible to configure which logTypes to include/exclude.
+        // excludeDriverLogs: ['*'], // pass '*' to exclude all driver session logs
+        // excludeDriverLogs: ['bugreport', 'server'],
+        // "appium:deviceName": "Pixel 2",
+        // "appium:platformName": "Android",
+        // "appium:platformVersion": "9"
```
christian-bromann: same as above .. let's just incorporate these

docs/ConfigurationFile.md
mgrybyk: I'm sorry, what do you mean to be honest.

excludeDriverLogs is only for driverlogs. 

If you want to disable particular logger just set it level to silent

packages/wdio-cli/src/templates/wdio.conf.tpl.ejs
mgrybyk: Do you want to set logLevel of wdio loggers (not driver logs) in capabilities?

Can't imagine why would we need to set wdio loggers per capability.

docs/ConfigurationFile.md
mgrybyk: if you do something like this
logLevels {
  'wdio-devtools-service': 'silent'
}
it will change log levels for 
wdio-devtools-service
and every logger that starts with wdio-devtools-service:
ex: wdio-devtools-service:CommandHandler

is it what are you asking for?

mgrybyk: https://github.com/webdriverio/webdriverio/pull/3467/files#diff-df13ae3e41b9e0b3e6427a57619a38c4R119

docs/ConfigurationFile.md
christian-bromann: I thought `excludeDriverLogs` would exclude the log types being fetched after the test by https://github.com/webdriverio/webdriverio/blob/master/packages/wdio-runner/src/index.js#L195-L241 , no?

mgrybyk: Yes (if we talking about same things).
We are getting logs supported by browser 
`        const logTypes = await global.browser.getLogTypes()`
then we are filtering them with excludeDriverLogs.

It has nothing to do with wdio loggers

mgrybyk:
oooh, finally, these tests take a lot of time :)

packages/wdio-logger/src/node.js
```diff
@@ -56,6 +57,9 @@ log.methodFactory = function (methodName, logLevel, loggerName) {
         }
 
         args = args.map((arg) => {
+            if (typeof arg === 'string') {
```
mgrybyk: remove this.

docs/ConfigurationFile.md
christian-bromann: I was just saying instead of adding a new capability to this demo config , let's just add it to one of the existing ones.

mgrybyk: I'll update both doc and template and let you know

mgrybyk:
@christian-bromann updated.

mgrybyk:
Have couple concerns regarding that:
- have to do mapping package name - logger name and increase complexity
- it is easier to see specific logger in log, copy paste its name and set level
- not all packages have its own logger, some packages reusing logger of other packages

If you still insist I can try of course but I can't see much benefits right now.
If we have logging configured so that each package has own logger then it might make sense.

christian-bromann:
> * not all packages have its own logger, some packages reusing logger of other packages

Where does this happen?

> * it is easier to see specific logger in log, copy paste its name and set level

Ah you are right. In this case this we should rename the logger to show the correct NPM package names that user are familiar with.

xlenz:
> Where does this happen?
At least with some reporters like allure reporter.

We can report this for future or just keep in mind and do later.
Rename loggers and avoid reusing loggers.

christian-bromann:
Created ticket for renaming the logger names https://github.com/webdriverio/webdriverio/issues/3616

PR #3468, opened 2019-01-31
Add automatically determination of the RDC hostname for US/EU

wswebcreation:
## Proposed changes
Previously if you wanted to use the RDC cloud of Sauce Labs you needed to provide your hostname yourself. This PR will set the hostname automatically by defaulting it to the US-hostname and if the region is provided, it is determined based on the region

## Types of changes

## Checklist


packages/wdio-config/src/lib/ConfigParser.js
```diff
@@ -52,10 +52,17 @@ export default class ConfigParser {
...
+             * If so, we need to provide a capabilities object to the config to set the hostname
+             */
+            const isRDC = this._capabilities.find(capability => 'testobject_api_key' in capability)
+            const capabilities = {capabilities: isRDC ? {testobject_api_key: 1} : {}}
+
             /**
              * detect Selenium backend
              */
-            this._config = merge(detectBackend(this._config), this._config, MERGE_OPTIONS)
+            this._config = merge(detectBackend({...this._config,...capabilities}), this._config, MERGE_OPTIONS)
```
christian-bromann: I think it will be easier to just pass in a new parameter called `isRDC`

packages/wdio-config/src/utils.js
```diff
@@ -7,19 +7,21 @@ const REGION_MAPPING = {
...
-    const dc = region ?
-        typeof REGION_MAPPING[region] !== 'undefined' ?
-            REGION_MAPPING[region] : (region + '.')
-        : ''
-    return `${dc}saucelabs.com`
+export function getSauceEndpoint (region, isRDC) {
+    const dcRegion = REGION_MAPPING[region] ? region : 'us'
+
+    if (isRDC){
+        return `${dcRegion}1.appium.testobject.com`
```
christian-bromann: this makes it difficult to allow to set arbitrary regions which is important when there will be a new datacenter like `us2`

packages/wdio-config/src/utils.js
```diff
@@ -47,10 +49,19 @@ export function detectBackend (options = {}) {
...
-    if (typeof user === 'string' && key.length === 36) {
+    const isRDC = capabilities && 'testobject_api_key' in capabilities
+    if ((typeof user === 'string' && key.length === 36) ||
+        // When SC is used a user needs to be provided and `isRDC` needs to be true
+        (typeof user === 'string' && isRDC) ||
+        // Or only RDC
+        isRDC
+    ) {
+        // For the VM cloud a prefix needs to be added, the RDC cloud doesn't have that
+        const preFix = isRDC ? '' : 'ondemand.'
```
christian-bromann: let's move the ondemand part to the `getSauceEndpoint` method then

packages/wdio-config/src/utils.js
wswebcreation: Ok, I understand what you mean, but what if the url will be like the `eu` one. I'd like to keep it simple, if a new DC would be added in the future I'd like to add it then. for now we have a stable solution that would not break if there is a typo and default to the us.
If you'd like it to break it (so not default), then I'm fine with it and revert the change

packages/wdio-config/src/utils.js
wswebcreation: The `getSauceEndpoint` is also used in the `@wdio/sauce-service` to determine the hostname for the rest api so moving it to there will break it.
Possible solution would be:

1. keep it here
2. create a new method that would only create the rest api hostname. Question for that one is where you want to have the method. In the `@wdio/config`, or in the `@wdio/sauce-service`. My reason for putting it in the `@wdio/sauce-service` would be that it already has the endpoint there (so it could belong in that scope), but I can also imagine if you would like to have it in the `@wdio/.config` for having all the configs there. If you'd choose for the last one then I'd also like advice to remove the `getSauceRestUrl` from the `@wdio/sauce-service` to the `@wdio/config`.

Like to hear your thoughts on that

packages/wdio-config/src/lib/ConfigParser.js
wswebcreation: Good suggestion, will fix that

packages/wdio-config/src/utils.js
christian-bromann: let's keep it then

packages/wdio-config/src/utils.js
```diff
@@ -47,10 +49,18 @@ export function detectBackend (options = {}) {
...
+
         return {
             protocol: protocol || 'https',
-            hostname: hostname || `ondemand.${getSauceEndpoint(region)}`,
+            hostname: hostname || `${preFix}${getSauceEndpoint(region, isRDC)}`,
```
christian-bromann:
```suggestion
            hostname: hostname || (preFix + getSauceEndpoint(region, isRDC)),
```

packages/wdio-config/src/utils.js
christian-bromann: I guess for now it is ok, but I think at some point we need to build a region handler for arbitrary cloud services.

packages/wdio-config/src/lib/ConfigParser.js
```diff
@@ -52,10 +52,16 @@ export default class ConfigParser {
 
             this._config = merge(this._config, fileConfig, MERGE_OPTIONS)
 
+            /**
+             * For Sauce Labs RDC we need to determine if the config file has a `testobject_api_key`
+             * If so, we need to provide a boolean to the `detectBackend` to set the correct hostname
+             */
+            const isRDC = this._capabilities.some(capability => 'testobject_api_key' in capability)
```
xlenz: this line brakes multiremote!

wswebcreation: Created a fix, my bad, I thought that `capabilites` was always an array

clample:
@christian-bromann - really it should be https and port `443`. The PR actually seems OK.

wswebcreation:
Ok,

Was kinda strange, because the PR works 😉 , so it's good now?

christian-bromann:
> so it's good now?

We should switch to https and and port 443

wswebcreation:
@christian-bromann and @clample 

Hmm, I might be missing something, but it already defaults to `https` and `443` if no protocol or port is provided, see https://github.com/webdriverio/webdriverio/blob/master/packages/wdio-config/src/utils.js#L53

Or shouldn't we allow the users to provide a protocol / port for RDC?

PR #3477, opened 2019-02-01
Extendable typings

ablok:
## Proposed changes

Proposal for adding types to frameworks/services etc...
Please have a look at the proposed changes here and let me know what you think. It does 2 things:
- The current typings make no difference in the `suite` and `test` description, regardless of framework. Actually, they are currently describing the mocha output. However, when running jasmine, there is different output available. With this PR all you would have to do is reference the framework package in your `tsconfig.json` and it would give you the correct typings for `suite` and `test` AND give you global typings for either mocha or jasmine.
- As an example I also added typings in this way for the `selenium-standalone-service`. To get the correct typings and allowed extra properties to the `config` object, all you need to do is install selenium-standalone-service and add a reference to it in the `tsconfig.json`.

Look at the `typescript.md` file for a description/example. This is just a proposal, but one that would work pretty nicely I think. Ideally, all wdio packages would come with typings files like this.

Please let me know what you think about this.

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/docs/TypeScript.md
+++ b/docs/TypeScript.md
@@ -86,5 +86,5 @@ Instead of having all type definitions globally available, you can also import o
 * These import the type definition for the `test` and `suite` variables that are available in
 * the beforeTest, afterTest, beforeSuite and afterSuite hooks.
 */
-import { Suite, Test } from "mocha-framework"  
+import { Suite, Test } from "@wdio/mocha-framework"  
 ```
--- a/docs/TypeScript.md
+++ b/docs/TypeScript.md
@@ -41,3 +41,53 @@ and your `tsconfig.json` needs to look like:
     ]
 }
 ```
+
+You can even use a typed configuration if you desire.
+Al you have to do is create a plain js config file that registers typescript and requires the typed config:
+
+```javascript
+require("ts-node/register")
+module.exports = require("wdio.conf.ts")
+```
+
+
+And in your typed configuration file:
+
+```typescript
+const config: WebdriverIO.Config = {
+    // Put your webdriverio configuration here
+}
+
+export { config }
+```
+
+
+Depending on the framework you use, you will need to add the typings for that framework to your `tsconfig.json` types property.
+For instance, if we decide to use the mocha framework, we need to add it like this to have all typings globally available:
+
+```json
+{
+    "compilerOptions": {
+        "baseUrl": ".",
+        "paths": {
+            "*": [ "./*" ],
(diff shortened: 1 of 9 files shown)


docs/TypeScript.md
```diff
@@ -41,3 +41,53 @@ and your `tsconfig.json` needs to look like:
     ]
 }
 ```
+
+You can even use a typed configuration if you desire.
+Al you have to do is create a plain js config file that registers typescript and requires the typed config:
```
```suggestion
All you have to do is create a plain js config file that registers typescript and requires the typed config:
```

docs/TypeScript.md
```diff
@@ -41,3 +41,53 @@ and your `tsconfig.json` needs to look like:
...
+
+```javascript
+require("ts-node/register")
+module.exports = require("wdio.conf.ts")
+```
+
+
```
```suggestion
```

docs/TypeScript.md
```diff
@@ -41,3 +41,53 @@ and your `tsconfig.json` needs to look like:
...
+```typescript
+const config: WebdriverIO.Config = {
+    // Put your webdriverio configuration here
+}
+
+export { config }
+```
+
+
```
```suggestion
```

docs/TypeScript.md
```diff
@@ -41,3 +41,53 @@ and your `tsconfig.json` needs to look like:
...
+        },
+        "types": ["node", "webdriverio", "@wdio/mocha-framework"]
+    },
+    "include": [
+        "./src/**/*.ts"
+    ]
+}
+```
+
+
```
```suggestion
```

docs/TypeScript.md
```diff
@@ -41,3 +41,53 @@ and your `tsconfig.json` needs to look like:
...
+Instead of having all type definitions globally available, you can also import only the typings that you need like this:
+
+```typescript
+/*
+* These import the type definition for the `test` and `suite` variables that are available in
+* the beforeTest, afterTest, beforeSuite and afterSuite hooks.
+*/
+import { Suite, Test } from "mocha-framework"
```
```suggestion
import { Suite, Test } from "@wdio/mocha-framework"  
```

ablok:
@christian-bromann I processed the review comments.
Do you agree with this approach? If so, I could try to add some more typings to other services in some more PR's.

christian-bromann:
> Do you agree with this approach? If so, I could try to add some more typings to other services in some more PR's.

Yes, I mean I am not a typescript expert but I think it makes sense to add sense and I don't worry about adding a small typings package to the subpackages.

Issue #3492, opened 2019-02-04
multiremote is broken in 5.4.13


xlenz:
**Environment (please complete the following information):**
 - **WebdriverIO version:** 5.4.13

**Config of WebdriverIO**
```
    capabilities: {
        myChromeBrowser: {
            capabilities: {
                browserName: 'chrome'
            }
        },
        myFirefoxBrowser: {
            capabilities: {
                browserName: 'firefox'
            }
        }
    }
```

**Describe the bug**
multiremote stopped working in 5.4.13 due to bug in isRDC

**To Reproduce**
launch multiremote session

**Expected behavior**
multiremote should work

**Log**
```
@wdio/config/build/lib/ConfigParser.js:84
      throw e;
      ^

TypeError: this._capabilities.some is not a function
    at ConfigParser.addConfigFile (/Users/mygr/sources/e2e/node_modules/@wdio/config/build/lib/ConfigParser.js:75:40)
```

**Additional context**
https://github.com/webdriverio/webdriverio/pull/3468
https://github.com/webdriverio/webdriverio/pull/3468/files#r253472523

wswebcreation:
@xlenz 

Hmm, I see, for multiremote you need to provide an object instead of an array according to the docs. I'll try to make a fix for it

christian-bromann:
we need a multiremote smoke test to prevent this from happening again

PR #3496, opened 2019-02-04
Fix/3492 fix multi remote

wswebcreation:
## Proposed changes

This PR fixes https://github.com/webdriverio/webdriverio/issues/3492. The new added line didn't look at multiremote support which should be an object instead of an array

## Types of changes

## Checklist

## Further comments

@webdriverio/technical-committee, I was wondering why for multiremote we need to haven an object, why aren't we providing an array with an object. This makes it confusing


packages/wdio-config/src/lib/ConfigParser.js
```diff
@@ -55,8 +55,10 @@ export default class ConfigParser {
             /**
              * For Sauce Labs RDC we need to determine if the config file has a `testobject_api_key`
              * If so, we need to provide a boolean to the `detectBackend` to set the correct hostname
+             *
+             * NOTE: This will not work for multi remote
              */
-            const isRDC = this._capabilities.some(capability => 'testobject_api_key' in capability)
+            const isRDC = Array.isArray(this._capabilities) && this._capabilities.some(capability => 'testobject_api_key' in capability)
```
xlenz: please add tests for this

wswebcreation: Will check it tomorrow, already evening here, unless someone else want's to add tests for this one

christian-bromann: we don't have a smoke test for multiremote (these are tests within the `/tests` directory). For the scope of this PR it would be sufficient to have a unit tests for this change.

wswebcreation: @christian-bromann 
I've added 2 tests but I couldn't find a better way then this, please check if you agree

Issue #3495, opened 2019-02-04
Cannot compile typescript properly: Cannot find name Suite, Test


xlenz:
**Environment (please complete the following information):**
 - **WebdriverIO version:** 5.4.13
- using TypeScript

**Config of WebdriverIO**
using typescript

**Describe the bug**
unable to compline typescript project properly 

**To Reproduce**
tsc

**Expected behavior**
should be possible to compile

**Log**
```
tsc
node_modules/webdriverio/webdriverio.d.ts:112:29 - error TS2304: Cannot find name 'Suite'.

112         beforeSuite?(suite: Suite): void;
                                ~~~~~

node_modules/webdriverio/webdriverio.d.ts:113:27 - error TS2304: Cannot find name 'Test'.

113         beforeTest?(test: Test): void;
                              ~~~~

node_modules/webdriverio/webdriverio.d.ts:135:28 - error TS2304: Cannot find name 'Suite'.

135         afterSuite?(suite: Suite): void;
                               ~~~~~

node_modules/webdriverio/webdriverio.d.ts:136:26 - error TS2304: Cannot find name 'Test'.

136         afterTest?(test: Test): void;
                             ~~~~


Found 4 errors.
```

**Additional Context**

happening after https://github.com/webdriverio/webdriverio/pull/3477

xlenz:
I'll do PR for this

mgrybyk:
This one is fixed by https://github.com/webdriverio/webdriverio/pull/3497

PR #3497, opened 2019-02-04
fix Suite and Test typings for #3495

mgrybyk:
## Proposed changes

fix Suite and Test typings 

## Types of changes

## Checklist

## Further comments

#3495

### DIFF
--- a/packages/wdio-mocha-framework/mocha-framework.d.ts
+++ b/packages/wdio-mocha-framework/mocha-framework.d.ts
@@ -1,7 +1,8 @@
 /// <reference types="mocha"/>
 
-declare namespace WebDriverIO {
+declare namespace WebdriverIO {
     interface Suite {
+        file: string;
         title: string;
         parent: string;
         fullTitle: string;
@@ -11,6 +12,10 @@ declare namespace WebDriverIO {
     interface Test extends Suite {
         currentTest: string;
         passed: boolean;
-        duration: any;
+        duration?: number;
     }
+}
+
+declare module "@wdio/mocha-framework" {
+    export default WebdriverIO
 }
--- a/packages/wdio-selenium-standalone-service/selenium-standalone-service.d.ts
+++ b/packages/wdio-selenium-standalone-service/selenium-standalone-service.d.ts
@@ -1,6 +1,6 @@
 import { InstallOpts, StartOpts } from "selenium-standalone";
 
-declare namespace WebdriverIO {
+declare module "webdriverio" {
     interface Config {
         seleniumLogs?: string;
         seleniumInstallArgs?: InstallOpts;
--- a/scripts/templates/webdriverio.tpl.d.ts
+++ b/scripts/templates/webdriverio.tpl.d.ts
@@ -56,7 +56,7 @@ declare namespace WebdriverIO {
             alpha?: number,
             type?: string,
             rgba?: string
-       }
+        }
     }
 
     interface Options {
@@ -79,6 +79,9 @@ declare namespace WebdriverIO {
         services?: (string | object)[],
         execArgv?: string[]
     }
+
+    interface Suite {}
+    interface Test {}
 
     interface Hooks {
 
@@ -191,7 +194,7 @@ declare namespace WebdriverIO {
 
     type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
 
-    type Config = Options & Omit<WebDriver.Options, "capabilities"> & Hooks;
+    interface Config extends Options, Omit<WebDriver.Options, "capabilities">, Hooks {}
 }
 
 declare var browser: WebDriver.Client<void> & WebdriverIO.Browser<void>;
--- a/packages/wdio-jasmine-framework/jasmine-framework.d.ts
+++ b/packages/wdio-jasmine-framework/jasmine-framework.d.ts
@@ -1,6 +1,6 @@
 /// <reference types="jasmine"/>
 
-declare namespace WebDriverIO {
+declare namespace WebdriverIO {
     interface Suite {
         title: string;
         fullName: string;
@@ -11,4 +11,8 @@ declare namespace WebDriverIO {
         parent: string;
         passed: boolean;
     }
+}
+
+declare module "@wdio/jasmine-framework" {
+    export default WebdriverIO
 }


packages/wdio-mocha-framework/mocha-framework.d.ts
```diff
@@ -1,6 +1,6 @@
 /// <reference types="mocha"/>
 
-declare namespace WebDriverIO {
+declare namespace WebdriverIO {
     interface Suite {
         title: string;
```
ablok: Can you add the `file: string` property here?

xlenz:
@ablok 
Shall I replace `type Config = Options & Omit<WebDriver.Options, "capabilities"> & Hooks;` with `interface Config extends Options, Omit<WebDriver.Options, "capabilities">, Hooks {}` from your PR?

packages/wdio-mocha-framework/mocha-framework.d.ts
xlenz: sure!

ablok:
> @ablok
> Shall I replace `type Config = Options & Omit<WebDriver.Options, "capabilities"> & Hooks;` with `interface Config extends Options, Omit<WebDriver.Options, "capabilities">, Hooks {}` from your PR?

For the `Config` object to be able to merge, they should both be interfaces afaik. Unless you know different, then please change it.

xlenz:
@ablok thanks! Updated PR.
P.S. I'm typescript expert neither, just've got super complaining IDE :)

packages/wdio-mocha-framework/mocha-framework.d.ts
```diff
@@ -1,7 +1,8 @@
 /// <reference types="mocha"/>
 
-declare namespace WebDriverIO {
+declare namespace WebdriverIO {
     interface Suite {
+        file: string;
```
xlenz: added file

scripts/templates/webdriverio.tpl.d.ts
```diff
@@ -56,7 +56,7 @@ declare namespace WebdriverIO {
             alpha?: number,
             type?: string,
             rgba?: string
-       }
+        }
```
xlenz: indentation was wrong

packages/wdio-selenium-standalone-service/selenium-standalone-service.d.ts
```diff
@@ -8,3 +8,7 @@ declare namespace WebdriverIO {
         skipSeleniumInstall?: boolean;
     }
 }
+
+declare module "@wdio/selenium-standalone-service" {
+    export default WebdriverIO
```
ablok: This export does not seem to work for me. 
My typescript says: Exports and export assignments are not permitted in module augmentations.

christian-bromann:
@mgrybyk can you address the comments from @ablok

scripts/templates/webdriverio.tpl.d.ts
```diff
@@ -80,6 +80,9 @@ declare namespace WebdriverIO {
         execArgv?: string[]
     }
 
+    interface Suite {}
```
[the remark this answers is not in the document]
mgrybyk: I'd better keep them

packages/wdio-selenium-standalone-service/selenium-standalone-service.d.ts
mgrybyk: That's fine, we using same approach everywhere, for example please see typing of allure reporter

mgrybyk: Do you have this error in wdio project or in your test project? 
If in wdio, it's ok, otherwise that's bad

xlenz: fixing...

xlenz: the problem is because selenium-standalone itself has some mess in typings

packages/wdio-selenium-standalone-service/selenium-standalone-service.d.ts
```diff
@@ -1,10 +1,13 @@
...
+export namespace WebdriverIO {
     interface Config {
         seleniumLogs?: string;
         seleniumInstallArgs?: InstallOpts;
         seleniumArgs?: StartOpts;
         skipSeleniumInstall?: boolean;
     }
 }
+// unable to export it as module "@wdio/selenium-standalone-service"
+// because selenium-standalone is not a module
```
christian-bromann: What do you mean by that?

xlenz: I mean that we have to do
`import { InstallOpts, StartOpts } from "selenium-standalone";`
instead of 
`/// <reference types="selenium-standalone"/>`

xlenz: @ablok please check if this solution works for you

ablok: @xlenz Does not seem to work for me. Don't know if it matters, but we are using typescript 3.2.2
The only way I can get this working is if in `webdriverio.d.ts` I export as follows:
```typescript
declare module "webdriverio" {
    export = WebdriverIO;
}
```
and then in `selenium-standalone-service.d.ts` I put:
```typescript
import { InstallOpts, StartOpts } from "selenium-standalone";

declare module "webdriverio" {
    interface Config {
        seleniumLogs?: string;
        seleniumInstallArgs?: InstallOpts;
        seleniumArgs?: StartOpts;
        skipSeleniumInstall?: boolean;
    }
}
```
This makes all errors go away. But don't ask me why ;)

xlenz: @ablok exporting without default export throws an error for me.
However changing module name to webdriverio in elenium-standalone-service.d.ts works fine for me.

Can you please check again?

xlenz: @ablok argh, I didn't add mocha selenium service to types, I was just importing file.

Now I see your problem,

```

  node_modules/@wdio/selenium-standalone-service/selenium-standalone-service.d.ts:13:20
    13     export default WebdriverIO
```

xlenz:
@ablok updated PR based on your feedback!

christian-bromann:
Happy to merge once you approve @ablok

ablok:
@christian-bromann @xlenz Looks good!

xlenz:
@christian-bromann @ablok shall we fix #3504 within this PR?

PR #3502, opened 2019-02-04
Fix Appium reloadSession and add ability to set capabilities for next session

### LINKED ISSUE #3494: reloadSession doesn't work in Appium
 - **WebdriverIO version:** 5.4.13
 - **Mode:** WDIO Testrunner

**Config of WebdriverIO**
Use Appium

**Describe the bug**
reloadSession doesn't work when using Appium because sessionId is not set (it is missing in response)

**To Reproduce**
reloadSession()

**Expected behavior**
reloadSession should get sessionId in same way as newSession. 
I mean this `(response.value.sessionId || response.sessionId)`

**Log**
not needed

**Additional context**
```
const response = await sessionRequest.makeRequest(this.options);
const newSessionId = response.sessionId;
// sessionId is missing in response. It should be value.sessionId in this case
value: {
  capabilities: {...}
  sessionId:"1cb617c2-31b7-4d7c-bfb8-25f7fb9f1219"
}
```

xlenz:
## Proposed changes

fix #3494 and add ability to reload session with different capabilities.

It is really useful if there is a need to switch between app(s), app/browser, adjust some capability, etc.

## Types of changes

## Checklist

### DIFF
--- a/packages/wdio-runner/tests/reporter.test.js
+++ b/packages/wdio-runner/tests/reporter.test.js
@@ -159,7 +159,7 @@ describe('BaseReporter', () => {
             reporterSyncTimeout: 100
         })
 
-        setTimeout(() => (reporter.reporters[0].inSync = true), 110)
+        setTimeout(() => (reporter.reporters[0].inSync = true), 112)
         await expect(reporter.waitForSync())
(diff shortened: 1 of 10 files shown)


xlenz:
created separate PRs to fix defect and implement new command, closing this one

Issue #3505, opened 2019-02-05
Ability to set capabilities for next session to be created


xlenz:
**Is your feature request related to a problem? Please describe.**
ability to set capabilities programmatically for next session to be created

**Describe the solution you'd like**
```
const originalCapabilities = { app: 'com.my.app' }
it('test', () => {
  $('~NATIVE_SELECTOR')
  browser.setNextSessionCapabilities({ browserName: 'safari' })
  browser.reloadSession()
  browser.url('https://my.app.com')
})
after(() => {
  browser.setNextSessionCapabilities(originalCapabilities)
})
```
**Describe alternatives you've considered**
any possible solution to change capabilities within same test

**Additional context**

christian-bromann:
Let's continue convo in #3507

PR #3507, opened 2019-02-05
Expose buildCapabilities

mgrybyk:
## Proposed changes

add setNextSessionCapabilities command

## Types of changes

## Checklist

## Further comments

#3505 

webdriverio typings and doc should be regenerated before releasing


mgrybyk:
@christian-bromann I'll provide you with one of my real life scenarios.

1. Test looks like this
Install app
Pass welcome screens, login, etc
Close app (app should be in closed state)
Switch to browser (here is a place where I change capabilities by removing app and adding browser safari)
Open site
Click on deep link 1
(App is opened)
Ensure app is opened in required state.

Repeat with other deep links

2. I'm doing full reset and restarting WDA for the first session, for next sessions I don't need to do that.

Can't find any other way rather than changing capabilities. In v4 i was doing it in init method.

christian-bromann:
@mgrybyk you should be able to just switch context to open up an app and switch back to a web context.

mgrybyk:
@christian-bromann it is not just a context. I have to open browser. It is not possible to open browser from application.
Anyway, once app is closed it is no longer possible to do anything on iOS, I have to recreate session.

christian-bromann:
Within the native context you can automate everything in the phone, e.g. opening the settings app etc. so with that you can open the browser or your native app. Oncee you are in the browser you can switch to web context and control the browser.

mgrybyk:
@christian-bromann let me try this tomorrow, have no spare iOS real devices.

mgrybyk:
How about case 2?

Change no reset and some other capabilities like noReset, useNewWda etc? 

@christian-bromann

christian-bromann:
> Change no reset and some other capabilities like noReset, useNewWda etc?

Why not setting these in the first initial session?

mgrybyk:
@christian-bromann because of Appium instability in combination with grid when there are 5-10 real android and 5-10 real iOS devices.

Quite often wda or iwdp hangs (mostly because of previous session is not terminated properly, but not always), as a result first session cannot be started properly. Have to restart one or both of them. 
On the other hand restarting them each time between tests increases duration.

The easiest way is to change these capabilities.

mgrybyk:
@christian-bromann 
I have forgotten one more case, I have also a case when I need to switch between two native apps.

Have to create session for first app, then create session for second app. As far as Appium is installing apps, I don't know how to deal without changing capabilities.

mgrybyk:
Is there a way to get some compromise?

I was thinking on using addCommand, but there is mo access to `this.options.requestedCapabilities`.

If I can get/set requestedCapabilities that are used when creating new session it would be perfect. 

Not sure which solution is better...

christian-bromann:
> Not sure which solution is better...

I would prefer something without adding a new command. Shouldn't this actually be already possible?

mgrybyk:
@christian-bromann As far as I know it is not possible.

`this` refers to the `browser` scope in addCommand and there is no access to params.requestedCapabilities in browser scope.

Please correct me if I'm wrong

mgrybyk:
@christian-bromann oops, my fault, I can access this.options.requestedCapabilities !

However I have to duplicate the way of transforming capabilities to w3c and jsonwp formats.

May we expose this function or handle this somehow?

```
    const [w3cCaps, jsonwpCaps] = capabilities && capabilities.alwaysMatch
        /**
         * in case W3C compliant capabilities are provided
         */
        ? [capabilities, capabilities.alwaysMatch]
        /**
         * otherwise assume they passed in jsonwp-style caps (flat object)
         */
        : [{ alwaysMatch: capabilities, firstMatch: [{}] }, capabilities]

    return { w3cCaps, jsonwpCaps }
```

mgrybyk:
@christian-bromann 

can you please check PR again. I'm not adding new command now, instead I just want to be able to build capabilities in a same way webdriverio does.

Here is my use case:

```

import { buildCapabilities } from "webdriver/build/utils"

browser.addCommand("setNextSessionCapabilities", function (capabilities) {
    const { w3cCaps, jsonwpCaps } = buildCapabilities(capabilities)
    this.options.requestedCapabilities.w3cCaps = w3cCaps
    this.options.requestedCapabilities.jsonwpCaps = jsonwpCaps
})
```

mgrybyk:
> Within the native context you can automate everything in the phone, e.g. opening the settings app etc. so with that you can open the browser or your native app. Oncee you are in the browser you can switch to web context and control the browser.

sorry to say it is not possible on iOS real device. Once app is closed it is no longer possible to do anything.

christian-bromann:
Sorry but I don't see really a reason to move these two things into utilities. It has no value add for the project. I would just say, copy these into your project.

mgrybyk:
Ok, already copied :)

Issue #3545, opened 2019-02-13
Ability to add custom wait* / is* commands


xlenz:
- **WebdriverIO version:** 5.4.17

**Is your feature request related to a problem? Please describe.**
I want to be able to add custom commands to element like
`const isDisplayed = $('myElem').isDiplayedWithin(5000)`

command example
```
browser.addCommand("isDisplayedWithin", function (timeout = TIMEOUTS.wait) {
   // also we have access here to this.selector, this.parent.$, this.parent.$$ which is cool!

    try {
        this.waitForDisplayed(timeout)
        return true
    } catch (err) {
        return false
    }
}, true)
```

that will wait for element visibility for some time if it is not visible and will return true or false afterwards.

It is super helpful when doing various assertions like
```
login()
expect($(header.username).toBeDisplayedWithin()).toBe(true)
```

waitForDisplayed is not suitable here.

**Describe the solution you'd like**

adjust
https://github.com/webdriverio/webdriverio/blob/6a6e5ba36d849fe279451d5217f6e74e1862967b/packages/webdriverio/src/middlewares.js#L21

so that it will be possible to add custom waitFor commands

```
.match(/(wait(Until|ForDisplayed|ForExist|ForEnabled)|waitFor(.*)|is(.*)Existing(.*)|is(.*)Displayed(.*))/)
```
So that user can add commands like:
- waitForSomething
- isFooDisplayed
- isExistingBar

**Describe alternatives you've considered**

- ability to suppress waitFor* exceptions and return value instead (true/false)
- pass params to isExisting, isDisplayed commands so that we will wait for element to be visible/existing if it is not visible/existing. And be able to invert it

**Additional context**

i'd better allow user to add custom commands

abjerstedt:
To clarify - the main request here is to be able to add a custom command that can skip the implicit wait in the middleware?

mgrybyk:
@abjerstedt  skip waitForExist in middlewere https://github.com/webdriverio/webdriverio/blob/6a6e5ba36d849fe279451d5217f6e74e1862967b/packages/webdriverio/src/middlewares.js#L31

I'm doing PR for hits

christian-bromann:
> waitForDisplayed is not suitable here.

Why?

mgrybyk:
waitForDisplayed doesn't return true/false and throws exception in case element is not found, behavior is correct, have no objections.

[...]

PR #3548, opened 2019-02-13
webdriverio: extend wait command prefixes

mgrybyk:
## Proposed changes

Avoid waiting for element in commands with prefixes: waitUntil, waitFor, waitToBe, isExisting, isDisplayed.

Useful to use with different assertion libraries like:
`assert.strictEqual($('foo').toBeDisplayed(), true)` or `assert.strictEqual($('foo').isExistingWithin(5000), true)`, etc

## Types of changes

## Checklist

## Further comments

implements #3545

### DIFF
--- a/packages/webdriverio/src/middlewares.js
+++ b/packages/webdriverio/src/middlewares.js
@@ -18,7 +18,7 @@ export const elementErrorHandler = (fn) => (commandName, commandFn) => {
              *  - elementId couldn't be fetched in the first place
              *  - command is not explicit wait command for existance or displayedness
              */
-            if (!this.elementId && !commandName.match(/(wait(Until|ForDisplayed|ForExist|ForEnabled)|isExisting|isDisplayed)/)) {
+            if (!this.elementId && !commandName.match(/(waitUntil|waitFor|isExisting|isDisplayed)/)) {
                 log.debug(
                     `command ${commandName} was called on an element ("${this.selector}") ` +
                     'that wasn\'t found, waiting for it...'
--- a/docs/CustomCommands.md
+++ b/docs/CustomCommands.md
@@ -16,7 +16,7 @@ browser.addCommand("getUrlAndTitle", function (customVar) {
 });
 ```
 
-Additionally, you can extend the element instance with your own set of commands, by passing 'true' as the final argument.
+Additionally, you can extend the element instance with your own set of commands, by passing 'true' as the final argument. By default element is expected to be existing in `waitforTimeout` milliseconds otherwise exception will be thrown.
 
 ```js
 browser.addCommand("waitAndClick", function () {
@@ -37,6 +37,19 @@ it('should use my custom command', () => {
     assert.strictEqual(result.title, 'GitHub · Where software is built');
     assert.strictEqual(result.customVar, 'foobar');
 });
+```
+
+If there is a need to control element existance in custom commands it is possible either to add command to browser and pass selector or add command to element with name that starts with one of: waitUntil, waitFor, isExisting, isDisplayed.
+
+```js
+browser.addCommand("isDisplayedWithin", function (timeout) {
+    try {
+        this.waitForDisplayed(timeout)
+        return true
+    } catch (err) {
+        return false
+    }
+}, true);
 ```
 
 __Note:__ if you register a custom command to the browser scope the command won't be accessible for elements. Likewise, if you register a command to the element scope, it won't be accessible at the browser scope:
--- a/packages/webdriverio/tests/middleware.test.js
+++ b/packages/webdriverio/tests/middleware.test.js
@@ -2,7 +2,19 @@ import logger from '@wdio/logger'
 import { remote } from '../src'
 import request from 'request'
 
+jest.mock('../src/commands/element/waitUntil', () => ({
+    __esModule: true,
+    default: jest.fn().mockImplementation(() => { return true })
+}))
+jest.mock('../src/commands/element/waitForDisplayed', () => ({
+    __esModule: true,
+    default: jest.fn().mockImplementation(() => { return true })
+}))
 jest.mock('../src/commands/element/waitForExist', () => ({
+    __esModule: true,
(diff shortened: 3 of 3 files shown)


packages/webdriverio/src/middlewares.js
```diff
@@ -18,7 +18,7 @@ export const elementErrorHandler = (fn) => (commandName, commandFn) => {
              *  - elementId couldn't be fetched in the first place
              *  - command is not explicit wait command for existance or displayedness
              */
-            if (!this.elementId && !commandName.match(/(wait(Until|ForDisplayed|ForExist|ForEnabled)|isExisting|isDisplayed)/)) {
```
mgrybyk: Previously commands like 'waitUntilFoo', 'isExistingFoo', 'isDisplayedFoo' were skipping waitForExist call. 
So I decided to add two more shortcuts (waitFor, waitToBe) and document all this.

abjerstedt:
I'm gonna let this one wait for @christian-bromann . But it makes sense to me.

mgrybyk:
@christian-bromann I want to add more custom commands like waitForExist, waitForDisplayed, etc to element.

Currently we are filtering commands by names, if name starts with waitUntil, waitForExist, waitForDisplayed and so on we are NOT waiting for element existence. In all the other cases first we wait for element to exist and only after we can manipulate with it
https://github.com/webdriverio/webdriverio/blob/6a6e5ba36d849fe279451d5217f6e74e1862967b/packages/webdriverio/src/middlewares.js#L31

I want my custom command to ignore waitForExist. The easiest way to do it is to adjust that regex https://github.com/webdriverio/webdriverio/blob/6a6e5ba36d849fe279451d5217f6e74e1862967b/packages/webdriverio/src/middlewares.js#L21

Ideally, we should not relay on command names (i think so) but it is good to start with regex

christian-bromann:
@mgrybyk ok so I assume you want to wait on the element to exist and then do the conditional wait like waitToBeDisplayed. Why not just use waitUntil for this:

```js
elem.addCommand('isDisplayedFoo', function () => {
  this.waitForExist()
  this.waitForDisplayed()
  return this.foo()
})
```

I am not sure if we need change some root behavior here.

mgrybyk:
@mgrybyk my bad, example is not really good. 
First of all, even though it is working, we don't have it described anywhere.

Secondly try naming commands like:
waitForNotExist
doUntilSomething


In such case your example won't work if element is not existing, because waitForExist will be called first.

(Ideally it should be parametrised not just by name, but at least we can define some naming convention and document it, should work fine as well)

christian-bromann:
A `waitForNotExist` is easily done by
```js
elem.addCommand('waitForNotExist', function () => {
  browser.waitUntil(() => !this.isExisting)
})
```

> First of all, even though it is working, we don't have it described anywhere.

I would rather prefer to document these snippets than adding a bunch of commands that do similar things in a slightly different fashion. For example there are people that don't mind to have `elem.waitForExist(undefined, true)` to wait until an element disappears. WebdriverIO is easy to extend to build such small helpers really easily.

abjerstedt:
this.waitForExist(undefined, true) is a better implementation imo. It inherits from the base waitForExist.

mgrybyk:
It would be better to add waitForNotExist to element instance rather that to browser.
Let's at least allow user to add commands like waitFor*.

christian-bromann:
@mgrybyk if you add `true` as last parameter it will be added to the element scope, so
```js
browser.addCommand('waitForNotExist', function() {
    return this.waitForExist(undefined, true)
}, true)
```
should do the trick!

mgrybyk:
@christian-bromann yeah, I'm aware about this parameter.

but you code will not work!

mgrybyk:
the problem is that 
```
browser.addCommand('waitForNotExist', function() {
    return this.waitForExist(undefined, true)
}, true)
```

is calling  waitForExist first and will be failing. 
https://github.com/webdriverio/webdriverio/blob/6a6e5ba36d849fe279451d5217f6e74e1862967b/packages/webdriverio/src/middlewares.js#L31

christian-bromann:
mhm .. the problem I have with that is that we have no `waitToBe` in our code and we can't allow people to add their command name patterns to the middleware. It is not maintainable.

In this case you will need to use
```js
browser.addCommand('waitForNotExist', function () => {
  browser.waitUntil(() => !this.isExisting)
}, true)
```
Which I agree we should recommend to user to use if they don't feel comfortable with the current command wording.

mgrybyk:
@christian-bromann so let me remove waitToBe from 
`/(waitUntil|waitFor|waitToBe|isExisting|isDisplayed)/` so that we have `/(waitUntil|waitFor|isExisting|isDisplayed)/`

Is this fine?

christian-bromann:
> I wish we can pass parameter to control this

```js
browser.addCommand('myCustomCommandForNotExist', function () => {
  browser.waitUntil(() => !this.isExisting)
}, true)
```
should already do this without passing along the parameter.

mgrybyk:
@christian-bromann 
but this won't work
```
browser.addCommand('myCustomCommandForNotExist', function () => {
  browser.waitUntil(() => !this.isExisting)
}, true)
```
because of waitForExist https://github.com/webdriverio/webdriverio/blob/6a6e5ba36d849fe279451d5217f6e74e1862967b/packages/webdriverio/src/middlewares.js#L31

Am I missing something? Looks like we are not on the same page

abjerstedt:
@mgrybyk @christian-bromann  Hmmmm i think i see his point now.

waitForNotExist() isn't in the matcher, so element.waitForNotExist() will wait for the element to exist and throw up before even beginning to run the custom function.

christian-bromann:
Ah gotcha, let's change the Regex to 

```
if (!this.elementId && !commandName.match(/(waitUntil|waitFor|isExisting|isDisplayed)/)) {
```
as you suggested. Something that allows generalization but doesn't require any user to add his command name to it.

mgrybyk:
@christian-bromann updated.

PR #3538, opened 2019-02-11
webdriverio: shadow$(selector) command

jrobinson01:
## Proposed changes
Adds the `shadow$(selector)` and `shadow$$(selector)` command to elements, so that an element's shadowRoot can be queried just like any other element. These commands simply call `$` and `$$` with function selectors, so should act exactly the same as the underlying commands. ~If the element does not have a shadowRoot, the element's light dom is queried, which _should_ work for browsers that don't yet support shadow DOM or are for whatever reason, using the shady DOM polyfills.~

command usage: 
```javascript
const myEl = $('some-element').shadow$('some-internal-element');
```

[//]: # (Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue.)

## Types of changes

[//]: # (What types of changes does your code introduce to WebdriverIO?)
[//]: # (_Put an `x` in the boxes that apply_)

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

[//]: # (_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._)

- [x] I have read the [CONTRIBUTING](https://github.com/webdriverio/webdriverio/blob/master/CONTRIBUTING.md) doc
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)

## Further comments
I added a test for the new command but I cannot get it to pass. I'm not 100% sure what I need to do with the mocked request to make it pass. Any guidance is appreciated. I'm sure I'm doing something stupid but not sure what the right thing is. I've verified that the command works in a separate project. It works in Chrome, Safari and Firefox on OSX. 

It should fallback to querying the light dom via `$` or `$$` if the element does not have a shadowRoot property. I will add this ASAP.

[//]: # (If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...)

### Reviewers: @webdriverio/technical-committee

PR #3563, opened 2019-02-15
new shadow dom service

jrobinson01:
## Proposed changes
adds a new package containing a service for working with shadow DOM elements. Supersedes #3538 

The service adds the `shadow$(selector)` and `shadow$$(selector)` element commands, so that an element's shadowRoot can be queried just like any other element. These commands simply call `$` and `$$` with function selectors, so should act exactly the same as the underlying commands. If the element does not have a shadowRoot, the element's light dom is queried, which _should_ work for browsers that don't yet support shadow DOM or are for whatever reason, using the shady DOM polyfills.

command usage: 
```javascript
const myEl = $('some-element').shadow$('some-internal-element');
```

## Types of changes

## Checklist

## Further comments


packages/wdio-shadow-dom-service/src/index.js
```diff
@@ -0,0 +1,29 @@
+export default class ShadowDomService {
```
abjerstedt: I'm not an export with the shadow Dom, but shouldn't we also be able to pierce the shadow dom directly from the browser object?

packages/wdio-shadow-dom-service/src/index.js
```diff
@@ -0,0 +1,29 @@
...
+          if (this.shadowRoot) {
+            return this.shadowRoot.querySelector${qsAll ? 'All' : ''}('${elementSelector}')
+          }
+          // fall back to querying the element directly if not
+          return this.querySelector${qsAll ? 'All' : ''}('${elementSelector}')
+        })`
+        return eval(strFn)
+    }
+
+    static async shadow$ (selector) {
```
abjerstedt: As its built, the user can do browser.$('foo').shadow$('bar').shadow$('bar2'). Have you tested this scenario to see if it works?

packages/wdio-shadow-dom-service/tests/service.test.js
```diff
@@ -0,0 +1,89 @@
+import ShadowDomService from '../src'
+
+describe('ShadowDomService', () => {
+
+    let service
+    beforeAll(() => {
```
abjerstedt: We will want to break this up into 2 different unit tests (probably browser and service).

packages/wdio-shadow-dom-service/tests/service.test.js
```diff
@@ -0,0 +1,89 @@
...
+            client = {$: jest.fn()}
+        })
+        it('should call client.$ with a function', () => {
+            ShadowDomService.shadow$.call(client, '#foo')
+            expect(client.$.mock.calls).toHaveLength(1)
+            expect(client.$.mock.calls[0]).toEqual([expect.any(Function)])
+        })
+    })
+
+    describe('shadow$$', () => {
```
abjerstedt: Take a look at the $ and $$() unit tests. We have a whole mocked out request system you can use. We need to ensure that the correct endpoint is actually hit.

Based on the selector you pass in, we can determine which mocked out response you get.

packages/wdio-shadow-dom-service/src/index.js
jrobinson01: I'm not quite following. Do you mean that the browser object should have the same commands? document or body shouldn't have a shadowRoot property. There definitely could be a command that starts at either (I don't know where `$` starts it's queries) but it would just be an alias for `$`.

packages/wdio-shadow-dom-service/src/index.js
jrobinson01: It should work the same as chaining `$` calls.

packages/wdio-shadow-dom-service/tests/service.test.js
jrobinson01: The only thing the package provides is a service. It could be broken into multiple tests I guess... one for the service bits that add the commands, and another for the commands themselves.

packages/wdio-shadow-dom-service/tests/service.test.js
jrobinson01: Is that necessary? The correct endpoints would be whatever $ and $$ use. These commands just call $ or $$ with a function selector.

jrobinson01:
@christian-bromann I originally created them as built in commands but opted for a separate package instead. I could go either way.

Given your example of rolling it into the $ command: 
```javascript
const myEl = $('some-element', 'some-internal-element');
```
would the outer element always be required? That's something I'm trying to avoid (querying for a deeply nested element requires starting at a top-most element). IOW, would this also work?

```javascript
const myEl = $('some-element');
const myInnerElement = myEl('some-internal-element');
```
I'm picturing `myEl` being inside a getter on a page-object.

christian-bromann:
Those are all good concerns and after some thinking I guess your initial approach is actually the best way moving forward. The `shadow$` and `shadow$$` need to be accessible __only__ from the [element scope](https://github.com/webdriverio/webdriverio/tree/master/packages/webdriverio/src/commands/element). With that you should be able to chain `$`, `$$`, `shadow$` and `shadow$$` as much as you want.

> I could go either way.

Let's make this an native command. I don't see a reason a service would be more beneficial.

Thanks for pushing this forward!

jrobinson01:
@christian-bromann new PR is here: https://github.com/webdriverio/webdriverio/pull/3586

ToygarM:
Hello @jrobinson01! According to your readme when i execute **npm install @wdio/shadow-dom-service --save-dev** on my cli i will get an error **Not Found: @wdio/shadow-dom-service@latest.** 

How to deal with this?

christian-bromann:
@ToygarM we never published this services .. please see https://webdriver.io/blog/2019/02/22/shadow-dom-support.html

PR #3577, opened 2019-02-19
typescript promise commands

### LINKED ISSUE #3574: Type definitions for webdriverio package missed a Promise<> as a return type for a lot of functions
**Pre-check**
- [x] I'm aware that I can [edit the docs](https://github.com/webdriverio/webdriverio/blob/master/CONTRIBUTING.md) and submit a pull request

**Describe the improvement**
`webdriverio.d.ts` and `webdriver.d.ts` don't have `Promise<>` at all as a return type. But I have two points:
1) All commands return `Promise<someValue>` as we [can see here](https://github.com/webdriverio/webdriverio/blob/master/packages/webdriver/src/command.js#L80). So, for example `.click()` by `webdriverio.d.ts` is `click(): undefined;`. But if you don't use `wdio/sync` library you should definitely resolve promise (for example `await someElement.click()`. And for sure return type should be `Promise<void>` instead of `undefined`. And it is true for each function that use `webdriver` command inside implementation 
2) Some function as `browser.pause()` returns `Promise` in the implementation:
```
export default function pause (milliseconds = 1000) {
    return new Promise((resolve) => setTimeout(resolve, milliseconds))
}
```
and return type here should be Promise<void> instead of `undefined;` as I see now in `webdriverio.d.ts`

I'd like to report
- [ ] Unclear documentation
- [x] A typo
- [ ] Missing documentation
- [ ] Other

**Description of the improvement / report**
I see that definitions for all commands are generated inside `webdriverio-generate-typings.js` and `webdriverio-generate-typings.js` files. So, it would be nice if someone figure out how to fix it. If I will have a clue how to do it I will write a comment that I can create a PR. By now it could grab someone who know how to deal with it.

mgrybyk:
## Proposed changes

webdriverio types are async by default.
webdriverio sync types are in @wdio/sync package

## Types of changes

## Checklist

## Further comments

- In order to make this PR work webdriverio, webdriver and @wdio/sync packages have to be released
- Please note that it is a braking change for existing TypeScript users that work in sync mode

@CrispusDH


CrispusDH:
I'd like first option. Two typing packages would help to avoid mess and it is easier to maintain i guess.

mgrybyk:
@CrispusDH 
1. Ok, I'll change to webdriverio for version with promises and webdriverio/sync to version with @wdio/sync.
Does this work?
2. Yes. 
elPromise = $('s') // is Promise (then, catch, ..)
el = await elPromise // result element
Same with methods
el.getAttribute('a') // is Promise
await el.getAttribute('a') // string

BorisOsipov:
1) No, please. Keep `webdriverio` typings as is and create a new for async. If we deliver new typings version with promises in the same package it will confuse a lot of users.

mgrybyk:
I'd better have async or promise version as default one (webdriverio). And sync version as webdriverio/sync because it requires installation of additional package

CrispusDH:
If we are discussing right way then I agree with @mgrybyk that we should keep such structure: pure webdriverio has async types -> webdriverio + sync has sync types, it seems logical for me. Also, for sync users they will need to change just type package name and add one more package and thats all, I think this is a sufficient effort for implementing types in a right way.

BorisOsipov:
>it seems logical for me.

Yeah I got your points and your arguments are clear.
Maybe some comments from @webdriverio/technical-committee?

Also @mgrybyk I see as additional step of this PR( or maybe for separate PR) - install webdriverio-sync typings package when user select sync mode on `wdio config` install wizard.

christian-bromann:
> @christian-bromann could you bring your opinion about structure?

I agree @mgrybyk and @CrispusDH opinion. The default behavior is async and only if you install `@wdio/sync` you get the sync version of commands.

mgrybyk:
@christian-bromann @wdio/sync package may have its own typings one day. Don't want to mix @wdio/sync with webdriverio typings.
Any other options? Keeping typing inside webdriverio is somewhat easier to handle.

not sure if it make sense to add package @wdio/webdriverio-sync

mgrybyk:
Please note that I'm going to do the same for webdriver package!

christian-bromann:
> not sure if it make sense to add package @wdio/webdriverio-sync

I would like to avoid that as much as possible. I am not sure how the user would benefit from `@wdio/sync` typings so why not putting both there?

mgrybyk:
@christian-bromann at least because we need to have typings for both webdriver and webdriverio.

It is not possible to export multiple typings and it is not possible to have packages like:
@wdio/sync/webdriver
@wdio/sync/webdriverio

mgrybyk:
@christian-bromann my bad, there is no need to introduce sync/async typing for webdriver explicitly.

christian-bromann:
ok, so I would say, let's make the standard webdriverio typings asynchronous and put the synchronous version into `@wdio/sync`

mgrybyk:
@christian-bromann updated.

NOTE: Have to copy webdriverio.d.ts to webdriverio and @wdio/sync because we can't refer to webdriverio as a module as far as we have to import either sync or async version and should not extend each other.

packages/wdio-sync/package.json
```diff
@@ -17,6 +17,7 @@
     "test:eslint": "eslint src test",
     "test:unit": "jest"
   },
+  "types": "./webdriverio-sync.d.ts",
```
christian-bromann: According to webdriverio-generate-typings.js this file needs to be just called `webdriverio.d.ts`

mgrybyk: we now have 
webdriverio.d.ts in both @wdio/sync and webdriverio
webdriverio-sync.d.ts in @wdio/sync
webdriverio-async.d.ts in webdriverio

christian-bromann: ah ok, makes sense. Could you update the docs to reflect this?

mgrybyk: yes, I have some tasks to do besides of docs:
* addCommand should not be a promise
* webDriver commands should return promise
* some command should be handled separately like WebDriver.newSession, remote, multiremote, etc.
* update docs

mgrybyk: @christian-bromann TypeScript docs updaed

mgrybyk:
PR updated, @BorisOsipov @CrispusDH @christian-bromann 

- In order to make this PR work webdriverio, webdriver and @wdio/sync packages have to be released
- Please note that it is a braking change for existing TypeScript users that work in sync mode

docs/TypeScript.md
```diff
@@ -42,6 +42,17 @@ and your `tsconfig.json` needs to look like:
...
+
+```json
+{
+    "compilerOptions": {
+        "types": ["node", "@wdio/sync"]
+    }
+}
+```
+
+
```
```suggestion
```

packages/wdio-sync/package.json
christian-bromann: Can we just name this `webdriverio.d.ts`? By being within the `@wdio/sync` package it should be hint enough that these are the sync types.

packages/webdriverio/package.json
```diff
@@ -44,7 +44,7 @@
     "cucumber",
     "testingbot"
   ],
-  "types": "./webdriverio.d.ts",
+  "types": "./webdriverio-async.d.ts",
```
christian-bromann: same, I would prefer to just have it `./webdriverio.d.ts`

packages/webdriverio/webdriverio-async.d.ts
```diff
@@ -0,0 +1,48 @@
+/// <reference types="./webdriverio"/>
+
+type BrowserObject = WebDriver.ClientOptions & WebDriver.ClientAsync & WebdriverIOAsync.Browser;
+
+
```
christian-bromann:
```suggestion
```

packages/webdriverio/webdriverio-async.d.ts
```diff
@@ -0,0 +1,48 @@
...
+type ElementPromise = Omit<WebdriverIO.Element, 'addCommand'>;
+
+// Element commands wrapper with Promise
+type ElementAsync = {
+    [K in keyof ElementPromise]: WrapWithPromise<WebdriverIO.Element[K]>
+}
+// Element commands that should not be wrapper with promise
+type ElementStatic = Pick<WebdriverIO.Element, 'addCommand'>
+
+
```
christian-bromann:
```suggestion
```

packages/wdio-sync/package.json
mgrybyk: There are files named webdriverio.d.ts in both webdriverio and @wdio/sync.

Should I rename them to webdriverio-core.d.ts? Will that work?

christian-bromann: > There are files named webdriverio.d.ts in both webdriverio and @wdio/sync.

Does this matter? Typescript is only looking what "types" is referencing too, no? So they can both be the same.

mgrybyk: Not really, but we have to keep two files physically.
I'll rename webdriverio to webdriverio-core, and webdriverio-sync and async to webdriverio.

mgrybyk: @christian-bromann updated

scripts/type-generation/webdriverio-generate-typings.js
```diff
@@ -66,8 +66,14 @@ const templateContents = fs.readFileSync(templatePath, 'utf8')
 let typingsContents = templateContents.replace('// ... element commands ...', allElementCommands)
 typingsContents = typingsContents.replace('// ... browser commands ...', allBrowserCommands)
 
-const outputFile = path.join(__dirname, '..', '..', 'packages/webdriverio', 'webdriverio.d.ts')
-fs.writeFileSync(outputFile, typingsContents, { encoding: 'utf-8' })
+const outputFileWebdriverio = path.join(__dirname, '..', '..', 'packages/webdriverio', 'webdriverio-core.d.ts')
```
mgrybyk: @christian-bromann renamed file to webdriverio-core

mgrybyk:
I haven't updated that. Should I change all function undefined return types to void? Looks good for me.
@CrispusDH do you have anything else?

I thought also about adding readonly flag to Element methods and forbid overriding them in addCommand. It is not supported / allowed anyway

CrispusDH:
@mgrybyk I'm not sure about all functions, but I think you will definitely figure out it :). Good idea about readonly.

mgrybyk:
@christian-bromann please don't forget to generate typings before releasing all 3 packages.
We have to mention that it is braking change for TypeScript sync users.

Thanks

christian-bromann:
Released as v5.6.0 and announced breaking change here: https://twitter.com/webdriverio/status/1098534277459918848

nami-varthakavi:
> @christian-bromann please don't forget to generate typings before releasing all 3 packages.
> We have to mention that it is braking change for TypeScript sync users.
> 
> Thanks

Hey, Will, you be able to add some documentation/examples on how to use these sync/async typings for $,$$ commands?

CrispusDH:
Here you could find how to add types for sync/async - https://webdriver.io/docs/typescript.html .

nami-varthakavi:
> Here you could find how to add types for sync/async - https://webdriver.io/docs/typescript.html .

Hey, Thanks. But I'm looking for a specific example on the usages for these commands(something like boilerplate). I'm might asking this too early because there is another fix coming in here #3618 which might answer my question.

mgrybyk:
@nami-varthakavi https://webdriver.io/docs/typescript.html

nami-varthakavi:
> @nami-varthakavi https://webdriver.io/docs/typescript.html

I'm using the version 5.6.0. I have added the "wdio/sync" to my tsconfig. I get this following error for my command were i use setValue() & click() . 

```Property 'setValue' does not exist on type 'Promise<Element>'. Did you forget to use 'await'?```

**Comment by nami-varthakavi** [comment]
> @nami-varthakavi https://webdriver.io/docs/typescript.html

Again I think you already have some updates coming up in #3618 so probably if i wait i should be fine :)

**Comment by mgrybyk** [comment]
@nami-varthakavi as it is mentioned in doc, if you are using async version (without @wdio/sync) you don't need to change anything, just keep using as it was before. Of course once @3618 is fixed.

```
async () => {
    const input = await $('input)
    input.setValue('qwe')
}
```

If you are using @wdio/sync you would need to replace webdriverio typings with @wdio/sync.

```
() => {
    const input = $('input)
    input.setValue('qwe')
}
```

nami-varthakavi:
> @nami-varthakavi as it is mentioned in doc, if you are using async version (without @wdio/sync) you don't need to change anything, just keep using as it was before. Of course once @3618 is fixed.
> 
> ```
> async () => {
>     const input = await $('input)
>     input.setValue('qwe')
> }
> ```
> If you are using @wdio/sync you would need to replace webdriverio typings with @wdio/sync.
> 
> ```
> () => {
>     const input = $('input)
>     input.setValue('qwe')
> }
> ```

Thanks for the update. Yeah, I kinda got from the discussions, initially, when I commented in this I was yet to discover that PR. anyway, thanks.

PR #3589, opened 2019-02-21
wdio-reporter: Output error messages on failure

WillBrock:
## Proposed changes

When using the dot reporter and an error occurs there is no information about the error.

Adds error messages to the dot reporter.

Adds a function to wdio-reporter that will return the error messages that other reporters can use.

## Types of changes

## Checklist

### DIFF
--- a/packages/wdio-reporter/src/index.js
+++ b/packages/wdio-reporter/src/index.js
@@ -238,7 +238,14 @@ export default class WDIOReporter extends EventEmitter {
      * function to write to reporters output stream
      */
     write (content) {
-        this.outputStream.write(content)
+        this.outputStream.write(content, 'standard')
+    }
+
+    /**
+     * Write data at the end of the report
+     */
+    writeEnd (content) {
+        this.outputStream.write(content, 'end')
     }
 
     /* istanbul ignore next */
--- a/packages/wdio-reporter/src/index.js
+++ b/packages/wdio-reporter/src/index.js
@@ -1,5 +1,6 @@
 import fs from 'fs'
 import fse from 'fs-extra'
+import chalk from 'chalk'
 import { format } from 'util'
 import EventEmitter from 'events'
 
@@ -19,7 +20,7 @@ export default class WDIOReporter extends EventEmitter {
             ? options.writeStream
             : fs.createWriteStream(this.options.logFile)
         this.failures = []
-        this.suites = {}
+        this.all_suites = {}
         this.hooks = {}
         this.tests = {}
         this.currentSuites = []
@@ -59,7 +60,7 @@ export default class WDIOReporter extends EventEmitter {
             const currentSuite = this.currentSuites[this.currentSuites.length - 1]
             currentSuite.suites.push(suite)
             this.currentSuites.push(suite)
-            this.suites[suite.uid] = suite
+            this.all_suites[suite.uid] = suite
             this.onSuiteStart(suite)
         })
 
@@ -145,7 +146,7 @@ export default class WDIOReporter extends EventEmitter {
         })
 
         this.on('suite:end',  /* istanbul ignore next */ (suite) => {
-            const suiteStat = this.suites[suite.uid]
+            const suiteStat = this.all_suites[suite.uid]
             suiteStat.complete()
             this.currentSuites.pop()
             this.onSuiteEnd(suiteStat)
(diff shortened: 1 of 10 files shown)


packages/wdio-reporter/tests/__mocks__/@wdio/reporter.js
```diff
@@ -24,6 +25,47 @@ export default class WDIOReporter extends EventEmitter {
...
+                    '',
+                    `${++failureLength}) ${suiteTitle} ${testTitle}`,
+                    chalk.red(test.error.message),
+                    ...test.error.stack.split(/\n/g).map(value => chalk.gray(value))
+                )
+            }
+        }
+
+        return output
+    }
```
christian-bromann: We should not reimplement functionality in a mock. We are just interested that it is executed with the right params and that it returns something

WillBrock: I can maybe unmock the reporter for those tests.

WillBrock:
It will print twice if someone uses the spec and dot reporter at the same time but only inside their reporter section. Do people use both at the same time though? I think the only way to prevent that would be to do the output after everything is done like in v4 in wdio-interface maybe or pass the config to the constructor of the reporter and look if both reporters are used.

dot-spec-reporter-output

packages/wdio-reporter/tests/__mocks__/@wdio/reporter.js
christian-bromann: Why, it should stay mocked. We want to isolate our unit tests.

WillBrock: I need to be able to verify the display and output of that method for the report so the mock won't work that way. I think I can maybe just create a util function with the output, call that in getFailureDisplay and unit test the util function for the report display. Then I can mock the call to getFailureDisplay.

christian-bromann: You should test `getFailureDisplay` as part of https://github.com/webdriverio/webdriverio/blob/master/packages/wdio-reporter/tests/reporter.test.js and then stub the methods in the mock and check that these are called.

christian-bromann:
I think it is fine that it prints it multiple times. It was the same behavior in v4. I just have one note for the dot reporter which is it would be cool if the error report could be printed after all `.` or `F` are printed on the screen. Do you think that is possible? I know this behavior is different in the spec reporter but it would be cool if the reporter can decide when to print these.

packages/wdio-reporter/tests/__mocks__/@wdio/reporter.js
WillBrock: oh duh, I'm an idiot :)

WillBrock:
I added the ability to print contents from a reporter after the main reporter output. As the code is now the dot reporter will act exactly the same as when it was in v4.

This adds a ```this.writeEnd()``` method which will print things after the main report.

Not sure if it would be helpful but instead of ```this.writeEnd``` we could do something like ```thts.write(`foo`, 2)``` where 2, 3, 4, etc. could be a print level so reporters could do as many levels as they would like.

Still need to add unit tests.

WillBrock:
I have to look at this a little bit more. The solution I had won't work for writing to files, need to come up with something for that.

WillBrock:
Two options come to mind with implementing this:

1) In v4 error messages would be spit out to the interface when tests failed, the reporter being used didn't have to implement this, e.g. spec and dot. It was built into the webdriverio code. We could maybe have this automatically done in wdio-reporter.

2) We could make the reporters(spec, dot, etc) have to implement this and the WDIOReporter class could provide the custom reporters a method to use that will output all of the messages at the end of the reporter output.

packages/wdio-reporter/src/index.js
```diff
@@ -183,11 +184,68 @@ export default class WDIOReporter extends EventEmitter {
...
+            ...suite.hooks
+                .filter((hook) => Boolean(hook.error))
+        ]
+    }
+
+    /**
+     * Get display for failed tests, e.g. stack trace
+     * @return {Array} Stack trace output
+     */
+    getFailureDisplay () {
```
nicholasbailey: Does this belong in wdio-reporter? It seems like each reporter should be responsible for determining the output format. This seems like it belongs in wdio-dot-reporter rather than the base reporter (which should be output format agnostic)

packages/wdio-reporter/src/index.js
```diff
@@ -183,11 +184,68 @@ export default class WDIOReporter extends EventEmitter {
...
+                }
+
+                const testTitle = test.title
+
+                // If we get here then there is a failed test
+                output.push(
+                    '',
+                    `${++failureLength}) ${suiteTitle} ${testTitle}`,
+                    chalk.red(test.error.message),
+                    ...test.error.stack.split(/\n/g).map(value => chalk.gray(value))
```
nicholasbailey: Can we please pull the update from wdio-spec reporter which handles multiple errors if they are present on the test? Jasmine will emit more than one error per test in an `errors`. See [https://github.com/webdriverio/webdriverio/pull/3702/files](https://github.com/webdriverio/webdriverio/pull/3702/files)

WillBrock:
@nicholasbailey The current code in this branch I am not planning on using, hence my last comment on the issue.

nicholasbailey:
Ok cool makes sense @WillBrock. Sorry about that!

mgrybyk:
Do we need this one to keep opened?

christian-bromann:
I don’t think so .. thanks dir putting this up @WillBrock

PR #3591, opened 2019-02-21
webdriverio: handle null returned by function selectors

jrobinson01:
## Proposed changes
Adds a `null` check to utils.getElementFromResponse so that commands like waitForExist, waitForDisplayed, etc can be used when $ is used with a function selector that returns null. This typically happens when the function selector uses `querySelector` and does not find an element.

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/packages/webdriverio/src/utils.js
+++ b/packages/webdriverio/src/utils.js
@@ -166,10 +166,16 @@ export const getPrototype = (scope) => {
 
 /**
  * get element id from WebDriver response
- * @param  {object} res         body object from response
- * @return {string|undefined}   element id or null if element couldn't be found
+ * @param  {?Object|undefined} res         body object from response or null
+ * @return {?string}   element id or null if element couldn't be found
  */
 export const getElementFromResponse = (res) => {
+    /**
+    * a function selector can return null
+    */
+    if (res === null || res === undefined) {
+        return null
+    }
     /**
      * deprecated JSONWireProtocol response
      */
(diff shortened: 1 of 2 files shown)


packages/webdriverio/src/utils.js
```diff
@@ -166,10 +166,16 @@ export const getPrototype = (scope) => {
...
+ * @param  {?Object|undefined} res         body object from response or null
+ * @return {?string}   element id or null if element couldn't be found
  */
 export const getElementFromResponse = (res) => {
+    /**
+    * a function selector can return null
+    */
+    if (res === null || res === undefined) {
+        return null
+    }
```
christian-bromann:
```suggestion
    }

```

packages/webdriverio/src/utils.js
```diff
@@ -166,10 +166,17 @@ export const getPrototype = (scope) => {
...
- * @param  {object} res         body object from response
- * @return {string|undefined}   element id or null if element couldn't be found
+ * @param  {?Object|undefined} res         body object from response or null
+ * @return {?string}   element id or null if element couldn't be found
  */
 export const getElementFromResponse = (res) => {
+    /**
+    * a function selector can return null
+    */
+    if (!res) {
```
mgrybyk: @jrobinson01 @christian-bromann can **res** be boolean?
I mean if res === false we will return null.

christian-bromann: It shouldn’t .. it is usually the request response body

PR #3600, opened 2019-02-22
Fix #3460

### LINKED ISSUE #3460: 'Socket hang up' errors started occurring consistently (but intermittently on different tests/steps) after 8-11 sceanarios in a wdio-cucumber-framework feature file
The following 'socket hang up' error occurs on different steps starting on the 8th to 11th scenario in a given feature file, where the previous scenarios, usually with the same sequence of steps as the failing one, run fine.  Up until last November, all were running fine and passing, for ~1yr.

The subsequent  4 tests also fail with 'socket hang up', while the following few tests will fail with error 'A session id is required for this command but wasn't found in the response payload', and then will start passing again.

A new session is started with each scenario, as browser.reload() gets called on starting a scenario.  The initial 'socket hang up' error occurs usually 4-5 steps into the scenario, while subsequent ones occur on the first step of the scenario.

This is on Chrome, and was working perfectly until ~last November.   Similar intermittent 'A session id is required for this command but wasn't found in the response payload' errors had always occurred on IE10 and IE11 (with no 'socket hang up').

Concurrent with the error, a browser window gets opened but doesn't get resized or attempt to navigate to the application url, and stays open while new windows get generated for new scenarios.

It happens with every feature file run, both using selenium-standalone locally, wdio testrunner and Selenium Grid Extras via Jenkins.

"Error: socket hang up
                    at createHangUpError (_http_client.js:302:15)
                    at Socket.socketOnEnd (_http_client.js:394:23)
                    at emitNone (events.js:91:20)
                    at Socket.emit (events.js:186:7)
                    at endReadableNT (_stream_readable.js:974:12)
                    at _combinedTickCallback (internal/process/next_tick.js:74:11)
                    at process._tickDomainCallback (internal/process/next_tick.js:122:9)"

selenium-standalone 3.141.0
 - **WebdriverIO version:** 4.8.0
 - **Mode:** Both 
 - **If WDIO Testrunner, running sync/async:**  sync
 - **Node.js version:**  8.9.0
 - **NPM version:** 5.6.0
 - **Browser name and version:**  Chrome 71.0.3578.98
   --Driver:  Chromedriver 2.45.0
 - **Platform name and version:** Any : Windows 10, Windows 8.1
wdio-cucumber-framework:  1.1.0

**Additional context**

With chromedriver versions <=2.43, the following concurrent error also occurs in the selenium log as noted below, with trace:  "java.lang.IllegalStateException: Committed".

With chromedriver 2.45, the same 'Socket hang up' error occurs as the step error in the wdio output, but the java.lang.illegalStateException: Committed is not output, and instead the selenium-standalone log shows multiple sessions created and removed without intervening output, as follows:

Starting ChromeDriver 2.45.615291 (ec3682e3c9061c10f26ea9e5cdcf3c53f3f74387) on port 42068
Only local connections are allowed.
15:51:45.018 INFO [ProtocolHandshake.createSession] - Detected dialect: OSS
15:51:45.033 INFO [RemoteSession$Factory.lambda$performHandshake$0] - Started new session ee949d62bb60cec80c97246fe5f21c13 (org.openqa.selenium.chrome.ChromeDriverService)
15:51:45.506 INFO [ProtocolHandshake.createSession] - Detected dialect: OSS
15:51:45.538 INFO [RemoteSession$Factory.lambda$performHandshake$0] - Started new session ff3758d7f2582b39ee02e3f6657ebabd (org.openqa.selenium.chrome.ChromeDriverService)
15:51:46.887 INFO [ProtocolHandshake.createSession] - Detected dialect: OSS
15:51:46.921 INFO [RemoteSession$Factory.lambda$performHandshake$0] - Started new session 40e0db5140996cd8a2e044f1901aeea6 (org.openqa.selenium.chrome.ChromeDriverService)
15:51:52.438 INFO [ActiveSessions$1.onStop] - Removing session 

This corresponds to logger output of messages being logged in before and after-scenario hooks, like

"in beforeScenario
in afterScenario
in beforeScenario
in afterScnenario"

without intervening logging, which normally occurs while a scenario is executed.  (The 'Socket hang up' error is output at then end of the run, as the Step/scenario error trace).

With the 'code to reproduce' zip file, as the application being tested is not available due to a firewall, a different baseUrl (with a valid server) is substituted.  In this case the 'Socket hang up' error is not, unfortunately, reproduced, but the same behavior with intermittent scenarios being aborted with an 'ECONNECT refused' error and a blank browser generated, while other scenarios attempt to get executed and only fail when failing to find specified selectors.  

Allure-report.zip where, if you filter out all passing and failed TCs, you can see each feature file with >9 scenarios hits the 'socket hang up' at TC8,9,10,11 or12:
[allure-report.zip](https://github.com/webdriverio/webdriverio/files/2828491/allure-report.zip)

mariocasciaro:
## Proposed changes

Lodash#merge modifies the first object provided as argument therefore it was modifying the default options in WebDriverRequest. This caused headers like 'Content-length' to propagate across requests causing problems such as a 'socket hang up' because the server is expecting more bytes than what the client is sending.

Should fix #3460 

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/packages/webdriver/src/request.js
+++ b/packages/webdriver/src/request.js
@@ -37,7 +37,7 @@ export default class WebDriverRequest extends EventEmitter {
     }
 
     makeRequest (options, sessionId) {
-        const fullRequestOptions = merge(this.defaultOptions, this._createOptions(options, sessionId))
+        const fullRequestOptions = merge({}, this.defaultOptions, this._createOptions(options, sessionId))
         this.emit('request', fullRequestOptions)
         return this._request(fullRequestOptions, options.connectionRetryCount)
     }
@@ -54,7 +54,7 @@ export default class WebDriverRequest extends EventEmitter {
          */
         if (this.body && (Object.keys(this.body).length || this.method === 'POST')) {
             requestOptions.body = this.body
-            requestOptions.headers = merge(requestOptions.headers, {
+            requestOptions.headers = merge({}, requestOptions.headers, {
                 'Content-Length': Buffer.byteLength(JSON.stringify(requestOptions.body), 'UTF-8')
             })
         }


osmolyar:
Hi @christian-bromann , thank you for the fix.  Will the above merge to webdriverio:master fix it in v4 as well?  (We cannot use v5 because we're using wdio-cucumber-framework).

christian-bromann:
> Will the above merge to webdriverio:master fix it in v4 as well?

No, feel free to make a PR to the v4 repo and I will release it.

osmolyar:
@christian-bromann - sorry, am confused about how to go about this (can't find v4 repo, only a tag in the releases tab).  

(though I notice @mariocasciaro 's comment on #3460 indicates that his scenario was already working perfectly on v4 but hopefully will be relevant to the original issue in #3460 also).

christian-bromann:
You can find the v4 repo here: https://github.com/webdriverio-boneyard/v4

osmolyar:
@christian-bromann , @mariocasciaro , there doesn't seem to be equivalent code to modify in v4 - there is no packages/webdriver/src/request.js, and lib/utils/RequestHandler.js doesn't have similar code either.  Perhaps this is not the fix to the original issue in #3460.  Will re-submit that issue in the v4 repo since it pertained to v4.

christian-bromann:
As mentioned before I don't have any capabilities to also maintain v4 so please provide a PR that I can release. Thanks!

osmolyar:
@mariocasciaro, if you are able to provide any pointers for investigation given that it doesn't appear to be a simple matter of finding analogous code in v4, it would be highly appreciated.

mariocasciaro:
@osmolyar I think it's very likely that the underlying issue is the same as the one in v5. Essentially for some reason the HTTP client must be sending a wrong 'Content-type' header or sending a Content-Type header with a GET request. The server is waiting indefinitely for data that never arrive.
Try to modify the v4 code to print out every outgoing header, it might give you some hints on what's going on.

osmolyar:
@mariocasciaro , thank you very much for the pointer (just seeing it now).  Will try to do as you suggest.

Incidentally, I've finally managed to reproduce the intermittent 'Socket hang up' issue in v4 using a non-proprietary page, Google Search, and have narrowed it down to the fact that the browser.reload() command must get called (every scenario, as in the original, or a few times a scenario as in the attached zip).
[google_search_reproduce_socket_hang_up.zip](https://github.com/webdriverio/webdriverio/files/2933389/google_search_reproduce_socket_hang_up.zip)

Also if you happen to have an idea of where the analogous code in v4 might be located that would also be very helpful.

osmolyar:
@mariocasciaro , thanks very much again for the tip to print out the request headers.  Unfortunately, even with the analogous fix in place as in the above mentioned pull request, the Socket hang up still occurs.

And when printing out newOptions.headers,  when the socket hang up occurs, the following Error on reload is sometimes output (same as before the fix), but there doesn't seem to be any issue with the request headers:
************newOptions.headers is *************{"Connection":"keep-alive","Accept":"application/json","User-Agent":"webdriverio/webdriverio/4.14.2"}
************newOptions.headers is *************{"Connection":"keep-alive","Accept":"application/json","User-Agent":"webdriverio/webdriverio/4.14.2"}
Setting viewport size
************newOptions.headers is *************{"Connection":"keep-alive","Accept":"application/json","User-Agent":"webdriverio/webdriverio/4.14.2"}
F.............In beforeScenario
.............************newOptions.headers is *************{"Connection":"keep-alive","Accept":"application/json","User-Agent":"webdriverio/webdriverio/4.14.2"}
.........************newOptions.headers is *************{"Connection":"keep-alive","Accept":"application/json","User-Agent":"webdriverio/webdriverio/4.14.2"}
Error in onReload hook: "Error: Don't call the 'init' command when using the wdio test runner. Your session will get initialised and closed automatically.
at execute() - C:\Users\osmolyar\AppData\Roaming\npm\node_modules\webdriverio\build\lib\commands\setViewportSize.js:87:12"
F........In beforeScenario
...............************newOptions.headers is *************{"Connection":"keep-alive","Accept":"application/json","User-Agent":"webdriverio/webdriverio/4.14.2"}

osmolyar:
@mariocasciaro , finally got the issue; it was the mere presence of a step definition file that contained this line

var req = http.request(testConfig.config.httpOptions, callback);

at the beginning of the file (rather than within a step definition) that was causing the issue, even if no steps defined in the file were being run.

 Thanks all for the help.

Issue #2962, opened 2018-09-20
Move away from calling Element Is Displayed endpoint for W3C-compliant drivers


burg:
## The problem

WebDriverIO currently expects to be able to use the endpoint GET /session/:sessionId/element/:elementId/displayed with W3C-compliant drivers. As this endpoint is not defined in the W3C specification, this is an invalid assumption. As a result, the request will return a 404. This is a serious functionality regression from non-W3C drivers.

The Selenium project has taken the approach of using the isDisplayed.js atom directly if the driver is speaking the W3C protocol dialect. I recommend taking this approach to maintain functionality for Safari 12 and other drivers that speak W3C protocol.

Selenium change: https://github.com/SeleniumHQ/selenium/commit/253fc6b579f23c799b618cfc778ddf1eed4f0d90

christian-bromann:
Thanks for notifying. I wasn't aware of this.

emilyrohrbough:
We have are experiencing this issue with using webdriver.io 4.13 while running against a selenium grid with IE nodes (v 3.14).

abjerstedt:
@burg - What am I missing on this? The endpoint listed above is indeed listed in the spec as the defacto method to ascertain element displayedness:
https://www.w3.org/TR/webdriver1/#element-displayedness
```
This function is typically exposed to GET requests with a URI Template of /session/{session id}/element/{element id}/displayed.
```

burg:
The algorithm is not specified with any semantics outside code, namely, the existing Selenium atom. There is no reason for drivers to implement a non-normative endpoint that can be polyfilled.

christian-bromann:
@burg what do you suggest? If the driver uses the Selenium atom anyway why should I need to polyfill it by myself?

christian-bromann:
Moving this out from v5 release as it doesn't block any functionality that needs to be ported.

christian-bromann:
@burg embedding Selenium Atoms into this project is not worthwhile. It adds too much unnecessary and undocumented code. Imo the definition of if an element is displayed or not should live on some sort of protocol level (maybe referencing it from a different spec). I understand that defining displayedness is not the job of the WebDriver protocol neither is it the job of the protocol consumer though.

burg:
Feel free to bring it up in the working group.

abjerstedt:
[...]

PR #3601, opened 2019-02-22
webdriverio: isDisplayed() change for browsers without the endpoint

abjerstedt:
## Proposed changes

This PR is changing the implementation of isDisplayed() to stop using the API, and instead use a script (provided by the apple team) to determine displayedness. The rationale for this, is that the webdriver spec no longer requires the usage of the api endpoint, and so, certain browsers such as edge and safari are no longer exposing it.

Resolves https://github.com/webdriverio/webdriverio/issues/2962.

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/packages/webdriverio/src/commands/element/isDisplayed.js
+++ b/packages/webdriverio/src/commands/element/isDisplayed.js
@@ -62,10 +62,13 @@ export default async function isDisplayed() {
         return false
     }
 
-    return browser.isChrome ?
-        await this.isElementDisplayed(this.elementId)
-        : await browser.execute(isElementDisplayedScript, {
+    //Only call the script on Safari and Edge
+    return Boolean(browser.capabilities.edge) ||
+    Boolean(browser.capabilities.safari)
+        ? await browser.execute(isElementDisplayedScript, {
             [ELEMENT_KEY]: this.elementId, // w3c compatible
             ELEMENT: this.elementId // jsonwp compatible
-        })
+        }) :
+        //Everyone else still uses the endpoint
+        await this.isElementDisplayed(this.elementId)
 }
--- a/packages/webdriverio/src/commands/element/isDisplayed.js
+++ b/packages/webdriverio/src/commands/element/isDisplayed.js
@@ -55,7 +55,12 @@ export default async function isDisplayed() {
         this.elementId = (await this.parent.$(this.selector)).elementId
     }
 
-    if (!this.elementId) return false
+    /*
+     * if element was still not found it also is not displayed
+    */
+    if (!this.elementId) {
+        return false
+    }
 
     return browser.isChrome ?
         await this.isElementDisplayed(this.elementId)
--- a/packages/webdriverio/src/commands/element/isDisplayed.js
+++ b/packages/webdriverio/src/commands/element/isDisplayed.js
@@ -44,7 +44,7 @@ import { ELEMENT_KEY } from '../../constants'
 import { getBrowserObject } from '../../utils'
 import isElementDisplayedScript from '../../scripts/isElementDisplayed'
 
-const noW3CEndpoint = ['MicrosoftEdge', 'safari']
+const noW3CEndpoint = ['microsoftedge', 'safari']
 
 export default async function isDisplayed() {
 
@@ -70,7 +70,7 @@ export default async function isDisplayed() {
      * no longer dictates it. In those instances, we pass the element through a script
      * that was provided by Brian Burg of safaridriver.
      */
-    return browser.isW3C && noW3CEndpoint.includes(browser.capabilities.browserName) ?
+    return browser.isW3C && noW3CEndpoint.includes(browser.capabilities.browserName.toLowerCase()) ?
         await browser.execute(isElementDisplayedScript, {
             [ELEMENT_KEY]: this.elementId, // w3c compatible
             ELEMENT: this.elementId // jsonwp compatible
--- a/packages/webdriverio/src/commands/element/isDisplayed.js
+++ b/packages/webdriverio/src/commands/element/isDisplayed.js
@@ -62,13 +62,19 @@ export default async function isDisplayed() {
         return false
     }
 
-    //Only call the script on Safari and Edge
-    return Boolean(browser.capabilities.edge) ||
-    Boolean(browser.capabilities.safari)
-        ? await browser.execute(isElementDisplayedScript, {
+
+    //Default is to use the endpoint
+    let useEndpoint = true
+
+    //Safari >= 12 and Edge 18
+    if (browser.isW3C && ['MicrosoftEdge', 'safari'].includes(browser.capabilities.browserName)) {
+        useEndpoint = false
+    }
+
+    return useEndpoint ?
+        await this.isElementDisplayed(this.elementId) :
+        await browser.execute(isElementDisplayedScript, {
             [ELEMENT_KEY]: this.elementId, // w3c compatible
             ELEMENT: this.elementId // jsonwp compatible
-        }) :
-        //Everyone else still uses the endpoint
-        await this.isElementDisplayed(this.elementId)
+        })
 }
--- a/packages/webdriverio/src/commands/element/isDisplayed.js
+++ b/packages/webdriverio/src/commands/element/isDisplayed.js
@@ -34,13 +34,19 @@
  * </example>
  *
  * @alias element.isDisplayed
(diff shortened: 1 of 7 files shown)


packages/webdriverio/tests/__mocks__/request.js
```diff
@@ -116,7 +116,9 @@ const requestMock = jest.fn().mockImplementation((params, cb) => {
     case `/wd/hub/session/${sessionId}/execute/sync`: {
         const script = Function(params.body.script)
         const args = params.body.args.map(arg => arg.ELEMENT || arg[ELEMENT_KEY] || arg)
-        value = script.apply(this, args) || {}
+        const result = script.apply(this, args)
+        //false and 0 are valid results
+        value = Boolean(result) || result === false || result === 0 ? result : {}
```
abjerstedt: || {} was swallowing valid responses

packages/webdriverio/src/commands/element/isDisplayed.js
```diff
@@ -34,20 +34,33 @@
...
+
     /*
      * This is only necessary as isDisplayed is on the exclusion list for the middleware
      */
     if (!this.elementId) {
         this.elementId = (await this.parent.$(this.selector)).elementId
     }
 
-    return this.elementId ? await this.isElementDisplayed(this.elementId) : false
+    if (!this.elementId) return false
```
christian-bromann: Let's do it with proper comment
```js
/*
 * if element was still not found it also is not displayed
 */
if (!this.elementId) {
    return false
}
```

abjerstedt: Done!

packages/webdriverio/src/commands/element/isDisplayed.js
```diff
@@ -34,20 +34,38 @@
...
 
-    return this.elementId ? await this.isElementDisplayed(this.elementId) : false
+    /*
+     * if element was still not found it also is not displayed
+    */
+    if (!this.elementId) {
+        return false
+    }
+
+    return browser.isChrome ?
```
wswebcreation: **This change it going to break all native app testing.** 
There should be a way to determine this is or Safari / Edge. Secondly we need to be aware of the fact that SafariDriver is also used for Hybrid apps. 

Challenges we have with apps:
We need to determine the if it is an app, but then we also got a challenge because of the following reasons

- app can be native, so no webview, and not using safari / chrome driver
- app is provided by the `app`-capability by default but there are cloud vendors that provide an extra capability to provide the app to there cloud, so we need to check multiple capabilities to determine if is is an app.

So determining if this is an app will be challenging. We might need to contact the Appium team to see if there is an easy way to determine if the safari driver is used during automation. Because you can also switch between context (native, where the XCUITest driver is user, or Webdriver where the safari driver is used.) during your tests

mgrybyk: Nice catch!

abjerstedt: Thanks @wswebcreation ! I wasn't exactly happy with this solution anyways, as I assumed it would cause more headaches than benefit.

packages/webdriverio/src/commands/element/isDisplayed.js
```diff
@@ -34,20 +34,47 @@
...
 
-    return this.elementId ? await this.isElementDisplayed(this.elementId) : false
+    /*
+     * if element was still not found it also is not displayed
+    */
+    if (!this.elementId) {
+        return false
+    }
+
+
```
christian-bromann:
```suggestion
```

2 double empty lines

packages/webdriverio/src/commands/element/isDisplayed.js
```diff
@@ -34,20 +34,47 @@
...
+    //Default is to use the endpoint
+    let useEndpoint = true
+
+    //Safari >= 12 and Edge 18
+    if (browser.isW3C && ['MicrosoftEdge', 'safari'].includes(browser.capabilities.browserName)) {
```
christian-bromann:
```suggestion
    if (browser.isW3C && browser.capabilities.browserName && ['microsoftedge', 'safari'].includes(browser.capabilities.browserName.toLowerCase())) {
```

also we can move `['microsoftedge', 'safari']` as constant to the top of the file.

packages/webdriverio/src/commands/element/isDisplayed.js
```diff
@@ -34,20 +34,47 @@
...
+        useEndpoint = false
+    }
+
+    return useEndpoint ?
```
christian-bromann: we don't use the `endpoint` variable anywhere else so just do the check above right here, also please add a comment

packages/webdriverio/src/commands/element/isDisplayed.js
abjerstedt: @wswebcreation Please re-review/ retest. 

I updated the logic to only run on Safari and Edge when running in w3c mode.  I'm honestly not really sure what to do about hybrid apps:
1.) If they are using safaridriver then the isDisplayed api no longer exists
2.) If they don't support execute, then they are kind of SOL

I was testing this last night, and I think we need to do a better job of determining who is w3c though. This Fix works on Edge 18 and Safari 12, but doesn't work on the previous versions of either (and I don't have the cleanest way to sniff the version other than browser.isW3C).

packages/webdriverio/src/commands/element/isDisplayed.js
abjerstedt: Done.

packages/webdriverio/src/commands/element/isDisplayed.js
abjerstedt: Done.

packages/webdriverio/src/commands/element/isDisplayed.js
abjerstedt: Done.

packages/webdriverio/src/commands/element/isDisplayed.js
```diff
@@ -34,20 +34,46 @@
...
+        return false
+    }
+
+    /*
+     * https://www.w3.org/TR/webdriver/#element-displayedness
+     * Certain drivers have decided to remove the endpoint as the spec
+     * no longer dictates it. In those instances, we pass the element through a script
+     * that was provided by Brian Burg of safaridriver.
+     */
+    return browser.isW3C && noW3CEndpoint.includes(browser.capabilities.browserName) ?
```
WillBrock: I don't think we need to the await's here since whatever calls isDisplayed will have to await on that anyways.

abjerstedt: Return await does a better job of error handling. I can pull up docs if you want (its why my middleware PR exposed those broken unit tests).

WillBrock: OK, I'm fine with how it is then :)

packages/webdriverio/src/commands/element/isDisplayed.js
```diff
@@ -34,20 +34,46 @@
...
  * @uses protocol/elements, protocol/elementIdDisplayed
  * @type state
  *
  */
 
+import { ELEMENT_KEY } from '../../constants'
+import { getBrowserObject } from '../../utils'
+import isElementDisplayedScript from '../../scripts/isElementDisplayed'
+
+const noW3CEndpoint = ['MicrosoftEdge', 'safari']
```
christian-bromann: What if people write this browser name with lower case? or just write `edge`. Are the different ways to define edge as a browser in the capabilities. What we definitely need to do though is to lowercase this.

abjerstedt: Done.

wswebcreation:
@christian-bromann and @abjerstedt 
This will still not work for hybrid apps where the safari browser is used  to automate the webview. We can't determine it because it's not in the capabilities (the driver is started with the app) and after switching the context the safari driver will be used.

I'll check if I have some time the coming days to see if I can validate that with the latest version.

christian-bromann:
@wswebcreation I think it will be impossible to determine if it is available with Appium, no? Shouldn't Appium fall back to the atom if isDisplayed is not available? We could also accept a flag to the isDisplayed command to manually decide which mode to use.

wswebcreation:
@christian-bromann 
Let me dive into it and see how it goes. I'll update you both on it

Issue #3608, opened 2019-02-23
element.getElementRect omits `y` coordinate


flunderpero:
**Environment (please complete the following information):**
 - **WebdriverIO version:** [e.g. 5.6.2]
 - **Mode:** [WDIO Testrunner]
 - **If WDIO Testrunner, running sync/async:** [sync]
 - **Node.js version:** [e.g. 10.13.0]
 - **NPM version:** [e.g. 6.4.1]
 - **Browser name and version:** [Safari 12.0.3]
 - **Platform name and version:** [OSX Mojave]
 
**Config of WebdriverIO**
``` 
{
    capabilities: [{browserName: "safari"}],
    port: 4447,
    seleniumArgs: {seleniumArgs: ["-port", "4447"]},
    services: [ "selenium-standalone"],
}

// Problem exists in BrowserStack Safari, too!
```
**Describe the bug**
Calling `getElementRect` never returns the `y`-value. Note: I do not call `getElementRect` myself
but rather ran into the bug while trying to use `element.moveTo()` which in turn calls `getElementRect` on Safari.

**To Reproduce**
```
    // This will not work because the moveTo-command tries to use a null value for y
    $("any element").moveTo()
``` 

**Expected behavior**
It should work. :-)

christian-bromann:
@flunderpero can you provide logs please?

flunderpero:
Sorry, I forgot. This is what happens when I try to call `element.moveTo()`:

```
2019-02-23T16:36:18.837Z INFO webdriver: COMMAND getElementRect("node-F85D9EB9-CBD9-4D7E-82D8-A29F2849EEE3")
2019-02-23T16:36:18.837Z INFO webdriver: [GET] http://127.0.0.1:4447/wd/hub/session/0DCB497F-76E4-4E29-93E5-4035A6E5669E/element/node-F85D9EB9-CBD9-4D7E-82D8-A29F2849EEE3/rect
2019-02-23T16:36:18.887Z INFO webdriver: RESULT { x: 102, width: 578, height: 11 }
2019-02-23T16:36:18.888Z INFO webdriver: COMMAND performActions(<object>)
2019-02-23T16:36:18.889Z INFO webdriver: [POST] http://127.0.0.1:4447/wd/hub/session/0DCB497F-76E4-4E29-93E5-4035A6E5669E/actions
2019-02-23T16:36:18.889Z INFO webdriver: DATA { actions:
   [ { type: 'pointer',
       id: 'finger1',
[...]

PR #3610, opened 2019-02-23
webdriverio: workaround for Safari to moveTo, getLocation, dragAndDrop, getSize

mgrybyk:
## Proposed changes

Because of defect in safari https://github.com/SeleniumHQ/selenium/issues/6637 `getElementRect` is not return `y`.
We are calculating x and y offset so that is one of x, y, width, height is not defined will get `NaN` as a result.

Proposal: try get rect with javascript or set 0 as default value in case on of x, y, width, height is not defined

Affected browser: Safari 12
Affected commands: moveTo, getLocation, dragAndDrop, getSize

fix for #3608 

## Types of changes

## Checklist

### DIFF
--- a/packages/webdriverio/src/utils.js
+++ b/packages/webdriverio/src/utils.js
@@ -3,9 +3,11 @@ import path from 'path'
 import cssValue from 'css-value'
 import rgb2hex from 'rgb2hex'
 import GraphemeSplitter from 'grapheme-splitter'
+import logger from '@wdio/logger'
 
 import { ELEMENT_KEY, W3C_SELECTOR_STRATEGIES, UNICODE_CHARACTERS } from './constants'
 
+const log = logger('webdriverio')
 const DEFAULT_SELECTOR = 'css selector'
 const DIRECT_SELECTOR_REGEXP = /^(id|css selector|xpath|link text|partial link text|name|tag name|class name|-android uiautomator|-ios uiautomation|-ios predicate string|-ios class chain|accessibility id):(.+)/
 const INVALID_SELECTOR_ERROR = new Error('selector needs to be typeof `string` or `function`')
@@ -392,4 +394,46 @@ export function verifyArgsAndStripIfElement(args) {
     }
 
     return !Array.isArray(args) ? verify(args) : args.map(verify)
+}
+
+/**
+ * getElementRect
+ */
+export async function getElementRect(scope) {
+    const rect = await scope.getElementRect(scope.elementId)
+
+    let defaults = { x: 0, y: 0, width: 0, height: 0 }
+
+    /**
+     * getElementRect workaround for Safari 12.0.3
+     * if one of [x, y, height, width] is undefined get rect with javascript
+     */
+    if (Object.keys(defaults).some(key => rect[key] == null)) {
+        /* istanbul ignore next */
+        const rectJs = await getBrowserObject(scope).execute(function (el) {
+            if (!el || !el.getBoundingClientRect) {
+                return
+            }
+            const { left, top, width, height } = el.getBoundingClientRect()
+            return {
+                x: left + this.scrollX,
+                y: top + this.scrollY,
+                width,
+                height
+            }
+        }, scope)
+
+        // try set proper value
+        Object.keys(defaults).forEach(key => {
+            if (rect[key] == null && typeof rectJs[key] === 'number') {
+                rect[key] = Math.floor(rectJs[key])
+            } else {
+                log.error('getElementRect', rect)
+                log.error('getBoundingClientRect', rectJs)
+                throw new Error('Failed to receive element rects via execute command')
+            }
+        })
+    }
+
+    return rect
 }
--- a/packages/webdriverio/tests/commands/element/moveTo.test.js
+++ b/packages/webdriverio/tests/commands/element/moveTo.test.js
@@ -35,6 +35,34 @@ describe('moveTo', () => {
             .toEqual({ type: 'pointerMove', duration: 0, x: 20, y: 30 })
     })
 
+    it('should do a moveTo with params if getElementRect returned empty object', async () => {
+        const browser = await remote({
+            baseUrl: 'http://foobar.com',
+            capabilities: {
+                browserName: 'foobar'
+            }
+        })
+
+        const elem = await browser.$('#elem')
+        request.setMockResponse([{}, { x: 5, y: 10, height: 33, width: 44 }])
+        await elem.moveTo(5, 10)
+        expect(request.mock.calls[4][0].body.actions[0].actions[0])
+            .toEqual({ type: 'pointerMove', duration: 0, x: 10, y: 20 })
+    })
+
+    it('should do a moveTo with params if getElementRect and getBoundingClientRect returned empty object', async () => {
+        const browser = await remote({
+            baseUrl: 'http://foobar.com',
+            capabilities: {
+                browserName: 'foobar'
+            }
+        })
+
+        const elem = await browser.$('#elem')
+        request.setMockResponse([{}, {}])
+        expect(elem.moveTo(5, 10)).rejects.toThrow('Failed to getElementRect')
+    })
+
     it('should do a moveTo without params (no-w3c)', async () => {
         const browser = await remote({
             baseUrl: 'http://foobar.com',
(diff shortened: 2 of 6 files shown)


mgrybyk:
We can also "improve" getLocation behavior to work in hacky way so that it will try get location with javascript if regular getLocation / getRect failed to do it. But I would not like to do that


```
// ...
    let locationJs = { x: 0, y: 0, width: 0, height: 0 }

    // if one of [x, y, height, width] is undefined get location with javascript
    if (Object.keys(locationJs).some(key => location[key] == null)) {
        locationJs = await browser.execute(function (el) {
            const { left, top, width, height } = el.getBoundingClientRect();
            return {
                x: left + this.scrollX,
                y: top + this.scrollY,
                width,
                height
            };
        }, this)

        // try set proper value
        Object.keys(locationJs).forEach(key => {
            if (location[key] == null) {
                location[key] = typeof locationJs[key] === 'number' ? Math.floor(locationJs[key]) : 0
            }
        })
    }
```

flunderpero:
This would help, next command I need is `dragAndDrop` which will likely not work either. Other commands might suffer from this bug, too.
In my humbled opinion it would be better to fix `getElementRect`. 

Tomorrow I will try to use the TechnologyPreview of Safari to see if the bug is gone with that.

mgrybyk:
@flunderpero getLocation and dragAndDrop will not work

mgrybyk:
I'm not sure if there is a value in this PR in such case.

mgrybyk:
@flunderpero this should fix `getElementRect`.

However I'm not sure if it is a good idea to add such hacks to the framework

abjerstedt:
I'm honestly not sure how i feel about this fix. I got blocked by this exact bug last week, but I don't think we should use an override just to support safari 12 (until the next release). 

I think the better behavior is to put a hack into any dependency commands to have a default value of 0 in the event that the driver returns a null.

mgrybyk:
I think both solutions are not really good. 
Return 0 instead of undefined will not work either.

It would be great to throw exception instead because we are expecting to get some value and failed.

Closing this PR for now

christian-bromann:
> Return 0 instead of undefined will not work either.

Why is that?

mgrybyk:
> Why is that?

imagine there is a button in the middle of the screen with coords x: 600, y: 600.

moveTo (and other commands that use getRect) won't work because x: 600, y: 0 are not valid coords of the element.

christian-bromann:
> moveTo (and other commands that use getRect) won't work because x: 600, y: 0 are not valid coords of the element.

ok, I was assuming that Safaridriver only fails returning `y` in certain edge cases not all the time.

abjerstedt:
@christian-bromann Yeahhh.... Safaridriver for v12 is just flat omitting the Y coordinate in getElementRect. 

I logged a bug directly with apple, and they asked for stacktraces :-/. However, it sounds like they already fixed it for the next safari release.

christian-bromann:
I think it makes sense to include this workaround for all safari sessions until majority of people use the fixed safaridriver, what do you think @mgrybyk ?

mgrybyk:
@christian-bromann I don't mind, it is up to you to decide.
For me fix looks more or less clean and we can always replace function `getElementRect` with `scope.getElementRect(scope.elementId)`

mgrybyk:
And even better, once it is fixed in Safari we can leave verification of getElementRect to be sure it has returned x, y, height and width

christian-bromann:
@mgrybyk Let's reopen then

christian-bromann:
@mgrybyk @abjerstedt any objections to merge this?

packages/webdriverio/tests/commands/element/moveTo.test.js
```diff
@@ -35,6 +35,21 @@ describe('moveTo', () => {
...
+    it('should do a moveTo with params if getElementRect returned empty object', async () => {
+        const browser = await remote({
+            baseUrl: 'http://foobar.com',
+            capabilities: {
+                browserName: 'foobar'
+            }
+        })
+
+        const elem = await browser.$('#elem')
+        request.setMockResponse([{}, { x: 5, y: 10, height: 33, width: 44 }])
```
christian-bromann: How is this testing the edge case behavior if all values are given? Shouldn't one be omitted?

mgrybyk: this is the case when getElementRect has not returned any value but execute script returned all the values.

@christian-bromann btw, what should we do in case we have failed to get one of x,y,width or height with JS. Currently it is fallback to 0. But I think we'd better throw an exception.

Can you judge this? :)
@abjerstedt @christian-bromann

christian-bromann: I think `el.getBoundingClientRect` will always return proper values therefor I would also say we can remove `if (!el || !el.getBoundingClientRect) return`, no?

mgrybyk: @christian-bromann I have a failing test because of that
`TypeError: el.getBoundingClientRect is not a function`

don't know how to mock that. Anyway it is DOM interaction

christian-bromann: so if the element isn't found we should throw or are you talking about webdriverio unit test. Also could you rebase the branch to current master?

mgrybyk: I'm talking about both.

First of all we should throw an exception if getElementRect failed and then getBoundingClientRect failed as well (for whatever reason)

Secondly I have problems with unit test, but I think additional verification is not a problem `if (!el || !el.getBoundingClientRect) return`

Doing rebase, added test for throw error, and will push after it soon

mgrybyk: @christian-bromann updated

packages/webdriverio/src/utils.js
```diff
@@ -393,3 +395,43 @@ export function verifyArgsAndStripIfElement(args) {
...
+        }, scope)
+
+        // try set proper value
+        Object.keys(defaults).forEach(key => {
+            if (rect[key] == null && typeof rectJs[key] === 'number') {
+                rect[key] = Math.floor(rectJs[key])
+            } else {
+                log.error('getElementRect', rect)
+                log.error('getBoundingClientRect', rectJs)
+                throw new Error('Failed to getElementRect')
```
christian-bromann:
```suggestion
                throw new Error('Failed to receive element rects via execute command')
```

packages/webdriverio/src/utils.js
```diff
@@ -393,3 +395,43 @@ export function verifyArgsAndStripIfElement(args) {
...
+                x: left + this.scrollX,
+                y: top + this.scrollY,
+                width,
+                height
+            }
+        }, scope)
+
+        // try set proper value
+        Object.keys(defaults).forEach(key => {
+            if (rect[key] == null && typeof rectJs[key] === 'number') {
```
christian-bromann:
```suggestion
            if (rect[key] === null && typeof rectJs[key] === 'number') {
```

packages/webdriverio/src/utils.js
```diff
@@ -393,3 +395,43 @@ export function verifyArgsAndStripIfElement(args) {
...
+export async function getElementRect(scope) {
+    const rect = await scope.getElementRect(scope.elementId)
+
+    let defaults = { x: 0, y: 0, width: 0, height: 0 }
+
+    /**
+     * getElementRect workaround for Safari 12.0.3
+     * if one of [x, y, height, width] is undefined get rect with javascript
+     */
+    if (Object.keys(defaults).some(key => rect[key] == null)) {
```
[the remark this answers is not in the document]
mgrybyk: @christian-bromann I want to perform check for null and undefined.
I can do it easily in one line rect[key] == null
or can do rect[key] === null and rect[key] === undefined

I'd better keep rect[key] == null, if you don't mind

packages/webdriverio/src/utils.js
mgrybyk: @christian-bromann same here

zuhito:
The fixed version of this problem is v5.x.x only?
I'd like to use the fixed version in v4.x.x. 🙏

christian-bromann:
> I'd like to use the fixed version in v4.x.x. 🙏

Please provide a PR and I can release it for v4.

zuhito:
I tried to change the code for v4.x.x but I seem not to be able to use getBoundingClientRect() in the code...

Issue #3602, opened 2019-02-22
Strip down element object when passed into execute commands


christian-bromann:
## Problem

With WebdriverIO you can pass in an element into the execute command to use it in the script , e.g.
```js
const body = $('body')
console.log(browser.execute((el) => el.nodeName, body)) // returns "BODY"
```
Right now WebdriverIO would serialize the whole element object including all properties and methods. This is unnecessary and requires to much bandwidth.

## Solution

In execute and executeAsync, check if any of the parameters is an element object and strip the object down to elementId (jsonwp and w3c conform).

baruchvlz:
Right now the callback returns the element id by default, no?

```javascript
const body = $('body')
console.log(browser.execute((el) => el, body))
// returns { 'element-6066-11e4-a52e-4f735466cecf': 'some-elem-123' }
```

Do you want this return to change to? (with all the other properties stripped)

```javascript
const body = $('body')
console.log(browser.execute((el) => el, body))
// returns { elementId: 'some-elem-123' }
```

christian-bromann:
@baruchvlz interesting point, I am not sure if we need to because if people want to fetch an element with a execute function they can just do:

```js
const body = $(() => document.body)
```

However I wonder if it would make sense to automatically transform an element response from execute into an actual wdio element. This would be valuable if you want to pass in parameters into the function call as you showed above. This would be definitely better handled in a different PR.

baruchvlz:
@christian-bromann This can be closed, eh?

christian-bromann:
jup .. thanks!

PR #3612, opened 2019-02-24
Strip Element objects from execute and executeAsync arguments

baruchvlz:
## Proposed changes

Ref: #3602 

## Types of changes

## Checklist

## Further comments

I'm not sure if the docs should be updated to tell the reader that Element object arguments will be stripped of all properties except `elementId`.

### DIFF
--- a/packages/webdriverio/src/utils.js
+++ b/packages/webdriverio/src/utils.js
@@ -379,7 +379,7 @@ export function verifyArgsAndStripIfElement(args) {
     function verify(arg) {
         if (arg.constructor.name === 'Element') {
             if (!arg.elementId) {
-              throw new Error('property "elementId" missing')
+                throw new Error(`The element with selector "${arg.selector}" you trying to pass into the execute method wasn't found`)
             }
 
             return {
--- a/packages/webdriverio/src/utils.js
+++ b/packages/webdriverio/src/utils.js
@@ -370,3 +370,11 @@ export async function findElements(selector) {
 
     throw INVALID_SELECTOR_ERROR
 }
+
+/**
+ * Strip element object of all propeties except `elementId`
+ */
+
+export function verifyArgsAndStripIfElement(args) {
+    return args.map(a => a.elementId ? { elementId: a.elementId } : a)
+}
--- a/packages/webdriverio/src/utils.js
+++ b/packages/webdriverio/src/utils.js
@@ -372,9 +372,24 @@ export async function findElements(selector) {
 }
 
 /**
- * Strip element object of all propeties except `elementId`
+ * Strip element object and return w3c and jsonwp compatible keys
  */
 
 export function verifyArgsAndStripIfElement(args) {
-    return args.map(a => a.elementId ? { elementId: a.elementId } : a)
+    function verify(arg) {
+        if (arg.constructor.name === 'Element') {
+            if (!arg.elementId) {
+              throw new Error('property "elementId" missing')
+            }
+
+            return {
+                [ELEMENT_KEY]: arg.elementId,
+                ELEMENT: arg.elementId
+            }
+        }
+
+        return arg
+    }
+
+    return !Array.isArray(args) ? verify(args) : args.map(verify)
 }
--- a/packages/webdriverio/src/commands/browser/execute.js
+++ b/packages/webdriverio/src/commands/browser/execute.js
@@ -1,3 +1,4 @@
+import { verifyArgsAndStripIfElement } from '../../utils'
 /**
  *
  * Inject a snippet of JavaScript into the page for execution in the context of the currently selected frame.
@@ -52,5 +53,5 @@ export default function execute (...args) {
         script = `return (${script}).apply(null, arguments)`
     }
 
-    return this.executeScript(script, args)
+    return this.executeScript(script, verifyArgsAndStripIfElement(args))
 }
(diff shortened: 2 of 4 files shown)


baruchvlz:
Force push was made to update the commit message in accordance with the contributing guide.

abjerstedt:
If our goal is to clean out webdriverio (not webdriver) element objects, wouldn't it be easier just to check the object name (since they are always named element) and then:
1.) Throw an error if the elementId doesn't exist
2.) Create a new object with the 2 element props

packages/webdriverio/src/utils.js
```diff
@@ -370,3 +370,11 @@ export async function findElements(selector) {
 
     throw INVALID_SELECTOR_ERROR
 }
+
+/**
+ * Strip element object of all propeties except `elementId`
+ */
+
+export function verifyArgsAndStripIfElement(args) {
+    return args.map(a => a.elementId ? { elementId: a.elementId } : a)
```
abjerstedt: This doesn't work. No browser consumes the elementId property. It needs to be ELEMENT, or the jsonwp version.

abjerstedt: import { ELEMENT_KEY } from '../../constants'

{
  [ELEMENT_KEY]: this.elementId, // w3c compatible
  ELEMENT: this.elementId // jsonwp compatible
}

baruchvlz: Yea, this makes sense. I'm still very new to all this w3c and jsonwp compatibility, I appreciate your feedback. I will try to adjust the code based on your comment and the suggestion in the review. Cheers.

baruchvlz: @abjerstedt Before I push anything, I want to be sure that I understood you correctly.

These are the tests I added based on your comments, is this what you meant?

```javascript
describe('verifyArgsAndStripIfElement', () => {
    it('returns the same value if it is not an element object', () => {
        expect(verifyArgsAndStripIfElement([1, 2, 3])).toEqual([1, 2, 3])
    })

    it('strips down properties if value is element object', () => {
        const fakeObj = new (class Element {
            constructor() {
                this.elementId = 'foo-bar'
                this.someProp = 123
                this.anotherProp = 'abc'
            }
        })()

        expect(verifyArgsAndStripIfElement([fakeObj, 'abc', 123])).toMatchObject([
            { [ELEMENT_KEY]: 'foo-bar', ELEMENT: 'foo-bar' },
            'abc',
            123
        ])
    })

    it('should work even if parameter is not of type Array', () => {
        const fakeObj = new (class Element {
            constructor() {
                this.elementId = 'foo-bar'
                this.someProp = 123
                this.anotherProps = 'abc'
            }
        })()

        expect(verifyArgsAndStripIfElement(fakeObj)).toMatchObject(
            { [ELEMENT_KEY]: 'foo-bar', ELEMENT: 'foo-bar' }
        )
        expect(verifyArgsAndStripIfElement('foo')).toEqual('foo')
    })

    it('throws error if element object is missing element id', () => {
        const fakeObj = new (class Element {
            constructor() {
                this.someProp = 123
                this.anotherProps = 'abc'
            }
        })()

        expect(() => verifyArgsAndStripIfElement(fakeObj)).toThrow('property "elementId" missing')
    })
})
```

I'm sure using `new (class Element {})()` is not ideal, but I don't know how else to mimic the class object in order to have `if (arg.constructor.name === 'Element')` in the function

christian-bromann: I think this looks fine, I would define the class once at the top of the file though instead of redefining it in every test.

packages/webdriverio/src/utils.js
```diff
@@ -370,3 +370,26 @@ export async function findElements(selector) {
...
+
+/**
+ * Strip element object and return w3c and jsonwp compatible keys
+ */
+
+export function verifyArgsAndStripIfElement(args) {
+    function verify(arg) {
+        if (arg.constructor.name === 'Element') {
+            if (!arg.elementId) {
+              throw new Error('property "elementId" missing')
```
christian-bromann: wrong indentation (add 2 spaces), also let's add a more meaningful message like `The element with selector "${arg.selector}" you trying to pass into the execute method wasn't found`.

baruchvlz:
@christian-bromann I had a fresh install of the repo in my machine and `npm install` is failing, same error as in the Travis buld.

Yarn works though.

christian-bromann:
> `npm install` is failing, same error as in the Travis buld.

yup, due to https://github.com/highlightjs/highlight.js/issues/1984 .. hope they can put out a fix soon.

baruchvlz:
@christian-bromann before you re-run the pipeline, I have to fix the spec for the error if `elementId` is missing, I forgot to update it after I changed the text to what you suggested.

mgrybyk:
Hey, just want to double check, in case I'm passing element to function, will it be still possible to get element properties and call functions, ex  element.dispatchEvent, etc?

baruchvlz:
@mgrybyk Aren't those methods attached during `executeScript`?

christian-bromann:
> Hey, just want to double check, in case I'm passing element to function, will it be still possible to get element properties and call functions, ex element.dispatchEvent, etc?

Yes. Everything the driver needs is to pass in the element (e.g. jsonwp: `{ ELEMENT: "..." }`) if the driver detects such object it will replace it with the actual DOM Node element. This PR strips out all the unnecessary wdio properties from the element which is not needed by the driver.

PR #3365, opened 2019-01-17
blog for file upload example for V5

caoxu2000:
## Proposed changes

[//]: # (Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue.)

## Types of changes

[//]: # (What types of changes does your code introduce to WebdriverIO?)
[//]: # (_Put an `x` in the boxes that apply_)

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

[//]: # (_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._)

- [ ] I have read the [CONTRIBUTING](https://github.com/webdriverio/webdriverio/blob/master/CONTRIBUTING.md) doc
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

## Further comments

[//]: # (If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...)

PR #3632, opened 2019-02-26
Upload blog post

christian-bromann:
copied #3365 from due to inability to make PR against fork.


packages/webdriverio/src/commands/browser/uploadFile.js
```diff
@@ -0,0 +1,47 @@
+/**
+ *
+ * Uploads a file to the selenium server by using the [`file`](/api/protocol/file.html) command. Note that
+ * this command might not be supported as it is an undocumented Selenium command.
```
christian-bromann: Can we add a code example with an actual page like https://the-internet.herokuapp.com/upload ?

website/blog/file-upload-example.md
```diff
@@ -0,0 +1,47 @@
...
+```
+If you want to upload file to a remote machine (e.g. SauceLab's VM) on which the browser is running you could use this Chrome only non official and undocumented Chromium command. Please note the file has to be Base64-encoded zip archive containing single file which to upload. Please refer [uploadFile](https://webdriver.io/docs/api/chromium.html#uploadfile)
+I haven't tested it yet but you can let me know if this works for you.
+```js
+const path = require('path');
+const filePath = path.join(__dirname, 'path/to/your/file');
+
+const remoteFilePath = browser.uploadFile(filePath);
+$('upload file input selector').setValue(remoteFilePath);
+
```
christian-bromann:
```suggestion
```

christian-bromann:
@caoxu2000 can you address the outlining things above and then we are good to go

erduna:
Is there an ETA on this merge? Or does it make sense for me just to include the code locally?

christian-bromann:
@caoxu2000 what is the state of this?

caoxu2000:
will reopen a new PR for this

erinev:
Hi, i wanted to know what is the status with this pull request ? we are onboarding webdriver-io v5 framework to our company and remote (selenium grid) file upload is blocker for us to fully integrate framework.  

any ETA when this PR will be merged/released ?  

will this remote file upload work for LINUX and WINDOWS platform ? (because in webdriver-io v4 different team faced issue that chooseFile is working only in LINUX platform)

xemicalas:
Hi, @christian-bromann @caoxu2000  any news when this will be resolved? This is a blocker for us, cause we cannot use remote file uploads.

erduna:
Also share same sentiment as last two commenters :)

caoxu2000:
@erinev @xemicalas @erduna sorry guys I understand the need for this but I couldn't make the integration test to work.  We have the following checks if the file command is available for the browser:
```
if (typeof this.file !== 'function') {
   throw new Error(`The uploadFile command is not available in ${this.capabilities.browserName}`)
 }
```
Currently this check fails for Chrome which is supposed to support.  I was stuck at this.
I may have some time down the road to take another look at it but no guarantee it's gonna happen.  You guys can take a stab at it if you want.  Or have somebody else to pick it up.  Trust me I'm frustrated on this too :(

erinev:
@caoxu2000 i managed to make it work in my local project with customCommand (I adapted your code). My solution:

1) i added custom command:
```
const fs = require('fs');
const path = require('path');
const archiver = require('archiver');

browser.addCommand('convertToRemoteFile', (localFilePath) => {
  const zipData = [];
  const source = fs.createReadStream(localFilePath);

  return new Promise((resolve, reject) => {
    archiver('zip')
      .on('error', err => reject(err))
      .on('data', data => zipData.push(data))
      .on('end', () => browser.uploadFile(Buffer.concat(zipData).toString('base64')).then(resolve, reject))
      .append(source, { name: path.basename(localFilePath) })
      .finalize((err) => {
        if (err) {
          reject(err);
        }
      });
  });
});
```
(note: i call `browser.uploadFile` instead `this.file`)  

2) usage is like this:
we have file input like this: `<input type="file" name="files" multiple="multiple">`
```
const localFilePath = 'C:\\Adform\\Development\\AdsFlow\\e2e-automated-tests-framework\\test-files\\imageBanner.png';
const remoteFile = browser.convertToRemoteFile(localFilePath);
$('input[type="file"]').setValue(remoteFile);
```

It works on this environment:
WIN 10 & 8.1
node: 10.15.0
Chrome browser: 74 & 58
webdriver-io: 5.8.1
local selenium hub & remote selenium hub

erinev:
as well i hought that it is inconvinient for users to use some convertToRemotFile command. I introduced one more custom command on all elements (but ofcourse it will work only for file input elements):  
```
  static _registerAddFileCommand() {
    browser.addCommand('addFile', function (localFilePath) {
      const remoteFile = browser.convertLocalFileToRemoteFile(localFilePath);
      this.setValue(remoteFile);
    }, this._enableForAllElements);
  }
```
and usage is more logical now: `$('input[type="file"]').addFile('C:\\someProject\\test-files\\image.png');`
i tested a bit more and it works on Chrome 64 (Linux) too

klamping:
@erinev thanks for sharing that code! will be helpful as I'm trying to tackle this exact issue.

packages/webdriverio/src/commands/browser/uploadFile.js
```diff
@@ -0,0 +1,47 @@
...
+ * @uses protocol/file
+ *
+ */
+import fs from 'fs'
+import path from 'path'
+import archiver from 'archiver'
+
+export default async function uploadFile (localPath) {
+    /**
+     * parameter check
```
CrispusDH: I'm not sure that it is useful comment

packages/webdriverio/src/commands/browser/uploadFile.js
```diff
@@ -0,0 +1,47 @@
...
+ * @param {String} localPath local path to file
+ * @type utility
+ * @uses protocol/file
+ *
+ */
+import fs from 'fs'
+import path from 'path'
+import archiver from 'archiver'
+
+export default async function uploadFile (localPath) {
```
CrispusDH: maybe we can name it in a same way as in protocols? just `file`

christian-bromann: the protocol expects a base64 encoded file and this command a file path so I think this is fine

packages/webdriverio/src/commands/browser/uploadFile.js
christian-bromann: it is because the `fs.createReadStream` would choke with a undescriptive error

packages/webdriverio/src/commands/browser/uploadFile.js
```diff
@@ -0,0 +1,59 @@
...
+     * parameter check
+     */
+    if (typeof localPath !== 'string') {
+        throw new Error('number or type of arguments don\'t agree with uploadFile command')
+    }
+
+    /**
+     * check if command is available
+     */
+    if (typeof this.file !== 'function') {
```
erwinheitzman: `this` refers to the Browser object no?

packages/webdriverio/src/commands/browser/uploadFile.js
```diff
@@ -0,0 +1,59 @@
+/**
+ *
+ * Uploads a file to the selenium server by using the [`file`](/api/protocol/file.html) command. Note that
```
erwinheitzman: `Note that this command might not be supported as it is an undocumented Selenium command.`
This is part still true?

packages/webdriverio/src/commands/browser/uploadFile.js
christian-bromann: yes

packages/webdriverio/src/commands/browser/uploadFile.js
christian-bromann: Actually we should mention here that it is only supported in Chrome and using Selenium standalone

artemv:
A v4-like chooseFile command for latest wdio (v5.16.7):
```
    browser.addCommand('chooseFile', function (localFilePath) {
        const remoteFile = browser.uploadFile(localFilePath);
        this.addValue(remoteFile);
    }, true);
```
Usage is like before:
```
$('input[type=file]').chooseFile(localFilePath)
```

Note that this uses `addValue` instead of `setValue` to avoid "element not interactable" error that happens if the input file is hidden/zero size (which is the case for Dropzone uploads).

PR #3644, opened 2019-02-27
Proposal: add onTestFramework hook

mgrybyk:
## Proposed changes

Proposal: add hook before test execution but after test framework is ready and its functions are wrapped with Fibers.

My use case. I'm decorating Mocha's describe and it objects to extend its functionality.
There is no place I can do it just before starting any test when test framework is initialised.

## Types of changes

## Checklist

## Further comments

Some tests and documentation are still missing, want to make I'm not doing something wrong before proceeding.

### DIFF
--- a/docs/ConfigurationFile.md
+++ b/docs/ConfigurationFile.md
@@ -336,6 +336,12 @@ exports.config = {
     onReload: function(oldSessionId, newSessionId) {
     },
     /**
+    * Gets executed once test framework (mocha, jasmine) has been initialised.
+    * At this moment global objects like `describe` or `it` are available.
+    */
+    onTestFramework: function() {
+    },
+    /**
      * Cucumber specific hooks
      */
     beforeFeature: function (feature) {
--- a/packages/wdio-config/src/constants.js
+++ b/packages/wdio-config/src/constants.js
@@ -59,6 +59,7 @@ export const DEFAULT_CONFIGS = {
     after: [],
     onComplete: NOOP,
     onReload: [],
+    onTestFramework: NOOP,
 
     /**
      * cucumber specific hooks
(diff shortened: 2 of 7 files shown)


mgrybyk:
@christian-bromann test framework's `describe` and `it` are not available it before hook.

mgrybyk:
Framework is initialised here https://github.com/webdriverio/webdriverio/blob/79b495a0f64f89ef06d569174b739f9d9f826f9f/packages/wdio-runner/src/index.js#L124

So, before hook is called just before mocha.run.
https://github.com/webdriverio/webdriverio/blob/79b495a0f64f89ef06d569174b739f9d9f826f9f/packages/wdio-mocha-framework/src/index.js#L55

But I need hook that is called after `mocha.run` is called and `it` functions are wrapped with Fibers, but before actual test execution (before `beforeSuite`)


Sam for jasmine

christian-bromann:
In mocha you can use [`mochaOpts.require`](https://github.com/webdriverio/webdriverio/tree/master/packages/wdio-mocha-framework#mochaoptsrequire-stringstring) to exactly do that, no?

mgrybyk:
@christian-bromann it would definitely work if `it` functions are not wrapped with `Fibers`
Flow is like following:
1. `before hook`
2. `mocha run`
3. [`mochaOpts.require`](https://github.com/webdriverio/webdriverio/tree/master/packages/wdio-mocha-framework#mochaoptsrequire-stringstring)
4. `it` functions are wrapped with Fibers
5. I want hook to be called here
6. ...

mgrybyk:
@christian-bromann maybe we can call https://github.com/webdriverio/webdriverio/blob/79b495a0f64f89ef06d569174b739f9d9f826f9f/packages/wdio-mocha-framework/src/index.js#L100
after wrapping `it` functions with Fibers?

christian-bromann:
> after wrapping `it` functions with Fibers?

If this helps with your use case, yes!

mgrybyk:
I'll do some testing around and let you know. Thanks a lot for help

mgrybyk:
@christian-bromann it works fine for me.
I'll update PR later on and re-open or create new one

PR #3654, opened 2019-02-28
webdriver: (W3C) Remove invalid capabilities from alwaysMatch

Varun-garg:
## Proposed changes

In `alwaysMatch` section of W3C capabilities we should only use the capabilities defined [here](https://w3c.github.io/webdriver/#capabilities) and the [extension capabilities](https://w3c.github.io/webdriver/#dfn-extension-capability). This PR will help in running tests on selenium servers where this validation is enforced.

## Types of changes

## Checklist

Signed-off-by: Varun Garg <varun.ga@browserstack.com>

### DIFF
--- a/packages/webdriver/src/constants.js
+++ b/packages/webdriver/src/constants.js
@@ -86,5 +86,13 @@ export const DEFAULTS = {
      */
     headers: {
         type: 'object'
-    }
+    },
+
+    /**
+     * w3c allowed capabilities
+     */
+    w3cAllowedCaps: [
+        'browserName', 'browserVersion', 'platformName', 'acceptInsecureCerts', 'pageLoadStrategy', 'proxy',
+        'setWindowRect', 'timeouts', 'strictFileInteractability', 'unhandledPromptBehavior'
+    ]
 }
--- a/packages/webdriver/src/constants.js
+++ b/packages/webdriver/src/constants.js
@@ -93,5 +93,6 @@ export const DEFAULTS = {
  */
 export const W3C_ALLOWED_CAPABILITIES = [
     'browserName', 'browserVersion', 'platformName', 'acceptInsecureCerts', 'pageLoadStrategy', 'proxy',
-    'setWindowRect', 'timeouts', 'strictFileInteractability', 'unhandledPromptBehavior'
+    'setWindowRect', 'timeouts', 'strictFileInteractability', 'unhandledPromptBehavior', 'keepBrowserName',
+    'mobileMode', 'baseUrl', 'jsonwpMode'
 ]
--- a/packages/webdriver/src/constants.js
+++ b/packages/webdriver/src/constants.js
@@ -86,13 +86,12 @@ export const DEFAULTS = {
      */
     headers: {
         type: 'object'
-    },
-
-    /**
-     * w3c allowed capabilities
-     */
-    w3cAllowedCaps: [
-        'browserName', 'browserVersion', 'platformName', 'acceptInsecureCerts', 'pageLoadStrategy', 'proxy',
-        'setWindowRect', 'timeouts', 'strictFileInteractability', 'unhandledPromptBehavior'
-    ]
+    }
 }
+/**
+ * w3c allowed capabilities
+ */
+export const W3C_ALLOWED_CAPABILITIES = [
+    'browserName', 'browserVersion', 'platformName', 'acceptInsecureCerts', 'pageLoadStrategy', 'proxy',
+    'setWindowRect', 'timeouts', 'strictFileInteractability', 'unhandledPromptBehavior'
+]
--- a/packages/webdriver/src/index.js
+++ b/packages/webdriver/src/index.js
@@ -4,7 +4,7 @@ import { validateConfig } from '@wdio/config'
 
 import webdriverMonad from './monad'
 import WebDriverRequest from './request'
-import { DEFAULTS } from './constants'
+import { DEFAULTS, W3C_ALLOWED_CAPABILITIES } from './constants'
 import { getPrototype, environmentDetector } from './utils'
 
 import WebDriverProtocol from '../protocol/webdriver.json'
@@ -37,12 +37,10 @@ export default class WebDriver {
              */
             : [{ alwaysMatch: params.capabilities, firstMatch: [{}] }, params.capabilities]
 
-        if (w3cCaps.alwaysMatch) {
-            w3cCaps.alwaysMatch = JSON.parse(JSON.stringify(w3cCaps.alwaysMatch))
-            Object.keys(w3cCaps.alwaysMatch)
-                .filter(key => !DEFAULTS.w3cAllowedCaps.includes(key) && !key.match(/^[\w-]+:.*$/))
-                .forEach(key => delete w3cCaps.alwaysMatch[key])
-        }
+        w3cCaps.alwaysMatch = JSON.parse(JSON.stringify(w3cCaps.alwaysMatch))
+        Object.keys(w3cCaps.alwaysMatch)
+            .filter(key => !W3C_ALLOWED_CAPABILITIES.includes(key) && !key.match(/^[\w-]+:.*$/))
+            .forEach(key => delete w3cCaps.alwaysMatch[key])
 
         const sessionRequest = new WebDriverRequest(
             'POST',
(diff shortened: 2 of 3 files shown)


packages/webdriver/src/constants.js
```diff
@@ -86,5 +86,13 @@ export const DEFAULTS = {
      */
     headers: {
         type: 'object'
-    }
+    },
+
+    /**
+     * w3c allowed capabilities
+     */
+    w3cAllowedCaps: [
```
christian-bromann: this property should not be part of the option defaults

packages/webdriver/src/index.js
```diff
@@ -37,6 +37,13 @@ export default class WebDriver {
              */
             : [{ alwaysMatch: params.capabilities, firstMatch: [{}] }, params.capabilities]
 
+        if (w3cCaps.alwaysMatch) {
```
christian-bromann: this is always true, just strip them directly

packages/webdriver/src/index.js
```diff
@@ -37,6 +37,11 @@ export default class WebDriver {
              */
             : [{ alwaysMatch: params.capabilities, firstMatch: [{}] }, params.capabilities]
 
+        w3cCaps.alwaysMatch = JSON.parse(JSON.stringify(w3cCaps.alwaysMatch))
```
christian-bromann: There is no reason to do that

Varun-garg:
Thanks for reviewing @christian-bromann . Not cloning alwaysMatch would also remove invalid caps from `params.capabilities`.  This would break backward compatibility with the old/extended caps/jars.

packages/webdriver/src/index.js
```diff
@@ -37,6 +38,14 @@ export default class WebDriver {
...
```
christian-bromann: Can we have it directly in the object creation? like
```js
/**
 * otherwise assume they passed in jsonwp-style caps (flat object)
 */
{ alwaysMatch: Object.keys(params.capabilities)
    .filter(key => !W3C_ALLOWED_CAPABILITIES.includes(key) && !key.match(/^[\w-]+:.*$/))
    .reduce((obj, key) => {
      obj[key] = params.capabilities[key]
      return obj
    }, {}), firstMatch: [{}] },
  params.capabilities]
```
instead of overwriting it afterwards.

christian-bromann:
ping @Varun-garg

Varun-garg:
Hey @christian-bromann. Sorry I for being late. While your solution looks more elegant, there were 2 issues.

1. Lines were increased after adding warning.
```
: [{
    alwaysMatch: Object.keys(params.capabilities)
        .filter(key => {
            if (W3C_ALLOWED_CAPABILITIES.includes(key) || key.match(/^[\w-]+:.*$/)) {
                return true
            } else {
                log.warn(`Skipping capability ${key} from alwaysMatch.`)
                return false
            }
        })
        .reduce((obj, key) => {
            obj[key] = params.capabilities[key]
            return obj
        }, {}), firstMatch: [{}]
}, params.capabilities]
```

2. We aren't filtering firstMatch if  `params.capabilities.alwaysMatch` exists.

I have rebased and fixed linter errors in test file.

christian-bromann:
> 1. Lines were increased after adding warning.

That is no problem, you can shorten this by doing:
```js
: [{
    alwaysMatch: Object.keys(params.capabilities)
        .filter(key => (W3C_ALLOWED_CAPABILITIES.includes(key) || key.match(/^[\w-]+:.*$/))
        .reduce((obj, key) => {
            obj[key] = params.capabilities[key]
            return obj
        }, {}), firstMatch: [{}]
}, params.capabilities]
```

> 1. We aren't filtering firstMatch if  `params.capabilities.alwaysMatch` exists.

If someone explicitly uses W3C conform capabilities and adds an invalid capability into it, we should not strip this invalid capability and let have the test fail.

> I have rebased and fixed linter errors in test file.

Build is still failing but I am not sure if it is because of this patch or some other reason.

Varun-garg:
> That is no problem, you can shorten this by doing:
Sorry for sounding stupid, I think you skipped the warnings.

> If someone explicitly uses W3C conform capabilities and adds an invalid capability into it, we should not strip this invalid capability and let have the test fail.
Noted this.

> Build is still failing but I am not sure if it is because of this patch or some other reason.
Will investigate this.

christian-bromann:
> Sorry for sounding stupid, I think you skipped the warnings.

Not sure if warnings are needed here

Varun-garg:
To let the users know of capabilities that we are dropping.
 
> I think it would be great to log a warning with capabilities that had to be removed because they are not W3C conform (only if there are any).

Varun-garg:
I have pushed some changes, I doubt if travis errors are caused by this change.

master:
```
Test Suites: 15 failed, 113 passed, 128 total
Tests:       36 failed, 997 passed, 1033 total
Snapshots:   0 total
```

current branch:
```
Test Suites: 15 failed, 113 passed, 128 total
Tests:       36 failed, 999 passed, 1035 total
Snapshots:   0 total
```

packages/webdriver/src/constants.js
```diff
@@ -88,3 +88,11 @@ export const DEFAULTS = {
         type: 'object'
     }
 }
+/**
+ * w3c allowed capabilities
+ */
+export const W3C_ALLOWED_CAPABILITIES = [
```
christian-bromann: there are a bunch of caps here that are not allowed capabilities, see here https://w3c.github.io/webdriver/#capabilities

Varun-garg:
Hey @christian-bromann there seemed to be some problem with my setup. I did need to whitelist some more capabilities not listed [here](https://w3c.github.io/webdriver/#capabilities). We may want to use the [extension capabilities fromat](https://w3c.github.io/webdriver/#dfn-extension-capability) to avoid differences with other drivers.

packages/webdriver/src/constants.js
Varun-garg: yup, just fixed 😄

christian-bromann: I still see invalid caps in there

christian-bromann:
> We may want to use the [extension capabilities fromat](https://w3c.github.io/webdriver/#dfn-extension-capability) to avoid differences with other drivers.

Yeah but this needs to be defined on the users end. If the user provides cap extensions it will not be stripped from the caps.

packages/webdriver/src/constants.js
Varun-garg: only to fix travis tests, they were being used in the existing webdriverio code.

Varun-garg:
> > We may want to use the [extension capabilities fromat](https://w3c.github.io/webdriver/#dfn-extension-capability) to avoid differences with other drivers.
> 
> Yeah but this needs to be defined on the users end. If the user provides cap extensions it will not be stripped from the caps.

Yup, I mean we can whitelist the present extensions (like these 4 caps) but ask users in future to use the extended format for new caps

christian-bromann:
> Yup, I mean we can whitelist the present extensions (like these 4 caps) but ask users in future to use the extended format for new cap

But this would speak against this whole PR which strips them away. Why should we not strip these away?

Varun-garg:
Hey, I also don't want those invalid caps. Remember the travis tests that were failing? I fixed them by finding these capabilities in the test cases and whitelisting them.

christian-bromann:
> Hey, I also don't want those invalid caps. Remember the travis tests that were failing? I fixed them by finding these capabilities in the test cases and whitelisting them.

Well you might need to modify some existing tests to match your changes.

christian-bromann:
ping @Varun-garg

christian-bromann:
@Varun-garg anything I can help with to move this forward?

Varun-garg:
@christian-bromann I will have to look into tests of other modules that are failing to remove those invalid caps. Will pick this back shortly

nextlevelbeard:
Adding some info to this because it might be relevant to the approach used.

Some invalid capabilities **need** to be passed to 3rd party services in order to fully use them and prefixes don't solve this issue, i.e. specifying
```javascript
capabilities: [
  {
    browserName: 'safari',
    'browserstack.local': true
  }
]
```
is needed to run tests in Safari on BrowserStack when the application under test is on the localhost or can only be accessed by the localhost.

maggiesavovska:
Hi! For crossbrowsertesting.com I pass "acceptSslCerts" cap. Can we add that as a valid cap? Thanks!

christian-bromann:
> Can we add that as a valid cap?

It is a valid cap

christian-bromann:
Closing due to inactivity. Please create a new PR if you think this patch is valuable to project. Thanks!

PR #3661, opened 2019-03-01
webdriver: Add strictSSL option

martinfrancois:
## Proposed changes

When executing a test behind a corporate proxy, even with correctly set `HTTP_PROXY` and `HTTPS_PROXY` environment variables, an error like:
```
Error: unable to verify the first certificate
```
or
```
Error: tunneling socket could not be established, cause=write EPROTO 8888:error:1408F10B:SSL routines:ssl3_get_record:wrong version number:openssl\ssl\record\ssl3_record.c:252:
```
can appear, if the corporate proxy is also doing SSL interception and inspection due to the certificate not being recognized as valid in https connections.

The [request node package](https://www.npmjs.com/package/request) has an option `strictSSL` which can be set to `false`, leading to SSL certificates not being validated in https connections and thus solving the issue.

The proposed change is to set the `strictSSL` option depending on the value of an environment variable `STRICT_SSL`. If it is set to `false`, `strictSSL` will also be defined as `false` and in all other cases, `strictSSL` will default to `true`.

## Types of changes

## Checklist

## Further comments

The functionality of this is similar to what exists for npm already: https://docs.npmjs.com/misc/config#strict-ssl
I considered passing the flag in via parameters, but wanted to keep the definition of the flag consistent with the other already existing proxy-related settings `HTTP_PROXY` and `HTTPS_PROXY`, as described in the proxy documentation.

### DIFF
--- a/docs/Proxy.md
+++ b/docs/Proxy.md
@@ -18,6 +18,15 @@ $ export HTTPS_PROXY=https://my.corp.proxy.com:9090
 $ wdio wdio.conf.js
 ```
 
+Additionally, if you run into errors concerning SSL certificates during test execution, you can set the `STRICT_SSL` environment variable to `false`, which will turn off SSL key validation when making requests with https:
+
+```sh
+$ export HTTP_PROXY=http://my.corp.proxy.com:9090
+$ export HTTPS_PROXY=https://my.corp.proxy.com:9090
+$ export STRICT_SSL=false
+$ wdio wdio.conf.js
+```
+
 If you use [Sauce Connect Proxy](https://wiki.saucelabs.com/display/DOCS/Sauce+Connect+Proxy) start it via:
 
 ```sh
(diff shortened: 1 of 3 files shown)


christian-bromann:
@martinfrancois please sign the CLA so this can be merged

martinfrancois:
Thanks, I just signed it!

christian-bromann:
it seems that the linter is not happy

martinfrancois:
just noticed it too - let's hope it's happy now 😉

martinfrancois:
Mhm.. the tests failing is a bit strange, they pass on my machine?

christian-bromann:
@martinfrancois make sure you run v8 as we use this in Travis too

martinfrancois:
Good point! Indeed, using node v8 it fails the same... I think the easiest would be to mock `process.env` here, but I didn't succeed... Any ideas?

martinfrancois:
Turned out there was an issue with the previous logic, which caused it to only work on Windows (since it treats environment variables case insensitive). I corrected the logic now to work on both platforms equally.

PR #3672, opened 2019-03-04
Issue 3416: pass multiple errors through to test stats for reporting if available

nicholasbailey:
## Proposed changes
Continuing the work started in [https://github.com/webdriverio/webdriverio/pull/3669](https://github.com/webdriverio/webdriverio/pull/3669). That PR allowed Jasmine to pass along more than one error on test run events. Here, we're handling more than one error in the base reporter, and assigning them to test stats for consumption by reporters. In packages/wdio-reporter/src/index.js we pass multiple errors if available to the test stats, and otherwise wrap the 'error' property of the event. Then in, the test/hook stats, we assign both a 'errors' property and an 'error' property (just the first element of the passed errors array). This should maintain backwards compatibility with any existing reporters/frameworks, while allowing reporters to optionally report multiple failures from a single tests.

There's also a lot of added test coverage in this PR, since I wanted to be very careful not to break anything. While we executed all of the event handlers in packages/wdio-reporter/src/index.js, a lot of the event handling behavior wasn't asserted against. Open to reducing the volume of tests if anyone thinks we're over-testing there.

## Types of changes

## Checklist

### DIFF
--- a/packages/wdio-reporter/src/index.js
+++ b/packages/wdio-reporter/src/index.js
@@ -3,6 +3,8 @@ import fse from 'fs-extra'
 import { format } from 'util'
 import EventEmitter from 'events'
 
+import { getErrorsFromEvent } from './utils'
+
 import SuiteStats from './stats/suite'
 import HookStats from './stats/hook'
 import TestStats from './stats/test'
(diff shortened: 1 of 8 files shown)


packages/wdio-reporter/src/index.js
```diff
@@ -11,6 +11,12 @@ import RunnerStats from './stats/runner'
 
 import { MOCHA_TIMEOUT_MESSAGE, MOCHA_TIMEOUT_MESSAGE_REPLACEMENT } from './constants'
 
+function getErrorsFromMessage(message) {
```
christian-bromann: Let's move this into `utils.js`

nicholasbailey: Sounds good

nicholasbailey: @christian-bromann fixed!

PR #3676, opened 2019-03-04
wdio-allure-reporter: start end step

mgrybyk:
## Proposed changes

- it should be possible to create stepFunc like it is described in allure doc https://docs.qameta.io/allure/#_steps_6
- it should be possible to add step with duration

To be able to do it we need to have start and end step events.

```
allure.startStep('long running wait for foo')
$('foo').waitForExist()
allure.endStep('passed')
```

## Types of changes

## Checklist

## Further comments

Another usage example:
```
const step = (step, func, attachment) => {
  allure.startStep(step)

  if (attachment) {
    allure.addAttachment(attachment)
  }

  let status
  try {
    func()
  } catch (error) {
    allure.addAttachment(error)
    if (error && error.name === 'AssertionError') {
      status = 'failed'
    } else {
      status = 'broken'
    }
  } finally {
    allure.endStep(status) // undefined => passed
  }
}

it('my test', () => {
  const el = $('foo')
  step('wait for foo', () => el.waitForExist())
  step('foo has modal class', () => assert.equals(el.getAttribute('class'), 'modal'))
})
```

### DIFF
--- a/packages/wdio-allure-reporter/src/index.js
+++ b/packages/wdio-allure-reporter/src/index.js
@@ -150,7 +150,7 @@ class AllureReporter extends WDIOReporter {
             }
             return
         }
-        
+
         // add hook as test to suite
         this.onTestStart(hook)
     }
@@ -259,16 +259,29 @@ class AllureReporter extends WDIOReporter {
         }
     }
 
+    startStep(title) {
+        if (!this.isAnyTestRunning()) {
+            return false
+        }
+        this.allure.startStep(title)
+    }
+
+    endStep(status) {
+        if (!this.isAnyTestRunning()) {
+            return false
+        }
+        this.allure.endStep(status)
+    }
+
     addStep({step}) {
         if (!this.isAnyTestRunning()) {
             return false
         }
-
-        this.allure.startStep(step.title)
+        this.startStep(step.title)
         if (step.attachment) {
             this.addAttachment(step.attachment)
         }
-        this.allure.endStep(step.status)
+        this.endStep(step.status)
     }
 
     addArgument({name, value}) {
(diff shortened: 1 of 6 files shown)


mgrybyk:
typings have to be generated before publishing this package

christian-bromann:
@BorisOsipov can you take a quick look on this?

mgrybyk:
@abjerstedt may it be useful for cucumber framework?

BorisOsipov:
>@abjerstedt may it be useful for cucumber framework as well?
@mgrybyk for cucumber we had additional report logic. See https://github.com/webdriverio-boneyard/wdio-allure-reporter/blob/master/lib/reporter.js#L28
it reports cucumber steps as allure steps(not tests). I din't port it to v5 because there is no wdio-cucumber for v5

mgrybyk:
@BorisOsipov yeah, I see. That's what I was trying to say. There is ongoing work on cucumber framework and it might be useful there.
Thanks

packages/wdio-allure-reporter/src/index.js
```diff
@@ -259,16 +259,29 @@ class AllureReporter extends WDIOReporter {
...
-        this.allure.startStep(step.title)
+    endStep(status) {
+        if (!this.isAnyTestRunning()) {
+            return false
+        }
+        this.allure.endStep(status)
+    }
+
+    addStep({step}) {
+        if (!this.isAnyTestRunning()) {
```
CrispusDH: Is it necessary here? Because next first function `this.startStep` has the same inside

mgrybyk:
Oops, sorry for that, have no idea how have I missed it

packages/wdio-allure-reporter/src/index.js
mgrybyk: To be honest I'm not really sure. I didn't change behaviour of existing steps, only added  startStep and endStep.

If I simply remove this verification  corresponding test will fail because it expects `false` as a return value while we are returning `undefined`.

mgrybyk: If you are sure that returning undefined instead of false in this case is fine, I can do the change.

mgrybyk:
@christian-bromann @CrispusDH updated

Issue #3537, opened 2019-02-10
startRecordingScreen and stopRecordingScreen don't work as expected


Gilad-WT:
**Environment (please complete the following information):**
 - **WebdriverIO version:** 5.2.8
 - **Mode:** [Standalone mode or WDIO Testrunner] Standalone 
 - **Node.js version:** 10.14.2
 - **NPM version:** 6.4.1
 - **Platform name and version:** Windows 10

**Config of WebdriverIO**
wdio.remote({
        hostname : opts.hostIp,
        port : opts.hostPort,
        logLevel : "silent",
        capabilities : {
            platformName: "Android",
            platformVersion: "8.1",
            deviceName: "Android Emulator",
            app: opts.app,
            automationName: "UiAutomator2",
            newCommandTimeout : 600
        }
    });

**Describe the bug**
startRecordingScreen and stopRecordingScreen don't work as expected

**To Reproduce**
```
async function test(){
let driver = await wdio.remote({
        hostname : opts.hostIp,
        port : opts.hostPort,
        logLevel : "silent",
        capabilities : {
            platformName: "Android",
            platformVersion: "8.1",
            deviceName: "Android Emulator",
            app: opts.app,
            automationName: "UiAutomator2",
            newCommandTimeout : 600
        }
    });
await driver.startRecordingScreen('./video.mp4', 20000000 *181, 1, 20000000)
await new Promise((resolve, reject) => {
            setTimeout(() => {
                resolve();
            }, 10000);
        });
await driver.stopRecordingScreen()
}
test()
```
The above code should according to the documentation here (https://webdriver.io/docs/api/appium.html) create a mp4 file named video.mp4 in the current dir but it does not 

Instead I have to use the following code to generate the file

```
async function test(){
let driver = await wdio.remote({
        hostname : opts.hostIp,
        port : opts.hostPort,
        logLevel : "silent",
        capabilities : {
            platformName: "Android",
            platformVersion: "8.1",
            deviceName: "Android Emulator",
            app: opts.app,
            automationName: "UiAutomator2",
            newCommandTimeout : 600
        }
    });
await driver.startRecordingScreen('./video.mp4', 20000000 *181, 1, 20000000)
await new Promise((resolve, reject) => {
            setTimeout(() => {
                resolve();
            }, 10000);
        });
let recordingAsBase64Str = await driver.stopRecordingScreen()
let videoBinary = Buffer.from(recordingAsBase64Str, 'base64')
[...]

PR #3677, opened 2019-03-05
Appium saveRecordingScreen

mgrybyk:
## Proposed changes

fix for #3537

- fixed Appium start / stop screen recording protocol commands 
- added saveRecordingScreen command similar to saveScreenshot
- moved getAbsoluteFilepath and assertDirectoryExists to utils to reuse them later on in all 3 files. Updated save screenshot tests accordingly

## Types of changes

## Checklist

## Further comments

typings should be generated for webdriver and webdriverio packages

### DIFF
--- a/packages/webdriverio/tests/commands/browser/saveRecordingScreen.test.js
+++ b/packages/webdriverio/tests/commands/browser/saveRecordingScreen.test.js
@@ -1,9 +1,8 @@
-
 import request from 'request'
 import { remote } from '../../../src'
 import * as utils from '../../../src/utils'
 
-describe('saveScreenshot', () => {
+describe('saveRecordingScreen', () => {
     jest.mock('fs')
     const fs = require('fs').default
     let browser, getAbsoluteFilepathSpy, assertDirectoryExistsSpy, writeFileSyncSpy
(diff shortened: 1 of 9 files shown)


packages/webdriverio/tests/commands/element/saveScreenshot.test.js
```diff
@@ -37,21 +55,4 @@ describe('saveScreenshot', () => {
             elem.saveScreenshot('./file.txt')
         ).rejects.toEqual(expectedError)
     })
-
-    it('should fail if not existing directory', async () => {
```
christian-bromann: I think this test is still valueable because it ensure that no one accidentally remove the file path assertion.

packages/webdriverio/tests/commands/browser/saveRecordingScreen.test.js
```diff
@@ -0,0 +1,50 @@
...
+        expect(video.toString()).toBe('some screenshot')
+    })
+
+    it('should fail if no filename provided', async () => {
+        const expectedError = new Error('saveRecordingScreen expects a filepath')
+
+        await expect(
+            browser.saveRecordingScreen()
+        ).rejects.toEqual(expectedError)
+    })
```
christian-bromann: Let's also have a check here that the existence of the filepath is being checked

packages/webdriverio/tests/commands/element/saveScreenshot.test.js
mgrybyk: @christian-bromann there is a test for it already but in more simple way.
See lines 31-34 https://github.com/webdriverio/webdriverio/pull/3677/files#diff-a2fdb7dc0f85292704af730512ae75c7R33

we have verification that assertion is called with proper parameters

christian-bromann: ahh, yeah this is fine then 👍

packages/webdriverio/tests/commands/browser/saveRecordingScreen.test.js
mgrybyk: adding it at the moment to save screenshots and save recording. Nice catch, thanks!

mgrybyk: @christian-bromann done

PR #3680, opened 2019-03-05
webdriverio: fix control key w3c compliancy

LukoyanovE:
## Proposed changes

Cannot send [w3c described](https://w3c.github.io/webdriver/#keyboard-actions) `Control` key.
Somewhy it's called `'Control Left'` in WDIO.

Also added `";", "=", "0", "1", "2", "3", "4", "5", "6", "7", "8", "9", "*", "+", ",", "-", ".", "/",` key codes.

There are some additional "right" keys in w3c spec, but I don't know how to add it to WDIO:
```
\uE050	"Shift"
\uE051	"Control"
\uE052	"Alt"
\uE053	"Meta"
\uE054	"PageUp"
\uE055	"PageDown"
\uE056	"End"
\uE057	"Home"
\uE058	"ArrowLeft"
\uE059	"ArrowUp"
\uE05A	"ArrowRight"
\uE05B	"ArrowDown"
\uE05C	"Insert"
\uE05D	"Delete"
```

## Types of changes

## Checklist

## Further comments


packages/webdriverio/src/constants.js
```diff
@@ -327,23 +328,41 @@ export const UNICODE_CHARACTERS = {
     'ArrowDown': '\uE015',
     'Insert': '\uE016',
     'Delete': '\uE017',
+    ';': '\uE018',
```
christian-bromann: Why do we need to define these twice?

LukoyanovE: [there](https://webdriver.io/docs/api/browser/keys.html) is doc says:

> You’ll find all supported characters [here](https://w3c.github.io/webdriver/webdriver-spec.html#keyboard-actions). 

And there are these codes: `";", "=", "0", "1", "2", "3", "4", "5", "6", "7", "8", "9", "*", "+", ",", "-", ".", "/",`.

I'm looking into `grapheme-splitter` and it it just split all and send such symbols as is.

Ok maybe, I should delete such symbols from PR.
I'll do, if you say.

christian-bromann: yes but these are all defined twice .. if you want to use these special characters you e.g. use `Semicolon` or just send `;` as a string.

LukoyanovE: Maybe doc should be rewritten to other key list, instead of w3c one?

CrispusDH:
Might be better to put at least one test for check random control key?

packages/webdriverio/src/constants.js
christian-bromann: Yes, we should link to the constant table. I honestly don't know what the difference between sending `Numpad 0` and `0`

LukoyanovE: `Numpad<N>` should be a keycode for separate Numpad part of keyboard

W3C spec includes it too
https://w3c.github.io/webdriver/#keyboard-actions

It should return the same code as `<N>` key, but when shifted, the same as `Insert`

LukoyanovE:
@CrispusDH , random control key?
it's question about all keys, or some special, like `Control`, `Right Control`, `Alt`, and others?

CrispusDH:
As I understand, currently, we have zero tests for that. So, at least one that will check each time random item from `UNICODE_CHARACTERS` it would be great :). @christian-bromann what do you think?

Issue #3700, opened 2019-03-08
isDisplayed doesn't work properly because of refetch


mgrybyk:
**Environment (please complete the following information):**
 - **WebdriverIO version:** 5.7.1

**Describe the bug**
isDisplayed fails once element is no longer exists

**To Reproduce**
```
// assuming element exists
const el = $('foo')
el.isDisplayed() // true/false
// remove element
el.isDisplayed() 
```

[Include code or an example repository that can easily be set up]

**Expected behavior**
el.isDisplayed() should return false immediately if element doesn't exist without refetching it.

**Log**
```
Error: Can't call isElementDisplayed on element with selector "input[type='submit']" because element wasn't found
     at implicitWait (/e2e/node_modules/webdriverio/build/utils/implicitWait.js:34:13)
     at Element.isDisplayed (/e2e/node_modules/webdriverio/build/commands/element/isDisplayed.js:89:19)
```

mgrybyk:
had to change regex to match isElementDisplayed as well
https://github.com/webdriverio/webdriverio/blob/master/packages/webdriverio/src/utils/implicitWait.js#L14
`if (!currentElement.elementId && !commandName.match(/(waitUntil|waitFor|isExisting|is?\w+Displayed)/)) {`
changed `isDisplayed` to `is?\w+Displayed`

Issue #3645, opened 2019-02-28
webdriverio: refetch doesn't work for shadow DOM elements in Safari


jrobinson01:
[//]: # NOTE: This repository only maintains packages that are listed in the [Readme](https://github.com/webdriverio/webdriverio/blob/master/README.md#packages). Please make sure that your issue is directly caused by one of these packages and if not file an issue in the correct 3rd party package repository.

**Environment (please complete the following information):**
 - **WebdriverIO version:** 5.6.2
 - **Mode:** testrunner
 - **If WDIO Testrunner, running sync/async:** sync
 - **Node.js version:** 8.9.1
 - **NPM version:** 5.5.1
 - **Browser name and version:** Safari 12.0.3
 - **Platform name and version:** OSX 10.13
 - **Additional wdio packages used (if applicable):** [e.g. @wdio/spec reporter, @wdio/selenium-standalone service]

**Config of WebdriverIO**
An example of how you configured WebdriverIO config

**Describe the bug**
After deleting an element and creating a new one, the refetchElement logic fails to find the new element.

**To Reproduce**
An example test in jasmine:
```javascript
it('shadow$ works with refetch', () => {
  const blogApp = browser.$('blog-app');
  const header = blogApp.shadow$('header');
  expect(header.$('a-link').getAttribute('href')).toEqual('/');
  // remove the main element and recreate
  browser.execute(() => {
    const oldEl = document.querySelector('blog-app');
    oldEl.parentElement.removeChild(oldEl);// removes the main element
    const newEl = document.createElement('blog-app');// create a new one, which contains header
    document.body.appendChild(newEl);
  });
  // header should be a stale reference at this point and invoke refetch
  expect(header.$('a-link').getAttribute('href')).toEqual('/');
});
```

**Expected behavior**
Safari webdriver should return a 'stale element reference' error response which wdio uses to invoke the refetch logic.

**Log**
https://gist.github.com/jrobinson01/7093fa9e53009c8bf20d6a0a30f5788a

**Additional context**

> Please use the correct markdown

abjerstedt:
@jrobinson01 Can you do me a favor get me some data on this?  
Set your logLevel to 'info' and paste the response the driver gives you for the stale element into this issue. I'll update the matcher to handle this case.

jrobinson01:
[...]

PR #3701, opened 2019-03-08
webdriverio: fix isDisplayed

### LINKED ISSUE #3699: blocker: isDisplayed doesn't work in Appium in 5.7.1
 - **WebdriverIO version:** 5.7.1

**Config of WebdriverIO**
Using Appium 1.11.1

```
capabilities: [{
...
    "appium:app": __dirname + '/../../apps/MoPlay.apk',
    "appium:appPackage": 'com.addisongloballtd.moplay.adhoc',
    "appium:appActivity": 'com.moplay.MainActivity',
}]
```

**Describe the bug**
isDisplayed doesn't work, works fine in 5.7.0

**To Reproduce**
`$('~BUTTON').waitForDisplayed()`

**Expected behavior**
isDisplayed, waitForDisplayed should work

**Log**
```
Error: waitUntil condition failed with the following reason: Cannot read property 'toLowerCase' of undefined
    at timer.catch.e (/e2e/node_modules/webdriverio/build/commands/browser/waitUntil.js:75:11)
    at Element.waitForDisplayed (/e2e/node_modules/webdriverio/build/commands/element/waitForDisplayed.js:51:15)
```

**Additional context**

mgrybyk:
## Proposed changes

fix #3699 #3700 #3645 #3709 

## Types of changes

## Checklist

## Further comments

code for testing PR
```
browser.url("http://adam.goucher.ca/parkcalc/")
const inputSelector = "input[type='submit']"

// already not existing
$("not-existing").waitForDisplayed(2000, true)
const el = $(inputSelector)
console.log(el.isDisplayed()) // true

// remove element in 3 seconds
browser.execute(function (selector) {
    const window = this
    window.setTimeout(() => {
        window.document.querySelector(selector).remove()
    }, 3000);
}, inputSelector)

// first existing, then not existing
$(inputSelector).waitForDisplayed(8000, true)
console.log(el.isDisplayed()) // false
```

### DIFF
--- a/packages/webdriverio/src/commands/element/isDisplayed.js
+++ b/packages/webdriverio/src/commands/element/isDisplayed.js
@@ -70,7 +70,7 @@ export default async function isDisplayed() {
      * no longer dictates it. In those instances, we pass the element through a script
      * that was provided by Brian Burg of safaridriver.
      */
-    return browser.isW3C && noW3CEndpoint.includes(browser.capabilities.browserName.toLowerCase()) ?
+    return browser.isW3C && !browser.isMobile && noW3CEndpoint.includes(browser.capabilities.browserName.toLowerCase()) ?
         await browser.execute(isElementDisplayedScript, {
             [ELEMENT_KEY]: this.elementId, // w3c compatible
             ELEMENT: this.elementId // jsonwp compatible
--- a/packages/webdriverio/src/middlewares.js
+++ b/packages/webdriverio/src/middlewares.js
@@ -15,7 +15,19 @@ export const elementErrorHandler = (fn) => (commandName, commandFn) => {
             this.elementId = element.elementId
 
             try {
-                return await fn(commandName, commandFn).apply(this, args)
+                const result = await fn(commandName, commandFn).apply(this, args)
+
+                /**
+                 * assume Safari responses like { error: 'no such element', message: '', stacktrace: '' }
+                 * as `stale element reference`
+                 */
+                if (result && result.error === 'no such element') {
+                    const err = new Error()
+                    err.name = 'stale element reference'
+                    throw err
+                }
+
+                return result
             } catch (error) {
                 if (error.name === 'stale element reference') {
                     const element = await refetchElement(this, commandName)
--- a/packages/webdriverio/src/utils/implicitWait.js
+++ b/packages/webdriverio/src/utils/implicitWait.js
@@ -11,7 +11,7 @@ const log = logger('webdriverio')
  */
 export default async function implicitWait (currentElement, commandName) {
 
-    if (!currentElement.elementId && !commandName.match(/(waitUntil|waitFor|isExisting|isDisplayed)/)) {
+    if (!currentElement.elementId && !commandName.match(/(waitUntil|waitFor|isExisting|is?\w+Displayed)/)) {
         log.debug(
             `command ${commandName} was called on an element ("${currentElement.selector}") ` +
             'that wasn\'t found, waiting for it...'
--- a/packages/webdriverio/tests/__mocks__/request.js
+++ b/packages/webdriverio/tests/__mocks__/request.js
@@ -50,7 +50,7 @@ const requestMock = jest.fn().mockImplementation((params, cb) => {
     case '/wd/hub/session':
         value = sessionResponse
 
-        if (params.body.capabilities.alwaysMatch.browserName.includes('noW3C')) {
+        if (params.body.capabilities.alwaysMatch.browserName && params.body.capabilities.alwaysMatch.browserName.includes('noW3C')) {
             value.desiredCapabilities = { browserName: 'mockBrowser' }
             delete value.capabilities
         }
@@ -239,8 +239,13 @@ const requestMock = jest.fn().mockImplementation((params, cb) => {
     /**
      * overwrite if manual response is set
      */
+    let statusCode = 200
     if (Array.isArray(manualMockResponse)) {
         value = manualMockResponse.shift() || value
+
+        if (typeof value.statusCode === 'number') {
+            statusCode = value.statusCode
+        }
 
         if (manualMockResponse.length === 0) {
             manualMockResponse = null
@@ -257,7 +262,7 @@ const requestMock = jest.fn().mockImplementation((params, cb) => {
 
     cb(null, {
         headers: { foo: 'bar' },
-        statusCode: 200,
+        statusCode,
         body: response
     }, response)
 })
--- a/packages/webdriverio/tests/commands/element/isDisplayed.test.js
+++ b/packages/webdriverio/tests/commands/element/isDisplayed.test.js
@@ -26,12 +26,45 @@ describe('isDisplayed test', () => {
         expect(request.mock.calls[0][0].uri.path).toBe('/wd/hub/session/foobar-123/element/some-elem-123/displayed')
     })
 
+    it('should allow to check if element is displayed in mobile mode without browserName', async () => {
+        browser = await remote({
+            baseUrl: 'http://foobar.com',
+            capabilities: {
+                keepBrowserName: true,
+                mobileMode: true
+            }
+        })
+        elem = await browser.$('#foo')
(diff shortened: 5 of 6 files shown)


packages/webdriverio/src/commands/element/isDisplayed.js
```diff
@@ -70,7 +70,7 @@ export default async function isDisplayed() {
      * no longer dictates it. In those instances, we pass the element through a script
      * that was provided by Brian Burg of safaridriver.
      */
-    return browser.isW3C && noW3CEndpoint.includes(browser.capabilities.browserName.toLowerCase()) ?
+    return browser.isW3C && !browser.isMobile && noW3CEndpoint.includes(browser.capabilities.browserName.toLowerCase()) ?
```
mgrybyk: fix #3699

packages/webdriverio/src/utils/implicitWait.js
```diff
@@ -11,7 +11,7 @@ const log = logger('webdriverio')
  */
 export default async function implicitWait (currentElement, commandName) {
 
-    if (!currentElement.elementId && !commandName.match(/(waitUntil|waitFor|isExisting|isDisplayed)/)) {
+    if (!currentElement.elementId && !commandName.match(/(waitUntil|waitFor|isExisting|is?\w+Displayed)/)) {
```
mgrybyk: fix #3700

abjerstedt:
Can iphone 12 safari still work with this? I think they still use safari driver and lose the endpoint.

mgrybyk:
@abjerstedt I'll test this really soon.

Are there any other cases to verify?

wswebcreation:
@mgrybyk 
Yes, iOS Hybrid app. I can verify the mobile parts if needed now. Give me 30 minutes

mgrybyk:
@wswebcreation I'll be verifying all the cases really soon as well.
Doing mobile testing on emulators, but I think it shouldn't be a problem in this case.
Missing IE and Edge.

wswebcreation:
Ok, then I'll proceed working for my boss 😜

abjerstedt:
@mgrybyk IE is unaffected by either the last or this PR. Edge should be fine... i don't think anyone actually tests mobile edge anyways.

mgrybyk:
@abjerstedt @wswebcreation 

- Android app **PASSED**
- Android browser chrome **PASSED**
- iOS app
- iOS browser safari
- Chrome **PASSED**
- Firefox **PASSED**
- Safari _FAILED_ !!!
test is not failing, however instead of `false` I'm receiving
`{ error: 'no such element', message: '', stacktrace: '' }`

abjerstedt:
@mgrybk yeah... theres already an issue with safari and refetch (from before my recent PRs).

I think safari is giving us a different status or response code on stales preventing us from matching and handling them.

mgrybyk:
I'm using following code for testing everything except of native app:
```
        browser.url("http://adam.goucher.ca/parkcalc/")
        const el = $("input[type='submit']")
        console.log(el.isDisplayed()) // should return true
        browser.execute(function () {
            this.document.querySelector("input[type='submit']").remove()
        })
        console.log(el.isDisplayed()) // should return false
```

mgrybyk:
@abjerstedt should I fix it as well within this PR?

abjerstedt:
Do we also want to add execute to this exclusion list for waiting?

abjerstedt:
@mgrybyk might as well. They are all slightly related.

mgrybyk:
@abjerstedt
> Do we also want to add execute to this exclusion list for waiting?

Not sure, from the first point of view it makes sense, but I might be wrong :)

mgrybyk:
@wswebcreation can you help with testing native app?
I'm having some issues with certificate and unable to verify it right now

mgrybyk:
@abjerstedt fixed Safari issue `{ error: 'no such element', message: '', stacktrace: '' }`

packages/webdriverio/src/commands/element/isDisplayed.js
```diff
@@ -70,10 +70,11 @@ export default async function isDisplayed() {
...
      * that was provided by Brian Burg of safaridriver.
      */
-    return browser.isW3C && noW3CEndpoint.includes(browser.capabilities.browserName.toLowerCase()) ?
+    const result = browser.isW3C && !browser.isMobile && noW3CEndpoint.includes(browser.capabilities.browserName.toLowerCase()) ?
         await browser.execute(isElementDisplayedScript, {
             [ELEMENT_KEY]: this.elementId, // w3c compatible
             ELEMENT: this.elementId // jsonwp compatible
         }) :
         await this.isElementDisplayed(this.elementId)
+    return result === true
```
mgrybyk: @abjerstedt fix of Safari's `{ error: 'no such element', message: '', stacktrace: '' }`

mgrybyk:
- Android app **PASSED**
- Android browser chrome **PASSED**
- iOS app  _unable to verify_
- iOS browser safari **PASSED**
- Chrome **PASSED**
- Firefox **PASSED**
- Safari **PASSED**

wswebcreation:
@mgrybyk 

I can do that in about 30-60 minutes, currently busy with some stuff for my boss

packages/webdriverio/src/commands/element/isDisplayed.js
mgrybyk: ignore this "fix".

abjerstedt: This won't fix the other commands though. We will want to update the response parser in webdriver.

mgrybyk:
Thanks a lot, @wswebcreation  

P.S. please don't merge this PR, have to do some more minor fixes and add tests

packages/webdriverio/src/middlewares.js
```diff
@@ -15,7 +15,19 @@ export const elementErrorHandler = (fn) => (commandName, commandFn) => {
...
 
             try {
-                return await fn(commandName, commandFn).apply(this, args)
+                const result = await fn(commandName, commandFn).apply(this, args)
+
+                /**
+                 * assume Safari responses like { error: 'no such element', message: '', stacktrace: '' }
+                 * as `stale element reference`
+                 */
+                if (result && result.error === 'no such element') {
```
mgrybyk: fix #3645

abjerstedt: Can we handle this inside webdriver instead?
packages/webdriver/src/utils.js
isSuccessfulResponse (statusCode, body) {

abjerstedt: ultimately, it would be nice if we could consume this inCustomRequestError and treat safari's stale elements the same as the rest

mgrybyk: @abjerstedt sorry to say there is no way to do it.
We have to keep this dirty hack.
The problem is that there is no difference between Safari's not found element and Safari's stale element exception.

My proposal would be to leave this code as is.

christian-bromann:
@mgrybyk can you remove the `DEL_ME` file? Feel free to add it to gitignore if you use it for dev purposes

mgrybyk:
@christian-bromann sorry for that. I'll delete it later today

mgrybyk:
@christian-bromann deleted.

christian-bromann:
@abjerstedt can you do another review?

packages/webdriverio/src/utils/implicitWait.js
mgrybyk: also fixes #3709

packages/webdriverio/src/middlewares.js
abjerstedt: @mgrybyk We still need to handle the safari stales upstream in webdriver (but i can put a diff pr up). 

We need to block the triple requests and just return an element not found. This should hypothetically cause the middleware to just work.

mgrybyk: Agree, I've noticed that requests are multiplied.
If you can assist it would be really awesome!

PR #3706, opened 2019-03-10
Enforce more eslint rules

christian-bromann:
## Proposed changes

We had some PRs recently where we were wondering why we didn't enforce certain code styles a bit more. I couldn't find the PR anymore but I went through the common eslint style definitions and enforced some rules that were used across the codebase already.

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,27 +1,41 @@
 module.exports = {
   extends: [
-    "eslint:recommended",
-    "plugin:import/errors",
-    "plugin:import/warnings"
+    'eslint:recommended',
+    'plugin:import/errors',
+    'plugin:import/warnings'
   ],
   env: {
     node: true,
     es6: true
   },
-  parser: "babel-eslint",
+  parser: 'babel-eslint',
   parserOptions: {
     ecmaVersion: 2016,
     sourceType: 'module'
   },
   rules: {
-    semi: ["error", "never"],
-    quotes: ["error", "single"],
+    semi: ['error', 'never'],
+    quotes: ['error', 'single'],
     indent: [2, 4],
 
-    "import/no-unresolved": [2, {commonjs: true, amd: true}],
-    "import/named": 2,
-    "import/namespace": 2,
-    "import/default": 2,
-    "import/export": 2
+    'import/no-unresolved': [2, {commonjs: true, amd: true}],
+    'import/named': 2,
+    'import/namespace': 2,
+    'import/default': 2,
+    'import/export': 2,
+
+    'no-multiple-empty-lines': [2, {'max': 2, 'maxEOF': 1}],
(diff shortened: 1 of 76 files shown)


.eslintrc.js
```diff
@@ -1,27 +1,41 @@
...
-    "import/namespace": 2,
-    "import/default": 2,
-    "import/export": 2
+    'import/no-unresolved': [2, {commonjs: true, amd: true}],
+    'import/named': 2,
+    'import/namespace': 2,
+    'import/default': 2,
+    'import/export': 2,
+
+    'no-multiple-empty-lines': [2, {'max': 2, 'maxEOF': 1}],
```
mgrybyk: @christian-bromann however this allows two lines
```
const waitForExist = require('../../src/commands/element/waitForExist')


describe('refetchElement', () => {
```
will be ignored

.eslintrc.js
```diff
@@ -1,27 +1,41 @@
 module.exports = {
```
mgrybyk: we have lines like `{ allowSingleLine: true }` and `{commonjs: true, amd: true}`. Can we use same style space-in-brackets style?

christian-bromann: I will set `space-in-brackets` to always because it has less errors (550 vs 2400)

.eslintrc.js
```diff
@@ -1,27 +1,42 @@
...
+    semi: ['error', 'never'],
+    quotes: ['error', 'single'],
     indent: [2, 4],
 
-    "import/no-unresolved": [2, {commonjs: true, amd: true}],
-    "import/named": 2,
-    "import/namespace": 2,
-    "import/default": 2,
-    "import/export": 2
+    'import/no-unresolved': [2, {commonjs: true, amd: true}],
```
mgrybyk: Minor styling issue, can you reformat this file?

.eslintrc.js
```diff
@@ -1,27 +1,42 @@
...
+    'import/named': 2,
+    'import/namespace': 2,
+    'import/default': 2,
+    'import/export': 2,
+
+    'no-multiple-empty-lines': [2, {'max': 1, 'maxEOF': 1}],
```
mgrybyk: Maybe I missed your answer,  what do you think on changing lines to 1?

.eslintrc.js
christian-bromann: reformat into what?

.eslintrc.js
christian-bromann: Than we would only get warnings if someone breaks that rule.

.eslintrc.js
mgrybyk: The file itself. There are issues with spaces in curly bracers

christian-bromann: mhm for some reason this wasn't caught

Issue #3693, opened 2019-03-07
Port wdio-appium-service


abjerstedt:
**Describe the solution you'd like**

In order to better support our native app customers, we should absorb wdio-appium-service into the monorepo. This would allow these users to get started on webdriverio with less overhead.

wswebcreation:
If we are going to do that, can we also make it a parallel solution, meaning that each Appium instance will get it's own port if multiple capabilities are provided?

saikrishna321:
@wswebcreation  

>meaning that each Appium instance will get it's own port if multiple capabilities are provided

We wouldn't need that as appium server can handle multiple capabilities request in single session

wswebcreation:
@saikrishna321 
Tnx, didn't knew that

christian-bromann:
We merged that service, great work by @MortenGregersen

PR #3717, opened 2019-03-12
Appium service

MortenGregersen:
## Proposed changes

The [wdio-appium-service](http://github.com/rhysd/wdio-appium-service) doesn't work with wdio v5. It was decided that instead of fixing the community package, it should be absorbed into the wdio monorepo (#3693).

## Types of changes

## Checklist

## Further comments

When creating this service I took a lot of inspiration from @wdio/selenium-standalone-service and [wdio-appium-service](http://github.com/rhysd/wdio-appium-service).

Most of the functionality is like [wdio-appium-service](http://github.com/rhysd/wdio-appium-service) but adapted for @wdio fx using the @wdio/logger. I also changed some naming and added some comments.

The `getFilePath.js` is a copy from @wdio/selenium-standalone-service. This should maybe be moved to @wdio/utils or something. I've copied the tests for it.

I'm not sure how much we should test in this service. I've created tests for all helper methods and also a test for checking that the Appium service was started. I couldn't get it to succeed all the time. I hope someone can help with this. I can see that the service is started by making a HTTP GET request to `http://127.0.0.1:4723/wd/hub/status`, but in the test I always get a connection refused.
I haven't created a test for the `exitCallback` as I couldn't find a way to trigger it. I think the service needs to be killed before the timeout.


MortenGregersen:
Maybe I should add something to the `README.md`... ☺️

packages/wdio-appium-service/src/launcher.js
```diff
@@ -0,0 +1,120 @@
+import logger from '@wdio/logger'
+import fs from 'fs-extra'
```
wswebcreation: You can also import all the functions instead of the complete library.

packages/wdio-appium-service/src/utils/getFilePath.js
```diff
@@ -0,0 +1,21 @@
+import path from 'path'
```
wswebcreation: Same here, you can use 'import {join, resolve} from "path"'

packages/wdio-appium-service/src/utils/getFilePath.js
```diff
@@ -0,0 +1,21 @@
...
+ * Resolves the given path into a absolute path and appends the default filename as fallback when the provided path is a directory.
+ * @param  {String} logPath         relative file or directory path
+ * @param  {String} defaultFilename default file name when filePath is a directory
+ * @return {String}                 absolute file path
+ */
+export default function getFilePath (filePath, defaultFilename) {
+    let absolutePath = path.resolve(filePath)
+
+    // test if we already have a file (e.g. selenium.txt, .log, log.txt, etc.)
+    // NOTE: path.extname doesn't work to detect a file, cause dotfiles are reported by node to have no extension
```
baruchvlz: Can't you use fs for this?

```
import { lstatSync } from 'fs';

lstatSync(filePath).isFile(); // this works for dot files
```

packages/wdio-appium-service/src/launcher.js
```diff
@@ -0,0 +1,120 @@
...
+        }
+        return cliOptionName
+    }
+
+    static _sanitizeCliOptionValue(value) {
+        const valueString = String(value)
+        // Encapsulate the value string in single quotes if it contains a white space
+        return /\s/.test(valueString) ? `'${valueString}'` : valueString
+    }
+}
```
baruchvlz: This is more of a question than anything else, but isn't it a convention to have static methods defined before the constructor? (not like it matters much other than just readability)

MortenGregersen: The methods didn't need to be static so I removed it. I think they should be in the bottom as they are helpers

packages/wdio-appium-service/src/utils/getFilePath.js
MortenGregersen: Fixed 👍

packages/wdio-appium-service/src/launcher.js
MortenGregersen: Fixed 👍

packages/wdio-appium-service/src/utils/getFilePath.js
MortenGregersen: It should not check if the file exists - just give a valid path.

christian-bromann:
@MortenGregersen can you check why the unit tests are failing?

MortenGregersen:
> @MortenGregersen can you check why the unit tests are failing?

I need some help figuring out how to fix the unit tests. Alternatively we should disable the test that checks if Appium is started as I wrote in the last paragraph of pull request description.

christian-bromann:
@MortenGregersen in your unit test please make sure to mock out every 3rd party dependency like `child_process`. You should not start an actual Appium server or a new process, you have to stub it out and just check of the child process command was called as expected.

MortenGregersen:
@christian-bromann I've mocked out all 3rd party dependencies and all unit tests are now green

packages/wdio-appium-service/tests/launcher.test.js
```diff
@@ -0,0 +1,177 @@
+import { AppiumLauncher } from '../src/launcher'
+import cp from 'child_process'
+import fs from 'fs-extra'
+import path from 'path'
+
+jest.mock('child_process', () => ({
+    spawn: jest.fn(),
+}))
+jest.mock('fs-extra', () => ({
```
christian-bromann: Can you move both of this into the mocks directory (`packages/wdio-appium-service/tests/__mocks__` which is the normal way to put these mocks into seperate files. Jest then automatically mocks them.

packages/wdio-appium-service/index.js
```diff
@@ -0,0 +1,2 @@
+const AppiumLauncher = require('./build/launcher').default
+module.exports = new AppiumLauncher()
```
christian-bromann: I thin we can remove this file, it doesn't seem to be used

packages/wdio-appium-service/src/launcher.js
```diff
@@ -0,0 +1,116 @@
...
+        this.command = ''
+        this.appiumArgs = []
+        this.waitStartTime = 0
+    }
+
+    async onPrepare(config) {
+        const appiumConfig = config.appium || {}
+
+        this.logPath = appiumConfig.logPath
+        this.command = appiumConfig.command || this._getAppiumCommand()
```
christian-bromann: shouldn't we use the Appium binary that has been downloaded as dependency (under `/packages/wdio-appium-service/node_modules/.bin/appium`?

packages/wdio-appium-service/tests/launcher.test.js
MortenGregersen: When I move the two `jest.mock()` to separate files (`child_process.js` and `fs-extra.js` in a `__mocks__` folder in the root of `packages/wdio-appium-service`- not in the `tests` folder) and add a `import jest from 'jest'` I get the following error. I haven't used jest and babel before, so I have no idea how to fix this. I placed the `jest.mock()` in the `launcher.test.js` file to match how `wdio-selenium-standalone-service` does.

```
Test suite failed to run

    babel-plugin-jest-hoist: The module factory of `jest.mock()` is not allowed to reference any out-of-scope variables.
    Invalid variable access: jest
    Whitelisted objects: Array, ArrayBuffer, Boolean, DataView, Date, Error, EvalError, Float32Array, Float64Array, Function, Generator, GeneratorFunction, Infinity, Int16Array, Int32Array, Int8Array, InternalError, Intl, JSON, Map, Math, NaN, Number, Object, Promise, Proxy, RangeError, ReferenceError, Reflect, RegExp, Set, String, Symbol, SyntaxError, TypeError, URIError, Uint16Array, Uint32Array, Uint8Array, Uint8ClampedArray, WeakMap, WeakSet, arguments, console, expect, isNaN, jest, parseFloat, parseInt, require, undefined, DTRACE_NET_SERVER_CONNECTION, DTRACE_NET_STREAM_END, DTRACE_HTTP_SERVER_REQUEST, DTRACE_HTTP_SERVER_RESPONSE, DTRACE_HTTP_CLIENT_REQUEST, DTRACE_HTTP_CLIENT_RESPONSE, global, process, Buffer, clearImmediate, clearInterval, clearTimeout, setImmediate, setInterval, setTimeout.
    Note: This is a precaution to guard against uninitialized mock variables. If it is ensured that the mock is required lazily, variable names prefixed with `mock` (case insensitive) are permitted.

      at invariant (node_modules/babel-plugin-jest-hoist/build/index.js:15:11)
      at newFn (node_modules/@babel/traverse/lib/visitors.js:193:21)
      at NodePath._call (node_modules/@babel/traverse/lib/path/context.js:53:20)
      at NodePath.call (node_modules/@babel/traverse/lib/path/context.js:40:17)
      at NodePath.visit (node_modules/@babel/traverse/lib/path/context.js:88:12)
      at TraversalContext.visitQueue (node_modules/@babel/traverse/lib/context.js:118:16)
      at TraversalContext.visitMultiple (node_modules/@babel/traverse/lib/context.js:85:17)
```

packages/wdio-appium-service/index.js
MortenGregersen: I've deleted the file :+1:

packages/wdio-appium-service/src/launcher.js
MortenGregersen: It now uses the path to the Appium binary in `node_modules`

packages/wdio-appium-service/tests/launcher.test.js
MortenGregersen: Can you help me out, @christian-bromann?

baruchvlz: @MortenGregersen  How are you importing the modules into your `__mocks__` file?

MortenGregersen: I didn't import the modules into a `__mocks__` file. I moved the `jest.mock()` code from the `launcher.test.js` to separate files (child_process.js and fs-extra.js in a `__mocks__` folder in the root of packages/wdio-appium-service- not in the tests folder) and added a `import jest from 'jest'`.

packages/wdio-appium-service/src/launcher.js
```diff
@@ -0,0 +1,117 @@
...
+        ensureFileSync(logFile)
+
+        log.debug(`Appium logs written to: ${logFile}`)
+        const logStream = createWriteStream(logFile, { flags: 'w' })
+        this.process.stdout.pipe(logStream)
+        this.process.stderr.pipe(logStream)
+    }
+
+    _getAppiumCommand() {
+        return join('node_modules', '.bin', 'appium')
```
christian-bromann: let's use `require.resolve('appium')` here

packages/wdio-appium-service/src/launcher.js
```diff
@@ -0,0 +1,117 @@
...
+
+            // Only non-boolean and non-null values are added as option values
+            if (typeof value !== 'boolean' && value !== null) {
+                cliArgs.push(this._sanitizeCliOptionValue(value))
+            }
+        }
+        return cliArgs
+    }
+
+    _lowerCamelToCliOptionName(camelCasedKey) {
```
christian-bromann: Can we replace this with https://github.com/blakeembrey/camel-case

packages/wdio-appium-service/src/launcher.js
MortenGregersen: Fixed 👍

packages/wdio-appium-service/src/launcher.js
MortenGregersen: I used `param-case` as it needs dashes instead of spaces.

MortenGregersen:
The build fails on test for other packages... How do we fix this, @christian-bromann?

packages/wdio-appium-service/tests/launcher.test.js
MortenGregersen: I got it to work somehow... 🤔

packages/wdio-appium-service/tests/launcher.test.js
```diff
@@ -0,0 +1,155 @@
+import { AppiumLauncher } from '../src/launcher'
+import childProcess from 'child_process'
+import fs from 'fs-extra'
+import path from 'path'
+
+jest.mock('child_process')
```
WillBrock: You can place the contents from the `__mocks__` directory here instead. An example can be found in wdio-selenium-standalone-service/tests/launcher.test.js. This way it will only mock child_process and fs for that file.

MortenGregersen: I did that in the first place but moved it to the `__mocks__` directory after @christian-bromann's comment (https://github.com/webdriverio/webdriverio/pull/3717#discussion_r266889489).
I've moved them back to the test file.

MortenGregersen:
Travis hangs trying to boot the virtual machine... 😕

christian-bromann:
https://www.traviscistatus.com/

PR #3742, opened 2019-03-20
Fix invalid urls

christian-bromann:
When navigating to pages like:
- google.com
- www.google.com
the driver will fail because the url is invalid.

Always prepend an http in front of the url.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -93,3 +93,4 @@ docs/_*.md
 .gradle
 wallaby.conf.js
 tags
+tests/.retry_succeeded
--- a/packages/webdriverio/src/commands/browser/url.js
+++ b/packages/webdriverio/src/commands/browser/url.js
@@ -34,6 +34,7 @@
  */
 
 import nodeUrl from 'url'
+import { validateUrl } from '../../utils'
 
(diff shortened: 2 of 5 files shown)


christian-bromann:
@CrispusDH @BorisOsipov can anyone review this small fix?

christian-bromann:
The current code will actually prevent navigating to urls such as `about:blank` .. not sure if this can have consequences, however people could still use the navigate command that comes without magic.

CrispusDH:
> what about special urls like `about:blank` or `data:text/html, <html contenteditable>` ?

@christian-bromann could we make exceptions for those kind of urls?

BorisOsipov:
We use at least `about:blank` to avoid some issues with SPA routing.

They are legal url schemas. Why should we forbid users to access them?

So I suppose adjust solution by replacing "Always prepend an http in front of the url." =>
"Prepend an http in front of the url *if schema is not specified*"

christian-bromann:
@BorisOsipov @CrispusDH thanks for the feedback, I think I found a good solution using the internal `URL` constructor. Please have a look!

PR #3746, opened 2019-03-21
Issue 3416: Support multiple errors in the Allure-reporter

nicholasbailey:
## Proposed changes
Over several PRs, we've introduced fixes to allow Jasmine to pass more than one failing expect to WebdriverIO. This matches Jasmine's soft assertion model, where expectation failures do not stop test execution. 

This PR passes those additional errors to Allure.

Unfortunately, Allure reporter itself assumes one error per test. The easiest way to work around that is to combine the errors into a single error holding their aggregated messages and stacks. It's not ideal, but allure doesn't offer us a great alternative.

## Types of changes

## Checklist

### DIFF
--- a/packages/wdio-allure-reporter/src/compoundError.js
+++ b/packages/wdio-allure-reporter/src/compoundError.js
@@ -1,0 +1,20 @@
+function indentAll(lines) {
+    return lines.split('\n').map(x => '    ' + x).join('\n')
+}
+
+/**
+ * An error that encapsulates more than one error, to support soft-assertions from Jasmine
+ * even though Allure's API assumes one error-per test
+ */
+export default class CompoundError extends Error {
+    constructor(...innerErrors) {
+        const message = ['CompoundError: One or more errors occurred. ---'].
+            concat(innerErrors.map(x => {
+                if (x.stack) return `${indentAll(x.stack)}\n--- End of stack trace ---`
+                else return `   ${x.message}\n--- End of error message ---`
+            })).join('\n')
+
+        super(message)
+        this.innerErrors = innerErrors
+    }
+}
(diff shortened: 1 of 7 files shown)


BorisOsipov:
@nicholasbailey Thanks for improving allure reporting. Please make linter happy for next review steps.

nicholasbailey:
@christian-bromann @BorisOsipov linter errors _should_ be fixed (sorry about that!)

christian-bromann:
@nicholasbailey something seems to be still off with the unit tests

nicholasbailey:
So there does. Fixing now

nicholasbailey:
@christian-bromann @BorisOsipov we should be good now (finally).
Sorry a lot of those tests passed locally because stack traces depend a bit on how you run the tests :)

BorisOsipov:
I don't forget about this. I am a bit busy now, I will see a few days later

nicholasbailey:
@CrispusDH just making sure I understand what you want. If there's only one error in a Jasmine test case, we still want to wrap that error in a CompoundError for reporting? 

While it's a little simpler from an implementation perspective, that seems like it would add noise to the reports themselves. I'm ok with that if that's folk's preference, but it's not what I would personally prefer.

Note that Jasmine test cases will always return an errors array, even if there is only one item in the array.

CrispusDH:
Let @BorisOsipov choose what option is preferable. I was thinking about what place should actual take in account this piece of `if .. else` logic:
```js
    if (test.errors && test.errors.length) {
        if (test.errors.length == 1) {
            return test.errors[0]
        } else {
            return new CompoundError(...test.errors)
        }
    } else {
        return test.error
    }
```
If you think that `CompoundError()` should work only with `test.errors` part of `test` entity it could be an option. In my opinion it should take `test` entity and decide what to do inside method instead of giving this this part of work to `getErrorFromFailedTest ()`.

PR #3756, opened 2019-03-25
Fail test if sauce service is used but no credentials are provided

christian-bromann:
Currently if you have the sauce services defined within your config but don't provide any sauce credentials the testrunner will try to run against a local instance without letting anyone know why.

I think it should be better propagated with at least a dominant warning that sauce service is used without sauce credentials. I am not sure if I want to break tests of people that already do it the wrong way. Therefor we should show that error somewhere explicit.

Just having the test fail completely but then again this could throw off tests from people that already use the sauce service even though they run locally.

**Additional context**
This should also be applied to the browserstack and testingbot service.

### DIFF
--- a/packages/wdio-browserstack-service/src/service.js
+++ b/packages/wdio-browserstack-service/src/service.js
@@ -11,11 +11,23 @@ export default class BrowserstackService {
         this.failures = 0
     }
 
+    beforeSession (config) {
+        /**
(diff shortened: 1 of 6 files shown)


abjerstedt:
@christian-bromann Do we also need to update the testingBot code to behave the same?

PR #3768, opened 2019-03-25
webdriverio: return promise-wrapped async elements from $/$$ in async mode

### LINKED ISSUE #3760: WebdriverIOAsync findElement(s) signatures have WebdriverIO return types
 - **WebdriverIO version:** 5.7.5
 - **Mode:** Standalone

**Describe the bug**

The use of `WrapWithPromise` to generate `WebdriverIOAsync` types from `WebdriverIO` types means that the return types of most methods are correctly typed as Promises rather than the resolved types of those Promises.

However, in the case of `WebdriverIOAsync.Browser.findElement(s)` (or `$` or `$$`), the resolved type is itself an element, but since these method signatures have been generated from `WebdriverIO.Browser`, that element is of type `WebdriverIO.Element`, *not* `WebdriverIOAsync.Element` as it should be.

**To Reproduce**
```
type T = ReturnType<WebdriverIOAsync.Browser["$"]>; // Promise<WebdriverIO.Element>
```

**Expected behavior**
```
type T = ReturnType<WebdriverIOAsync.Browser["$"]>; // Promise<WebdriverIOAsync.Element>
```

So that we can do the following without a type error: 

```
const element = await browser.$("selector");
await element.$("sub-selector").then((subElement) => subElement.click());
```

richsilv:
## Proposed changes

`WebdriverIO.Async` types are updated so that the `$` and `$$` methods on `Element` and `Browser` return one or an array of `WebdriverIOAsync.Element` objects wrapped in a promise, rather than one of an array of `WebdriverIO.Element` objects wrapped in a promise (which is inconsistent with the API).

Resolves #3760 

## Types of changes

## Checklist

## Further comments

N/A

### DIFF
--- a/packages/webdriverio/webdriverio.d.ts
+++ b/packages/webdriverio/webdriverio.d.ts
@@ -3,21 +3,36 @@
 type BrowserObject = WebDriver.ClientOptions & WebDriver.ClientAsync & WebdriverIOAsync.Browser;
 
 // Element commands that should be wrapper with Promise
-type ElementPromise = Omit<WebdriverIO.Element, 'addCommand'>;
+type ElementPromise = Omit<WebdriverIO.Element, 'addCommand' | '$' | '$$'>;
 
 // Element commands wrapper with Promise
 type ElementAsync = {
     [K in keyof ElementPromise]: WrapWithPromise<ElementPromise[K], ReturnType<ElementPromise[K]>>
+} & {
+  $(
+    selector: string | Function
+  ): Promise<WebdriverIOAsync.Element>;
+  $$(
+    selector: string | Function
+  ): Promise<WebdriverIOAsync.Element[]>;
 }
+
 // Element commands that should not be wrapper with promise
 type ElementStatic = Pick<WebdriverIO.Element, 'addCommand'>
 
 // Browser commands that should be wrapper with Promise
-type BrowserPromise = Omit<WebdriverIO.Browser, 'addCommand' | 'options'>;
+type BrowserPromise = Omit<WebdriverIO.Browser, 'addCommand' | 'options' | '$' | '$$'>;
 
 // Browser commands wrapper with Promise
 type BrowserAsync = {
     [K in keyof BrowserPromise]: WrapWithPromise<BrowserPromise[K], ReturnType<BrowserPromise[K]>>
+} & {
+  $(
+    selector: string | Function
+  ): Promise<WebdriverIOAsync.Element>;
+  $$(
+    selector: string | Function
+  ): Promise<WebdriverIOAsync.Element[]>;
 }
 
 // Browser commands that should not be wrapper with promise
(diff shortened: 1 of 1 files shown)


christian-bromann:
@richsilv can you please sign the CLA?

packages/webdriverio/webdriverio.d.ts
```diff
@@ -3,21 +3,36 @@
 type BrowserObject = WebDriver.ClientOptions & WebDriver.ClientAsync & WebdriverIOAsync.Browser;
 
 // Element commands that should be wrapper with Promise
-type ElementPromise = Omit<WebdriverIO.Element, 'addCommand'>;
+type ElementPromise = Omit<WebdriverIO.Element, 'addCommand' | '$' | '$$'>;
 
 // Element commands wrapper with Promise
 type ElementAsync = {
     [K in keyof ElementPromise]: WrapWithPromise<ElementPromise[K], ReturnType<ElementPromise[K]>>
+} & {
```
CrispusDH: could we somehow reduce duplication of this code?

richsilv: :thumbsup: makes sense. Should I amend the commit, or add another?

CrispusDH: I think it is up to you. I prefer additional commit :)

richsilv: as requested

christian-bromann:
@CrispusDH I am not a typescript expert so if you are cool with the change, I will merge it

CrispusDH:
Looks good for me

mgrybyk:
moment, taking a look

packages/webdriverio/webdriverio.d.ts
```diff
@@ -3,22 +3,33 @@
 type BrowserObject = WebDriver.ClientOptions & WebDriver.ClientAsync & WebdriverIOAsync.Browser;
 
 // Element commands that should be wrapper with Promise
-type ElementPromise = Omit<WebdriverIO.Element, 'addCommand'>;
+type ElementPromise = Omit<WebdriverIO.Element, 'addCommand' | '$' | '$$'>;
+
+// Methods which return async element(s) so non-async equivalents cannot just be promise-wrapped
+interface AsyncSelectors {
+    $(
```
mgrybyk: small suggestion, can we move $ and $$ to types and reuse types in global variables declarations, something like this:
```
type $ = (selector: string | Function) => Promise<WebdriverIOAsync.Element>;
type $$ = (selector: string | Function) => Promise<WebdriverIOAsync.Element[]>;
interface AsyncSelectors {
    $: $;
    $$: $$;
}
```
...
```
declare var browser: BrowserObject;
declare var $: $;
declare var $$: $$;
```

What do you think?

richsilv: I can see the point of the types, but for my benefit could you explain why these should be declared globally? Are they supposed to be accessible that way?

christian-bromann: @richsilv if you use the wdio testrunner yes

richsilv: Makes sense, thanks! I'll make the requested change.

richsilv: okay, done

mgrybyk:
I think it is required to rebase in order to resolve codecov failure, @richsilv

PR #3782, opened 2019-03-27
validate typings

mgrybyk:
## Proposed changes

POC: typings validation

We had repetitive problems because of typos in jsdoc/etc that causes TypeScript compiler to fail and block users. With this PR I want to introduce some tests to prevent merging of broken typings files.

## Types of changes

- [x] Polish

## Checklist

## Further comments

TODOs
- [ ] fix typings validation for jasmine
- [ ] fix typings validation for selenium-standalone-service
- [ ] raise ticket for execute and call typings in async mode


mgrybyk:
@CrispusDH may you take a look?

scripts/type-generation/test/copy.js
```diff
@@ -0,0 +1,46 @@
...
+
+    '@types/selenium-standalone': 'packages/wdio-selenium-standalone-service/node_modules/@types/selenium-standalone',
+    '@wdio/selenium-standalone-service': 'packages/wdio-selenium-standalone-service',
+}
+
+/**
+ * copy package.json and typings from package to type-generation/test/.../node_modules
+ */
+async function copy() {
+    outDirs.forEach(async outDir => {
```
CrispusDH: You have to use either `for ... of` or `await Promise.all(packages.map())` construction. I prefer first option. `forEach` method does not in a proper way with async ... await https://stackoverflow.com/questions/37576685/using-async-await-with-a-foreach-loop

christian-bromann: keeping this open until this is resolved.

mgrybyk: my bad, fixed

mgrybyk:
working on TODOs right now

christian-bromann:
@mgrybyk please remove the wip label when ready to review

CrispusDH:
I think it might be useful to create a short README.md about "How we check types". Some steps or ideas. It would be helpful for contributors if future. @mgrybyk , what do you think? I mean it could be difficult to figure out what commands are executes and what happens in files.

mgrybyk:
@CrispusDH added README.md. Is it helpful? What kind of information is it missing?

christian-bromann:
@mgrybyk actually would you mind to move the type generation tests to `/tests/`? I think having smoke and type checks in one place is better than having them across the project.

mgrybyk:
I'll take a look in next couple of days or even earlier

PR #3786, opened 2019-03-28
[docs] typo fix

Armanio:
Mini typo fix 🤷‍♂️

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -21,7 +21,7 @@ it('can handle commands synchronously', () => {
 you need return the command promise so it gets resolved properly as well as access the value when the promise got resolve:
 
 ```js
-it('handles commands as promises', ()  =>{
+it('handles commands as promises', ()  => {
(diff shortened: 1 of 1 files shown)


docs/API.md
```diff
@@ -21,7 +21,7 @@ it('can handle commands synchronously', () => {
 you need return the command promise so it gets resolved properly as well as access the value when the promise got resolve:
 
 ```js
-it('handles commands as promises', ()  =>{
+it('handles commands as promises', ()  => {
```
christian-bromann:
```suggestion
it('handles commands as promises', () => {
```

while we at it 😉

CrispusDH: hm, I will check whether exists or not eslint rule for it :)

BorisOsipov: It is docs. Linter doesn't process `.md` files

Armanio: Fixed!

PR #3794, opened 2019-03-30
Implement react$ and react$$ command

christian-bromann:
## Proposed changes

We want to simplify the way people fetch elements within React applications. This patch introduces `react$` and `react$$` that allow to query and React component bz y its actual component name and filter by prop and state properties.

## Types of changes

## Checklist

## Further comments

This is currently work in progress. There are some things that have to be changed in `resq` to get this working. A modified version of that package allowed to run the following test:

```js
await browser.url('https://ahfarmer.github.io/calculator/');

const btn7 = await browser.react$('t', { name: '7' })
const btn6 = await browser.react$('t', { name: '6' })
const btnMult = await browser.react$('t', { name: 'x' })
const btnEqual = await browser.react$('t', { name: '=' })
const display = await browser.$('.component-display')

await btn7.click()
await btnMult.click()
await btn6.click()
await btnEqual.click()

console.log(await display.getText()); // prints "42"
```

### DIFF
--- a/packages/webdriverio/src/scripts/resq.js
+++ b/packages/webdriverio/src/scripts/resq.js
@@ -1,0 +1,38 @@
+export const waitToLoadReact = function waitToLoadReact () {
+    window.resq.waitToLoadReact()
+}
+
+export const react$ = function react$ (selector, props) {
+    const elems = window.resq.resq$$(selector)
+
+    if (Object.keys(props).length === 0) {
+        return elems[0]
+    }
+
+    let elem = [...elems].find((elem) => {
+        for (const [name, value] of Object.entries(props)) {
+            // logs.push(`check ${name} and ${value} for ${elem[name]}`)
+            if (elem.props[name] === value) {
+                return true
+            }
+        }
+
+        return false
+    })
+
+    if (!elem || (!elem.node && elem.children.length === 0)) {
+        return
+    }
+
+    if (elem.node) {
+        return elem.node
+    }
+
+    elem = elem.node || elem.children[0].node
+
+    if (!elem) {
+        throw new Error('NotFound')
+    }
+
+    return elem
+}
--- a/packages/webdriverio/package.json
+++ b/packages/webdriverio/package.json
@@ -64,6 +64,7 @@
     "lodash.isplainobject": "^4.0.6",
     "lodash.merge": "^4.6.1",
     "lodash.zip": "^4.2.0",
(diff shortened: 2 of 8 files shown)


packages/webdriverio/src/scripts/resq.js
```diff
@@ -0,0 +1,38 @@
...
+    const elems = window.resq.resq$$(selector)
+
+    if (Object.keys(props).length === 0) {
+        return elems[0]
+    }
+
+    let elem = [...elems].find((elem) => {
+        for (const [name, value] of Object.entries(props)) {
+            // logs.push(`check ${name} and ${value} for ${elem[name]}`)
+            if (elem.props[name] === value) {
```
baruchvlz: If you use `resq$(selector).byProps(props)` you will get the first element in which the props match. Accordingly, you could use `resq$$` to select all elements that match selector and props. 

Keep in mind, that as of right now, `byProps` and `byState` are deep matchers.

packages/webdriverio/src/scripts/resq.js
```diff
@@ -0,0 +1,38 @@
...
+
+    if (!elem || (!elem.node && elem.children.length === 0)) {
+        return
+    }
+
+    if (elem.node) {
+        return elem.node
+    }
+
+    elem = elem.node || elem.children[0].node
```
baruchvlz: As I mentioned to you on gitter. This will not quite work as expected if the JSX is wrapped in a Fragment.

With this implementation, this is how it'll work:

```javascript
import React from 'react';

const MyCmp = () => (
  <React.Fragment>
    <div>something</div>
    <div> something else </div>
  </React.Fragment>
)


//

react$('MyCmp') // <div>something</div>
```

baruchvlz: I'm working on a feature to fix this, it's still raw and untested, but here's the gist of it https://github.com/baruchvlz/resq/commit/7b9ffd1272da79ec4ace45768ff0282e93fcad5f

packages/webdriverio/src/scripts/resq.js
christian-bromann: That's awesome, I didn't know about this.

christian-bromann:
Have to merge into a new branch since git seems to not like branch names with `$` in it.

christian-bromann:
Follow up PR #3797

PR #3797, opened 2019-04-01
Implement react$ and react$$ command

christian-bromann:
## Proposed changes

We want to simplify the way people fetch elements within React applications. This patch introduces `react$` and `react$$` that allow to query and React component bz y its actual component name and filter by prop and state properties.

## Types of changes

## Checklist

## Further comments

This is currently work in progress. There are some things that have to be changed in `resq` to get this working. A modified version of that package allowed to run the following test:

```js
await browser.url('https://ahfarmer.github.io/calculator/');

const btn7 = await browser.react$('t', { name: '7' })
const btn6 = await browser.react$('t', { name: '6' })
const btnMult = await browser.react$('t', { name: 'x' })
const btnEqual = await browser.react$('t', { name: '=' })
const display = await browser.$('.component-display')

await btn7.click()
await btnMult.click()
await btn6.click()
await btnEqual.click()

console.log(await display.getText()); // prints "42"
```

### DIFF
--- a/packages/webdriverio/src/commands/browser/react$$.js
+++ b/packages/webdriverio/src/commands/browser/react$$.js
@@ -10,12 +10,8 @@
     it('should calculate 7 * 6', () => {
         browser.url('https://ahfarmer.github.io/calculator/');
 
-        browser.react$('t', { name: '7' }).click()
-        browser.react$('t', { name: 'x' }).click()
-        browser.react$('t', { name: '6' }).click()
-        browser.react$('t', { name: '=' }).click()
-
-        console.log($('.component-display').getText()); // prints "42"
+        const orangeButtons = browser.react$$('t', { orange: true })
+        console.log(orangeButtons.map((btn) => btn.getText())); // prints "[ '÷', 'x', '-', '+', '=' ]"
     });
  * </example>
  *
--- a/packages/webdriverio/src/commands/browser/react$$.js
+++ b/packages/webdriverio/src/commands/browser/react$$.js
@@ -27,7 +27,7 @@ import { waitToLoadReact, react$$ as react$$Script } from '../../scripts/resq'
 
 const resqScript = fs.readFileSync(require.resolve('resq'))
 
-export default async function react$ (selector, props = {}, state = {}) {
+export default async function react$$ (selector, props = {}, state = {}) {
     await this.executeScript(resqScript.toString(), [])
     await this.execute(waitToLoadReact)
     const res = await this.execute(react$$Script, selector, props, state)
--- a/packages/webdriverio/src/commands/browser/react$$.js
+++ b/packages/webdriverio/src/commands/browser/react$$.js
@@ -1,7 +1,7 @@
 /**
  *
- * The `react$` command is a useful command to query React Components by their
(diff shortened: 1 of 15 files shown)


christian-bromann:
@baruchvlz this is now ready for review

packages/webdriverio/src/scripts/resq.js
```diff
@@ -0,0 +1,55 @@
...
+    /**
+     * see https://github.com/baruchvlz/resq/issues/19
+     */
+    if (typeof elems === 'string') {
+        throw new Error(elems)
+    }
+
+    elems = elems
+        .byProps(props)
+        .byState(state)
```
baruchvlz: Hm, I don't think this will work as expected.

Take this scenario

```javascript
return (
  <div>
    <MyCmp s="b" />
    <MyCmp s="b" />
  </div>
);
```

Searching `byProps({})` we would expect to return the component, but currently this is the behaviour:

```javascript
resq$$('MyCmp').byProps({}).length // 0
resq$$('MyCmp').byProps({ s: 'b' }).length // 2
```

I think I'll release a patch later today to fix this so that if we pass an empty object the filter should work as if it's not being filtered at all. What do you think?

packages/webdriverio/src/commands/browser/react$$.js
```diff
@@ -0,0 +1,35 @@
...
+ * @return {Element}
+ *
+ */
+import fs from 'fs'
+import { getElements } from '../../utils/getElementObject'
+import { waitToLoadReact, react$$ as react$$Script } from '../../scripts/resq'
+
+const resqScript = fs.readFileSync(require.resolve('resq'))
+
+export default async function react$ (selector, props = {}, state = {}) {
```
baruchvlz: Shouldn't the name be `react$$`?

packages/webdriverio/src/scripts/resq.js
```diff
@@ -0,0 +1,55 @@
...
+}
+
+export const react$ = function react$ (selector, props, state) {
+    let elems = window.resq.resq$$(selector)
+
+    /**
+     * see https://github.com/baruchvlz/resq/issues/19
+     */
+    if (typeof elems === 'string') {
+        throw new Error(elems)
```
baruchvlz: I'll update this in the next patch along side the update to the filtering methods so you don't have to deal with it here.

packages/webdriverio/src/scripts/resq.js
christian-bromann: > I think I'll release a patch later today to fix this so that if we pass an empty object the filter should work as if it's not being filtered at all. What do you think?

Yeah I think that would make sense. If the behavior is that if you pass in a filter like `{a: 'foobar'}` it finds components that have a property `a` equals `foobar` __and__ also others rather than finding only components that contain only on property `a` equals `foobar`. strict vs non strict and I think if you pass in a filter atm it is non strict so should it behave passing an empty object.

christian-bromann:
https://github.com/baruchvlz/resq/pull/20 needs to be released first before merging this.

christian-bromann:
merged in #3862

PR #3819, opened 2019-04-04
updates to resq script

baruchvlz:
## Proposed changes

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/packages/webdriverio/src/scripts/resq.js
+++ b/packages/webdriverio/src/scripts/resq.js
@@ -3,25 +3,25 @@ export const waitToLoadReact = function waitToLoadReact () {
 }
 
 export const react$ = function react$ (selector, props, state) {
-    let elem = window.resq.resq$(selector)
+    let element = window.resq.resq$(selector)
         .byProps(props)
         .byState(state)
 
-    if (!elem.name) {
+    if (!element.name) {
         return { message: `React element with selector "${selector}" wasn't found` }
     }
 
-    return elem.node || elem.children[0].node
+    return element.node || element.children[0].node
 }
 
 export const react$$ = function react$$ (selector, props, state) {
-    let elems = window.resq.resq$$(selector)
+    let elements = window.resq.resq$$(selector)
         .byProps(props)
         .byState(state)
 
-    if (!elems.length) {
+    if (!elements.length) {
         return { message: `React elements with selector "${selector}" wasn't found` }
     }
 
-    return [...elems].map(elem => elem.node || elem.children[0].node)
+    return [...elements].map(element => element.node || element.children[0].node)
 }
--- a/packages/webdriverio/src/scripts/resq.js
+++ b/packages/webdriverio/src/scripts/resq.js
@@ -20,7 +20,7 @@ export const react$$ = function react$$ (selector, props, state) {
         .byState(state)
 
     if (!elements.length) {
-        return { message: `React elements with selector "${selector}" wasn't found` }
+        return []
     }
 
     return [...elements].map(element => element.node || element.children[0].node)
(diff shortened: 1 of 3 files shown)


packages/webdriverio/src/scripts/resq.js
```diff
@@ -3,35 +3,25 @@ export const waitToLoadReact = function waitToLoadReact () {
 }
 
 export const react$ = function react$ (selector, props, state) {
-    let elems = window.resq.resq$$(selector)
+    let elem = window.resq.resq$(selector)
```
CrispusDH: It might be clearer to use full word `element` instead of `elem`.

packages/webdriverio/src/scripts/resq.js
```diff
@@ -3,35 +3,25 @@ export const waitToLoadReact = function waitToLoadReact () {
...
     let elems = window.resq.resq$$(selector)
         .byProps(props)
         .byState(state)
 
-    return [...elems].map((elem) => {
-        if (elem.node) {
-            return elem.node
-        }
+    if (!elems.length) {
+        return { message: `React element with selector "${selector}" wasn't found` }
```
CrispusDH: I think will be more precise `React elements...` instead of `element`

packages/webdriverio/src/scripts/resq.js
```diff
@@ -3,35 +3,25 @@ export const waitToLoadReact = function waitToLoadReact () {
...
-            return elem.node
-        }
+    if (!elems.length) {
+        return { message: `React elements with selector "${selector}" wasn't found` }
+    }
 
-        elem = elem.node || elem.children[0].node
-        return elem
-    })
+    return [...elems].map(elem => elem.node || elem.children[0].node)
```
CrispusDH: Should be here check `!element.name`?

baruchvlz: The reason I check for `element.name` in the `react$` function is because the resq library returns an object with at least one private property, `_nodes`. So I assume that if the object doesn't have a property `name` then library didn't find a component.

In contrast, for `react$$` the library will return an empty object if no components have been found. That's why it's safe here to assume that there is at least one component in the array since `elems.length` is true.

Hopefully this makes sense.

packages/webdriverio/src/scripts/resq.js
baruchvlz: Adjusted here https://github.com/webdriverio/webdriverio/pull/3819/commits/230caf4b1cc81bcfb5cd9d60aace6466bebb8677#diff-6e003f06aad6f9b928c5d4923fcd06caR52

packages/webdriverio/src/scripts/resq.js
baruchvlz: I agree. Changed d9cc9c6

packages/webdriverio/src/scripts/resq.js
```diff
@@ -3,35 +3,25 @@ export const waitToLoadReact = function waitToLoadReact () {
...
+    let elements = window.resq.resq$$(selector)
         .byProps(props)
         .byState(state)
 
-    return [...elems].map((elem) => {
-        if (elem.node) {
-            return elem.node
-        }
+    if (!elements.length) {
+        return { message: `React elements with selector "${selector}" wasn't found` }
```
christian-bromann: for fetching elements we should comply to the webdriver spec which is just return an empty array. Only the `fineElement` command would return an error if it can not be found.

baruchvlz: Return an empty array in both functions?

christian-bromann: No, just for react$$

baruchvlz: 4fb8094

PR #3832, opened 2019-04-09
Ensure ms edge runs on w3c on Sauce

christian-bromann:
## Proposed changes

Seems that `webdriver` doesn't properly determine MS Edge sessions on Sauce as W3C sessions.

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/packages/webdriver/src/utils.js
+++ b/packages/webdriver/src/utils.js
@@ -198,11 +198,19 @@ export function isW3C (capabilities) {
      * assume session to be a WebDriver session when
      * - capabilities are returned
      *   (https://w3c.github.io/webdriver/#dfn-new-sessions)
-     * - a `platform` capability is not given but a platformName is returned
-     *   which is not defined in the JSONWire protocol
+     * - it is an Appium session (since Appium is full W3C compliant)
      */
     const isAppium = capabilities.automationName || capabilities.deviceName
-    return Boolean((!capabilities.platform && capabilities.platformName) || isAppium)
+    const hasW3CCaps = (
+        capabilities.platformName &&
+        capabilities.browserVersion &&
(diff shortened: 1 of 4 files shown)


mgrybyk:
Does every W3C browser have **browserVersion** and **setWindowRect**?

christian-bromann:
> Does every W3C browser have **browserVersion** and **setWindowRect** in capabilities?

No, I was looking for the responses I tested and some did not contain one or the other but either one of these were available. I seems though that `platformName` and `browserVersion` is always there.

packages/webdriver/src/utils.js
```diff
@@ -198,11 +198,19 @@ export function isW3C (capabilities) {
...
     const isAppium = capabilities.automationName || capabilities.deviceName
-    return Boolean((!capabilities.platform && capabilities.platformName) || isAppium)
+    const hasW3CCaps = (
+        capabilities.platformName &&
+        capabilities.browserVersion &&
+        /**
+         * local safari doesn't provide platformVersion therefor
+         * check also for setWindowRect
+         */
+        (capabilities.platformVersion || capabilities.setWindowRect)
```
mgrybyk: One more question, does all w3c browsers have either platform version or setWindowRect?

christian-bromann: yes both caps are part of the W3C spec and __not__ within jsonwp. Usually `platformVersion` should exist in any session response but I don't see it only for running local Safari sessions which is why I added another check for `setWindowRect` which is also commonly set.

christian-bromann: this is definitely not a 100% bullet proof solution yet but improves the problem we have with Edge on Sauce.

Issue #2866, opened 2018-08-24
Implement "Element is not clickable" autofix


christian-bromann:
_From @christian-bromann on March 10, 2018 16:38_

## The problem

Sometimes commands on elements like (click) can't be executed due to the "Element is not clickable at xy" exception. Let's implement a couple of workarounds (explained in [this great SO post](https://stackoverflow.com/a/44916498/765174)) as part of the [element handler error middleware](https://github.com/webdriverio/v5/blob/master/packages/webdriverio/src/middlewares.js#L17) to prevent users from running into this issue.

_Copied from original issue: webdriverio/v5#21_

christian-bromann:
_From @seadb on March 27, 2018 19:0_

I will be sew happy when this is implemented :)

christian-bromann:
👏 nice, let me know if you have any questions on this

christian-bromann:
_From @thephilwells on April 3, 2018 20:33_

Is this issue claimed? I'd like to give it a shot if I may.

christian-bromann:
@seadb any progress on this? Or can @thephilwells take over?

christian-bromann:
Let's ensure we include the work that is already done by @thephilwells

christian-bromann:
Removing it from the v5 release as it is currently also not implemented in v4 and I want to get v5 out asap.

thephilwells:
Makes sense. Apologies for dragging my feet on this.

christian-bromann:
@thephilwells feel free to propose a PR

WillBrock:
@abjerstedt is this already implemented in v5?

christian-bromann:
@WillBrock nope, it should be implemented in the webdriverio middleware. There are a couple of steps we can try to avoid that error, e.g. scroll element into the center of the viewport. However we should not try to remove the element or move it somewhere else because we shouldn't reject the meaning of that error.

mgrybyk:
@CrispusDH is this one fixed by your PR?

CrispusDH:
@mgrybyk it is one piece of SO answer. Also, my workaround will work if overlapping element changes position during scrolling. But in this specific case it works for me. I have implement example (link in PR I think).

christian-bromann:
I think there are various scenarios where an element is not clickable (see issues above).

erwinheitzman:
Just want to get something clear.
Are we allowing elements to be clicked that in reality could never be clicked (due to overlapping elements)?

[...]

PR #3865, opened 2019-04-17
webdriverio: improve middlewares

CrispusDH:
## Proposed changes
Improve `middleware` a bit by handling `Element <td>...</td> is not clickable at point (116, 704)` error. 

Also, I saw that we have some issues with IE and Edge about error text, so I decided to add `toLowerCase()`.

## Types of changes

## Checklist

## Further comments

I faced with issue in my project that I get error:
```js
unknown error: Element <td>...</td> is not clickable at point (116, 704). Other element would receive the click: <a href="#" class="js-send-form" title="">...</a>
```
it was fixed by adding `await element.scrollIntoView();` before `.click()`. I have remembered about #2866 . So, I decided improve a bit middleware. It does not do any "magic", just trying to scroll into view and do the same command.

I'm not familiar with unit tests a lot, so I did not get how to create unit tests for checking it.  But I will try to figure it out if you will think that it's necessary.

### DIFF
--- a/packages/webdriverio/src/middlewares.js
+++ b/packages/webdriverio/src/middlewares.js
@@ -29,11 +29,15 @@ export const elementErrorHandler = (fn) => (commandName, commandFn) => {
 
                 return result
             } catch (error) {
-                if (error.name === 'stale element reference') {
+                if (error.name.toLowerCase() === 'stale element reference') {
                     const element = await refetchElement(this, commandName)
                     this.elementId = element.elementId
                     this.parent = element.parent
 
+                    return await fn(commandName, commandFn).apply(this, args)
+                }
+                if (error.name.toLowerCase().includes('is not clickable at point')) {
+                    await element.scrollIntoView()
                     return await fn(commandName, commandFn).apply(this, args)
                 }
                 throw error


BorisOsipov:
I am not sure about this.

1) Selenium itself scrolls before click
2) In some cases scroll won't help for "Element is not clickable at point"(overlays and etc)
3) I will always against any magic that I can't control(disable\enable).

Instead we can refetch element as we did for stale reference error several time and hope clickable issue goes. 
From my experience refetching solves most of not clickable issues. Other issues have to be solved by end users code(they should decide what to do - scroll, or remove overlaping element.)

CrispusDH:
1. Refetch is redundant for `Element is not clickable...` because element has been already exists and clickable, you don't need to refetch it.
2. My additional flow will not be a magic. It will not rid off error in case if whole element is overlayed by another element. Need to understand how click action executes. Without my PR it will scroll to element before any part of it will appear. With my PR it will scroll before all element appears in a view point. I see it helpful.

Even Alexei Barantsev was talking about willing to write some kind of wrapper for click action to make click not once in the middle of element but clicking 5-8 times on different parts of element.

I'm open to hear your additional questions and concerns :)

BorisOsipov:
>Refetch is redundant for Element is not clickable... because element has been already exists and clickable, you don't need to refetch it.

yes, refetch don't need, but retries for executing click will help.

>My additional flow will not be a magic.
It will. I call `click` protocol command from my code. There is not additional scroll in my cod, but you introduce it.

>Even Alexei Barantsev was talking about willing to write some kind of wrapper for click action to make click not once in the middle of element but clicking 5-8 times on different parts of element.
1) Please, proof your points with links to video or etc. Or mention Alexei to proof. I don't have such info.
2) Btw it indeed good for wrappers. Wrappers is good when you write\use them by yourself and understood what they do. But in our docs we are clearly describe what command do https://webdriver.io/docs/api/element/click.html
We execute protocol command. No magic here

>I'm open to hear your additional questions and concerns :)

totally agree with you and happy to disscuss ypur approach to help with `Element is not clickable` issue.

christian-bromann:
> wrapper for click action to make click not once in the middle of element but clicking 5-8 times on different parts of element.

This would be indeed a good workaround. Element can be for many reasons not clickable. I have doubts that just scrolling it into view and clicking again will solve the problem. Do you have an example that we could use to test your changes?

CrispusDH:
@BorisOsipov  Barantsev speech on Heinzenberg conf (RUS) - https://youtu.be/4dh--iD_zK8?t=2978

I will try create some demo to check will my PR help or not.

CrispusDH:
I create small Demo App and two tests inside that showing case when `Element isn't clickable...` error can be handled by `scrollIntoView` method. Instructions in README. I want to mention that in my working project I faced it with chrome browser in mobile emulation mode. In Demo App I use common chrome browser.

link to the repo - https://github.com/CrispusDH/element-isnt-clickable

abjerstedt:
> > wrapper for click action to make click not once in the middle of element but clicking 5-8 times on different parts of element.
> 
> This would be indeed a good workaround. Element can be for many reasons not clickable. I have doubts that just scrolling it into view and clicking again will solve the problem. Do you have an example that we could use to test your changes?

This shouldn't be in the middleware imo. We are getting a little hacky and using brute force at that point (or at least make it config option).

CrispusDH:
Thanks @abjerstedt for sharing your opinion. I'm really appreciate it 👍

abjerstedt:
Oops i quoted both people... I meant that at the 5-8 random clicks.

CrispusDH:
In this PR I want to discuss adding `scrollIntoView` only in case of specific error.

CrispusDH:
I think I will close it. I don't see that it would significant improve middleware. I will think about other ways to do it :)

Issue #3867, opened 2019-04-18
 When loglevel for specific loggers are set to silent, info and debug logs appear


kmnaid:
**Environment:**
 - **WebdriverIO version:** @wdio/cli@5.7.14
 - **Mode:** WDIO Testrunner
 - **If WDIO Testrunner, running sync/async:** sync
 - **Node.js version:** v10.15.2
 - **NPM version:** 6.4.1
 - **Browser name and version:** Chrome Version 73.0.3683.86 
 - **Platform name and version:** OSX 10.14.4 
 - **Additional wdio packages used (if applicable):** 
```
├─┬ @wdio/cli@5.7.14
│ ├─┬ @wdio/config@5.7.13
│ │ ├── @wdio/logger@5.7.8 deduped
│ ├─┬ @wdio/logger@5.7.8
│ ├─┬ @wdio/utils@5.7.8
│ │ ├── @wdio/logger@5.7.8 deduped
│ │ ├── @wdio/config@5.7.13 deduped
│ │ ├── @wdio/logger@5.7.8 deduped
│ │ ├── @wdio/repl@5.7.13 deduped
│ │   ├── @wdio/config@5.7.13 deduped
│ │   ├── @wdio/logger@5.7.8 deduped
├─┬ @wdio/dot-reporter@5.7.8
│ ├── @wdio/reporter@5.7.8 deduped
├─┬ @wdio/local-runner@5.7.14
│ ├── @wdio/logger@5.7.8 deduped
│ ├─┬ @wdio/repl@5.7.13
│ │ └── @wdio/config@5.7.13 deduped
│ ├─┬ @wdio/runner@5.7.14
│ │ ├── @wdio/config@5.7.13 deduped
│ │ ├── @wdio/logger@5.7.8 deduped
│ │ ├── @wdio/utils@5.7.8 deduped
├─┬ @wdio/mocha-framework@5.7.14
│ ├── @wdio/config@5.7.13 deduped
│ ├── @wdio/logger@5.7.8 deduped
├─┬ @wdio/reporter@5.7.8
├─┬ @wdio/sauce-service@5.7.8
│ ├── @wdio/logger@5.7.8 deduped
├─┬ @wdio/selenium-standalone-service@5.7.8
├─┬ @wdio/spec-reporter@5.7.13
│ ├── @wdio/reporter@5.7.8 deduped
├─┬ @wdio/sync@5.7.13
│ ├── @wdio/config@5.7.13 deduped
[...]

PR #3872, opened 2019-04-23
apply wdioLogLevel before starting workers

mgrybyk:
## Proposed changes

Want to change current behaviour of applying `logLevel`.

Current:
1. Initialise loggers with default `logLevel` (because config is not yet parsed)
2. Set `logLevels`
3. Set `logLevel` when workers start

As a result `logLevel` is ignored for some loggers and it is not possible to override `logLevel` in conf file using ENV variable.

New:
1. Initialise loggers with default `logLevel` (because config is not yet parsed)
2. Set `logLevel` if was not provided by user with ENV var `WDIO_LOG_LEVEL`
3. Set `logLevels`

In such case `logLevel` is applied before printing any messages to log

---

fixes some more issues mentioned in #3867 

## Types of changes

## Checklist

## Further comments

Output example having only `logLevel` set to silent:

```
Test Suites:     1 passed, 1 total (100% completed)
Time:            🕐  8.02s
```

### DIFF
--- a/packages/wdio-logger/src/node.js
+++ b/packages/wdio-logger/src/node.js
@@ -5,7 +5,7 @@ import chalk from 'chalk'
 import prefix from 'loglevel-plugin-prefix'
 import ansiStrip from 'strip-ansi'
 
-const DEFAULT_LEVEL = 0
+const DEFAULT_LEVEL = 'trace'
 const COLORS = {
     error: 'red',
     warn: 'yellow',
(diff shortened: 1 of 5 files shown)


packages/wdio-logger/src/node.js
```diff
@@ -126,20 +126,36 @@ export default function getLogger (name) {
...
+    Object.keys(loggers).forEach(logName => {
+        const logLevelName = getLogLevelName(logName)
 
-        if (logLevelsConfig[logLevelName] !== undefined) {
-            loggers[loggerName].setLevel(logLevelsConfig[logLevelName])
-        }
+        /**
+         * either apply log level from logLevels object or use global logLevel
+         */
+        const logLevel = logLevelsConfig[logLevelName] !== undefined ? logLevelsConfig[logLevelName] : process.env.WDIO_LOG_LEVEL
```
christian-bromann:
```suggestion
        const logLevel = logLevelsConfig[logLevelName] || process.env.WDIO_LOG_LEVEL
```

mgrybyk: @christian-bromann I'm afraid about case when `logLevelsConfig[logLevelName] === 0`.
In such case this ternary won't work 
`logLevelsConfig[logLevelName] || process.env.WDIO_LOG_LEVEL`

christian-bromann: yeah true, even tho it seems unlikely, how about doing

```
typeof logLevelsConfig[logLevelName] !== 'undefined'
```

I am kind of careful about comparing variables like this, e.g. `[undefined] == '' // true`

xlenz: I'll change to `typeof logLevelsConfig[logLevelName] !== 'undefined'` in the evening

Issue #3927, opened 2019-05-05
Typescript typing for waitUntil method


MunGell:
**Environment (please complete the following information):**
 - **WebdriverIO version:** 5.7.15
 - **Mode:** Standalone mode
 - **Node.js version:** v10.4.1
 - **NPM version:** 6.9.0
 - **Browser name and version:** Chrome
 - **Platform name and version:** Linux
 - **Additional wdio packages used (if applicable):** @wdio/selenium-standalone service

**Config of WebdriverIO**

N/A

**Describe the bug**

At the moment typing for `waitUntil` method is [like this](https://github.com/webdriverio/webdriverio/blob/8dc24c4f4cf4ccf774f2b0c15fc57550a3ba0af4/scripts/templates/webdriverio.tpl.d.ts#L179):

```typescript
waitUntil(
    condition: () => boolean,
    timeout?: number,
    timeoutMsg?: string,
    interval?: number
): void
```

However, in `waitForDisplayed` method it is used [like this](https://github.com/webdriverio/webdriverio/blob/8dc24c4f4cf4ccf774f2b0c15fc57550a3ba0af4/packages/webdriverio/src/commands/element/waitForDisplayed.js#L48):

```
return this.waitUntil(async () => reverse !== (await this.isDisplayed()), ms, errorMsg);
```

In this implementation it returns `Promise<boolean>`, while typing suggests it needs to be just `boolean`.

When I try to implement similar logic in my code - it throws an error.

I believe typing needs to be updated.

**To Reproduce**

N/A

**Expected behavior**

N/A

**Log**

N/A

**Additional context**

N/A

MunGell:
Also, I believe `waitForDisplayed` typing needs to be updated as well:

```typescript
waitForDisplayed(
    ms?: number,
    reverse?: boolean,
    error?: string
): boolean;
```

While it's implementation is returns the output of `waitUntil` method.

`waitUntil` is a void method, so `waitForDisplayed` will return the same

Correct me if I am wrong. Can create a separate issue if needed.

mgrybyk:
Closed via https://github.com/webdriverio/webdriverio/pull/3928

PR #3928, opened 2019-05-05
Change return type waitUntil to boolean

tadashi0713:
## Proposed changes

https://github.com/webdriverio/webdriverio/issues/3927

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/packages/webdriverio/src/commands/browser/waitUntil.js
+++ b/packages/webdriverio/src/commands/browser/waitUntil.js
@@ -29,6 +29,7 @@
  * @param {Number=}  timeout    timeout in ms (default: 500)
  * @param {String=}  timeoutMsg error message to throw when waitUntil times out
  * @param {Number=}  interval   interval between condition checks (default: 500)
+ * @return {Boolean} true if condition is fulfilled
  * @uses utility/pause
  * @type utility
  *
--- a/scripts/templates/webdriverio.tpl.d.ts
+++ b/scripts/templates/webdriverio.tpl.d.ts
@@ -181,7 +181,7 @@ declare namespace WebdriverIO {
             timeout?: number,
             timeoutMsg?: string,
             interval?: number
-        ): void
+        ): boolean
         // ... browser commands ...
     }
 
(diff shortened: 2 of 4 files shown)


tadashi0713:
@CrispusDH Thanks for review!

I fixed according to this issue.
https://github.com/webdriverio/webdriverio/issues/3927

It`s wired that return type of `waitUntil` is void although `waitForDisplayed` or `waitForExisting` return type are boolean (They use waitUntil function as return).

What do you think where we should fix?

CrispusDH:
I think we should firstly discuss it here. I don't remember why it returns `boolean` but it was some circumstances :)

tadashi0713:
Agree
Or maybe we missed `boolean` handling in `waitForDisplayed` and `waitForExist`, although `waitUntil` just returns `void`

packages/webdriverio/src/commands/browser/waitUntil.js
```diff
@@ -29,6 +29,7 @@
  * @param {Number=}  timeout    timeout in ms (default: 500)
  * @param {String=}  timeoutMsg error message to throw when waitUntil times out
  * @param {Number=}  interval   interval between condition checks (default: 500)
+ * @return {Boolean} true if condition is fulfilled
```
MunGell: This is false since `timer.catch()` (Promise catch handler) will return undefined in all situations

mgrybyk:
boolean return type looks good to me.

waitFor* and waitUntil command either returns true or throws an exception.

MunGell:
@mgrybyk but it is **not** what it returns at the moment.

At the moment, it returns `undefined`, see here: https://github.com/webdriverio/webdriverio/blob/master/packages/webdriverio/src/commands/browser/waitUntil.js

Even if we will merge this PR it will have incorrect typings.

mgrybyk:
@MunGell just tried it, returns true

MunGell:
@mgrybyk you are right, just looked at the code again - it returns the result of the `fn` function that is passed as an argument to the method.

mgrybyk:
Thank you, @tadashi0713

Issue #3869, opened 2019-04-19
Don't see console.log('hello') from onComplete hook in the console


prathameshnyt:
**Environment (please complete the following information):**
 - **WebdriverIO version:** 5.7.8
- **Mode:** WDIO Testrunner
 - **Node.js version:** v10.15.2
 - **NPM version:** 6.4.1
 - **Browser name and version:** Chrome 73
 - **Platform name and version:** Mac Os 10.14.4 (18E226

**Config of WebdriverIO**
An example of how you configured WebdriverIO config

**Describe the bug**
As discussed on Gitter, creating issue for us to further followup. I don't see console.log('hello') from onComplete hook message in the console.

**To Reproduce**
Add console.log('hello') in onComplete hook. If I console.log(config) then sometimes it does output it in the console, but not the complete object.

christian-bromann:
@prathameshnyt can you provide a reproducible example?

prathameshnyt:
Thanks @christian-bromann I will look into best way to send you an example. In meantime, I have basicTest from the the guide and running it with test runner with following in onComplete hook. @WillBrock was able to reproduce it as well.

```
    onComplete: function(exitCode, config, capabilities, results) {
        console.log('Hello');
    },
```

christian-bromann:
Was able to reproduce it too

prathameshnyt:
Ok. Thanks for confirming.

mariobianco:
Not sure if this helps provide further info, but I narrowed down the bug to being related to `@wdio/cli version 5.7.9+`. Specifically the bug first showed up in 5.7.9 perhaps something related to [this PR](https://github.com/webdriverio/webdriverio/pull/3638).

baruchvlz:
Any log before [this line](https://github.com/webdriverio/webdriverio/blob/master/packages/wdio-cli/src/launcher.js#L83) works. It seems that [runMode](https://github.com/webdriverio/webdriverio/blob/master/packages/wdio-cli/src/launcher.js#L99) is not resolving / rejecting in some cases. I have added two logs, [here](https://github.com/baruchvlz/webdriverio/blob/fix/oncomplete/packages/wdio-cli/src/launcher.js#L84) and [here](https://github.com/baruchvlz/webdriverio/blob/fix/oncomplete/packages/wdio-cli/src/launcher.js#L156), which are never executed.


Interestingly enough, this onComplete executes and logs.:
```
onComplete: function() {
  console.log('onComplete')
[...]

Issue #3826, opened 2019-04-08
Logs not outputting in real-time + duplication at end 


klamping:
**Environment (please complete the following information):**
 - **WebdriverIO version:** 5.7.10
 - **Mode:** Testrunner
 - **If WDIO Testrunner, running sync/async:** sync
 - **Node.js version:** v8.11.3
 - **NPM version:** 6.8.0
 - **Browser name and version:** Chrome 73.0.3683.86/Firefox 65.0.1
 - **Platform name and version:** OSX 10.13.6
 - **Additional wdio packages used (if applicable):**
```
"@wdio/local-runner": "^5.7.9",
"@wdio/mocha-framework": "^5.7.8",
"@wdio/selenium-standalone-service": "^5.7.8",
"@wdio/spec-reporter": "^5.7.9",
"@wdio/sync": "^5.7.9"
```

**Config of WebdriverIO**
https://github.com/klamping/wdio-log-issue/blob/master/wdio.conf.js

**Describe the bug**
Two issues:
1. Most of the messages don't log until the end
2. I'm still getting some duplication at the end

![](https://thumbs.gfycat.com/OptimalAngelicAardwolf-size_restricted.gif)
Full version: https://gfycat.com/ifr/OptimalAngelicAardwolf

**To Reproduce**
1. Download/Clone https://github.com/klamping/wdio-log-issue
2. `npm install`
3. `npx wdio`

**Expected behavior**
Logs output as steps are completed, no duplicate logs at the end

**Additional context**
Tried this using iTerm 2, Build 3.2.8, run through oh-my-zsh. Also was able to duplicate the issue using just Mac's Terminal

Related PR
https://github.com/webdriverio/webdriverio/pull/3638

vgrigoruk:
Output looks very bad when you run tests on CircleCI. There is a lot of duplication and `CI=true` env variable is not respected for some reason.
See example job here: https://circleci.com/gh/vgrigoruk/wdio-circleci-example/4
Source code is here: https://github.com/vgrigoruk/wdio-circleci-example

christian-bromann:
I guess we need to check CI using https://github.com/watson/ci-info#ci-info

SubJunk:
Just to add to the comment about Circle-CI, my builds fail on Travis recently because of the logspam. Travis ends with:

`The job exceeded the maximum log length, and has been terminated.`

It started happening sometime between 5.4 (which is the version I'm successfully still using) and 5.7

[...]

PR #3950, opened 2019-05-08
new logger - initial version

mgrybyk:
## Proposed changes

Removed usage of **log-update**

image

**Videos**
- [single instance](https://drive.google.com/file/d/1ahxkqm8GDM5uObuN6UMjeRYrqK1mjtxr/view?usp=sharing)
- [2 instances](https://drive.google.com/file/d/1456DUehcSRuxl42uGITCCilICNPYT8EW/view?usp=sharing)

should fix these ones: #3826 #3869

## Types of changes

## Checklist

## Further comments

**Added tests**

Styling suggestions are really welcomed!

### DIFF
--- a/packages/wdio-cli/src/interface.js
+++ b/packages/wdio-cli/src/interface.js
@@ -80,18 +80,20 @@ export default class WDIOCLInterface extends EventEmitter {
     }
 
     onJobComplete(cid, job, retries, message) {
-        const details = []
+        const details = [`[${cid}]`, message]
         if (job) {
-            const filename = Array.isArray(job.specs) ? job.specs.join(', ').replace(process.cwd(), '') : ''
-            const retryCount = retries > 0 ? `(${retries} retries)` : ''
-            details.push('in', getRunnerName(job.caps), filename ? `- ${filename}` : '', retryCount)
+            details.push('in', getRunnerName(job.caps), this.getFilenames(job.specs))
         }
+        details.push(retries > 0 ? `(${retries} retries)` : '')
 
-        this.log(
-            `[${cid}]`,
-            message,
-            ...details
-        )
+        this.log(...details)
+    }
+
+    getFilenames(specs = []) {
+        if (specs.length > 0) {
+            return '- ' + specs.join(', ').replace(new RegExp(`${process.cwd()}`, 'g'), '')
+        }
+        return ''
     }
 
     /**
--- a/packages/wdio-cli/src/interface.js
+++ b/packages/wdio-cli/src/interface.js
@@ -9,6 +9,12 @@ const log = logger('@wdio/cli')
 export default class WDIOCLInterface extends EventEmitter {
     constructor (config, specs, totalWorkerCnt, isWatchMode = false) {
         super()
+
+        /**
+         * Colors can be forcibly enabled/disabled with env variable `FORCE_COLOR`
+         * `FORCE_COLOR=1` - forcibly enable colors
+         * `FORCE_COLOR=0` - forcibly disable colors
+         */
         this.hasAnsiSupport = !!chalk.supportsColor.hasBasic
         this.specs = specs
         this.config = config
--- a/packages/wdio-cli/src/interface.js
+++ b/packages/wdio-cli/src/interface.js
@@ -42,11 +42,7 @@ export default class WDIOCLInterface extends EventEmitter {
             /**
              * messages from worker reporters
              */
-            reporter: {},
-            /**
-             * messages from worker itself
-             */
-            worker: {}
+            reporter: {}
         }
 
         this.onStart()
@@ -84,9 +80,11 @@ export default class WDIOCLInterface extends EventEmitter {
         if (job) {
             details.push('in', getRunnerName(job.caps), this.getFilenames(job.specs))
         }
-        details.push(retries > 0 ? `(${retries} retries)` : '')
+        if (retries > 0) {
+            details.push(`(${retries} retries)`)
+        }
 
-        this.log(...details)
+        return this.log(...details)
     }
 
     getFilenames(specs = []) {
@@ -133,11 +131,10 @@ export default class WDIOCLInterface extends EventEmitter {
     /**
      * for testing purposes call console log in a static method
      */
-    /* istanbul ignore next */
     log (...args) {
-        /* istanbul ignore next */
         // eslint-disable-next-line no-console
         console.log(...args)
+        return args
     }
 
     /**
@@ -155,21 +152,21 @@ export default class WDIOCLInterface extends EventEmitter {
             return this.inDebugMode
         }
 
-        if (!event.origin || !this.messages[event.origin]) {
+        if (!event.origin) {
             return log.warn(`Can't identify message from worker: ${JSON.stringify(event)}, ignoring!`)
+        }
+
+        if (event.origin !== 'reporter') {
+            return this.log(event.cid, event.origin, event.name, event.content)
         }
 
         if (!this.messages[event.origin][event.name]) {
             this.messages[event.origin][event.name] = []
         }
 
-        if (event.origin === 'reporter') {
-            this.messages[event.origin][event.name].push(event.content)
-            if (this.isWatchMode) {
-                this.printReporters()
-            }
-        } else {
-            this.log(event.cid, event.origin, event.name, event.content)
+        this.messages[event.origin][event.name].push(event.content)
+        if (this.isWatchMode) {
+            this.printReporters()
         }
     }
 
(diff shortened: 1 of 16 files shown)


mgrybyk:
added videos

abjerstedt:
So we are still not printing the reporters until the very end of the run?

mgrybyk:
Not, but I can change that in one condition or make it configurable. 
As you like.
Currently reporters are printed straight away in watch mode only

caoxu2000:
why the 2nd test '05.ts' results got printed twice?

mgrybyk:
Because it was retried and failed both times

packages/wdio-cli/src/interface.js
```diff
@@ -55,19 +49,58 @@ export default class WDIOCLInterface extends EventEmitter {
...
+        this.onJobComplete(cid, job, retries, chalk.bold.green('PASSED'))
+    }
+
+    onSpecFailure(cid, job, retries) {
+        this.onJobComplete(cid, job, retries, chalk.bold.red('FAILED'))
+    }
+
+    onJobComplete(cid, job, retries, message) {
+        const details = []
+        if (job) {
```
CrispusDH: Do you suggest that `job` could be not exist?

packages/wdio-cli/src/interface.js
```diff
@@ -164,204 +180,37 @@ export default class WDIOCLInterface extends EventEmitter {
...
-         */
-        if (this.jobs.size) {
-            this.display.push('')
-        }
-
-        this.printStdout(10)
-        this.printSummary()
-        this.updateClock()
+        const isFinished = this.jobs.size === 0
+        const shutdownMessage = !isFinished ? 'Ending WebDriver sessions gracefully ...\n' + '(press ctrl+c again to hard kill the runner)' : 'Ended WebDriver sessions gracefully after a SIGINT signal was received!'
```
CrispusDH: It might be better use `isRunning` instead of `!isFinished`

packages/wdio-cli/src/interface.js
mgrybyk: Yeah, there is such test on launcher.test.js, not sure how to reproduce this in real life.
In one of the cases job doesn't exist

packages/wdio-cli/src/interface.js
mgrybyk: I'll check, I have copied most of the logic, like these two lines.

packages/wdio-cli/src/interface.js
mgrybyk: @CrispusDH this test is failing https://github.com/webdriverio/webdriverio/blob/8de7f1a3b12d97282ed4ee2e25e4c3c9a8940ac1/packages/wdio-cli/tests/launcher.test.js#L105

job is undefined in this test both for old and current interface implementations.
It was added by @bennieswart 3 months ago, not sure how to get such situation in real life.

packages/wdio-cli/src/interface.js
mgrybyk: changed to isRunning

packages/wdio-cli/src/interface.js
```diff
@@ -55,19 +48,58 @@ export default class WDIOCLInterface extends EventEmitter {
...
+    onSpecFailure(cid, job, retries) {
+        this.onJobComplete(cid, job, retries, chalk.bold.red('FAILED'))
+    }
+
+    onJobComplete(cid, job, retries, message) {
+        const details = []
+        if (job) {
+            const filename = Array.isArray(job.specs) ? job.specs.join(', ').replace(process.cwd(), '') : ''
+            const retryCount = retries > 0 ? `(${retries} retries)` : ''
+            details.push('in', getRunnerName(job.caps), filename ? `- ${filename}` : '', retryCount)
```
CrispusDH: It might be cleaner to put as argument just `job` variable and change `getRunnerName` to:
```js
export function getRunnerName ({caps = {}}) {
   ...
}
```

packages/wdio-cli/src/interface.js
```diff
@@ -55,19 +48,58 @@ export default class WDIOCLInterface extends EventEmitter {
...
+    }
+
+    onSpecFailure(cid, job, retries) {
+        this.onJobComplete(cid, job, retries, chalk.bold.red('FAILED'))
+    }
+
+    onJobComplete(cid, job, retries, message) {
+        const details = []
+        if (job) {
+            const filename = Array.isArray(job.specs) ? job.specs.join(', ').replace(process.cwd(), '') : ''
```
CrispusDH: I think it makes sense to create a function for getting `filename` and then remove the inline variable completely. In this case we will get cleaner main function and remove condition in line 87.
```js
function filename({specs}) {
   const result = Array.isArray(specs) ? specs.join(', ').replace(process.cwd(), '') : '';
   return result ? `- ${result}` : ''
}
```
after it line 87 would look like:
`details.push('in', getRunnerName(job.caps), filename(job), retryCount)`

erwinheitzman:
> Because it was retried and failed both times.
> 
> Do we need to change this? I'm not sure to be honest because we have really executed that file twice

We used to only show one result though if we want to change this, now would be a good time.

erwinheitzman:
Why are some of the tests skipped? If they are no longer relevant, remove them instead.

mgrybyk:
@erwinheitzman I haven't started working on the tests yet, was just reviewing existing. Want to implement other suggestions first

mgrybyk:
@erwinheitzman result was printed in case of retry before as well. 

image

packages/wdio-cli/src/interface.js
mgrybyk: Can't get what do you mean here. I can't modify getRunnerName function that way. It accepts capabilities object, not an object containing capabilities.
```{
 browserName: 'Chrome'
}```
not
```{
  capabilities: {
     browserName: 'Chrome'
  }
}
```

CrispusDH: but you insert argument like `job.caps`. So, if you will insert `job` object with `caps` property it will work.

CrispusDH:
Would it make a sense to have such algorithm:
1. Run all tests
2. Show list of passed and failed
3. Retry failed tests
4. Show list of passed and failed
....
In this case would be easier to get what tests were retried, what tests were failed twice, etc. Currently, it would be pretty hard to understand. For example you will get at the end such message:
```js
125 passed, 13 retries, 7 failed
```
So, after retry 6 of 13 first failed tests were passed passed and 7 were failed again. But what exactly? How to find it?

bennieswart:
I've been following this with half an ear since I got mentioned so I'm not up to date with the whole context, but I see there is some uncertainty about the retries. It might be of value to review the code here https://github.com/webdriverio/webdriverio/pull/3381, and specifically have a look at the fields added to `onRunnerStart` (`retry`) and `onRunnerEnd` (`retries`) here https://github.com/webdriverio/webdriverio/pull/3381/files?file-filters%5B%5D=.js&file-filters%5B%5D=.md#diff-f9fbbc3907f0d68d7bf074ad66870d03 .

If a runner starts with `retry > 0` then it has previously run and failed. If a runner fails with `retries > 0`, then it will be retried again.

I hope this helps.

bennieswart:
In reponse to @CrispusDH's suggestion, I'd prefer if tests were rescheduled asap upon failure, reason being that we take great care regarding the order of execution of our tests to run longest tests first, so as not to wait for a single long-running test at the end after all other tests have completed. When such a test fails the first time, it should immediately be rescheduled for the same reason.

mgrybyk:
Anyway runner logic is not a part of logger, if you'd like to change flow you can submit another ticket for that

packages/wdio-cli/src/interface.js
mgrybyk: I think we are talking about different things here, getRunnerName doesn't accept job, it accepts capabilities

packages/wdio-cli/src/interface.js
mgrybyk: @CrispusDH please check again

mgrybyk:
What do you think regarding adding timestamp to workers output?

image

mgrybyk:
PR updated, removed WIP label, added tests.

packages/wdio-cli/src/interface.js
```diff
@@ -1,31 +1,22 @@
...
         this.config = config
         this.totalWorkerCnt = totalWorkerCnt
-        this.sigintTriggered = false
         this.isWatchMode = false
         this.inDebugMode = false
-        this.stdout = stdout
-        this.stdoutBuffer = []
-        this.stderr = stderr
-        this.stderrBuffer = []
+        this.specFileRetries = config.specFileRetries || 0
```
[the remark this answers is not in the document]
BorisOsipov: Yes in runner configuration docs

christian-bromann:
@mgrybyk will test this today

mgrybyk:
Tested logging on Jenkins and it works fine! In pipelines colored logs are shown in realtime without duplications. Same for regular jobs, but logs are not colored by default, added docs on how to use `FORCE_COLOR` in case there is some plugin like AnsiColors installed.

Also fixed issues with watch mode.

mgrybyk:
no changes made, just synced with master

christian-bromann:
@mgrybyk can you please rebase on top of master?

@webdriverio/project-committers can I get another 👍 so we can merge this?

mgrybyk:
rebasing in 10 minutes

mgrybyk:
@christian-bromann rebased, can be merged!

Issue #3942, opened 2019-05-07
Configuration inconsistencies


MunGell:
**Describe your question with as much detail as possible**

Based on the [`webdriverio.d.ts`](https://github.com/webdriverio/webdriverio/blob/a2ca188b62356b30bfe088da28385413fc0ba72c/packages/webdriverio/webdriverio.d.ts#L36) file, `remote()` function expects `WebDriver.Options` interface object to be passed to it as the first argument.

In the body of the `remote()` function we do validation of the received options/configuration against [`WDIO_DEFAULTS`](https://github.com/webdriverio/webdriverio/blob/a2ca188b62356b30bfe088da28385413fc0ba72c/packages/webdriverio/src/constants.js#L31) constant, which describes object structure of [`WebdriverIO.Options` ](https://github.com/webdriverio/webdriverio/blob/a2ca188b62356b30bfe088da28385413fc0ba72c/scripts/templates/webdriverio.tpl.d.ts#L46) interface.

However, later in the `remote()` code we [pass unfiltered `params` object](https://github.com/webdriverio/webdriverio/blob/a2ca188b62356b30bfe088da28385413fc0ba72c/packages/webdriverio/src/index.js#L36) down to [`WebDriver.newSession`](https://github.com/webdriverio/webdriverio/blob/a2ca188b62356b30bfe088da28385413fc0ba72c/packages/webdriver/src/index.js#L17).

Typings suggest (and code confirms) that `WebDriver.newSession` expects object of type `WebDriver.Options` on its input.

It sounds like the only way to make it work is to pass merged `WebdriverIO.Options & WebDriver.Options` object, from which each method will pick its own configuration.

Should the typings be updated or should we consider re-thinking this piece of code?

**Environment (please complete the following information):**
 - **WebdriverIO version:** master branch on 8 May 2019
 - **Mode:** Standalone mode

**Additional context**

Answer to this question will help fixing `attach()` function type definition.
Based on the code, the right definition at the moment is:

```typescript
interface AttachSessionOptions extends WebDriver.Options {
    sessionId: string;
    isW3C?: boolean;
}
```

CC @christian-bromann

PR #3953, opened 2019-05-09
webdriverio: Update types for attach and remote methods

MunGell:
## Proposed changes

This is a proposal to address some part of #3942

## Types of changes

## Checklist

- [ ] **N/A** I have added tests that prove my fix is effective or that my feature works
- [ ] **N/A** I have added necessary documentation (if appropriate)

## Further comments

This PR contains solution where we suggest passing merged options (WebdriverIO and WebDriver) to the `remote()` function, since this is what code is expecting on the input.

The other option would be to re-engineer the way we work with `options`, but this most probably will lead to a breaking change in public API.

### DIFF
--- a/packages/webdriverio/webdriverio.d.ts
+++ b/packages/webdriverio/webdriverio.d.ts
@@ -33,12 +33,12 @@ type BrowserAsync = {
 type BrowserStatic = Pick<WebdriverIO.Browser, 'addCommand' | 'options'>;
 declare namespace WebdriverIOAsync {
     function remote(
-        options?: WebDriver.Options,
+        options?: WebDriver.Options & WebdriverIO.Options,
         modifier?: (...args: any[]) => any
     ): BrowserObject;
 
     function attach(
-        options?: WebDriver.Options,
+        options?: WebDriver.AttachSessionOptions,
     ): BrowserObject;
 
     function multiremote(
--- a/scripts/templates/webdriver.tpl.d.ts
+++ b/scripts/templates/webdriver.tpl.d.ts
@@ -232,8 +232,20 @@ declare namespace WebDriver {
         key?: string;
     }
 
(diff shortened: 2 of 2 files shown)


ost12666:
Is this version released? When I install the last version with npm I still get the error here:
    options?: WebDriver.**AttachSessionOptions**

christian-bromann:
@ost12666 this should be implemented

ost12666:
I saw the PR but when I do npm install webdriverio I get 5.8.4 with the problem. Any idea?

node_modules/webdriverio/webdriverio.d.ts:41:29 - error TS2694: Namespace 'WebDriver' has no exported member 'AttachSessionOptions'.

41         options?: WebDriver.AttachSessionOptions,
                               ~~~~~~~~~~~~~~~~~~~~

mgrybyk:
@christian-bromann @ost12666 in order this one to be fixed webdriver package has to be released as well.
We have released webdriverio package only.

ost12666:
So any suggestion what to do for now? I tried to set noEmitOnError to false in tsconfig but it still stops with error.

christian-bromann:
> We have released webdriverio package only.

Yeah I think every change to the TS definition should come with a change of the definition inside the package (adding a comment or anything) so that Lerna detects a change there as well and releases it. I will make sure to point this out moving forward.

ost12666:
thanks! any idea for a work around now since I am kind of stuck...

mgrybyk:
Define your own typings, similar to how it is in webdriver package.
And remove workaround when new version of webdriver package is released

ost12666:
thanks I guess that can work

ost12666:
Could not do that since I am not using WebDriver directly but this does the trick in tsconfig.json

    "skipLibCheck": true,

PR #3957, opened 2019-05-09
Typescript: update wait functions return types

### LINKED ISSUE #3940: Typescript types for waitUntil
 - **WebdriverIO version:** master
 - **Mode:** Standalone
 - **Node.js version:** 10.14.1
 - **NPM version:** 6.9.0
 - **Browser name and version:** Firefox 66.0.3
 - **Platform name and version:** OS/X Mojave 10.14.4

**Describe the bug**

```ts
        waitUntil(
            condition: () => boolean,
            timeout?: number,
            timeoutMsg?: string,
            interval?: number
        ): boolean
```

`condition` here is a function that returns a boolean, but in order for this to be useful in standalone mode, I want to write a function like:

```js
browser.waitUntil(
            function() {
                return browser.execute("return document.readyState;")
                    .then(function(result) {
                        state = result.value;
                        return result.value === 'complete';
                    });
            },
            ms,
            `Page did not load after ${ms} - ${state}`
        );
```

So this should probably be:

```ts
waitUntil(
    condition: () => boolean | Promise<boolean>,
    timeout?: number,
    timeoutMsg?: string,
    interval?: number
): boolean
```

Also, why is the return type `boolean` instead of `Promise<boolean>`?  Or, is there perhaps some way to get a different set of types for standalone mode?

MunGell:
## Proposed changes

Update `waitUntil` and all `waitFor*` functions to return `Promise<boolean>` instead of just `boolean`.

Fixes #3940 

## Types of changes

## Checklist

- [ ] **N/A** I have added tests that prove my fix is effective or that my feature works
- [ ] **N/A** I have added necessary documentation (if appropriate)

## Further comments

All these functions return Promise of the callback that is passed to `waitUntil`.

In order to make typing generation work I had to add additional logic to work with incorrect output of `dox` module: `Promise.<boolean>` instead of `Promise<boolean>`.
I have put it in the end of `changeType` function to ensure `Promise` will be capitalised.

Additionally, `forEach` was replaced with `map` as it is more appropriate in this situation.

### DIFF
--- a/packages/webdriverio/src/commands/browser/waitUntil.js
+++ b/packages/webdriverio/src/commands/browser/waitUntil.js
@@ -29,7 +29,7 @@
  * @param {Number=}  timeout    timeout in ms (default: 500)
  * @param {String=}  timeoutMsg error message to throw when waitUntil times out
  * @param {Number=}  interval   interval between condition checks (default: 500)
- * @return {Boolean} true if condition is fulfilled
+ * @return {Promise<boolean>} true if condition is fulfilled
  * @uses utility/pause
  * @type utility
  *
--- a/packages/webdriverio/src/commands/element/waitForDisplayed.js
+++ b/packages/webdriverio/src/commands/element/waitForDisplayed.js
@@ -28,7 +28,7 @@
  * @param {Number=}  ms       time in ms (default: 500)
  * @param {Boolean=} reverse  if true it waits for the opposite (default: false)
  * @param {String=}  error    if exists it overrides the default error message
- * @return {Boolean} true     if element is displayed (or doesn't if flag is set)
+ * @return {Promise<boolean>} true if element is displayed (or doesn't if flag is set)
  * @uses utility/waitUntil, state/isDisplayed
  * @type utility
  *
--- a/packages/webdriverio/src/commands/element/waitForEnabled.js
+++ b/packages/webdriverio/src/commands/element/waitForEnabled.js
@@ -27,7 +27,7 @@
  * @param {Number=}  ms       time in ms (default: 500)
  * @param {Boolean=} reverse  if true it waits for the opposite (default: false)
  * @param {String=}  error    if exists it overrides the default error message
- * @return {Boolean} true     if element is (dis/en)abled
+ * @return {Promise<boolean>} true if element is (dis/en)abled
  * @uses utility/waitUntil, state/isEnabled
  * @type utility
  *
--- a/packages/webdriverio/src/commands/element/waitForExist.js
+++ b/packages/webdriverio/src/commands/element/waitForExist.js
@@ -29,7 +29,7 @@
  * @param {Number=}  ms       time in ms (default: 500)
  * @param {Boolean=} reverse  if true it instead waits for the selector to not match any elements (default: false)
  * @param {String=}  error    if exists it overrides the default error message
- * @return {Boolean} true     if element exists (or doesn't if flag is set)
+ * @return {Promise<boolean>} true if element exists (or doesn't if flag is set)
  * @uses utility/waitUntil, state/isExisting
  * @type utility
  *
--- a/scripts/templates/webdriverio.tpl.d.ts
+++ b/scripts/templates/webdriverio.tpl.d.ts
@@ -181,7 +181,7 @@ declare namespace WebdriverIO {
             timeout?: number,
             timeoutMsg?: string,
             interval?: number
-        ): boolean
+        ): Promise<boolean>
         // ... browser commands ...
     }
 
--- a/scripts/type-generation/generate-typings-utils.js
+++ b/scripts/type-generation/generate-typings-utils.js
@@ -22,10 +22,7 @@ const changeType = (text) => {
 }
 
 const getTypes = (types, alwaysType) => {
-    types.forEach((type, index, array) => {
-        array[index] = changeType(type)
-    })
-    types = types.join(' | ')
+    types = types.map(changeType).join(' | ')
     if (types === '' && !alwaysType) {
         types = 'void'
     } else if (types === '*' || (types === '' && alwaysType)) {
--- a/scripts/type-generation/generate-typings-utils.js
+++ b/scripts/type-generation/generate-typings-utils.js
@@ -16,8 +16,12 @@ const changeType = (text) => {
     default: {
         text = text.toLowerCase()
     }
-
     }
+
+    if (text.indexOf('promise.<') === 0) {
+        text = text.replace('promise.', 'Promise')
+    }
+
     return text
 }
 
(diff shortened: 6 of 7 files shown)


packages/webdriverio/src/commands/element/waitForExist.js
```diff
@@ -29,7 +29,7 @@
  * @param {Number=}  ms       time in ms (default: 500)
  * @param {Boolean=} reverse  if true it instead waits for the selector to not match any elements (default: false)
  * @param {String=}  error    if exists it overrides the default error message
- * @return {Boolean} true     if element exists (or doesn't if flag is set)
+ * @return {Promise<boolean>} true if element exists (or doesn't if flag is set)
```
christian-bromann: This is not true .. it always depends if you run WebdriverIO sync or asynchronously, We should not have this here.

packages/webdriverio/src/commands/browser/waitUntil.js
```diff
@@ -29,7 +29,7 @@
  * @param {Number=}  timeout    timeout in ms (default: 500)
  * @param {String=}  timeoutMsg error message to throw when waitUntil times out
  * @param {Number=}  interval   interval between condition checks (default: 500)
- * @return {Boolean} true if condition is fulfilled
+ * @return {Promise<boolean>} true if condition is fulfilled
```
mgrybyk: it should be boolean, not promises here

packages/webdriverio/src/commands/element/waitForDisplayed.js
```diff
@@ -28,7 +28,7 @@
  * @param {Number=}  ms       time in ms (default: 500)
  * @param {Boolean=} reverse  if true it waits for the opposite (default: false)
  * @param {String=}  error    if exists it overrides the default error message
- * @return {Boolean} true     if element is displayed (or doesn't if flag is set)
+ * @return {Promise<boolean>} true if element is displayed (or doesn't if flag is set)
```
mgrybyk: same

packages/webdriverio/src/commands/element/waitForEnabled.js
```diff
@@ -27,7 +27,7 @@
  * @param {Number=}  ms       time in ms (default: 500)
  * @param {Boolean=} reverse  if true it waits for the opposite (default: false)
  * @param {String=}  error    if exists it overrides the default error message
- * @return {Boolean} true     if element is (dis/en)abled
+ * @return {Promise<boolean>} true if element is (dis/en)abled
```
mgrybyk: same

packages/webdriverio/src/commands/element/waitForExist.js
mgrybyk: same

scripts/templates/webdriverio.tpl.d.ts
```diff
@@ -181,7 +181,7 @@ declare namespace WebdriverIO {
             timeout?: number,
             timeoutMsg?: string,
             interval?: number
-        ): boolean
+        ): Promise<boolean>
```
mgrybyk: please don't change this here, it is already wrapped with promise

scripts/type-generation/generate-typings-utils.js
```diff
@@ -16,16 +16,17 @@ const changeType = (text) => {
     default: {
         text = text.toLowerCase()
     }
+    }
 
+    if (text.indexOf('promise.<') === 0) {
```
mgrybyk: please don't do any promise related changes here

tests/typings/sync/sync.ts
```diff
@@ -2,7 +2,7 @@ import allure from '@wdio/allure-reporter'
 
 // browser
 browser.pause(1)
-const waitUntil: boolean = browser.waitUntil(() => true, 1, '', 1)
+const waitUntil: Promise<boolean> = browser.waitUntil(() => true, 1, '', 1)
```
mgrybyk: sync mode has nothing to do with promises!

MunGell: Got it, need to figure out the dark magic of sync/async conversion in this library.

At the moment it is clear to me that this function is asynchronous, but I see that you use Fibers to steam line it for test-runners.

Is sync version only applicable to test runner and standalone is always async?

scripts/type-generation/generate-typings-utils.js
MunGell: Mhm, regardless of the outcome of sync/async discussion, I believe this code is important to make Typescript typing work for promises (any of them, really).

Why do you think I should not change anything here?

MunGell:
Thanks for the offer @mgrybyk , I will ping you privately in Gitter

mgrybyk:
Great, I'll help with direction really soon!

mgrybyk:
Please see gitter for more details, hope that helps!

CrispusDH:
@MunGell do you have some additional questions?

scripts/type-generation/generate-typings-utils.js
MunGell: @CrispusDH this question is still relevant, not sure why this function should not be touched

mgrybyk: As we have already discussed, you should not do any changes in generate typings function

MunGell: HI @mgrybyk 

We indeed have discussed that typings should be changed and splitter between the two packages, but we haven't touched the topic of *why* code that generates incorrect Typescript typings should not be touched.

I believe you might be confused with the purpose of this change, it is partially described here: https://github.com/webdriverio/webdriverio/pull/3957/commits/266bedde2efc99353d42982b515a6e9376a94d37

At the moment our automatic typing generation scripts will generate `promise.<string>` type for Promises, which is not correct way of defining promise type in Typescript. The issue comes from the libraries we depend on (see commit message) and looks like it is an intentional behaviour based on fixtures.

For us to be able to generate correct typings for functions that return promises we need the change I've made here.

Could you please explain *why* should I not change this code?

Thanks!

mgrybyk: Can you please provide an example? 
Secondly, we can have another PR to fix typings related issues, this one is about waitUntil, and there is nothing to be changed in typing generation for it.

MunGell: Yes, of course, as I said this function without my fix will generate incorrect Promise typing like this: `Promise.<script>` (notice dot in the middle).
Feel free to switch to my branch, revert this particular change and run `npm run generate:typings`. Check the generated `d.ts` file and how it has `Promise.<T>` return types.

Otherwise, in master branch of this repository change return type for any function in jsdoc to `Promise<T>` and see the output in the generated `d.ts` file.

Change in how we are handling type generation is required for fixing `waitUntil()` typing issue. I don't think your suggestion to split it into 2 pull request is a good idea as it will create sequential interdependency between the PRs.

Both commit message and PR description are transparent on this change and the motivation for it.

Hope that explains the issue.

mgrybyk: I can't see such problem, can you please provide an example? Which commands have such issue?

MunGell: @mgrybyk I think I've exhausted my ability to explain the issue.  If you can't see the problem feel free to update this PR as required or decline it otherwise.

mgrybyk: Specific example would be really helpful, sorry for exhausting :+1:

mgrybyk:
@MunGell thank you for your suggestion, it would be really great to specify exactly what are you going to fix to provide better visibility. Feel free to raise another issue for that.

Closing this PR for now because of https://github.com/webdriverio/webdriverio/pull/4009 that resolves waitUntil issue in particular.

PR #3958, opened 2019-05-09
fix multiremote instances spawning

### LINKED ISSUE #2825: Multiremote support for multiple specs
## The problem

Missing feature of being able to run multiremote tests in locally and cloud services like Saucelabs or Browserstack correctly.  When you have more than one spec defined as multiremote test then you want Saucelabs or BrowserStack to spin up new VMs per spec. Currently it uses the same VM for all the multiremote specs, and soon exhausts itself.  This is seen locally also.

## Environment

* WebdriverIO version: 4.13.1 and v5
* Node.js version: 8.6.0
* [Standalone mode or wdio testrunner](http://webdriver.io/guide/getstarted/modes.html): wdio
* if wdio testrunner, running synchronous or asynchronous tests: sync
* Additional wdio packages used (if applicable):  `wdio-sauce-service`

## Details

Run multiremote tests (more than 1 spec) on saucelabs or browserstack and the wdio launcher will use the initial pair (or however many you declare) of VMs across all your tests.  It should dispose of the VMs at the end of the test and create new ones by default.  It is not expected to run the tests in parallel so this is not required (maxInstances support).  Looking at `launcher.js` multiremote tests will only every run one `processNumber` with a single `cid` despite different tests starting.  The process count remains the same.

N/A

Run the example multiremote in saucelabs from wdio runner, but duplicate the test to have more than 1 multiremote spec to run.

mgrybyk:
## Proposed changes

Change way how multiremote mode spawns instances.

**Current behaviour**
- all the spec files are running in same worker. This causing issues with reporters and limits spawning of instances (so parallelization works bad)
- instances are not killed in the very end, browsers are hanging opened

**New behaviour**
- another worker spawned for each spec file
- spec files can run in parallel
- browser instances are killed after every spec file (same as in non multiremote mode)

fixes #2825

## Types of changes

## Checklist

### DIFF
--- a/packages/wdio-cli/src/launcher.js
+++ b/packages/wdio-cli/src/launcher.js
@@ -98,19 +98,9 @@ class Launcher {
      */
     runMode (config, caps) {
         /**
-         * if it is an object run multiremote test
-         */
-        if (this.isMultiremote) {
-            return new Promise((resolve) => {
-                this.resolve = resolve
-                this.startInstance(this.configParser.getSpecs(), caps, 0)
-            })
-        }
-
-        /**
          * fail if no caps were found
          */
-        if (!caps || !caps.length) {
+        if (!caps || (!this.isMultiremote && !caps.length)) {
             return new Promise((resolve) => {
                 log.error('Missing capabilities, exiting with failure')
                 this.interface.updateView()
@@ -122,15 +112,31 @@ class Launcher {
          * schedule test runs
          */
         let cid = 0
-        for (let capabilities of caps) {
+        if (this.isMultiremote) {
+            /**
+             * Multiremote mode
+             */
             this.schedule.push({
                 cid: cid++,
-                caps: capabilities,
-                specs: this.configParser.getSpecs(capabilities.specs, capabilities.exclude).map(s => ({ files: [s], retries: config.specFileRetries })),
-                availableInstances: capabilities.maxInstances || config.maxInstancesPerCapability,
-                runningInstances: 0,
-                seleniumServer: { hostname: config.hostname, port: config.port, protocol: config.protocol }
+                caps,
+                specs: this.configParser.getSpecs(caps.specs, caps.exclude).map(s => ({ files: [s], retries: config.specFileRetries })),
+                availableInstances: config.maxInstances || 1,
+                runningInstances: 0
             })
+        } else {
+            /**
+             * Regular mode
+             */
+            for (let capabilities of caps) {
+                this.schedule.push({
+                    cid: cid++,
+                    caps: capabilities,
+                    specs: this.configParser.getSpecs(capabilities.specs, capabilities.exclude).map(s => ({ files: [s], retries: config.specFileRetries })),
+                    availableInstances: capabilities.maxInstances || config.maxInstancesPerCapability,
+                    runningInstances: 0,
+                    seleniumServer: { hostname: config.hostname, port: config.port, protocol: config.protocol }
+                })
+            }
         }
 
         return new Promise((resolve) => {
@@ -342,16 +348,16 @@ class Launcher {
         }
         this.interface.emit('job:end', { cid, passed, retries })
 
-        // Update schedule now this process has ended
-        if (!this.isMultiremote) {
-            // get cid (capability id) from rid (runner id)
-            cid = parseInt(cid, 10)
+        /**
+         * Update schedule now this process has ended
+         */
+        // get cid (capability id) from rid (runner id)
+        cid = parseInt(cid, 10)
 
-            this.schedule[cid].availableInstances++
-            this.schedule[cid].runningInstances--
-        }
+        this.schedule[cid].availableInstances++
+        this.schedule[cid].runningInstances--
 
-        if (!this.isMultiremote && !this.runSpecs()) {
+        if (!this.runSpecs()) {
             return
         }
 
--- a/packages/wdio-runner/src/index.js
+++ b/packages/wdio-runner/src/index.js
@@ -270,11 +270,12 @@ export default class Runner extends EventEmitter {
      * within a hook by the user
      */
     async endSession (shutdown) {
+        const hasSessionId = global.browser && (this.isMultiremote ? global.browser.instances.some(i => global.browser[i] && global.browser[i].sessionId) : global.browser.sessionId)
         /**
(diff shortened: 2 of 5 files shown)


packages/wdio-cli/src/launcher.js
```diff
@@ -97,20 +97,10 @@ class Launcher {
...
-                this.resolve = resolve
-                this.startInstance(this.configParser.getSpecs(), caps, 0)
-            })
-        }
-
         /**
          * fail if no caps were found
          */
-        if (!caps || !caps.length) {
+        if (!caps || (!this.isMultiremote && !caps.length)) {
```
erwinheitzman: can be simplified to:
`if (!caps || !caps.length)`

packages/wdio-cli/src/launcher.js
```diff
@@ -122,15 +112,31 @@ class Launcher {
...
+             * Multiremote mode
+             */
             this.schedule.push({
                 cid: cid++,
-                caps: capabilities,
-                specs: this.configParser.getSpecs(capabilities.specs, capabilities.exclude).map(s => ({ files: [s], retries: config.specFileRetries })),
-                availableInstances: capabilities.maxInstances || config.maxInstancesPerCapability,
-                runningInstances: 0,
-                seleniumServer: { hostname: config.hostname, port: config.port, protocol: config.protocol }
+                caps,
```
erwinheitzman: This is now the object containing all configured browsers, is this what we need here?

packages/wdio-cli/src/launcher.js
```diff
@@ -122,15 +112,31 @@ class Launcher {
...
+                specs: this.configParser.getSpecs(caps.specs, caps.exclude).map(s => ({ files: [s], retries: config.specFileRetries })),
+                availableInstances: config.maxInstances || 1,
```
erwinheitzman: Why 1 instead of maxInstancesPerCapability?

packages/wdio-cli/src/launcher.js
```diff
@@ -122,15 +112,31 @@ class Launcher {
...
+            /**
+             * Multiremote mode
+             */
             this.schedule.push({
                 cid: cid++,
-                caps: capabilities,
-                specs: this.configParser.getSpecs(capabilities.specs, capabilities.exclude).map(s => ({ files: [s], retries: config.specFileRetries })),
-                availableInstances: capabilities.maxInstances || config.maxInstancesPerCapability,
-                runningInstances: 0,
-                seleniumServer: { hostname: config.hostname, port: config.port, protocol: config.protocol }
```
erwinheitzman: This line isn't needed for multi remote?

packages/wdio-cli/src/launcher.js
```diff
@@ -342,16 +348,16 @@ class Launcher {
         }
         this.interface.emit('job:end', { cid, passed, retries })
 
-        // Update schedule now this process has ended
-        if (!this.isMultiremote) {
-            // get cid (capability id) from rid (runner id)
-            cid = parseInt(cid, 10)
+        /**
```
erwinheitzman: Seems like the code below will now get executed for both multi remote as well as regular mode, is this correct?

packages/wdio-runner/src/index.js
```diff
@@ -270,26 +270,45 @@ export default class Runner extends EventEmitter {
      * within a hook by the user
      */
     async endSession (shutdown) {
+        /**
+         * For Multiremote - make sure that all the instances have sessionId
+         */
+        const hasSessionId = global.browser && (this.isMultiremote ? !global.browser.instances.some(i => global.browser[i] && !global.browser[i].sessionId) : global.browser.sessionId)
```
erwinheitzman: Pretty long line, perhaps divide over more lines for readability.

packages/wdio-cli/src/launcher.js
mgrybyk: `caps` is an object in Multiremote, while in regular mode it is an array.

Before current PR this block was unreachable in Multiremote, so have to add this condition

packages/wdio-cli/src/launcher.js
mgrybyk: Before it was allowed to have only 1 instance in multiremote, even maxInstances was ignored.
I've noticed that maxInstances value is 100 and was afraid it might brake something in Multiremote, so I decided to use 1 instead of maxInstancesPerCapability

I can change it to something like `maxInstancesPerCapability / (Object.keys(caps).length || 1)` to avoid spawning to much browser instances in Multiremote. Not sure if it makes sense to do that.

packages/wdio-cli/src/launcher.js
mgrybyk: It was not used before, most likely it is not needed because this line might differ per Multiremote instance

packages/wdio-cli/src/launcher.js
mgrybyk: absolutely! There is no more difference here now between regular and multiremote modes

packages/wdio-runner/src/index.js
mgrybyk: agree, fixing

packages/wdio-cli/src/launcher.js
mgrybyk: yes, that is how multiremote works

packages/wdio-cli/src/launcher.js
erwinheitzman: I can only find a few references to maxInstancesPerCapability, this actually seems deprecated: https://webdriver.io/docs/options.html

packages/wdio-runner/src/index.js
mgrybyk: @erwinheitzman fixed!

packages/wdio-cli/src/launcher.js
mgrybyk: ah sorry, one more comment, it makes no sense to use `maxInstancesPerCapability`, we should only use `maxInstances`, so that's fine.
Sorry for confusing, `maxInstancesPerCapability` only makes sense in regular mode

mgrybyk:
@erwinheitzman do you have some more comments?

mgrybyk:
@christian-bromann can we merge this one?

PR #3964, opened 2019-05-10
webdriverio: Modify findStrategy.js to work with Youi.tv selectors.

richmond998999:
Our team currently works with Youi.tv, but some of the selectors used by their [driver](https://github.com/YOU-i-Labs/appium-youiengine-driver#selector-strategies) is not implemented in webdriverio. I've added a few changes in findStrategy.js to accommodate the selector strategy Youi.tv uses.

You can learn more about You.i TV [here](https://www.youi.tv/).

## Proposed changes

Modify findStrategy.js to work with Youi.tv selectors. 

- Add id selector. (Current implementation of id uses id in css selector)
- Add youi tv CYI class name selectors. 
- Add allowing attribute in name selectors. 
- Add unit test for changes.
- Add documentation for changes

## Types of changes

- [X ] New feature (non-breaking change which adds functionality)

## Checklist

- [X ] I have read the [CONTRIBUTING](https://github.com/webdriverio/webdriverio/blob/master/CONTRIBUTING.md) doc
- [X ] I have added tests that prove my fix is effective or that my feature works
- [X ] I have added necessary documentation (if appropriate)

## Further comments

### DIFF
--- a/docs/Selectors.md
+++ b/docs/Selectors.md
@@ -105,7 +105,7 @@ const classNameAndText = $('<my-element />');
 console.log(classNameAndText.getText()); // outputs: "WebdriverIO is the best"
 ```
 
-## ID
+## id
 
 To find an element using an element using their id, query it by using `id=` in front of the query string (e.g. `id=driver`).
 Note: you can use id in css selectors by using `#` before the id name. See [element with certain text](#element-with-certain-text).
--- a/docs/Selectors.md
+++ b/docs/Selectors.md
@@ -105,6 +105,22 @@ const classNameAndText = $('<my-element />');
 console.log(classNameAndText.getText()); // outputs: "WebdriverIO is the best"
 ```
 
+## ID
+
+To find an element using an element using their id, query it by using `id=` in front of the query string (e.g. `id=driver`).
+Note: you can use id in css selectors by using `#` before the id name. See [element with certain text](#element-with-certain-text).
+
+```html
+<a id="driver" href="https://webdriver.io">WebdriverIO</a>
+```
+
+You can query this element with id by calling:
+
+```js
+const id = $('id=driver');
+console.log(id.getText()); // outputs: "WebdriverIO"
+```
+
 ## xPath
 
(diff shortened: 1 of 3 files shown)


packages/webdriverio/tests/findStrategy.test.js
```diff
@@ -31,6 +31,12 @@ describe('selector strategies helper', () => {
         expect(element.value).toBe('search.input')
     })
 
+    it('should find an element using "name" method with a . in the name', () => {
+        const element = findStrategy('[name="searchinput[@isDisplayed=\'true\']"]')
```
CrispusDH: I don't see `.` symbol in the selector that you put

packages/webdriverio/src/utils/findStrategy.js
```diff
@@ -44,6 +44,10 @@ const defineStrategy = function (selector) {
     if (selector.startsWith('*=')) {
         return 'partial link text'
     }
+    // Use id strategy if the selector starts with id=
+    if (selector.startsWith('id=')) {
```
CrispusDH: Could you share why do you want to add it? If more common would be use `#` symbol.

docs/Selectors.md
```diff
@@ -105,6 +105,22 @@ const classNameAndText = $('<my-element />');
 console.log(classNameAndText.getText()); // outputs: "WebdriverIO is the best"
 ```
 
+## ID
```
CrispusDH: ID => Id

packages/webdriverio/tests/findStrategy.test.js
richmond998999: Sorry, the description should be different.

packages/webdriverio/src/utils/findStrategy.js
richmond998999: The reason for this is when using the `#` symbol, findStrategy uses the css selector to look for the element using id. The youi tv appium driver does not allow using css selector but allows id. Currently, webdriver io does not use id selectors with the $ and $$ function. 

We can change the default id selector to use `#` but I think it would still be better to have both so people can still use `#` to look for id using css selectors and `id=` to look for id using id selector.

docs/Selectors.md
richmond998999: Changed the title from ID to id.

Issue #3976, opened 2019-05-15
queryAppState not working correctly


nextlevelbeard:
**Environment (please complete the following information):**
 - **WebdriverIO version:** 5.8.4
 - **Mode:** WDIO Testrunner
 - **If WDIO Testrunner, running sync/async:** sync
 - **Node.js version:** 10
 - **Browser name and version:** Chrome
 - **Platform name and version:** Android
 - **Appium version:** 1.13.0

**Describe the bug**
The `queryAppState` command is not sending the required parameters due to using an outdated method. It is trying to send a parameters payload for a GET command.
The command was changed from a GET to a POST 
=> https://github.com/appium/appium-base-driver/pull/185

**To Reproduce**
```javascript
    it.only('should return status 4 when invoking queryAppState on Android', function() {
        browser.isAndroid || this.skip();
        browser.url(`https://www.webdriver.io`);
        const state = browser.queryAppState('com.android.chrome');
        expect(state).to.be.eql(4);
    });
```

**Expected behavior**
The command should be executed successfully by sending the correct payload to Appium.
The reproducible test should pass.

**Log**
Using the command `browser.queryAppState('com.android.app')` sends a GET command to Appium with no payload, Appium spits out:
```
[HTTP] --> GET /wd/hub/session/[REDACTED]/appium/device/app_state {}
[debug] [W3C] Bad parameters: BadParametersError: Parameters were incorrect. We wanted {"required":[["appId"],["bundleId"]]} and you sent []
[HTTP] <-- GET /wd/hub/session/[REDACTED]/appium/device/app_state 400 15 ms - 11695 
```

christian-bromann:
Let's continue convo in #3977

PR #3977, opened 2019-05-15
Fix queryAppState protocol

nextlevelbeard:
## Proposed changes

Fixes https://github.com/webdriverio/webdriverio/issues/3976

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/packages/webdriver/protocol/appium.json
+++ b/packages/webdriver/protocol/appium.json
@@ -541,7 +541,7 @@
     }
   },
   "/session/:sessionId/appium/device/app_state": {
-    "GET": {
+    "POST": {
       "command": "queryAppState",
       "description": "Get the given app status on the device",
       "ref": "http://appium.io/docs/en/commands/device/app/app-state/",


nextlevelbeard:
Just noticed but it may be out of date.
A POST method has been added to the JSONWP Specification in https://github.com/appium/appium-base-driver/pull/185.

https://github.com/appium/appium-base-driver/blob/master/lib/protocol/routes.js#L472 .

Could you try to reproduce #3976 ?

mgrybyk:
According to https://github.com/appium/appium-base-driver/blob/master/lib/protocol/routes.js#L472 both **GET** and **POST** methods should be supported. Most likely one of them is outdated and exists for compatibility.

However according to http://appium.io/docs/en/commands/device/app/app-state/ it should be **GET**, which is not necessarily correct because **queryAppState** request has **payload**.

I don't mind changing it from GET to POST if POST works with latest Appium version, which seems to be true.

nextlevelbeard:
> I don't mind changing it from GET to POST if POST works with latest Appium version, which seems to be true.

It seems to be true for me and just to reinforce: this isn't a nitpicking thing like "we should use POST cause I feel like it". By using GET,  the `queryAppState` command _is not working for me right now_, as I've detailed in #3976. An empty payload is sent instead of the `appId`/`bundleId`.

PR #3978, opened 2019-05-15
Fix attach to session w3c handling

### LINKED ISSUE #3933: isW3C options is not working for attaching session
 - **WebdriverIO version:** 5.8.2
 - **Mode:** Standalone mode
 - **Node.js version:** 10.11.0
 - **NPM version:** 6.4.1
 - **Browser name and version:** Firefox 66.0.4
 - **Platform name and version:** macOS Mojave
 - **Additional wdio packages used (if applicable):** nothing special

**Config of WebdriverIO**
Not using any config file. Just a single test case can reproduce the issue.

**Describe the bug**
- When attaching to an existing session, `isW3C` option is ignored.
- Moreover the default value (true) is ignored, so we cannot use some commands like `getWindowHandles()` in firefox or IE11 in the attached session.

**To Reproduce**
- Activate Geckodriver
- Create a new session for firefox
- Create a new client and attach to the existing session with `option.isW3C=true`
- Call any commands that have different specification between Json Wire Protocol and WebDriver Protocol

[Include code or an example repository that can easily be set up]
https://gist.github.com/nkns165/32a92f22ec5e6501c7a0c01ce0becfd9
This test code works with mocha.

**Expected behavior**
The webdriver works with W3C WebDriver Protocol (use `session/:sessionId/window/handles`) 

**Log**
https://gist.github.com/nkns165/1ec42e0dc64bc759ef1e2468dd25a0ba

**Additional context**
I investigated the reason a little. Maybe the argument here should be changed.

https://github.com/webdriverio/webdriverio/blob/3fe38b5d9946372d4243d07dbd8cadf05389b19a/packages/webdriver/src/index.js#L92
```
const prototype = Object.assign(getPrototype({isW3C: options.isW3C}), userPrototype)
```

mgrybyk:
## Proposed changes

Currently `attachToSession` ignores W3C option and assumes that it is always **false**.

This PR fixes it.
`isW3C` default value is now true and `attachToSession` respects isW3C passed value.

## Types of changes

## Checklist

## Further comments

fixes #3933

### DIFF
--- a/packages/webdriver/src/index.js
+++ b/packages/webdriver/src/index.js
@@ -88,8 +88,8 @@ export default class WebDriver {
         logger.setLevel('webdriver', options.logLevel)
 
         options.capabilities = options.capabilities || {}
-        options.isW3C = options.isW3C || true
-        const prototype = Object.assign(getPrototype(options.isW3C), userPrototype)
+        options.isW3C = options.isW3C === false ? false : true
+        const prototype = Object.assign(getPrototype({ isW3C: options.isW3C }), userPrototype)
         const monad = webdriverMonad(options, modifier, prototype)
         return monad(options.sessionId, commandWrapper)
     }
--- a/packages/webdriver/src/index.js
+++ b/packages/webdriver/src/index.js
@@ -71,7 +71,7 @@ export default class WebDriver {
             isChrome: { value: isChrome }
         }
 
-        const protocolCommands = getPrototype({ isW3C, isMobile, isIOS, isAndroid, isChrome, isSauce })
+        const protocolCommands = getPrototype({ isW3C, isChrome, isMobile, isSauce })
         const prototype = merge(protocolCommands, environmentFlags, userPrototype)
         const monad = webdriverMonad(params, modifier, prototype)
         return monad(response.value.sessionId || response.sessionId, commandWrapper)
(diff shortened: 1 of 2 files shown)


packages/webdriver/src/index.js
```diff
@@ -88,8 +88,8 @@ export default class WebDriver {
         logger.setLevel('webdriver', options.logLevel)
 
         options.capabilities = options.capabilities || {}
-        options.isW3C = options.isW3C || true
-        const prototype = Object.assign(getPrototype(options.isW3C), userPrototype)
+        options.isW3C = options.isW3C === false ? false : true
+        const prototype = Object.assign(getPrototype({ isW3C: options.isW3C }), userPrototype)
```
christian-bromann: I think it also makes sense to add the other flags in here too , like `isChrome` , `isSauce` etc, no?

mgrybyk: yeah, `getPrototype` accepts `{ isW3C, isChrome, isMobile, isSauce }`. 
Agree that it makes sense to pass all of them

mgrybyk: @christian-bromann  updated

packages/webdriver/src/index.js
```diff
@@ -88,8 +88,8 @@ export default class WebDriver {
         logger.setLevel('webdriver', options.logLevel)
 
         options.capabilities = options.capabilities || {}
-        options.isW3C = options.isW3C || true
-        const prototype = Object.assign(getPrototype(options.isW3C), userPrototype)
+        options.isW3C = options.isW3C === false ? false : true
+        const prototype = Object.assign(getPrototype({ ...options, isW3C: options.isW3C }), userPrototype)
```
christian-bromann:
```suggestion
        const prototype = Object.assign(getPrototype({ ...options }), userPrototype)
```

mgrybyk: @christian-bromann done

PR #3992, opened 2019-05-20
Updates to Performance Features

christian-bromann:
## Proposed changes

I updated the DevTools service to better collect performance metrics. I removed the Lighthouse code and imported Puppeteer for the DevTools interface interactions. All metrics are much more reliable now and it is all easier to use, e.g.:

```js
const assert = require('assert')

describe('JSON.org page', () => {
    before(() => {
        browser.enablePerformanceAudits()
    })

    it('should load within performance budget', () => {

        browser.url('http://json.org')

        let metrics = browser.getMetrics()
        assert.ok(metrics.speedIndex < 1500) // check that speedIndex is below 1.5ms

        let score = browser.getPerformanceScore() // get Lighthouse Performance score
        assert.ok(score >= .99) // Lighthouse Performance score is at 99% or higher

        $('=Esperanto').click()

        metrics = browser.getMetrics()
        assert.ok(metrics.speedIndex < 1500)
        score = browser.getPerformanceScore()
        assert.ok(score >= .99)
    })

    after(() => {
        browser.disablePerformanceAudits()
    })
})
```

I also added a custom command that allows to use Puppeteer directly:

```js
describe('use Puppeteer', () => {
    it('by wrapping commands with call', () => {
        browser.url('http://json.org')

        const puppeteer = browser.getPuppeteer()
        const page = browser.call(() => puppeteer.browser.pages())[0]
        console.log(browser.call(() => page.title()))
    })

    it('by using async/await', async () => {
        const puppeteer = browser.getPuppeteer()
        const pages = await puppeteer.browser.pages()
        console.log(await pages[0].title())
    })
})
```

## Types of changes

## Checklist

## Further comments

I will follow up with a blogpost describing ways how you can/should set up a test suite with performance budgets etc.


packages/wdio-devtools-service/src/auditor.js
```diff
@@ -0,0 +1,131 @@
...
+
+import Diagnostics from 'lighthouse/lighthouse-core/audits/diagnostics'
+import MainThreadWorkBreakdown from 'lighthouse/lighthouse-core/audits/mainthread-work-breakdown'
+import Metrics from 'lighthouse/lighthouse-core/audits/metrics'
+import TTFBMetric from 'lighthouse/lighthouse-core/audits/time-to-first-byte'
+
+import { quantileAtValue } from './utils'
+import logger from '@wdio/logger'
+
+const log = logger('Auditor')
```
mgrybyk: I think we agreed to name loggers like `@wdio/devtools-service:Auditor`

packages/wdio-devtools-service/src/auditor.js
```diff
@@ -0,0 +1,131 @@
...
+
+/**
+ * metric scoring to calculate Lighthouse Performance Score
+ * contains: [media, falloff]
+ * see https://docs.google.com/spreadsheets/d/1_iDW0B870vZF6dAhf1Icdher0gX_KFxCd6Df0_fZ6do/edit#gid=283330180
```
mgrybyk: Hey,
I don't have access to https://docs.google.com/spreadsheets/d/1_iDW0B870vZF6dAhf1Icdher0gX_KFxCd6Df0_fZ6do/edit#gid=283330180

Can we have some public available url?

packages/wdio-devtools-service/src/auditor.js
```diff
@@ -0,0 +1,131 @@
...
+ */
+const METRIC_SCORING = {
+    FMP: [4000, 1600], // first meaningful paint
```
mgrybyk: Are these FMP, FI, etc described somewhere? (maybe in doc above)

packages/wdio-devtools-service/src/auditor.js
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    _audit (AUDIT, params = {}) {
+        const auditContext = {
+            options: Object.assign({}, AUDIT.defaultOptions),
+            ...SHARED_AUDIT_CONTEXT
+        }
+
+        try {
+            return AUDIT.audit({
```
mgrybyk: styling question (not even suggestion): is it good to have multiple returns this way? Or it is better to have `result` variable, assign value to it and return in the very end?

packages/wdio-devtools-service/src/auditor.js
```diff
@@ -0,0 +1,131 @@
...
+                devtoolsLogs: { defaultPass: this.devtoolsLogs },
+                ...params
+            }, auditContext)
+        } catch (e) {
+            log.error(e)
+            return {}
+        }
+    }
+
+    updateCommands (browser) {
```
mgrybyk: Mind adding comment to explain what is going on here?

packages/wdio-devtools-service/src/auditor.js
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    async getDiagnostics () {
+        const result = await this._audit(Diagnostics)
+        return result.details.items[0]
+    }
+
+    async getMetrics () {
+        const timeToFirstByte = await this._audit(TTFBMetric, { URL: this.url })
+        const result = await this._audit(Metrics)
```
mgrybyk: isn't it posible that `this._audit(Metrics)` return `{}`? In such case next line will fail `result.details.items[0]`, no?

packages/wdio-devtools-service/README.md
```diff
@@ -191,6 +290,28 @@ console.log(browser.getPageWeight())
...
+
+The service uses Puppeteer for its automation under the hood. You can get access to the used instance by calling the `getPuppeteer` command. __Note:__ Puppeteer commands are async and either needs to be called within the `call` command or handled via `async/await`:
+
+```js
+describe('use Puppeteer', () => {
+    it('by wrapping commands with call', () => {
+        browser.url('http://json.org')
+
+        const puppeteer = browser.getPuppeteer()
+        const page = browser.call(() => puppeteer.browser.pages())[0]
```
CrispusDH: why here you use `page` and in async example you use `pages`. I think it would be better to keep one flow in example

packages/wdio-devtools-service/src/auditor.js
```diff
@@ -0,0 +1,131 @@
...
+    constructor (traceLogs, devtoolsLogs) {
+        this.devtoolsLogs = devtoolsLogs
+        this.traceLogs = traceLogs
+        this.url = traceLogs.pageUrl
+        this.loaderId = traceLogs.loaderId
+    }
+
+    _audit (AUDIT, params = {}) {
+        const auditContext = {
+            options: Object.assign({}, AUDIT.defaultOptions),
```
CrispusDH: it might be clearer to use spread syntax `options: {...AUDIT.defaultOptions}`

packages/wdio-devtools-service/src/auditor.js
CrispusDH: I prefer having the `result` variable

packages/wdio-devtools-service/src/auditor.js
```diff
@@ -0,0 +1,131 @@
...
+
+    async getDiagnostics () {
+        const result = await this._audit(Diagnostics)
+        return result.details.items[0]
+    }
+
+    async getMetrics () {
+        const timeToFirstByte = await this._audit(TTFBMetric, { URL: this.url })
+        const result = await this._audit(Metrics)
+        const metrics = result.details.items[0] || {}
```
CrispusDH: Would it work?
```js
const {estimatedInputLatency, observedDomContentLoaded, observedFirstVisualChange, firstContentfulPaint, firstMeaningfulPaint, observedLastVisualChange, firstCPUIdle, interactive, observedLoad, speedIndex} = result.details.items[0] || {}
return {
  estimatedInputLatency,
  Math.round(timeToFirstByte.rawValue, 10),
  domContentLoaded: observedDomContentLoaded,
  ...
  firstContentfulPaint,
  ...
}
```
or maybe:
```js
const metrics = result.details.items[0] || {}
return {
            timeToFirstByte: Math.round(timeToFirstByte.rawValue, 10),
            ...metrics
        }
```

packages/wdio-devtools-service/src/driver.js
```diff
@@ -0,0 +1,65 @@
...
+        this.browser = browser
+
+        /**
+         * ToDo (Christian): handle target changes
+         * this should usually never happens as we only connect within one
+         * target within the VM so maybe we can discard this
+         */
+    }
+
+    static async attach (browserURL) {
```
CrispusDH: Would be understandable just `url` instead of `browserURL`?

CrispusDH:
will look further a little bit later

packages/wdio-devtools-service/src/auditor.js
christian-bromann: made if public, it is a document from google anyway

packages/wdio-devtools-service/src/auditor.js
christian-bromann: they come from the document described above

packages/wdio-devtools-service/src/auditor.js
christian-bromann: as that variable wouldn't be used anyway, I prefer to return directly

packages/wdio-devtools-service/tests/__fixtures__/tracelog.json
```diff
@@ -0,0 +1 @@
...485,"tid":24067,"ts":67663877250,"ph":"X","cat":"devtools.timeline","name":"RunTask","args":{},"dur":14,"tdur":13,"tts":22829},{"pid":61485,"tid":24067,"ts":67663877269,"ph":"X","cat":"devtools.timeline","name":"RunTask","args":{},"dur":33,"tdur":32,"tts":22848},{"pid":61485,"tid":24067,"ts":67663880569,"ph":"X","cat":"devtools.timeline","name":"RunTask","args":{},"dur":45,"tdur":43,"tts":22911},{"pid":61485,"tid":24067,"ts":67663882083,"ph":"X","cat":"devtools.timeline","name":"RunTask","args":{},"dur":35,"tdur":34,"tts":22980},{"pid":61485,"tid":24067,"ts":67663882126,"ph":"X","cat":"devtools.timeline","name":"RunTask","args":{},"dur":10,"tdur":9,"tts":23022},{"pid":61485,"tid":24067,"ts":67663882626,"ph":"X","cat":"devtools.timeline","name":"RunTask","args":{},"dur":34,"tdur":34,"tts":23049},{"pid":61485,"tid":24067,"ts":67663882669,"ph":"X","cat":"devtools.timeline","name":"RunTask","args":{},"dur":16,"tdur":16,"tts":23092},{"pid":61485,"tid":24067,"ts":67663884133,"ph":"X","cat":"devtools.timeline","name":"RunTask","args":{},"dur":26,"tdur":26,"tts":23130},{"pid":61485,"tid":24067,"ts":67663884247,"ph":"X","cat":"devtools.timeline","name":"RunTask","args":{},"dur":56,"tdur":56
```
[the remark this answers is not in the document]
christian-bromann: it doesn't matter if formatted or not .. it is to big to read anyway

packages/wdio-devtools-service/src/utils.js
```diff
@@ -70,3 +72,59 @@ export async function readIOStream (cdp, stream) {
...
+ * @customfunction
+ */
+export function quantileAtValue (median, falloff, value) {
+    const location = Math.log(median)
+
+    // The "falloff" value specified the location of the smaller of the positive
+    // roots of the third derivative of the log-normal CDF. Calculate the shape
+    // parameter in terms of that value and the median.
+    const logRatio = Math.log(falloff / median)
+    const shape = Math.sqrt(1 - 3 * logRatio - Math.sqrt((logRatio - 3) * (logRatio - 3) - 8)) / 2
```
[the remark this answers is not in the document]
christian-bromann: I pretty much copy&pasted the code from the google sheet, did you test this?

packages/wdio-devtools-service/src/auditor.js
christian-bromann: no, the metrics names are sometimes different: `domContentLoaded` vs. `observedDomContentLoaded`

christian-bromann:
@CrispusDH @mgrybyk thanks for reviewing! Fixed issues highlighted by you, can you please take another look?

packages/wdio-devtools-service/src/index.js
```diff
@@ -29,23 +36,133 @@ export default class DevToolsService {
...
+            session.on('Page.frameNavigated', ::this.traceGatherer.onFrameNavigated)
+
+            const page = await this.devtoolsDriver.getActivePage()
+            page.on('requestfailed', ::this.traceGatherer.onFrameLoadFail)
+
+            /**
+             * enable domains for client
+             */
+            await Promise.all(['Page', 'Network', 'Console'].map(
+                (domain) => Promise.all([
```
CrispusDH: it might be clearer to avoid second `Promise.all()`. It would look like:
```js
await Promise.all(['Page', 'Network', 'Console'].map(
   async (domain) => {
       await session.send(`${domain}.enable`)
       this.client[domain]['enable']()
    }
...
```

packages/wdio-devtools-service/src/index.js
```diff
@@ -29,23 +36,133 @@ export default class DevToolsService {
...
+
+        /**
+         * update custom commands once tracing finishes
+         */
+        this.traceGatherer.once('tracingComplete', (traceEvents) => {
+            const auditor = new Auditor(traceEvents, this.devtoolsGatherer.getLogs())
+            auditor.updateCommands(global.browser)
+        })
+
+        return new Promise((resolve) => {
```
CrispusDH: I think using `Promise` directly it is not necessary here

packages/wdio-devtools-service/src/index.js
```diff
@@ -29,23 +36,133 @@ export default class DevToolsService {
...
+        this.cpuThrottling = cpuThrottling
+        this.cacheEnabled = Boolean(cacheEnabled)
+        this.shouldRunPerformanceAudits = true
+    }
+
+    /**
+     * custom command to disable performance audits
+     */
+    _disablePerformanceAudits () {
+        delete this.networkThrottling
```
CrispusDH: It might be better reset these parameters to some default values instead of wiping from `this`

packages/wdio-devtools-service/src/index.js
christian-bromann: This way, it would take twice as long because both of the commands aren't run in parallel.

packages/wdio-devtools-service/src/index.js
christian-bromann: it is, we need to defer the continuation until the tracing has been finished .. this can only be done by returning a promise

packages/wdio-devtools-service/src/index.js
christian-bromann: they are automatically defaulted when enabling performance audits .. I can remove the deletes though

christian-bromann:
Any other comments?

packages/wdio-devtools-service/src/utils.js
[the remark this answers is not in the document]
mgrybyk: yeah, just copied internalErf_ and quantileAtValue functions to browser console and called quantileAtValue like this:
`quantileAtValue(0, 1, 2)` -> NaN // median === 0
`quantileAtValue(1, 1, 1)` -> NaN // falloff => value
`quantileAtValue(0, 1, 2)` -> NaN // falloff >= value

However I'm not sure if it make sense to do something here because I don't know what arguments can be received by `quantileAtValue`

packages/wdio-devtools-service/README.md
```diff
@@ -191,6 +290,28 @@ console.log(browser.getPageWeight())
 // }
 ```
 
+### Access Puppeteer Instance
+
+The service uses Puppeteer for its automation under the hood. You can get access to the used instance by calling the `getPuppeteer` command. __Note:__ Puppeteer commands are async and either needs to be called within the `call` command or handled via `async/await`:
```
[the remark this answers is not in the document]
christian-bromann: That is not possible unfortunately. We could overwrite the whole API but I don't know if that makes much sense .. people should just use it with `async/await`

packages/wdio-devtools-service/src/utils.js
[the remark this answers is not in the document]
christian-bromann: me neither .. as I said, it is copy pasted from the google spreadsheet and I don't think it will cause issues, at least I haven't experienced any

christian-bromann:
@CrispusDH @mgrybyk pushed the test coverage. Are we good now?

Issue #3925, opened 2019-05-04
Specs & exclude configuration under capabilities are ignored and hence not working well with suite and spec command line args.


yamkay:
**Environment:**
 - **WebdriverIO version:** 5.8.1
 - **Mode:** WDIO Testrunner
 - **If WDIO Testrunner, running sync/async:** sync
 - **Node.js version:** 8.11.2
 - **NPM version:** 5.8.0
 - **Additional wdio packages used (if applicable):** wdio-spec-reporter, wdio-allure-reporter, wdio-sauce-service

**Config of WebdriverIO**
```
exports.config = {
    suites: {
        p0: [
            './tests/**/*.p0.test.js'
        ],
       p1: [
            './tests/**/*.p1.test.js'
        ],

    },
    capabilities = [
      {
    browserName: "googlechrome",
    platformName: "macOS 10.13",
    browserVersion: "72.0",
    exclude: ["./tests/mobile/**"]
    },
  {
    browserName: "Chrome",
    appiumVersion: "1.9.1",
    deviceName: "Samsung Galaxy S9 WQHD GoogleAPI Emulator",
    deviceOrientation: "portrait",
    platformVersion: "8.0",
    platformName: "Android",
    exclude: ["./tests/desktop/**"]
  }
]
}
```

**Describe the bug**
Whenever we pass the command line arguments (such as --suite, --spec), the specs & exclude configurations under each capability are ignored.

**To Reproduce**

1. Create a folder "tests"
2. Under tests folder, create 2 folders - desktop, mobile
3. Under desktop folder, add a spec file with the name "file1.p0.test.js"
4. Under mobile folder, add a spec file with the name "file2.p1.test.js"
5. Say you are using the above configuration in a config file called "wdio.config.js"
6. Now run only p0 tests with command line arguments - **wdio wdio.config.js --suite p0**
7. **You can observe that file1.p0.test.js (under desktop folder) will be run on both of the above capabilities though it has been excluded from the 2nd capability.**

**Expected behavior**
If the spec file from a suite/spec, which is passed from command line, is excluded under a capability, then it should not execute on it.

Similarly, if there are specific specs configured under a capability, then only those specs should be run on the same capability without running the other tests.

**Additional context**

If the method "getSpec" in the below file is refactored to the following solution, then I guess it should work. Please confirm.

https://github.com/webdriverio/webdriverio/blob/master/packages/wdio-config/src/lib/ConfigParser.js

**Solution**

```
[...]

PR #3995, opened 2019-05-20
wdio-config: include and exclude at capability level

ergbouncex:
## Proposed changes

Allowing spec include/exclude at the capability level to filter properly while passing multiple suites of tests. Previously the capability include/exclude only worked if no suites were passed.

## Types of changes

## Checklist

## Further comments
Fixes issue #3925

### DIFF
--- a/packages/wdio-config/src/lib/ConfigParser.js
+++ b/packages/wdio-config/src/lib/ConfigParser.js
@@ -162,6 +162,7 @@ export default class ConfigParser {
             for (let suiteName of suites) {
                 // ToDo: log warning if suite was not found
                 let suite = this._config.suites[suiteName]
+
                 if (suite && Array.isArray(suite)) {
                     suiteSpecs = suiteSpecs.concat(ConfigParser.getFilePaths(suite))
                 }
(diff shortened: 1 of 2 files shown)


packages/wdio-config/tests/configparser.test.js
```diff
@@ -102,7 +102,7 @@ describe('ConfigParser', () => {
             const specs = configParser.getSpecs()
             expect(specs).toContain(__filename)
             expect(specs).toContain(INDEX_PATH)
-            expect(specs).toContain(path.join(__dirname, 'validateConfig.test.js'))
+            expect(specs).not.toContain(path.join(__dirname, 'validateConfig.test.js'))
```
mgrybyk: Why do we need to invert this assertion?

packages/wdio-config/src/lib/ConfigParser.js
```diff
@@ -174,9 +175,18 @@ export default class ConfigParser {
...
+                tmpSpecs = tmpSpecs.concat(ConfigParser.getFilePaths(capSpecs))
+            }
+
+            if (Array.isArray(capExclude)) {
+                exclude = exclude.concat(ConfigParser.getFilePaths(capExclude))
+            }
 
-            return [...new Set(tmpSpecs)]
+            specs = [...new Set(tmpSpecs)]
+            return specs.filter(spec => exclude.indexOf(spec) < 0)
```
mgrybyk: Can be replaced with !exclude.inlcudes(spec)

packages/wdio-config/tests/configparser.test.js
ergbouncex: The original assertion was not actually checking the _exclude_ defined within FIXTURES_CONF which has:
```
    exclude: [
        path.join(TEST_ROOT, '/validateConfig.test.js')
    ]
```

packages/wdio-config/src/lib/ConfigParser.js
ergbouncex: Sure. The same syntax is used outside of the suite loop too. I will update both

PR #4009, opened 2019-05-22
Typings: wrap waitUntil condition function return type with Promise

### LINKED ISSUE #3940: Typescript types for waitUntil

mgrybyk:
## Proposed changes

wrap **waitUntil** condition function return type with Promise
`() =>  boolean` -> `() => Promise<boolean>`

fix #3940

## Types of changes

## Checklist

### DIFF
--- a/packages/wdio-sync/webdriverio.d.ts
+++ b/packages/wdio-sync/webdriverio.d.ts
@@ -11,6 +11,15 @@ declare namespace WebdriverIO {
     function multiremote(
         options: WebdriverIO.MultiRemoteOptions
     ): WebDriver.Client;
+
+    interface Browser {
+        waitUntil(
+            condition: () => boolean,
+            timeout?: number,
+            timeoutMsg?: string,
+            interval?: number
+        ): boolean
+    }
 }
 
 declare var browser: BrowserObject;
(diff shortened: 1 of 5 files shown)


CrispusDH:
Looks good. :)

PR #4014, opened 2019-05-24
Simplify getting started

mgrybyk:
## Proposed changes

I'd like to simplify getting started guid as much as possible, please don't be strict :)

- First of all I'm really lazy and want to run `./node_modules/.bin/wdio` and hit _Enter_ button until cli stops bothering me with tricky questions. 

- Secondly it's too boring for me to read in general, especially about downloading drivers and environment variables.

- Finally I don't care about configuring anything in config, install additional packages or stuff like that, I just want my tests run and that's it.

## Types of changes

- [x] Docs

## Checklist

## Further comments

I'm new to creating pages in docs, can anyone please take a look and point what am I doing wrong. Naming suggestions are very welcomed :)

P.S. Looks like there are no tests for `setup.js` and `config.js`. Skipping it for now.

### DIFF
--- a/packages/wdio-cli/src/config.js
+++ b/packages/wdio-cli/src/config.js
@@ -23,17 +23,17 @@ export const SUPPORTED_REPORTER = [
 export const SUPPORTED_SERVICES = [
     ' sauce - https://www.npmjs.com/package/@wdio/sauce-service',
     ' testingbot - https://www.npmjs.com/package/@wdio/testingbot-service',
-    ' firefox-profile - https://www.npmjs.com/package/@wdio/firefox-profile-service',
     ' selenium-standalone - https://www.npmjs.com/package/@wdio/selenium-standalone-service',
+    ' chromedriver - https://www.npmjs.com/package/wdio-chromedriver-service',
     ' devtools - https://www.npmjs.com/package/@wdio/devtools-service',
     ' applitools - https://www.npmjs.com/package/@wdio/applitools-service',
     ' browserstack - https://www.npmjs.com/package/@wdio/browserstack-service',
     ' appium - https://www.npmjs.com/package/@wdio/appium-service',
-    ' chromedriver - https://www.npmjs.com/package/wdio-chromedriver-service',
     ' intercept - https://www.npmjs.com/package/wdio-intercept-service',
     ' zafira-listener - https://www.npmjs.com/package/wdio-zafira-listener-service',
     ' reportportal - https://www.npmjs.com/package/wdio-reportportal-service',
-    ' docker - https://www.npmjs.com/package/wdio-docker-service'
+    ' docker - https://www.npmjs.com/package/wdio-docker-service',
+    ' firefox-profile - https://www.npmjs.com/package/@wdio/firefox-profile-service',
 ]
 
 export const SUPPORTED_RUNNERS = [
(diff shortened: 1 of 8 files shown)


packages/wdio-cli/src/config.js
```diff
@@ -290,7 +291,25 @@ export const QUESTIONNAIRE = [{
     name: 'services',
     message: 'Do you want to add a service to your test setup?',
     choices: SUPPORTED_SERVICES,
-    filter: filterPackageName('service')
+    default: SUPPORTED_SERVICES.filter(service => service.includes('wdio-chromedriver-service')),
+    filter: filterPackageName('service'),
+    validate: (answers) => {
+        let result = true
+        const exclusiveServices = {
```
mgrybyk: Move to constants!

packages/wdio-cli/src/setup.js
```diff
@@ -21,6 +21,9 @@ export default function setup () {
         }
         if (answers.installServices) {
             packagesToInstall = packagesToInstall.concat(answers.services)
+            if (answers.services.includes('wdio-chromedriver-service')) {
```
mgrybyk: I think this should be moved to wdio chromedriver service itself

packages/wdio-cli/src/config.js
```diff
@@ -290,7 +298,19 @@ export const QUESTIONNAIRE = [{
     name: 'services',
     message: 'Do you want to add a service to your test setup?',
     choices: SUPPORTED_SERVICES,
-    filter: filterPackageName('service')
+    default: SUPPORTED_SERVICES.filter(service => service.includes('wdio-chromedriver-service')),
```
christian-bromann: Why not just?
```suggestion
    default: 'wdio-chromedriver-service',
```

packages/wdio-cli/src/config.js
```diff
@@ -278,6 +285,7 @@ export const QUESTIONNAIRE = [{
     name: 'reporters',
     message: 'Which reporter do you want to use?',
     choices: SUPPORTED_REPORTER,
+    default: SUPPORTED_REPORTER.filter(reporter => reporter.includes('spec-reporter')),
```
christian-bromann: same
```suggestion
    default: '@wdio/spec-reporter',
```

packages/wdio-cli/src/config.js
mgrybyk: @christian-bromann that's what I've tried first :)

First of all it should be an array for checkboxes, secondly I have to specify full name which is `' chromedriver - https://www.npmjs.com/package/wdio-chromedriver-service'`

I can change `SUPPORTED_SERVICES.filter(service => service.includes('wdio-chromedriver-service'))` to `' chromedriver - https://www.npmjs.com/package/wdio-chromedriver-service'`

Same for reporters

christian-bromann: ah got it, nah this is fine then

mgrybyk:
@christian-bromann updated sidebars, however I was not sure which section is the best for newly added files and decided to change structure a bit.

Is it legal to do such changes?

packages/wdio-cli/src/setup.js
baruchvlz: This is the correct way to do this, but I why can't `chromedriver` be a dependency of `wdio-chromedriver-service` that way the user doesn't have to manually add it.

PR #4037, opened 2019-05-30
Fix build

christian-bromann:
Our typescript tests fail due to:

```
../../../node_modules/typescript/lib/lib.es5.d.ts:1469:6 - error TS2300: Duplicate identifier 'Omit'.

1469 type Omit<T, K extends keyof any> = Pick<T, Exclude<keyof T, K>>;
          ~~~~

  node_modules/webdriver/webdriver.d.ts:7:6
    7 type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
           ~~~~
    'Omit' was also declared here.

node_modules/webdriver/webdriver.d.ts:7:6 - error TS2300: Duplicate identifier 'Omit'.

7 type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
       ~~~~

  ../../../node_modules/typescript/lib/lib.es5.d.ts:1469:6
    1469 type Omit<T, K extends keyof any> = Pick<T, Exclude<keyof T, K>>;
              ~~~~
    'Omit' was also declared here.


Found 2 errors.
```

This removes the definition of `Omit` which is not used anyway.

### DIFF
--- a/packages/webdriverio/webdriverio.d.ts
+++ b/packages/webdriverio/webdriverio.d.ts
@@ -5,7 +5,7 @@ type $ = (selector: string | Function) => Promise<WebdriverIOAsync.Element>;
 type $$ = (selector: string | Function) => Promise<WebdriverIOAsync.Element[]>;
 
 // Element commands that should be wrapper with Promise
-type ElementPromise = Omit<WebdriverIO.Element, 'addCommand' | '$' | '$$'>;
+type ElementPromise = WdioOmit<WebdriverIO.Element, 'addCommand' | '$' | '$$'>;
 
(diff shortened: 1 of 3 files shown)


ianwremmel:
When will this be released? It's preventing me from upgrading to TypeScript 3.5.

christian-bromann:
I can do a release

timmywil:
The webdriver package itself needs a version bump as well.

mgrybyk:
We may ask @christian-bromann to help with it.

@wdio/sync, webdriver and webdriverio should be released all together in this case.

christian-bromann:
on it

mgrybyk:
Thank you

PR #4049, opened 2019-05-31
Overwrite native commands

mgrybyk:
## Proposed changes

Ability to overwrite native commands.

**Element**
```
browser.overwriteCommand('addValue', function (origAddValue, value, message) {
  console.log(message)
  return origAddValue(value)
}, true)
```

**Browser**
```
browser.overwriteCommand('pause', function (origPause, ms, msg) {
  console.log(`Sleeping for ${ms}`)
  origPause(ms)
})
```

## Types of changes

## Checklist

- [x] I have added unit tests
- [x] I have added smoke tests (_in progress_)
- [x] I have added necessary documentation `CustomCommands.md`
- [x] I have added types and tests for types

### DIFF
--- a/docs/CustomCommands.md
+++ b/docs/CustomCommands.md
@@ -3,6 +3,8 @@ id: customcommands
 title: Custom Commands
 ---
 
+### Adding custom commands
+
 If you want to extend the browser instance with your own set of commands there is a method called `addCommand` available from the browser object. You can write your command in a synchronous (default) way the same way as in your specs or asynchronous (like when using WebdriverIO in standalone mode). The following example shows how to add a new command that returns the current url and title as one result only using synchronous commands:
 
 ```js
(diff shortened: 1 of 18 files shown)


packages/webdriver/src/monad.js
```diff
@@ -68,30 +77,60 @@ export default function WebDriver (options, modifier, propertiesObject = {}) {
...
                     })
                 }
+
                 this.__propertiesObject__[name] = { value: customCommand }
             } else {
                 unit.lift(name, customCommand, proto)
             }
         }
 
+        client.overwriteCommand = function (name, func, attachToElement = false, proto, instances) {
```
CrispusDH: Would be possible to overwrite e.g. `click()`? And use it like `element.click()`. Or it is just for `BrowserObject`?

mgrybyk:
@CrispusDH yeah, it should be possible to overwrite browser and element level commands,  including `click`

docs/CustomCommands.md
```diff
@@ -96,3 +98,71 @@ it('execute external library in a sync way', () => {
...
+Overwriting commands on element level is almost the same, simply pass `true` as 3rd argument to `overwriteCommand`
+
+```js
+/**
+ * Attempt to scroll to element if it is not clickable.
+ * Pass { force: true } to click with JS even if element is not visible or clickable.
+ */
+// "click"            - command to be overwritten
+// origClickFunction  - original click function
+browser.overwriteCommand("click", function (origClickFunction, { force = false } = {}) {
```
mgrybyk: @CrispusDH take a look

docs/CustomCommands.md
```diff
@@ -96,3 +98,71 @@ it('execute external library in a sync way', () => {
 ```
 
 Note that the result of your custom command will be the result of the promise you return. Also there is no support for synchronous commands in standalone mode therefore you always have to handle asynchronous commands using promises.
+
+### Overwriting native commands
+
+You can also overwrite native commands with `overwriteCommand`. It is highly unrecommended to do that because it may lead to unpredictable behavior of framework! Overall approach is similar to `addCommand`, the only difference is that first argument in command function is original function, please see some examples below.
```
christian-bromann:
```suggestion
You can also overwrite native commands with `overwriteCommand`. It is not recommended to do this because it may lead to unpredictable behavior of the framework! The overall approach is similar to `addCommand`, the only difference is that the first argument in the command function is the original function that you are about to overwrite. Please see some examples below:
```

docs/CustomCommands.md
```diff
@@ -96,3 +98,71 @@ it('execute external library in a sync way', () => {
...
+
+#### Overwriting browser commands
+
+```js
+/**
+ * print milliseconds before pause and return its value.
+ */
+// "pause"            - command to be overwritten
+// origPauseFunction  - original pause function
+browser.overwriteCommand("pause", function (origPauseFunction, ms) {
```
christian-bromann:
```suggestion
browser.overwriteCommand('pause', function (origPauseFunction, ms) {
```

docs/CustomCommands.md
christian-bromann:
```suggestion
browser.overwriteCommand('click', function (origClickFunction, { force = false } = {}) {
```

packages/webdriver/src/monad.js
```diff
@@ -68,30 +77,69 @@ export default function WebDriver (options, modifier, propertiesObject = {}) {
...
+                } else {
+                    /**
+                     * regular mode
+                     */
+                    this.__propertiesObject__.__elementOverrides__.value[name] = customCommand
+                }
+            } else if (client[name] && origCommands[name]) {
+                delete client[name]
+                unit.lift(name, customCommand, proto, (...args) => origCommands[name].apply(this, args))
+            } else {
```
christian-bromann: if we return above then we don't need an else and save in indentation

docs/CustomCommands.md
CrispusDH: thanks for example 👍

packages/webdriver/src/monad.js
mgrybyk: in such case we need two return statements in `if` and `else if` blocks.
Is it better this instead of `else` block? 
Don't mind change it, just want to double check. 
@christian-bromann

mgrybyk:
@christian-bromann thanks for helping me with the doc

christian-bromann:
Can you resolve conflicts?

mgrybyk:
Yes, in next 2 hours

mgrybyk:
rebased!

PR #4050, opened 2019-06-02
Fix call, execute, executeAsync, remote, multiremote

### LINKED ISSUE #3796: execute and call typings have invalid return type in WebdriverIOAsync
 - **WebdriverIO version:** 5.7.8

**Describe the bug**
Execute and Call typings have invalid return type after wrapping with Promise, instead of generic T it is {}

**To Reproduce**
There is a test for it already
https://github.com/webdriverio/webdriverio/blob/master/tests/typings/webdriverio/async.ts#L9

just uncomment it and run TypeScript compiler.

**Expected behavior**
Execute and Call typings should return proper value, like in @wdio/sync 

**Additional context**
please uncomment test to prove the fix.

We can't wrap generic functions like this atm, waiting for https://github.com/Microsoft/TypeScript/issues/5453
So, have to duplicate these generic functions

mgrybyk:
## Proposed changes

With this PR I'd like to
- update minimum TypeScript version to 3.5.1 (this might be **breaking** change for TS users with lower versions!)
- fixed `call`, `execute`, `executeAsync` in async mode (wrapped with Promises)
- changed `BrowserObject` to interface so that it can be extended
- fixed `remote` to use proper capabilities object
- changed `multiremote` return type to `BrowserObject` from `WebDriver.ClientAsync`, hope it make sense
- changed `attach` Options to be required as far as sessionId is required
- **removed** `remote`, `multiremote`, `attach` from _sync_ mode (please correct me if it is possible to use these commands in _sync_ mode)
- added tests and examples for using custom commands

## Types of changes

## Checklist

## Further comments

Fix #3796

### DIFF
--- a/docs/TypeScript.md
+++ b/docs/TypeScript.md
@@ -4,7 +4,7 @@ title: TypeScript Setup
 ---
 
 Similar to Babel setup, you can register [TypeScript](http://www.typescriptlang.org/) to compile your .ts files in your before hook of your config file. You will need [ts-node](https://github.com/TypeStrong/ts-node) and [tsconfig-paths](https://github.com/dividab/tsconfig-paths) as the installed devDependencies.
-Minimal TypeScript version is 3.2.1
+Minimal TypeScript version is 3.5.1
 
 ```js
 // wdio.conf.js
@@ -53,7 +53,9 @@ For sync mode (@wdio/sync) `webdriverio` types have to be replaced with `@wdio/s
 }
 ```
 
-Please avoid importing webdriverio or @wdio/sync explicitly. `WebdriverIO` and `WebDriver` typings are accessible from anywhere once added to types in `tsconfig.json`.
+Please avoid importing webdriverio or @wdio/sync explicitly. `WebdriverIO` and `WebDriver` types are accessible from anywhere once added to types in `tsconfig.json`.
+
+### Typed Configuration
 
 You can even use a typed configuration if you desire.
 All you have to do is create a plain js config file that registers typescript and requires the typed config:
@@ -73,8 +75,10 @@ const config: WebdriverIO.Config = {
 export { config }
 ```
 
-Depending on the framework you use, you will need to add the typings for that framework to your `tsconfig.json` types property.
-For instance, if we decide to use the mocha framework, we need to add it like this to have all typings globally available:
+### Framework types
+
+Depending on the framework you use, you will need to add the types for that framework to your `tsconfig.json` types property.
+For instance, if we decide to use the mocha framework, we need to add it like this to have all types globally available:
 
 ```json
 {
@@ -92,7 +96,7 @@ For instance, if we decide to use the mocha framework, we need to add it like th
 }
 ```
 
-Instead of having all type definitions globally available, you can also import only the typings that you need like this:
+Instead of having all type definitions globally available, you can also import only the types that you need like this:
 
 ```typescript
 /*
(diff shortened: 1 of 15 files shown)


docs/TypeScript.md
```diff
@@ -101,3 +105,41 @@ Instead of having all type definitions globally available, you can also import o
...
+        browserCustomCommand: (arg) => void
+    }
+}
+```
+
+**Async mode**
+
+```
+declare module WebdriverIOAsync {
+    adding command to `$()`
```
```suggestion
    // adding command to `$()`
```

packages/webdriver/protocol/jsonwp.json
```diff
@@ -710,7 +710,7 @@
         "name": "args",
         "type": "(string|object|number|boolean|undefined)[]",
         "description": "the script arguments",
-        "required": true
+        "required": false
```
christian-bromann:
```suggestion
```

packages/webdriver/protocol/webdriver.json
```diff
@@ -651,7 +651,7 @@
         "name": "args",
         "type": "(string|object|number|boolean|undefined)[]",
         "description": "an array of JSON values which will be deserialized and passed as arguments to your function",
-        "required": true
+        "required": false
```
christian-bromann:
```suggestion
```

packages/webdriverio/webdriverio.d.ts
```diff
@@ -1,11 +1,10 @@
 /// <reference types="webdriverio/webdriverio-core"/>
 
-type BrowserObject = WebDriver.ClientOptions & WebDriver.ClientAsync & WebdriverIOAsync.Browser;
 type $ = (selector: string | Function) => Promise<WebdriverIOAsync.Element>;
 type $$ = (selector: string | Function) => Promise<WebdriverIOAsync.Element[]>;
 
 // Element commands that should be wrapper with Promise
-type ElementPromise = WdioOmit<WebdriverIO.Element, 'addCommand' | '$' | '$$'>;
+type ElementPromise = Omit<WebdriverIO.Element, 'addCommand' | '$' | '$$'>;
```
christian-bromann: We renamed this because it caused some conflicts with a typescript definition, has that been resolved?

packages/webdriverio/webdriverio.d.ts
```diff
@@ -15,50 +14,65 @@ interface AsyncSelectors {
...
+    [K in keyof BrowserPromise]:
+    (...args: Parameters<BrowserPromise[K]>) => Promise<ReturnType<BrowserPromise[K]>>;
 } & AsyncSelectors;
 
 // Browser commands that should not be wrapper with promise
 type BrowserStatic = Pick<WebdriverIO.Browser, 'addCommand' | 'options'>;
 declare namespace WebdriverIOAsync {
     function remote(
-        options?: WebDriver.Options & WebdriverIO.Options,
+        options?: WebdriverIO.RemoteOptions,
```
CrispusDH: Should be here `WebdriverIOAsync` instead of `WebdriverIO`?

mgrybyk: RemoteOptions is defined in WebdriverIO, so no.

packages/webdriverio/webdriverio.d.ts
mgrybyk: If we change minimum typescript version, then no. I've also removed our implementation of Omit.

packages/webdriver/protocol/jsonwp.json
mgrybyk: @christian-bromann I found 73 usages of `"required: false"` in all 5 json files. I supposed the idea is to specify if property is required or not explicitly. 
Don't mind removing it if I'm wrong, double checking.
Thanks

packages/webdriverio/webdriverio.d.ts
mgrybyk: https://github.com/webdriverio/webdriverio/pull/4050/files#diff-20adb64d960c251a4c323b411bec1f91R73

packages/webdriver/protocol/jsonwp.json
christian-bromann: ah ok, I wasn't sure anymore. It is fine then.

christian-bromann:
@mgrybyk how is this a breaking change?

mgrybyk:
Typescript users with version below 3.5.1 will have compilation error(s).
We can just notify them and ask to upgrade version

christian-bromann:
> We can just notify them and ask to upgrade version

Yeah, I guess we should tweet it out and do a minor semver update

PR #4056, opened 2019-06-04
wdio-mocha-framework: fix pending test test:end

### LINKED ISSUE #4018: Custom Reporter: Wrong data provided to onTestEnd when last test is a skipped test

 - **WebdriverIO version:** 5.9.1
 - **Mode:** WDIO Testrunner
 - **Node.js version:** 11.9.0
 - **NPM version:** Using yarn 1.15.2
 - **Browser name and version:** Chrome
 - **Platform name and version:** MacOS 

```
"devDependencies": {
    "@wdio/cli": "^5.7.1",
    "@wdio/dot-reporter": "^5.7.0",
    "@wdio/local-runner": "^5.7.1",
    "@wdio/mocha-framework": "^5.7.0",
    "@wdio/sync": "^5.7.2",
    "jest-matchers": "^20.0.3",
    "webdriverio": "^5.7.1",
     "wdio-json-reporter": "^1.1.0",
    "wdio-mochawesome-reporter": "^3.0.0"
  },
```

**Config of WebdriverIO**
Pretty much defaults except the reporters and my spec file
```
reporters: [
        'dot',
        ['mochawesome',{ 
            outputDir: './Results', 
            includeScreenshots: true,
            outputFileFormat: function(opts) { 
            return `wdio-mochawesome-${opts.cid}.${opts.capabilities.browserName}.json`
        }}]
    ],
```

**Describe the bug**
When the last test in a suite is a skipped test, the test object provided to `onTestEnd` has the same information as the test run just prior to the skipped test.

**To Reproduce**
Minimal Spec File:
```
const expect = require('jest-matchers')
describe('Sample Suite 1',function(){

    it('should report fail',function(){
        browser.url('https://www.google.com')
        expect(true).toBe(false)
    })

    it.skip('should report skip',function(){
        browser.url('https://www.google.com')
    })
})
```

Minimal Custom Reporter:
```
const WDIOReporter = require('@wdio/reporter').default

export default class CustomReporter extends WDIOReporter {
    onTestStart (test) { console.log('onTestStart',test) }
    onTestSkip (test) { console.log('onTestSkip',test) }
    onTestEnd (test) { console.log('onTestEnd',test) }
};
```

**Expected behavior**
One of two behaviors:

1) A onTestEnd event for a skipped test should provide a test object that matches the skipped test

2) No onTestEnd event should be raised for skipped tests.  This seems logical because the onTestStart is not raised.

**Log**
https://gist.github.com/fijijavis/504e684dcfa2a760f17d6f298d4f3e25

There are 2 log files in this gist.

emit.log
I added a console.log to the `emit` function in @ wdio/mocha-framework/index.js.  In the log you will see 2 `test:end` events.  One for the failed test and a second for the skipped test.  What is interesting to me is that BOTH have the same `uid` value.

reporter.log
The result of the console.log statements from a custom reporter (like above).  You will see 2 log statements for `onTestEnd` and they have the same test object.

**Additional context**
I think the ideal state would be to not emit a `onTestEnd` event for a skipped test.  
* The `onTestStart` event is not emitted so this would match.
* The test didn't run so logically it couldn't have "ended"

I tried (unsuccessfully) to figure out where that change should be made.  If there are any pointers then I can try to submit a PR for a fix.

Thanks!

mgrybyk:
## Proposed changes

Suggestion: increment uid for pending test in order to fix `test:end` event.

fix #4018 4018

I'm not 100% sure who is an original caller of `test:end` event after `test:pending`, as far as I can see it is **mocha** itself. If my assumption is correct, then current solution is fine. There is no need in changing mocha behavior as far as it may affect some reporters.

-- alternative solutions (more risky)

2nd option is not calling `test:end` for pending tests at all, but I'm not sure how can it affect Jasmine.
```
this.on('test:end',  /* istanbul ignore next */ (test) => {
    if (test.pending) { return }
    const testStat = this.tests[test.uid]
    this.onTestEnd(testStat)
})
```

3rd option is to ignore `test:end` event only for mocha
```
emit(event, payload, err) {
    ...
    if (message.pending) { return }
    this.reporter.emit(message.type, message);
}
```

## Types of changes

## Checklist

## Further comments

a bit changed reporter provided by @fijijavis (printing message only instead of test object)
```
class CustomReporter extends WDIOReporter {
    onTestStart (test) { console.log('onTestStart',test.title) }
    onTestSkip (test) { console.log('onTestSkip',test.title) }
    onTestEnd (test) { console.log('onTestEnd',test.title) }
}
```
**Result**
image


BorisOsipov:
>I'm not 100% sure who is an original caller of test:end event after test:pending, as far as I can see it is mocha itself. 

Me too. It is cornerstone how different adapters reports pending tests. AFAIK in my reporters I had workarounds for sch cases.

BorisOsipov:
Will be nice if you have look\debug at least how allure reporter handle such case

packages/wdio-devtools-service/tests/gatherer/trace.test.js
```diff
@@ -217,7 +217,7 @@ test('onLoadEventFired: using min trace time', async () => {
 
     const start = Date.now()
     await traceGatherer.onLoadEventFired()
-    expect(Date.now() - start).toBeGreaterThanOrEqual(299)
+    expect(Date.now() - start).toBeGreaterThanOrEqual(250)
```
mgrybyk: @christian-bromann according to https://github.com/webdriverio/webdriverio/blob/bc6ff2b374aab716d4f4d4300b82c89f01b5df20/packages/wdio-devtools-service/src/gatherer/trace.js#L156 
`waitForNetworkIdleEvent` and `waitForCPUIdleEvent` are running in parallel which means that overall time should be `100 + 150` instead of `50 + 100 + 150`. Please correct me if I'm wrong

mgrybyk: added this fix because the test was failing for me with values 294, 298

mgrybyk:
@BorisOsipov allure results looks good for me
image

packages/wdio-devtools-service/tests/gatherer/trace.test.js
christian-bromann: sounds good to me

packages/wdio-mocha-framework/src/index.js
```diff
@@ -274,7 +274,7 @@ class MochaAdapter {
         if (message.type === 'hook:end') {
             return this.getSyncEventIdEnd('hook')
         }
-        if (message.type === 'test:start') {
+        if (['test:start', 'test:pending'].includes(message.type)) {
             return this.getSyncEventIdStart('test')
         }
         if (['test:pending', 'test:end', 'test:pass', 'test:fail'].includes(message.type)) {
```
christian-bromann: I think we can remove `test:pending` here

mgrybyk:
I'll add tests soon

packages/wdio-mocha-framework/src/index.js
mgrybyk: Looks like yes. I'll double check

mgrybyk: you are right, removing it.

PR #4063, opened 2019-06-06
Minor fix to add a try-catch block if the delete-session throws an error.

nami-varthakavi:
## Proposed changes

## Types of changes

## Checklist

## Further comments


mgrybyk:
When does it throw an exception?

nami-varthakavi:
> When does it throw an exception?

We have special application were we use multiremote in which both drivers are embedded in single application. So if you kill one automatically the second one is killed.

nami-varthakavi:
> Please add unit tests

Yes I can.

christian-bromann:
> So if you kill one automatically the second one is killed.

If you run multiremote you have to call deleteSession for both browser because you run 2 sessions in parallel. I have my reservations with this patch. Can you please explain in more detail your situation?

nami-varthakavi:
Sure. The application has two components 
1. chromium based application controlled using chromedriver
2. A qt component which is controlled using qtdriver and the driver lib is already embedded in it. 

Even though multi-remote config is to interact between two different apps. In this scenario multi remote works out of the box. 

For example when I kill the chromedriver the application and the session is ended. And after that when the deleteSession tries to delete the qtDriver session, it has been already killed by the first delete session. So it throws an error.

nami-varthakavi:
If you guys are ok I may suggest one more option to add a new boolean ‘config’ option to provide a choice for the user to handle the clean up instead of handling it by wdio.

nami-varthakavi:
> Sure. The application has two components
> 
> 1. chromium based application controlled using chromedriver
> 2. A qt component which is controlled using qtdriver and the driver lib is already embedded in it.
> 
> Even though multi-remote config is to interact between two different apps. In this scenario multi remote works out of the box.
> 
> For example when I kill the chromedriver the application and the session is ended. And after that when the deleteSession tries to delete the qtDriver session, it has been already killed by the first delete session. So it throws an error.

Can I get this one reviewed please? @christian-bromann @mgrybyk.

mgrybyk:
@nami-varthakavi there reason why it is not yet merged is that introducing try/catch there looks for me like a hacky workaround (a bit).

Maybe introducing config option on whether to close session or not is more reasonable? In such case you'll be able to end session for only one of your multiremote sessions.

What are your thoughts regarding this one @christian-bromann ? Is it good to merge or we should think on different solution?

christian-bromann:
It seems like a workaround for me too. Usually if a session is started it needs to be closed again. Even though the browser process was killed the session still exists in the driver. I would like to explore what is causing the error and who is closing the session in the first place. Then apply a fix that would only apply for that specific scenario.

nami-varthakavi:
> It seems like a workaround for me too. Usually if a session is started it needs to be closed again. Even though the browser process was killed the session still exists in the driver. I would like to explore what is causing the error and who is closing the session in the first place. Then apply a fix that would only apply for that specific scenario.

I can probably explain it to u directly if that is fine because I don't want to share too many details in here

nami-varthakavi:
> @nami-varthakavi there reason why it is not yet merged is that introducing try/catch there looks for me like a hacky workaround (a bit).
> 
> Maybe introducing config option on whether to close session or not is more reasonable? In such case you'll be able to end session for only one of your multiremote sessions.
> 
> What are your thoughts regarding this one @christian-bromann ? Is it good to merge or we should think on different solution?

Yeah I can make the necessary changes to create a new config option

nami-varthakavi:
> It seems like a workaround for me too. Usually if a session is started it needs to be closed again. Even though the browser process was killed the session still exists in the driver. I would like to explore what is causing the error and who is closing the session in the first place. Then apply a fix that would only apply for that specific scenario.

DMed to describe the scenario we are encountering. If you still feel, are you cool with updating to use a new config-option?

christian-bromann:
> use a new config-option

I am not a fan of adding a config option. I want to rather explore why the session is being closed by the QT driver and maybe fix the error there.

nami-varthakavi:
> > use a new config-option
> 
> I am not a fan of adding a config option. I want to rather explore why the session is being closed by the QT driver and maybe fix the error there.

It's how the application has been built because both the drivers are just part of a single application. 
Anyway, I will make the changes and update the PR soon.

nami-varthakavi:
> > use a new config-option
> 
> I am not a fan of adding a config option. I want to rather explore why the session is being closed by the QT driver and maybe fix the error there.

The application is built in such a way that the qt driver will be only listening to port 9517 if the application is up and running. When the driver session communicating chromium part of the application is deleted, it kills the application altogether. Since the application is killed, there is no longer existence of the qtDriver application.

nami-varthakavi:
@christian-bromann is it okay to add the config option. I understand try-catch is a workaround but adding a config option is a much safer way to do this. And providing an option to the user to do the clean up by either wdio/themselves can be considered as a feature request?

christian-bromann:
@nami-varthakavi to get more context: are you automating the same application with two drivers?

nami-varthakavi:
> @nami-varthakavi to get more context: are you automating the same application with two drivers?

yeah

nami-varthakavi:
> > @nami-varthakavi to get more context: are you automating the same application with two drivers?
> 
> yeah

christian-bromann:
> yeah! The desktop application's toolbar is controlled by QT and the chrome driver controls the application inside like the main menus and basic contents.

How about not using multiremote then and instead use WebdriverIO to create a separate standalone session calling the `remote` function in the before hook?

nami-varthakavi:
> > yeah! The desktop application's toolbar is controlled by QT and the chrome driver controls the application inside like the main menus and basic contents.
> 
> How about not using multiremote then and instead use WebdriverIO to create a separate standalone session calling the `remote` function in the before hook? I don't think multiremote is the best approach here given that you automate a single application and given the fact that apparently the qtDriver closes the session when the application closes.
> 
> Also please file an issue in the qtDriver project. A session in the driver should not be automatically closed if the application (e.g. the browser) is being closed. You can try this with Chrome and Chromedriver where you still have a session even you quit the browser window.

It's not basically a problem of qt driver. We have packaged the qt driver and its libraries within the application, that's why it is killing the application whenever we kill the driver. Our devs have tried many ways to fix this but the solution was only to embed within the application. I don't think there is a way around that, 

I will anyway try doing it in the before the block and see.

nami-varthakavi:
Currently, in our end, this fix has been deprioritized. I'm closing this PR for now.

PR #4089, opened 2019-06-13
update appium service

### LINKED ISSUE #4081: Appium service: move appium to dev dependency and update cli
Appium service has `appium` as a dependency. This is a restriction for users who want to use appium different from the one is shipped.

- move appium to dev dependency 
- also install appium same as chromedriver on initial setup if it is not installed globally

mgrybyk:
## Proposed changes

Remove `appium` from `@wdio/appium-service` dependencies and install `appium` same as `chromedriver` if it is not installed globally.

fixed #4081 

also
- use outputDir as folder for appium logs if it is not specified in appium
- fixed `chromedriver` condition in `wdio.conf.tpl.ejs`
- `chromedriver` and `appium` packages are added when using `wdio install service ...`
- fixed some failing tests because of Date.now()

## Types of changes

## Checklist

## Further comments

tested changes in my project.

### DIFF
--- a/packages/wdio-appium-service/package.json
+++ b/packages/wdio-appium-service/package.json
@@ -33,8 +33,10 @@
   "bugs": {
     "url": "https://github.com/webdriverio/webdriverio/issues"
   },
+  "devDependencies": {
+    "appium": "^1.13.0"
+  },
   "dependencies": {
-    "appium": "^1.12.0",
     "fs-extra": "^8.0.1",
     "param-case": "^2.1.1"
   },
--- a/packages/wdio-appium-service/package.json
+++ b/packages/wdio-appium-service/package.json
@@ -33,9 +33,6 @@
   "bugs": {
     "url": "https://github.com/webdriverio/webdriverio/issues"
   },
-  "devDependencies": {
-    "appium": "^1.13.0"
-  },
   "dependencies": {
     "fs-extra": "^8.0.1",
     "param-case": "^2.1.1"
--- a/packages/wdio-appium-service/src/launcher.js
+++ b/packages/wdio-appium-service/src/launcher.js
@@ -18,7 +18,7 @@ export class AppiumLauncher {
     async onPrepare(config) {
         const appiumConfig = config.appium || {}
 
-        this.logPath = appiumConfig.logPath
+        this.logPath = appiumConfig.logPath || config.outputDir
         this.command = appiumConfig.command || this._getAppiumCommand()
         this.appiumArgs = this._cliArgsFromKeyValue(appiumConfig.args || {})
 
--- a/packages/wdio-appium-service/src/launcher.js
+++ b/packages/wdio-appium-service/src/launcher.js
@@ -70,7 +70,15 @@ export class AppiumLauncher {
     }
 
     _getAppiumCommand(moduleName = 'appium') {
-        return require.resolve(moduleName)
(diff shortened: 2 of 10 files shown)


packages/wdio-appium-service/package.json
```diff
@@ -33,8 +33,10 @@
   "bugs": {
     "url": "https://github.com/webdriverio/webdriverio/issues"
   },
+  "devDependencies": {
+    "appium": "^1.13.0"
```
christian-bromann: I am not sure if we need Appium whatsoever because in dev I don't really need it too

mgrybyk:
> If neither appium is installed globally nor as dev dependency, can we install it for the user?

https://github.com/webdriverio/webdriverio/pull/4089/files#diff-67c78548af6da581e175d98914982e23R42

packages/wdio-appium-service/package.json
mgrybyk: let me check and remove if possible

mgrybyk: there were some changes required in test but I've removed appium.

mgrybyk: there were some changes required in test but I've removed appium.

christian-bromann:
> it is done same as for chromedriver, we are installing appium if it is not installed globally

Yeah but this applies only if people to the setup wizard. What happens if they add it to their existing framework.

mgrybyk:
Got your point, I'll check. Completely forgot about that, thank you

mgrybyk:
@christian-bromann if people add either appium or chromedriver service no additional packages will be installed right now

christian-bromann:
@mgrybyk with this change for the appium service, yes. That is why I would suggest to install appium or download chromedriver for these cases, or have a clear error message saying what is needed to do.

mgrybyk:
ok, I'll change behavior for appium and chromedriver

mgrybyk:
@baruchvlz you might be interested as well.

mgrybyk:
@christian-bromann 
`./bin/wdio.js install service appium` installs also `appium` if it is not globally installed, otherwise prints message saying that it is needed to add `appium: { command: 'appium' }`

`./bin/wdio.js install service chromedriver` installs also `chromedriver`. It should mention that `path` have to be changes to `/` -> `path: '/'`

packages/wdio-appium-service/src/launcher.js
```diff
@@ -69,8 +69,16 @@ export class AppiumLauncher {
         this.process.stderr.pipe(logStream)
     }
 
-    _getAppiumCommand() {
-        return require.resolve('appium')
+    _getAppiumCommand(moduleName = 'appium') {
+        try {
+            return require.resolve(moduleName)
+        } catch (err) {
+            log.error('appium is not installed locally.\n' +
```
christian-bromann:
```suggestion
            log.error('appium is not installed locally. Please install Appium via `npm install appium`.\n' +
```

packages/wdio-appium-service/src/launcher.js
```diff
@@ -69,8 +69,16 @@ export class AppiumLauncher {
...
+            'If you use globally installed appium please add\n' +
```
christian-bromann:
```suggestion
            'If you use globally installed Appium please add\n' +
```

PR #4102, opened 2019-06-19
wdio-appium-service: fix args array, print startup failure message

mgrybyk:
## Proposed changes

- fix args array support
- print startup failure message (previously it was only static message)
image

## Types of changes

## Checklist

### DIFF
--- a/packages/wdio-appium-service/src/launcher.js
+++ b/packages/wdio-appium-service/src/launcher.js
@@ -40,6 +40,7 @@ export class AppiumLauncher {
     _startAppium(command, args, waitStartTime, callback) {
         log.debug(`Will spawn Appium process: ${command} ${args.join(' ')}`)
         let process = spawn(command, args, { stdio: ['ignore', 'pipe', 'pipe'] })
+        let error
 
         process.stdout.on('data', (data) => {
             if (data.includes('Appium REST http interface listener started')) {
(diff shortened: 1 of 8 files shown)


packages/wdio-appium-service/tests/launcher.test.js
```diff
@@ -122,7 +123,8 @@ describe('Appium launcher', () => {
             } catch (e) {
                 error = e
             }
-            const expectedError = new Error("Appium exited before timeout (exit code: 2) - Check that you don't already have a running Appium service.")
+            const expectedError = new Error('Appium exited before timeout (exit code: 2)\n' +
```
christian-bromann: Can we add a test that would have the error from `process.stderr.once('data', err => { error = err })`? It's the branch that makes the test failing due to missing coverage.

If it is too difficult to test this feel free to improve coverage somewhere else where it is simpler.

packages/wdio-appium-service/src/launcher.js
```diff
@@ -48,10 +49,15 @@ export class AppiumLauncher {
...
+        // only capture first error to print it in case Appium failed to start.
+        process.stderr.once('data', err => { error = err })
+
         process.once('exit', (exitCode) => {
             let errorMessage = `Appium exited before timeout (exit code: ${exitCode})`
             if (exitCode == 2) {
-                errorMessage += ' - Check that you don\'t already have a running Appium service.'
+                errorMessage += '\n' + (error || 'Check that you don\'t already have a running Appium service.')
+                // eslint-disable-next-line no-console
+                console.error(errorMessage)
```
christian-bromann:
```suggestion
                log.error(errorMessage)
```

mgrybyk: I'm not sure why, but if I replace console.error with log.error I can't see it anywhere. 
Maybe there is some race condition or we are not waiting for log to be flushed. 

I'll check

mgrybyk: changed it to `log.error`.
log level was set to `warn` in my project and for some reason because of that error messages were not showing up. 
Changed it to `info` fixes the problem, have no idea why is that happening, have to investigate separately.

PR #4116, opened 2019-06-25
Fix bug where environment flags aren't attached to session

christian-bromann:
## Proposed changes

Currently if you use `attachSession` or `attach` our environment flags aren't properly set on the attached session. This patches fixes this.

## Types of changes

## Checklist

## Further comments

This bug was detected during Appiums development on badge processes: https://github.com/appium/appium-base-driver/pull/328/files

### DIFF
--- a/packages/webdriver/src/index.js
+++ b/packages/webdriver/src/index.js
@@ -5,7 +5,7 @@ import { validateConfig } from '@wdio/config'
 import webdriverMonad from './monad'
 import WebDriverRequest from './request'
 import { DEFAULTS } from './constants'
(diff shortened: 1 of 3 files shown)


christian-bromann:
> so it was just a simple matter of the code not existing in the attach flow.

Yes, also we didn't created the prototype object correctly, it was

```js
{
   isW3C: <boolean>,
   ...
}
```
instead of:
```js
{
   isW3C: { value: <boolean> },
   ...
}
```

erwinheitzman:
getEnvironmentVars still returns the object like so: isW3C: { value: isW3C }
is this correct then?

PR #4128, opened 2019-06-28
webdriverio: fix isDisplayed script to work with shadow dom

jrobinson01:
## Proposed changes

Using the `isDisplayed` and `waitForDisplayed` commands with elements inside shadow dom, with the latest version of chromedriver (w3c) incorrectly reports elements as not visible. This PR fixes that by modifying the `isElementDisplayed`.

## Types of changes

## Checklist

## Further comments

 I would like to add some tests in the /examples directory but am not 100% sure how that works. I have a test page and some tests written in webdriverio locally that I could use. Guidance would be appreciated!

### DIFF
--- a/packages/webdriverio/src/scripts/isElementDisplayed.js
+++ b/packages/webdriverio/src/scripts/isElementDisplayed.js
@@ -70,6 +70,11 @@ export default function isElementDisplayed(element) {
     function cascadedStylePropertyForElement(element, property) {
         if (!element || !property)
             return null
+        // if document-fragment, skip it and use element.host instead. This happens
+        // when the element is inside a shadow root.
+        // window.getComputedStyle errors on document-fragment.
+        if (element instanceof DocumentFragment)
+            element = element.host
 
         let computedStyle = window.getComputedStyle(element)
         let computedStyleProperty = computedStyle.getPropertyValue(property)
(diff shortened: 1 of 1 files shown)


packages/webdriverio/src/scripts/isElementDisplayed.js
```diff
@@ -152,12 +157,21 @@ export default function isElementDisplayed(element) {
             return isElementSubtreeHiddenByOverflow(childNode)
         })
     }
+    // walk up the tree testing for a shadow root
+    function isElementInsideShadowRoot(element) {
+        if (!element) {
```
mgrybyk: Can element  be undefined at this point?

jrobinson01: I think it can be null. Are you questioning the need for the conditional of suggesting it is more explicit?

mgrybyk: If it can be null, I'm fine with it, thank you

jrobinson01:
@christian-bromann I’d be happy to add a test. I’m just not sure quite where it would go. There doesn’t appear to be any other tests for the script I changed. I Would like to use a simple html page with a custom-element to run tests against but not sure where to put it in the project. Any guidance would be much appreciated!

jrobinson01:
@christian-bromann Here's what I'm thinking. This is a separate repo, but should give you an idea of what I'd like to do as far as tests go (this could be expanded to cover the shadow$ and shadow$$ commands as well). A simple html page that needs to be served somehow: https://github.com/jrobinson01/wdio-boilerplate/blob/v5-shadow-root-command/test-site/index.html
and then a wdio test for that page:
https://github.com/jrobinson01/wdio-boilerplate/blob/v5-shadow-root-command/tests/shadow.spec.js

mgrybyk:
We can add unit tests for `isElementDisplayed.js` file, but I'm not sure if it makes much sense to do that (correct me if I'm wrong).

There is no problem to host some page, but there are problems with browser. I don't know if we can run browser tests in Travis CI, I think no.

jrobinson01:
I agree @mgrybyk that unit tests probably don't make a ton of sense here. I was under the impression that the project ran some tests using webdriverio but I could be mistaken.

PR #4149, opened 2019-07-08
wdio-allure-reporter: Add capability to include attachment in afterhook

PippoRaimondi:
## Proposed changes

Adding a simple flag to a reporter to allow us to add an existing attachment in the after hook

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/packages/wdio-allure-reporter/src/index.js
+++ b/packages/wdio-allure-reporter/src/index.js
@@ -249,9 +249,11 @@ class AllureReporter extends WDIOReporter {
         test.setDescription(description, type)
     }
 
-    addAttachment({ name, content, type = 'text/plain' }) {
-        if (!this.isAnyTestRunning()) {
-            return false
+    addAttachment({ name, content, type = 'text/plain' }, loadExistingAttachmentInAfterHook = false) {
+        if (!loadExistingAttachmentInAfterHook) {
+            if (!this.isAnyTestRunning()) {
+                return false
+            }
         }
 
         if (type === 'application/json') {
--- a/packages/wdio-allure-reporter/tests/allureFeatures.test.js
+++ b/packages/wdio-allure-reporter/tests/allureFeatures.test.js
@@ -138,6 +138,33 @@ describe('reporter runtime implementation', () => {
         expect(addAttachment).toHaveBeenCalledWith('foo', Buffer.from('bar'), 'baz')
     })
 
+    it('should not call add attachment if no test is running', () => {
+        const reporter = new AllureReporter({ stdout: true })
+        const addAttachment = jest.fn()
+        reporter.allure = {
+            getCurrentSuite: jest.fn(() => false),
+            getCurrentTest: jest.fn(() => false),
+            addAttachment
+        }
+
+        reporter.addAttachment({ name: 'foo', content: 'bar', type: 'baz' })
+        expect(addAttachment).toHaveBeenCalledTimes(0)
+    })
+
+    it('should call add attachment if loading an attachment', () => {
+        const reporter = new AllureReporter({ stdout: true })
+        const addAttachment = jest.fn()
+        reporter.allure = {
+            getCurrentSuite: jest.fn(() => false),
+            getCurrentTest: jest.fn(() => false),
+            addAttachment
+        }
+
+        reporter.addAttachment({ name: 'foo', content: 'bar', type: 'baz' }, true)
+        expect(addAttachment).toHaveBeenCalledTimes(1)
+        expect(addAttachment).toHaveBeenCalledWith('foo', Buffer.from('bar'), 'baz')
+    })
+
     it('should correct add "application/json" attachment', () => {
         const reporter = new AllureReporter({ stdout: true })
         const dumpJSON = jest.fn()


packages/wdio-allure-reporter/src/index.js
```diff
@@ -249,9 +249,11 @@ class AllureReporter extends WDIOReporter {
         test.setDescription(description, type)
     }
 
-    addAttachment({ name, content, type = 'text/plain' }) {
-        if (!this.isAnyTestRunning()) {
-            return false
+    addAttachment({ name, content, type = 'text/plain' }, loadExistingAttachmentInAfterHook = false) {
+        if (!loadExistingAttachmentInAfterHook) {
```
christian-bromann:
```suggestion
        if (!loadExistingAttachmentInAfterHook || !this.isAnyTestRunning()) {
```
no need for two nested ifs

BorisOsipov:
>Adding a simple flag to a reporter to allow us to add an existing attachment in the after hook

can you describe your which issue you solve? For me it is not clear why we need some boolean flag. Looks like this condition can be easy implemented in your own code..

PippoRaimondi:
> > Adding a simple flag to a reporter to allow us to add an existing attachment in the after hook
> 
> can you describe your which issue you solve? For me it is not clear why we need some boolean flag. Looks like this condition can be easy implemented in your own code..

I am going to use this report to attach screenshots from my visual regression test suite so it's easier to debug in case something goes wrong. And so I don't use the `addAttachment` method in each and every test rather than in a global hook, I need to ensure it doesn't return... does this make sense?

packages/wdio-allure-reporter/src/index.js
PippoRaimondi: No, we need to ensure the nested condition is not interpreted at all if the flag is true...  it's logically incorrect to use `||` or `&&` in this case.

christian-bromann:
@PippoRaimondi I agree with @BorisOsipov you could just instead of calling
```js
const loadExistingAttachmentInAfterHook = false
// ...
addAttachment({ ... }, loadExistingAttachmentInAfterHook)
```
do
```js
if (loadExistingAttachmentInAfterHook) {
    addAttachment({ ... })
```

PippoRaimondi:
> @PippoRaimondi I agree with @BorisOsipov you could just instead of calling
> 
> ```js
> const loadExistingAttachmentInAfterHook = false
> // ...
> addAttachment({ ... }, loadExistingAttachmentInAfterHook)
> ```
> 
> do
> 
> ```js
> if (loadExistingAttachmentInAfterHook) {
>     addAttachment({ ... })
> }
> ```

Please note the wdio allure reporter instance method `addAttachment` is a wrapper to allures method `addAttachment`, which is what I am trying to get invoked here... The below condition prevents allures instance `addAttachment` from being called.

```
    if (!this.isAnyTestRunning()) {
        return false
    }
```

In the `afterEach` hook `this.isAnyTestRunning()` returns `false`... hence I needed a way to bypass that so I could attach my screenshots... I don't have a lot of context why we added that check first place, so with the little information I had I added a flag to ensure we didn't break any existing functionality...

christian-bromann:
> In the `afterEach` hook

Which afterEach hook? The one in your code or in one of the plugins?

PippoRaimondi:
> > In the `afterEach` hook
> 
> Which afterEach hook? The one in your code or in one of the plugins?

In my code, within wdio test suite...

christian-bromann:
> In my code, within wdio test suite

So then add an additional `if` in your test suite as I mentioned in my comment above.

PippoRaimondi:
> > In my code, within wdio test suite
> 
> So then add an additional `if` in your test suite as I mentioned in my comment above.

Sorry, I might have not explained it very well the reason why it won't work if I add an `if` in my test suite.

Let me test something out (might add a new method all together) then I will come back to you with a more detailed reason why I am doing it in the package.

PippoRaimondi:
> > > In my code, within wdio test suite
> > 
> > 
> > So then add an additional `if` in your test suite as I mentioned in my comment above.
> 
> Sorry, I might have not explained it very well the reason why it won't work if I add an `if` in my test suite.
> 
> Let me test something out (might add a new method all together) then I will come back to you with a more detailed reason why I am doing it in the package.

Ok, so this will never work by changing `wdio-allure-reporter` package alone.

Quick recap, the flag I was adding was attempting to bypass condition that returns `false` from the method below. This was one of the reasons why calling the addAttachment out of the test context wouldn't work, i.e outside of the `it` block... What I was trying to avoid was have to call `addAttachment` in each test (within the `it` block)...

```
addAttachment({ name, content, type = 'text/plain' }) {
    **if (!this.isAnyTestRunning()) {
        return false
    }**

    if (type === 'application/json') {
        this.dumpJSON(name, content)
    } else {
        this.allure.addAttachment(name, Buffer.from(content), type)
    }
}
```

Although it was a bit sloppy at my end thinking the condition in bold above was the only thing needed to be changed (also the update was incomplete, I had to change the static reference as well to actually get the flag to "work")...

I looked deeper and the real blocker is with `allure-js-common`... by the end of the test or suite it calls `endCase` or `endSuite` and unless the architecture of that bit is changed the methods can never really be called in hooks as they lose all the test context... To fix this would require a bit of work and not sure there is enough value as I have a workaround to this issue.

Unless you have any other idea of how to accomplish this, it's safe to close this PR.

christian-bromann:
Thanks for looking into it. Closing!

PR #4179, opened 2019-07-12
Add EsLint check for dependencies

christian-bromann:
Yesterday we released a new version where a dependency was added to a sub package but not added as a dependency in the package.json. For some reason an Eslint check for this is not in place. This should be fixed so we don't introduce such bugs again.


christian-bromann:
I found the issue why these issues were not detected by eslint and fixed them.

Issue #4165, opened 2019-07-10
Not possible to call original function on different element in overwriteCommand


akloeber:
**Environment (please complete the following information):**
 - **WebdriverIO version:** @5.10.7
 - **Mode:** WDIO Testrunner
 - **If WDIO Testrunner, running sync/async:** sync
 - **Node.js version:** 10.16.0
 - **NPM version:** 6.9.0
 - **Browser name and version:** Chrome 75
 - **Platform name and version:** Mac OS 10.14.5
 - **Additional wdio packages used (if applicable):**
```
 $ yarn list --pattern wdio
yarn list v1.16.0
├─ @wdio/cli@5.10.7
├─ @wdio/config@5.9.4
├─ @wdio/junit-reporter@5.9.3
├─ @wdio/local-runner@5.10.7
├─ @wdio/logger@5.9.3
├─ @wdio/mocha-framework@5.10.1
├─ @wdio/repl@5.9.4
├─ @wdio/reporter@5.9.3
├─ @wdio/runner@5.10.7
├─ @wdio/spec-reporter@5.9.3
├─ @wdio/sync@5.10.1
├─ @wdio/utils@5.9.3
├─ wdio-chromedriver-service@5.0.2
├─ wdio-mochawesome-reporter@3.1.0
├─ webdriver@5.10.9
│  └─ @wdio/config@5.10.8
└─ webdriverio@5.10.9
   ├─ @wdio/config@5.10.8
   └─ @wdio/repl@5.10.8
✨  Done in 0.34s.
```

**Describe the bug**
We are using Kendo UI components and want to override `getValue()` to deliver corrects values for those components. In oder to do this we have to invoke the orginal `getValue()` function on a nested element:
```js
browser.overwriteCommand('getValue', function (origGetValueFn) {
  switch (this.getTagName()) {
    case 'kendo-numerictextbox':
    case 'kendo-datepicker': {
      return origGetValueFn.call(this.$('.k-input'));
    }
    case 'kendo-dropdownlist':
      return this.$('.k-input').getText();
    default:
      return origGetValueFn.call(this);
  }
}, true);
```

Unfortunately it looks like `origGetValueFn` is always bound to the original element und thus the the call of `origGetValueFn` returns `null` in this case as the main element of the Kendo component (e.g. `<kendo-numerictextbox>`) does not have a value itself.

**Expected behavior**
It should be possible to "reuse" `origGetValueFn` on a different element determined inside the override-function.

[...]

PR #4186, opened 2019-07-12
webdriver: support rebinding of context when invoking origFn in element.overwriteCommand

akloeber:
webdriver: support rebinding of context when invoking origFn in `element.overwriteCommand`, fixes https://github.com/webdriverio/webdriverio/issues/4165

## Proposed changes

It is now possible to call the original function in `element.overwriteCommand` with a different context. If no context is set the element is used to keep this backwards compatible.

## Types of changes

## Checklist

### DIFF
--- a/packages/webdriver/tests/utils.test.js
+++ b/packages/webdriver/tests/utils.test.js
@@ -290,14 +290,33 @@ describe('utils', () => {
     describe('overwriteElementCommands', () => {
         it('should overwrite command', function () {
             const context = {}
+            const origFnMock = jest.fn(() => 1)
             const propertiesObject = {
-                foo: { value() { return 1 } },
+                foo: { value: origFnMock },
                 __elementOverrides__: {
                     value: { foo(origCmd, arg) { return [origCmd(), arg] } }
                 }
             }
             overwriteElementCommands.call(context, propertiesObject)
             expect(propertiesObject.foo.value(5)).toEqual([1, 5])
+            expect(origFnMock.mock.calls.length).toBe(1)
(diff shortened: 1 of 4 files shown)


packages/webdriver/tests/utils.test.js
```diff
@@ -290,14 +290,33 @@ describe('utils', () => {
...
                     value: { foo(origCmd, arg) { return [origCmd(), arg] } }
                 }
             }
             overwriteElementCommands.call(context, propertiesObject)
             expect(propertiesObject.foo.value(5)).toEqual([1, 5])
+            expect(origFnMock.mock.calls.length).toBe(1)
+            expect(origFnMock.mock.instances[0]).toBe(propertiesObject.foo)
+        })
+
+        it('should support rebinding when invoking original fn', function () {
```
christian-bromann: Can you modify the test so that it returns a different value as expected because it is bound to a different instance?

akloeber: You mean in addition to the check of jest.mock.instances which already tracks this-bindings on invocations? Or do you want replace the checks on the jest-mock completely with checks against the return value?

christian-bromann: I haven't used `jest.fn().mock.instances` before so it wasn't clear for me that it theoretical also checks that. Might adding another smoke test for this here: https://github.com/webdriverio/webdriverio/blob/master/tests/mocha/test.js#L128 ?

akloeber: Sure, I'll do that.

akloeber: I have added a smoketest.

PR #4194, opened 2019-07-15
isSauce should not depend on hostname

christian-bromann:
## Proposed changes

Removing the hostname check in the `isSauce` utility as the package should also mark this flag to `true` if the user runs tests through Sauce Connect.

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/packages/webdriver/src/utils.js
+++ b/packages/webdriver/src/utils.js
@@ -286,14 +286,10 @@ export function isAndroid (caps) {
  */
 export function isSauce (hostname, caps) {
     return Boolean(
-        hostname &&
-        hostname.includes('saucelabs') &&
+        caps.extendedDebugging ||
         (
-            caps.extendedDebugging ||
-            (
-                caps['sauce:options'] &&
-                caps['sauce:options'].extendedDebugging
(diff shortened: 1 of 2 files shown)


packages/webdriver/src/utils.js
```diff
@@ -286,14 +286,10 @@ export function isAndroid (caps) {
  */
 export function isSauce (hostname, caps) {
     return Boolean(
-        hostname &&
```
wswebcreation: This will not work for people who are still using Sauce without W3C because not all users are using `extendedDebugging`

christian-bromann: @wswebcreation the flag only indicates if the custom Sauce commands can be applied to the client. It was doing this before too. I agree with you though that this can be confusing for people that don't run extended debugging jobs and want to use the flag. I would like to tackle this in a different PR thought to clean this flag up and have two flags for running on Sauce and for running on Sauce with extended debugging.

christian-bromann: https://github.com/webdriverio/webdriverio/issues/4196

PR #4204, opened 2019-07-16
Print link to Sauce Labs job details page when using Sauce Connect

### LINKED ISSUE #4197: @wdio/spec-reporter: make sure test link is also provided when running tests with SauceLabs using Sauce Connect
Currently if you run tests through Sauce Connect the link to the job details page is not provided as the [`getTestLink`](https://github.com/webdriverio/webdriverio/blob/master/packages/wdio-spec-reporter/src/index.js#L98) method checks the hostname in the config which can be localhost when running via Sauce Connect. Let's change this and make sure this also works in that case. A good additional indicator to check would be checking the capabilities if `sauce:options` is set or not (which has to be when running SC as a tunnelIdentifier has to be set).

christian-bromann:
## Proposed changes

When using Sauce Connect the spec reporter wouldn't provide the link to the job details page. This link fixes it.

## Types of changes

## Checklist

## Further comments

closes #4197

### DIFF
--- a/packages/wdio-spec-reporter/tests/index.test.js
+++ b/packages/wdio-spec-reporter/tests/index.test.js
@@ -15,6 +15,14 @@ import {
 
 const reporter = new SpecReporter({})
 
+const fakeSessionId = 'ba86cbcb70774ef8a0757c1702c3bdf9'
+const getRunnerConfig = (config) => {
+    return Object.assign({}, RUNNER, {
+        config,
+        sessionId: fakeSessionId
+    })
+}
+
 describe('SpecReporter', () => {
     let tmpReporter = null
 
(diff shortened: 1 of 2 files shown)


packages/wdio-spec-reporter/tests/index.test.js
```diff
@@ -167,7 +173,52 @@ describe('SpecReporter', () => {
...
+                },
+                sessionId: 'ba86cbcb70774ef8a0757c1702c3bdf9'
+            })
+            printReporter.printReport(runner)
+
+            expect(printReporter.write).toBeCalledWith(SAUCELABS_REPORT)
+        })
+
+        it('should print link to SauceLabs job details page if run with Sauce Connect (w3c)', () => {
+            printReporter.suiteUids = SUITE_UIDS
```
mgrybyk: can `printReporter` related things be moved out to before each?

packages/wdio-spec-reporter/tests/index.test.js
```diff
@@ -167,7 +173,52 @@ describe('SpecReporter', () => {
...
+            printReporter.suites = SUITES
+            printReporter.stateCounts = {
+                passed : 4,
+                failed : 1,
+                skipped : 1,
+            }
+
+            const runner = Object.assign({}, RUNNER, {
```
mgrybyk: I see that runner is defined most of the times with same/similar set of parameters.
Do you think it make sense to define a function for doing it to reduce amount of code?

christian-bromann:
@mgrybyk I improved the tests according to your suggestions, what do you think?

PR #4210, opened 2019-07-17
Allow to choose between WebDriver and Chrome DevTools Protocol as automation backend

christian-bromann:
## Proposed changes

Within the last years, we have seen broad adoption of the Chrome DevTools Protocol (CDP) in other browsers like Edge as well as support in Firefox. Therefor it is worth adding it as automation backend especially since [Puppeteer](https://pptr.dev) makes it really easy for us to have a protocol binding for it that works across all browser. This patch adds a new package called `devtools` which exposes the WebDriver interface to the user but runs all automation commands using Puppeteer. For example while the following script runs a WebDriver test:

```sh
chromedriver --port=4444 --url-base=/wd/hub --verbose
```
```js
const { remote } = require('./packages/webdriverio')

let client;

(async function () {
    client = await remote({
        capabilities: { browserName: 'chrome' }
    })

    await client.url('https://webdriver.io')
    console.log(await client.getTitle())
    await client.deleteSession()
})().catch(async (e) => {
    console.error(e.stack)
    await client.deleteSession()
})
```

with the output:

```
2019-07-17T00:37:07.040Z INFO webdriverio: Initiate new session using the webdriver protocol
2019-07-17T00:37:07.045Z INFO webdriver: [POST] http://localhost:4444/wd/hub/session
2019-07-17T00:37:07.045Z INFO webdriver: DATA { capabilities: { alwaysMatch: { browserName: 'chrome' }, firstMatch: [ {} ] },  desiredCapabilities: { browserName: 'chrome' } }
2019-07-17T00:37:08.850Z INFO webdriver: COMMAND navigateTo("https://webdriver.io/")
2019-07-17T00:37:08.853Z INFO webdriver: [POST] http://localhost:4444/wd/hub/session/7d6938784058d0dd0c05df4ec99dba2c/url
2019-07-17T00:37:08.853Z INFO webdriver: DATA { url: 'https://webdriver.io/' }
2019-07-17T00:37:10.721Z INFO webdriver: COMMAND getTitle()
2019-07-17T00:37:10.722Z INFO webdriver: [GET] http://localhost:4444/wd/hub/session/7d6938784058d0dd0c05df4ec99dba2c/title
2019-07-17T00:37:10.729Z INFO webdriver: RESULT WebdriverIO · Next-gen WebDriver test framework for Node.js
WebdriverIO · Next-gen WebDriver test framework for Node.js
2019-07-17T00:37:10.730Z INFO webdriver: COMMAND deleteSession()
2019-07-17T00:37:10.731Z INFO webdriver: [DELETE] http://localhost:4444/wd/hub/session/7d6938784058d0dd0c05df4ec99dba2c
```

By adding `automationProtocol: 'devtools'` as option, WebdriverIO will try to require the package and applies the same command interface as the `webdriver` package but with a different implementation where instead of making a WebDriver request, one or multiple Puppeteer commands are called. With this design the test looks the same but runs a complete different automation technology behind the scenes:

```js
const { remote } = require('./packages/webdriverio')

let client;

(async function () {
    client = await remote({
        automationProtocol: 'devtools', // CHANGE PROTOCOL HERE
        capabilities: { browserName: 'chrome' }
    })

    await client.url('https://webdriver.io')
    console.log(await client.getTitle())
    await client.deleteSession()
})().catch(async (e) => {
    console.error(e.stack)
    await client.deleteSession()
})
```

with the output:

```
2019-07-17T00:42:56.931Z INFO webdriverio: Initiate new session using the devtools protocol                            
2019-07-17T00:42:57.333Z INFO devtools: Launch Chrome with flags: --disable-translate --disable-extensions --disable-background-networking --disable-sync --metrics-recording-only --disable-default-apps --mute-audio --no-first-run --disable-hang-monitor --disable-prompt-on-repost --disable-web-resources --disable-client-side-phishing-detection --enable-logging --log-level=0 --password-store=basic --use-mock-keychain --test-type=webdriver --force-fieldtrials=SiteIsolationExtensions/Control --window-position=0,0 --window-size=1200,900                                                          
2019-07-17T00:42:58.252Z INFO devtools: Connect Puppeteer with browser on port 57823                                   
2019-07-17T00:42:59.264Z INFO devtools: COMMAND navigateTo("https://webdriver.io/")                                    
2019-07-17T00:43:00.439Z INFO remotedriver: RESULT null                                                                
2019-07-17T00:43:00.440Z INFO devtools: COMMAND getTitle()                                                             
2019-07-17T00:43:00.444Z INFO remotedriver: RESULT WebdriverIO · Next-gen WebDriver test framework for Node.js
WebdriverIO · Next-gen WebDriver test framework for Node.js                                                            
2019-07-17T00:43:00.444Z INFO devtools: COMMAND deleteSession()                                                        
2019-07-17T00:43:00.445Z INFO remotedriver: RESULT null
```

This is especially useful when it comes to local test execution where the browser is on the same machine as the test. Tests running on the DevTools protocol locally are not only faster, they tend to be more reliable due to the bidirectional connection and all Puppeteer magic. For the user nothing will change as the API will still work exactly the same.

Currently it is designed in a way that user have to "opt-in" to this. This means that if you want to run tests using CDP you will have to manually install the devtools package from NPM and apply the option to your config. At the moment `puppeteer-firefox` comes with a custom compiled Firefox. If we would make the `devtools` package a direct dependency all user would always download that custom Firefox binary even though they don't want to use it. Moving forward I think at some point Firefox will not be automatically shipped with the binary so we can make it a direct dependency (and maybe even the default mode for running tests locally since no driver is required anymore).

Currently I have named the CDP package `devtools`. Unfortunately this NPM package is already used by @bmeck. I reached out to him and asked if I could use it given that there hasn't been any releases for the last 5 years. (@bmeck if you read this please respond, would love to use this package name 🙏).

Since the CDP package will use Puppeteer under the hood users can easily switch between the WebdriverIO API and the Puppeteer one, to e.g. do use native browser APIs like tracing etc. This is how that looks like:

```js
// from https://github.com/GoogleChrome/puppeteer/blob/master/examples/pdf.js
const { remote } = require('./packages/webdriverio')

let client;

(async function () {
    client = await remote({
        automationProtocol: 'devtools',
        capabilities: { browserName: 'chrome' }
    })

    await client.url('https://news.ycombinator.com')
    const browser = client.puppeteerBrowser
    const page = (await browser.pages())[0]
    await page.pdf({
        path: 'hn.pdf',
        format: 'letter'
    })
    await client.deleteSession()
})().catch(async (e) => {
    console.error(e.stack)
    await client.deleteSession()
})
```

There are some WebDriver commands that can't be implemented with Puppeteer (e.g. `maximizeWindow`). In case there is a demand for that command we can try to find different ways to implement it, maybe NPM has a package for that too. Generally I don't think these commands are used much.

## Types of changes

Here is a general overview on all changes made in this PR:

- created a `@wdio/protocols` sub package that contains only the protocol json files to be consumed by `webdriver` and the new `devtools` package
- moved the monad implementation to `@wdio/utils` to allow it to get consumed by `webdriver` and the new `devtools` package
- created a new package called `devtools` that is build with the same interface as `webdriver` just with custom command implementation using Puppeteer
- implemented almost all WebDriver commands using Puppeteer

## Checklist

## Further comments

Things that I would like to add to this PR before merging:

- [x] allow to receive element ids from executeAsyncScript command
- [x] minor cleanups and refactorings
- [x] support xPath selector strategy
- [ ] unit tests to new `devtools` package
- [ ] e2e command tests on Chrome headless in CI
- [ ] documentation around the new automation protocol
- [ ] a blog post on this
- [ ] implement actions API with Puppeteer


packages/devtools/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "devtools",
+  "version": "5.7.5",
+  "description": "A WebDriver extension that runs commands on the DevTools protocol",
```
mgrybyk: Is it really a WebDriver extension? I think puppeteer has nothing to do with it, no?

packages/devtools/src/commands/status.js
```diff
@@ -0,0 +1,17 @@
+import path from 'path'
+
+const puppeteerPath = require.resolve('puppeteer-core')
+const puppeteerPkg = require(`${path.dirname(puppeteerPath)}/package.json`)
+const puppeteerFirefoxPath = require.resolve('puppeteer-firefox')
+const puppeteerFirefoxPkg = require(`${path.dirname(puppeteerFirefoxPath)}/package.json`)
+
+export default async function navigateTo () {
```
mgrybyk: I think function name should be `status`

packages/devtools/src/commands/deleteSession.js
```diff
@@ -0,0 +1,4 @@
+export default async function navigateTo () {
```
mgrybyk: I think function name should be `deleteSession`

packages/devtools/src/commands/getTitle.js
```diff
@@ -0,0 +1,4 @@
+export default async function navigateTo () {
```
mgrybyk: I think function name should be `getTitle`

packages/devtools/src/commands/getUrl.js
```diff
@@ -0,0 +1,4 @@
+export default async function navigateTo () {
```
mgrybyk: I think function name should be `getUrl`

packages/devtools/src/commands/getWindowHandles.js
```diff
@@ -0,0 +1,3 @@
+export default async function getWindowHandle () {
```
mgrybyk: I think function name should be `getWindowHandles` (`s` is missing in the end)

packages/devtools/src/commands/takeElementScreenshot.js
```diff
@@ -0,0 +1,12 @@
+export default async function takeScreenshot ({ elementId }) {
```
mgrybyk: I think function name should be `takeElementScreenshot`

christian-bromann:
@mgrybyk thanks for taking an initial look, I will address the issue as soon as possible

mgrybyk:
Will you be able to resolve conflicts as well?

christian-bromann:
🙌 I am owning the [`devtools`](https://www.npmjs.com/package/devtools) now

christian-bromann:
UPDATE: addressed issues mentioned by @mgrybyk , receiving elements from `executeAsync` still doesn't work but is next on my list (updated ToDo in the thread)

mgrybyk:
Thank you, ping me when PR is ready for review!

christian-bromann:
@mgrybyk besides documentation and unit tests I have fixed all outstanding issues. Can you please take another look?

packages/devtools/src/commands/createWindow.js
```diff
@@ -0,0 +1,12 @@
+import uuidv4 from 'uuid/v4'
+
+export default async function createWindow () {
```
mgrybyk: Create window doesn't work if no arguments were passed
```
Usage: createWindow(type)

Property Description:
  "type" (string): Set to 'tab' if the newly created window shares an OS-level window with the current browsing context, or 'window' otherwise.

For more info see https://w3c.github.io/webdriver/#new-window
```

I think there is something wrong with building protocol object as far as devtools command should override webdriver commands.

Also, types missing for this command, I mean the one where createWindow has no arguments AND where createWindow returns `string` but not `object`.

image

packages/devtools/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "url": "https://github.com/webdriverio/webdriverio/issues"
+  },
+  "dependencies": {
+    "@wdio/logger": "^5.7.0",
+    "@wdio/protocols": "^0.0.0",
+    "@wdio/utils": "^5.11.1",
+    "chrome-launcher": "^0.11.1",
+    "lodash.merge": "^4.6.2",
+    "puppeteer-core": "^1.18.1",
+    "puppeteer-firefox": "^0.5.0"
```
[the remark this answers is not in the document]
christian-bromann: Unfortunately yes, I will make a follow up PR in the puppeteer-firefox project to make this optional.

christian-bromann:
Fixed comments.

> also I have issues with test execution stuck, here are steps to reproduce

I added a timeout for such cases.

packages/devtools/src/scripts/executeScript.js
```diff
@@ -0,0 +1,36 @@
...
+ * is suppose to be an element handle
+ *
+ * @param  {HTMLElement} _            $eval fetched element
+ * @param  {String}      script       user script
+ * @param  {String}      dataProperty property name for elements being returned
+ * @param  {String}      dataFlag     flag for element
+ * @param  {Object[]}    args         user arguments for custom script
+ * @return {Object}                   result of custom script
+ */
+export default function executeScript (_, script, dataProperty, dataFlag, ...args) {
```
mgrybyk: this won't accept element as argument, it changes it to body if function is not arrow function.

It should be changed to same way as executeAsyncScript is designed.

`export default (_, script, dataProperty, dataFlag, ...args) => {`

packages/devtools/src/commands/createWindow.js
mgrybyk: @christian-bromann thank you for changing return type from string to object.

However argument is still required for `createWindow()`

If it cannot be done right now, we can create task for it

packages/devtools/src/index.js
```diff
@@ -0,0 +1,90 @@
...
+        const prototype = { ...protocolCommands, ...environmentPrototype, ...userPrototype }
+
+        const monad = webdriverMonad(params, modifier, prototype)
+        return monad(sessionId, customCommandWrapper)
+    }
+
+    static async reloadSession (instance) {
+        const { session } = sessionMap.get(instance.sessionId)
+        const { w3cCaps } = instance.options.requestedCapabilities
+        const browser = await launch(w3cCaps)
```
mgrybyk: reload session create browser with more and more times every time!
startup: 1 tab
reloadSession: 3 tabs (+2 tabs)
reloadSession: 5 tabs (+2 tabs)
...

mgrybyk: wrong capabilities passed here, I think.

startup capabilities:
```
{ browserName: 'chrome',
  'goog:chromeOptions':
   { w3c: true,
     args:
      [ '--window-size=1920,1080',
        '--disable-translate',
        '--disable-extensions',
        '--disable-background-networking',
        '--disable-sync',
        '--disable-default-apps',
        '--mute-audio',
        '--no-first-run',
        '--disable-prompt-on-repost' ] } }
```

reload session capabilities:
```
{ browserName: 'Chrome',
  browserVersion: '75.0.3770.142',
  platformName: 'darwin',
  platformVersion: '18.6.0' }
```

packages/devtools/src/commands/createWindow.js
christian-bromann: > However argument is still required for `createWindow()`

an argument is also required if you use WebDriver.

packages/devtools/src/constants.js
```diff
@@ -0,0 +1,78 @@
...
+     * maxConnectionRetries in chrome-launcher
+     */
+    connectionRetryCount: {
+        type: 'number',
+        default: 50
+    }
+}
+
+export const DEFAULT_IMPLICIT_TIMEOUT = 0
+export const DEFAULT_PAGELOAD_TIMEOUT = 5 * 60 * 1000 // 5 min
```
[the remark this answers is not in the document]
christian-bromann: I have taken this from Chromedriver, I can the values from Geckodriver

packages/devtools/src/commands/elementClick.js
```diff
@@ -0,0 +1,22 @@
...
+    const elementHandle = this.elementStore.get(elementId)
+
+    if (!elementHandle) {
+        throw new Error(`Couldn't find element with id ${elementId} in cache`)
+    }
+
+    /**
+     * ensure to fulfill the click promise if the click has triggered an alert
+     */
+    await new Promise((resolve) => {
```
mgrybyk: if element is no longer visible or something is wrong I'm getting
```
[0-0] (node:10140) UnhandledPromiseRejectionWarning: Error: Node is either not visible or not an HTMLElement
    at ElementHandle._clickablePoint (/Users/mygr/sources/e2e/node_modules/puppeteer-core/lib/JSHandle.js:199:13)
    at process._tickCallback (internal/process/next_tick.js:68:7)
  -- ASYNC --
    at ElementHandle.<anonymous> (/Users/mygr/sources/e2e/node_modules/puppeteer-core/lib/helper.js:111:15)
    at resolve (/Users/mygr/sources/e2e/node_modules/devtools/build/commands/elementClick.js:26:26)
    at new Promise (<anonymous>)
    at DevToolsDriver.elementClick (/Users/mygr/sources/e2e/node_modules/devtools/build/commands/elementClick.js:22:9)
    at Element.<anonymous> (/Users/mygr/sources/e2e/node_modules/devtools/build/devtoolsdriver.js:88:47)
    at Element.runCommand (/Users/mygr/sources/e2e/node_modules/@wdio/sync/build/wrapCommand.js:101:21)
    at Element.wrapCommandFn (/Users/mygr/sources/e2e/node_modules/@wdio/sync/build/wrapCommand.js:35:25)
    at Element.runCommand (/Users/mygr/sources/e2e/node_modules/@wdio/sync/build/wrapCommand.js:101:21)
    at Element.wrapCommandFn (/Users/mygr/sources/e2e/node_modules/@wdio/sync/build/wrapCommand.js:35:25)
    at Element.elementErrorHandlerCallbackFn (/Users/mygr/sources/e2e/node_modules/webdriverio/build/middlewares.js:27:57)
[0-0] (node:10140) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:10140) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

Can we catch the error or something?

packages/devtools/src/commands/createWindow.js
mgrybyk: Let me clarify, the argument is not needed for devtools but required for webdriver. 

Typescript types we have assumes that argument is always required. 
We need to extend this to make argument optional when using devtools.

We can ship these types along with devtools package to avoid mixing everything

christian-bromann: I modified the command so that it requires a value now. WebDriver and DevTools need to have the same interface.

mgrybyk: Can you point me to that change? I can't see arg here
`export default async function createWindow () {`

mgrybyk: thank you, got it!

packages/devtools/src/commands/isElementDisplayed.js
```diff
@@ -0,0 +1,9 @@
+export default function isElementDisplayed ({ elementId }) {
```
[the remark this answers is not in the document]
christian-bromann: let me change the behavior in WebdriverIO since Puppeteer can actually determine visibility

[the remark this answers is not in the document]
christian-bromann: so we will call `isElementDisplayed` in the `isVisibleWithinViewport` command of WebdriverIO so that the command is not unused.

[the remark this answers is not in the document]
mgrybyk: Can't understand fully, why don't you just change a bit condition in isDisplayed.js to call protocol command instead of using script?
Am I understanding something wrong?

packages/devtools/src/commands/getElementAttribute.js
```diff
@@ -0,0 +1,10 @@
+export default async function getElementAttribute ({ elementId, name }) {
+    const elementHandle = this.elementStore.get(elementId)
+
+    if (!elementHandle) {
+        throw new Error(`Couldn't find element with id ${elementId} in cache`)
+    }
+
+    const jsHandle = await elementHandle.getProperty(name)
+    return jsHandle.jsonValue()
```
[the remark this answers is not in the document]
christian-bromann: What do you mean by more than 1 attribute?

packages/devtools/src/commands/getAllCookies.js
```diff
@@ -0,0 +1,4 @@
+export default async function getAllCookies () {
+    const page = this.windows.get(this.currentWindowHandle)
+    return page.cookies()
```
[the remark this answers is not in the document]
christian-bromann: yeah, seen that error too. I think it is a bug in Puppeteer.

packages/devtools/src/commands/createWindow.js
```diff
@@ -0,0 +1,12 @@
...
+    const newPage = await this.browser.newPage()
+    const handle = uuidv4()
+
+    await newPage.bringToFront()
```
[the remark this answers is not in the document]
christian-bromann: I assume, yes!

packages/devtools/src/commands/isElementDisplayed.js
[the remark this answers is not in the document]
christian-bromann: The puppeteer isVisible protocol command can just check if the element is within the viewport: https://pptr.dev/#?product=Puppeteer&version=v1.19.0&show=api-elementhandleisintersectingviewport

packages/devtools/src/commands/getElementAttribute.js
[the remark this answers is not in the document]
mgrybyk: I mean attribute value
Foo=bar1 bar2 bar3

Results array of bar1 bar2 bar3 instead of string, as in webdriver protocol

packages/devtools/src/commands/isElementDisplayed.js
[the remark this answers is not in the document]
mgrybyk: I don't mind, but it won't be used neither if we don't change condition in webdriverio isDisplayed.js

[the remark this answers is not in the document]
mgrybyk: I don't mind, but it won't be used neither if we don't change condition in webdriverio isDisplayed.js

packages/devtools/src/commands/getElementAttribute.js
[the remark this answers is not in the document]
mgrybyk: Actually we should use something different from `getProperty`.
https://github.com/GoogleChrome/puppeteer/issues/1451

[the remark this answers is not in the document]
mgrybyk: @christian-bromann here is what is working for me
```
const page = this.windows.get(this.currentWindowHandle);
    const result = await page.$eval('html', (html, elem, attr) => {
        return elem.getAttribute(attr);
    }, elementHandle, name);
    return result || ''
```

packages/devtools/src/commands/isElementDisplayed.js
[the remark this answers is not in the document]
christian-bromann: @mgrybyk I added a condition is WebdriverIOs `isVisibleWithinViewport` command

packages/devtools/src/commands/createWindow.js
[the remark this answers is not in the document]
mgrybyk: marking it as resolved according to https://aslushnikov.github.io/ispuppeteerfirefoxready/

packages/devtools/src/commands/getAllCookies.js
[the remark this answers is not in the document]
mgrybyk: cookies seems to be implemented https://aslushnikov.github.io/ispuppeteerfirefoxready/ and I don't see any issue here, I might be missing something https://github.com/GoogleChrome/puppeteer/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+cookies

packages/devtools/src/commands/elementClick.js
christian-bromann: fixed the issue and made it so it throws a proper stale element exception

christian-bromann:
@mgrybyk I implemented the xPath selector strategy which should now allow to use commands like `selectByAttribute`, `selectByIndex` or `selectByVisibleText`. Next is to implement `link text` and `partial link text` and then move on with writing tests for this beast. Another review would be appreciated 🙏

packages/devtools/src/commands/switchToFrame.js
```diff
@@ -0,0 +1,46 @@
+import { switchFrame } from '../utils'
+
+export default async function switchToFrame ({ id }) {
```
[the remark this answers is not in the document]
mgrybyk: @christian-bromann any update on this one?

packages/devtools/src/commands/getElementProperty.js
```diff
@@ -0,0 +1,5 @@
+import getElementAttribute from './getElementAttribute'
+
+export default async function getElementProperty (...args) {
+    return getElementAttribute.apply(this, args)
```
mgrybyk: @christian-bromann we can't use `getElementAttribute` here.

It won't work, for example, when getting `value`.
`<input value="qwe">`
lets type _foobar_, see `value` attribute, it is still _qwe_, however input.value is _foobar_

See also https://pptr.dev/#?product=Puppeteer&version=v1.19.0&show=api-jshandlegetpropertypropertyname

packages/devtools/src/constants.js
```diff
@@ -0,0 +1,86 @@
+export const DEFAULT_WIDTH = 1200
+export const DEFAULT_HEIGHT = 900
+export const DEFAULT_X_POSITION = 0
+export const DEFAULT_Y_POSITION = 0
+
+export const ELEMENT_KEY = 'element-6066-11e4-a52e-4f735466cecf'
+
+export const DEFAULT_FLAGS = [
```
mgrybyk: what do you think on adding `--disable-features=NetworkService` to flags?

christian-bromann: What does it do?

packages/devtools/src/constants.js
```diff
@@ -0,0 +1,86 @@
...
+    // Mute any audio
+    '--mute-audio',
+    // Skip first run wizards
+    '--no-first-run',
+    // chromedriver flags
+    '--disable-hang-monitor',
+    '--disable-prompt-on-repost',
+    '--disable-web-resources',
+    '--disable-client-side-phishing-detection',
+    '--enable-logging',
```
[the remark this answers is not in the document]
christian-bromann: it is copied from what Chromedriver is using

packages/devtools/src/constants.js
```diff
@@ -0,0 +1,86 @@
...
+    '--log-level=0',
+    '--password-store=basic',
+    '--use-mock-keychain',
+    '--test-type=webdriver',
```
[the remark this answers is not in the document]
christian-bromann: not really sure what this flag stands for .. I assume it shows this message at the top saying that this browser is controlled by an automation driver, that's it.

package.json
```diff
@@ -33,6 +34,7 @@
     "test:smoke": "babel-node ./tests/smoke.runner.js",
     "ts": "run-p ts:*",
     "ts:sync": "cd tests/typings/sync && tsc --incremental",
+    "ts:sync": "cd tests/typings/devtools && tsc --incremental",
```
mgrybyk: should be ts:devtools

packages/devtools/src/commands/switchToFrame.js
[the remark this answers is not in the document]
christian-bromann: `id` is what the [protocol](https://github.com/webdriverio/webdriverio/blob/master/packages/webdriver/protocol/webdriver.json#L207-L212) expects to receive as param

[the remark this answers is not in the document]
mgrybyk: It's just a name, id can be an element object that has certain properties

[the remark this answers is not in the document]
christian-bromann: yeah but it is defined like this is in the protocol json. Are you referring to a different command maybe? This command has no idea about any WebdriverIO concepts.

christian-bromann:
This now ready to get the final review. I will follow up with a blog post as well as more documentation what I want to separate this out given the size this PR already has gotten. I it helps I can push everything that I did in devtools in a different PR.

packages/devtools/src/commands/switchToFrame.js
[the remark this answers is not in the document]
mgrybyk: Sorry to say can't verify right now. I just want switch to frame work same as in webdriverio without a need to somehow hack devtools command to make it work.
I have a gist with switch to iframe example
https://gist.github.com/mgrybyk/4a1cdc60323caa041cc6465fcdde3576

mgrybyk:
Will you be able to fix get element attribute? It doesn't work as expected https://github.com/webdriverio/webdriverio/pull/4210/files#discussion_r309273860

packages/devtools/src/commands/switchToFrame.js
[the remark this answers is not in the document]
mgrybyk: And the problem is that currently switch to frame fails when passing element to it

christian-bromann:
> Will you be able to fix get element attribute?

Done

packages/devtools/src/scripts/getElementAttribute.js
```diff
@@ -0,0 +1,3 @@
+export default function getElementCSSValue (_, elem, attr) {
```
mgrybyk: Can you please rename function after copying?

mgrybyk:
@christian-bromann sorry for not informative review, have to run soon, here are some issues I've faced:

1. 
```
el.click()
expect(el.isFocused()).eq(true)
expect(el.isSelected()).eq(false)
```
getting `expected null to equal false`

2. all selectBy* methods don't work for me, getting `Node is either not visible or not an HTMLElement`: selectByVisibleText, selectByAttribute, selectByIndex

3. I'm still not able to switch to iframe `browser.switchToFrame($("#result iframe"))`
```
Could not switch frame, unknwon id: [object Object]
Error: Could not switch frame, unknwon id: [object Object]
     at DevToolsDriver.switchToFrame (node_modules/devtools/src/commands/switchToFrame.js:45:11)
```

gist used for testing https://gist.github.com/mgrybyk/4a1cdc60323caa041cc6465fcdde3576

christian-bromann:
> getting `expected null to equal false`

fixed with last commit

> 1. all selectBy* methods don't work for me, getting `Node is either not visible or not an HTMLElement`: selectByVisibleText, selectByAttribute, selectByIndex

This is indeed an issue because clicking like in WebDriver doesn't work in Puppeteer. I will have to find a way to fix this.

> 2\. I'm still not able to switch to iframe `browser.switchToFrame($("#result iframe"))`

Was able to reproduce, working on a fix.

packages/devtools/src/commands/isElementSelected.js
```diff
@@ -0,0 +1,9 @@
+import getElementAttribute from './getElementAttribute'
+import getElementTagName from './getElementTagName'
+
+export default async function isElementSelected ({ elementId }) {
+    const tagName  = await getElementTagName.call(this, { elementId })
+    const name = tagName === 'option' ? 'selected' : 'checked'
+    const isSelected = await getElementAttribute.call(this, { elementId, name })
```
mgrybyk: I think we should use `getElementProperty` here instead of get element attribute.
Select option doesn't necessarily has `selected` attributed even if selected, while property `selected` always reflects actual state

packages/devtools/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "devtools",
+  "version": "5.7.5",
```
mgrybyk: will the version automatically be updated on release to 5.12 (or higher)?

christian-bromann: yes, once I release it in a bundle with the other projects

christian-bromann:
@mgrybyk I was able to fix all issues you pointed out!

mgrybyk:
Thank you, however I won't be able verify it this week.
I wanted also to verify some other cases like perform actions, especially after scrolling.
I don't mind merging it sooner and fixing minor issues later on

christian-bromann:
> Thank you, however I won't be able verify it this week.

No worries! I am working on the e2e tests right now. I would like to have at least one simple test for every command to ensure it behaves as described in the protocol.

packages/devtools/src/commands/acceptAlert.js
```diff
@@ -0,0 +1,8 @@
+export default async function acceptAlert () {
+    if (!this.activeDialog) {
+        throw new Error('no such alert')
+    }
+
+    await this.activeDialog.accept()
+    return null
```
erwinheitzman: Why return `null`everywhere, can't we return undefined like it does by default if you do not return anything?

packages/devtools/src/commands/addCookie.js
```diff
@@ -0,0 +1,10 @@
+export default async function addCookie ({ cookie }) {
+    const page = this.windows.get(this.currentWindowHandle)
+
+    if (!Object.keys(cookie).includes('name') || !Object.keys(cookie).includes('value')) {
```
erwinheitzman: Would be cleaner and less repetitive if we do it like this
```
const { name, value } = Object.keys(cookie)

if (!name || !value) {
    throw new Error('Provided cookie object is missing either "name" or "value" property')
}
```

packages/devtools/src/commands/back.js
```diff
@@ -0,0 +1,5 @@
+export default async function back () {
+    const page = this.windows.get(this.currentWindowHandle)
```
erwinheitzman: no need to store something that is used once

packages/devtools/src/commands/deleteCookie.js
```diff
@@ -0,0 +1,11 @@
+export default async function deleteCookie ({ name }) {
+    const page = this.windows.get(this.currentWindowHandle)
+    const cookies = await page.cookies()
+    const cookieToDelete = cookies.find((cookie) => cookie.name === name)
+
+    if (cookieToDelete) {
```
erwinheitzman: should we throw if cookie is not found?

packages/devtools/src/commands/elementClear.js
```diff
@@ -0,0 +1,13 @@
+import command from '../scripts/elementClear'
+
+export default async function elementClear ({ elementId }) {
+    const elementHandle = this.elementStore.get(elementId)
+
+    if (!elementHandle) {
```
erwinheitzman: if it's not in cache, should we fetch it?

packages/devtools/src/commands/executeScript.js
```diff
@@ -0,0 +1,37 @@
...
+        script,
+        SERIALIZE_PROPERTY,
+        SERIALIZE_FLAG,
+        ...transformExecuteArgs.call(this, args)
+    )
+
+    const timeoutPromise = new Promise((_, reject) => {
+        setTimeout(
+            () => {
+                const additionalTimeoutInfo = (
```
erwinheitzman: this is `false` or a `string`, I think we should use a ternary here so it will either return the said `string`or a empty `string`

erwinheitzman: then we can also simplify the code below it:
```
return reject(new Error(`script timeout ${additionalTimeoutInfo}`))
```

packages/devtools/src/commands/findElement.js
```diff
@@ -0,0 +1,35 @@
...
+export default async function findElement ({ using, value }) {
+    if (!SUPPORTED_SELECTOR_STRATEGIES.includes(using)) {
+        throw new Error(`selector strategy "${using}" is not yet supported`)
+    }
+
+    const page = this.windows.get(this.currentWindowHandle)
+    let needsCleanUp = false
+
+    if (using === 'xpath') {
+        const foundElement = await page.$eval('html', findElementByXPath, value, null, SERIALIZE_PROPERTY)
```
erwinheitzman: nothing is done with the element found?

packages/devtools/src/commands/elementClick.js
```diff
@@ -0,0 +1,39 @@
...
+        throw new Error(`Couldn't find element with id ${elementId} in cache`)
+    }
+
+    /**
+     * in order to allow clicking on option elements (as WebDriver does)
+     * we need to check if the element is such a an element and select
+     * it instead of actually executing the click
+     */
+    const tagName = await getElementTagName.call(this, { elementId })
+    if (tagName === 'option') {
```
mgrybyk: I'm a bit worried about this change.
Isn't it possible that user may want to actually click on `option` element instead of making it selected?

Maybe select script should be used only when selectBy* commands are called?

christian-bromann: > Isn't it possible that user may want to actually click on `option` element instead of making it selected?

Well if the user clicks on the element he will select it which is what happens here.

packages/devtools/src/commands/acceptAlert.js
christian-bromann: Please see the [protocol](https://w3c.github.io/webdriver/), it all of these cases it says:

> Return success with data null.

packages/devtools/src/commands/deleteCookie.js
christian-bromann: no, there is nothing that explicitly says that in the protocol

packages/devtools/src/commands/elementClear.js
christian-bromann: No, this functionality is handled in WebdriverIO. We should stay as close to the WebDriver protocol as possible.

packages/devtools/src/commands/findElement.js
christian-bromann: nope, the way it works is that we use `$eval` to run some JS that finds elements using an xPath. All these elements get a data attribute so that they can be found using CSS selectors in a secondary call. At the end we clean up the data attributes and return the value.

christian-bromann: `foundElement` is a boolean in this case

christian-bromann: I will add a comment

christian-bromann:
@mgrybyk @erwinheitzman addressed comments

packages/devtools/src/commands/elementClick.js
mgrybyk: yeah, but some implementations expects mouse click event to be triggered.
We can do same for input type checkbox and just change `checked` flag instead of clicking on it...
I just wanted to say that it looks a bit hacky to me even though I fully agree that such implementation is way simpler. 
Click is click, and we are mixing click logic with selecting option.

christian-bromann: > Click is click, and we are mixing click logic with selecting option.

In my first implementation I had an actual click but unfortunately this was throwing an error which is why I switched to this option. I doubt that developer have click listeners on option elements. Usually you would register a onChange handler on the select element.

mgrybyk: I'm resolving this for now then if no one else minds

christian-bromann:
sure thing

package.json
```diff
@@ -31,8 +32,10 @@
     "test:typings": "npm run generate:typings && node tests/typings/copy && npm run ts && npm run clean:tests",
     "test:coverage": "jest --coverage",
     "test:smoke": "babel-node ./tests/smoke.runner.js",
+    "test:e2e": "jest --config e2e/jest.config.js --detectOpenHandles",
```
mgrybyk: Detect open handles prevents from running tests in parallel, I think it can be removed as far as the root cause has been fixed

christian-bromann: done

packages/devtools/src/scripts/executeAsyncScript.js
```diff
@@ -0,0 +1,45 @@
...
+
+            if (tmpResult.find((r) => r instanceof HTMLElement)) {
+                tmpResult = tmpResult.map((r, i) => {
+                    if (r instanceof HTMLElement) {
+                        const dataPropertyValue = `${dataFlag}_${i}`
+                        r.setAttribute(dataProperty, dataPropertyValue)
+                        return dataPropertyValue
+                    }
+
+                    return result
```
Krinkle: Should this be `return r` instead?

christian-bromann: good question, our e2e tests are fine and return elements properly. Are you experiencing a bug?

Krinkle: This intends to map a result array and change the ones that are HTMLElement, and presumably leave the others unchanged.

However, I would expect the following to fail:

```js
result = [
 HTMLDivElement,
 42,
 HTMLAnchorElement
];
```

This would become:
```js
result = [
  dataValue,
  [ HTMLDivElement, 42, HTMLAnchorElement ],
  dataValue
]
```
instead of

```js
result = [
  dataValue,
  42,
  dataValue
]
```

I'm still in the midst of upgrading to v5 and (after that) trying out devtools, so I haven't yet run into this (potential) bug. Was reviewing the changes and came across this.

Issue #3877, opened 2019-04-23
MaxListenersExceededWarning: Possible EventEmitter memory leak detected


StephenABoyd:
[//]: # NOTE: This repository only maintains packages that are listed in the [Readme](https://github.com/webdriverio/webdriverio/blob/master/README.md#packages). Please make sure that your issue is directly caused by one of these packages and if not file an issue in the correct 3rd party package repository.

**Environment (please complete the following information):**
 - **WebdriverIO version:** 5.7.14
 - **Mode:** WDIO Testrunner
 - **If WDIO Testrunner, running sync/async:** sync
 - **Node.js version:** 10.9.0
 - **NPM version:** 6.2.0
 - **Browser name and version:** Chrome 74
 - **Platform name and version:** Windows NT 10
 - **Additional wdio packages used (if applicable):** 
@wdio/junit-reporter, 
@wdio/cli, 
@wdio/mocha-framework, 
@wdio/local-runner, 
@wdio/spec-reporter, 
@wdio/sync

**Describe the bug**
During my CI build I'm seeing this happen. The build continues as normal after this
```
(node:4488) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 end listeners added. Use emitter.setMaxListeners() to increase limit
(node:4488) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 unpipe listeners added. Use emitter.setMaxListeners() to increase limit
(node:4488) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 drain listeners added. Use emitter.setMaxListeners() to increase limit
(node:4488) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 error listeners added. Use emitter.setMaxListeners() to increase limit
(node:4488) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 close listeners added. Use emitter.setMaxListeners() to increase limit
(node:4488) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 finish listeners added. Use emitter.setMaxListeners() to increase limit
(node:4488) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 unpipe listeners added. Use emitter.setMaxListeners() to increase limit
(node:4488) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 drain listeners added. Use emitter.setMaxListeners() to increase limit
[...]

PR #4223, opened 2019-07-19
Fix MaxListenersExceededWarning

mgrybyk:
## Proposed changes

Fix `MaxListenersExceededWarning: Possible EventEmitter memory leak detected.` 
#3877 

## Types of changes

## Checklist

## Further comments

Certain event listeners are added automatically when Readable / Writable stream is piped.
As far as we pipe all worker streams with process stdout, stderr and stdin, lots of event listeners are created.

### DIFF
--- a/packages/wdio-local-runner/src/stdStream.js
+++ b/packages/wdio-local-runner/src/stdStream.js
@@ -1,0 +1,28 @@
+import { Transform } from 'stream'
+import { removeLastListener } from './utils'
+
+export default class RunnerStream extends Transform {
+    constructor () {
+        super()
+
+        /**
+         * Remove events that are automatically created by Writable stream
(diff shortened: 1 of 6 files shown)


nickwangyoui:
Is there any change this could go into 4.x? We have plans to migrate to 5.x, but we have a bunch of suites and it would be helpful if this worked for our existing projects :x

christian-bromann:
> Is there any change this could go into 4.x?

This issue was specific on v5 and might have a different reason in v4. Feel free to propose a PR for v4 but I would recommend to spend that time to upgrade to v5.

nickwangyoui:
Ah! I see, I'll have to investigate further then, thank you for the quick reply @christian-bromann.

PR #4229, opened 2019-07-21
Browser/element prototype performance improvements

mgrybyk:
## Proposed changes

`lodash.merge` is a bit slow or not needed in some cases. 
Use `Object.assign` or [spread(...)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) where possible.

perf improvement measurements in real project:
- minor, when getting element `$()`: **8%** 
- **Super-duper** significant, when `$$("*")[49].$('not-existing')`: **~100000%**

## Types of changes

## Checklist


mgrybyk:
Added one more fix for issue with getting elements:
```
browser.url("http://adam.goucher.ca/parkcalc/")
const elems = $$("*") // 50 elements
const elem = elems[49] // index is important!!
const elemC = elem.$("not-existing")
```

before fix: calls - more than 8573, duration - more than 3 minutes!! 
after fix: calls - 2, duration - 80

performance improvement - ∞, lol

packages/wdio-runner/tests/index.test.js
```diff
@@ -3,10 +3,14 @@ import fs from 'fs'
 import { attach } from 'webdriverio'
 import WDIORunner from '../src'
 
-jest.mock('fs')
+const writeFile = fs.writeFile
```
christian-bromann: What is the reason for this change?

packages/webdriverio/tests/commands/browser/saveRecordingScreen.test.js
```diff
@@ -1,11 +1,16 @@
+import fs from 'fs'
 import request from 'request'
 import { remote } from '../../../src'
 import * as utils from '../../../src/utils'
 
+const writeFileSync = fs.writeFileSync
```
christian-bromann: same

mgrybyk: let's proceed convo above

packages/wdio-runner/tests/index.test.js
mgrybyk: the reason is that we not can't mock `fs` now before `webdriverio/src/utils.js` has been called.
https://github.com/webdriverio/webdriverio/pull/4229/files#diff-fe019f9e87ddc9284183a1356b00a163

christian-bromann: I don't understand, why is that? Afaik doing 

```js
const writeFile = fs.writeFile
// ...
    beforeAll(() => {
        fs.writeFile = jest.fn()
    })
```
is the same as if we would do:

```js
jest.mock('fs')
```
right away!

mgrybyk: @christian-bromann taking a look.

packages/webdriverio/tests/__mocks__/fs.js
```diff
@@ -1,4 +1,7 @@
+const { readdirSync } = jest.requireActual('fs')
 const fs = jest.genMockFromModule('fs')
 fs.writeFileSync = jest.fn()
 fs.existsSync = jest.fn(() => true)
+fs.readdirSync.mockImplementation(readdirSync)
```
mgrybyk: have to use original function here to be able to read commands from file system

packages/webdriverio/tests/commands/browser/saveRecordingScreen.test.js
```diff
@@ -1,10 +1,10 @@
 import request from 'request'
-import { remote } from '../../../src'
```
mgrybyk: have to move imports after `fs` is mocked, otherwise can't mock properly `fs` imports inside commands.

mgrybyk:
@christian-bromann updated `fs` mocking according to your comment.

packages/webdriverio/tests/commands/browser/uploadFile.test.js
```diff
@@ -1,9 +1,8 @@
 import path from 'path'
 import archiver from 'archiver'
 
-import { remote } from '../../../src'
-
 describe('uploadFile', () => {
+    const { remote } = require('../../../src')
```
christian-bromann: Why is it here other way around? Because in this test we actually want to have the mocked version and in others we don't?

mgrybyk: I'll try to explain how it works after my changes in case we keep `remote` import as it was before.

When `remote` is imported all the commands are loaded from file system.
If `fs` was fully mocked before no commands are loaded.
If `fs` was not mocked and `remote` is imported, then there is no way to mock https://github.com/webdriverio/webdriverio/blob/3aa49945eba252aa70641088c8bf816129b49ccd/packages/webdriverio/src/commands/browser/uploadFile.js#L25 because commands were loaded before `fs` is mocked.

So I decided to mock fs, restore readdirSync and then import everything else. I can dig a bit more into it in the evening.

If you have any hints on how can this be improved please let me know

mgrybyk:
@christian-bromann thank you for review!

christian-bromann:
So all of this is necessary because of the change in `packages/webdriverio/src/utils.js`, right? What is the performance impact when we load the commands within the function call instead of the module level. Is this the reason why the unit tests has become slow? If so can we use some sort of caching here?

mgrybyk:
1. Yes
2. In real project I'm getting better performance. Not sure regarding unit tests. 
3. I'll take a look why unit tests become slower.
4. Caching is a good idea!

christian-bromann:
> 3\. I'll take a look why unit tests become slower.

I was under the assumption that due to slower unit tests you decided to make the changes to `packages/webdriverio/src/utils.js`. If we don't need them and would remove them we also can remove the changes to all test files here.

mgrybyk:
@christian-bromann changes to src/utils.js are required to improve performance of real life project, not tests themselves. It really improves performance of $('') by 5% :)

Anyway, I have measured performance of unit tests before and after my changes, here are the results:
before changes:
```
real    1m23.186s
user    1m43.025s
sys     0m6.957s
```
after changes
```
real    1m24.893s
user    1m43.294s
sys     0m7.739s
```

christian-bromann:
> changes to src/utils.js are required to improve performance of real life project, not tests themselves. It really improves performance of $('') by 5% :)

Can we do some research as to why that is? Technically this operation (reading all JS files to apply the functions to the prototype) should happen only once in the worker process. If we (have to) do it multiple times we should work with some caching mechanism do not have to do this more.

mgrybyk:
@christian-bromann before it was 
```
const applyScopePrototype = (prototype, scope) => {
    const dir = path.resolve(__dirname, 'commands', scope)
    const files = fs.readdirSync(dir)
    for (let filename of files) {
        const commandName = path.basename(filename, path.extname(filename))
        prototype[commandName] = { value: require(path.join(dir, commandName)).default }
    }
}
```

it is a regular function which is called for every element/browser instance and obviously it would read files from file system for every `$('foo')`.

And I want to call it, as you have mentioned, only once per worker

mgrybyk:
test used to measure performance of getting element:
```
describe("perf", () => {
    it("getting $ performance", () => {
        browser.url("http://adam.goucher.ca/parkcalc/")
        $('body').waitForExist()
        browser.pause(2000)

        const start = Date.now()
        for (let i = 0; i < 100; i++) {
            $('body')
        }
        console.log(Date.now() - start)

        // tests on my local machine
        // before:  1200
        // after:   1100
        //
        // (before - after) / after * 100 = 8%
    })
})
```

packages/webdriverio/src/utils.js
```diff
@@ -10,16 +10,21 @@ import { URL } from 'url'
...
-    for (let filename of files) {
-        const commandName = path.basename(filename, path.extname(filename))
-        prototype[commandName] = { value: require(path.join(dir, commandName)).default }
-    }
+const scopes = {
+    browser: browserCommands,
+    element: elementCommands
+}
+
+const applyScopePrototype = async (prototype, scope) => {
```
CrispusDH: Should arrow function be `async`?

mgrybyk: no, it's a defect

mgrybyk: @CrispusDH thank you, fixed

mgrybyk:
if there are no more other comments, it can be merged

Issue #4231, opened 2019-07-22
Speedup build


mgrybyk:
**Is your feature request related to a problem? Please describe.**
`npm run build` takes much time

**Describe the solution you'd like**
`npm run build` should not take that much time. 
We might think on speeding up build by tuning babel config and/or moving some packages outside of monorepo.

I think packages lots of packages can be moved out, like reporters, or/and only be built under certain condition (locally, I mean).

Maybe we can tag packages somehow, ex: core-package, feature-packages, or something.

**Describe alternatives you've considered**
Any alternative to speedup build

christian-bromann:
> I think packages lots of packages can be moved out

There are definitely packages that haven't changed a lot but given all the advantages we got from moving them into a mono-repo would get lost if we would move them out again. I agree that setting up the project by running `npm run setup` is slow and uses a lot of CPU but moving out packages would be my last option. I rather would see if we can build something that improves our current situation (NPM cache, build improvements etc.).

PR #4237, opened 2019-07-23
speedup build, clean, bootstrap

mgrybyk:
## Proposed changes

changed commands:
- speed up build a bit by running it in parallel
- speed up clean up a bit by reducing amount of `rimraf` calls

added commands:
- `bootstrap-fast` which is super fast because it is not cleaning up node_modules of each package;
- `build-fast` which is a bit faster because calls `rimraf` only once, but is less safe because there might be some build steps missed in future
- `setup-fast` it is a link to bootstrap-fast and build-fast

### Testing

_Assuming `npm run setup` has been performed_

`npm run bootstrap` - 47s
`npm run bootstrap-fast` - 2s

`npm run build` - before: 19s, after: 14s
`npm run build-fast` - 9s

## Types of changes

## Checklist

## Further comments

closing #4231 for now


mgrybyk:
@christian-bromann done

package.json
```diff
@@ -2,20 +2,24 @@
   "private": true,
   "scripts": {
     "create": "node ./scripts/generateSubPackage",
-    "bootstrap": "npm run clean && lerna bootstrap --no-ci && npm run link",
-    "build": "lerna run build",
+    "bootstrap-full": "npm run clean && lerna bootstrap --no-ci && npm run link",
+    "bootstrap": "run-p clean:* && lerna bootstrap --force-local && npm run link",
```
christian-bromann:
```suggestion
    "bootstrap": "run-s clean:* && lerna bootstrap --force-local && npm run link",
```

I don't think it makes sense to clean packages while lerna is building them

mgrybyk: Only clean is running in parallel, starting from && the next command goes

mgrybyk: Only clean is running in parallel, starting from && the next command goes

christian-bromann: ah right

Issue #4244, opened 2019-07-24
wdio-sync getAlertText unloads synchronous commands?


Lusito:
**Environment (please complete the following information):**
 - **WebdriverIO version:** 5.11.7
 - **Mode:** WDIO Testrunner
 - **If WDIO Testrunner, running sync/async:** sync
 - **Node.js version:** 10.15.1
 - **NPM version:** 6.9.0
 - **Browser name and version:** chrome 75
 - **Platform name and version:** ubuntu 18.04
 - **Additional wdio packages used (if applicable):**

```
"@wdio/config": "^5.11.0",
"@wdio/cucumber-framework": "^5.11.7",
"@wdio/junit-reporter": "^5.11.7",
"@wdio/local-runner": "^5.11.7",
"@wdio/selenium-standalone-service": "^5.11.2",
"@wdio/spec-reporter": "^5.11.7",
"@wdio/sync": "^5.11.7",
"chromedriver": "^75.1.0",
"cucumber": "^5.1.0",
"wdio-chromedriver-service": "^5.0.2",
"wdio-cucumber-selected-steps": "file:../wdio-cucumber-selected-steps",
```

**Config of WebdriverIO**

```
    services: ['selenium-standalone'],
    framework: 'cucumber',
    sync: true,
```

**Describe the bug**
After using `getAlertText()` in a no-alert situation, `$(...).click`, etc. don't work anymore, since `$(...)` returns a promise.

**To Reproduce**

- open a page with no alertbox
- call `browser.getAlertText()`
- click on an element that exists. => boom, `$(...)click` is not a function.

https://github.com/Lusito/wdio-issues

**Expected behavior**
Even when the alert doesn't exist, it should not break the following `$(...)`.
Obviously, when I use `isAlertOpen()`, I can prevent the "wrong" call to `getAlertText()`, but I wouldn't expect this kind of behavior in the first place.

christian-bromann:
Can you provide your test code?

Lusito:
Here you go:
https://bitbucket.jambit.com/users/spfingsten/repos/wdio-issues/browse

The setup has been hacked together since I can't release my original code here right now.

christian-bromann:
@Lusito can you just share a gist? I can't access that page.

Lusito:
ah, damnit, the page is only accessible from inside the company lan.. gimme a second

mgrybyk:
I'll fix it today

christian-bromann:
[...]

PR #4221, opened 2019-07-18
Fix slow before/after command hook in sync mode

mgrybyk:
## Proposed changes

Fix slow before/after command hook in sync mode. #4133 
Avoid calling before/after hook for all nested commands.

**93% faster now!**

See test used https://github.com/webdriverio/webdriverio/pull/4221#issuecomment-513987108

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/webdriverio/webdriverio/blob/master/CONTRIBUTING.md) doc
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

## Further comments

To make this fix working properly we need to merge https://github.com/webdriverio/webdriverio/pull/4229 first and rebase

PR #4249, opened 2019-07-24
Fix commands error handling in sync mode

mgrybyk:
## Proposed changes

After https://github.com/webdriverio/webdriverio/pull/4221 error handling of commands was broken.

Should fix:
- promise was not resolved after error in previous command;  #4244 
- stacktrace is not full;

## Types of changes

## Checklist

### DIFF
--- a/packages/wdio-sync/src/wrapCommand.js
+++ b/packages/wdio-sync/src/wrapCommand.js
@@ -12,13 +12,10 @@ import { sanitizeErrorMessage } from './utils'
  * @return {Function}   actual wrapped function
  */
 export default function wrapCommand (commandName, fn) {
-    let stackError
     return function wrapCommandFn (...args) {
         // save error for getting full stack in case of failure
         // should be before any async calls
-        if (!stackError) {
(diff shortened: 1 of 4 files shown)


packages/wdio-sync/src/wrapCommand.js
```diff
@@ -12,13 +12,19 @@ import { sanitizeErrorMessage } from './utils'
  * @return {Function}   actual wrapped function
  */
 export default function wrapCommand (commandName, fn) {
+    let stackError
     return function wrapCommandFn (...args) {
+        // save error for getting full stack in case of failure
+        // should be before any async calls
+        if (!stackError) {
+            stackError = new Error()
```
christian-bromann: wouldn’t this mean that the error would be always the same?

mgrybyk: the error - no, but the stack trace yes in some cases, need to fix it

mgrybyk:
If there are no comments, let's merge!

Issue #4254, opened 2019-07-25
findElement sometimes returns Promise in sync mode


diachedelic:
**Environment (please complete the following information):**
 - **WebdriverIO version: v5.11.7
 - **Mode: WDIO Testrunner
 - **If WDIO Testrunner, running sync/async: sync
 - **Node.js version: v11.15.0
 - **NPM version: v6.7.9
 - **Browser name and version: 76.0.3809.80 (Official Build) beta (64-bit)
 - **Platform name and version: MacOS 10.14.5 (18F132)
 - **Additional wdio packages used (if applicable):** 
 ```

    "@wdio/allure-reporter": "^5.11.7",
    "@wdio/cli": "^5.11.8",
    "@wdio/local-runner": "^5.11.7",
    "@wdio/mocha-framework": "^5.11.0",
    "@wdio/selenium-standalone-service": "^5.11.2",
    "@wdio/spec-reporter": "^5.11.7",
    "@wdio/sync": "^5.11.9",
```

**Config of WebdriverIO**
See repro

**Describe the bug**
Given the correct circumstances, calling `element.$()` returns a Promise in
sync mode.

**To Reproduce**

```
git clone git@github.com:diachedelic/wdio-repros.git &&
  cd wdio-repros &&
  git checkout sync-promises &&
  npm i &&
  npm test
```

**Expected behavior**
I expect `$()` to always return an Element

**Log**
https://gist.github.com/diachedelic/9af6ccc9a4beedb8918f2ab5616fe004

**Comment by mgrybyk** [comment]
Please update to the latest version.

If you still experience the issue, please provide with steps to reproduce

**Comment by diachedelic** [comment]
@mgrybyk which version is that?

**Comment by mgrybyk** [comment]
Hey, I think I missed something. I though you are still on previous one.

I'll take a look soon.
Thank you for steps to reproduce

**Comment by diachedelic** [comment]
@mgrybyk I am on the latest that NPM will give me - I'm pretty sure the latest `webdriverio` and `webdriver` are `v5.11.7`, even though some other packages are at `v5.11.9`

**Comment by diachedelic** [comment]
@mgrybyk If I pin `webdriver` to `v5.11.5` (via Yarn's "resolutions" feature) the issue is resolved, e.g.:
```
// package.json
...
  "devDependencies": {
    ...
    "webdriverio": "^5.11.7"
  },
  "resolutions": {
    "webdriver": "5.11.5"
  },
...

[...]

PR #4256, opened 2019-07-25
Fix element chaining in async mode

mgrybyk:
## Proposed changes

Fix issue when chaining in some cases return promise.
#4254 

## Types of changes

## Checklist

### DIFF
--- a/packages/wdio-webdriver-mock-service/src/index.js
+++ b/packages/wdio-webdriver-mock-service/src/index.js
@@ -45,6 +45,7 @@ export default class WebdriverMockService {
         global.browser.addCommand('waitForDisplayedScenario', ::this.waitForDisplayedScenario)
         global.browser.addCommand('cucumberScenario', ::this.cucumberScenario)
         global.browser.addCommand('clickScenario', ::this.clickScenario)
+        global.browser.addCommand('isExistingScenario', ::this.isExistingScenario)
     }
 
     clickScenario() {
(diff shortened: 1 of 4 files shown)


mgrybyk:
added also smoke test for it

packages/wdio-webdriver-mock-service/src/index.js
```diff
@@ -55,6 +56,15 @@ export default class WebdriverMockService {
         this.command.elementClick(ELEMENT_ID).once().reply(200, { value: null })
     }
 
+    isExistingScenario() {
+        this.nockReset()
+        const elemResponse = { 'element-6066-11e4-a52e-4f735466cecf': ELEMENT_ID }
+
+        this.command.findElement().times(1).reply(200, { value: elemResponse })
+        this.command.findElementFromElement(ELEMENT_ID).times(2).reply(200, { value: elemResponse })
+        this.command.findElementsFromElement(ELEMENT_ID).times(2).reply(200, { value: [elemResponse] })
```
christian-bromann: this is not needed
```suggestion
```

tests/mocha/test.js
```diff
@@ -21,6 +21,14 @@ describe('Mocha smoke test', () => {
         $('elem').click()
     })
 
+    it('should chain properly', () => {
+        browser.isExistingScenario()
+
+        const el = browser.$('body')
+        el.$('.selector-1').isExisting()
+        el.$('.selector-2').isExisting()
```
christian-bromann: can we also assert the value to ensure that the promises got resolved?

mgrybyk: before the fix `el.$('.selector-2')` was Promise, so calling isExisting() was not working.
Anyway it makes sense to add additional verification

packages/wdio-webdriver-mock-service/src/index.js
mgrybyk: can remove this one because calling isExisting() calls findElementsFromElement underneath

christian-bromann: right

PR #4269, opened 2019-07-30
Fix waitUntil first iteration is promise

mgrybyk:
## Proposed changes

First iteration of `browser.waitUntil` doesn't respect promises.
```
const results = []
browser.waitUntil(() => {
    flag = !flag
    results.push(browser.pause(1))
    return flag
})
// results: [ Promise, undefined ]
```

## Types of changes

## Checklist

### DIFF
--- a/tests/mocha/test.js
+++ b/tests/mocha/test.js
@@ -32,15 +32,13 @@ describe('Mocha smoke test', () => {
         assert.equal(el.$('.selector-2').isExisting(), true)
     })
 
-    it('should waitUntil', () => {
-        let flag = true
+    it('should handle promises in waitUntil callback funciton', () => {
         const results = []
         browser.waitUntil(() => {
-            flag = !flag
-            results.push(browser.pause(1))
-            return flag
+            results.push(browser.getUrl())
+            return results.length > 1
         })
-        assert.deepEqual(results, [undefined, undefined])
+        assert.deepEqual(results, ['https://mymockpage.com', 'https://mymockpage.com'])
     })
 
     describe('middleware', () => {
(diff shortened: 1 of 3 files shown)


tests/mocha/test.js
```diff
@@ -32,6 +32,17 @@ describe('Mocha smoke test', () => {
         assert.equal(el.$('.selector-2').isExisting(), true)
     })
 
+    it('should waitUntil', () => {
```
christian-bromann: This smoke test is not clear to me. Can we create a more real world scenario where the first 3 request fail (e.g. for isExisting) and then the 4th works and we check if the condition was executed 4 times and the return value of waitUntil is true?

mgrybyk: Sorry to say have no better ideas to verify if wait until command promises are resolved. 
However I can rename the test to reflect what it actually does, is this fine?

mgrybyk: argh, no, I can't use isExisting, because it is element call, and I need browser call, anyway will tune test a bit

mgrybyk: @christian-bromann take a look

mgrybyk: I can't say it is more meaningful but at least test name reflects what it does

tests/mocha/test.js
```diff
@@ -32,6 +32,15 @@ describe('Mocha smoke test', () => {
         assert.equal(el.$('.selector-2').isExisting(), true)
     })
 
+    it('should handle promises in waitUntil callback funciton', () => {
+        const results = []
+        browser.waitUntil(() => {
```
christian-bromann: This looks good to me, one more note: let's do 
```suggestion
        const result = browser.waitUntil(() => {
```
and make an assertion on the result: `expect(result).toBe(true)`

mgrybyk: nice catch, missed that

mgrybyk: @christian-bromann done.

PR #4276, opened 2019-07-30
Use index when refetching element

### LINKED ISSUE #4270: elements fetched with $$ lose their context
 - **WebdriverIO version:** 5.11.10
 - **Mode:** WDIO Testrunner
 - **Node.js version:** v11.15.0
 - **NPM version:** yarn 1.15.2
 - **Browser name and version:** 75.0.3770.142 (Official Build) (64-bit)
 - **Platform name and version:** MacOS 10.14.5
```
@wdio/cli
@wdio/local-runner
@wdio/mocha-framework
@wdio/spec-reporter
@wdio/sync
wdio-chromedriver-service
```

**Config of WebdriverIO**
See repro

**Describe the bug**
After calling commands on non-existant children of the nth Element returned from
`$$`, the Element "loses it's place" in the array, and reverts to the first
item.

**To Reproduce**
```
git clone git@github.com:diachedelic/wdio-repros.git &&
  cd wdio-repros &&
  git checkout elements-context &&
  yarn &&
  yarn test
```

**Expected behavior**
Elements fetched with `$$` retain their ordering context no matter what.

**Log**
https://gist.github.com/diachedelic/303a3eddfdee4aa9dce9d2f491061563

** Additional context **
I was only able to reproduce using `wdio-chromedriver-service`, rather than Selenium, so it may be a bug in that package but I don't know how to tell.

mgrybyk:
## Proposed changes

Fix #4270 
refetch `$$('foo')[2].$('bar')` didn't work because we used to look for the first element.
This PR fixes it.

Cases when elements amount or ordering has changed are not handled!

## Types of changes

## Checklist

### DIFF
--- a/packages/webdriverio/tests/utils/refetchElement.test.js
+++ b/packages/webdriverio/tests/utils/refetchElement.test.js
@@ -43,6 +43,14 @@ describe('refetchElement', () => {
         expect(JSON.stringify(refetchedElement)).toBe(JSON.stringify(subSubElem))
     })
 
+    it('should not lose element index', async () => {
+        const elem = await browser.$('#foo')
+        const subElems = await elem.$$('#subfoo')
+        const subElem = subElems[1]
+        const refetchedElement = await refetchElement(subElem, '$')
+        expect(refetchedElement.elementId).toEqual(subElem.elementId)
+    })
+
     it('should successfully refetch an element that isn\'t immediately present', async () => {
         const elem = await browser.$('#foo')
         request.retryCnt = 0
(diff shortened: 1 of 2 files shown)


packages/webdriverio/tests/utils/refetchElement.test.js
```diff
@@ -43,6 +43,14 @@ describe('refetchElement', () => {
         expect(JSON.stringify(refetchedElement)).toBe(JSON.stringify(subSubElem))
     })
 
+    it('should not lose element index', async () => {
```
mgrybyk: the test was failing before the fix, of course.

PR #4281, opened 2019-07-30
Speedup internal tests

mgrybyk:
## Proposed changes

Reduce time taken by tests execution. Results first, ofcourse!
```
before: 1m 20 sec
after:  0m 33 sec
```

- run jest tests in parallel by removing [detectopenhandles](https://jestjs.io/docs/en/cli#detectopenhandles), 40s -> 14s
- speedup eslint _locally_ by using cache, 9s -> 1s
- speedup types test _locally_ by using --incremental, 7s -> 4s
- run smoke tests in parallel, 22s -> 10s

Also removed comments from generated files

## Types of changes

- [x] internal

## Checklist


tests/smoke.runner.js
```diff
@@ -5,29 +5,24 @@ import assert from 'assert'
 import launch from './helpers/launch'
 import { SERVICE_LOGS, LAUNCHER_LOGS, REPORTER_LOGS } from './helpers/fixtures'
 
-const sleep = (ms) => new Promise(resolve => setTimeout(resolve, ms));
+const sleep = (ms) => new Promise(resolve => setTimeout(resolve, ms))
 
-(async () => {
-    /**
-     * Mocha wdio testrunner tests
-     */
```
christian-bromann: I don't think these comments add a lot of time to the execution, I rather would keep them

tests/smoke.runner.js
```diff
@@ -156,6 +147,22 @@ const sleep = (ms) => new Promise(resolve => setTimeout(resolve, ms));
...
+        cucumberTestrunner(),
+        cucumberFailAmbiguousDefinitions(),
+        customService(),
+        customReporterString(),
+        customReporterObject(),
+        multiremote(),
+        retryFail(),
+        retryPass(),
+    ]
+    await Promise.all(tests)
```
christian-bromann: I rather prefer to run these tests sequentially so in case one test is failing it doesn't make the stdout too noisy

package.json
```diff
@@ -27,17 +27,17 @@
...
+    "test:coverage": "jest --coverage",
     "test:smoke": "babel-node ./tests/smoke.runner.js",
     "ts": "run-p ts:*",
-    "ts:sync": "cd tests/typings/sync && tsc",
-    "ts:sync-mocha": "cd tests/typings/sync-mocha && tsc",
-    "ts:sync-jasmine": "cd tests/typings/sync-jasmine && tsc",
-    "ts:webdriverio": "cd tests/typings/webdriverio && tsc",
-    "ts:webdriverio-mocha": "cd tests/typings/webdriverio-mocha && tsc",
-    "ts:webdriverio-jasmine": "cd tests/typings/webdriverio-jasmine && tsc",
+    "ts:sync": "cd tests/typings/sync && tsc --incremental",
```
christian-bromann: What does `--incremental` doing?

mgrybyk: > Enable incremental compilation by reading/writing information from prior compilations to a file on disk.

> TypeScript 3.4 introduces a new flag called --incremental which tells TypeScript to save information about the project graph from the last compilation. The next time TypeScript is invoked with --incremental, it will use that information to detect the least costly way to type-check and emit changes to your project

mgrybyk: it generates `tsconfig.tsbuildinfo` file in /tests/types/**/dist folder

tests/smoke.runner.js
mgrybyk: in case some of these fail we can easily comment out others and re-run, it runs more than two times faster!

tests/smoke.runner.js
mgrybyk: yeap, will revert this

tests/smoke.runner.js
mgrybyk: are there any other options that can work?

tests/smoke.runner.js
mgrybyk: @christian-bromann done.

tests/smoke.runner.js
christian-bromann: > in case some of these fail we can easily comment out others and re-run

Yeah but this doesn't work in CI

> are there any other options that can work?

Maybe allow to run a specific smoke test by applying a cli parameter?

mgrybyk: > Yeah but this doesn't work in CI

what if we run tests sequentially in CI? Does Travis CI has environment variable so that I can understand if we running test in CI or not?

mgrybyk: Added condition `if (process.env.CI)` to run sequentially.

Need to make sure that we have that env variable

mgrybyk: yeah, it works!

mgrybyk:
@christian-bromann one second, I wanted to run jest in parallel on travis CI as well

mgrybyk:
@christian-bromann that's it for now, no more changes planned.
Thank you for reviewing it

PR #4299, opened 2019-08-03
Fix chrome getLogs and getLogTypes

### LINKED ISSUE #4297: How to get browser logs in Chromedriver version 76 ?
How do we get browser console logs in chromedriver v76. Since getLogs does not exist anymore as mentioned over [here ](https://github.com/webdriverio/webdriverio/issues/4275). Is there workaround to get browser console logs?

Thanks in advance!!

mgrybyk:
## Proposed changes

Since chromedriver 75 get log commands were deprecated in w3c mode.
Since chromedriver 76 get log commands were added again, see https://bugs.chromium.org/p/chromedriver/issues/detail?id=2947

Fix chrome `getLogs` and `getLogTypes` by adding these commands to chromium protocol.

fixes #4297

## Types of changes

## Checklist

### DIFF
--- a/packages/webdriver/protocol/chromium.json
+++ b/packages/webdriver/protocol/chromium.json
@@ -424,7 +424,7 @@
       "returns": {
         "type": "String[]",
         "name": "logTypes",
-        "description": "The list of available log types."
+        "description": "The list of available log types, example: browser, driver."
       }
     }
   },
--- a/packages/webdriver/protocol/chromium.json
+++ b/packages/webdriver/protocol/chromium.json
@@ -414,5 +414,36 @@
         "description": "The base64-encoded PNG image data comprising the screenshot of the visible region of an element’s bounding rectangle after it has been scrolled into view."
       }
     }
+  },
+  "/session/:sessionId/se/log/types": {
+    "GET": {
+      "command": "getLogTypes",
+      "description": "Get available log types.",
+      "ref": "https://github.com/SeleniumHQ/selenium/wiki/JsonWireProtocol#sessionsessionidlogtypes",
+      "parameters": [],
+      "returns": {
+        "type": "String[]",
+        "name": "logTypes",
+        "description": "The list of available log types."
+      }
+    }
+  },
+  "/session/:sessionId/se/log": {
+    "POST": {
+      "command": "getLogs",
+      "description": "Get the log for a given log type. Log buffer is reset after each request.",
+      "ref": "https://github.com/SeleniumHQ/selenium/wiki/JsonWireProtocol#sessionsessionidlog",
+      "parameters": [{
+        "name": "type",
+        "type": "string",
+        "description": "the log type",
+        "required": true
+      }],
+      "returns": {
+        "type": "Object[]",
+        "name": "logs",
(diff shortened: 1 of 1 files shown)


packages/webdriver/protocol/chromium.json
```diff
@@ -414,5 +414,36 @@
         "description": "The base64-encoded PNG image data comprising the screenshot of the visible region of an element’s bounding rectangle after it has been scrolled into view."
       }
     }
+  },
+  "/session/:sessionId/se/log/types": {
```
christian-bromann:
```suggestion
  "/session/:sessionId/log/types": {
```

packages/webdriver/protocol/chromium.json
```diff
@@ -414,5 +414,36 @@
...
+      "ref": "https://github.com/SeleniumHQ/selenium/wiki/JsonWireProtocol#sessionsessionidlogtypes",
+      "parameters": [],
+      "returns": {
+        "type": "String[]",
+        "name": "logTypes",
+        "description": "The list of available log types."
+      }
+    }
+  },
+  "/session/:sessionId/se/log": {
```
christian-bromann:
```suggestion
  "/session/:sessionId/log": {
```

packages/webdriver/protocol/chromium.json
```diff
@@ -414,5 +414,36 @@
...
+  "/session/:sessionId/se/log/types": {
+    "GET": {
+      "command": "getLogTypes",
+      "description": "Get available log types.",
+      "ref": "https://github.com/SeleniumHQ/selenium/wiki/JsonWireProtocol#sessionsessionidlogtypes",
+      "parameters": [],
+      "returns": {
+        "type": "String[]",
+        "name": "logTypes",
+        "description": "The list of available log types."
```
christian-bromann: Let's add the few log types as example in here

packages/webdriver/protocol/chromium.json
christian-bromann: > it should be with se in url according to issue mentioned in PR, it works with se, I have verified it

sorry I missed that

mgrybyk: @christian-bromann one more note here, getLogs will stop working for chrome versions (not chromedriver!) below 76.

The only way to workaround it is to check if chrome version is below 76 in runtime using `browser.capabilities.version` in jsonwp mode and `browser.capabilities.browserVersion` in w3c mode.

I don't think we really need it as far as it is better to use latest chrome, but I might be wrong.

What do you think?

packages/webdriver/protocol/chromium.json
mgrybyk: done

packages/webdriver/protocol/chromium.json
christian-bromann: I think it is worth implementing such feature and decide on a pattern for the protocols (e.g. chromium.json is base and chromium_64.json overwrites everything from version 64 upwards). Given that we make a session request and then create the instance prototype the effort should be relatively small. That said, it should not be part of this PR. I will create an issue with this we some more details in a sec.

Issue #4298, opened 2019-08-03
Typings for webdriver commands are not correct if `elemenId` in `variables`


jayrepo:
**Environment (please complete the following information):**
 - **WebdriverIO version:** 5.11.7
 - **Mode:**  WDIO Testrunner
 - **If WDIO Testrunner, running sync/async:** sync
 - **Node.js version:** 12.6.0
 - **NPM version:** 6.9.0
 - **Browser name and version:** Chrome 68
 - **Platform name and version:** Windows 10
 - **Additional wdio packages used (if applicable):** @wdio/sync

**Config of WebdriverIO**
 Using TypeScript.

**Describe the bug**
For commands require `elementId`, the corresponding typings do not have `elementId` as first arg

**To Reproduce**
e.g.
```js
// command
browser.getElementRect(elementId)

// typing in webdriver.d.ts
getElementRect(): object;
```

**Expected behavior**
 if a command has `elementId` in `variables` array, `elementId` should be prepend as first arg. 

**Additional context**
I just checked a few commands.

If all commands in `webdriverio/packages/webdriver/protocol/*.json` are binding to browser/driver, remove filtering of `elementId` in `webdriverio/scripts/type-generation/webdriver-generate-typings.js` should fix this issue

> Please use the correct markdown

mgrybyk:
@jayrepo please avoid using protocol commands, use `$('elem').getSize()`

Question, does `browser.getElementRect($('elem'))` work at all? In all browsers?

jayrepo:
@mgrybyk 
Yes I can use `getSize()` and `getLocation()` for `getElementRect()`.
Does all protocol commands have variants on Element? I can't find that for `browser.takeElementScreenshot()`

`browser.getElementRect($('elem').elementId)` works on native Android app. not sure about browsers.

I'm experimenting on typescript with webdriverio. Another problem with the auto generated typing is browser commands return `object`.
```
// Property 'height' does not exist on type '{}'.ts(2339)
const { height, width } = browser.getWindowRect()

// Have to workaround with
const { height, width }: any = browser.getWindowRect()
```

mgrybyk:
@jayrepo all W3C compliant commands have variants on Element, if not please let us know.
To take element screenshot you can just `$('').saveScreenshot('')`

Regarding issue with rect please raise another defect, you may describe it really shortly, just put webdrverio version for history

mgrybyk:
[...]

PR #4300, opened 2019-08-03
TypeScript: add elementId as parameter to protocol commands with variable

mgrybyk:
## Proposed changes

Add elementId as parameter to protocol commands with variable

fixes: #4298

## Types of changes

## Checklist

### DIFF
--- a/packages/webdriverio/webdriverio.d.ts
+++ b/packages/webdriverio/webdriverio.d.ts
@@ -4,7 +4,15 @@ type $ = (selector: string | Function) => Promise<WebdriverIOAsync.Element>;
 type $$ = (selector: string | Function) => Promise<WebdriverIOAsync.Element[]>;
 
 // Element commands that should be wrapper with Promise
-type ElementPromise = Omit<WebdriverIO.Element, 'addCommand' | '$' | '$$' | 'selector' | 'elementId'>;
+type ElementPromise = Omit<WebdriverIO.Element,
+    'addCommand'
+    | '$'
+    | '$$'
+    | 'selector'
+    | 'elementId'
+    | 'element-6066-11e4-a52e-4f735466cecf'
+    | 'ELEMENT'
+>;
 
 // Methods which return async element(s) so non-async equivalents cannot just be promise-wrapped
 interface AsyncSelectors {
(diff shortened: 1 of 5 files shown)


packages/webdriverio/webdriverio.d.ts
```diff
@@ -19,7 +27,13 @@ type ElementAsync = {
 } & AsyncSelectors;
 
 // Element commands that should not be wrapper with promise
-type ElementStatic = Pick<WebdriverIO.Element, 'addCommand' | 'selector' | 'elementId'>
+type ElementStatic = Pick<WebdriverIO.Element,
+    'addCommand'
+    | 'selector'
+    | 'elementId'
+    | 'element-6066-11e4-a52e-4f735466cecf'
```
CrispusDH: could you explain how it would works with it `element-6066-11e4-a52e-4f735466cecf`?

mgrybyk: @CrispusDH see https://www.w3.org/TR/webdriver1/#elements

`element-6066-11e4-a52e-4f735466cecf` is constant, same as `ELEMENT` in JSONWP

scripts/type-generation/webdriver-generate-typings.js
```diff
@@ -15,8 +15,13 @@ for (const [protocolName, definition] of Object.entries(PROTOCOLS)) {
...
             const vars = variables
-                .filter((v) => v.name != 'sessionId' && v.name != 'elementId')
-                .map((v) => `${v.name}: string`)
+                .filter((v) => v.name != 'sessionId')
+                .map((v) => {
+                    if (v.name === 'elementId') {
+                        return `${v.name}: ElementId`
+                    }
+                    return `${v.name}: string`
+                })
```
CrispusDH: could you add comments why you filter bu `sessionId` and have if block with condition `elementId`. I think it would be helpful for folks who, possibly, would improve it or change.

mgrybyk: hey @CrispusDH I'll add comment, however I've noticed that I made a mistake here.

Let me fix it :)

PR #4305, opened 2019-08-04
Run travis CI with Node.js 10

tadashi0713:
## Proposed changes
Since many developers use Node.js 10 for WebdriverIO, and Appium minimum Node.js version has bumped to v10, it`s better to run travisCI with Node.js 10
https://github.com/appium/appium/releases/tag/v1.14.0

## Types of changes

## Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ language: node_js
 
 node_js:
   - 8
+  - 10
+  - 12
 
 script:
   - npm run ci
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ language: node_js
 node_js:
   - 8
   - 10
-  - 12
 
 script:
   - npm run ci


tadashi0713:
@mgrybyk Thanks for comment. 

> At this point there is no need to drop nodejs 8, I think.

I didn't drop nodejs 8, just added running on nodejs10

https://travis-ci.org/webdriverio/webdriverio/builds/567468284

As you can see, tests are running both nodejs 8 and nodejs10

I am sorry if I misunderstood

mgrybyk:
Yeah, missed that, does it doubles execution time? If yes I don't think we need it because pipeline will run for 11 minutes

christian-bromann:
@tadashi0713 any reason why you would like to run tests in v10 as well? I agree with @mgrybyk to not add it given the increase in execution time.

tadashi0713:
@mgrybyk @christian-bromann I believe they should be run parallel(I'm now rerunning)
I believe it should be run v10 as well, because as I wrote in PR.

> Since many developers use Node.js 10 for WebdriverIO, and Appium minimum Node.js version has bumped to v10, it's better to run travisCI with Node.js 10

christian-bromann:
> Since many developers use Node.js 10 for WebdriverIO, and Appium minimum Node.js version has bumped to v10, it`s better to run travisCI with Node.js 10

Can you please elaborate why we should test on Node v10 too? We don't have any version specific logic in WebdriverIO. In addition to that do we not test Appium in our pipeline so any changes in Appium would not be caught here.

tadashi0713:
hmm so there's no need to check in CI for multiple(including future) language versions?
I believe it's better to run CI to do in order to guarantee its package in each language versions, and many OSS project do so.

tadashi0713:
If so, I will close this PR

mgrybyk:
@tadashi0713 it's definitely make sense, however we don't want to increase pipeline duration if possible. Also, as you have mentioned, most of people use nodejs 10, so developers run test locally.

patthiel:
> Can you please elaborate why we should test on Node v10 too? We don't have any version specific logic in WebdriverIO. In addition to that do we not test Appium in our pipeline so any changes in Appium would not be caught here.

This isn't the case, unfortunately.

@christian-bromann we're unable to use the `wdio-sync` package with node 8 due to the Fibers v4 package requiring node 10.  (see https://www.npmjs.com/package/fibers/v/4.0.0#supported-platforms )

mgrybyk:
Well, that's another topic, if you are right we have to change minimum supported version to 10

patthiel:
Yep, another topic, but related to this issue. I'm happy to open up a separate one to raise that. I'm  surprised it wasn't caught by the CI build, but I also don't know the intricacies of how this project and it's repos are built since it started leveraging lerna.

mgrybyk:
CI passes with nodejs `v8.16.0`

patthiel:
@mgrybyk digging into this a bit further. I tried running the CI build from a fresh node8 container, it seems to be reliant on globally installed dependencies? 

Reproduceable by the following:
1. Run a fresh node8 container and bash into that jawn `docker run -it node:8.16.0-jessie /bin/bash` 
2. clone webdriverio
3. run `npm run ci`
4. Outputs:
```
sh: 1: lerna: not found
npm ERR! file sh
npm ERR! code ELIFECYCLE
npm ERR! errno ENOENT
npm ERR! syscall spawn
npm ERR! @ ci: `lerna bootstrap --no-ci && run-s build-full test:eslint test:typings test:coverage link test:smoke`
npm ERR! spawn ENOENT
```

Am i missing something ? I would think all these commands in package json would be called via `npx`.

Back to the original issue: I can submit a PR downgrading Fibers to version 3. That should get us back node 8 compatibility.

christian-bromann:
@patthiel lerna is a [project dependency](https://github.com/webdriverio/webdriverio/blob/master/package.json#L73) and is installed when you do `npm install` after the clone.

> I can submit a PR downgrading Fibers to version 3

I am not sure if this a good idea. I agree however that we need to better communicate when an update of Fibers requires a new minimal NodeJS version.

patthiel:
> @patthiel lerna is a project dependency and is installed when you do npm install after the clone.

yup, totally understand that. I, however, didn't realize that `npm run` scripts automatically add `./node_modules/.bin` to the PATH when being executed (i usually use `npx` to run those bins). Been a while since i've worked on a node project 😉.

Anywho, instead of downgrading fibers, it sounds like we do need to update the minimum node version requirement. I am still concerned that the CI build missed this, but here's a PR updating the node version: https://github.com/webdriverio/webdriverio/pull/4356

mgrybyk:
closed due to https://github.com/webdriverio/webdriverio/pull/4356

Issue #4008, opened 2019-05-22
Using @wdio/jasmine-framework does not support spec retry on failure


SanthoshBonala:
**Pre-check**
- [x] I'm aware that I can [edit the docs](https://github.com/webdriverio/webdriverio/blob/master/CONTRIBUTING.md) and submit a pull request

**Describe the improvement**

I'd like to report
- [x] Unclear documentation
- [ ] A typo
- [x] Missing documentation
- [ ] Other

**Description of the improvement/report**
When I use @wdio/jasmine-framework as a plugin for jasmine. The docs say that I can pass test retry count as the last parameter for the spec or hook. But when I try to do that it picking it as a timeout parameter instead of test retry which is the behaviour of jasmine framework.

** package.json
```
  "dependencies": {
    "@wdio/cli": "^5.7.15",
    "@wdio/jasmine-framework": "^5.7.13",
    "@wdio/local-runner": "^5.7.15",
    "webdriverio": "^5.7.13"
  }
```
** test.js
```
describe('Webdriverio', function () {
    it('Testing async function',  async () => {
        // redirect to url so that we can exceed default 1ms timeout
        let title;
        browser.url('https://webdriver.io/docs/retry.html');
        title  = await browser.getUrl();
        browser.url('https://webdriver.io/docs/retry.html');
        title  = await browser.getUrl();
        browser.url('https://webdriver.io/docs/retry.html');
        title  = await browser.getUrl();
        browser.url('https://webdriver.io/docs/retry.html');
        title  = await browser.getUrl();
        browser.url('https://webdriver.io/docs/retry.html');
        title  = await browser.getUrl();
        browser.url('https://webdriver.io/docs/retry.html');
        title  = await browser.getUrl();
        browser.url('https://webdriver.io/docs/retry.html');
        title  = await browser.getUrl();
        browser.url('https://webdriver.io/docs/retry.html');
        title  = await browser.getUrl();
        expect(title).toBeTruthy();
    }, 1);
});
```
Error from @wdio/jasmine-framework adapter:
```
 "Error: Timeout - Async callback was not invoked within 1ms (custom timeout)
    at <Jasmine>
    at ontimeout (timers.js:436:11)
    at tryOnTimeout (timers.js:300:5)
    at listOnTimeout (timers.js:263:5)
    at Timer.processTimers (timers.js:223:10)"
```

mike-d-davydov:
[...]

PR #4308, opened 2019-08-05
Fix for async mode in wdio-sync

SanthoshBonala:
## Proposed changes

Fix for the [Issue ](https://github.com/webdriverio/webdriverio/issues/4008)

## Types of changes

## Checklist

### DIFF
--- a/packages/wdio-sync/src/index.js
+++ b/packages/wdio-sync/src/index.js
@@ -118,7 +118,7 @@ const runHook = function (hookFn, origFn, before, after, repeatTest = 0) {
             /**
              * user wants handle async command using promises, no need to wrap in fiber context
              */
-            if (hookFn.name === 'async') {
+            if (hookFn.constructor.name === 'AsyncFunction') {
                 return executeAsync.call(this, hookFn, repeatTest, filterSpecArgs(hookArgs))
             }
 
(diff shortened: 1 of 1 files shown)


packages/wdio-sync/src/index.js
```diff
@@ -118,7 +118,7 @@ const runHook = function (hookFn, origFn, before, after, repeatTest = 0) {
             /**
              * user wants handle async command using promises, no need to wrap in fiber context
              */
-            if (hookFn.name === 'async') {
+            if (hookFn.constructor.name === 'AsyncFunction') {
                 return executeAsync.call(this, hookFn, repeatTest, filterSpecArgs(hookArgs))
```
mgrybyk: Is it possible that hookFn can be undefined or something without constructor property? 
In such case we'll get cannot read property name of undefined?
Do you think it makes sense to verify if constructor property exists in hookFn?

SanthoshBonala: sure updating the pr to check for constructor property on function prototype

mgrybyk: Do you think we still need to keep the old behavior? 
Is it possible at all to name function async? Isn't it a keyword? (Can't verify right now, just asking)

SanthoshBonala: it is possible to name fn as aync eventhough async is a keyword. Having an old workflow can be used as a fallback approach? Do you want me to remove old behavior?

mgrybyk: Not sure if anyone used it, we can keep it, I don't mind. Let's see if anyone else has some input of this.

I still think we can remove this one `typeof hookFn === 'function' `

SanthoshBonala: sure updating

mgrybyk: Thanks a lot!

SanthoshBonala: no problem

Issue #2859, opened 2018-08-24
Implement tag management for all frameworks


christian-bromann:
_From @christian-bromann on May 14, 2018 8:36_

Many people want to run tests having a specific set of tags. In v4 there was no tag detection and if you would use the framework functionality it would spawn up a browser and immediately close it.

Let's implement native support for this.

_Copied from original issue: webdriverio/v5#68_

christian-bromann:
_From @BorisOsipov on May 14, 2018 8:37_

am I right, it is related to issue https://github.com/webdriverio/webdriverio/issues/1247 ?

christian-bromann:
@BorisOsipov yes

christian-bromann:
I doubt this is possible by using framework native capabilities. However if we could do that it would be amazing and way faster than some sort of file content regex matching.

christian-bromann:
_From @id4automation on August 14, 2018 23:31_

Hiya, Any inputs on this is much appreciated.

christian-bromann:
@id4automation Please see webdriverio/webdriverio#1247

mgrybyk:
hey guys, I'm using simple wrapper function to manage tags like this:
```
it("my test", () => {
    // some test
}, { tags: [browser.isMobileApp] })

// or 

describe("my suite", () => {
    it("my test 1", () => { })
    it("my test 2", () => { })
}, { tags: [APP_REGION !== REGIONS.ASIA, !browser.isIOS] })
```

I'm passing optional array of booleans to `it` and `describe` blocks to control if suite/test should be executed. If one of condition is **false** test/suite is excluded from run, it is not marked as skipped.

It helps me a lot to organize tests in functional way and avoid grouping them by platforms, reports for different test configurations won't contain skipped reports of unsupported platforms.

baruchvlz:
@christian-bromann Are we still thinking about implementing this?

mgrybyk:
I'd say yes! And now it's finally possible thanks to #4531 

You wanna deal with it?

baruchvlz:
@mgrybyk I'll try, will let you know if I come up across any roadblocks.

mgrybyk:
Do you have any vision on how it should work in the end?

christian-bromann:
it would be great if we could leverage the tagging functionality of the frameworks and combine it with the use case of cross browser testing, e.g. allow to tag specific tests to be only run in browserX and skipped in other browser. This is just an idea though, first we should really define what we want to do here.

[...]

PR #4326, opened 2019-08-06
wdio-cucumber-framework:  Filter cucumber features before process spawned

### LINKED ISSUE #1247: Possibility to configure test spec process spawning?
## The problem

According to https://github.com/webdriverio/webdriverio/issues/1106#issuecomment-184728530 `each test spec gets executed in a single child process`.
We have currently 101 individual `.feature files` (we are using Cucumber) and cannot put them into one big file.
When we launch our test script, my Selenium Server is trying to spawn 101 process, this results in a very heavy lag and I need to kill the processes.

Is there a way to configure this behaviour? Do we need to adjust our wdio.conf?
## Environment

**updated on 07/05/16**
- WebdriverIO version: 4.1.1
- Node.js version: 4.4.6
- Selenium Server 2.53.0
- using wdio testrunner
- running synchronous or asnychronous tests: does not matter
- Additional wdio packages used (if applicable):
  wdio-cucumber-framework: 0.2.0
## Details

Even with the cucumber option tags like `-cucumberOpts.tags=@foo`, it will spawn 101 processes, even though only 1 feature was tagged with `@foo`
(Note: The setup is working fine with wdio 3.4.0)

jayrepo:
## Proposed changes

Fix #1247
Do an additional filtering  in `wdio-cli/src/launcher.js` before process spawned

Export a utility function to filter features by tag expression
Use with `specsFilter` option in config `specFilter: (config) => string[]`

```js
// wdio.conf.js
exports.config = {
  // ...
  specsFilter: require('@wdio/cucumber-framework').filterSpecsByTag
  // ...
  cucumberOpts: {
      // ...
    tagExpression: '@smoke and @stable'
      // ...
  }
  // ...
}
```
A simple example of filter by file name using `specsFilter`:
```js
const { ConfigParser } = require('@wdio/config')
exports.config = {
  // ...
  namePattern: 'moduleA',
  specsFilter: (config) => {
      const specs = ConfigParser.getFilePaths(config.specs, false)
      return specs.filter(spec => spec.includes(config.namePattern))
  }
  // ...
}
```
Other frameworks should exported their own filters, as each framework has its own implementation.

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/packages/wdio-cli/src/launcher.js
+++ b/packages/wdio-cli/src/launcher.js
@@ -60,6 +60,13 @@ class Launcher {
     async run () {
         let config = this.configParser.getConfig()
         let caps = this.configParser.getCapabilities()
+        // /**
+        //  * run filter on specs
+        //  */
+        if (Array.isArray(config.specs) && config.specs.length && typeof config.specsFilter === 'function') {
+            config.specs = await config.specsFilter(config)
+        }
+
         const launcher = initialiseServices(config, caps, 'launcher')
 
         /**
--- a/packages/wdio-cli/src/utils.js
+++ b/packages/wdio-cli/src/utils.js
@@ -158,6 +158,7 @@ export function replaceConfig(
 
         return config.replace(text, buildNewConfigArray(text, type, name))
     }
+    return config
 }
 
 export function addServiceDeps(names, packages, update) {
--- a/packages/wdio-cli/tests/launcher.test.js
+++ b/packages/wdio-cli/tests/launcher.test.js
@@ -305,12 +305,12 @@ describe('launcher', () => {
             expect(launcher.runSpecs()).toBe(false)
             expect(launcher.getNumberOfRunningInstances()).toBe(5)
             expect(launcher.getNumberOfSpecsLeft()).toBe(4)
-            expect(launcher.schedule[0].runningInstances).toBe(3)
-            expect(launcher.schedule[0].availableInstances).toBe(47)
+            expect(launcher.schedule[0].runningInstances).toBe(2)
+            expect(launcher.schedule[0].availableInstances).toBe(48)
             expect(launcher.schedule[1].runningInstances).toBe(2)
             expect(launcher.schedule[1].availableInstances).toBe(58)
-            expect(launcher.schedule[2].runningInstances).toBe(0)
-            expect(launcher.schedule[2].availableInstances).toBe(70)
+            expect(launcher.schedule[2].runningInstances).toBe(1)
+            expect(launcher.schedule[2].availableInstances).toBe(69)
         })
 
         it('should not allow to schedule more runner if no instances are available', () => {
--- a/packages/wdio-cli/tests/launcher.test.js
+++ b/packages/wdio-cli/tests/launcher.test.js
@@ -434,8 +434,5 @@ describe('launcher', () => {
 
             expect(await launcher.run()).toBe(1)
         })
-        it('filter specs', async () => {
-
-        })
     })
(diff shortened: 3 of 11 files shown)


mgrybyk:
Can you please help understanding what is the difference between your implementation and built in tagExpression filter?

jayrepo:
@mgrybyk It's almost the same as the built in filter.

The problem is that when built in tag filter starts to work, processes are already spawned for each feature file. Even if the feature does not contain a matching tag expression, the process will still start a seesion, run no test, close session and report pass

It's really annoying if you have hundreds of feature files. For mobile app test, it takes around 1 min start a session.

This proposal did a pre-processing to filter the features before everything starts.

mgrybyk:
I'd better have it working natively for all frameworks! We have the same issue with mocha

If it's not possible, can we tune somehow built in tag expression filter, to work like your implementation or add another option to cucumberOpts and reuse existing  code as much as possible? 

Even though current solution work for you  it doesn't look clean enough, from my point of view

jayrepo:
@mgrybyk 
> I'd better have it working natively for all frameworks! We have the same issue with mocha

My original thought was add a config like `specsFilter: (config) => string[]`, so other framework can also export a filter function to use it.

This requires changes to the `wdio-config` package. As I'm not familiar with other frameworks and configs, I didn't touch it.
Is it OK to apply the filter just before return? https://github.com/webdriverio/webdriverio/blob/91bc19ae25477184efbd89e72b8dd64417fbd433/packages/wdio-config/src/lib/ConfigParser.js#L201

mgrybyk:
Let me take a look

mgrybyk:
not sure how to handle it better right now

jayrepo:
Made some changes, need to do some e2e test

jayrepo:
Add a config.
Filter the specs in `launcher` of `wdio-cli`

mgrybyk:
There is one more approach (suggestion) to handle it more widely, it is not a part of current PR!

We may change a bit the way how browser object is created. 
Currently we init browser before framework.
It might be changed, we may first create browser object, then init framework (but not run!) and verify if there are tests to run, if so, create browser session, and then run framework.

christian-bromann:
I prefer @mgrybyk solution here. We can change the interface of the wdio framework packages so that hey have an `init` and `run` function. This is related to #2859.

jayrepo:
@mgrybyk @christian-bromann 

That will be great. what I proposed is just a temporary workaround.

If my understanding is correct, your solution will work like this:
```js
const browser = new Browser()  // not start session
const specs = framework.init()  // parsing specs with framework's native tag implementation
if (specs.length) {
  brower.startSession()
  framework.run()
}
```

mgrybyk:
Yes.
As I've mentioned it requires refactoring in framework core, so I don't think it gonna be soon. If no one can pick it up I'll start working on it somewhere in September if there are no critical tasks.
Additionally I'm waiting for devtools PR to be merged first to avoid tons of conflicts.

codiophile:
Interestingly enough, we have this feature in our framework, although to solve a different problem. We filter the feature files based on tags before sending the list of spec files to the launcher. I do like @mgrybyk's proposed solution; I just wanted to point out that until it arrives you can implement the filtering outside of webdriverio.

mgrybyk:
@jayrepo please see https://github.com/webdriverio/webdriverio/pull/4531

mgrybyk:
Closing the PR.
Let's proceed convo in https://github.com/webdriverio/webdriverio/pull/4531

Issue #4322, opened 2019-08-05
$(element).moveTo() moves cursor to [0, 0] on Safari


thephilwells:
**Environment (please complete the following information):**
 - **WebdriverIO version:** 5.11.7
 - **Mode:** Testrunner
 - **If WDIO Testrunner, running sync/async:** sync
 - **Node.js version:** 8.9.0
 - **NPM version:** 5.5.1
 - **Browser name and version:** Safari 12.1.2
 - **Platform name and version:** MacOS 10.13
 - **Additional wdio packages used (if applicable):** @wdio/sauce-service (reproduced in Sauce and on local)

**Config of WebdriverIO**
Sauce config (W3C spec compliant):

```js
if (config.get('SAUCE_ACCESS_KEY')) {
  services.push('sauce')
  user = config.get('SAUCE_USERNAME')
  key = config.get('SAUCE_ACCESS_KEY')
  capability['sauce:options'] = {
    build: `${config.get('PROJECT_NAME')}:${config.get('BUILD_NUMBER') || new Date()}`,
    platform: config.get('PLATFORM_OS'),
    version: config.get('PLATFORM_BROWSER_VERSION').toString(),
    screenResolution: config.get('PLATFORM_SCREEN_RESOLUTION'),
    parentTunnel: config.get('PLATFORM_PARENT_TUNNEL'),
    tunnelIdentifier: config.get('PLATFORM_TUNNEL_ID'),
    extendedDebugging: config.get('PLATFORM_EXTEND_DEBUG'),
    commandTimeout: config.get('PLATFORM_COMMAND_TIMEOUT'),
    idleTimeout: config.get('PLATFORM_IDLE_TIMEOUT'),
    timeZone: config.get('PLATFORM_TIME_ZONE'),
  }
}
```

from package.json:

```
    "@wdio/cli": "^5.11.13",
    "@wdio/local-runner": "^5.11.0",
    "@wdio/mocha-framework": "^5.11.0",
    "@wdio/reporter": "^5.11.0",
    "@wdio/sauce-service": "^5.11.0",
    "@wdio/selenium-standalone-service": "^5.11.0",
    "@wdio/spec-reporter": "^5.11.0",
    "@wdio/sync": "^5.11.0",
    "webdriverio": "^5.11.0"
```

**Describe the bug**
[...]

PR #4330, opened 2019-08-07
webdriverio: remove actions.release from moveTo command

thephilwells:
## Proposed changes

Removes the explicit `browser.releaseActions()` from `element.moveTo()`

In SafariDriver, the act of releasing the actions was causing the mouse pointer to hop back to `0,0` on the page, invalidating the mouse hover state the user would have wanted.

## Types of changes

## Checklist

## Further comments

This fixes https://github.com/webdriverio/webdriverio/issues/4322

### DIFF
--- a/packages/webdriverio/src/commands/element/moveTo.js
+++ b/packages/webdriverio/src/commands/element/moveTo.js
@@ -35,5 +35,5 @@ export default async function moveTo (xoffset, yoffset) {
         id: 'finger1',
         parameters: { pointerType: 'mouse' },
         actions: [{ type: 'pointerMove', duration: 0, x: newXoffset, y: newYoffset }]
-    }]).then(() => this.releaseActions())
+    }])
 }


mgrybyk:
@thephilwells one question, if it possible to move to element 1 and then move to element 2?

I want to make sure that by removing releaseActions we are not breaking anything

thephilwells:
I'll check. Let me see if I can put a test together for that scenario, too.

mgrybyk:
You won't be able, we have no e2e at the moment, if you can verify it locally, that would be enough. Thank you!

thephilwells:
Good thing you asked me to check. In fact, in Safari, it now creates a second pointer for the second action. Back to the drawing board!

thephilwells:
It works as expected if we pass in the args for all the actions at once:

```js
async function moveTo(coords) // something like [{x: 50, y: 50}, {x: 25, y: 25}]
  let actions = []
  coords.forEach(set => {
    actions.push({
      type: 'pointerMove',
      duration: 0,
      x: set.x,
      y: set.y
    })
  }

  return this.performActions([{
    type: 'pointer',
    id: 'finger1',
    parameters: { pointerType: 'mouse' },
    actions
  }])
}
```

But that still rules out cases where a user moves the mouse, wants to do something else, then wants to move the mouse again.

mgrybyk:
Ok, what if we make releasing actions configurable?
change function signature a bit so that it won't break anything to existing users:
```
export default async function moveTo (xoffset, yoffset, { releaseActions = true } = {}) {
```
and use it Safari like
`$('foo').moveTo(x, y, { releaseActions: false })`
then releaseActions manually, when needed `browser.releaseActions()`


Please note that it is just a suggestion! I'd better have some working solution without additional flags or magic, but

thephilwells:
That seems like a step in the right direction. I wonder if we need to check
if any actions are left open from prior calls to `performAction()`.

mgrybyk:
Good idea, can you try if that works for you?

thephilwells:
Even when I don't `releaseActions()` explicitly, the second action is broken in SafariDriver, with a second pointer ending up at 0,0. Going to try some more workarounds..

thephilwells:
I'm not really seeing a way to fix the "two actions in a row" case on Safari, short of the workaround I posted above. As I see it, this PR covers the cases like so:

- ✅ Single `moveTo()` call causes mouse pointer to go to `[xoffset, yoffset]` and stay there
- ❌ Multiple `moveTo()` calls cause multiple pointers to spawn on Safari, with all but the first pointer (which places itself correctly) appearing at `[0, 0]`

Maybe we can use this PR and open a new issue for the second case? That second issue might have to be fixed within SafariDriver.

christian-bromann:
> That second issue might have to be fixed within SafariDriver.

Can you try using Safari Developer Preview?

thephilwells:
@christian-bromann I have Tech Preview installed, but how do I get SafariDriver to use that on my local instead of plain old Safari in wdio?

thephilwells:
Sorry for the delay. Running against Safari Technology Preview is not straightforward! Will report back when I get this up and running.

christian-bromann:
@thephilwells any updates?

thephilwells:
I had trouble getting the beta working in my local and then the dayjob
stuff for away from me. I'll try again tomorrow.

What do we think about passing a flag like we talked about above?

packages/webdriverio/src/commands/element/moveTo.js
```diff
@@ -14,7 +14,7 @@
 
 import { getElementRect, getScrollPosition } from '../../utils'
 
-export default async function moveTo (xoffset, yoffset) {
+export default async function moveTo (xoffset, yoffset, { releaseActions = true } = {}) {
```
thephilwells: This optional flag would do no harm to existing code, but fixes the two-in-a-row `moveTo()` calls breakage on Safari.

Note that this is still not tested on Developer Preview.

christian-bromann:
Closing this PR as it seems to become stale. Feel free just create a new one if you are still interested in this patch. Cheers!

kartikeypokhriyal7000:
Is this issue fixed? I'm still facing issues with moveto @christian-bromann

christian-bromann:
> Is this issue fixed?

As you can see this PR was closed. Are your issues connected with this and are you using Safari?

kartikeypokhriyal7000:
I'm on safari and moveTo doesn't work @christian-bromann

christian-bromann:
@kartikeypokhriyal7000 feel free to pick up this PR and address outstanding comments.

PR #4332, opened 2019-08-07
End process normally

### LINKED ISSUE #4328: @wdio/local-runner can not kill the WebDriver sessions after completing test
 - **WebdriverIO version:** [^5.11.12]
 - **Mode:** [chromedriverr]
 - **Node.js version:** [10]
 - **NPM version:** [6]
 - **Browser name and version:** [Chrome 76]
 - **Platform name and version:** [mac OS 10.14.6]
 - **Additional wdio packages used (if applicable):** [@wdio/local-runner, @wdio/cucumber-framework, wdio-chromedriver-service]

**Config of WebdriverIO**
https://github.com/flyHe/UI_H5_Test/blob/master/tests/wdio.conf.js

**Describe the bug**
@wdio/local-runner can not kill the WebDriver sessions after completing test, the gulp is pending after completing all the tasks

**To Reproduce**
repo: https://github.com/flyHe/UI_H5_Test
`npm i`
`npm run test:search:local`

**Expected behavior**
no need to kill the task by [process.exit(0);](https://github.com/flyHe/UI_H5_Test/blob/master/gulp-tasks/test.js#L85)

**Log**
Screen Shot 2019-08-05 at 5 35 59 PM
kill manually
Screen Shot 2019-08-05 at 5 36 08 PM

mgrybyk:
## Proposed changes

End process normally instead of killing it with `process.exit`

fixed #4328

## Types of changes

## Checklist

### DIFF
--- a/packages/wdio-cli/src/config.js
+++ b/packages/wdio-cli/src/config.js
@@ -157,6 +157,7 @@ export const CLI_PARAMS = [{
     description: 'Cucumber options'
 }]
 
+/* istanbul ignore next */
 export const QUESTIONNAIRE = [{
     type: 'list',
     name: 'runner',
--- a/packages/wdio-cli/src/interface.js
+++ b/packages/wdio-cli/src/interface.js
@@ -22,6 +22,11 @@ export default class WDIOCLInterface extends EventEmitter {
         this.isWatchMode = isWatchMode
         this.inDebugMode = false
         this.specFileRetries = config.specFileRetries || 0
+
+        /**
+         * `Launcher.exitHandler` is always called so we need to identify
+         * if `SIGINT` was actually called
+         */
         this.endedNormally = false
 
         this.on('job:start', ::this.addJob)
--- a/packages/wdio-cli/src/interface.js
+++ b/packages/wdio-cli/src/interface.js
@@ -22,6 +22,7 @@ export default class WDIOCLInterface extends EventEmitter {
         this.isWatchMode = isWatchMode
         this.inDebugMode = false
         this.specFileRetries = config.specFileRetries || 0
+        this.endedNormally = false
 
         this.on('job:start', ::this.addJob)
         this.on('job:end', ::this.clearJob)
@@ -191,7 +192,7 @@ export default class WDIOCLInterface extends EventEmitter {
         /**
          * allow to exit repl mode via Ctrl+C
          */
-        if (this.inDebugMode) {
+        if (this.inDebugMode || this.endedNormally) {
             return false
         }
 
@@ -232,6 +233,7 @@ export default class WDIOCLInterface extends EventEmitter {
             return
         }
 
+        this.endedNormally = true
         this.printReporters()
         this.printSummary()
     }
--- a/packages/wdio-cli/src/launcher.js
+++ b/packages/wdio-cli/src/launcher.js
@@ -62,13 +62,13 @@ class Launcher {
         /**
          * catches ctrl+c event
          */
-        exitHook(:: this.exitHandler)
+        exitHook(::this.exitHandler)
         let exitCode
         let onCompleteResults
 
         try {
-            let config = this.configParser.getConfig()
-            let caps = this.configParser.getCapabilities()
+            const config = this.configParser.getConfig()
+            const caps = this.configParser.getCapabilities()
             const launcher = initialiseServices(config, caps, 'launcher')
 
             /**
--- a/packages/wdio-cli/src/launcher.js
+++ b/packages/wdio-cli/src/launcher.js
@@ -55,10 +55,9 @@ class Launcher {
 
     /**
      * run sequence
-     * @param   {Function}  callback    callback
      * @return  {Promise}               that only gets resolves with either an exitCode or an error
      */
-    async run (callback) {
+    async run () {
         /**
          * catches ctrl+c event
          */
@@ -101,9 +100,15 @@ class Launcher {
             await logger.waitForBuffer()
 
             this.interface.finalise()
-            return callback(exitCode)
+            return exitCode
         } catch (err) {
-            return callback(exitCode, err, !!onCompleteResults)
+            if (!onCompleteResults) {
+                /**
+                 * `shutdown` event is added by `exitHook`
+                 */
+                process.emit('shutdown', 1)
+            }
+            throw err
         }
     }
 
--- a/packages/wdio-cli/src/launcher.js
+++ b/packages/wdio-cli/src/launcher.js
@@ -55,49 +55,56 @@ class Launcher {
 
     /**
      * run sequence
-     * @return  {Promise} that only gets resolves with either an exitCode or an error
+     * @param   {Function}  callback    callback
+     * @return  {Promise}               that only gets resolves with either an exitCode or an error
      */
-    async run () {
-        let config = this.configParser.getConfig()
-        let caps = this.configParser.getCapabilities()
-        const launcher = initialiseServices(config, caps, 'launcher')
-
-        /**
-         * run pre test tasks for runner plugins
-         * (e.g. deploy Lambda function to AWS)
-         */
-        await this.runner.initialise()
-
-        /**
-         * run onPrepare hook
-         */
-        log.info('Run onPrepare hook')
-        await runOnPrepareHook(config.onPrepare, config, caps)
-        await runServiceHook(launcher, 'onPrepare', config, caps)
-
+    async run (callback) {
         /**
          * catches ctrl+c event
          */
-        exitHook(::this.exitHandler)
+        exitHook(:: this.exitHandler)
+        let exitCode
+        let onCompleteResults
 
-        let exitCode = await this.runMode(config, caps)
+        try {
+            let config = this.configParser.getConfig()
+            let caps = this.configParser.getCapabilities()
+            const launcher = initialiseServices(config, caps, 'launcher')
 
-        /**
-         * run onComplete hook
-         * even if it fails we still want to see result and end logger stream
-         */
-        log.info('Run onComplete hook')
-        await runServiceHook(launcher, 'onComplete', exitCode, config, caps)
+            /**
+             * run pre test tasks for runner plugins
+             * (e.g. deploy Lambda function to AWS)
+             */
+            await this.runner.initialise()
 
-        const onCompleteResults = await runOnCompleteHook(config.onComplete, config, caps, exitCode, this.interface.result)
+            /**
+             * run onPrepare hook
+             */
+            log.info('Run onPrepare hook')
+            await runOnPrepareHook(config.onPrepare, config, caps)
+            await runServiceHook(launcher, 'onPrepare', config, caps)
 
-        // if any of the onComplete hooks failed, update the exit code
-        exitCode = onCompleteResults.includes(1) ? 1 : exitCode
+            exitCode = await this.runMode(config, caps)
 
-        await logger.waitForBuffer()
+            /**
+             * run onComplete hook
+             * even if it fails we still want to see result and end logger stream
+             */
+            log.info('Run onComplete hook')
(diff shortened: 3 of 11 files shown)


tests/smoke.runner.js
```diff
@@ -208,10 +208,7 @@ const retryPass = async () => {
         await Promise.all(tests.map(test => test()))
     }
 
-    /**
-     * for some reason the process get stuck therefor exit it
```
mgrybyk: process should not get stuck anymore

tests/helpers/launch.js
```diff
@@ -3,30 +3,33 @@ import path from 'path'
 
 import Launcher from '../../packages/wdio-cli/src/launcher.js'
 
-export default function launch (...args) {
```
mgrybyk: only changes promises to callback

packages/wdio-local-runner/src/worker.js
```diff
@@ -69,7 +68,7 @@ export default class WorkerInstance extends EventEmitter {
             cwd: process.cwd(),
             env: runnerEnv,
             execArgv,
-            silent: true
+            stdio: ['inherit', 'pipe', 'pipe', 'ipc']
```
mgrybyk: pipe only stdout and stderr

packages/wdio-cli/src/run.js
```diff
@@ -64,9 +64,24 @@ export default function run (params) {
...
 
+/**
+ * launcher callback function
+ * @param {number|undefined}    exitCode    exit code
+ * @param {Error|undefined}     err         Error
+ * @param {boolean|undefined}   isComplete  if all tests and hooks were executed
+ */
+const launchCallback = (exitCode, err, isComplete) => {
+    if (!err) {
+        return { exitCode }
```
mgrybyk: return values used in unit tests only

packages/wdio-cli/src/launcher.js
```diff
@@ -55,49 +55,56 @@ class Launcher {
 
     /**
      * run sequence
-     * @return  {Promise} that only gets resolves with either an exitCode or an error
+     * @param   {Function}  callback    callback
+     * @return  {Promise}               that only gets resolves with either an exitCode or an error
      */
-    async run () {
```
mgrybyk: - run function now accepts callback as argument and calls it in the end;
- moved `exitHook(:: this.exitHandler)` to the top so that I can emit `shutdown` event in callback
- no other changes, just wrapped everything with try/catch block

packages/wdio-cli/src/interface.js
```diff
@@ -191,7 +192,7 @@ export default class WDIOCLInterface extends EventEmitter {
         /**
          * allow to exit repl mode via Ctrl+C
          */
-        if (this.inDebugMode) {
+        if (this.inDebugMode || this.endedNormally) {
```
mgrybyk: before this message was not printed in the very end because we used to kill process.

As far as we always call SIGTERM, we need to determine if process ended normally or Ctrl+C was really called.

packages/wdio-cli/src/run.js
```diff
@@ -64,9 +64,24 @@ export default function run (params) {
...
+    }
+    console.error(err)
+    if (!isComplete) {
+        process.emit('shutdown', 1)
```
mgrybyk: if an exception was thrown before tests and hooks were completed we need to send `shutdown` event to end process normally. 

It should be `shutdown`, not `end` because we use [async-exit-hook](https://www.npmjs.com/package/async-exit-hook) to do something in the very end.

packages/wdio-cli/src/run.js
```diff
@@ -64,9 +64,24 @@ export default function run (params) {
...
+ * launcher callback function
+ * @param {number|undefined}    exitCode    exit code
+ * @param {Error|undefined}     err         Error
+ * @param {boolean|undefined}   isComplete  if all tests and hooks were executed
+ */
+const launchCallback = (exitCode, err, isComplete) => {
+    if (!err) {
+        return { exitCode }
+    }
+    console.error(err)
```
mgrybyk: **IMPORTANT**:
If error was thrown we are not printing out reporters!!!  I think this behavior is invalid! 
I can change it, if anyone find it useful.

packages/wdio-cli/src/config.js
```diff
@@ -157,6 +157,7 @@ export const CLI_PARAMS = [{
     description: 'Cucumber options'
 }]
 
+/* istanbul ignore next */
```
mgrybyk: added task https://github.com/webdriverio/webdriverio/issues/4339

packages/wdio-cli/src/run.js
```diff
@@ -5,6 +5,7 @@ import Launcher from './launcher.js'
 import Watcher from './watcher'
 import setup from './setup'
 
+/* istanbul ignore next */
```
mgrybyk: added task https://github.com/webdriverio/webdriverio/issues/4339

packages/wdio-cli/src/setup.js
```diff
@@ -8,6 +8,7 @@ import yarnInstall from 'yarn-install'
 import { CONFIG_HELPER_INTRO, CONFIG_HELPER_SUCCESS_MESSAGE, QUESTIONNAIRE } from './config'
 import { getNpmPackageName, getPackageName, addServiceDeps } from './utils'
 
+/* istanbul ignore next */
```
mgrybyk: https://github.com/webdriverio/webdriverio/issues/4339

packages/wdio-cli/src/launcher.js
```diff
@@ -55,49 +55,56 @@ class Launcher {
...
 
-        /**
-         * run onPrepare hook
-         */
-        log.info('Run onPrepare hook')
-        await runOnPrepareHook(config.onPrepare, config, caps)
-        await runServiceHook(launcher, 'onPrepare', config, caps)
+        try {
+            let config = this.configParser.getConfig()
+            let caps = this.configParser.getCapabilities()
```
CrispusDH: caps and config can be const

packages/wdio-cli/src/run.js
```diff
@@ -64,9 +65,24 @@ export default function run (params) {
...
+    }
+    return { err, isComplete }
+}
+
+export function launch (wdioConf, params) {
     const launcher = new Launcher(wdioConf, params)
-    launcher.run().then(
-        (code) => process.nextTick(() => process.exit(code)),
-        (e) => process.nextTick(() => { throw e }))
+    launcher.run(launchCallback)
```
CrispusDH: `run` is async function, we need to resolve promise

mgrybyk: We use callback instead

mgrybyk: I can change implementation to use promises instead of callback

packages/wdio-cli/src/launcher.js
mgrybyk: Yeah, I just copy pasted existing code, didn't want to improve it or modify it. 
Let me change it

packages/wdio-cli/src/run.js
CrispusDH: I think it would be more convenient.

mgrybyk: Hey, sorry for confusing, we are not waiting for any result from async function, so there is no need in then/catch here

packages/wdio-cli/src/launcher.js
mgrybyk: @CrispusDH fixed

packages/wdio-cli/src/run.js
CrispusDH: In my opinion it would be better to do in separate PR.

mgrybyk: agree

tests/helpers/launch.js
```diff
@@ -2,11 +2,13 @@ import fs from 'fs'
 import path from 'path'
 
 import Launcher from '../../packages/wdio-cli/src/launcher.js'
+import { launchCallback } from '../../packages/wdio-cli/src/run.js'
 
 export default function launch (...args) {
     const launcher = new Launcher(...args)
 
-    return launcher.run().then((isFailing) => {
+    return launcher.run(launchCallback).then(({ err, exitCode }) => {
```
christian-bromann: the `onFulfilled` handle should not have this err, result payload. We are mixing callback with promise patterns here.

mgrybyk: can you please help with this one? 
I should not do here `.then(result => function() {})` ?

packages/wdio-cli/src/run.js
```diff
@@ -64,9 +65,28 @@ export default function run (params) {
...
 }
 
-function launch (wdioConf, params) {
+/**
+ * launcher callback function
+ * @param {number|undefined}    exitCode    exit code
+ * @param {Error|undefined}     err         Error
+ * @param {boolean|undefined}   isComplete  if all tests and hooks were executed
+ */
+export const launchCallback = (exitCode, err, isComplete) => {
```
christian-bromann: I am not sure if we need this function. Doing `process.exit(code)` was definitely not a great idea. Ideally the `launch` method should end with:

```js
return launcher.run()
```

The reason why processes in Node hang around is that not all event listeners are removed. We should try to ensure that clear up all event listeners so that processes can shutdown gracefully.

mgrybyk: @christian-bromann we are calling `shutdown` event that ends child processes gracefully, even if there are some listeners it will be killed.
The idea of this callback is to identify if need to call it or no

mgrybyk: the reason why it was hanging before is that we were not releasing properly STDIN after piping it.

tests/helpers/launch.js
mgrybyk: removed callback

packages/wdio-cli/src/run.js
```diff
@@ -64,9 +65,7 @@ export default function run (params) {
     })
 }
 
-function launch (wdioConf, params) {
+export function launch (wdioConf, params) {
     const launcher = new Launcher(wdioConf, params)
-    launcher.run().then(
-        (code) => process.nextTick(() => process.exit(code)),
-        (e) => process.nextTick(() => { throw e }))
+    return launcher.run().catch(err => { console.error(err) })
```
christian-bromann: Why do we log the error instead of letting the user deal with it?

packages/wdio-cli/src/launcher.js
```diff
@@ -55,49 +55,61 @@ class Launcher {
...
-        this.interface.finalise()
-        return exitCode
+            await logger.waitForBuffer()
+
+            this.interface.finalise()
+            return exitCode
+        } catch (err) {
+            if (!onCompleteResults) {
+                /**
+                 * `shutdown` event is added by `exitHook`
```
christian-bromann: Do you mean is caught by `exitHook`?

mgrybyk: it is defined few lines above, see https://github.com/webdriverio/webdriverio/pull/4332/files#diff-03f312cc95c9f2a6588676367c2fcbe8R64

packages/wdio-cli/src/run.js
mgrybyk: sorry, missed that after debugging, removed.

mgrybyk:
forgot to remove WIP label

mgrybyk:
@christian-bromann do you have any comments/suggestions on this one?

mgrybyk:
@christian-bromann it works, see 
image

mgrybyk:
We need it here https://github.com/webdriverio/webdriverio/blob/f5fa98472a3deedcc2c059cf73ab8b7240c035b5/packages/wdio-local-runner/src/stdStream.js#L2

@christian-bromann

PR #4347, opened 2019-08-09
type: fix `object` type returned by protocol command

### LINKED ISSUE #4301: Property does not exist on type if commands' type def return `object`
 - **WebdriverIO version:** 5.11.7

**Config of WebdriverIO**
Typescript

**Describe the bug**
getWindowRect() for example. It occur on all commands which have return type as `obejct` in type def.

```js
// Property 'height' does not exist on type '{}'.ts(2339)
const { height, width } = browser.getWindowRect()

// type in webdriver.d.ts
getWindowRect(): object;

// Have to workaround with
const { height, width }: any = browser.getWindowRect()
```

JieC:
## Proposed changes
Fix #4301
Mapped `object` returned in `webdriver.json` and `appium.json` with proper types
Others to `ProtocolCommandResponse` as workaround

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/scripts/templates/webdriver.tpl.d.ts
+++ b/scripts/templates/webdriver.tpl.d.ts
@@ -397,16 +397,14 @@ declare namespace WebDriver {
     }
 
     // webdriver.json
-    interface SessionReturn {
-        sessionId: string,
-        capabilities: {
-            [key: string]: any
-        }
+    interface SessionReturn extends DesiredCapabilities {
+        // addtional capabilities return by remote
+        [key: string]: any;
     }
 
     interface StatusReturn {
-        ready: boolean,
-        message: string,
+        ready?: boolean,
+        message?: string,
         [key: string]: any
     }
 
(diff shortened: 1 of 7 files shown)


scripts/templates/webdriver.tpl.d.ts
```diff
@@ -391,6 +391,46 @@ declare namespace WebDriver {
...
+        width: number,
+        height: number
+    }
+    // appium.json
+
+    interface StringsReturn {
+        [key:string]: string
+    }
+
+    interface SettingsReturn {
```
CrispusDH: SettingsReturn and ProtocolCommandResponse are the same, it might be better leave just one with some more abstract naming

JieC: Changed `SettingsReturn`
If use `unknown`, we cannot access prop on it.

scripts/templates/webdriver.tpl.d.ts
```diff
@@ -391,6 +391,61 @@ declare namespace WebDriver {
...
+        width: number,
+        height: number
+    }
+    // appium.json
+
+    interface StringsReturn {
+        [key:string]: string
+    }
+
+    interface SettingsReturn{
```
mgrybyk: missing space before `{`

scripts/templates/webdriver.tpl.d.ts
```diff
@@ -391,6 +391,61 @@ declare namespace WebDriver {
...
+
+    interface RectReturn {
+        x: number,
+        y: number,
+        width: number,
+        height: number
+    }
+    // appium.json
+
+    interface StringsReturn {
```
mgrybyk: Can you please have only one interface for `[key:string]: string` and extended other interfaces?
ex:
`interface SessionReturn extends DesiredCapabilities, ProtocolCommandResponse { }`

jayrepo: Done

mgrybyk: Can we remove this one and use just ProtocolCommandResponse ?

scripts/templates/webdriverio.tpl.d.ts
```diff
@@ -16,6 +16,8 @@ declare namespace WebdriverIO {
         height: number
     }
 
+    interface RectReturn extends WebDriver.RectReturn {}
```
mgrybyk: Why do we need to declare it here? Can we just use WebDriver.RectReturn in command?

scripts/templates/webdriver.tpl.d.ts
JieC: The value here is `string`, while `any` in `ProtocolCommandResponse`.

scripts/templates/webdriverio.tpl.d.ts
JieC: Yes, you are right, I will remove it

mgrybyk:
Looks good to me

PR #4354, opened 2019-08-12
Wrap test function with before after step hooks

### LINKED ISSUE #4004: WDIO hook for it blocks
 - **WebdriverIO version:** [5.8.6]
 - **Mode:** [WDIO Testrunner]
 - **Node.js version:** [10.15.3]
 - **NPM version:** [6.4.1]
 - **Browser name and version:** [Chrome 64]
 - **Platform name and version:** [macOs Mojave]
 - **Additional wdio packages used (if applicable):** [@wdio/mocha-framework": "5.8.1"]

**Describe the bug**
It's very difficult to take screenshot in test failing time because if I have next test structure
```js
describe('test', () => {
afterEach(() => {
 after commands block
})
```

```js
it('step1', ()=> {
  commands block
  expect(smth)
})
})
```

And in `wdio.conf`:
```js
afterTest: function (test) {
    if (test.error !== undefined) {
      browser.takeScreenshot();
    }
  }
```

Screenshot makes after "after commands block" and I can't see test failing reasons

**To Reproduce**
You can reproduce it by next steps:

***test.js***
```js
describe('test', () => {
afterEach(() => {
 browser.pause(2000)
})

it('step1', ()=> {
  console.log('into test ', new Date())
  expect(false).to.be.true
})
})
```

***wdio.conf.js***
```js
...
  afterTest: function (test) {
    console.log('after test ', new Date())
  },
...
```

So you see something like:
```
[0-0] into test 2019-05-22T09:20:35.450Z
[0-0] after test  2019-05-22T09:20:37.521Z
```

And

**Expected behavior**
`afterTest` function runs after "it"

mgrybyk:
## NOTE

Please **bump version of all the packages** when doing release!

## Problem

Currently we use mocha/jasmine before/afterEach hooks as wdio before/afterTest hooks.
This has certain limitations, for example if test looks like this
```
describe('suite', () => {
  describe('sub suite', () => {
    it('test', () => {
      // some test
    })

    afterEach(() => {
      // reload page
    })
  })
})
```
there is no way to capture screenshot, because screenshot is taken after `afterEach` hook.

## Proposed changes

- Wrap mocha/jasmine framework functions (`it`, `beforeEach` and others) with wdio hooks like this:
``` 
global.it = function (..args) {
  // call wdio before test/hook
  // call original `it`/`afterEach`/etc function
  // call wdio after test/hook
}
```
fixes #4004

- Additionally mocha users can now access mocha runnable context that adds more flexibility for custom reporters (ex. Mochawesome with screenshots)
```
afterTest: function (test, context) {
    context // { test, _runnable }
},
```
same for hooks
```
afterHook: function (test, context) {
    context // { test, _runnable, currentTest }
},
```

- have same approach for all test frameworks: mocha, jasmine and cucumber. See file `packages/wdio-utils/src/test-framework/testFnWrapper.js`

- allow mixing async and sync like this:
```
it('async test', async () => {
  await browser.pause(1)
})

it('sync test' () => {
  browser.pause(1) // it was Promise before
})
```

## Types of changes

there might be some regressions for mocha/jasmine users that strongly rely on beforeTest/afterTest in some very specific way, sorry to say I might miss some use cases, it should be easy to deal with them.
Cucumber users are not affected.

## TODO

- [x] have same approach to run before/after test hooks for all frameworks (mocha, cucumber, jasmine)
- [x] verify before/after test hooks in async mode
- [x] add @wdio/sync unit tests
- [x] verify before/afterTest hook functions signature (ideally to avoid breaking changes)
- [x] add smoke tests
- [x] update before/after hook signature in configs and types
- [x] avoid adding dependencies to @wdio/utils
- [x] reorganize smoke added test


mgrybyk:
unit tests were not updated, however smoke tests are passing.
Have to finish TODO items before updating unit tests

mgrybyk:
unit tests were not updated, however smoke tests are passing.
Have to finish TODO items before updating unit tests

packages/wdio-utils/src/testFnWrapper.js
```diff
@@ -0,0 +1,104 @@
+import { isFunctionAsync } from './utils'
+
+/**
+ * wraps test framework spec/hook function with WebdriverIO before/after hooks
+ *
+ * @param   {object} wrapFunctions  executeHooksWithArgs, executeAsync, runSync
+ * @param   {string} type           Test/Step or Hook
+ * @param   {object} spec           specFn and specFnArgs
```
mgrybyk: not sure it makes much sense to pass spec, before, after as objects

mgrybyk:
PR is ready for review, will do docs changes a bit later.

packages/wdio-mocha-framework/src/index.js
```diff
@@ -97,14 +95,24 @@ class MochaAdapter {
         const match = MOCHA_UI_TYPE_EXTRACTOR.exec(options.ui)
         const type = (match && INTERFACES[match[1]] && match[1]) || DEFAULT_INTERFACE_TYPE
 
+        const hookArgsFn = (context) => {
+            const test = context.test || {}
+            const parentTitle = test.parent ? test.parent.title : ''
+            return [{ ...test, parent: parentTitle || '' }, context]
```
CrispusDH: I think parentTitle will always exists, you initialized it line before. Or you could think that test.parent.title can not exist even if test.parent will exist?

packages/wdio-utils/src/test-framework/testFnWrapper.js
```diff
@@ -0,0 +1,68 @@
...
+ * @param   {object} before         beforeFn and beforeFnArgs
+ * @param   {object} after          afterFn and afterFnArgs
+ * @param   {string} cid            cid
+ * @param   {number} repeatTest     number of retries if test fails
+ * @return  {*}                     specFn result
+ */
+export const testFrameworkFnWrapper = async function (
+    { executeHooksWithArgs, executeAsync, runSync },
+    type,
+    { specFn, specFnArgs },
```
CrispusDH: Should `specFnArgs` have default value as `beforeFnArgs` and `afterFnArgs`?

packages/wdio-utils/src/test-framework/testInterfaceWrapper.js
```diff
@@ -0,0 +1,121 @@
...
+import { testFnWrapper } from '@wdio/config'
+import { filterSpecArgs } from '../utils'
+
+/**
+ * runs a hook within fibers context (if function name is not async)
+ * it also executes before/after hook hook
+ *
+ * @param  {Function} hookFn        function that was passed to the framework hook
+ * @param  {Function} origFn        original framework hook function
+ * @param  {Function} beforeFn      before hook hook
```
CrispusDH: Double`hook` in description

packages/wdio-utils/src/test-framework/testInterfaceWrapper.js
```diff
@@ -0,0 +1,121 @@
...
+ * @param  {Function} beforeFnArgs  function that returns args for `beforeFn`
+ * @param  {Function} afterFn       after hook hook
```
CrispusDH: Same as above

packages/wdio-mocha-framework/src/index.js
mgrybyk: I think test.parent.title might not exist in some cases, but I'm not sure.
Have to check if test.parent.title value exists for root suite for both it and before

packages/wdio-utils/src/test-framework/testFnWrapper.js
mgrybyk: I have set default values for after fn args because I'm modifying it.

packages/wdio-mocha-framework/src/index.js
mgrybyk: @CrispusDH updated.

christian-bromann:
Please remove the work-in-progress label if it is ready for review

packages/wdio-mocha-framework/tests/adapter.test.js
```diff
@@ -322,52 +334,52 @@ test('getUID', () => {
     expect(adapter.getUID({ type: 'hook:start' })).toBe('hook-0-1')
     expect(adapter.getUID({ type: 'hook:end' })).toBe('hook-0-1')
     expect(adapter.getUID({ type: 'suite:start' })).toBe('suite-0-0')
-        expect(adapter.getUID({ type: 'hook:start' })).toBe('hook-00-0')
```
mgrybyk: oops, have to rollback the change, accidentally reformated the file

mgrybyk:
PR is ready for review

packages/wdio-config/src/shim.js
```diff
@@ -25,7 +25,7 @@ export let executeHooksWithArgs = async function executeHooksWithArgsShim (hooks
             result = hook.apply(null, args)
         } catch (e) {
             log.error(e.stack)
-            return resolve()
+            return resolve(e)
```
christian-bromann: Are we using this error somewhere? I am not sure if it is a good idea to pass the error as result.

packages/wdio-config/src/shim.js
```diff
@@ -35,7 +35,7 @@ export let executeHooksWithArgs = async function executeHooksWithArgsShim (hooks
         if (result && typeof result.then === 'function') {
             return result.then(resolve, (e) => {
                 log.error(e.stack)
-                resolve()
+                resolve(e)
```
christian-bromann: same

packages/wdio-cucumber-framework/tests/adapter.test.js
```diff
@@ -25,7 +25,14 @@ const wdioReporter = {
...
 const adapterFactory = (cucumberOpts = {}) => new CucumberAdapter(
     '0-2',
-    { cucumberOpts },
+    {
+        cucumberOpts,
+        beforeStep: 'beforeStep',
+        afterStep: 'afterStep',
+        beforeHook: 'beforeHook',
+        afterHook: 'afterHook',
+
```
christian-bromann:
```suggestion
```

packages/wdio-cucumber-framework/tests/adapter.test.js
```diff
@@ -25,7 +25,14 @@ const wdioReporter = {
 
 const adapterFactory = (cucumberOpts = {}) => new CucumberAdapter(
     '0-2',
-    { cucumberOpts },
+    {
+        cucumberOpts,
+        beforeStep: 'beforeStep',
+        afterStep: 'afterStep',
+        beforeHook: 'beforeHook',
+        afterHook: 'afterHook',
```
christian-bromann:
```suggestion
        afterHook: 'afterHook'
```

packages/wdio-jasmine-framework/src/index.js
```diff
@@ -75,15 +76,23 @@ class JasmineAdapter {
...
-            fnName
-        ))
+        INTERFACES['bdd'].forEach((fnName) => {
+            const isTest = TEST_INTERFACES.includes(fnName)
+            runTestInFiberContext(
+                isTest,
+                isTest ? this.config.beforeTest : this.config.beforeHook,
+                hookArgsFn,
+                isTest ? this.config.afterTest : this.config.afterHook,
+                hookArgsFn,
```
christian-bromann: Same args for both hooks? Shouldn't the `afterTest` hook contain the result of the test?

packages/wdio-mocha-framework/src/index.js
```diff
@@ -97,14 +95,22 @@ class MochaAdapter {
...
             runTestInFiberContext(
-                [testCommand, testCommand + '.only'],
-                this.config.beforeHook,
-                this.config.afterHook,
-                fnName
+                isTest,
+                isTest ? this.config.beforeTest : this.config.beforeHook,
+                hookArgsFn,
+                isTest ? this.config.afterTest : this.config.afterHook,
+                hookArgsFn,
```
christian-bromann: same

packages/wdio-utils/package.json
```diff
@@ -31,6 +31,7 @@
     "url": "https://github.com/webdriverio/webdriverio/issues"
   },
   "dependencies": {
+    "@wdio/config": "^5.13.0-alpha.0",
```
christian-bromann: I will release all packages again, we need to change this.

tests/mocha/service.js
```diff
@@ -1,7 +1,54 @@
 import assert from 'assert'
 
-describe('my feature', () => {
-    it('should do stuff', () => {
+const sleep = (ms) => new Promise(resolve => setTimeout(resolve, ms))
+
+before(async () => {
+    await sleep(10)
```
christian-bromann: For what are we sleeping for?

tests/helpers/config.js
```diff
@@ -66,4 +68,53 @@ exports.config = {
...
+            throw new Error('wdio afterTest error: ' + test.title)
+        }
+    },
+    beforeHook(test) {
+        if (global.WDIO_SERVICE_TEST_HOOK_DURATION === 0) {
+            throw new Error('wdio beforeHook error: ' + test.title)
+        }
+    },
+    async afterHook(test, context, { error, duration, passed }) {
+        await sleep(20)
```
christian-bromann: needed?

tests/mocha/service.js
```diff
@@ -1,7 +1,54 @@
...
+})
+
+describe('my feature 1', () => {
```
christian-bromann: I wonder if we should better structure these smoke tests. This test here was suppose to check of commands are being run synchronously. I think it makes sense to create an extra test file that tests the hooks.

packages/wdio-config/src/shim.js
mgrybyk: Yes, we use it, I think we have just missed this behavior in shim.
It is used in logHookError

packages/wdio-jasmine-framework/src/index.js
mgrybyk: It should and it's appended later on, we can't get result at this point.
BeforeTest and AfterTest hooks are executed while test is in progress

packages/wdio-utils/package.json
mgrybyk: Yeap, will rebase later on

tests/helpers/config.js
mgrybyk: We aren't waiting for anything here, I just want to make sure that async command is executed properly. 
I think it would be better to replace it with browser.sleep, am I right?

tests/mocha/service.js
mgrybyk: Nothing, just make sure async command works.
It can be replaced with browser.pause

mgrybyk:
> Do you think we can reorg this?

I'll think on how can it be improved.

tests/helpers/config.js
christian-bromann: yes

tests/mocha/service.js
christian-bromann: does it mean that they are executed at the same time?

packages/wdio-jasmine-framework/src/index.js
christian-bromann: > It should and it's appended later on

Where does it get appended? Is this a breaking change?

mgrybyk: It's not a breaking change even though the signature has changed. I'd say it's new feature. 

See 
https://github.com/mgrybyk/webdriverio/blob/ee9a217e3330ddafdf02889a6e0b7530243a28f2/packages/wdio-utils/src/test-framework/testFnWrapper.js

Here is an example for mocha.

Original `it` is replaced by `testFrameworkFnWrapper` that calls not just original `it` but also runs beforeTest before and afterTest after.

We know original `it` result once it is executed in `testFrameworkFnWrapper`, now result can be appended to afterTest args.

Users can use hook as before
```
afterTest(test) {
  if (test.passed) {}
}
```
or in new format (same as afterStep in cucumber works)
```
afterTest(test, context, result) {
  if (result.passed) {}
}
```

Finally we now have same approach to run hooks across all test frameworks and modes (sync, async).

tests/mocha/service.js
mgrybyk: I think I missed something here.

I just want to make sure that mixing async and sync modes in same spec is not causing any unpredictable behavior. 
The reason for doing such test is that testFrameworkFnWrapper runs async function without fibers, nothing new here, it worked same way before, but still have to be tested.

packages/wdio-cucumber-framework/tests/adapter.test.js
mgrybyk: we can add a linter rule to either have or not the comma in the end. The comma is required in one of my projects, always missing it now

mgrybyk:
@christian-bromann updated PR according to comments

mgrybyk:
added more unit tests and removed `runFnInFiberContextWithCallback` from @wdio/sync as far as it is not used anywhere

mgrybyk:
@christian-bromann rebased, added temporary pac-resolver mock https://github.com/webdriverio/webdriverio/pull/4354/commits/f5ae88d85aa71afb29d274dfd02c98c0b14a3c7f to prove that all tests are passing

packages/wdio-config/tests/shim-sync.test.js
```diff
@@ -0,0 +1,52 @@
+import { executeHooksWithArgs, runFnInFiberContext, wrapCommand, hasWdioSyncSupport, executeSync, executeAsync, runSync } from '../src/shim'
+
+jest.mock('@wdio/sync', () => ({
```
christian-bromann: I am not sure about the value of testing module exports. If an export is missing I am sure other unit tests will fail as well.

christian-bromann: I don't mind to keep it as is

packages/wdio-sync/src/wrapCommand.js
```diff
@@ -35,9 +35,8 @@ export default function wrapCommand (commandName, fn) {
         }
         /**
          * all named nested functions run in parent Fiber context
-         * except of debug and waitUntil
          */
-        this._NOT_FIBER = fn.name !== '' && fn.name !== 'debug' && commandName !== 'waitUntil'
+        this._NOT_FIBER = fn.name !== ''
```
christian-bromann: What is the reason behind this change?

mgrybyk: It's enough to delete flag here https://github.com/mgrybyk/webdriverio/blob/f5ae88d85aa71afb29d274dfd02c98c0b14a3c7f/packages/wdio-sync/src/runFnInFiberContext.js to make waitUntil, debug and other functions that use runFnInFiberContext work properly.
Ideally I'd not rely on function name at all, but I don't have a solution right now.

packages/wdio-config/tests/shim-sync.test.js
mgrybyk: Good point, I'll check

mgrybyk: @christian-bromann I'd insist to keep the test.
If I comment this line https://github.com/webdriverio/webdriverio/pull/4354/files#diff-ab58d4080942109b9f399dc7e8bf88feR70

all the unit tests are passing. Even though smoke tests are failing it would take an additional time to debug why failure like `Expected [object Promise] to be 'Mock Page Title'.` has happened.

mgrybyk:
@christian-bromann updated PR according to comments.

Please note that due to lot's of changes in imports I'd recommend to release all the packages after merging this one

Issue #4346, opened 2019-08-09
Does all minor versions of webdriverio v5 require nodejs >=10.0.0 ?


pk-saxena:
The documentation says v5 requires nodejs version v8.11.2 or higher. But on building the project, I see an error that one of the dependency of wdio-sync, namely fibers 4.0.1 requires nodejs 10 or higher. Does any minor version of v5 supports running with nodejs < 10 ?

The `package.json` file in our project has following content
```
  "engines": {
    "node": ">=8.x",
    "yarn": ">=1.9.x"
  }
  "dependencies": {
    "@wdio/cli": "^5.11.0",
 }
  "devDependencies": {
    "@wdio/dot-reporter": "^5.11.0",
    "@wdio/local-runner": "^5.11.2",
    "@wdio/mocha-framework": "^5.11.0",
    "@wdio/selenium-standalone-service": "^5.11.2",
    "@wdio/spec-reporter": "^5.11.0",
    "@wdio/sync": "^5.11.0",
    "wdio-chromedriver-service": "^5.0.2",
    "webdriverio": "^5.11.0",
 }
```
As part of building the project on CI, we run the command:
`yarn install --frozen-lockfile`
On doing this, I see the error:
`error fibers@4.0.1: The engine "node" is incompatible with this module. Expected version ">=10.0.0". Got "8.16.0"`

yarn.lock file contains:
```
"@wdio/sync@^5.11.0":
  version "5.11.0"
  resolved "https://registry.yarnpkg.com/@wdio/sync/-/sync-5.11.0.tgz#aec0dfd75c8b1cc7a1de6276fa5fd66c9e8aad51"
  integrity sha512-YRrdmLtJaRvJV4h9Bg33oI90g8qbPFechLBGjq4cPEohJa+gIS3vabcyjZ7kvHxEST+/ITieSV/CNXnGA7u44w==
  dependencies:
    "@wdio/config" "^5.11.0"
    "@wdio/logger" "^5.11.0"
    fibers "^4.0.0"
```

patthiel:
Encountered this as well. Mentioned in here: https://github.com/webdriverio/webdriverio/pull/4305

The CI build should have caught this, as it uses node8. 

I'm happy to submit a PR to downgrade Fibers to 3, which should make the project compatible with node 8 again. However, I do not know if there is specific stuff we want / need from Fibers ^4.

mgrybyk:
I'd just drop node8

mgrybyk:
[...]

PR #4356, opened 2019-08-13
Update minimum node version requirement to Node 10

patthiel:
## Proposed changes

As raised in https://github.com/webdriverio/webdriverio/pull/4305 and https://github.com/webdriverio/webdriverio/issues/4346, the minimum node version for webdriverio should be updated to Node 10 due to [dependency](https://www.npmjs.com/package/fibers#supported-platforms)(s) that require it.

I've tried to update the node version wherever I could find it, but i'm sure there's a spot or two that i missed.. Please let me know and i'll update this PR. The CI build passes locally and in a local `node:10-jessie` container.

## Types of changes

- [x] Breaking change 

## Checklist

## Further comments


kellyselden:
Does this project use SemVer? If so, anyone using webdriver unpinned on node 8 will start breaking (as we did). This should be reverted in v5 and rereleased as v6.

patthiel:
@kellyselden I kind of agree, that's why I marked the change as breaking (but i'm also not sure if this project uses SemVer).

However, wdio-sync already relies on a dependency that is only compatible with  >=node 10 and any version of that library beyond `wdio-sync "5.7.13"` will throw a node compatibility error when upgrading. This change makes official, a [change](https://github.com/webdriverio/webdriverio/pull/3898) in compatibility that was made months ago.

As to whether we should reverted this change and re-release it as v6, I would defer to @webdriverio/technical-committee. If reverted, i would recommend downgrading Fibers to ensure this whole monorepo is compatible with node8.

christian-bromann:
@patthiel is right here. The compatibility problem was already caused by the fibers update month ago. So we messed up already at that point which makes it useless to revert this change now. We are definitely using semver in the project and gonna upgrade accordingly the next time when we drop any Node.js support. I apologies if this has caused any inconvenience.

harlentan:
This commit just break my npm install!

My project run in node 8 for a long time.

harlentan:
What is your problem!

@christian-bromann

wswebcreation:
@harlentan,

What do you mean with your remark?

kellyselden:
I would like to argue for sticking to SemVer. It would be a benefit to your consumers with not much work (reverting the two commits). Even though this is an open project run by volunteers and can obviously do whatever it wants, here is my case:

1. My internal company project that wraps webdriver supports all active Node LTSs, so I would have to pin to the version before SemVer break.
1. My org (Ember.js) also supports all active Node LTSs, so my plans to start promoting a bridge between the two will either have to wait until December (Node 8 EOL), or release with a old pinned version (lame).
1. People would be losing out on your bugfixes until December.
1. Basically I'm arguing that breaking SemVer and dropping Node early is more impactful to your consumers than you may realize.

I'm saying this with love as a fellow open-source junkie. Please reconsider :)

I love the project BTW.

mgrybyk:
I'm against reverting PR but we may raise another PR with changing minimum version to 8 and downgrading fibers version to 3.0.1

mgrybyk:
Here you go, let's discuss it here https://github.com/webdriverio/webdriverio/pull/4360

christian-bromann:
Wrote a short post mortem on this issue: https://gist.github.com/christian-bromann/1b6245c04b894b5b989accdc90eb250d

PR #4360, opened 2019-08-14
Proposal: change minimum node version to 8

mgrybyk:
## Proposed changes

First of all I'd better stay with nodejs 10 as minimum version.

However some customers complaining on it and there are no critical reasons to have version 10 as minimum.

Additionally we may publish some post / message saying that node 8 will be dropped at December.

## Types of changes

## Checklist


packages/wdio-sync/package.json
```diff
@@ -36,7 +36,7 @@
   "dependencies": {
     "@wdio/config": "^5.12.0",
     "@wdio/logger": "^5.12.0",
-    "fibers": "^4.0.0"
+    "fibers": "^3.1.0"
```
mgrybyk: fibers is now on 3.1.0

mgrybyk:
requesting change here:
we need to install appropriate node-fibers version based on nodejs version, something like:
if nodejs version < 10 install node-fibers v3
otherwise install node-fibers v4

christian-bromann:
@kellyselden if you use Node v8 how can you run WebdriverIO anyway if Fibers is not supported for that Node version? This change to require Node v10 shouldn't have affected anyone given that we run on Fibers v4 for a while now.

kellyselden:
```
$ yarn why fibers
yarn why v1.15.2
[1/4] 🤔  Why do we have the module "fibers"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
error We couldn't find a match!
```

@christian-bromann I guess because I'm only the `webdriverio` NPM package. Then updating the lockfile yesterday gave me

```
error webdriverio@5.12.0: The engine "node" is incompatible with this module. Expected version ">= 10.13.0". Got "8.16.0"
error Found incompatible module
```

christian-bromann:
@kellyselden ok, so you are not using WebdriverIO in sync mode. We currently investigate if Fibers v4 would work in Node 10

mgrybyk:
tested postinstall script on MacOS and Window 10 Pro 1903, 64 bit.

kellyselden:
I think your solution will still break SemVer. Anyone that has `npm config set ignore-scripts true` because of [security concerns](https://blog.npmjs.org/post/141702881055/package-install-scripts-vulnerability) will now be breaking.

christian-bromann:
> I think your solution will still break SemVer.

This will be the minority thought. I rather take this risk instead of having people manually upgrade to v6 and missing all the upcoming bugfixes.

christian-bromann:
Wrote a short post mortem on this issue: https://gist.github.com/christian-bromann/1b6245c04b894b5b989accdc90eb250d

PR #4361, opened 2019-08-14
use `optionalDependencies` instead of `postinstall` for `fibers`

kellyselden:
## Proposed changes

Here is my attempt to address https://github.com/webdriverio/webdriverio/pull/4360#issuecomment-521254675. It avoids the `npm config set ignore-scripts true` issue, and also has a fallback for Node 10 users that install with `--no-optional`.

The optional dependency will succeed in Node 10, and therefore be used. It will fail gracefully in Node 8, and won't be used. Node 8 users that install with `--ignore-engines` also won't get it because of the runtime Node version check.

Unfortunately, I had to use a commit SHA instead of a git tag because `fibers` stopped tagging versions. I won't be upset if you don't accept this.

## Types of changes

## Checklist

## Further comments


kellyselden:
I'm getting lint errors, and I don't know where I went wrong. Do you see anything obvious?

```
   1:20  error  Unable to resolve path to module 'fibers'           import/no-unresolved

   2:21  error  Unable to resolve path to module 'fibers/future'    import/no-unresolved

  12:25  error  Unable to resolve path to module 'fibersv4'         import/no-unresolved

  13:26  error  Unable to resolve path to module 'fibersv4/future'  import/no-unresolved
```

christian-bromann:
@kellyselden thanks for the patch! I looked into it and it seems that the behavior is as expected with Yarn but fails when installing things with NPM. Our CI uses NPM and with that no fibers package is being installed. I believe it is because both packages are having the same name. So I went ahead and pushed a new package [`fibersv4`](https://www.npmjs.com/package/fibersv4) to NPM and tried the same. My dependencies looked as follows:

```json
  "dependencies": {
    "@wdio/config": "^5.12.1",
    "@wdio/logger": "^5.12.1",
    "@wdio/utils": "^5.12.1",
    "fibersv4": "^4.0.1"
  },
  "optionalDependencies": {
    "fibers": "^3.1.0"
  },
```

The problem with that is it fails installing it with Yarn and Node v8 and exchanging fibersv4 and fibers fails when using NPM and Node 12. 🤷‍♂

christian-bromann:
I think I found a solution. If I put just both of them into `optionalDependencies` it will just ignore if one of the installs fails. See https://github.com/webdriverio/webdriverio/compare/cb-fibers-fix?expand=1

christian-bromann:
> and also has a fallback for Node 10 users that install with `--no-optional`.

Dang, my solution will fail on that.

kellyselden:
Looks like you were right. Using your package, it now passes linting. Does this version work in your test cases?

christian-bromann:
closing this in favor of #4365. Thanks for pointing out the opportunity using this workaround.

PR #4365, opened 2019-08-15
use `optionalDependencies` instead of `postinstall` for `fibers

christian-bromann:
My solution for #4361


mgrybyk:
Is it required to have minimum node version 10 for applitools and devtools services?

mgrybyk:
Can test coverage be increased?

packages/wdio-sync/src/fibers.js
```diff
@@ -0,0 +1,49 @@
...
+try {
+    /**
+     * try original fibers package first
+     */
+    // eslint-disable-next-line import/no-unresolved
+    Fiber = require('fibers')
+    // eslint-disable-next-line import/no-unresolved
+    Future = require('fibers/future')
+} catch (e) {
+    log.warn('Couldn\'t load fibers package for Node v10 and above')
```
mgrybyk: I think log level here may be trace or debug.

christian-bromann:
> Is it required to have minimum node version 10 for applitools and devtools services?

Yes, applitools requires for whatever reason selenium-webdriver, I made an issue for that here: eyes.webdriverio.javascript5#21 and devtools service requires Lighthouse which has a node dependency on 10.x

> Can test coverage be increased?

Which part of the code are you referring to?

christian-bromann:
> Can test coverage be increased?

Added unit test for new file.

kellyselden:
I don't understand how this is better. It goes from throwing for people that use `npm config set ignore-scripts`, to throwing for people that use `--no-optional`. You need the common denominator (fibers 3) to be in `dependencies` https://github.com/webdriverio/webdriverio/pull/4361/files#diff-d4bda84032a665bb0361a2532fe3460fR40.

mgrybyk:
We can't use fibers 3 for nodejs 12

mgrybyk:
Do you know any way to install dependency based on nodejs version that won't break anything @kellyselden

christian-bromann:
> You need the common denominator (fibers 3) to be in `dependencies`

This would throw the same way for people using Yarn and Node.js v12. This solution throws for no one. You are right that it breaks if you have `--no-optional` set. However I think not that many people use this flag and it definitely less people than people using Yarn and Node v12.

kellyselden:
Here is how I perceive things

v3 in optional with `--no-optional`:

 * Node 12: Doesn't work
 * Node 8: Doesn't work

v3 in deps with `--no-optional`:

 * Node 8: Works
 * Node 12: Doesn't work

v3 in optional without `--no-optional`:

 * Node 12: Works
 * Node 8: Works

v3 in deps without `--no-optional`:

 * Node 8: Works
 * Node 12: Works

It seems to me like it could be better.

christian-bromann:
There is one major difference with v3 in deps vs optional: everyone on Node v12 using Yarn will have a failing install with having v3 in deps. My patch assumes that the number of people installing Node dependencies with `--no-optional` is much smaller than people running Node v12 with Yarn.

kellyselden:
Oh I wasn't aware yarn couldn't install v3. that's strange.

christian-bromann:
Yeah, unfortunately if a package install fails with Yarn the whole dependency install fails and will be rolled back.

PR #4369, opened 2019-08-15
Make WebdriverIO to support native app testing on Windows with WinAppDriver

licanhua:
## Proposed changes

[WinAppDriver ](https://github.com/Microsoft/WinAppDriver) is a service to support Selenium-like UI Test Automation on Windows Applications.
For some message like click, the return message is `{"sessionId":"19C1CE26-1887-4DA1-AD04-521307B5445C","status":0}` which doesn't include value. then WebDriverIO consider it as failed message and will retry.
This pullrequest try to consider value is optional when status=0 to workaround WinAppDriver problem.

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/packages/webdriver/src/utils.js
+++ b/packages/webdriver/src/utils.js
@@ -26,8 +26,7 @@ export function isSuccessfulResponse (statusCode, body) {
             /**
             * A workaround for WinAppDriver. For some messages, WinAppDriver doesn't return value when status is 0.
             */
-        }
-        else {
+        } else {
             log.debug('request failed due to missing body')
             return false
         }
--- a/packages/webdriver/src/utils.js
+++ b/packages/webdriver/src/utils.js
@@ -22,7 +22,7 @@ export function isSuccessfulResponse (statusCode, body) {
      * response contains a body
      */
     if (!body || typeof body.value === 'undefined') {
-        if (body.status && body.status === 0) {
+        if (body && body.status === 0) {
             /**
             * A workaround for WinAppDriver. For some messages, WinAppDriver doesn't return value when status is 0.
             */
--- a/packages/webdriver/src/utils.js
+++ b/packages/webdriver/src/utils.js
@@ -22,8 +22,15 @@ export function isSuccessfulResponse (statusCode, body) {
      * response contains a body
      */
     if (!body || typeof body.value === 'undefined') {
-        log.debug('request failed due to missing body')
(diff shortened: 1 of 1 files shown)


packages/webdriver/src/utils.js
```diff
@@ -22,8 +22,14 @@ export function isSuccessfulResponse (statusCode, body) {
      * response contains a body
      */
     if (!body || typeof body.value === 'undefined') {
```
mgrybyk: I think overall condition can be improved to avoid nested conditions

licanhua: I thought about that, and my first local copy is in one line.
splitting it to nested condition would make easy to read. Otherwise I have to write test case on it to provide more context.

licanhua:
> Got some thoughts on it.
> Can it break somehow something else? When status is 0 for other drivers except of WinApp?

I have no knowledge about other drivers.

packages/webdriver/src/utils.js
```diff
@@ -22,8 +22,14 @@ export function isSuccessfulResponse (statusCode, body) {
...
-        log.debug('request failed due to missing body')
-        return false
+        if (body && body.status === 0) {
+            /**
+            * A workaround for WinAppDriver. For some messages, WinAppDriver doesn't return value when status is 0.
```
christian-bromann: This should be fixed in WinAppDriver then. WebdriverIO follows the W3C standard and so should the driver.

christian-bromann:
Will close this as the issue should rather be fixed in WinAppDriver to follow the protocol rather than here.

licanhua:
For people who have the same problem, you can workaround the problem by customized webdriver.
I always create pull request to WinAppDriver, but we don't have ETA for the final release date. Please check if [WinAppDriver](https://github.com/Microsoft/WinAppDriver/releases) there is a new release, this problem may be fixed. 

```
    "webdriverio": "^5.10.1",
    "webdriver": "git+https://github.com/react-native-windows/webdriver.git",
```

mihailDamchevski:
Hey guys! Waiting on this fix since January, any progress?

christian-bromann:
> For people who have the same problem, you can workaround the problem by customized webdriver.

I do __not__ recommend using this custom `webdriver` build as it us more than 3 years old and not compatible with any recent WebdriverIO setup.

PR #4391, opened 2019-08-21
e2e tests for DevTools package

christian-bromann:
To ensure that the DevTools driver is running according the WebDriver spec we should run some e2e test. Given that having an e2e test for every single command with all edge cases would be a bit to much therefor I only cover the most important commands (ignore things like timeouts, getWindowRect etc).

Yes, this will add another test to the pipeline but since it is running in headless mode it runs pretty fast and only adds 11.361s to the overall test.


christian-bromann:
@webdriverio/project-committers this is ready for review

e2e/devtools.test.js
```diff
@@ -0,0 +1,385 @@
...
+    })
+
+    it('findElements with css selector', async () => {
+        const boxes = await browser.findElements('css selector', '.box')
+        expect(boxes).toHaveLength(5)
+    })
+
+    it('find nested elements with css selector', async () => {
+        const nested = await browser.findElement('css selector', '.nested')
+        const spans = await browser.findElementsFromElement(nested[ELEMENT_KEY], 'css selector', 'span')
```
mgrybyk: what do you think on adding `findElementsFromElement` for xpath strategy as well?

packages/devtools/src/commands/elementClick.js
```diff
@@ -29,11 +30,27 @@ export default async function elementClick ({ elementId }) {
...
+         */
+        page.once('framenavigated', () => {
+            clearTimeout(waitForPageLoadTimeout)
+            page.once('load', () => resolve(null))
+        })
+
         const dialogHandler = () => resolve()
         page.once('dialog', dialogHandler)
         return elementHandle.click().then(() => {
             page.removeListener('dialog', dialogHandler)
```
mgrybyk: should we remove `framenavigated` as well?

packages/devtools/src/commands/findElement.js
```diff
@@ -8,9 +8,17 @@ export default async function findElement ({ using, value }) {
         throw new Error(`selector strategy "${using}" is not yet supported`)
     }
 
-    const page = this.windows.get(this.currentWindowHandle)
+    const page = this.getPageHandle()
     let needsCleanUp = false
 
+    if (using === 'link text') {
+        using = 'xpath'
+        value = `.//a[normalize-space() = "${value}"]`
```
mgrybyk: why do we need `.` here? There should be no nested element when using findElement(s), I think.

e2e/devtools.test.js
```diff
@@ -0,0 +1,385 @@
+import DevTools from '../packages/devtools/src/index'
```
mgrybyk: what do you think on splitting this files into several ones to run tests in parallel?

christian-bromann: I wasn't sure how Jest is doing parallelization but it seems to work 👍 split test suite into 2 files

christian-bromann:
@mgrybyk addressed comments

christian-bromann:
any further comments?

packages/devtools/src/commands/elementClick.js
```diff
@@ -35,15 +35,17 @@ export default async function elementClick ({ elementId }) {
         /**
          * check if page load has happened due to click
          */
-        page.once('framenavigated', () => {
+        const frameNavigatedHandler = () => {
             clearTimeout(waitForPageLoadTimeout)
```
mgrybyk: @christian-bromann  should `dialog` event be removed here as well?

christian-bromann: Ahh, I completely forgot why I had implemented it that way. I added comments to better explain how this is suppose to work.

packages/devtools/src/commands/elementClick.js
```diff
@@ -29,11 +30,43 @@ export default async function elementClick ({ elementId }) {
...
         page.once('dialog', dialogHandler)
         return elementHandle.click().then(() => {
+            /**
+             * no modals popped up, so clean up the listener
+             */
             page.removeListener('dialog', dialogHandler)
-            resolve(null)
+
+            /**
+             * wait for at least 150ms to see if a page load was triggered,
```
mgrybyk: does it mean that we'll wait for 150ms after every click?

christian-bromann: Yes .. I know it is not ideal but the only way we can guarantee we have stable page transitions

mgrybyk: I don't know how to handle it better neither, we may improve it later on

mgrybyk:
LGTM but please fix failing smoke tests

PR #4234, opened 2019-08-28
@wdio/sync: throw if command is executed outside of a fiber context

christian-bromann:
People fall into the trap of calling commands outside of `it` blocks or hooks of a test framework. Instead of returning a promise and confuse the user we should immediately throw an error with a helpful message to tell customers to only call commands within expected areas.


christian-bromann:
@mgrybyk I decided to only print a warning to allow handling commands using promises. Not sure if it is visible enough though but we could think about how we can properly display such important warnings in the testrunner.

PR #4424, opened 2019-08-29
support direct connect functionality for load balanced session redirection

jlipps:
## Proposed changes

Bringing the functionality of https://github.com/appium/python-client#direct-connect-urls to wdio.

## Types of changes

- [ ] Documentation update

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works - adding an e2e test would be difficult since we'd have to rely on a cloud provider. I did test this with HeadSpin, which supports this cap, and it worked as expected.

### DIFF
--- a/packages/webdriver/src/utils.js
+++ b/packages/webdriver/src/utils.js
@@ -415,3 +415,25 @@ export function getEnvironmentVars({ isW3C, isMobile, isIOS, isAndroid, isChrome
         isSeleniumStandalone: { value: isSeleniumStandalone }
     }
 }
+
+/**
+ * Decorate the params object with host updates based on the presence of
+ * directConnect capabilities in the new session response. Note that this
+ * mutates the object.
+ * @param  {Object} params    post-new-session params used to build driver
+ */
+export function setupDirectConnect(params) {
+    log.info(JSON.stringify(params))
+    const { directConnectProtocol, directConnectHost, directConnectPort,
+        directConnectPath } = params.capabilities
+    if (directConnectProtocol && directConnectHost && directConnectPort &&
+        directConnectPath) {
+        log.info('Found direct connect information in new session response. ' +
+                 `Will connect to server at ${directConnectProtocol}://` +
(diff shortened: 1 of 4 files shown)


jlipps:
By the way, I was unable to commit and push this change with git hook verification enabled because of many lint errors unrelated to my change.

jlipps:
yes, can you point me to the right place for that?

christian-bromann:
Give the location of the code you should add a test for it here: https://github.com/webdriverio/webdriverio/blob/master/packages/webdriver/tests/utils.test.js

packages/webdriver/src/utils.js
```diff
@@ -415,3 +415,25 @@ export function getEnvironmentVars({ isW3C, isMobile, isIOS, isAndroid, isChrome
...
 }
+
+/**
+ * Decorate the params object with host updates based on the presence of
+ * directConnect capabilities in the new session response. Note that this
+ * mutates the object.
+ * @param  {Object} params    post-new-session params used to build driver
+ */
+export function setupDirectConnect(params) {
+    log.info(JSON.stringify(params))
```
christian-bromann: Do we need this info, if we have another one a couple lines down?

packages/webdriver/src/utils.js
```diff
@@ -415,3 +415,25 @@ export function getEnvironmentVars({ isW3C, isMobile, isIOS, isAndroid, isChrome
...
+    const { directConnectProtocol, directConnectHost, directConnectPort,
+        directConnectPath } = params.capabilities
+    if (directConnectProtocol && directConnectHost && directConnectPort &&
```
mgrybyk: can one of these, ex `directConnectPath`, be an empty string?

packages/webdriver/src/utils.js
jlipps: nope, this was for debugging. will remove

jlipps:
ok, unit tests added.

packages/webdriver/src/utils.js
jlipps: good point, empty string is probably a valid case.

PR #4429, opened 2019-08-30
Only print warning if user is running wdio testrunner

christian-bromann:
A small addition to PR #4234


christian-bromann:
moving ahead as this is only a tiny modification to an already approved PR

Issue #4430, opened 2019-08-31
sync mode broken inside multiremote customCommands


js-kyle:
**Environment (please complete the following information):**
 - **WebdriverIO version:** 5.12.5
 - **Mode:** WDIO Testrunner
 - **If WDIO Testrunner, running sync/async:** sync
 - **Node.js version:** 10.14.1
 - **NPM version:** 6.4.1
 - **Browser name and version:** Chrome 76
 - **Platform name and version:** MacOS 10.14.4
 - **Additional wdio packages used (if applicable):** 
```
    "@wdio/cli": "^5.12.4",
    "@wdio/devtools-service": "^5.12.3",
    "@wdio/local-runner": "^5.12.4",
    "@wdio/mocha-framework": "^5.12.1",
    "@wdio/spec-reporter": "^5.12.1",
    "@wdio/sync": "^5.12.3",
    "wdio-chromedriver-service": "^5.0.2",
    "webdriverio": "^5.12.5"
```
**Config of WebdriverIO**
```
exports.config = {
  port: 9515 ,
  path: '/',
  sync: true,
  maxInstances: 1,
  services: [
    'chromedriver',
  ],
  capabilities: {
    browserA: {
      capabilities: {
        browserName: 'chrome'
      }
    },
    browserB: {
      capabilities: {
        browserName: 'chrome'
      }
    }
  }
}
```

**Describe the bug**
It appears that when using customCommands in multiremote mode, that the sync feature does not work correctly.

**To Reproduce**

These commands work, and log the HTML as expected:
```
  it('working demo', function() {
    browserA.url('https://google.com');
    let el = browserA.$('body');
    console.log(el.getHTML()); // this logs HTML
  });
```
**The same commands as above, do not resolve the promises when inside a customCommand:**
```
  it('broken demo', function() {

    browserA.addCommand('testCommand', function() {
      browserA.url('https://google.com');
      let el = browserA.$('body');
      console.log(el); This logs Promise { <pending> }
      console.log(el.getHTML()); TypeError: el.getHTML is not a function
    });
    browserA.testCommand();
  });
```

**Expected behavior**
Commands should return the value, rather than a pending Promise

**Additional context**
The problem is not seen when running with a single browser/capability. The issue can only be reproduced inside customCommands.

If the customCommand is changed to a async function which uses `await`, the values are resolved, which leads me to believe it is an issue to do with @wdio/sync.

[...]

Issue #4133, opened 2019-07-01
Sync mode work slow in some situations when defined hooks beforeCommand and afterCommand


frg-tarasov:
**Environment:**
 - **@wdio/cli: 5.10.9**
 - **@wdio/local-runner: 5.10.9**
 - **@wdio/mocha-framework: 5.10.8**
 - **@wdio/selenium-standalone-service: 5.0.2**
 - **@wdio/spec-reporter: 5.9.3**
 - **Node.js version: 10.16.0**
 - **NPM version: 6.9.0**
 - **Browser name and version: Chromium 75.0.3770.100 Linux**
 - **Platform name and version: Linux archlinux 5.1.15-arch1-1-ARCH**

**Config of WebdriverIO and etc**
I prepared repository with full example for reproduce problem
https://github.com/frg-tarasov/webdriverio-sync-slow-example

**Describe the bug**
Sync mode work slow in some situations when defined hooks beforeCommand and afterCommand. Query element by selector may work very slow if elements many.

**Expected behavior**
Time of query element should not dependency hooks and run mode tests.

BorisOsipov:
@frg-tarasov can you set  `logLevel: 'trace'` in wdio.conf.js and add updated logs?

frg-tarasov:
I'm added logs.

christian-bromann:
> I'm added logs.

where?

frg-tarasov:
In repository. https://github.com/frg-tarasov/webdriverio-sync-slow-example/tree/master/slow/logs
As soon as asked added.

christian-bromann:
Can you provide logs using `sync` and without?

BorisOsipov:
Logs are in repo with sample project

frg-tarasov:
Normal folder is without sync mode, slow folder with sync mode. Each folder contains logs.

mgrybyk:
@frg-tarasov first of all there are some problems with your example, not all promises are properly resolved.
Secondly there are no big difference between sync mode and normal mode.

For sure sync mode is a bit slower because we need to wrap functions and / or maybe because of Fibers.

I'm posting my results:

![image](https://user-images.githubusercontent.com/25589559/61234457-67505780-a733-11e9-899c-51b2d166d4aa.png)

![image](https://user-images.githubusercontent.com/25589559/61234482-7505dd00-a733-11e9-80bd-eb375e77261e.png)

christian-bromann:
[...]

PR #4438, opened 2019-09-02
Fix custom command and command overwrites in multiremote

christian-bromann:
## Proposed changes

This patch fixes issues that got reported in #4430.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/packages/wdio-sync/src/wrapCommand.js
+++ b/packages/wdio-sync/src/wrapCommand.js
@@ -49,6 +49,8 @@ export default function wrapCommand (commandName, fn) {
             inFiber(this)
             return futureResult
         } catch (e) {
+            inFiber(this)
+
             /**
              * in case some 3rd party lib rejects without bundling into an error
              */
@@ -64,7 +66,6 @@ export default function wrapCommand (commandName, fn) {
                 return result
             }
 
-            inFiber(this)
             throw e
         }
     }
--- a/packages/wdio-sync/src/wrapCommand.js
+++ b/packages/wdio-sync/src/wrapCommand.js
@@ -49,8 +49,6 @@ export default function wrapCommand (commandName, fn) {
             inFiber(this)
             return futureResult
         } catch (e) {
-            inFiber(this)
-
             /**
              * in case some 3rd party lib rejects without bundling into an error
              */
@@ -66,6 +64,7 @@ export default function wrapCommand (commandName, fn) {
                 return result
             }
 
+            inFiber(this)
             throw e
         }
     }
@@ -119,6 +118,17 @@ function isNotInFiber (context, fnName) {
  * @param {object} context browser or element
  */
 function inFiber(context) {
+    if (context.constructor.name === 'MultiRemoteDriver') {
+        return context.instances.forEach(instance => {
+            context[instance]._NOT_FIBER = false
+            let parent = context[instance].parent
+            while (parent && parent._NOT_FIBER) {
+                parent._NOT_FIBER = false
+                parent = parent.parent
+            }
+        })
+    }
+
     context._NOT_FIBER = false
     let parent = context.parent
     while (parent && parent._NOT_FIBER) {
--- a/tests/multiremote/test.js
+++ b/tests/multiremote/test.js
@@ -32,6 +32,7 @@ describe('smoke test multiremote', () => {
                 return `Title: ${title}`
             })
             assert.strictEqual(global.browserA.foobar(), 'Title: Mock Page Title')
+            assert.equal(typeof global.browserB.foobbar, 'undefined')
         })
 
         it('should throw if promise rejects', () => {
--- a/tests/multiremote/test.js
+++ b/tests/multiremote/test.js
@@ -32,7 +32,7 @@ describe('smoke test multiremote', () => {
                 return `Title: ${title}`
             })
             assert.strictEqual(global.browserA.foobar(), 'Title: Mock Page Title')
-            assert.equal(typeof global.browserB.foobbar, 'undefined')
+            assert.equal(typeof global.browserB.foobar, 'undefined')
         })
 
         it('should throw if promise rejects', () => {
--- a/tests/multiremote/test.js
+++ b/tests/multiremote/test.js
@@ -5,6 +5,8 @@ describe('smoke test multiremote', () => {
         assert.equal(
             JSON.stringify(browser.getTitle()),
             JSON.stringify(['Mock Page Title', 'Mock Page Title']))
+        assert.equal(browser.browserB.getTitle(), 'Mock Page Title')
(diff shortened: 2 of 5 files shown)


SubJunk:
Thanks for the quick work!

packages/wdio-sync/src/wrapCommand.js
```diff
@@ -49,6 +49,8 @@ export default function wrapCommand (commandName, fn) {
             inFiber(this)
             return futureResult
         } catch (e) {
+            inFiber(this)
```
mgrybyk: Do you know why should it be moved for multiremote custom commands?

moving `inFiber` before returning a result affects performance and might cause a regression of https://github.com/webdriverio/webdriverio/issues/4133

christian-bromann: @mgrybyk I just ran the example from the issue and had similar results as to what is expected from the fast example. So I think this change didn't affect the performance of it.

christian-bromann: This change is responsible for allowing user to run commands from single instances in multiremote synchronously. Before this would return only a promise.

mgrybyk: Got it, I want to verify it locally today

mgrybyk: Ok, I've verified it locally and have to say that it affects performance.

Before the change I have 15 beforeCommand calls and after the change 44.

My test:
```
describe("main suite", () => {
    beforeEach(() => {
        browser.url("http://guinea-pig.webdriver.io/")
    })

    it("remove before wait for not exist", () => {
        const el = $$('body')[0].$('aside').$('span')

        browser.execute(function (el) { el.remove() }, el)
        el.waitForNotDisplayed(2000)
    })
})
```

tests/multiremote/test.js
```diff
@@ -23,6 +25,16 @@ describe('smoke test multiremote', () => {
...
         })
 
+        it('should respect promises if command was added to single browser', () => {
+            browser.customCommandScenario(Object.keys(browser.instances).length)
+            global.browserA.addCommand('foobar', () => {
+                const title = global.browserA.getTitle()
+                return `Title: ${title}`
+            })
+            assert.strictEqual(global.browserA.foobar(), 'Title: Mock Page Title')
+            assert.equal(typeof global.browserB.foobbar, 'undefined')
```
mgrybyk: is it a typo in `foobbar`? two `b` letters

mgrybyk: is it a typo in `foobbar`? two `b` letters

christian-bromann: typo, good catch 👍

packages/wdio-sync/src/wrapCommand.js
mgrybyk: I'm not able to reproduce defect locally after reverting back position of `inFiber(this)`.

However smoke tests failing for me in such case.

Here is my test that is passing in my working project:
```
   it('mr', () => {
        // @ts-ignore
        assert.equal(browser.browserB.getTitle(), 'WebdriverJS Testpage')
        // @ts-ignore
        assert.equal(browser.browserA.getTitle(), 'WebdriverJS Testpage')
        // @ts-ignore
        // const browserA = browser
        browserA.addCommand('testCommand', function () {
            // @ts-ignore
            browserA.url('http://guinea-pig.webdriver.io/');
            // @ts-ignore
            let el = browserA.$('aside');
            // console.log(el);
            console.log(el.getHTML());
        });
        // @ts-ignore
        browserA.testCommand();

        // @ts-ignore
        global.browserA.addCommand('foobar', () => {
            // @ts-ignore
            const title = global.browserA.getTitle()
            return `Title: ${title}`
        })
        // @ts-ignore
        assert.strictEqual(global.browserA.foobar(), 'Title: WebdriverJS Testpage')


        // @ts-ignore
        browserA.overwriteCommand('getTitle', function (origCommand: any) {
            return `Title: ${origCommand()}`
        })
        // @ts-ignore
        assert.equal(browserA.getTitle(), 'Title: WebdriverJS Testpage')

    })
```

mgrybyk: I think it happens because protocol commands are wrapped twice in Multiremote mode for some reason

mgrybyk: Managed to reproduce it locally with
```
    it('mr', () => {
        assert.equal(
            JSON.stringify(browser.getTitle()),
            JSON.stringify(['WebdriverJS Testpage', 'WebdriverJS Testpage']))
        // @ts-ignore
        assert.equal(browser.browserB.getTitle(), 'WebdriverJS Testpage')
        // @ts-ignore
        assert.equal(browser.browserA.getTitle(), 'WebdriverJS Testpage')
    })
```

mgrybyk: @christian-bromann I've figure out what is going on. We have to set `_NOT_FIBER` for every instance

mgrybyk: So it is required to revert the change and add something like this to `inFiber` function:
```
    if (Array.isArray(context.instances)) {
        context.instances.forEach(instance => {
            context[instance]._NOT_FIBER = false
        })
    }
```

christian-bromann:
@mgrybyk I applied the fix now at the correct position, can you try again?

mgrybyk:
I think you still have to do this https://github.com/webdriverio/webdriverio/pull/4438#discussion_r320177401

mgrybyk:
Ah, missed your change, checking

packages/wdio-sync/src/wrapCommand.js
```diff
@@ -118,6 +118,17 @@ function isNotInFiber (context, fnName) {
  * @param {object} context browser or element
  */
 function inFiber(context) {
+    if (context.constructor.name === 'MultiRemoteDriver') {
```
mgrybyk: I think this one might not work if we are in element context.
Ex: 
browser // MultiRemoteDriver
browser.$('foo') // ?

packages/wdio-sync/src/wrapCommand.js
```diff
@@ -118,6 +118,17 @@ function isNotInFiber (context, fnName) {
...
+        return context.instances.forEach(instance => {
```
mgrybyk: Shouldn't we set `_NOT_FIBER` for both context and context instances?

mgrybyk:
Added two comments, have to think on how to verify that...

packages/wdio-sync/src/wrapCommand.js
mgrybyk: now I know that it can be also `Element` and `Browser`

PR #4434, opened 2019-09-02
Unable to get element parent by xpath

christian-bromann:
 - **WebdriverIO version:** 5.12.5 
 - **Mode:** WDIO Testrunner
 - **Node.js version:** 10
 - **Browser name and version:** Chrome 76

**Config of WebdriverIO**
Add `devtools` package from master (pre-release version).
Then enable devtools: `automationProtocol: 'devtools'`

**Describe the bug**
Unable to get element parent by xpath. Have to use `$(function () { return this.elementParent })` to workaround it.

**To Reproduce**
```
describe("main suite", () => {
    it("some test", () => {
        browser.url("http://guinea-pig.webdriver.io/")
        $("#purplebox").$("..")
    })
})
```

**Expected behavior**
- should be possible to get parent same as without `automationProtocol: 'devtools'`
- please add smoke test

**Log**
```
Error: failed to find element matching selector "*"
at DevToolsDriver.findElementFromElement (/devtools/build/commands/findElementFromElement.js:46:46)
```

## Solution

- replaced workaround of setting data attribute with usage of `$x` command


mgrybyk:
Will you be able to increase test coverage?

christian-bromann:
@mgrybyk done

christian-bromann:
I will move ahead even though codecov is complaining but I removed bunch of code which makes it hard to stay within the threshold

Issue #4448, opened 2019-09-03
isDisplayedInViewport fails if element doesn't exist


mgrybyk:
**Environment (please complete the following information):**
 - **WebdriverIO version:** 5.12.5 
 - **Mode:** WDIO Testrunner
 - **If WDIO Testrunner, running sync/async:** sync
 - **Node.js version:** 10
 - **Browser name and version:** Chrome 76

**Config of WebdriverIO**
Add `devtools` package from master (pre-release version).
Then enable devtools: `automationProtocol: 'devtools'`

**Describe the bug**
isDisplayedInViewport fails if element doesn't exist.

**To Reproduce**
```
describe("main suite", () => {
    it("some test", () => {
        browser.url("http://guinea-pig.webdriver.io")
        $('foobar').isDisplayedInViewport()
    })
})
```

**Expected behavior**
should work same as without `automationProtocol: 'devtools'`.

**Log**
```
Error: Malformed type for "elementId" parameter of command isElementDisplayed
Expected: string
Actual: undefined
For more info see https://w3c.github.io/webdriver/#element-displayedness
    at validate (/e2e/node_modules/devtools/build/utils.js:52:13)
    at Element.<anonymous> (/e2e/node_modules/devtools/build/devtoolsdriver.js:79:43)
```

christian-bromann:
Fixed by #4459

Issue #4447, opened 2019-09-03
isDisplayed is not working as expected in headless mode


mgrybyk:
**Environment (please complete the following information):**
 - **WebdriverIO version:** 5.12.5 
 - **Mode:** WDIO Testrunner
 - **If WDIO Testrunner, running sync/async:** sync
 - **Node.js version:** 10
 - **Browser name and version:** Chrome 76

**Config of WebdriverIO**
Add `devtools` package from master (pre-release version).
Then enable devtools: `automationProtocol: 'devtools'`

**Describe the bug**
isDisplayed is not working as expected in headless mode if element appeared outside of viewport.

**To Reproduce**
```
describe("main suite", () => {
    it("some test", () => {
        browser.url("http://guinea-pig.webdriver.io")

        browser.execute(function () {
            this.setTimeout(() => {
                const mySpan = this.document.createElement("span")
                mySpan.id = 'foobarTest'
                mySpan.innerText = 'displayed!'
                this.document.querySelector('body').appendChild(mySpan)
            }, 2000);
        })

        $('#foobarTest').waitForDisplayed(4000)
    })
})
```

**Expected behavior**
should work same as without `automationProtocol: 'devtools'` or same as without headless mode.

**Log**
```
2019-09-03T14:38:40.060Z INFO devtools: RESULT Error: Element with selector "#foobarTest" not found
    at DevToolsDriver.findElement (/Users/mygr/sources/e2e/node_modules/devtools/build/utils.js:99:12)
    at process._tickCallback (internal/process/next_tick.js:68:7)
2019-09-03T14:38:40.858Z INFO devtools: COMMAND findElement("css selector", "#foobarTest")
2019-09-03T14:38:40.861Z INFO devtools: RESULT Error: Element with selector "#foobarTest" not found
    at DevToolsDriver.findElement (/Users/mygr/sources/e2e/node_modules/devtools/build/utils.js:99:12)
    at process._tickCallback (internal/process/next_tick.js:68:7)
2019-09-03T14:38:41.657Z INFO devtools: COMMAND findElement("css selector", "#foobarTest")
2019-09-03T14:38:41.660Z INFO devtools: RESULT { 'element-6066-11e4-a52e-4f735466cecf': 'ELEMENT-1' }
2019-09-03T14:38:41.664Z INFO devtools: COMMAND isElementDisplayed("ELEMENT-1")
2019-09-03T14:38:41.668Z INFO devtools: RESULT false
[...]

Issue #4443, opened 2019-09-03
Navigation fails with "Execution context was destroyed"


mgrybyk:
**Environment (please complete the following information):**
 - **WebdriverIO version:** 5.12.5 
 - **Mode:** WDIO Testrunner
 - **If WDIO Testrunner, running sync/async:** sync
 - **Node.js version:** 10
 - **Browser name and version:** Chrome 76

**Config of WebdriverIO**
Add `devtools` package from master (pre-release version).
Then enable devtools: `automationProtocol: 'devtools'`

**Describe the bug**
Test fails after clicking on link

**To Reproduce**
```
describe("main suite", () => {
    beforeEach(() => {
        browser.url("https://webdriver.io/")
        console.log(browser.getTitle())
    })

    it("foobar test", () => {
        $(".nav-site a[href*='api']").click()
        console.log(browser.getTitle())

        $(".post a[href*='CONTRIBUTING']").click()
        console.log(browser.getTitle())
    })
})
```

**Expected behavior**
should work same as without `automationProtocol: 'devtools'`

**Log**
```
Execution context was destroyed, most likely because of a navigation.
    at Frame.<anonymous> (/e2e/node_modules/puppeteer-core/lib/helper.js:111:15)
    at Page.title (/e2e/node_modules/puppeteer-core/lib/Page.js:995:29)
    at Page.<anonymous> (/e2e/node_modules/puppeteer-core/lib/helper.js:112:23)
    at DevToolsDriver.getTitle (/e2e/node_modules/devtools/build/commands/getTitle.js:10:15)
    at Browser.<anonymous> (/e2e/node_modules/devtools/build/devtoolsdriver.js:83:47)
```

christian-bromann:
Fixed by #4459

PR #4459, opened 2019-09-05
Bugfixes for DevTools package

christian-bromann:
## Proposed changes

This patch has fixes for #4443, #4447 and #4448.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

Not quite done yet, have to check how I can incorporate tests for these fixes.


packages/devtools/src/commands/isElementDisplayed.js
```diff
@@ -1,11 +1,24 @@
...
     const elementHandle = this.elementStore.get(elementId)
 
     if (!elementHandle) {
         throw getStaleElementError(elementId)
     }
 
+    /**
+     * scroll element into view
+     */
+    await executeScript.call(this, {
```
mgrybyk: I don't really like the idea to use scrollIntoView in isDisplayed (I'll provide an example why in next comment).
Are there any alternatives to it?
I assumed that isDisplayed should work even if element is not in current viewport.

mgrybyk: Open https://go.moplay.com/de/sports/registration-pop-up/?Offer=Bet10Get30GenSports
wait for site to load and form to open, then do in browser
`$$('[data-step="1"] .stepwise-registration__ssl')[0].scrollIntoView()`
That's it, form is broken :(

I use either `scrollIntoViewIfNeeded(false)` or `scrollIntoView(false)` here.
Maybe we can at least change `elem.scrollIntoView()` to `elem.scrollIntoViewIfNeeded()` ?

christian-bromann: So for the protocol command it doesn't matter if the element is within the viewport or not for determine the visibility. Since Puppeteer only allows to check if an element is in the viewport, I have to do this action before calling the isVisible command.

christian-bromann:
So I think I added necessary unit tests, this is ready for review.

packages/devtools/src/commands/isElementDisplayed.js
```diff
@@ -1,11 +1,24 @@
+import executeScript from './executeScript'
+
 import { getStaleElementError } from '../utils'
+import { ELEMENT_KEY } from '../constants'
+
+const SCROLL_INTO_VIEW_SCRIPT = 'return (function scrollIntoView (elem) { elem.scrollIntoView() }).apply(null, arguments)'
```
mgrybyk: What do you think on changing scrollIntoView to scrollIntoViewIfNeeded?

mgrybyk: However scrollIntoViewIfNeeded command is non standard.
Ok then, can we get element visibility, if it is not visible then scroll to it and try again?

So my complain is that I won't be able to use `isDisplayed` command anymore with current implementation :(

packages/webdriverio/src/commands/element/isDisplayedInViewport.js
```diff
@@ -45,7 +45,7 @@ export default function isDisplayedInViewport () {
      * if devtools package is used we can use Puppeteers ability to
      * check visibility within the viewport
      */
-    if (this.isDevTools) {
+    if (this.isDevTools && this.elementId) {
```
mgrybyk: Can you add unit test for this one?

packages/devtools/src/commands/isElementDisplayed.js
christian-bromann: > So my complain is that I won't be able to use isDisplayed command anymore with current implementation :(

Why is that?

mgrybyk: https://github.com/webdriverio/webdriverio/pull/4459/files#r321205421

> Open https://go.moplay.com/de/sports/registration-pop-up/?Offer=Bet10Get30GenSports
wait for site to load and form to open, then do in browser
$$('[data-step="1"] .stepwise-registration__ssl')[0].scrollIntoView()
That's it, form is broken :(

So I'm not able to use isDisplayed in forms like that

christian-bromann:
@mgrybyk I removed the `isDisplayed` capability from the DevTools package. I think it is difficult to define what visible means which is why neither Puppeteer nor WebDriver has command for it. That said, in order to have a single source of truth let's have them both execute the JS code we have to determine visibility for both devtools and webdriver

e2e/protocol.test.js
```diff
@@ -269,6 +269,22 @@ describe('executeScript', () => {
     })
 })
 
+it('can handle navigation and command execution', async () => {
+    await browser.navigateTo('http://guinea-pig.webdriver.io')
+    expect(typeof (await browser.getTitle())).toContain('string')
```
mgrybyk: why it is not `'WebdriverJS Testpage'` ?

e2e/protocol.test.js
```diff
@@ -269,6 +269,22 @@ describe('executeScript', () => {
...
+
+    const link = await browser.findElement('link text', 'two')
+    await browser.elementClick(link[ELEMENT_KEY])
+    expect(typeof (await browser.getTitle())).toContain('string')
+
+    await browser.back()
+    expect(typeof (await browser.getTitle())).toContain('string')
```
mgrybyk: aren't we back to `'WebdriverJS Testpage'` page?

e2e/protocol.test.js
```diff
@@ -269,6 +269,22 @@ describe('executeScript', () => {
...
+
+    const githubLink = await browser.findElement('link text', 'GitHub Repo')
+    await browser.elementClick(githubLink[ELEMENT_KEY])
```
mgrybyk: is it possible to verify that title is `GitHub`?

christian-bromann:
> I think we should be more strict when verifying page titles

No because this is not part of the protocol, technically if you click on a link in WebDriver and get the title, chances are that you still get the old title. Therefor it is just important for me here, that we get a result

e2e/protocol.test.js
christian-bromann: see comment below

e2e/protocol.test.js
christian-bromann: see comment below

e2e/protocol.test.js
christian-bromann: see comment below

christian-bromann:
This will also fix the flaky tests on master

mgrybyk:
Why then with Webdriver the test always passing (waiting for page to load?) when click triggers navigation while with devtools protocol it always fail?

christian-bromann:
> when click triggers navigation while with devtools protocol it always fail?

Do you have a reproducible example?

mgrybyk:
Investigating it

mgrybyk:
@christian-bromann 
I know it's `getUrl` instead of `getTitle`, is it same issue?
```
const assert = require('assert')

describe("remote driver", () => {
    it("foobar", () => {

        $("#githubRepo").click()
        const url = browser.getUrl()
        assert.strictEqual(url, 'https://github.com/')
    })
})
```

christian-bromann:
I increased the timeout for frame navigation and updated the e2e test to actually check the title. What do you think?

mgrybyk:
Taking a look. 
In a meanwhile isDisplayed also have to be fixed

mgrybyk:
@christian-bromann to be honest I don't like it, now every click takes one second, also not only click can lead to site loading, ex: it can be any javascript call.

What if we fully remove everything related to site load from click and implement something like webdriver's page load strategy?

christian-bromann:
@mgrybyk I agree with you, I wasn't really happy with it either. I check the Puppeteer API and found that they have an [`ExecutionContext`](https://pptr.dev/#?product=Puppeteer&version=v1.19.0&show=api-class-executioncontext) class and it can be used to check before every command if and execution context exists. Unfortunately that doesn't help completely to avoid issues especially when a transition happens randomly which is why I kept the fallback where we end up re-executing the command. What do you think?

mgrybyk:
It is better now, thank you! However I still have to wait for page to load manually and I think it can be improved.

Here is what I've figured out while debugging:
```
// add listeners for debugging in devtoolsdriver.js
// page.on('load', () => console.log('EVENT: page load'));
// page.on('domcontentloaded', () => console.log('EVENT: page domcontentloaded'));
// page.on('framenavigated', () => console.log('EVENT: page framenavigated'));

// open some long loading link (with click or javascript)
browser.execute(function () {
                this.document.location = 'https://secure.gamblingcommission.gov.uk/PublicRegister/Search/Detail/50970'
})

browser.waitUntil(() => {
    const readyState: string = browser.execute('return document.readyState')
    console.log('readyState', readyState)
    return ['complete', 'interactive'].includes(readyState)
}, 15000, 'oh snap', 300)

browser.pause(2000)
console.log('readyState', browser.execute('return document.readyState'))
```

Here is an output:
```
[0-0] EVENT: page framenavigated
[0-0] readyState loading // have to wait manually
[0-0] readyState loading
[0-0] readyState loading
[0-0] EVENT: page domcontentloaded // same as readyState interactive?
[0-0] readyState interactive  // even though site has loaded some JS is executing
[0-0] EVENT: page load // same as readyState complete?
[0-0] readyState complete // now I can do everything
```

What I wanted to say, do you think it makes sense to wait for `document.readyState` to be interactive at least? 
We may allow user to configure page load strategy in similar way to webdriver' drivers (eager, normal, none) https://stackoverflow.com/a/43737358/2475987

It may be done with another PR later on but I think it's quite useful.

christian-bromann:
Good call @mgrybyk, added this to the PR, please have a look

mgrybyk:
Hey, I'll review later today.

mgrybyk:
@christian-bromann I'm getting `page.mainFrame is not a function`
for line `const executionContext = await page.mainFrame().executionContext()`

Looking for steps to reproduce

mgrybyk:
@christian-bromann 
```
it("foobar", () => {
    browser.url("https://fiddle.jshell.net/westonruter/6mSuK/show/")
    $("#result iframe").waitForDisplayed()
    browser.pause(1000)
    assert.strictEqual(($("iframe").getAttribute("src")).includes("fiddle"), true)
    browser.switchToFrame($("#result iframe"))
})
```

Execution fails, browser stays opened.
Log:
```
ERROR @wdio/local-runner: Failed launching test session: TypeError: page.mainFrame is not a function
    at DevToolsDriver.checkPendingNavigations (/e2e/node_modules/devtools/build/devtoolsdriver.js:152:41)
    at Browser.wrappedCommand (/e2e/node_modules/devtools/build/devtoolsdriver.js:79:18)
    at Runner.endSession (/e2e/node_modules/@wdio/runner/build/index.js:241:26)
    at Runner.run (/e2e/node_modules/@wdio/runner/build/index.js:135:18)
```

christian-bromann:
@mgrybyk this should be fixed with #4444 which is why I rebased the branch

christian-bromann:
nevermind, I see the error in the build .. taking a look

christian-bromann:
Added a fix for it

mgrybyk:
I don't have any other comments for this PR except of adding e2e tests for switching to iframe that can be done separately later on.
Verified cases from #4443, #4447, #4448 

Thank you!

PR #4462, opened 2019-09-05
Print not implemented command

mgrybyk:
## Proposed changes

Print not implemented command name for better visibility

## Types of changes

- [x] Polish

## Checklist

- [x] Updated test

### DIFF
--- a/packages/devtools/src/devtoolsdriver.js
+++ b/packages/devtools/src/devtoolsdriver.js
@@ -59,7 +59,7 @@ export default class DevToolsDriver {
          * check if command is implemented
          */
         if (typeof this.commands[command] !== 'function') {
-            return () => { throw new Error('Not yet implemented') }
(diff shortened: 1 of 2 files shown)


packages/devtools/src/devtoolsdriver.js
```diff
@@ -59,7 +59,7 @@ export default class DevToolsDriver {
          * check if command is implemented
          */
         if (typeof this.commands[command] !== 'function') {
-            return () => { throw new Error('Not yet implemented') }
+            return () => { throw new Error(`Command "${command}" is not yet implemented`) }
```
CrispusDH: I think you have to use `has` instead of `is` in present Perfect

christian-bromann: I think it is fine

mgrybyk: What if it is present simple?

PR #4463, opened 2019-09-05
Retried specfiles are inserted at the start of the specs queue

bennieswart:
## Proposed changes

In our case we take care to list specfiles in decreasing order of expected runtime, to avoid the overall runtime being unnecessarily extended by a single long-running specfile at the end.
In such cases, whenever a specfile fails, it is beneficial to retry it as soon as possible by inserting it into the start of the queue instead of the end.

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/docs/ConfigurationFile.md
+++ b/docs/ConfigurationFile.md
@@ -170,6 +170,8 @@ exports.config = {
     //
     // The number of times to retry the entire specfile when it fails as a whole
     specFileRetries: 1,
+    // Whether or not retried specfiles should be retried immediately or deferred to the end of the queue
+    specFileRetriesDeferred: false,
     //
     // Test reporter for stdout.
     // The only one supported by default is 'dot'
--- a/docs/Options.md
+++ b/docs/Options.md
@@ -148,10 +148,16 @@ Type: `Number`<br>
 Default: `0` (don't bail; run all tests)
 
 ### `specFileRetries`
-The number of retry attempts for an entire specfile when it fails as a whole.
+The number of times to retry an entire specfile when it fails as a whole.
 
 Type: `Number`<br>
 Default: `0`
+
+### `specFileRetriesDeferred`
+Whether or not retried specfiles should be retried immediately or deferred to the end of the queue.
+
+Type: `Boolean`<br>
+Default: `true`
 
 ### `waitforTimeout`
 Default timeout for all `waitFor*` commands. (Note the lowercase `f` in the option name.) This timeout __only__ affects commands starting with `waitFor*` and their default wait time.
--- a/packages/wdio-cli/tests/launcher.test.js
+++ b/packages/wdio-cli/tests/launcher.test.js
@@ -169,6 +169,19 @@ describe('launcher', () => {
             launcher.endHandler({ cid: '0-5', exitCode: 1, retries: 1, specs: ['a.js'] })
             expect(launcher.schedule).toMatchObject([{ cid: 0, specs: [{ rid: '0-5', files: ['a.js'], retries: 0 }] }])
         })
+
+        it('should requeue retried specfiles at beginning of queue', () => {
+            launcher.configParser.getConfig = jest.fn().mockReturnValue({ specFileRetriesDeferred: false })
+            launcher.schedule = [{ cid: 0, specs: [{ files: ['b.js'] }] }]
+            launcher.endHandler({ cid: '0-5', exitCode: 1, retries: 1, specs: ['a.js'] })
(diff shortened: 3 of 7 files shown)


CrispusDH:
Hm, it can be additional point in config, because I can't tell what approach is right. In another hand, retry mechanism I always see only after run and as for me it's better way. You could easily see what was failed in a first run. If you have a lot of tests and retries will be immediately it would be total mess in output

christian-bromann:
> what do you think?

Yeah, sounds reasonable

bennieswart:
Good arguments both ways. I will add a setting and a test.

docs/ConfigurationFile.md
```diff
@@ -162,6 +162,8 @@ exports.config = {
     //
     // The number of times to retry the entire specfile when it fails as a whole
     specFileRetries: 1,
+    // Retried specfiles are inserted at the beginning of the queue.
+    specFileRequeue: 'first',
```
CrispusDH: I don't like string here, because it's obvious what should be here. You will go to the documentation each time. If we have only two states it could be boolean true or false. For example retriesRequeueAtBegginning: true or false by default.

docs/Options.md
```diff
@@ -141,11 +141,18 @@ Type: `Number`<br>
...
 
 ### specFileRetries
-The number of times to retry an entire specfile when it fails as a whole
+The number of times to retry an entire specfile when it fails as a whole.
 
 Type: `Number`<br>
 Default: `0`
 
+### specFileRequeue
+Where retried specfiles should be reinserted into the queue.
```
CrispusDH: It's still not clear what behavior will be. Could you explain a little bit more. You wrote `where` but it's unclear how it will work

packages/wdio-cli/tests/launcher.test.js
```diff
@@ -143,6 +143,17 @@ describe('launcher', () => {
             launcher.endHandler({ cid: '0-5', exitCode: 1, retries: 1, specs: ['a.js'] })
             expect(launcher.schedule).toMatchObject([{ cid: 0, specs: [{ rid: '0-5', files: ['a.js'], retries: 0 }] }])
         })
+
+        it('should requeue retried specfiles correctly', () => {
```
CrispusDH: "correctly" it's not clear definition. You could create two tests for covering two cases and write specific description

docs/ConfigurationFile.md
bennieswart: I agree - I cannot imagine any reason why retries would ever be inserted somewhere other than at the very beginning or very end. I did it this way mainly because I could not come up with a non-clumsy name for the boolean setting at the time.

The term "defer" is a nice boolean descriptor for now vs later - how about something like `specFileRetriesDeferred`? Then "true" means defer retries to the end of the queue and "false" means don't defer and retry now.

I also considered changing `specFileRequeue` into an object that could group all current and future retry-related settings, for example:
```js
specFileRetries: {
    count: 1,
    defer: true
}
```
We could add backwards compatibility by transforming `specFileRetries: 1` to `specFileRetries: { count: 1 }` internally.
What are your thoughts on this?

CrispusDH: 1) I like new name specFileRetriesDeferred
2) I don't like transform it to object, because it would affect to much users.
So, in my opinion would be enough to rename property and set boolean inside :)

bennieswart:
That was my initial thought as well. @CrispusDH, mind weighing in?

CrispusDH:
@christian-bromann I have added my points above and you [agree with me](https://github.com/webdriverio/webdriverio/pull/4463#issuecomment-528745302) :)

christian-bromann:
@CrispusDH

> I think it would increase mess in output and would be uncommon for our users.

Why do you see it creating a mess in the output?

IMO a spec retry can happen right after its failure. It makes more sense to me. I don't see why someone would want to configure it to run it after all tests.

CrispusDH:
Let's imagine you have 100 tests. After tests were run you want to check what tests were retried. If after first run you have list of failed tests and then second run with this scope it's quite easy to see all failed tests in a first run.
What if you retried tests immediately. How you will find this info?

christian-bromann:
@bennieswart what is the state of this PR?

bennieswart:
@christian-bromann I just rebased this onto latest master. If you're happy with the current state, you can merge.

njota08:
Is this actually working?
I'm setting **specFileRetriesDeferred** to false and my tests are being retried at the end. Same behavior if I set it as true.

PR #4460, opened 2019-09-08
Add unit tests for wrapCommand

christian-bromann:
I'd like to add tests for `wdio-sync/src/wrapCommand.js`

We missed to add unit tests with fix for multiremote custom commands https://github.com/webdriverio/webdriverio/pull/4438

### DIFF
--- a/packages/wdio-sync/tests/wrapCommand.test.js
+++ b/packages/wdio-sync/tests/wrapCommand.test.js
@@ -118,6 +118,44 @@ describe('wrapCommand:runCommand', () => {
         expect(context._hidden_changes_).toEqual([false, false])
     })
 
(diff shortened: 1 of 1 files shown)


mgrybyk:
Is it ok that coverage for that file is not 100%? I remember there is a place with string error that is not covered with tests, if so it's fine

christian-bromann:
> Is it ok that coverage for that file is not 100%?

Absolutely. This test had no full unit test coverage in the first place so every patch only needs to add unit tests for the execution path it is changing. 

If you are ok with this PR mind approving it?

mgrybyk:
Can't approve in mobile mode, LGTM, approving in comment mode

PR #4444, opened 2019-09-08
Execution fails when running tests in parallel

christian-bromann:
 - **WebdriverIO version:** 5.12.5 
 - **Mode:** WDIO Testrunner
 - **Node.js version:** 10
 - **Browser name and version:** Chrome 76

**Config of WebdriverIO**
Add `devtools` package from master (pre-release version).
Then enable devtools: `automationProtocol: 'devtools'`

**Describe the bug**
Some test(s) are failing when running multiple tests in parallel

**To Reproduce**
1. Create 4-8 files like this:
```
describe("main suite 1", () => {
    it("foobar test", () => {
    })
})
```
2. run tests in parallel, ex `maxInstances: 4`

NOTE: sometimes it fails even with `maxInstances: 2`, there more tests and instances we have the more tests are failing.

**Expected behavior**
should work same as without `automationProtocol: 'devtools'`

**Log**
```
TypeError: Cannot read property 'setDefaultTimeout' of undefined
    at DevToolsDriver.setTimeouts (/e2e/node_modules/devtools/build/devtoolsdriver.js:116:10)
    at new DevToolsDriver (/e2e/node_modules/devtools/build/devtoolsdriver.js:53:10)
    at Function.newSession (/e2e/node_modules/devtools/build/index.js:47:20)
```

## Solution

Wait for pages to be available and resolve `createSession` call after that. I am not sure if it is just technically not possible to run more than 4 headless Chrome instances at a time but with this simple patch it resolves the issue. I don't think that user will run a significant number of parallel Chrome sessions on a single machine due to CPU limitations.

### DIFF
--- a/packages/devtools/src/utils.js
+++ b/packages/devtools/src/utils.js
@@ -223,7 +223,7 @@ export async function getPages (browser, retryInterval = 100) {
         log.info('no browser pages found, retrying...')
 
         /**
-         * wait for 200ms to try again
+         * wait for some milliseconds to try again
          */
         await new Promise((resolve) => setTimeout(resolve, retryInterval))
         return getPages(browser)
--- a/packages/devtools/src/utils.js
+++ b/packages/devtools/src/utils.js
@@ -206,3 +206,28 @@ export function getStaleElementError (elementId) {
     error.name = 'stale element reference'
     return error
 }
+
+/**
+ * Helper function to get a list of Puppeteer pages from a Chrome browser.
+ * In case many headless browser are run in parallel there are situations
+ * where there are no pages because the machine is busy booting the headless
+ * browser.
+ *
+ * @param  {Puppeteer.Browser} browser  browser instance
+ * @return {Puppeteer.Page[]}           list of browser pages
+ */
+export async function getPages (browser, retryInterval = 100) {
+    const pages = await browser.pages()
+
+    if (pages.length === 0) {
+        log.info('no browser pages found, retrying...')
+
+        /**
+         * wait for 200ms to try again
+         */
+        await new Promise((resolve) => setTimeout(resolve, retryInterval))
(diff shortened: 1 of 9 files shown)


packages/devtools/src/utils.js
```diff
@@ -206,3 +206,28 @@ export function getStaleElementError (elementId) {
...
+ * @return {Puppeteer.Page[]}           list of browser pages
+ */
+export async function getPages (browser, retryInterval = 100) {
+    const pages = await browser.pages()
+
+    if (pages.length === 0) {
+        log.info('no browser pages found, retrying...')
+
+        /**
+         * wait for 200ms to try again
```
CrispusDH: Default retry interval is 100 not 200

packages/devtools/src/utils.js
```diff
@@ -206,3 +206,28 @@ export function getStaleElementError (elementId) {
...
+         */
+        await new Promise((resolve) => setTimeout(resolve, retryInterval))
+        return getPages(browser)
```
CrispusDH: Could it be endless recursion?

mgrybyk (commented):
👍

mgrybyk:
The main issue is fixed, now I'm able to run tests in parallel.

packages/devtools/src/utils.js
christian-bromann: I will fix the comment. I had 200ms first and then lowered it to 100ms.

packages/devtools/src/utils.js
christian-bromann: yes .. however it will be like the browser not booting up and at some point the user will just cancel the run. I don't think this will happen a lot of times but if someone is reporting it we can look into having a timeout here.

PR #4266, opened 2019-09-08
Convert protocol elements to WebdriverIO elements

christian-bromann:
There are some commands like `getActiveElement` that returns something like `element-6066-11e4-a52e-4f735466cecf:"ELEMENT-57"`

There is no way to interact with such object like with WebdriverIO element

**Goal**
I'd like to be able to wrap such responses, ideally automatically somehow

**Suggested solutions**
- add command WebdriverIO for every protocol command, like `getActiveElement` to wrap response with WebdriverIO element;
- add some command, like `browser.wrap(browser.getActiveElement())` to be able to convert protocol element objects to WebdriverIO elements
- automatically convert responses that have only `element-6066-11e4-a52e-4f735466cecf` in body to WebdriverIO elements.

### DIFF
--- a/packages/webdriverio/src/commands/browser/$.js
+++ b/packages/webdriverio/src/commands/browser/$.js
@@ -2,7 +2,9 @@
  * The `$` command is a short way to call the [`findElement`](/docs/api/webdriver.html#findelement) command in order
  * to fetch a single element on the page. It returns an object that with an extended prototype to call
  * action commands without passing in a selector. However if you still pass in a selector it will look
- * for that element first and call the action on that element.
+ * for that element first and call the action on that element. You can also pass in an object as selector
+ * where the object contains a property `element-6066-11e4-a52e-4f735466cecf` with the value of a reference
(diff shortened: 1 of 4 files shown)


christian-bromann:
@webdriverio/project-committers what do you all think about this solution?

CrispusDH:
I think this solution is good enough for such use case.

packages/webdriverio/src/commands/browser/$.js
```diff
@@ -32,18 +34,32 @@
         });
         console.log(text.$$('li')[2].$('a').getText()); // outputs: "API"
     });
+
+    it('should allow to convert protocol result of an element into a WebdriverIO element', () => {
+        const activeElement = browser.getActiveElement();
+        console.log($(activeElement).getTagName); // outputs active element
```
mgrybyk: .getTagName should be a function call in all occurrences

mgrybyk:
Not sure why but I can't neither approve nor request changes for PRs both from mobile and desktop, that's why leaving comments only

PR #4473, opened 2019-09-09
Fix devtools types

mgrybyk:
## Proposed changes

devtools types doesn't work in combination with webdriverio, ex: `browser.isDevTools` won't work

## Types of changes

## Checklist


CrispusDH:
Could we check types in the same way as webdriverio types?

mgrybyk:
@CrispusDH yes, I think we should do it one day.

PR #4476, opened 2019-09-10
Fix interaction with iframe with devtools protocol

mgrybyk:
## Proposed changes

There were some issues while in iframe context:
- click doesn't work
- page close doesn't work
- and some other issues

... because `frame` doesn't have lot's of `page` methods, see https://github.com/GoogleChrome/puppeteer/blob/v1.19.0/docs/api.md#class-frame

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

please note that I haven't added tests for every affected method to work in iframe.

### DIFF
--- a/packages/devtools/src/commands/refresh.js
+++ b/packages/devtools/src/commands/refresh.js
@@ -1,7 +1,7 @@
 export default async function refresh () {
     delete this.currentFrame
 
-    let page = this.getPageHandle()
+    const page = this.getPageHandle()
     await page.reload()
     return null
 }
--- a/packages/devtools/src/commands/refresh.js
+++ b/packages/devtools/src/commands/refresh.js
@@ -1,15 +1,7 @@
 export default async function refresh () {
+    delete this.currentFrame
+
     let page = this.getPageHandle()
-
-    /**
-     * if reload is not a function we are currently in a frame and need
-     * to move back to the page scope
-     */
-    if (typeof page.reload !== 'function') {
-        delete this.currentFrame
-        page = this.getPageHandle()
-    }
-
     await page.reload()
     return null
 }
--- a/packages/devtools/src/devtoolsdriver.js
+++ b/packages/devtools/src/devtoolsdriver.js
@@ -133,8 +133,8 @@ export default class DevToolsDriver {
         page.setDefaultTimeout(this.timeouts.get('pageLoad'))
     }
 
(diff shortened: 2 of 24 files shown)


packages/devtools/src/commands/refresh.js
```diff
@@ -1,15 +1,7 @@
...
-     * if reload is not a function we are currently in a frame and need
-     * to move back to the page scope
-     */
-    if (typeof page.reload !== 'function') {
-        delete this.currentFrame
-        page = this.getPageHandle()
-    }
+    delete this.currentFrame
 
+    let page = this.getPageHandle()
```
mgrybyk: change to const

packages/devtools/src/devtoolsdriver.js
```diff
@@ -133,8 +133,8 @@ export default class DevToolsDriver {
         page.setDefaultTimeout(this.timeouts.get('pageLoad'))
     }
 
-    getPageHandle () {
-        if (this.currentFrame) {
+    getPageHandle ({ frame = false } = {}) {
```
christian-bromann: I guess this way it is extendable but not sure how many properties we will add to this method, therefor I would suggest to just have a parameter called `isInIFrame` which would simplify this a bit

mgrybyk: ok, I'll rename `frame` to `isInFrame` instead of `isInIFrame` because context may be not necessary an `iframe`

mgrybyk: @christian-bromann done

packages/devtools/src/devtoolsdriver.js
```diff
@@ -133,8 +133,8 @@ export default class DevToolsDriver {
         page.setDefaultTimeout(this.timeouts.get('pageLoad'))
     }
 
-    getPageHandle () {
-        if (this.currentFrame) {
+    getPageHandle ({ isInFrame = false } = {}) {
```
christian-bromann: I was more thinking about that:

```suggestion
    getPageHandle (isInFrame) {
```

mgrybyk: got it, I've started with that, but then changed to parameter only because `getPageHandle(true)` doesn't look really informative.

Let me change it back

mgrybyk: @christian-bromann done

PR #4491, opened 2019-09-11
Update travis ci dist to xenial and set @babel/core version to 7.5.5

mgrybyk:
## Proposed changes

- Fix coverage issue by setting fixed @babel/core version 7.5.5
- Fix e2e tests that fails because chrome cannot be installed.

fixes: `Protocol error (Target.getBrowserContexts): \'Target.getBrowserContexts\' wasn\'t found` caused by `dpkg: error processing archive /tmp/google-chrome-stable_current_amd64.deb (--install):
 subprocess dpkg-deb --control returned error exit status 2`

## Types of changes

- [x] CI / build related

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
   },
   "devDependencies": {
     "@babel/cli": "^7.2.3",
-    "@babel/core": "^7.4.0",
+    "@babel/core": "7.5.5",
     "@babel/node": "^7.2.2",
     "@babel/plugin-proposal-class-properties": "^7.4.0",
     "@babel/plugin-proposal-function-bind": "^7.2.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -49,7 +49,7 @@
   },
   "devDependencies": {
     "@babel/cli": "^7.2.3",
-    "@babel/core": "^7.4.0",
+    "@babel/core": "7.5.5",
```
mgrybyk: have to set fixed `@babel/core` version to `7.5.5` to avoid using version `7.6.0` where `/* istanbul ignore next */` is ignored.

related issue https://github.com/babel/babel/issues/10432

Issue #4286, opened 2019-07-31
Add standalone REPL for Appium


simongranger:
**Is your feature request related to a problem? Please describe.**
At the moment we can use the webdriver.io REPL interface by adding driver.debug() in our tests. Unfortunately it's not super easy to use when all you want it try out commands or make a manual demo.

**Describe the solution you'd like**
There is a standalone console app for ruby lib that reads your caps from a appium.txt file. It would be great if we could have a similar solution for webdriver.io

https://github.com/appium/ruby_console

**Describe alternatives you've considered**
I write my tests with webdriver.io but always use Appium Ruby Console for sending manual commands or doing demos. I'd rather be able to do it in webdriver.io

christian-bromann:
This will land in #4286. Let's continue the conversation there.

PR #4510, opened 2019-09-19
@wdio-cli: Appium REPL configurations #4286

lamkovod:
## Proposed changes

Previously REPL worked only with browser driver, but some users asked for Appium debugging features to make it easier and faster.
#4286 

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/docs/repl.md
+++ b/docs/repl.md
@@ -16,10 +16,23 @@ wdio repl chrome -u $SAUCE_USERNAME -k $SAUCE_ACCESS_KEY
 ```
 
 ```sh
-wdio repl appium --config './path/appium.conf.js'
+wdio repl android
+OR
+wdio repl ios
 ```
 
-This would open appium session with desiredCapabilities from provided configuration file. Make sure Appium running on port `4444` in order to initiate the session.
+This would open Chrome/Safari session on connected device/emulator/simulator. Make sure Appium running on port `4444` in order to initiate the session.
+
+```sh
+wdio repl './path/to/your_app.apk'
+```
+
+This would open App session on connected device/emulator/simulator. Make sure Appium running on port `4444` in order to initiate the session.
+
+Capabilities for iOS device can be passed with arguments:
+--ver     - platformVersion
+--device  - deviceName
+--udid    - udid (for real devices)
 
 You can apply any options (see `wdio --help`) available for your REPL session.
 
--- a/docs/repl.md
+++ b/docs/repl.md
@@ -15,6 +15,12 @@ This would open a Chrome browser that you can control with the REPL interface. M
 wdio repl chrome -u $SAUCE_USERNAME -k $SAUCE_ACCESS_KEY
 ```
 
+```sh
+wdio repl appium --config './path/appium.conf.js'
+```
+
+This would open appium session with desiredCapabilities from provided configuration file. Make sure Appium running on port `4444` in order to initiate the session.
+
 You can apply any options (see `wdio --help`) available for your REPL session.
 
 ![WebdriverIO REPL](https://webdriver.io/img/repl.gif)
--- a/packages/wdio-cli/src/commands/repl.js
+++ b/packages/wdio-cli/src/commands/repl.js
@@ -1,15 +1,23 @@
 import { remote } from 'webdriverio'
+import fs from 'fs'
 
 export const command = 'repl <browserName>'
 export const desc = 'Run WebDriver session in command line'
+const DEFAULT_CONFIG_PATH = 'appium.conf.js'
 
 export const handler = async (argv) => {
-    const { browserName } = argv
+    var caps = { browserName: argv.browserName }
+
+    if (caps.browserName === 'appium') {
+        const pathUrl = argv.config ? argv.config : DEFAULT_CONFIG_PATH
+        if(!fs.existsSync(pathUrl)) {
+            throw new Error ('File was not found. Check provided config path')
+        }
+        caps = JSON.parse(fs.readFileSync(pathUrl, 'utf8'))
+    }
 
     const client = await remote(Object.assign(argv, {
-        capabilities: {
-            browserName
-        }
+        capabilities: caps
     }))
 
     global.$ = ::client.$
--- a/packages/wdio-cli/src/commands/repl.js
+++ b/packages/wdio-cli/src/commands/repl.js
@@ -1,20 +1,11 @@
 import { remote } from 'webdriverio'
-import fs from 'fs'
+import { getCapabilities } from '../utils'
 
-export const command = 'repl <browserName>'
+export const command = 'repl <option> [capabilities]'
 export const desc = 'Run WebDriver session in command line'
-const DEFAULT_CONFIG_PATH = 'appium.conf.js'
 
 export const handler = async (argv) => {
-    var caps = { browserName: argv.browserName }
-
-    if (caps.browserName === 'appium') {
-        const pathUrl = argv.config ? argv.config : DEFAULT_CONFIG_PATH
-        if(!fs.existsSync(pathUrl)) {
-            throw new Error ('File was not found. Check provided config path')
-        }
-        caps = JSON.parse(fs.readFileSync(pathUrl, 'utf8'))
-    }
+    let caps = getCapabilities(argv)
 
     const client = await remote(Object.assign(argv, {
         capabilities: caps
--- a/packages/wdio-cli/src/config.js
+++ b/packages/wdio-cli/src/config.js
@@ -55,14 +55,16 @@ WebdriverIO CLI runner
 
 Usage: wdio [options] [configFile]
 Usage: wdio config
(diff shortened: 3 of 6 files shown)


packages/wdio-cli/src/commands/repl.js
```diff
@@ -1,15 +1,23 @@
 import { remote } from 'webdriverio'
+import fs from 'fs'
 
 export const command = 'repl <browserName>'
 export const desc = 'Run WebDriver session in command line'
+const DEFAULT_CONFIG_PATH = 'appium.conf.js'
 
 export const handler = async (argv) => {
-    const { browserName } = argv
+    var caps = { browserName: argv.browserName }
```
mgrybyk: I'd suggest using let instead of var.

docs/repl.md
```diff
@@ -15,6 +15,12 @@ This would open a Chrome browser that you can control with the REPL interface. M
 wdio repl chrome -u $SAUCE_USERNAME -k $SAUCE_ACCESS_KEY
 ```
 
+```sh
+wdio repl appium --config './path/appium.conf.js'
```
mgrybyk: Can you please add an example of appium.conf.js ?

packages/wdio-cli/src/commands/repl.js
```diff
@@ -1,15 +1,23 @@
 import { remote } from 'webdriverio'
+import fs from 'fs'
 
 export const command = 'repl <browserName>'
 export const desc = 'Run WebDriver session in command line'
+const DEFAULT_CONFIG_PATH = 'appium.conf.js'
```
mgrybyk: Looks like it should be json instead of js if you are parsing file as json

packages/wdio-cli/src/config.js
```diff
@@ -55,14 +55,16 @@ WebdriverIO CLI runner
 
 Usage: wdio [options] [configFile]
 Usage: wdio config
-Usage: wdio repl <browserName>
+Usage: wdio repl <browserName> [appiumConfigFile]
```
christian-bromann: I rather prefer a Chrome example here
```suggestion
Usage: wdio repl [options] <browserName>
```

christian-bromann:
This patch as it is seems for me like a workaround and is not generalized enough. I see the need to be able to also spin up Appium test in repl mode but not like this. I am open for other suggestions.

lamkovod:
> This patch as it is seems for me like a workaround and is not generalized enough. I see the need to be able to also spin up Appium test in repl mode but not like this. I am open for other suggestions.

Well, yes, I totally agree that it should be generalized. 
The easiest option is - provide to `handler` path to JSON configuration file with config parameter (get rid of <browserName>) and read all capabilities from it without parsing. 
Than it would be breaking change, but fast and easy refactor without any patching.
What do you think @christian-bromann ?

christian-bromann:
> What do you think @christian-bromann ?

How about something around the lines of:
```sh
$ wdio repl ./path/to/app.apk
```
or for mobile web tests
```sh
$ wdio repl safari --mobile
```
where it checks if Appium is running on port `4723` and fails if not.

lamkovod:
`$ wdio repl ./path/to/app.apk`
Well, not sure that launch repl on some static simulator/emulator is a good idea. 

To make life easier I suggest pass just config file or use default one to be able to handle it in terms of capabilities, because debugging of native apps often requires specific device.
Like this:
`wdio --config repl.conf.json`

repl.conf.json:
```
{
    "platformName": "iOS",
    "deviceName": "iPhone 7",
    "platformVersion": "11.3",
    "browserName": "Safari"
}
```
or
```
{
    "automationName": "XCUITest",
    "platformName": "iOS",
    "deviceName": "iPhone 7",
    "platformVersion": "11.3",
    "app": "/Users/avod/Documents/AG/e2e/apps/MoPlay.app"
}
```

 For user it would be easier to create .json for debugging only once and then just launch`wdio repl`.

christian-bromann:
> Well, not sure that launch repl on some static simulator/emulator is a good idea.

Well it can detect if you want to run an iOS or Android app and starts Appium with the right capabilities.

> For user it would be easier to create .json for debugging only once and then just launch`wdio repl`.

What is the difference of creating full config then? The idea of a repl is to start something/anything with a simple command to debug. It shouldn't matter which kind of device or OS version. It is pretty much about start me something in this environment. It should be the task of WebdriverIO to figure out what is available and start that.

mgrybyk:
@christian-bromann @lamkovod for iOS version or udid is mandatory.
From my point of view syntax may look like this (from high level perspective):
- real device + safari `wdio repl ios --udid foobar` 
- simulator + app `wdio repl ios --version 12.2 path_to_app`

It's really important to specify ios version / udid, otherwise repl would be pretty much useless.

I think there are no required properties for Android, so it may look like this:
- real device + chrome `wdio repl android` 
- simulator + app `wdio repl android path_to_apk`

christian-bromann:
@mgrybyk shouldn't it be possible to get all available UDIDs? If so we can just take the first available on if not specified.

mgrybyk:
Can't say that for sure but I'd avoid calling bash commands if possible. 
Let's investigate it properly first before coming up with final decisions. 

@lamkovod can you assist on that? What is required to run appium session for android/iOS simulator/real device browser/app?

lamkovod:
Android session we need only these capabilities:
`{ platformName: 'Android', automationName: 'UiAutomator2', deviceName: 'Test' }`. It can be easily used with any real device or launched emulator. It will take first Android device with adb command.
 
iOS capabilities, at least:
```
{
    "automationName": "XCUITest",
    "platformName": "iOS",
    "deviceName": "iPhone 7",
    "browserName": "Safari"
}
``` we should provide `udid` OR `deviceName` for simulator (last version will be launched), so for iOS we got 2 options:
- check for available simulators(Xcode)
- ask user to provide at least name of device

**Comment by christian-bromann** [comment]
> Android session we need only these capabilities:

We can assume these values if someone calls the repl command on an apk. They don't need to be provided by the user.

> ask user to provide at least name of device

Let's try to assume first and then ask if we can't find a device to run in.

File: packages/wdio-cli/src/utils.js [hunk 5]
```diff
@@ -196,3 +197,38 @@ export function addServiceDeps(names, packages, update) {
...
+            }
+        )
+    }
+    return finalCapabilities
+}
+
+function getPlatformCaps(option) {
+    let platformConfig = {}
+    switch (option) {
+    case /.*\.(apk|app|ipa)$/.test(option) && option:
```
baruchvlz: Why check if option exists? Doing `.test` on a falsy value returns false. Also if option is a falsy value, the switch will default.

packages/wdio-cli/src/utils.js
```diff
@@ -196,3 +197,38 @@ export function addServiceDeps(names, packages, update) {
...
+        platformConfig = Object.assign({ app: option }, option.endsWith('apk') ? ANDROID_CONFIG : IOS_CONFIG)
+        break
+    case 'android':
+        platformConfig = Object.assign({ browserName: 'Chrome' }, ANDROID_CONFIG)
+        break
+    case 'ios':
+        platformConfig = Object.assign({ browserName: 'Safari' }, IOS_CONFIG)
+        break
+    default:
+        platformConfig = { browserName: option }
```
baruchvlz: Instead of reassigning `platformConfig` and then returning it, you could just return the assignment from the switch / case statement. Also instead of using `Object.assign` you could just destructure the object, but this is more a matter of preferability.

Example:

```js
case /.*\.(apk|app|ipa)$/.test(option) && option:
    return {
        app: option,
        ...(option.endsWith('apk') ? ANDROID_CONFIG : IOS_CONFIG)
        // or ...ANDROID_CONFIG / ...IOS_CONFIG for the other 2 cases
    }
```

packages/wdio-cli/tests/utils.test.js
```diff
@@ -248,3 +250,32 @@ describe('addServiceDeps', () => {
...
+            ...IOS_CONFIG,
+            browserName: 'Safari',
+            platformVersion: null,
+            udid: null })
+    })
+
+    it('should return driver with capabilities for desktop', () => {
+        expect(getCapabilities({ option: 'chrome' })).toEqual({ browserName: 'chrome' })
+    })
+})
```
baruchvlz: The object checks could be snapshots instead.

For example
```js
expect(getCapabilities({ option: 'chrome' })).toMatchSnapshot()
```

This way we don't have to worry about changing the objects in the tests if we change something in the function's return. We just update the snapshots.

christian-bromann:
I will close this PR. I like the general idea of spinning up Appium sessions via Repl but I don't like the solution outlined in this patch. As mentioned in the commends above the repl should figure out how to run the session, no config file required.

docs/repl.md
mrister: anybody have a working example maybe ? @mgrybyk

PR #4516, opened 2019-09-20
Fix Error in "undefined" in Jasmine framework

mgrybyk:
## Proposed changes

Print test title on test failure.

before: `Error in "undefined"`
after: `Error in "some test ROOT"`

## Types of changes

## Checklist


mgrybyk:
@christian-bromann no idea what is going on with codecov.
Maybe I should do a force push or something

PR #4544, opened 2019-09-27
wdio-cli: add glob support for the exclude param

schuttsm:
## Proposed changes

On wdio-cli, there's a bug where exclude param is a regex match, not a glob match.  My fix will change it so it attempts a glob match first, if there are no matches, it will attempt a regex match, so as to be backwards compatible.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/wdio-config/src/lib/ConfigParser.js
+++ b/packages/wdio-config/src/lib/ConfigParser.js
@@ -218,7 +218,7 @@ export default class ConfigParser {
             if (fs.existsSync(filteredFile) && fs.lstatSync(filteredFile).isFile()) {
                 filesToFilter.add(path.resolve(process.cwd(), filteredFile))
             } else if (globMatchedFiles.length) {
-                globMatchedFiles.forEach(file => filesToFilter.add(file))
+                globMatchedFiles.forEach(filesToFilter.add)
             } else {
                 fileList.forEach(file => {
                     if (file.match(filteredFile)) {
--- a/packages/wdio-config/src/lib/ConfigParser.js
+++ b/packages/wdio-config/src/lib/ConfigParser.js
@@ -218,7 +218,7 @@ export default class ConfigParser {
             if (fs.existsSync(filteredFile) && fs.lstatSync(filteredFile).isFile()) {
                 filesToFilter.add(path.resolve(process.cwd(), filteredFile))
             } else if (globMatchedFiles.length) {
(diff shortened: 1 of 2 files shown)


packages/wdio-config/src/lib/ConfigParser.js
```diff
@@ -214,8 +214,17 @@ export default class ConfigParser {
         const filesToFilter = new Set()
         const fileList = ConfigParser.getFilePaths(config)
         cliArgFileList.forEach(filteredFile => {
+            let matchedFiles = ConfigParser.getFilePaths(glob.sync(filteredFile))
+            let globMatchedFiles = []
+            matchedFiles.forEach(file => {
+                if (matchedFiles.indexOf(file) != -1) {
```
baruchvlz: I don't get this statement. You're verifying that the array you're looping has the value that you got from that same array?

packages/wdio-config/src/lib/ConfigParser.js
```diff
@@ -214,8 +214,17 @@ export default class ConfigParser {
...
+                    globMatchedFiles.push(file)
+                }
+            })
```
baruchvlz: This whole loop is weird. Could you explain the reason behind it?

packages/wdio-config/src/lib/ConfigParser.js
```diff
@@ -214,8 +214,17 @@ export default class ConfigParser {
...
             if (fs.existsSync(filteredFile) && fs.lstatSync(filteredFile).isFile()) {
                 filesToFilter.add(path.resolve(process.cwd(), filteredFile))
+            } else if (globMatchedFiles.length) {
+                globMatchedFiles.forEach(file => {filesToFilter.add(file)})
```
baruchvlz: You can do something like this if you want to do a one liner:

```js
globMatchedFiles.forEach(filesToFilter.add)
```

packages/wdio-config/src/lib/ConfigParser.js
schuttsm: @baruchvlz - it was left over from a previous refactoring.  You're right, it doesn't make sense.

Let me remove that section & clean it up.

schuttsm:
@baruchvlz thank you for the feedback.  I removed the loop you asked about and I think its ready for another look

christian-bromann:
Thanks @schuttsm for the contribution! 👍

Issue #3572, opened 2019-02-18
Missing type definitions for `@wdio/reporter` package


AnnikaNissen:
**Environment (please complete the following information):**
 - **WebdriverIO version:** [5.4.17
 - **Mode:** WDIO Testrunner
 - **If WDIO Testrunner, running sync/async:** sync
 - **Additional wdio packages used (if applicable):** @wdio/reporter

**Describe the bug**
When trying to compile a custom reporter class from TypeScript to JS, the TS compiler complains about missing type definitions for the @wdio/reporter.

My code:
```
import WDIOReporter from '@wdio/reporter';

export default class CustomReporter extends WDIOReporter {
   constructor (options) {
       options = Object.assign(options, { stdout: true });
       super(options);
   }
};
```

The compiler error: 
```
error TS7016: Could not find a declaration file for module '@wdio/reporter'. './node_modules/@wdio/reporter/build/index.js' implicitly has an 'any' type.
 Try `npm install @types/wdio__reporter` if it exists or add a new declaration (.d.ts) file containing `declare module '@wdio/reporter';`
``` 

**Expected behavior**
There should be type definitions available for the `@wdio/reporter` package, so that the compiler passes and the development gets easier because of proper code completion.

dziraf:
I personally got it compile by using `require` instead of `import`:
```
// tslint:disable-next-line:no-var-requires
const WDIOReporter = require("@wdio/reporter").default;
```
However, it would be nice to have type definitions.

mgrybyk:
@lamkovod thank you for dealing with it

PR #4550, opened 2019-09-30
wdio-reporter: Type definitions

lamkovod:
## Proposed changes

Typings for WDIOReporter.
+
added `pageLoad` to `WebdriverIO.Options`

## Types of changes
- [ ] Documentation update

## Checklist

Code example:
```
import WDIOReporter from '@wdio/reporter';

export default class CustomReporter extends WDIOReporter {
   constructor (options: WDIOReporter.Options) {
       options = Object.assign(options, { stdout: true });
       super(options);
   }
};
```
#3572 

### DIFF
--- a/packages/wdio-reporter/reporter.d.ts
+++ b/packages/wdio-reporter/reporter.d.ts
@@ -1,0 +1,68 @@
+declare namespace WDIOReporter {
+    type TestState = 'passed' | 'pending' | 'failed' | 'skipped';
+    type ErrorType = 'AssertionError' | 'Error';
+
+    class Reporter {
+        constructor (options: Options);
+
+        onRunnerStart(): void;
+        onBeforeCommand(): void;
+        onAfterCommand(): void;
+        onScreenshot(): void;
+        onSuiteStart(suite: Suite): void;
+        onHookStart(hook: Hook): void;
+        onHookEnd(hook: Hook): void;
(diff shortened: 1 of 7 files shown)


packages/wdio-reporter/reporter.d.ts
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    interface Options {
+        configFile: string;
+        logFile: string;
+        logLevel: string;
+        stdout?: boolean;
+    }
+
+    interface Suite {
```
mgrybyk: Can we somehow reuse Suite, Test, Hook (or at least something) from Webdriverio types

mgrybyk:
May you please also add some very basic types test

packages/wdio-reporter/reporter.d.ts
lamkovod: I don't think that we can do that. 
It's another package

scripts/templates/webdriverio.tpl.d.ts
```diff
@@ -63,6 +63,7 @@ declare namespace WebdriverIO {
         specFileRetries?: number,
         waitforTimeout?: number,
         waitforInterval?: number,
+        pageLoad?: number,
```
christian-bromann:
```suggestion
```

This is not a WebdriverIO option

Issue #4540, opened 2019-09-27
wdi-cli: chromedriver and selenium-standalone services are not exclusive 


mgrybyk:
[//]: # NOTE: This repository only maintains packages that are listed in the [Readme](https://github.com/webdriverio/webdriverio/blob/master/README.md#packages). Please make sure that your issue is directly caused by one of these packages and if not file an issue in the correct 3rd party package repository.

**Environment (please complete the following information):**
 - **WebdriverIO version:** 5.13.2

**Describe the bug**
chromedriver and selenium-standalone are not exclusive anymore.
This is a regression.

**To Reproduce**
1. run `wdio config`
2. choose chromedriver and selenium-standalone on services

Actual Result:
it is allowed to choose both chromedriver and selenium-standalone

**Expected behavior**
chromedriver and selenium-standalone should be exclusive

mgrybyk:
@baruchvlz can you please take a look?
It is a regression after https://github.com/webdriverio/webdriverio/commit/5272de38dfeb4f4fc0232a81fff116f6725d6c4a

I think it is somehow related to package filtering, not sure

baruchvlz:
@mgrybyk Yea I'll take a look at it.

Are chromedriver and selenium-standalone the only ones that are exclusive? Or are there other combinations?

mgrybyk:
@baruchvlz well, choosing these two together is a common mistake.
There gonna be different cases for sure, can't provide you with such information right now

christian-bromann:
I think general cases are choosing 2 drivers (e.g. chromedriver and selenium standalone service) or having two or more cloud provider services also makes no sense. I guess this is difficult to generalise. Maybe we can add categories to them to help better identify their potential usage.

baruchvlz:
> It is a regression after 5272de3 

I was looking at the commit to see if I could find what was removed, but I can't find it.

> I think it is somehow related to package filtering, not sure

Not quite, this can easily be done via inquirer with the `validate` option on the answers that require this type of check.

I have started working on this, but we'll need to merge #4402 first.

PR #4402, opened 2019-08-25
Updates/wdio cli

baruchvlz:
## Proposed changes

[//]: # (Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue.)

## Types of changes

[//]: # (What types of changes does your code introduce to WebdriverIO?)
[//]: # (_Put an `x` in the boxes that apply_)

- [x] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

[//]: # (_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._)

- [x] I have read the [CONTRIBUTING](https://github.com/webdriverio/webdriverio/blob/master/CONTRIBUTING.md) doc
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)

## Further comments

[//]: # (If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...)

Checklist:
- [x] Replace array of strings with collection of supported packages
- [x] Move shared files to `utils` folder
- [x] Introduce `run` command
- [x] Add missing `help` flags

PR Issue: #4388

PR #4559, opened 2019-10-01
wdio-cli: validate exclusive services

baruchvlz:
Related issue: #4540 

With changes:

Screenshot 2019-10-01 at 20 15 48

*Comments*
If you're wondering why we're using the `$--$` hash, refer to [this](https://github.com/webdriverio/webdriverio/pull/4402#discussion_r318231344) comment.

### DIFF
--- a/packages/wdio-cli/src/constants.js
+++ b/packages/wdio-cli/src/constants.js
@@ -1,7 +1,7 @@
 /* istanbul ignore file */
 export const EXCLUSIVE_SERVICES = {
     'wdio-chromedriver-service': {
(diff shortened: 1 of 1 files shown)


christian-bromann:
> If you're wondering why we're using the `$--$` hash, refer to [this](https://github.com/webdriverio/webdriverio/pull/4402#discussion_r318231344) comment.

Can we remove it from the error message at least?

baruchvlz:
> Can we remove it from the error message at least? 

Yea, good catch. I left the `EXCLUSIVE_SERVICES` as it was and adjusted the validator.

PR #4561, opened 2019-10-01
Remove args filter in executeHooksWithArgs

mgrybyk:
## Proposed changes

Remove args filter in executeHooksWithArgs

## Types of changes

## Checklist

## Further comments

Have no idea why have I added that filter, maybe by mistake or while debugging something.
We didn't have the filter before and I can't see any reason to have it.

If we need to filter out args typeof functions at the point (why?) I can change the filter to `.filter((hook) => typeof hook !== 'function')` and add a comment why we need it.

### DIFF
--- a/packages/wdio-sync/src/executeHooksWithArgs.js
+++ b/packages/wdio-sync/src/executeHooksWithArgs.js
@@ -24,7 +24,7 @@ export default function executeHooksWithArgs (hooks = [], args) {
      * make sure args is an array since we are calling apply
      */
     if (!Array.isArray(args)) {
-        args = [args].filter((hook) => typeof hook === 'function')
+        args = [args]
     }
 
     hooks = hooks.map((hook) => new Promise((resolve) => {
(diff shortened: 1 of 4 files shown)


mgrybyk:
Do we need to release `@wdio/mocha-framework`, `@wdio/jasmine-framework`, `@wdio/cucumber-framework` as well? These packages might be used with wrong `@wdio/sync` or `@wdio/utils` versions. Not sure if we need to handle it

anto-ac:
I've upgraded all packages to the latest version and I'm now getting

```
1-0 worker error { name: 'TypeError',
  message: '(0 , _config.executeHooksWithArgs) is not a function',
  stack:
   'TypeError: (0 , _config.executeHooksWithArgs) is not a function\n    at MochaAdapter.run 
```

As @mgrybyk pointed out, other packages may need to be released.

mgrybyk:
@anto-ac you have upgraded your packages incorrectly. Please update all wdio related packages, you may need to delete package lock json

PR #4566, opened 2019-10-03
Report errors in beforeAll and afterAll with Jasmine

### LINKED ISSUE #3084: Why does not Webdriverio report an error in beforeAll with Jasmine?
I use with Jasmine.
When an error occur in `beforeAll`, WebdriverIO does not report the error, but exit code is 1.
I expect it to report error and test failed.
When an error occur in 'beforeEach', WebdriverIO report it and the test failed.

example code

```
describe('sample', () => {
  beforeAll(() => {
    throw new Error("ERROR")
  })

  it('sample', () => {
    browser.url('http://www.google.com')
  })
})
```

 - **WebdriverIO version: 4.14.0
 - **Mode:** WDIO Testrunner
 - **If WDIO Testrunner, running sync/async:"" sync
 - **Node.js version: 8.6.0
 - **NPM version:** 6.4.1
 - **Browser name and version:** Chrome 70
 - **Platform name and version: mac OS 10.13.6
 - **Additional wdio packages used (if applicable):** [e.g. @wdio/spec reporter, @wdio/selenium-standalone service]
   *    "chromedriver": "2.44.0",
   *   "wdio-chromedriver-service": "0.1.3",
   *  "wdio-jasmine-framework": "0.3.7",

martinfrancois:
## Proposed changes

Closes #3084 
Jasmine currently doesn't report issues in WebdriverIO, if there is an error in `beforeAll` or `afterAll`. This PR causes those errors to get reported correctly.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

- I decided to report errors in `beforeAll` and `afterAll` in a similar way as it is implemented for Mocha currently. This means:
  - `"before all" hook` or `"after all" hook` appear only in the reporting if there is a failure in one of them, respectively
  - If at least one of them fails, the failure count gets increased by one (since the suite failed)
- Currently, the `spec` reporter shows the hooks at the bottom of the suite's report. With Mocha, if `before` fails, the rest of the test suite will not be run. There is no way to change this behavior with Mocha currently. This means, up until now it would either show the `before` (and `after`) hooks that failed with no specs with Mocha or only the specs with the `after` hook, if it failed. However with Jasmine, it will continue to run by default, as long as the `failFast` option is not being used. This means there could be the scenario, where a `beforeAll` hook fails, while there are some specs that have run.
  - I changed the spec reporter to first check if a `beforeAll` hook is present which has failed, and if this is the case, it will be shown at the top of the suite.
  - This shows the timeline better, as the `beforeAll` hook is run first in the suite and it would be confusing if the specs would all come first, with the `beforeAll` and `afterAll` hooks at the bottom of the report.

### DIFF
--- a/packages/wdio-jasmine-framework/src/reporter.js
+++ b/packages/wdio-jasmine-framework/src/reporter.js
@@ -11,6 +11,7 @@ export default class JasmineReporter {
         this.parent = []
         this.failedCount = 0
         this.startedTest = null
+        this.specEventQueue = []
     }
 
     suiteStarted (suite) {
@@ -36,7 +37,7 @@ export default class JasmineReporter {
         test.type = 'test'
         test.start = new Date()
         this.parent[this.parent.length - 1].tests++
-        this.emit('test:start', test)
+        this.queueEvent('test:start', test)
     }
 
     specDone (test) {
@@ -62,9 +63,9 @@ export default class JasmineReporter {
         const e = 'test:' + test.status.replace(/ed/, '')
         test.type = 'test'
         test.duration = new Date() - this.testStart
-        this.emit(e, test)
+        this.queueEvent(e, test)
         this.failedCount += test.status === 'failed' ? 1 : 0
-        this.emit('test:end', test)
+        this.queueEvent('test:end', test)
     }
 
     suiteDone (suite) {
@@ -88,6 +89,7 @@ export default class JasmineReporter {
                 start: Date.now(),
                 status: 'failed'
             })
+            this.emitQueue()
         }
 
         // errors on beforeAll are swallowed by Jasmine and can only be accessed later on
@@ -108,6 +110,8 @@ export default class JasmineReporter {
 
         this.emitHookEnd(suite, 'after all', errorsAfterAll)
 
+        this.emitQueue()
+
         this.parent.pop()
         suite.type = 'suite'
         suite.duration = new Date() - this.suiteStart
@@ -115,6 +119,20 @@ export default class JasmineReporter {
 
         // if suite is failed due to a failure in beforeAll / afterAll, increase failed count
         this.failedCount += suite.status === 'failed' ? 1 : 0
+    }
+
+    /**
+     * To ensure chronological ordering of the events, since Jasmine only exposes the errors in beforeAll at
+     * suiteDone, we queue the events from the specs until we reach suiteDone. Then, we send out the beforeAll events
+     * first, followed by the queued events from the specs.
+     */
+    queueEvent(event, payload) {
+        this.specEventQueue.push({ event: event, payload: payload })
+    }
+
+    emitQueue() {
+        this.specEventQueue.forEach(item => this.emit(item.event, item.payload))
+        this.specEventQueue = []
     }
 
     emit (event, payload) {
--- a/packages/wdio-jasmine-framework/src/reporter.js
+++ b/packages/wdio-jasmine-framework/src/reporter.js
@@ -20,15 +20,9 @@ export default class JasmineReporter {
 
         this.emit('suite:start', suite)
 
-        const hookBeforeAll = Object.assign({}, suite)
-        hookBeforeAll.type = 'hook'
-        hookBeforeAll.description = '"before all" hook'
-        this.emit('hook:start', hookBeforeAll)
+        this.emitHookStart(suite, 'before all')
 
-        const hookAfterAll = Object.assign({}, suite)
-        hookAfterAll.type = 'hook'
-        hookAfterAll.description = '"after all" hook'
-        this.emit('hook:start', hookAfterAll)
+        this.emitHookStart(suite, 'after all')
 
         this.parent.push({
             description: suite.description,
@@ -112,25 +106,9 @@ export default class JasmineReporter {
             errorsAfterAll = errors.filter(error => error.stack.includes('UserContext.afterAll'))
         }
 
-        const hookBeforeAll = Object.assign({}, suite)
-        hookBeforeAll.type = 'hook'
-        hookBeforeAll.parent = parentSuite
-        hookBeforeAll.description = '"before all" hook'
-        if (errorsBeforeAll) {
-            hookBeforeAll.errors = errorsBeforeAll
-            hookBeforeAll.error = errorsBeforeAll[0]
-        }
-        this.emit('hook:end', hookBeforeAll)
+        this.emitHookEnd(suite, parentSuite, 'before all', errorsBeforeAll)
 
-        const hookAfterAll = Object.assign({}, suite)
-        hookAfterAll.type = 'hook'
-        hookAfterAll.parent = parentSuite
-        hookAfterAll.description = '"after all" hook'
-        if (errorsAfterAll) {
-            hookAfterAll.errors = errorsAfterAll
-            hookAfterAll.error = errorsAfterAll[0]
-        }
-        this.emit('hook:end', hookAfterAll)
+        this.emitHookEnd(suite, parentSuite, 'after all', errorsAfterAll)
 
         this.parent.pop()
         suite.type = 'suite'
@@ -170,6 +148,25 @@ export default class JasmineReporter {
 
     getUniqueIdentifier (target) {
         return target.description + target.id
+    }
+
+    emitHookStart(suite, name) {
+        const hook = Object.assign({}, suite)
+        hook.type = 'hook'
+        hook.description = '"' + name + '" hook'
+        this.emit('hook:start', hook)
+    }
+
+    emitHookEnd(suite, parentSuite, name, errors) {
+        const hook = Object.assign({}, suite)
+        hook.type = 'hook'
+        hook.parent = parentSuite
+        hook.description = '"' + name + '" hook'
+        if (errors) {
+            hook.errors = errors
+            hook.error = errors[0]
+        }
+        this.emit('hook:end', hook)
     }
 
     cleanStack (error) {
--- a/packages/wdio-jasmine-framework/src/reporter.js
+++ b/packages/wdio-jasmine-framework/src/reporter.js
@@ -21,10 +21,6 @@ export default class JasmineReporter {
 
         this.emit('suite:start', suite)
 
-        this.emitHookStart(suite, 'before all')
-
-        this.emitHookStart(suite, 'after all')
-
         this.parent.push({
             description: suite.description,
             id: suite.id,
@@ -69,6 +65,29 @@ export default class JasmineReporter {
     }
 
     suiteDone (suite) {
+        // errors on beforeAll are swallowed by Jasmine and can only be accessed at suiteDone
+        let errorsBeforeAll = []
+        let errorsAfterAll = []
+        if (suite.failedExpectations.length) {
+            let errors = suite.failedExpectations
+
+            if (this.shouldCleanStack) {
+                errors = suite.failedExpectations.map(x => this.cleanStack(x))
+            }
+
+            errorsBeforeAll = errors.filter(error => error.stack.includes('UserContext.beforeAll'))
+            errorsAfterAll = errors.filter(error => error.stack.includes('UserContext.afterAll'))
+        }
+
+        // since we cannot know if there was a beforeAll when there is no error, only send out hooks if there are errors
+        if (errorsBeforeAll.length) {
+            this.emitHookStart(suite, 'before all')
+            this.emitHookEnd(suite, 'before all', errorsBeforeAll)
+        }
+
+        // send out queued spec events after beforeAll
+        this.emitQueue()
+
         /**
          * in case there is a runtime error within one of the specs
          * create an empty test to attach the error to it
@@ -92,25 +111,11 @@ export default class JasmineReporter {
             this.emitQueue()
         }
 
-        // errors on beforeAll are swallowed by Jasmine and can only be accessed later on
-        let errorsBeforeAll
-        let errorsAfterAll
-        if (suite.failedExpectations.length) {
-            let errors = suite.failedExpectations
-
-            if (this.shouldCleanStack) {
-                errors = suite.failedExpectations.map(x => this.cleanStack(x))
-            }
-
-            errorsBeforeAll = errors.filter(error => error.stack.includes('UserContext.beforeAll'))
-            errorsAfterAll = errors.filter(error => error.stack.includes('UserContext.afterAll'))
+        // since we cannot know if there was a afterAll when there is no error, only send out hooks if there are errors
+        if (errorsAfterAll.length) {
+            this.emitHookStart(suite, 'after all')
+            this.emitHookEnd(suite, 'after all', errorsAfterAll)
         }
-
-        this.emitHookEnd(suite, 'before all', errorsBeforeAll)
-
-        this.emitHookEnd(suite, 'after all', errorsAfterAll)
-
-        this.emitQueue()
 
         this.parent.pop()
         suite.type = 'suite'
@@ -182,10 +187,8 @@ export default class JasmineReporter {
         hook.type = 'hook'
         hook.parent = this.getParentSuite()
         hook.description = '"' + name + '" hook'
-        if (errors) {
-            hook.errors = errors
-            hook.error = errors[0]
-        }
+        hook.errors = errors
+        hook.error = errors[0]
         this.emit('hook:end', hook)
     }
 
--- a/packages/wdio-jasmine-framework/src/reporter.js
+++ b/packages/wdio-jasmine-framework/src/reporter.js
@@ -101,7 +101,7 @@ export default class JasmineReporter {
         // errors on beforeAll are swallowed by Jasmine and can only be accessed later on
         let errorsBeforeAll
         let errorsAfterAll
-        if (suite.failedExpectations && suite.failedExpectations.length) {
+        if (suite.failedExpectations.length) {
             let errors = suite.failedExpectations
 
             if (this.shouldCleanStack) {
--- a/packages/wdio-jasmine-framework/src/reporter.js
+++ b/packages/wdio-jasmine-framework/src/reporter.js
@@ -68,13 +68,11 @@ export default class JasmineReporter {
     }
 
     suiteDone (suite) {
-        const parentSuite = this.parent[this.parent.length - 1]
-
         /**
          * in case there is a runtime error within one of the specs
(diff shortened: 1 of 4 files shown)


mgrybyk:
I don't really like the approach, first of all not every user wants to see before after All hooks in reports, secondly it's tighten spec reporter. 
Can it be more generic? Does printing errors  works for you?

martinfrancois:
@mgrybyk I see what you mean.
First of all, the before and after hooks only appear in the reports if there are errors (at least in the spec reporter, since every reporter handles these kinds of details by themselves).

mgrybyk:
Got it, so we need to to have failed before after All hook in report.
Let me check one more option quickly. 

Anyway I agree that it would be nice to have hook error shown based on timeline. 
Noticed this not just for jasmine but for Cucumber as well.
Is there a way to avoid relying on hook name and use timestamp or something to keep chronological order in spec reporter?

martinfrancois:
@mgrybyk sure! I don't know how it is with Cucumber, but it would make sense of course.
I was thinking about this as well, however one thing that is clear: without changing the implementation of the spec reporter at all, this would not be possible, as it currently just puts the hooks at the bottom.

Considering the chronological approach, this would be a possibility, however it would mean the following:
- We would need some kind of workaround for Jasmine anyways, we would need to:
  - either "fake" the time stamp of the `hook:end` for `beforeAll` to be the same as the timestamp of `hook:start` (`hook:end` can only be sent at `suiteDone` as the error cannot be accessed from Jasmine earlier) so that it would chronologically get added before the specs that run
  - or we include logic in the specs reporter that would take care of that, however this would kind of be the approach it is now
  - we couldn't just use the time stamp of the `hook:start` as a measure instead, since the `afterAll` hook needs to start before the specs to work
- if we go with the chronological approach, we would need to "merge" both the hooks and the suites together, before we sort them.
  - I'm not sure if this would cause any problems with other frameworks, do you know if there is one where having hooks at the end of the suite would be preferred to having them in chronological order? I can't think of any right now, but I don't know all of them that report in WebdriverIO either...
- Chronologically sorting them would make the spec reporter slower. I don't know how big the impact is, it would depend on the amount of specs of course, the more specs the longer it would take.

As you see both approaches have their advantages and disadvantages... It's not ideal Jasmine is handling it in this way, causing us to come up with workarounds to these issues... Which approach do you think it would make sense to go more? I decided to go the with not sorting it chronologically for the PR, as I felt it would introduce the least amount of breaking changes. But I would also be okay with going the other approach here or in a separate PR!

packages/wdio-jasmine-framework/src/reporter.js
```diff
@@ -87,10 +98,47 @@ export default class JasmineReporter {
             })
         }
 
+        // errors on beforeAll are swallowed by Jasmine and can only be accessed later on
+        let errorsBeforeAll
+        let errorsAfterAll
+        if (suite.failedExpectations && suite.failedExpectations.length) {
```
mgrybyk: suite.failedExpectations is always an array

packages/wdio-jasmine-framework/src/reporter.js
```diff
@@ -87,10 +98,47 @@ export default class JasmineReporter {
...
+
+        const hookBeforeAll = Object.assign({}, suite)
+        hookBeforeAll.type = 'hook'
+        hookBeforeAll.parent = parentSuite
+        hookBeforeAll.description = '"before all" hook'
+        if (errorsBeforeAll) {
+            hookBeforeAll.errors = errorsBeforeAll
+            hookBeforeAll.error = errorsBeforeAll[0]
+        }
+        this.emit('hook:end', hookBeforeAll)
```
mgrybyk: is it possible that we don't have before/after All hooks at all?

packages/wdio-jasmine-framework/src/reporter.js
```diff
@@ -87,10 +98,47 @@ export default class JasmineReporter {
...
+
+            if (this.shouldCleanStack) {
+                errors = suite.failedExpectations.map(x => this.cleanStack(x))
+            }
+
+            errorsBeforeAll = errors.filter(error => error.stack.includes('UserContext.beforeAll'))
+            errorsAfterAll = errors.filter(error => error.stack.includes('UserContext.afterAll'))
+        }
+
+        const hookBeforeAll = Object.assign({}, suite)
```
mgrybyk: this block can be moved to a function that accepts before/after as an argument

packages/wdio-jasmine-framework/src/reporter.js
```diff
@@ -87,10 +98,47 @@ export default class JasmineReporter {
...
+        }
+        this.emit('hook:end', hookAfterAll)
+
         this.parent.pop()
         suite.type = 'suite'
         suite.duration = new Date() - this.suiteStart
         this.emit('suite:end', suite)
+
+        // if suite is failed due to a failure in beforeAll / afterAll, increase failed count
+        this.failedCount += suite.status === 'failed' ? 1 : 0
```
mgrybyk: will the suite be treated as failed in such case?

packages/wdio-spec-reporter/src/index.js
```diff
@@ -147,17 +147,29 @@ class SpecReporter extends WDIOReporter {
...
-             */
-            ...suite.tests,
-            /**
-             * and only hooks that failed
-             */
-            ...suite.hooks
-                .filter((hook) => Boolean(hook.error))
-        ]
+        const beforeAllTitle = '"before all" hook'
+        const beforeAll = suite.hooks.find(hook => Boolean(hook.error) && hook.title === beforeAllTitle)
```
mgrybyk: can we have more than one beforeAll?

packages/wdio-jasmine-framework/src/reporter.js
```diff
@@ -19,6 +19,17 @@ export default class JasmineReporter {
         suite.start = new Date()
 
         this.emit('suite:start', suite)
+
+        const hookBeforeAll = Object.assign({}, suite)
```
mgrybyk: can be moved to a function that accepts hook type as arg

packages/wdio-jasmine-framework/src/reporter.js
```diff
@@ -19,6 +19,17 @@ export default class JasmineReporter {
...
+        hookBeforeAll.type = 'hook'
+        hookBeforeAll.description = '"before all" hook'
+        this.emit('hook:start', hookBeforeAll)
```
mgrybyk: same as with hook:end, not sure that we always have before/after All and not sure if it's only one.

martinfrancois:
Another way to go that would solve the issue without having to sort them after the fact based on the timestamp would be to get rid of the differentiation between hooks and suites, as in just have one array in `wdio-reporter` for the events of both suites and hooks, just filling them in as they come. We could then differentiate between hook or suite in the objects themselves.
This would probably be the cleanest solution, in `wdio-jasmine-framework` we could queue the events of the test specs until we reach `suiteDone`, then emit the hooks for the start and end of `beforeAll`, then emit all the events that were queued, followed by the events of `afterAll`. But it would mean the most breaking changes as well, and this would be something we would need to have a look at in a separate PR for sure.

mgrybyk:
We can't just get rid of them, it's gonna be a breaking change. But we can introduce one more list as I've mentioned above.

packages/wdio-jasmine-framework/src/reporter.js
martinfrancois: Yes. But that's part of the problem... I couldn't come up with a way to find out if there are tests in `beforeAll` or `afterAll` with Jasmine, as Jasmine keeps them separated before the run only, then they all get put in a queue and get run all together. At this point I couldn't find a way to differentiate them again.

packages/wdio-jasmine-framework/src/reporter.js
martinfrancois: Will do!

packages/wdio-jasmine-framework/src/reporter.js
martinfrancois: Yes. This is at least the current behavior with Mocha. `beforeAll` and `afterAll` are pre-/postconditions of the suite, so if one of them fails, the suite can be considered failed as well.

packages/wdio-spec-reporter/src/index.js
martinfrancois: No, there can only be one `beforeAll`

packages/wdio-jasmine-framework/src/reporter.js
martinfrancois: Will do!

martinfrancois:
Ah I see! Yes, that would be a possibility. But this would mean we would need to maintain one combined and two separate arrays, which increases the complexity. Let's see what @christian-bromann thinks about this

packages/wdio-jasmine-framework/src/reporter.js
mgrybyk: We can detect before after All hooks in the end and trigger start event just before stop.
At least in such case we wont add not existing hooks

packages/wdio-jasmine-framework/src/reporter.js
martinfrancois: Is it guaranteed that Jasmine will always provide it? I was not sure, which is why I added the check, but I could also remove it

packages/wdio-jasmine-framework/src/reporter.js
martinfrancois: See comment from before here as well ;-)

packages/wdio-jasmine-framework/src/reporter.js
mgrybyk: According to their docs on custom reporter it should be

packages/wdio-jasmine-framework/src/reporter.js
martinfrancois: I think there must be a misunderstanding... We can only know for sure at the end of the suite if there was a before / after hook if there was a failure in one of them. Else if there was a `beforeAll` that passed, we cannot differentiate it from a regular spec anymore, except if we can guarantee the name of the `beforeAll` was not defined. However, assuming nobody would specify a name for `beforeAll` seemed a bit dangerous to me. That's why I decided to consistently use the hooks. We could also just send them out if there was a failure, but this would still not be an accurate representation...

packages/wdio-jasmine-framework/src/reporter.js
martinfrancois: Just removed it and added empty `failedExpectation` arrays to each of the unit tests as well with `suiteDone` so they won't fail

packages/wdio-jasmine-framework/src/reporter.js
martinfrancois: Just refactored this

packages/wdio-jasmine-framework/src/reporter.js
martinfrancois: Just refactored this one as well

packages/wdio-jasmine-framework/src/reporter.js
```diff
@@ -19,6 +19,11 @@ export default class JasmineReporter {
         suite.start = new Date()
 
         this.emit('suite:start', suite)
+
+        this.emitHookStart(suite, 'before all')
```
mgrybyk: I still think that it, as well as emit hook end, can be moved to `if (suite.failedExpectations.length) {` block, so that we only add failed hooks to report

packages/wdio-jasmine-framework/src/reporter.js
```diff
@@ -87,10 +90,31 @@ export default class JasmineReporter {
...
+
+            if (this.shouldCleanStack) {
+                errors = suite.failedExpectations.map(x => this.cleanStack(x))
+            }
+
+            errorsBeforeAll = errors.filter(error => error.stack.includes('UserContext.beforeAll'))
+            errorsAfterAll = errors.filter(error => error.stack.includes('UserContext.afterAll'))
+        }
+
+        this.emitHookEnd(suite, 'before all', errorsBeforeAll)
```
mgrybyk: I'd rename hooks to beforeAll / afterAll to avoid possible overlapping with mocha hooks

martinfrancois: This was actually intentional. By going with this approach, the other reporters need less adjustment. For example, Allure shows the hooks in the report only if it matches this name exactly (since it matches for it). Since mocha and jasmine cannot be used together anyways, I don't think there would be any issues.

martinfrancois:
> * error in `beforeAll` hook added to root scope is still ignored.

Can you please explain that further, maybe with an example? I'm not sure what you mean..

martinfrancois:
Based on the current situation I would propose the following changes, please tell me if you agree:
- change the hooks only to get sent out if there is an error
- queue the events from the specs until `suiteDone` is called, where the `beforeAll` hook would be sent first if there is an error, followed by emitting the events in the queue, followed by an event to the `afterAll` hook, should there be an error
  - this way when we decide to implement the "chronological" approach by using the combined array, we wouldn't need to change anything in the jasmine framework itself, it would already work.
- we would leave the changes to the spec reporter like it is now and in a separate PR, (if @christian-bromann agrees with the combined array approach) we would adjust the wdio-reporter to implement the combined array and wdio-spec-reporter to use the combined array to show the hooks / suites in chronological order
  - since we need to change the function I adjusted here in the spec reporter anyways during these changes, I think it would be okay to leave them as it is now

This way we can leave the other PR for the details regarding the change on `wdio-reporter`, as we may stumble upon challenges that we didn't think of now, so we don't delay this functionality too much.

What do you think?

mgrybyk:
@martinfrancois just add `beforeAll` hook outside describe (in a root scope).

mgrybyk:
I like the idea to avoid affecting spec reporter and improve it later on so that it can print tests/hooks in chronological way. It is also helpful for Cucumber framework

martinfrancois:
> @martinfrancois just add `beforeAll` hook outside describe (in a root scope).

@mgrybyk is this even supported by Jasmine? I never thought of using `beforeAll` outside of a `describe` before and I couldn't find any code snippets online that used `beforeAll` in this way... But I could be missing something.
The docs from Jasmine are not very specific about this. While I couldn't find an example of it there as well, the docs say the following about `beforeAll`:

> Run some shared setup once before all of the specs in the `describe` are run.

It's not clear though if this means that the `beforeAll` must be in the `describe` or not... 

Is it something you have seen used somewhere and if so I would be interested to know what the reason for putting it outside of the `describe` block.

packages/wdio-spec-reporter/src/index.js
```diff
@@ -147,17 +147,29 @@ class SpecReporter extends WDIOReporter {
...
+        if (beforeAll) {
+            // since Jasmine allows test cases to run even if beforeAll failed, ensure beforeAll is on the top
+            return [beforeAll,
+                // report all tests
+                ...suite.tests,
+                // and only hooks that failed, without beforeAll
+                ...suite.hooks.filter(hook => Boolean(hook.error) && hook.title !== beforeAllTitle)
+            ]
+        } else {
+            return [
```
christian-bromann: no else needed here

martinfrancois: just removed it. I was just wondering, if there is an ESLint rule for that, we should add it!

martinfrancois: Found it: https://github.com/webdriverio/webdriverio/pull/4568

mgrybyk:
two point here. 
- I'd avoid changing spec reporter with this PR and raise an issue to fix it so that it would work for both Jasmine and Cucumber instead of applying workaround.
This is not related to the issue, the core problem is that before/after All hook errors are not printed and are missing reporters.

- I find a bit easier way to deal with it that would also cover cases when before/after All is in root scope AND errors will be printed when they happen instead of in the end of the suite.

martinfrancois:
1. Point: Alright, makes sense, let's keep the discussion around that in #4571. I will revert the changes to the spec reporter here.
2. Point:
> errors are printed when they happen instead of in the end of the suite.

I agree, this would be better, but this is not possible due to the way Jasmine works. If you read through https://github.com/webdriverio/webdriverio/issues/3084 you will see that Jasmine swallows the exceptions and failures until `suiteDone` is called, where they are inserted back again. This means it simply isn't possible for us to do this without changing the behavior of Jasmine itself, which is beyond the scope of this PR. In my opinion, given the implementation of this PR uses the correct chronological order of the calls, it's good enough and already provides a big value over what is the case now (no errors are visible at all in the reporting).

 > I find a bit easier way to deal with it that would also cover cases when before/after All is in root scope

I see.. I had a look at this issue and had a look at how Jasmine handles errors in `beforeAll` that is in root scope. It seems like those are not reported to `suiteDone` at all, as the `failedExpectations` array doesn't include them... This is probably because at root scope, there is no suite that Jasmine could report something to. Since the error cannot be accessed by WebdriverIO, we would need to figure out how to handle these issues at Jasmine first, too I think...

mgrybyk:
> errors are printed when they happen instead of in the end of the suite.

It's possible, already made it in my PR

martinfrancois:
> > errors are printed when they happen instead of in the end of the suite.
> 
> It's possible, already made it in my PR.

Oh, I wasn't aware of that!

> > don't think this should be an issue
> 
> Maybe, but I'd prefer it to work to avoid fixing it in future, especially if it doesn't take an extra effort

Sure, it's just I didn't find a way to do it at all, so it's not like it was no extra effort for me :wink:

How should we proceed? Should I integrate your changes into mine, or do you think you need to re-do it from scratch?

mgrybyk:
If the decision is not urgent I'd better wait a bit for someone's else input.
Maybe there is even better way to handle it

martinfrancois:
Okay!

mgrybyk:
@martinfrancois anyway thanks for dealing with it, all these fixes are happening with your help!

martinfrancois:
@mgrybyk You're welcome, I'm glad they are helping! Thank you too for your support!

christian-bromann:
@martinfrancois a lot of discussions on this thread, can you summarize what will change with this patch?

mgrybyk:
I think we can  close it.
Gonna do it with https://github.com/webdriverio/webdriverio/pull/4566

christian-bromann:
Sounds good to me. Thanks @martinfrancois for your work here!

martinfrancois:
I wanted to suggest the same. Thanks!

Issue #4558, opened 2019-10-01
Add unit test for /packages/wdio-cli/src/index.js


christian-bromann:
We recently had a regression where changes to the default wdio command changed the default behavior. Let's add some testing around it to make sure it doesn't happen again.

christian-bromann:
This work has been done in #4587

PR #4570, opened 2019-10-05
Add tests to cli index and clean up index file

baruchvlz:
Issue: #4558

I'm also cleaning up a bit of the index file per the comment [here](https://github.com/baruchvlz/webdriverio/blob/chore/cli-tests/packages/wdio-cli/src/index.js#L25-L32)

With current implementation:

`npx wdio wdio.conf.js -l=silent`

Screenshot 2019-10-05 at 10 52 00

### DIFF
--- a/packages/wdio-cli/src/index.js
+++ b/packages/wdio-cli/src/index.js
@@ -46,19 +46,22 @@ export const run = async () => {
         .map((file) => file.slice(0, -3))
 
     if (!params._.find((param) => supportedCommands.includes(param))) {
-        params.configPath = path.join(process.cwd(), params._[0] || DEFAULT_CONFIG_FILENAME)
-        params._.push(fs.existsSync(params.configPath) ? 'run' : 'config')
-        return handler(params).catch(async (err) => {
-            const output = await new Promise((resolve) => {
-                yargs.parse('--help', (err, argv, output) => resolve(output))
-            })
+        const configPath = params._[0]
+        const workingDir = configPath.startsWith('/') ? '' : process.cwd()
 
-            console.error(`${output}\n\n${err.stack}`)
-            process.exit(1)
+        params.configPath = path.join(workingDir, configPath || DEFAULT_CONFIG_FILENAME)
+
+        return handler(params).catch(async (err) => {
+            yargs.parse('--help', (_, __, output) => {
+                console.error(`${output}\n\n${err.stack}`)
+
+                if (!process.env.JEST_WORKER_ID) {
+                    /* istanbul ignore next */
+                    process.exit(1)
+                }
+            })
         })
     }
-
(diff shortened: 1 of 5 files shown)


packages/wdio-cli/src/index.js
```diff
@@ -25,16 +22,21 @@ export const run = async () => {
...
     /**
-     * if yargs doesn't run a command from the command directory
-     * we verify if a WDIO config file exists in the given "path" parameter
-     * if so, we assume the user ran `wdio path/to/wdio.conf.js` and we execute `wdio run` command
+     * The only way we reach this point is if the user runs the binary without a command (i.e. wdio wdio.conf.js)
+     * We can safely assume that if this is the case, the user is trying to execute the `run` command as it
+     * was previous to https://github.com/webdriverio/webdriverio/pull/4402
+     *
+     * Since the `run` command verifies if the configuration file exists before executing
+     * we don't have to check that again here.
```
baruchvlz: @christian-bromann Do you think it'll make sense to verify if the config file exists before executing yargs? This check would be ignored if the user runs the config command.

mgrybyk: what is the expected result if config exists and user runs wdio config?
Should we ask to overwrite config or something?
(sorry if I misunderstood the question)

baruchvlz: @mgrybyk  If you run `wdio config` and there is a config already then you will overwrite it. That's the behaviour since a while if I recall correctly. We could add a check to see if a wdio.conf.js file exists and either exit or ask to overwrite.

My question was if we should verify if the config file exists before running a command other than `config` instead of verifying it inside each individual command. Currently we do this check in the [run](https://github.com/webdriverio/webdriverio/blob/master/packages/wdio-cli/src/commands/run.js#L117) and [install](https://github.com/webdriverio/webdriverio/blob/master/packages/wdio-cli/src/commands/install.js#L60) commands.

packages/wdio-cli/src/index.js
```diff
@@ -34,25 +33,32 @@ export const run = async () => {
...
-        params._.push(fs.existsSync(params.configPath) ? 'run' : 'config')
+        const configPath = params._[0]
+
+        params.configPath = path.resolve(process.cwd(), configPath || DEFAULT_CONFIG_FILENAME)
+
         return handler(params).catch(async (err) => {
-            const output = await new Promise((resolve) => {
-                yargs.parse('--help', (err, argv, output) => resolve(output))
-            })
+            yargs.parse('--help', (_, __, output) => {
```
christian-bromann: let's keep this as is as it has less indentation and should work the same way

christian-bromann: I will fix that in my PR

PR #4572, opened 2019-10-05
Report errors in beforeAll and afterAll with Jasmine

mgrybyk:
## Proposed changes

print and report errors in `beforeAll` and `afterAll` hooks.
- Errors are printed when they happen (not in the end of suite)
- `beforeAll` and `afterAll` hook errors in root scope are printed as well

image

image

## Types of changes

## Checklist

## Further comments

another approach to deal with #3084 

ref: @martinfrancois  https://github.com/webdriverio/webdriverio/pull/4566


mgrybyk:
going to add unit tests soon

martinfrancois:
@mgrybyk feel free to copy the unit tests from my branch (from my original PR), the behavior should be the same 😉

mgrybyk:
had to decouple `'should properly set up jasmine'` because it became too huge.

tests/jasmine/test.js
```diff
@@ -1,4 +1,16 @@
+beforeAll(() => {})
+afterAll(() => {
+    throw new Error('afterAll failure in root scope')
```
christian-bromann: two questions:

1. shouldn't this fail the smoke tests?
2. if not can we somehow assert the spec report that it outputs proper logs?

mgrybyk: 1. beforeAll / afterAll hook failures are not failing the suite
2.1. root scope hooks and tests don't appear in any reporters including spec (for jasmine)
2.2. don't know if we can validate spec reporter result, let me check that

mgrybyk: point 2.1 is a defect I think (not related to the PR)

mgrybyk: we can validate reporter output with `wdio-smoke-test-reporter`, let me do that

mgrybyk: @christian-bromann added smoke test with reporter and raised a defect https://github.com/webdriverio/webdriverio/issues/4582

PR #4587, opened 2019-10-08
add examples and epilogue to cli commands

christian-bromann:
## Proposed changes

Additional information to the CLI command interface.

* `wdio --help` (before/after)

Screen Shot 2019-10-08 at 11 14 46 AM
Screen Shot 2019-10-08 at 11 35 59 AM

* `wdio config --help` (before/after)

Screen Shot 2019-10-08 at 11 16 35 AM
Screen Shot 2019-10-08 at 11 16 49 AM

* `wdio install --help` (before/after)

Screen Shot 2019-10-08 at 11 17 32 AM
Screen Shot 2019-10-08 at 11 17 47 AM

* `wdio repl --help` (before/after)

Screen Shot 2019-10-08 at 11 18 32 AM
Screen Shot 2019-10-08 at 11 18 45 AM

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

Tests will be added by @baruchvlz


christian-bromann:
> Are you going to add tests?

Yes, it will be happen as part of #4570

mgrybyk:
got it, let's wait for the tests then otherwise build will start failing

packages/wdio-cli/src/index.js
```diff
@@ -1,39 +1,65 @@
...
     const params = { ...argv.argv }
-    if (!params._.find((param) => SUPPORTED_COMMANDS.includes(param))) {
-        params.configPath = path.join(process.cwd(), params._[0] || DEFAULT_CONFIG_FILENAME)
-        params._.push(fs.existsSync(params.configPath) ? 'run' : 'config')
-        return handler(params)
+    const supportedCommands = fs
+        .readdirSync(path.join(__dirname, 'commands'))
+        .map((file) => file.slice(0, -3))
+
+    if (!params._.find((param) => supportedCommands.includes(param))) {
```
mgrybyk: what is `params._` ?

christian-bromann: This is the command that is being passed in. Not sure if there is a different way to access it, I can check

christian-bromann: This is not an internal property and recommended to use by args: https://github.com/yargs/yargs/blob/master/docs/examples.md#and-non-hyphenated-options-too-just-use-argv_

mgrybyk:
Is it expected to have `Not enough non-option arguments: got 0, need at least 1` in the end instead of epilogue?
image

I don't mind, just want to make sure it works as expected.

christian-bromann:
> I don't mind, just want to make sure it works as expected.

What did you executed?

I am currently running manual tests to check functionality of it.

mgrybyk:
`./node_modules/.bin/wdio run`

christian-bromann:
it should ask you if you want to create a config file if you don't have one

mgrybyk:
for me error message like `Not enough non-option arguments: got 0, need at least 1` makes sense.

I wanted to run conf file but didn't proved it, so pretty much fair.

The question was only about epilogue if it should be in the very end in such case or no, both cases works for me

christian-bromann:
I think if the epilogue is always at the end such error message can get lost so I am fine with the observed behavior

mgrybyk:
Can you increase unit test coverage a bit?

PR #4591, opened 2019-10-08
wdio-devtools-service: better browser name and version check

### LINKED ISSUE #4589: Devtools: Unable to run performance test
 - **WebdriverIO version:** 5.12
 - **Mode:** WDIO Testrunner
 - **If WDIO Testrunner, running sync/async:** Sync
 - **Node.js version:** v10.16.2
 - **NPM version:** v6.11.3
 - **Browser name and version:** Chrome 77 ran through Browserstack
 - **Platform name and version:** Windows 10
 - **Additional wdio packages used (if applicable):** @wdio/devtools-service

**Config of WebdriverIO**
```
exports.config = {
   user: bstack_user,
   key: bstack_password,
   updateJob: false,
    maxInstances: 1,
    specs: [
        './tests/partner_portal/**/policy_secondary_*.spec.js'
    ],
capabilities: [{
        'os': 'Windows',
        'os_version': '10',
        'browserName': 'Chrome',
        'browser_version': '77.0',
        build: 'partner_portal-test',
        'project': 'Partner_Portal',
        acceptInsecureCerts: true,
        'browserstack.local': true,
        resolution: '1920x1080'
    }],
services: [
        'devtools'
    ],
logLevel: 'warn',
    coloredLogs: true,
    screenshotPath: './errorShots/',
    baseUrl: partnerURL,
    waitforTimeout: 20000,
    connectionRetryTimeout: 10000,
    host: 'hub.browserstack.com',
    framework: 'mocha',
    mochaOpts: {
        ui: 'bdd',
        timeout: 100000,
        compilers: ['@babel/register']
    },
    before: function () {
        var chai = require('chai');
        global.expect = chai.expect;
        chai.Should();
    },
    onPrepare: function (config, capabilities) {
        console.log("Connecting local");
        return new Promise(function (resolve, reject) {
            exports.bs_local = new browserstack.Local();
            exports.bs_local.start({
                'key': exports.config.key
            }, function (error) {
                if (error) return reject(error);
                console.log('Connected. Now testing...');

                resolve();
            });
        });
    },
    // Code to stop browserstack local after end of test
    onComplete: function () {
        exports.bs_local.stop(function () {});
    }
}
}
```

**Describe the bug**
`The @wdio/devtools-service currently only supports Chrome version 63 and up`

I receive this error even with browserstack capabilities set to Chrome 77.

**To Reproduce**
Please see below example performance file

```
describe('Testing the partner portal\'s initial paint time', () => {
    before(function() {
        browser.enablePerformanceAudits();
    });

    after(function() {
        browser.disablePerformanceAudits()
    });

    it('Should load within performance constraints that will be set at a later point in time', () => {
        browser.url(url);
        let metrics = browser.getMetrics();
        assert.ok(metrics.speedIndex < 1500);

        let score = browser.getPerformanceScore();
        console.log(score);

        $('#AgencyCode').setValue(partner.PARTNER_PORTAL_AGENCY_CODE);
        $('#frmRouter > div > div > div > div:nth-child(2) > div > input:nth-child(2)').click();

        metrics = browser.getMetrics()
        assert.ok(metrics.speedIndex < 1500)
        score = browser.getPerformanceScore()
        assert.ok(score >= .99)
    });
});
```

**Expected behavior**
According to the docs this should run as long as 'devtools' is added to services array

**Log**
```
Connecting local
Connected. Now testing...
[0-0] RUNNING in Chrome - C:\Users\xxx\Desktop\projects\webdriver-xxx-test\performance\partner_perf.js
[0-0] 2019-10-08T17:36:07.714Z **ERROR @wdio/devtools-service: The @wdio/devtools-service currently only supports Chrome version 63 and up**
[0-0] initialised custom reporter with the following reporter options: { logLevel: 'warn',
  setLogFile: [Function: NOOP],
  someOption: 'foobar',
  logFile: undefined,
  writeStream: { write: [Function: write] } }
Some log line
[0-0] We are beginning the test!
[0-0] TypeError in "Testing the partner portal's initial paint time "before all" hook for "Should load within performance constraints that will be set at a later point in time""
browser.enablePerformanceAudits is not a function
[0-0] We've completed the suite!
[0-0] TypeError in "Testing the partner portal's initial paint time "after all" hook for "Should load within performance constraints that will be set at a later point in time""
browser.disablePerformanceAudits is not a function
```

mgrybyk:
## Proposed changes

Tune supporter version verification to be case insensitive and also support other possible version property names

fixes #4589

## Types of changes

## Checklist

### DIFF
--- a/packages/wdio-devtools-service/src/utils.js
+++ b/packages/wdio-devtools-service/src/utils.js
@@ -9,6 +9,7 @@ const log = logger('@wdio/devtools-service:utils')
 
 const RE_DEVTOOLS_DEBUGGING_PORT_SWITCH = /--remote-debugging-port=(\d*)/
 const RE_USER_DATA_DIR_SWITCH = /--user-data-dir=([^-]*)/
+const VERSION_PROPS = ['browserVersion', 'browser_version', 'version']
 
 /**
  * Find Chrome DevTools Interface port by checking Chrome switches from the chrome://version
@@ -134,8 +135,21 @@ export function quantileAtValue (median, falloff, value) {
  * @param {object} caps capabilities
  * @param {number} minVersion minimal chrome browser version
  */
-export function isBrowserVersionLower(caps, minVersion) {
-    const versionProps = ['browserVersion', 'browser_version', 'version']
-    const browserVersion = caps[versionProps.find(prop => caps[prop])]
+export function isBrowserVersionLower (caps, minVersion) {
+    const browserVersion = getChromeMajorVersion(caps[VERSION_PROPS.find(prop => caps[prop])])
     return typeof browserVersion === 'number' && browserVersion < minVersion
+}
+
+/**
+ * get chromedriver major version
+ * @param   {string|*}      version chromedriver version like `78.0.3904.11` or just `78`
+ * @return  {number|*}              either major version, ex `78`, or whatever value is passed
+ */
+export function getChromeMajorVersion (version) {
+    let majorVersion = version
+    if (typeof version === 'string') {
+        majorVersion = Number(version.split('.')[0])
+        majorVersion = isNaN(majorVersion) ? version : majorVersion
+    }
+    return majorVersion
 }
--- a/packages/wdio-devtools-service/src/utils.js
+++ b/packages/wdio-devtools-service/src/utils.js
@@ -128,3 +128,14 @@ export function quantileAtValue (median, falloff, value) {
     const standardizedX = (Math.log(value) - location) / (Math.SQRT2 * shape)
     return (1 - internalErf_(standardizedX)) / 2
 }
+
+/**
+ * check if browser version is lower than `minVersion`
+ * @param {object} caps capabilities
+ * @param {number} minVersion minimal chrome browser version
+ */
+export function isBrowserVersionLower(caps, minVersion) {
+    const versionProps = ['browserVersion', 'browser_version', 'version']
+    const browserVersion = caps[versionProps.find(prop => caps[prop])]
+    return typeof browserVersion === 'number' && browserVersion < minVersion
+}
--- a/packages/wdio-devtools-service/src/index.js
+++ b/packages/wdio-devtools-service/src/index.js
@@ -5,7 +5,7 @@ import DevToolsDriver from './driver'
 import Auditor from './auditor'
 import TraceGatherer from './gatherer/trace'
 import DevtoolsGatherer from './gatherer/devtools'
-import { findCDPInterface, getCDPClient } from './utils'
+import { findCDPInterface, getCDPClient, isBrowserVersionLower } from './utils'
 import { NETWORK_STATES, DEFAULT_NETWORK_THROTTLING_STATE } from './constants'
 
 const log = logger('@wdio/devtools-service')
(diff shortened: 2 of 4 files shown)


packages/wdio-devtools-service/src/utils.js
```diff
@@ -128,3 +128,14 @@ export function quantileAtValue (median, falloff, value) {
...
     return (1 - internalErf_(standardizedX)) / 2
 }
+
+/**
+ * check if browser version is lower than `minVersion`
+ * @param {object} caps capabilities
+ * @param {number} minVersion minimal chrome browser version
+ */
+export function isBrowserVersionLower(caps, minVersion) {
+    const versionProps = ['browserVersion', 'browser_version', 'version']
```
christian-bromann: We can move this up to the top of the file?

packages/wdio-devtools-service/src/utils.js
```diff
@@ -128,3 +128,14 @@ export function quantileAtValue (median, falloff, value) {
...
+    const browserVersion = caps[versionProps.find(prop => caps[prop])]
+    return typeof browserVersion === 'number' && browserVersion < minVersion
```
christian-bromann: What if someone has a browser version like `65.1.234.2`?

mgrybyk: Missed that!

packages/wdio-devtools-service/src/utils.js
```diff
@@ -128,3 +129,27 @@ export function quantileAtValue (median, falloff, value) {
...
     return (1 - internalErf_(standardizedX)) / 2
 }
+
+/**
+ * check if browser version is lower than `minVersion`
+ * @param {object} caps capabilities
+ * @param {number} minVersion minimal chrome browser version
+ */
+export function isBrowserVersionLower (caps, minVersion) {
+    const browserVersion = getChromeMajorVersion(caps[VERSION_PROPS.find(prop => caps[prop])])
```
christian-bromann:
```suggestion
    const chromeVersion = caps[VERSION_PROPS.find(prop => caps[prop])]
    
    if (!chromeVersion) {
        log.warn('No Chrome version can be detected in the capabilities')
        return true
    }
    
    const browserVersion = getChromeMajorVersion(chromeVersion)
```

mgrybyk:
> Let's add this check in case no browser version is provided.

There is such case already or I'm missing something

packages/wdio-devtools-service/src/utils.js
mgrybyk: I don't think we need to warn that no browser version is provided, it's not mandatory

christian-bromann: you right

mgrybyk: the function returns false if there is no version provided OR browser version is above 63.

otherwise it returns `true`.

mgrybyk:
argh, missed to rename `browserVersion` to `chromeVersion`.

Shall I raise a PR for it or let's just keep it as is?

christian-bromann:
I think it is fine. I hope the service can be used by other browsers soon too.

Issue #4289, opened 2019-08-01
Ability to click with particular button, ex right mouse button


mgrybyk:
**Is your feature request related to a problem? Please describe.**
I want to be able to click on element with right, middle or any other buttons

**Describe the solution you'd like**
Click with desired mouse button on element with `performActions`
Example:
`$('selector').click({ button: 'right' })` same as `$('selector').click({ button: 1 })`

There should be no changes to `$('selector').click()`, it should use `elementClick` as before.

**Describe alternatives you've considered**
Do not modify click, add new command to click with perform actions.

**Additional context**

lexeek:
I hope this feature will be implemented soon.

christian-bromann:
@lexeek please don't wait until it happens and rather help us getting in implemented ;-)

caoxu2000:
@christian-bromann I can tackle this but can't promise anything yet.  Do you have any insight on how this should be implemented?  2 ways at high level I can think of: 1) call protocol api directly 2) just call performActions command.  Please let me know your thoughts when you get a chance.

caoxu2000:
Here is all the research I did from v4 & v5:
in v4 https://github.com/webdriverio-boneyard/v4/blob/master/lib/commands/rightClick.js
it uses handleMouseButtonCommand.js below
https://github.com/webdriverio-boneyard/v4/blob/master/lib/helpers/handleMouseButtonCommand.js
but there is protocol js:
https://github.com/webdriverio-boneyard/v4/blob/master/lib/helpers/handleMouseButtonProtocol.js
not sure which one to use for v5 if we need this helper

in v5 for example this doubleClick uses performActions:
https://github.com/webdriverio/webdriverio/blob/e7770403ed1ed41e976b3e04ef7325ea878328e5/packages/webdriverio/src/commands/element/doubleClick.js

christian-bromann:
@caoxu2000 you have to use performAction is the `isW3C` flag is set to `true` otherwise you can use the logic that you've found in v4.

jprateek:
is there any plan to implement this soon?

mgrybyk:
@jprateek @lexeek there are plans to do it, will try to provide with some details soon.

Until that you may just add custom command

christian-bromann:
This landed in #4592 and will be released next.

lexeek:
Thank you for implementation! Despite we have this feature implemented on our own we are really happy with these improvements!

PR #4592, opened 2019-10-09
Webdriverio expand click command

erwinheitzman:
## Proposed changes

Implementation of #4289 which expands the click command.

## Types of changes

- [x] Documentation update

## Checklist

## Further comments

This change shouldn't affect anyone.

### DIFF
--- a/packages/webdriverio/src/commands/element/click.js
+++ b/packages/webdriverio/src/commands/element/click.js
@@ -48,7 +48,7 @@
     })
     it('should demonstrate a right click passed as number', () => {
         const myButton = $('#myButton')
-        myButton.click({ button: 'right' }) // opens the contextmenu at the location of the button
+        myButton.click({ button: 2 }) // opens the contextmenu at the location of the button
     })
  * </example>
  *
@@ -77,7 +77,7 @@ export default async function click (options) {
         return this.elementClick(this.elementId)
     }
 
-    if (button === 0 || button === 1 || button === 2) {
+    if ([0, 1, 2].includes(button)) {
         if (this.isW3C) {
             await this.performActions([{
                 type: 'pointer',
--- a/packages/webdriverio/src/commands/element/click.js
+++ b/packages/webdriverio/src/commands/element/click.js
@@ -42,6 +42,35 @@
  *
  */
 
-export default function click () {
-    return this.elementClick(this.elementId)
+export default async function click (options) {
+    const { button } = options || {}
+    const isLeftButton = button === 'left' || button === 0 || !button
+    const isMiddleButton = button === 'middle' || button === 1
+    const isRightButton = button === 'right' || button === 2
+
+    if (isLeftButton) {
+        return this.elementClick(this.elementId)
+    }
+
+    if (isRightButton || isMiddleButton) {
+        if (this.isW3C) {
+            await this.performActions([{
+                type: 'pointer',
+                id: 'pointer1',
+                parameters: { pointerType: 'mouse' },
+                actions: [
+                    { type: 'pointerMove', origin: this, x: 0, y: 0 },
+                    { type: 'pointerDown', button: isRightButton ? 2 : 1 },
+                    { type: 'pointerUp', button: isRightButton ? 2 : 1 }
+                ]
+            }])
+
+            return this.releaseActions()
+        }
+
+        await this.moveTo()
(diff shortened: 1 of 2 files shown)


packages/webdriverio/src/commands/element/click.js
```diff
@@ -37,11 +37,40 @@
...
 
-export default function click () {
-    return this.elementClick(this.elementId)
+export default async function click (options) {
+    const { button } = options || {}
+    const isLeftButton = button === 'left' || button === 0 || !button
+    const isMiddleButton = button === 'middle' || button === 1
+    const isRightButton = button === 'right' || button === 2
+
+    if (isLeftButton) {
```
mgrybyk: I'd use perform actions if user passed button in any case. There might be a case when element is not clickable and perform actions will do the trick

packages/webdriverio/src/commands/element/click.js
```diff
@@ -37,11 +37,40 @@
...
+        return this.elementClick(this.elementId)
+    }
+
+    if (isRightButton || isMiddleButton) {
+        if (this.isW3C) {
```
mgrybyk: Will you be able to add devtools protocol support or you want to deal with it with another ticket?

packages/webdriverio/src/commands/element/click.js
christian-bromann: so we don't want to use the `elementClick` command at all?

packages/webdriverio/src/commands/element/click.js
erwinheitzman: Good point, since I will be leaving for my vacation tomorrow I think it's best to do it in a separate ticket :)

packages/webdriverio/src/commands/element/click.js
erwinheitzman: Alright, I can see that being useful. I'll push the change when I can.

packages/webdriverio/src/commands/element/click.js
mgrybyk: I'll raise a ticket, have a great vacation

packages/webdriverio/src/commands/element/click.js
christian-bromann: makes sense to me

packages/webdriverio/src/commands/element/click.js
```diff
@@ -37,11 +37,40 @@
...
+            await this.performActions([{
+                type: 'pointer',
+                id: 'pointer1',
+                parameters: { pointerType: 'mouse' },
+                actions: [
+                    { type: 'pointerMove', origin: this, x: 0, y: 0 },
+                    { type: 'pointerDown', button: isRightButton ? 2 : 1 },
```
christian-bromann: Should we have a bit pause in here? I've seen action commands that use pause to simulate the click interaction

erwinheitzman: This is only used in a doubleClick in between the two clicks so I don't think it is needed. I tested this in Chrome 77 and all buttons function as expected.

packages/webdriverio/src/commands/element/click.js
```diff
@@ -37,11 +37,49 @@
...
+
+    if (button === 'right') {
+        button = 2
+    }
+
+    if (typeof button === 'undefined') {
+        return this.elementClick(this.elementId)
+    }
+
+    if (button === 0 || button === 1 || button === 2) {
```
mgrybyk: maybe `[0, 1, 2].includes(button)`? Just a proposal, not sure if it's better

packages/webdriverio/src/commands/element/click.js
```diff
@@ -36,12 +36,66 @@
...
+    :example.html
+    <button id="myButton">Click me</button>
+    :example.js
+    it('should demonstrate a right click passed as string', () => {
+        const myButton = $('#myButton')
+        myButton.click({ button: 'right' }) // opens the contextmenu at the location of the button
+    })
+    it('should demonstrate a right click passed as number', () => {
+        const myButton = $('#myButton')
+        myButton.click({ button: 'right' }) // opens the contextmenu at the location of the button
```
christian-bromann:
```suggestion
        myButton.click({ button: 2 }) // opens the contextmenu at the location of the button
```

Issue #4578, opened 2019-10-06
Would be great to add a custom locator strategy


gemelgb:
**Is your feature request related to a problem? Please describe.**
I m using a UI framework that generates a virtual dom and I want to be able to add a new locator or custom strategy to achieve the retrieval of the dom elements based on the virtual dom properties and attributes.

**Describe the solution you'd like**
More or less an implementation exists for protractor called addLocator, something similiar for WDIO would be awesome.

**Describe alternatives you've considered**
Custom commands are limited and does not help me to achieve this. More or less would be possible if I can use the standard WDJS locator by.js. The WDIO provides an alternative to use js functions but I cannot pass any parameters. On the other hand if I use browser.execute still seem not to work because it cannot return an element which I can use especially for an Array of elements. Any alternative that i might not known?

mgrybyk:
Can you please be more specific and provide with some detailed examples?

christian-bromann:
> Any alternative that i might not known?

You can always create an element object by having a reference of it, e.g.

```js
const elemRef = browser.execute(() => ...)
const elem = $(elemRef)
```

> More or less an implementation exists for protractor called addLocator

I am not aware what Protractor is doing. Can you please explain?

CrispusDH:
Here documentation from source:
```js
  /**
   * Add a locator to this instance of ProtractorBy. This locator can then be
   * used with element(by.locatorName(args)).
   *
   * @view
   * <button ng-click="doAddition()">Go!</button>
   *
   * @example
   * // Add the custom locator.
   * by.addLocator('buttonTextSimple',
   *     function(buttonText, opt_parentElement, opt_rootSelector) {
   *   // This function will be serialized as a string and will execute in the
   *   // browser. The first argument is the text for the button. The second
   *   // argument is the parent element, if any.
   *   var using = opt_parentElement || document,
   *       buttons = using.querySelectorAll('button');
   *
   *   // Return an array of buttons with the text.
   *   return Array.prototype.filter.call(buttons, function(button) {
   *     return button.textContent === buttonText;
   *   });
   * });
   *
   * // Use the custom locator.
   * await element(by.buttonTextSimple('Go!')).click();
   *
   * @alias by.addLocator(locatorName, functionOrScript)
   * @param {string} name The name of the new locator.
   * @param {Function|string} script A script to be run in the context of
   *     the browser. This script will be passed an array of arguments
   *     that contains any args passed into the locator followed by the
[...]

PR #4603, opened 2019-10-10
custom locator strategy

baruchvlz:
## Proposed changes
Related issue: #4578

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/packages/webdriverio/src/commands/browser/custom$$.js
+++ b/packages/webdriverio/src/commands/browser/custom$$.js
@@ -32,6 +32,7 @@ async function custom$$ (strategyName, strategyArgument) {
     }
 
     let res = await this.execute(strategy, strategyArgument)
+
     if (!Array.isArray(res)) {
         res = [res]
     }
--- a/packages/webdriverio/src/commands/browser/custom$$.js
+++ b/packages/webdriverio/src/commands/browser/custom$$.js
@@ -3,9 +3,9 @@
  * The `customs$` allows you to use a custom strategy declared by using `browser.addLocatorStrategy`
  *
  * <example>
-    :pause.js
+    :example.js
     it('should get all the plugin wrapper buttons', () => {
-        browser.url('https://webdriver.io');
+        browser.url('https://webdriver.io')
         browser.addLocatorStrategy('myStrat', (selector) => {
             return document.querySelectorAll(selector)
         })
@@ -13,7 +13,7 @@
         const pluginWrapper = browser.custom$$('myStrat', '.pluginWrapper')
 
         console.log(pluginWrapper.length) // 4
-    });
+    })
  * </example>
  *
  * @alias custom$$
--- a/packages/webdriverio/src/commands/browser/custom$$.js
+++ b/packages/webdriverio/src/commands/browser/custom$$.js
@@ -6,11 +6,11 @@
     :pause.js
     it('should get all the plugin wrapper buttons', () => {
         browser.url('https://webdriver.io');
-        browser.useLocatorStrategy('myStrat', (selector) => {
+        browser.addLocatorStrategy('myStrat', (selector) => {
             return document.querySelectorAll(selector)
         })
 
-        const pluginWrapper = browser.custom$$('.pluginWrapper')
+        const pluginWrapper = browser.custom$$('myStrat', '.pluginWrapper')
 
         console.log(pluginWrapper.length) // 4
     });
@@ -30,7 +30,11 @@ async function custom$$ (strategyName, strategyArgument) {
     if (!strategy) {
         throw Error('No strategy found for ' + strategyName)
     }
-    const res = await this.execute(strategy, strategyArgument)
+
+    let res = await this.execute(strategy, strategyArgument)
+    if (!Array.isArray(res)) {
+        res = [res]
+    }
 
     return await getElements.call(this, strategy, res)
 }
--- a/packages/webdriverio/src/commands/browser/custom$$.js
+++ b/packages/webdriverio/src/commands/browser/custom$$.js
@@ -22,6 +22,7 @@
  * @return {Element}
  */
 import { getElements } from '../../utils/getElementObject'
+import { ELEMENT_KEY } from '../../constants'
 
 async function custom$$ (strategyName, strategyArgument) {
     const strategy = this.strategies.get(strategyName)
@@ -32,11 +33,18 @@ async function custom$$ (strategyName, strategyArgument) {
 
     let res = await this.execute(strategy, strategyArgument)
 
+    /**
+     * if the user's script return just one element
+     * then we convert it to an array as this method
+     * should return multiple elements
+     */
     if (!Array.isArray(res)) {
         res = [res]
     }
 
-    return await getElements.call(this, strategy, res)
+    res = res.filter(el => !!el && typeof el[ELEMENT_KEY] === 'string')
+
+    return res.length ? await getElements.call(this, strategy, res) : []
 }
 
 export default custom$$
--- a/packages/webdriverio/src/commands/browser/custom$$.js
+++ b/packages/webdriverio/src/commands/browser/custom$$.js
@@ -1,0 +1,38 @@
+/**
+ *
+ * The `customs$` allows you to use a custom strategy declared by using `browser.addLocatorStrategy`
+ *
+ * <example>
+    :pause.js
+    it('should get all the plugin wrapper buttons', () => {
+        browser.url('https://webdriver.io');
+        browser.useLocatorStrategy('myStrat', (selector) => {
+            return document.querySelectorAll(selector)
+        })
+
+        const pluginWrapper = browser.custom$$('.pluginWrapper')
+
+        console.log(pluginWrapper.length) // 4
+    });
+ * </example>
+ *
+ * @alias custom$
+ * @param {String} strategyName
+ * @param {Any} strategyArguments
+ * @return {Element}
+ */
+import { runFnInFiberContext } from '@wdio/utils/build/shim'
+import { getElements } from '../../utils/getElementObject'
+
+async function custom$$ (strategyName, strategyArgument) {
+    const strategy = this.strategies.get(strategyName)
+
+    if (!strategy) {
+        throw Error('No strategy found for ' + strategyName)
+    }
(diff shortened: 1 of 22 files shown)


mgrybyk:
How should it work in the end, can you please provide with examples?

baruchvlz:
@mgrybyk 

@christian-bromann and I had a conversation about how to approach this. Essentially this is what we came up with.

The user will have to register a locator somewhere in their tests like so

```ts
browser.addLocatorStrategy(name: string, callback: (args: any[]) => HTMLElement))
```

When they want to use this strategy, they can call it as such:

```ts
browser.custom$(name: string, strategyArgs: any[]) => WDIOElement
```

Any suggestions / concerns let me know.

CrispusDH:
@baruchvlz , I think might be better to create two methods:
```js
browser.custom$(name: string, strategyArgs: any[]) => WDIOElement
browser.custom$$(name: string, strategyArgs: any[]) => WDIOElement[]
```
in such case it would be convenient to work with types and be predictable what you expect to get.
What do you think?

baruchvlz:
@CrispusDH Yea that makes sense, I'll add the commands.

baruchvlz:
@CrispusDH 
Sorry, on second thought, having both `custom$` and `custom$$` might be confusing since you have to add the strategies before you use them. It would be better to consider renaming the command to some more explicit like`useLocatorStrategy`, this way its easier to understand while writing tests, and we always return one element since it depends on the script passed by the user.

Alternatively, we could add an extra flag to the command for selecting multiple elements.

```ts
useLocatorStrategy<T = {}>(name: string, strategyArgument: string | T, multiple: boolean) => WDIOElement | WDIOElement[]
```

Adding a dynamic type will allow the user to pass multiple arguments to their script if they require extra logic.

I think `useLocatorStrategy` is a good name and its intuitive since the user will attach strategies via `browser.addLocatorStrategy`.

christian-bromann:
> on second thought, having both `custom$` and `custom$$` might be confusing since you have to add the strategies before you use them.

In what way is this confusing?

> It would be better to consider renaming the command to something more explicit like `useLocatorStrategy`

My concern is that it is not aligned with other command names such as `react$/$$` or `shadow$/$$`

> and we always return one element since it depends on the script passed by the user.

This is not ensured by the user script. If for whatever reason the script fetches multiple elements, multiple elements will be returned.

baruchvlz:
There is no real difference between `custom$/$$` and `useLocatorStrategy` and you have a good point on keeping the methods consistent. I'll add the methods to both browser and element objects.

packages/webdriverio/src/commands/browser/custom$$.js
```diff
@@ -0,0 +1,43 @@
...
+        throw Error('No strategy found for ' + strategyName)
+    }
+
+    let res = await this.execute(strategy, strategyArgument)
+
+    if (!Array.isArray(res)) {
+        res = [res]
+    }
+
+    return await getElements.call(this, strategy, res)
```
christian-bromann:
```suggestion
    return await getElements.call(this, strategy, res.filter((el) => Boolean(el) && typeof el[ELEMENT_ID] === 'string'))
```

Let's add a check to ensure that the returned object has an actual element reference.

packages/webdriverio/src/commands/browser/custom$.js
```diff
@@ -0,0 +1,42 @@
...
+
+async function custom$ (strategyName, strategyArguments) {
+    const strategy = this.strategies.get(strategyName)
+
+    if (!strategy) {
+        throw Error('No strategy found for ' + strategyName)
+    }
+
+    let res = await this.execute(strategy, strategyArguments)
+
```
christian-bromann: same here

packages/webdriverio/src/commands/browser/custom$$.js
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export default runFnInFiberContext(custom$$)
```
christian-bromann: This is not necessary. We don't do that with any command.

packages/webdriverio/src/commands/element/custom$$.js
```diff
@@ -0,0 +1,46 @@
...
+ * @alias custom$$
+ * @param {String} strategyName
+ * @param {Any} strategyArguments
+ * @return {Element}
+ */
+import { runFnInFiberContext } from '@wdio/utils/build/shim'
+import { getElements } from '../../utils/getElementObject'
+import { getBrowserObject } from '../../utils'
+
+async function custom$$ (strategyName, strategyArgument) {
```
christian-bromann: since it is the same implementation you can just require it from the browser directory, no?

baruchvlz: @christian-bromann It's missing the `parent` element as in the `element.custom$` function. Will fix.

packages/webdriverio/src/commands/browser/custom$$.js
baruchvlz: @christian-bromann I wasn't able to run this in sync mode without the fiber context wrapper. Will look for another way.

packages/webdriverio/src/commands/browser/custom$$.js
baruchvlz: @christian-bromann  Not entirely sure what the goal of this is. Wouldn't getElements error if the res doesn't contain a WDIO element? If not, should `custom$/$$` throw the error?

packages/webdriverio/src/commands/browser/custom$$.js
```diff
@@ -0,0 +1,50 @@
+/**
+ *
+ * The `customs$` allows you to use a custom strategy declared by using `browser.addLocatorStrategy`
+ *
+ * <example>
+    :pause.js
```
christian-bromann:
```suggestion
    :example.js
```

packages/webdriverio/src/commands/browser/custom$$.js
```diff
@@ -0,0 +1,50 @@
...
+    it('should get all the plugin wrapper buttons', () => {
+        browser.url('https://webdriver.io');
+        browser.addLocatorStrategy('myStrat', (selector) => {
+            return document.querySelectorAll(selector)
+        })
+
+        const pluginWrapper = browser.custom$$('myStrat', '.pluginWrapper')
+
+        console.log(pluginWrapper.length) // 4
+    });
```
christian-bromann:
```suggestion
    })
```

packages/webdriverio/src/commands/browser/custom$$.js
```diff
@@ -0,0 +1,50 @@
+/**
+ *
+ * The `customs$` allows you to use a custom strategy declared by using `browser.addLocatorStrategy`
+ *
+ * <example>
+    :pause.js
+    it('should get all the plugin wrapper buttons', () => {
+        browser.url('https://webdriver.io');
```
christian-bromann:
```suggestion
        browser.url('https://webdriver.io')
```

packages/webdriverio/src/commands/browser/custom$.js
```diff
@@ -0,0 +1,52 @@
+/**
+ *
+ * The `custom$` allows you to use a custom strategy declared by using `browser.addLocatorStrategy`
+ *
+ * <example>
+    :pause.js
```
christian-bromann:
```suggestion
    :example.js
```

packages/webdriverio/src/commands/browser/custom$.js
```diff
@@ -0,0 +1,52 @@
...
+    it('should fetch the project title', () => {
+        browser.url('https://webdriver.io');
```
christian-bromann:
```suggestion
        browser.url('https://webdriver.io')
```

packages/webdriverio/src/commands/element/custom$$.js
```diff
@@ -0,0 +1,67 @@
+/**
+ *
+ * The `customs$` allows you to use a custom strategy declared by using `browser.addLocatorStrategy`
+ *
+ * <example>
+    :pause.js
```
christian-bromann:
```suggestion
    :example.js
```

packages/webdriverio/src/commands/element/custom$$.js
```diff
@@ -0,0 +1,67 @@
...
+    it('should get all the plugin wrapper buttons', () => {
+        browser.url('https://webdriver.io');
```
christian-bromann:
```suggestion
        browser.url('https://webdriver.io')
```

packages/webdriverio/src/commands/element/custom$$.js
```diff
@@ -0,0 +1,67 @@
...
+        browser.addLocatorStrategy('myStrat', (selector) => {
+            return document.querySelectorAll(selector)
+        })
+
+        const pluginRowBlock = browser.custom$('myStrat', '.pluginRowBlock')
+        const pluginWrapper = pluginRowBlock.custom$$('myStrat', '.pluginWrapper')
+
+        console.log(pluginWrapper.length) // 4
+    });
```
christian-bromann:
```suggestion
    })
```

packages/webdriverio/src/index.js
```diff
@@ -61,6 +61,14 @@ export const remote = async function (params = {}, remoteModifier) {
         )
     }
 
+    instance.addLocatorStrategy = (name, script) => {
```
christian-bromann: How about we move this function into a utility as follows:
```js
instance.addLocatorStrategy = addLocatorStrategyHandler(instance)
```

the utility then looks like

```js
export function addLocatorStrategyHandler (scope) {
    if (scope.strategies.get(name)) {
        throw new Error(`Strategy ${name} already exists`)
    }

    scope.strategies.set(name, script)
}
```

baruchvlz: Wouldn't it be more like this?

```js
export function addLocatorStrategyHandler (scope) {
    return (name, script) => {
        if (scope.strategies.get(name)) {
            throw new Error(`Strategy ${name} already exists`)
        }

        scope.strategies.set(name, script)
    }
}
```

christian-bromann: yes

christian-bromann:
@baruchvlz can you rebase?

PR #4609, opened 2019-10-11
reset _NOT_FIBER flag on Timer timeout

### LINKED ISSUE #4599: isExisting returns object instead of boolean
 - **WebdriverIO version:** 5.14.5
 - **Mode:** WDIO Testrunner
 - **Node.js version:** v11.15.0
 - **NPM version:** v6.7.0
 - **Browser name and version:** 77.0.3865.90 (Official Build) (64-bit)
 - **Platform name and version:** MacOS 10.14.6
```
@wdio/cli
@wdio/local-runner
@wdio/mocha-framework
@wdio/spec-reporter
@wdio/sync
wdio-chromedriver-service
```

**Config of WebdriverIO**
See repro

**Describe the bug**
In certain circumstances, `element.isExisting()` returns an object instead of a boolean value.

**To Reproduce**
```
git clone git@github.com:diachedelic/wdio-repros.git &&
  cd wdio-repros &&
  git checkout zombie-existing &&
  npm i &&
  npm run test

```

**Log**
See `logs/` directory in repro

mgrybyk:
## Proposed changes

fixes #4599

## Types of changes

## Checklist

## Further comments

`_NOT_FIBER` should be somehow reseted on timeout in Timer. Have no idea how to deal with it better

### DIFF
--- a/packages/wdio-sync/src/wrapCommand.js
+++ b/packages/wdio-sync/src/wrapCommand.js
@@ -7,6 +7,27 @@ import { sanitizeErrorMessage } from './utils'
 const log = logger('@wdio/sync')
 let inCommandHook = false
 
+const timers = []
+const elements = new Set()
+
+/**
+ * resets `_NOT_FIBER` if Timer has timed out
+ */
+process.on('WDIO_TIMER', (payload) => {
+    if (payload.start) {
+        return timers.push(payload.id)
+    }
+    if (timers.includes(payload.id)) {
+        while (timers.pop() !== payload.id);
+    }
+    if (payload.timeout) {
+        elements.forEach(element => { delete element._NOT_FIBER })
+    }
+    if (timers.length === 0) {
+        elements.clear()
+    }
+})
+
 /**
  * wraps a function into a Fiber ready context to enable sync execution and hooks
  * @param  {Function}   fn             function to be executed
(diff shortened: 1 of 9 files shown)


mgrybyk:
tests are failing because of intermittent failure in devtools e2e tests

mgrybyk:
ok, test are passing, want to re-run few times to make sure it's not my changes are breaking things

christian-bromann:
@mgrybyk can you explain a bit what problem you have discovered and your solution? This looks like a more complex change that I would like to have more context on.

mgrybyk:
The problem
```

const parent = $('body')
const child = parent.$('section')
const elem = child.$('a#secondPageLink')
elem.isDisplayed()
browser.execute(function (el) { el.remove() }, elem)

try {
    browser.waitUntil(() => {
        elem.click()
        return true
    }, 2000)
} catch (err) { }

console.log(browser._NOT_FIBER, parent._NOT_FIBER, child._NOT_FIBER, elem._NOT_FIBER)
```

currently output of that code is `false false false true`. One of `_NOT_FIBER` flags is still `true`, because timeout error has happened in Timer and aborted execution.
As a result `wrapCommand` hasn't finished properly and `_NOT_FIBER` is still `true`.

I'll describe in next comment how am I dealing with it

mgrybyk:
We need to reset `_NOT_FIBER` flag of every element we were interacting with in `waitUntil` if timeout occurred in Timer.
I decided that the best solution would be to send some event saying to clear `_NOT_FIBER` flag.
To do that I need to know which elements are possibly affected, so I'm adding every element to `elements` Set when Timer has started (and clearing the Set when Timer has finished).

mgrybyk:
alternative solution is to somehow get rid of `_NOT_FIBER` at all.
Trying to do that time to time but still have no working solution

Issue #4292, opened 2019-08-02
Add unit tests for @wdio/sync - executeAsync


mgrybyk:
**Is your feature request related to a problem? Please describe.**
I'd like to add tests for @wdio/sync, these files in particular:
- `executeAsync` in `index.js`

**Describe the solution you'd like**
Currently most of the functions are marked with `/* istanbul ignore next */`. 
We have to remove that and test all that functions.

MatthewBurstein:
Do the tests for executeAsync still need adding? I'd be happy to pick this up :)

mgrybyk:
gonna be fixed with https://github.com/webdriverio/webdriverio/pull/4612

@MatthewBurstein thank you

PR #4612, opened 2019-10-13
Wdio sync executeAsync tests

MatthewBurstein:
## Proposed changes

Add tests for executeAsync in @wdio/sync as per [this issue](https://github.com/webdriverio/webdriverio/issues/4292)

Additional slight refactor of executeAsynx to fix a bug.

## Types of changes

- [x] Documentation update

## Checklist

## Further comments
Added tests for the executeAsync function. In doing so, I realised that if the hook returns a  promise which rejects, the stack filtering functionality was unreachable. I had a go at refactoring this, and I _think_ it should be fine, but there might well be implications to this that I have missed/don't know about. Let me know what you think.

Also fixed one or two typos in comments etc.

### DIFF
--- a/packages/wdio-sync/src/index.js
+++ b/packages/wdio-sync/src/index.js
@@ -11,7 +11,7 @@ import { STACKTRACE_FILTER_FN } from './constants'
  *
  * @param  {Function} fn         spec or hook method
  * @param  {Number}   repeatTest number of retries
- * @param  {Array}    args       number of retries
+ * @param  {Array}    args       arguments passed to hook
  * @return {Promise}             that gets resolved once test/hook is done or was retried enough
  */
 const executeSync = async function (fn, repeatTest = 0, args = []) {
@@ -53,14 +53,15 @@ const executeSync = async function (fn, repeatTest = 0, args = []) {
  *
  * @param  {Function} fn         spec or hook method
  * @param  {Number}   repeatTest number of retries
+ * @param  {Array}    args       arguments passed to hook
  * @return {Promise}             that gets resolved once test/hook is done or was retried enough
  */
-/* istanbul ignore next */
+
 const executeAsync = function (fn, repeatTest = 0, args = []) {
     let result, error
 
     /**
-     * if a new hook gets executed we can assume that all commands should have finised
+     * if a new hook gets executed we can assume that all commands should have finished
      * with exception of timeouts where `commandIsRunning` will never be reset but here
      */
     // commandIsRunning = false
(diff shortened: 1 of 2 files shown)


packages/wdio-sync/src/index.js
```diff
@@ -83,23 +84,23 @@ const executeAsync = function (fn, repeatTest = 0, args = []) {
...
     /**
-     * handle promise response
+     * if we don't retry just return result
      */
-    return result.catch((e) => {
-        if (repeatTest) {
-            return executeAsync(fn, --repeatTest, args)
-        }
-
-        e.stack = e.stack.split('\n').filter(STACKTRACE_FILTER_FN).join('\n')
```
MatthewBurstein: Previously I think this was unreachable because if `repeatTest` is falsey then it must be 0 which would be handled in the previous if statement.

packages/wdio-sync/src/index.js
```diff
@@ -83,23 +84,23 @@ const executeAsync = function (fn, repeatTest = 0, args = []) {
...
-        return Promise.reject(e)
-    })
+    return new Promise(resolve => resolve(result))
```
baruchvlz: Is it possible to make the function `async` instead of using the Promise object? Then you can just `return result` and for the rejection throw an error.

packages/wdio-sync/src/index.js
```diff
@@ -83,23 +84,23 @@ const executeAsync = function (fn, repeatTest = 0, args = []) {
     }
 
     /**
-     * if we don't retry just return result
+     * handle promise response
      */
-    if (repeatTest === 0 || !result || typeof result.catch !== 'function') {
-        return new Promise(resolve => resolve(result))
+    if (result && typeof result.catch === 'function') {
```
mgrybyk: let's change to `if (result > 0 && typeof result.catch === 'function') {`

MatthewBurstein: Not sure I follow sorry. `result` here is the value returned from the hook so is probably not a number? We check it exists to prevent `typeof result.catch` throwing in error if `result` is `undefined`, `null` etc. Could you clarify?

packages/wdio-sync/src/index.js
MatthewBurstein: This actually simplifies it a lot, the only problem is it makes it hard to differentiate between the two error cases, one when `fn.apply(this, args)` throws an error, and one when it returns a promise which rejects. At the moment we only filter the stack trace in the second case. 

Is it ok to filter it in both cases? Or is there a way to tell the two cases apart in the catch block?

packages/wdio-sync/src/index.js
mgrybyk: Damn my fault, though about repeatTest

packages/wdio-sync/src/index.js
mgrybyk: Yes, sounds good.
I think executeSync works like this, it was changed to async function recently

Issue #4615, opened 2019-10-13
Avoid starting some services in worker processes


mgrybyk:
**Is your feature request related to a problem? Please describe.**
Currently we load all the services in child processes that leads to increasing memory consumption.

**Describe the solution you'd like**
Avoid starting selenium-standalone and appium services in worker processes

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

christian-bromann:
Let's remove all services from the services list if they don't have any class for the worker.

mgrybyk:
@christian-bromann I'll remove services according to

| Service  | Load in worker |
| ------------- | ------------- |
| chromedriver  | no  |
| selenium-standalone  | no  |
| appium  | no  |
| reportportal  | no |
| firefox-profile  | no  |
| sauce  | **yes**  |
| testingbot  | **yes** |
| browserstack  | **yes** |
| crossbrowsertesting  | **yes** |
| docker  | **yes** |
| zafira-listener  | **yes** |
| devtools  | **yes** |
| applitools  | **yes**  |

PR #4623, opened 2019-10-14
Filter services that should not run in worker process

mgrybyk:
## Proposed changes

avoid starting `'chromedriver', 'selenium-standalone', 'appium', 'reportportal', 'firefox-profile'` services in worker process

#4615 

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/packages/wdio-config/src/constants.js
+++ b/packages/wdio-config/src/constants.js
@@ -74,3 +74,10 @@ export const SUPPORTED_HOOKS = [
     'beforeFeature', 'beforeScenario', 'beforeStep', 'afterStep', 'afterScenario', 'afterFeature',
     'onReload', 'onPrepare', 'onComplete'
 ]
+
+/**
+ * these services should not be started in worker process
+ */
+export const NON_WORKER_SERVICES = [
+    'chromedriver', 'selenium-standalone', 'appium', 'reportportal', 'firefox-profile'
+]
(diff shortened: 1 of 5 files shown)


packages/wdio-config/src/constants.js
```diff
@@ -74,3 +74,10 @@ export const SUPPORTED_HOOKS = [
     'beforeFeature', 'beforeScenario', 'beforeStep', 'afterStep', 'afterScenario', 'afterFeature',
     'onReload', 'onPrepare', 'onComplete'
 ]
+
+/**
+ * these services should not be started in worker process
+ */
+export const NON_WORKER_SERVICES = [
```
christian-bromann: Instead of having to maintain a list like that can we just check after we required the launcher whether the service also offers a worker service?

mgrybyk: Not sure how to achieve that. I mean we can check everything after requiring the launcher, however the goal is to avoid importing services so that they won't load any modules, as a result I can't import launcher. 

Maybe in v6 we can change the way how services are loaded?

christian-bromann: > Maybe in v6 we can change the way how services are loaded?

Yeah. I guess what I suggested is just more of a hacky workaround. Let's use v6 to introduce a clear way to determine whether or not a service has to be also get required in a worker process.

Issue #4478, opened 2019-09-10
@wdio/cucumber-framework: Issues with multiple tables in Scenarios


nextlevelbeard:
**Environment:**
 - **WebdriverIO version:** 5.15.0

**Config of WebdriverIO**
Any + steps to reproduce

~**Issue 1**: Cannot use row data in the Scenario name when the Scenario has multiple tables. Only the first example is printed by the spec reporter and an error is thrown, test/Scenario still runs.~ SOLVED
**Issue 2**: Trying to execute a Scenario that has multiple tables when the Examples section has tags will throw an error if that tag is used in a tagged hook.

**To Reproduce Issue 1**: 
feature file:
```
@tag
Feature: Foo
  foobar 1
  foobar 2
  foobar 3

  Background:
    Given stub

  @tag
  Scenario Outline: Logged in with user <userId>
    Given stub
    Then User <userId> with <password> is logged in
      Examples:
        | userId    | password |
        | 345       | pw3      |

      Examples:
        | userId    | password |
        | 456       | pw4      |
```
Steps:
```js
Given(/^User (.*) with (.*) is logged in$/, { wrapperOptions: { retry: 2 } }, function (userId, password) {
    console.log(userId, password)
})

Given('stub', function () {
    console.log('stub')
})
```

**To Reproduce Issue 2**: 
feature file:
```
@tag
Feature: Foo
    foobar 1
    foobar 2
    foobar 3

  Backgroud:
      Given stub

  @tag
  Scenario Outline: Bar
    Given stub
    Then User <userId> with <password> is logged in
      Examples:
        | userId    | password |
        | 345       | pw3      |

      @tag2
      Examples:
        | userId    | password |
        | 456       | pw4      |
```
Steps:
```js
const { Given, Before } = require("cucumber");

// The tag here is important, to reproduce the issue, it has to be an used tag
Before("@tag2", function() {
  this.user = 'foobar';
});

Given(/^User (.*) with (.*) is logged in$/, { wrapperOptions: { retry: 2 } }, function (userId, password) {
    console.log(userId, password)
})

Given('stub', function () {
    console.log('stub')
})
```

**Expected behavior**
The tests should pass and not error, the spec reporter should present both Scenarios.

**Log**
**Issue 1 (console)**:
```js
0-0 worker error {
  name: 'TypeError',
  message: "Cannot read property 'complete' of undefined",
  stack: "TypeError: Cannot read property 'complete' of undefined\n    at " +
[...]

PR #4629, opened 2019-10-15
Fix tagged hooks in Cucumber

mgrybyk:
## Proposed changes

Fix tagged hooks in Cucumber

fixes issue 2 in #4478

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

Test data and result are here https://github.com/webdriverio/webdriverio/issues/4478#issuecomment-542228835

### DIFF
--- a/packages/wdio-cucumber-framework/src/utils.js
+++ b/packages/wdio-cucumber-framework/src/utils.js
@@ -147,38 +147,6 @@ export function compareScenarioLineWithSourceLine(scenario, sourceLocation) {
     return scenario.location.line === sourceLocation.line
 }
 
-export function getStepFromFeature(feature, pickle, stepIndex, sourceLocation) {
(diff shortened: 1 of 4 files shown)


packages/wdio-cucumber-framework/src/utils.js
```diff
@@ -147,38 +147,6 @@ export function compareScenarioLineWithSourceLine(scenario, sourceLocation) {
     return scenario.location.line === sourceLocation.line
 }
 
-export function getStepFromFeature(feature, pickle, stepIndex, sourceLocation) {
```
mgrybyk: had to remove this way of getting steps, it didn't at all in some cases.

We should rely on steps received in `testCasePreparedEvent` instead of relying on feature and mixing up steps in a hacky way. Didn't know this before

PR #4656, opened 2019-10-18
Check if element is clickable

### LINKED ISSUE #4646: isDisplayedInViewport returns true for non existing element or for hidden element
 - **WebdriverIO version:**5.15.1
 - **Mode:** wdio
 - **Node.js version:** v10.16.0
 - **NPM version:** v.6.9.0
 - **Browser name and version:** chrome 77
 - **Platform name and version:** mac/windows

**Describe the bug**
isDisplayedInViewport() method returns true for non existing element in DOM or for hidden element

**To Reproduce**
~**Scenario 1**~ should be fixed with https://github.com/webdriverio/webdriverio/pull/4647
``` 
browser.url('https://www.google.com/');
console.log($('.hello-world').isDisplayedInViewport()); // returns true
```

~*Expected behavior*~
~Should return false as element does not exist and not displayed~

**Scenario 2**
When element in DOM, but hidden under dropdown. element.isDisplayed() returns true, but element.isDisplayedInViewport() should return false as element not in viewport

mgrybyk:
## Proposed changes

Add `isClickable` command to element. Works like this:
1. check if element exist
2. check if element is visible
3. check if element is within viewport (if no - scroll to the element)
4. check if element's center is not overlapped 
5. check if element is not disabled

had to change `isDisplayedInViewport` command because it didn't work. The logic is more or less the same:
1. check if element exist
2. check if element is visible
3. check if element is within viewport (removed everything related to visibility)

## Types of changes

- [x] New feature

## Checklist

## Further comments

ref #4646


mgrybyk:
I think Ihave to rollback `isDisplayed` changes, I'm afraid it might be breaking.

Imagine the case when user want's to check if element is shown and don't care if it's overlapped or not but just want to click on certain position.

It might affect `isDisplayedInViewport` as well, what do you think on making overlap verification optional?

igerasimov23:
@mgrybyk Just a small comment or you can think of it as a user test case:
1. isDisplayed() command usually used in automation before clicking on element. Most of us have a wrappers over click() command such as waitAndClick() which waits for element to be  displayed and then click() on that. If element will be overlapped, isDisplayed() command will return true (this is how it worked before) as element on the screen, but clicking on that element will return exception as selenium was not able to click it. Really looking for a way to somehow to return isDisplayed() method `false` if element overlapped and not clickable.
2. Second user test scenario as user does not care at all it overlapped or not, he wants just to verify element displayed on the page or not

Just my 2 cents

mgrybyk:
I agree on both points, however I'd avoid breaking changes because of point 2. 

I think it's also applicable to is displayed in viewport command. 
Not sure what to do. New command for that? Make both is displayed commands parametrized?

Ping @christian-bromann

CrispusDH:
Could be situation when element displayed but fully overlapped?

mgrybyk:
Yes

CrispusDH:
I think in this case isDisplayed should be false. In other situation... hm, we could define some %. For example by default isDisplayed will be false if element is overlapped more than 50%, it could be configurable. What do you think about?

igerasimov23:
My opinion, if click() returns exception it means overlapped, otherwise it's not

wswebcreation:
I haven't checked the code, but based on the description and the comments here are my 2 cents

If the behavior of a method that is used a lot is changed, or the outcome is changed then this would be **breaking**. 
Secondly the question is how the behavior would differ from other frameworks that use something like isDisplayed.

I would suggest to create a new method even though a lot of comments make sense

igerasimov23:
@wswebcreation I agree that we should not change behavior of current isDisplayed() method as it behaves the same as in the other frameworks and if we do it will break many tests. Instead, we can create option inside that method that will check if it overlapped or not or we can just create new method.

mgrybyk:
I don't mind adding new command to check if element is overlapped or not instead of changing isDisplayed behavior.

Any objections? Or we don't want to introduce such a command and let user deal with it?

wswebcreation:
My reason for having a separate command is to prevent:

- code complexity
- questions from the users which might have a higher impact in the amount of time you need to put on Gitter in debugging

But that's my 2 cents

mgrybyk:
Yeah, I agree. I just not sure if we need this command at all. It might be useful to understand if element is clickable or something

WillBrock:
I kind of like the idea of a new command as well.

christian-bromann:
I see two options:

1. allowing to add arguments to the `isDisplayed` command allowing it to be more strict.
2. creating a new command `isClickable` that does the check

Given that I always prefer extending to existing commands to keep command list short and powerful I would advocate for the first option. What are your opinions?

mgrybyk:
1. We can't be fully sure that element is clickable if it's not in the current viewport. 
Assuming that I'd leave command isDisplayed as is but only change isDisplayed in viewport command. 

2. Is it possible that element is not overlapped but still not clickable? If so maybe we should think on different name. It would be nice to have isClickable command, lots of users want it

mgrybyk:
@christian-bromann @WillBrock @wswebcreation @igerasimov23 @CrispusDH 
- removed changes from isDisplayed command.
- added `isClickable` command to check if element is overlapped

christian-bromann:
> not sure if it's a good idea to scroll to element

I know that a bunch of commands are scrolling the element into the viewport before doing anything with it, see also:

> The element interaction commands provide a high-level instruction set for manipulating form controls. Unlike Actions, they will implicitly scroll elements into view and check that it is an interactable element.

https://w3c.github.io/webdriver/#interaction

mgrybyk:
Ok, the flow then can be like this
Check if displayed
Check if displayed in viewport, if no scroll to element 
Check if element is overlapped

christian-bromann:
> Check if element is overlapped

What does overlapped mean exactly? As far as I can see it currently checks if the center of the element is overlapped by another element. Should we be concerned for different cases, e.g. the center is overlapped but some areas are clickable etc.?

mgrybyk:
@christian-bromann by default element click triggered in the center of element.
Example: https://jsfiddle.net/y1ex5prg/8/

image

Such element is not clickable because of 1x1 square in the center. If we cover everything else but center element is still clickable

christian-bromann:
> Such element is not clickable because of 1x1 red dot in the center. If we cover everything else but center, element is still clickable

That makes sense to me but it should be clearly documented.

mgrybyk:
Oh, it appeared that `isDisplayedInViewport` doesn't work correctly, have to fix it to reuse in isClickable.

Very simple example.
1. Go to google https://www.google.com/search?q=foo
image

2. `$("//h2[text() = 'People also ask']").isDisplayedInViewport()`
Actual Result: `isDisplayedInViewport` returns `false` (but should return `true`!)

mgrybyk:
-fixed is `isDisplayedInViewport`

updated https://github.com/webdriverio/webdriverio/pull/4656#issue-329768863

mgrybyk:
- removed `isElementOverlapped` script at all
- added instead `isElementClickable` script instead

mgrybyk:
ready for review

packages/webdriverio/src/commands/element/isClickable.js
```diff
@@ -0,0 +1,43 @@
...
+ * - exists;
+ * - visible;
+ * - is within viewport (if not try scroll to it);
+ * - it's center is not overlapped with another element.
+ *
+ * otherwise return false.
+ *
+ * Please note that `isClickable` works only in web and webviews, it doesn't work in mobile app native context.
+ *
+ * <example>
```
christian-bromann: Can we add a `waitUntil` example with this command? I can see people would be really interested in that.

packages/webdriverio/src/scripts/isElementInViewport.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * check if element is within the viewport
+ * https://gist.github.com/davidtheclark/5515733#gistcomment-2113205
```
christian-bromann: Can we add a comment noting that any changes to this file need to be reflected in `isElementClickable`?

packages/webdriverio/src/commands/element/isClickable.js
mgrybyk: Maybe just add waitForClickable command?

christian-bromann: Sounds good to me. This has been asked for a lot therefor I am fine with it

packages/webdriverio/src/scripts/isElementInViewport.js
mgrybyk: done

packages/webdriverio/src/commands/element/isClickable.js
mgrybyk: added `waitUntil` example for now

mgrybyk:
added:
- check if element is disabled in `isClickable`

packages/webdriverio/tests/commands/element/isClickable.test.js
```diff
@@ -0,0 +1,42 @@
+/**
+ * @jest-environment jsdom
```
christian-bromann: Any reason why we need this? We don't have it for other tests either.

packages/webdriverio/tests/scripts/isElementClickable.test.js
```diff
@@ -0,0 +1,75 @@
+/**
+ * @jest-environment node
```
christian-bromann: same

packages/webdriverio/tests/scripts/isElementInViewport.test.js
```diff
@@ -0,0 +1,105 @@
+/**
+ * @jest-environment node
```
christian-bromann: same

packages/webdriverio/tests/commands/element/isClickable.test.js
mgrybyk: Copy paste, removing

packages/webdriverio/tests/scripts/isElementClickable.test.js
mgrybyk: Should be here, we have it for script tests

packages/webdriverio/tests/scripts/isElementInViewport.test.js
mgrybyk: Should be here, we have it for script tests

packages/webdriverio/tests/scripts/isElementClickable.test.js
mgrybyk: well, it had to be `@jest-environment jsdom` so that there is no need to create our own `document` and `window`, but in these tests it's not needed because we mock everything

mgrybyk:
@christian-bromann removed `@jest-environment` where it's not needed

PR #4663, opened 2019-10-20
Shared store service to exchange data between workers

mgrybyk:
## Proposed changes

An ability to share some data between main and child processes (specs).
I've seen already multiple times request to share something across child processes.

- may be used to have some object to share between child processes, ex cookies
- some flag to execute something only N amount of times
- might be useful in reporters
- might be used to store session capabilities in watch mode

### works like this
set: `browser.sharedStore.set('foo', 'bar')` // set in main or any child process
get: `browser.sharedStore.get('foo')` // outputs 'bar' in main or any child process

image

## Types of changes

## Checklist

- [x] test in windows
- [x] unit tests
- [x] smoke tests

## Further comments

This is something very experimental, want to get feedback, please don't judge strictly!

if the overall idea is ok, I'll proceed working on, adding tests, etc


packages/wdio-shared-store-service/src/client.js
```diff
@@ -0,0 +1,15 @@
+const axios = require('axios').default
```
christian-bromann:
```suggestion
import axios from 'axios'
```

packages/wdio-runner/src/index.js
```diff
@@ -60,6 +60,8 @@ export default class Runner extends EventEmitter {
         this.configParser.filterWorkerServices()
 
         this.config = this.configParser.getConfig()
+        this.config.sharedStoreServicePort = sharedStoreServicePort
```
christian-bromann: where is this used?

packages/wdio-shared-store-service/src/server.js
```diff
@@ -0,0 +1,38 @@
+const polka = require('polka')
+const { json } = require('body-parser')
+
+const store = {}
+
+const app = polka()
```
christian-bromann: why do we need a web server if we can just store the data in memory?

mgrybyk: I don't know how to do it, to be honest. 
How can I share data between two child process? Don't want use process events for that

packages/wdio-runner/src/index.js
mgrybyk: It's used in the service part of the service. I want to run only one instance of server on a random port to avoid conflicts but it's not trivial to pass data (port) from main process to child processes. Anyway I'd need to send an event with payload to child process or something like that.

mgrybyk: anyway I've removed it, added it temporary because it was easier (for me) to pass data this way.

wswebcreation:
Hi @mgrybyk 

This looks really great and I believe the mentioned usecases can benefit of it. There is only one thing that worries me regarding testautomation and the way people might _miss-use_ it. If you write proper testcases they should be atomic but also autonomous, especially the last one will not be met because, if I understand this functionality correctly, use data between instances. This is a bad practice and not advised if we look at _test automation best practices_.
Even with a disclaimer people will _miss-use_ it. What do you think of it, or do I understand the functionality incorrect?

mgrybyk:
Absolutely agree, the idea to have this service is not to do something in the tests but rather have more control over environment where the tests are running on, ex store cookies, users, etc. Sometimes it's not possible to create fully isolated environment for e2e test.
Btw, see recently submitted issue 4665

wswebcreation:
I agree with that, a disclaimer might be useful, but on the other hand, if you are referring to 

> Btw, see recently submitted issue 4665

Then I would suggest providing a property in the config that would read for example testdata from an array of data and attach it to each instance instead of this. Something like:

```js
exports.config = {
  //...
  
  // this is the property, the object can have a different structure
  testData: [
    {
      username: 'John Doe',
      password: 'foo-bar-1901'
    },
    // and so on
  ],
  //...
}
```

and then use it like this

```js
$('#elementID').setValue(browser.testData.username);
```

We could also get questions like `I want to be in control which data is connected to which instance`, but I don't know how to solve that one for now.

I'm just worried the current implementation will be used in a way it is not intended to be used

mgrybyk:
What can you do if you can use the user only once?

BorisOsipov:
>I'm just worried the current implementation will be used in a way it is not intended to be used

+1
Feature is great, but I am sure people will abuse this feature to make coupled tests.
Heard the same concerns from @CrispusDH

I'd like @wswebcreation proposal to allow store test data somewhere in config object, but will better create some wdio service which provide test data queue.

Use case example:

User provides final set of credentials <login, password> with N values in config.
When starts test user calls `browser.sharedStore().acquireData()` and receive next credentials from queue if it exists. If no free users  will have to retry pull credentials sometime to wait until free credentials appears.
When tests ends user have to back credentials into queue `browser.sharedStore().releaseData(credentials)`.

Sorry for chaotic description, but I think only way to avoid misusing is restrict making changes in test data from test. Only working with existing data (but it still leaves some ways to abuse)

BorisOsipov:
>What can you do if you can use the user only once?

in my idea with queue - you just not release credentials after test :)

packages/wdio-shared-store-service/src/server.js
christian-bromann: You can launch the storage part as a single identity in the launcher. The launcher is run only once in the launcher process. In the launcher you can then add the port to the wdio config which gets propagated into all child processes.

For data transportation between storage and child processes you are right that an simple server seems to be the simplest solution but there might be also other options.

mgrybyk:
Agree on that, but it's not that flexible and only covers very llimited amount of cases.

I think that we should not limit user, believe me even without such feature it's possible to design test extremely terrible

packages/wdio-shared-store-service/src/server.js
mgrybyk: The port added to config is not propagated to child processes, sorry to say.
That was something I've tried first

BorisOsipov:
> but it's not that flexible and only covers very llimited amount of cases.

yes it is.

>believe me even without such feature it's possible to design test extremely terrible

Sad but true.

packages/wdio-shared-store-service/src/server.js
mgrybyk: Yes, for capabilities it works

mgrybyk:
What is the decision here? Are you guys ok with adding the feature?

packages/wdio-shared-store-service/src/server.js
```diff
@@ -0,0 +1,50 @@
...
+    })
+
+const startServer = async () => new Promise((resolve, reject) => {
+    /**
+     * run server on a random port, `0` stands for random port
+     * > If port is omitted or is 0, the operating system will assign
+     * > an arbitrary unused port, which can be retrieved by using `server.address().port`
+     * https://nodejs.org/api/net.html#net_server_listen_port_host_backlog_callback
+     */
+    app.listen(0, (err) => {
```
BorisOsipov: Small question, does this call occurs before selenium server started(I mean e.g. selenium-standalone service)?
I am worried about very rare case when random server port by coincidence will be `4444` when usually selenium server will start.

mgrybyk: It starts same time all the services are started (in parallel)

mgrybyk: Before proceeding I'll check how it works on windows as well

mgrybyk: The ranges should be between 49152 and 65535

@BorisOsipov

BorisOsipov: @mgrybyk nice! thanks for research.

mgrybyk:
I like more syntax `browser.sharedStore.get('foo')` instead of `browser.sharedStore().get('foo')`. 
Is it ok to add `sharedStore` directly to browser object instead of using `addCommand`?

CrispusDH:
> I like more syntax `browser.sharedStore.get('foo')` instead of `browser.sharedStore().get('foo')`.
> Is it ok to add `sharedStore` directly to browser object instead of using `addCommand`?

I like more `browser.sharedStore.get('foo')` syntax also

packages/wdio-shared-store-service/src/server.js
```diff
@@ -0,0 +1,50 @@
...
+    // routes
+    .post('/get', (req, res) => {
+        res.end(JSON.stringify({ value: store[req.body.key] }))
+    })
+    .post('/set', (req, res) => {
+        store[req.body.key] = req.body.value
+        return res.end()
+    })
+
+const startServer = async () => new Promise((resolve, reject) => {
```
CrispusDH: `async` is redundant here

mgrybyk:
ready for review!

mgrybyk:
@christian-bromann I think I should replace axios with got here?

mgrybyk:
replaced `axios` with `got`

PR #4668, opened 2019-10-21
set agent.keepAlive based on Connection header

kellyselden:
## Proposed changes

I attempted to turn off the default keep-alive behaviour like this:

```js
      browser = await remote({
        headers: {
          Connection: 'close',
        },
      });
```

but I was still seeing `Connection: 'keep-alive'` in the browser. I tracked down our handoff to the `request` package, and the custom header was correctly `Connection: 'close'`. Turns out, if you send both `Agent.keepAlive` and `Connection: 'close'` to `request`, it will overwrite any custom headers with what is in your agent config. This fixes our agent keep-alive assumption to match the Connection header.

Unfortunately, since we have to now mutate the agents, they can't be globally stored anymore.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/webdriver/src/request.js
+++ b/packages/webdriver/src/request.js
@@ -18,8 +18,8 @@ const DEFAULT_HEADERS = {
 
 const log = logger('webdriver')
 const agents = {
-    http: new http.Agent({ keepAlive: true }),
-    https: new https.Agent({ keepAlive: true })
+    http: http.Agent,
+    https: https.Agent
 }
 
 export default class WebDriverRequest extends EventEmitter {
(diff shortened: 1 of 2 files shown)


christian-bromann:
> but I was still seeing `Connection: 'keep-alive'` in the browser.

What do you mean by this? The headers options defines header properties between WebdriverIO and the browser driver.

kellyselden:
Ah, you're right. This doesn't fix the issue.

kellyselden:
Is there a way to get a `browser.url` call to not use `Connection: keep-alive`?

christian-bromann:
> call to not use `Connection: keep-alive`?

It is how your user would request your page, why would you want to modify this?

kellyselden:
My test server is closing via

```js
    await new Promise(resolve => {
      this.server.close(resolve);
    });
```

but it is taking 5 seconds to close because of the default https://nodejs.org/api/http.html#http_server_keepalivetimeout. If I set keepAliveTimeout to `1`, it works as expected, but I was looking for a better way to do this.

kellyselden:
This sped up my tests a lot. I was just trying to come up with a better solution. https://github.com/CrowdStrike/faltest/pull/129

Issue #4688, opened 2019-10-25
@wdio/selenium-standalone-service cannot find IE driver


tugend:
**Config of WebdriverIO**
I followed the vanilla getting started example here; https://webdriver.io/docs/gettingstarted.html.
Using a Windows 10 machine, that runs IE smoothly using Nightwatch and Selenium, or even webdriver via codecept. 

```
npm init -y
npm i @wdio/cli
.\node_modules\.bin\wdio config (yes to all and pick selenium-standalone)
Manualle chagne capabilities to `{browserName: 'internet explorer'}`
.\node_modules\.bin\wdio wdio.conf.js
```

As far as I understand this should work.
I then spent 5 hours trying various blog post suggestions to no avail.

I have tried to manually update selenium with the right driver
`.\node_modules\.bin\selenium-standalone install --drivers.ie.arch=ia32 --version=3.0.0 --drivers.ie.baseURL=https://selenium-release.storage.googleapis.com`

I have tried to add options to the selenium standalone sevice

```
services: [
      [
          'selenium-standalone',
          {
              seleniumLogs: "./",
              seleniumArgs: { 
                  drivers: {
                      ie: {
                          version: "3.141.0",
                          arch: "ia32", // forces use of 32 bit driver
                          baseURL: "https://selenium-release.storage.googleapis.com"
                      },
                  },
              },
              seleniumInstallArgs: {
                  drivers: {
                      ie: {
                          version: "3.141.0",
                          arch: "ia32", // forces use of 32 bit driver
                          baseURL: "https://selenium-release.storage.googleapis.com"
                      },
                  },
              }
          }
      ]
  ],
 ``` 

I have also tried various options for the ie driver it self

```
maxInstances: 5,
version: '3.0.0',
version: 11,
browserName: 'internet explorer',
acceptUntrustedCertificates: true,
ignoreProtectedModeSettings: true,    //only applicable to IE browser
ignoreZoomSetting: true,              //only applicable to IE browser
ensureCleanSession: true,
``` 

**Describe the bug**
Regardless of what I do, I get the same error as if I misspell the browser name, i.e. `"browserName": "internet explorer"` or `"browserName": "Vanilla Sandwich"` results in the same stacktrace which basically boils down to `Cannot read property 'capabilities' of null` and a bunch of default error page html from selenium.

See https://gist.github.com/tugend/7812b343177c1305496dc6cf93397ca4 for the full stacktrace

**To Reproduce**
See the above. 

[...]

PR #4689, opened 2019-10-26
wdio-cli: Update documentation wrt. overwriting the path variable

tugend:
## Proposed changes

Updated the documentation wrt. overwriting the path variable. I spent a lot of time and confusion figuring out that it should be changed (or outcommented) before I could run IE or EDGE drivers.

## Types of changes

- [x] Documentation update

## Checklist

## Further comments

### DIFF
--- a/docs/DriverBinaries.md
+++ b/docs/DriverBinaries.md
@@ -103,7 +103,8 @@ So the top of your file should look like:
 exports.config = {
     runner: 'local',
     //
-    // Uncomment line below to override default path ('/wd/hub') for usage of driver binary directly, ex: chromedriver or geckodriver.
+    // Override default path ('/wd/hub') for usage of driver binary directly, ex: chromedriver or geckodriver.
+    // If you want to run IE or Edge drivers, set it to '/wd/hub' instead.
     path: '/',
 }
 ```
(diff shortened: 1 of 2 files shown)


mgrybyk:
I'm not sure if the PR is useful. All you need to make iedriver work with selenium standalone is remove path from config. It's in the docs, I've mentioned it in the issue as well. Not sure if the change make it more clear

tugend:
That's fair. Can you point me in the direction of the doc that mentions it?
Because I missed it entirely. 

It works fine either removing it or setting the path explicityly to /wd/hub, I thought it was a bit cleaner to change it explicitly.

tugend:
Previous discussion can be found here; https://github.com/webdriverio/webdriverio/issues/4688#issuecomment-546586689

mgrybyk:
I'd keep the files you've changed untouched and add hint to remove path here https://webdriver.io/docs/selenium-standalone-service.html

By default path is commented out so that selenium standalone users don't need to do anything, while it's only mentioned for chromedriver users to uncomment path https://webdriver.io/docs/wdio-chromedriver-service.html#configuration

tugend:
I went and retested. If I use config and only pick selenium-standalone, you're right. So I have to agree, the setup makes sense. If one picks both chromedriver and selenium-standalone the path variables is included and the confusion arises.

Hence, my mistake. Thank you for commenting. 
I'll close the issue.

```
npm i --save-dev @wdio/cl
./node_modules/.bin/wdio config // pick selenium-standalone only
```

PR #4690, opened 2019-10-27
Handle session errors

mgrybyk:
## Proposed changes

Print understandable message if session was not started, example:
image

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/packages/webdriver/src/utils.js
+++ b/packages/webdriver/src/utils.js
@@ -426,11 +426,18 @@ export const getSessionError = (err) => {
         return "Run edge driver on 127.0.0.1 instead of localhost, ex: --host=127.0.0.1, or set `hostname: 'localhost'` in your wdio.conf.js"
     }
 
-    // Illegal w3c capability
+    const w3cCapMessage = '\nMake sure to add vendor prefix like "goog:", "appium:", "moz:", etc to non W3C capabilities.' +
+        '\nSee more https://www.w3.org/TR/webdriver/#capabilities'
+
+    // Illegal w3c capability passed to selenium standalone
     if (err.message.includes('Illegal key values seen in w3c capabilities')) {
-        return err.message +
-            '\nMake sure to add vendor prefix like "goog:", "appium:", "moz:", etc.' +
-            '\nSee more https://www.w3.org/TR/webdriver/#capabilities'
+        return err.message + w3cCapMessage
+    }
+
+    // wrong host/port, port in use, illegal w3c capability passed to selenium grid
+    if (err.message === 'Response has empty body') {
+        return 'Make sure to connect to valid hostname:port or the port is not in use.' +
+            '\nIf you use a grid server ' + w3cCapMessage
     }
 
     return err.message
--- a/packages/webdriver/src/utils.js
+++ b/packages/webdriver/src/utils.js
@@ -16,6 +16,13 @@ const MOBILE_CAPABILITIES = [
     'device-orientation', 'deviceOrientation', 'deviceName'
 ]
 
+const BROWSER_DRIVER_ERRORS = [
+    'unknown command: wd/hub/session', // chromedriver
+    'HTTP method not allowed', // geckodriver
+    "'POST /wd/hub/session' was not found.", // safaridriver
+    'Command not found' // iedriver
+]
+
 /**
  * start browser session with WebDriver protocol
  */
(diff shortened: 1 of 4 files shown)


packages/webdriver/src/utils.js
```diff
@@ -382,3 +396,35 @@ export function setupDirectConnect(params) {
...
+    }
+
+    // browser driver / service is not started
+    if (err.code === 'ECONNREFUSED') {
+        return `Unable to connect to "${err.address}:${err.port}", make sure browser driver is running on that address.`
+    }
+
+    // wrong path: selenium-standalone
+    if (err.message.includes('Whoops! The URL specified routes to this help page.')) {
+        return 'Make sure there is no `path` in wdio.conf!'
```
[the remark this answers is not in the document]
christian-bromann: Given that the `path` not always has to be set to `/wd/hub` we should explain that it usually is set to that value or if not how one can find out where this path is set

packages/webdriver/src/utils.js
```diff
@@ -382,3 +396,35 @@ export function setupDirectConnect(params) {
         params.path = directConnectPath
     }
 }
+
+/**
+ * get human readable message from response error
+ * @param {Error} err response error
+ */
+export const getSessionError = (err) => {
+    if (!err.message) {
```
christian-bromann: let's move this section below `if (err.code === 'ECONNREFUSED') {`

packages/webdriver/src/utils.js
```diff
@@ -382,3 +396,35 @@ export function setupDirectConnect(params) {
...
+    }
+
+    // wrong path: chromedriver, geckodriver, etc
+    if (BROWSER_DRIVER_ERRORS.some(m => err.message.includes(m))) {
+        return "Make sure `path: '/'` exists in config!"
+    }
+
+    // edge driver on localhost
+    if (err.message.includes('Bad Request - Invalid Hostname') && err.message.includes('HTTP Error 400')) {
+        return 'Run browser driver on 127.0.0.1 instead of localhost, ex: --host=127.0.0.1'
```
christian-bromann: The WebDriver option I think you are referring here is `hostname`

mgrybyk: for edge it's --host
image

packages/webdriver/src/utils.js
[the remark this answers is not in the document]
mgrybyk: can you please take a look again?

packages/webdriver/src/utils.js
mgrybyk: edge returns ugly html page if host is 127.0.0.1 instead of localhost by default

packages/webdriver/src/utils.js
```diff
@@ -382,3 +396,35 @@ export function setupDirectConnect(params) {
...
+        return `Unable to connect to "${err.address}:${err.port}", make sure browser driver is running on that address.`
+    }
+
+    if (!err.message) {
+        return 'See logs for more information.'
+    }
+
+    // wrong path: selenium-standalone
+    if (err.message.includes('Whoops! The URL specified routes to this help page.')) {
+        return "Make sure there is no `path: '/'` or path set to valid uri, ex: '/wd/hub', in wdio.conf!"
```
christian-bromann: How about:
```suggestion
        return "It seems you are running a Selenium Standalone server and point to a wrong path. Please set `path: '/wd/hub'` in your wdio.conf.js!"
```

christian-bromann: we could also try to fix the path automatically, maybe in a different PR

mgrybyk: Agree, depending on service used. Actually we can do it in the services!

packages/webdriver/src/utils.js
```diff
@@ -382,3 +396,35 @@ export function setupDirectConnect(params) {
...
+    }
+
+    // wrong path: chromedriver, geckodriver, etc
+    if (BROWSER_DRIVER_ERRORS.some(m => err.message.includes(m))) {
+        return "Make sure `path: '/'` exists in config!"
```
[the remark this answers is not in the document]
mgrybyk: That's gonna be breaking change. Also, default path for selenium grid is wd hub

packages/webdriver/src/utils.js
mgrybyk: I'll change in selenium standalone and raise pr or issue to chrome and firefox services

packages/webdriver/src/utils.js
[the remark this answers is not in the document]
christian-bromann: that is true, I will create an issue to do this for v6 release

mgrybyk:
wanna add one more case
image

mgrybyk:
Just figured out I missed one more scenario
image

Don't have anything else to be added, we can merge the PR if strings are ok

Issue #4697, opened 2019-10-29
this.skip() in async tests causes "Cannot read property 'split' of undefined"


colinbendell:
Prior to 5.15.1, calls to `this.skip()` in an async mocha test would properly pass through as a skipped test. However, now with the asyncTest refactoring to use async/await, all calls to `this.skip()` result in:
> Cannot read property 'split' of undefined

This is because `this.skip()` calls `throw new Pending('sync skip')`. This used to fall out of executeAsync() by being wrapped in a Promise.reject(). How, it is caught and rethrown after prettying the `e.stack`. However, `e.stack` doesn't exist and thus causes a new error.

christian-bromann:
Let's continue convo in #4698

PR #4698, opened 2019-10-29
pretty e.stack only if it exists

colinbendell:
Addresses #4697.  skipped tests in mocha result in `throw new Pending('sync skip')` that do not have stack traces.

## Proposed changes

Mocha throws a `new Pending('sync skip)` when an async test is `this.skip()`ed. This used to pass through wdio, but now gets evaluated and the stack prettied. However, the stack doesn't always exist - particularly from this non exception use case - and will result in a new exception `Cannot read property 'split' of undefined` from wdio. 

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/packages/wdio-sync/src/index.js
+++ b/packages/wdio-sync/src/index.js
@@ -71,7 +71,9 @@ const executeAsync = async function (fn, repeatTest = 0, args = []) {
             return await executeAsync(fn, --repeatTest, args)
         }
 
-        if (e.stack) e.stack = e.stack.split('\n').filter(STACKTRACE_FILTER_FN).join('\n')
+        if (e.stack) {
+           e.stack = e.stack.split('\n').filter(STACKTRACE_FILTER_FN).join('\n')
+        }
         throw e
     }
 }
--- a/packages/wdio-sync/src/index.js
+++ b/packages/wdio-sync/src/index.js
@@ -72,7 +72,7 @@ const executeAsync = async function (fn, repeatTest = 0, args = []) {
         }
 
         if (e.stack) {
-           e.stack = e.stack.split('\n').filter(STACKTRACE_FILTER_FN).join('\n')
+            e.stack = e.stack.split('\n').filter(STACKTRACE_FILTER_FN).join('\n')
         }
         throw e
     }
--- a/packages/wdio-sync/src/index.js
+++ b/packages/wdio-sync/src/index.js
@@ -70,7 +70,9 @@ const executeAsync = async function (fn, repeatTest = 0, args = []) {
         if(repeatTest > 0) {
             return await executeAsync(fn, --repeatTest, args)
         }
-
+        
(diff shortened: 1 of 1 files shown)


packages/wdio-sync/src/index.js
```diff
@@ -71,7 +71,7 @@ const executeAsync = async function (fn, repeatTest = 0, args = []) {
             return await executeAsync(fn, --repeatTest, args)
         }
 
-        e.stack = e.stack.split('\n').filter(STACKTRACE_FILTER_FN).join('\n')
+        if (e.stack) e.stack = e.stack.split('\n').filter(STACKTRACE_FILTER_FN).join('\n')
```
baruchvlz:
```suggestion
        if (e.stack) {
           e.stack = e.stack.split('\n').filter(STACKTRACE_FILTER_FN).join('\n')
        }
```

mgrybyk:
Are tests properly marked as skipped with the fix?

mgrybyk:
@colinbendell can you please sign CLA

colinbendell:
> @colinbendell can you please sign CLA

done.

packages/wdio-sync/src/index.js
```diff
@@ -71,7 +71,9 @@ const executeAsync = async function (fn, repeatTest = 0, args = []) {
             return await executeAsync(fn, --repeatTest, args)
         }
 
-        e.stack = e.stack.split('\n').filter(STACKTRACE_FILTER_FN).join('\n')
+        if (e.stack) {
+           e.stack = e.stack.split('\n').filter(STACKTRACE_FILTER_FN).join('\n')
```
mgrybyk: it should be 12 spaces instead of 11 here
```
/home/travis/build/webdriverio/webdriverio/packages/wdio-sync/src/index.js
  75:1  error  Expected indentation of 12 spaces but found 11  indent
```

colinbendell:
> Can we add a comment why stack is missing here?

in the code or in the PR? As mentioned above, this is because in mocha, `this.skip()` does a `throw new Pending('sync skip)` which is not an instance of `Error` and therefore doesn't have a stack.

mgrybyk:
In the code, please

packages/wdio-sync/src/index.js
```diff
@@ -70,8 +70,12 @@ const executeAsync = async function (fn, repeatTest = 0, args = []) {
         if(repeatTest > 0) {
             return await executeAsync(fn, --repeatTest, args)
         }
-
-        e.stack = e.stack.split('\n').filter(STACKTRACE_FILTER_FN).join('\n')
+
```
christian-bromann:
```
/home/travis/build/webdriverio/webdriverio/packages/wdio-sync/src/index.js
  73:1   error  Trailing spaces not allowed  no-trailing-spaces
  74:93  error  Trailing spaces not allowed  no-trailing-spaces
```

PR #4700, opened 2019-10-29
webdriverio: expanded click command with offset

erwinheitzman:
## Proposed changes

Expand click command with an `x` and `y` offset (can be combined with the `button` property)

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/packages/webdriverio/src/commands/element/click.js
+++ b/packages/webdriverio/src/commands/element/click.js
@@ -24,19 +24,27 @@
         const myButton = $('#myButton')
         myButton.click()
         const myText = $('#someText')
-        const text = myText.getText();
-        assert(text === 'I was clicked'); // true
+        const text = myText.getText()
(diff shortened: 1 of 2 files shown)


packages/webdriverio/src/commands/element/click.js
```diff
@@ -46,56 +54,82 @@
...
+            }]
+        }])
 
-        await this.moveTo()
-        return this.positionClick(button)
+        return this.releaseActions()
     }
 
-    throw new Error('Button type not supported.')
+    const { width, height } = await this.getElementSize(this.elementId)
```
mgrybyk: shouldn't we use https://webdriver.io/docs/api/element/getSize.html here instead?

christian-bromann: not really because it is definitely not the w3c case here

Issue #4674, opened 2019-10-24
JSDoc for IDE's


pako88:
**Is your feature request related to a problem? Please describe.**
It would be nice to have the JSDoc documentation visible in IDE's like Visual Studio Code.

**Describe the solution you'd like**
I want documentation of every command visible in IDE's.

**Additional context**
![image](https://user-images.githubusercontent.com/28978631/67466122-ca149180-f646-11e9-8925-7a239f6ad148.png)
Documentation should be available in webdriverio-core.d.ts

mgrybyk:
will be implemented with the PR

PR #4706, opened 2019-10-30
Types refactoring

mgrybyk:
## Proposed changes

wrapping commands with Promises increased complexity too much so I decided to generate two separate files, one for async and another for sync versions.

- avoid wrapping commands with Promise 
- generate separate files for sync and async versions so there should not be any hidden/tricky/unobvious issues
- add generated types files to the repo to detect changes in webdriver, webdriverio and @wdio/sync packages automatically
- ~dropped WebdriverIOAsync, so it has to be replaced with WebdriverIO~ let's do it in v6 https://github.com/webdriverio/webdriverio/issues/4720
- added js docs https://github.com/webdriverio/webdriverio/issues/4674

## Types of changes

- [x] Documentation update

## Checklist


mgrybyk:
Do you think it's a serious change to remove WebdriverIOAsync and use only WebdriverIO?

ping @CrispusDH as far as he is one of the typescript + async mode users

mgrybyk:
webdriverio command
image

webdriverio command from specific commands
image

protocol command
image

mgrybyk:
@CrispusDH yes. As a typescript + async user, is it difficult to replace WebdriverIOAsync with WebdriverIO in your project?
I want to understand can we do it now or not

christian-bromann:
Happy to merge this one after getting feedback from @CrispusDH

mgrybyk:
@christian-bromann what is you opinion that the PR is breaking? 
Sorry to say there is no way to keep both old and new behavior. 
We may postpone it till v6, of course I'd like to have it now, hehe

christian-bromann:
Let's understand first if it would break @CrispusDH tests. Do you see a way to make this an opt-in thing?

mgrybyk:
I have tried in multiple ways but can't make it optin. I haven't tried on more thing I have on my mind which is to leave async typings as is and replace them with new way in v6

erwinheitzman:
Will have a look at this tomorrow @mgrybyk

CrispusDH:
> @CrispusDH yes. As a typescript + async user, is it difficult to replace WebdriverIOAsync with WebdriverIO in your project?
> I want to understand can we do it now or not

I can do it very easily. Using WebStorm I will replace `WebdriverIOAsync` with `WebdriverIO`. I think VSCode should have some similar functionality for refactoring.

mgrybyk:
Same here, but it's still breaking  :(

christian-bromann:
Let's hold this for v6 release and plan this release earlier.

mgrybyk:
now it's not be a breaking change, created https://github.com/webdriverio/webdriverio/issues/4720 to do it for v6

scripts/templates/webdriverio.tpl.d.ts
```diff
@@ -176,17 +180,25 @@ declare namespace WebdriverIO {
...
+         * overwrite `browser` or `element` command
+         */
         overwriteCommand(
             name: string,
             func: (origCommand: Function, ...args: any[]) => any,
             attachToElement?: boolean
         ): void;
-        options: RemoteOptions;
         // ... browser commands ...
     }
```
baruchvlz: We should add the type definitions for `addLocatorStrategy` missed these in my PR.

mgrybyk: Will you be able to raise a PR for that?

PR #4722, opened 2019-11-03
Add test retry attempts to after hooks

### LINKED ISSUE #4662: Send number of retries to afterTest() hook
Mocha and WebdriverIO now does not report number of retries in afterTest() hook in which test right now during execution. But it will be very useful to help identify **flaky** tests. Many reporting system support retry option (such as https://reportportal.io/) but we can not send anything to that from WebdriverIO during execution

afterTest() hook should send number of retry in which test during execution. For example, if we have globally set Retry option to `3`, test failed on 2nd we should send number `2` and so on.

mgrybyk:
## Proposed changes

- add TEST retries limit and attempts to after test / hook.

image

- add SPEC FILE retries attempts to config `browser.config.specFileRetryAttempts`. We already have `specFileRetries` in config defined by user.
image

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

works same in mocha, cucumber, etc

fixes #4662

### DIFF
--- a/docs/ConfigurationFile.md
+++ b/docs/ConfigurationFile.md
@@ -3,7 +3,7 @@ id: configurationfile
 title: Testrunner Configuration
 ---
 
-The configuration file contains all necessary information to run your test suite. It’s just a NodeJS module that exports a JSON. 
+The configuration file contains all necessary information to run your test suite. It’s just a NodeJS module that exports a JSON.
 
 Here is an example configuration with all supported properties and additional information:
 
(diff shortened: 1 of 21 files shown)


mgrybyk:
`specFileRetryAttempts` is not available in hooks in `this` because context of hooks is not `config` but a copy of config, see https://github.com/webdriverio/webdriverio/blob/master/packages/wdio-config/src/lib/ConfigParser.js#L49

we can use `browser.config.specFileRetryAttempts` instead

PR #4733, opened 2019-11-04
Add docs to describe difference between sync and async modes

mgrybyk:
## Proposed changes

It's quite frequent question about differences between sync and async modes, browser.call, mixing sync and async functions.

It would be good to have it documented and just send link to the doc instead of explaining again

## Types of changes

- [x] Documentation update

## Checklist

## Further comments

image

### DIFF
--- a/docs/SyncVsAsync.md
+++ b/docs/SyncVsAsync.md
@@ -3,6 +3,10 @@ id: sync-vs-async
 title: Sync mode vs async
 ---
 
+## How to enable/disable sync mode
+
+To enable sync mode it's only required to install `@wdio/sync` package, to disable sync mode uninstall `@wdio/sync`.
+
 ## Async mode
 
 Every browser/element function call returns a Promise and needs to be awaited to get result.
--- a/docs/SyncVsAsync.md
+++ b/docs/SyncVsAsync.md
@@ -1,0 +1,67 @@
+---
+id: sync-vs-async
+title: Sync mode vs async
+---
+
+## Async mode
+
+Every browser/element function call returns a Promise and have to be awaited to get result.
+
+It's not allowed to chain methods.
+
+Example
+
+```js
+describe('suite async', () => {
+    it('test', async () => { // pay attention to `async` keyword
+        const el = await $('body') // note `await` keyword
+        await el.click()
+
+        await browser.pause(500)
+
+        console.log(browser.capabilities) // static properties should not be awaited
+
+        await $('body').click() // WON'T WORK! You can't chain methods like this.
+    })
+})
+```
+
+## Sync mode
+
+If you have `@wdio/sync` installed you can avoid awaiting browser/element calls. It is still required to deal with Promises from 3rd-party libraries, use [browser.call](/api/browser/call.html) to do that.
+
+Example
+
+```js
+// https://www.npmjs.com/package/got#gotgeturl-options
+const { get } = require('got')
+
+describe('suite sync', () => {
+    it('test sync',
+    // make sure to remove `async` keyword otherwise function treated as async
+    // and you have to await every browser call.
+    // The same is applicable to hooks in `wdio.conf.js` as well.
+    () => {
+        browser.pause(500)
+
+        // wrap 3rd-party library calls with `browser.call`
+        const response = browser.call(() => get('https://cat-fact.herokuapp.com/facts/'))
+
+        $('body').click() // You can chain methods in sync mode
+    })
+
+    it('async test in sync mode',
+    // even if you have `@wdio/sync` installed you can use async functions.
+    // However in such case you have to await every browser/element call like in async mode!
+    async () => {
+        await browser.pause(500)
+
+        const response = await get('https://cat-fact.herokuapp.com/facts/')
+
+        const el = await $('body')
+        await el.click()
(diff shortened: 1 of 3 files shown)


mgrybyk:
Help with grammar is appreciated!

docs/SyncVsAsync.md
```diff
@@ -0,0 +1,67 @@
+---
+id: sync-vs-async
+title: Sync mode vs async
+---
+
+## Async mode
+
+Every browser/element function call returns a Promise and have to be awaited to get result.
```
pmerwin: Maybe this instead:
Every browser/element function call returns a Promise and needs to be awaited to get result.

docs/SyncVsAsync.md
```diff
@@ -0,0 +1,67 @@
...
+    () => {
+        browser.pause(500)
+
+        // wrap 3rd-party library calls with `browser.call`
+        const response = browser.call(() => get('https://cat-fact.herokuapp.com/facts/'))
+
+        $('body').click() // You can chain methods in sync mode
+    })
+
+    it('async test in sync mode',
```
pmerwin: Though this is correct we might discourage this as it tends to get confusing in sync mode.

mgrybyk: I got the same feeling here. How do you think it can be improved?

pmerwin: Maybe something like this:

    // even if you have `@wdio/sync` installed and configured, it is still possible to use async functions.
    // However, in such case you have to await every browser/element call like in async mode, and this can get 
    // confusing when other tests are sync, so we discourage mixing modes, but it is possible to do.

mgrybyk:
@pmerwin updated according to your comments, also added `Common issues` sections to sync and async modes

pmerwin:
> @pmerwin updated according to your comments, also added `Common issues` sections to sync and async modes


Awesome, thanks!!!

docs/SyncVsAsync.md
```diff
@@ -0,0 +1,93 @@
+---
+id: sync-vs-async
+title: Sync mode vs async
+---
+
+## Async mode
+
+Every browser/element function call returns a Promise and needs to be awaited to get result.
+
+It's not allowed to chain functions.
```
baruchvlz: This statement seems kind of vague, maybe adding why is not allowed might help with understanding.

docs/SyncVsAsync.md
```diff
@@ -0,0 +1,93 @@
...
+        console.log(browser.capabilities) // static properties should not be awaited
+
+        await $('body').click() // WON'T WORK! You can't chain functions like this.
+    })
+})
+```
+
+### Common issues in async mode
+
+- `await $('body').click()` throws `$(...).click is not a function` because the element was not awaited. To fix this first await element then do click, like this:
```
baruchvlz:
```suggestion
- `await $('body').click()` throws `$(...).click is not a function` because the element was not awaited. To fix this, first await element then trigger the click, like so:
```

docs/SyncVsAsync.md
```diff
@@ -0,0 +1,93 @@
...
+- previous command was not awaited:
+```js
+const el = await $('body')
+el.waitForExist() // await is missing here, you'll get `Unhandled promise rejection`.
+await el.click()
+```
+
+## Sync mode
+
+If you have `@wdio/sync` installed you can avoid awaiting browser/element calls. It is still required to deal with Promises from 3rd-party libraries, use [browser.call](/api/browser/call.html) to do that.
```
baruchvlz:
```suggestion
If you're using`@wdio/sync` then you can avoid awaiting for browser/element calls. It is still required to deal with Promises from 3rd-party libraries, you should use [browser.call](/api/browser/call.html) for this.
```

docs/SyncVsAsync.md
```diff
@@ -0,0 +1,93 @@
...
+
+        const el = await $('body')
+        await el.click()
+    })
+})
+```
+
+### How does sync mode work?
+
+Every browser/element command is wrapped with [fibers](https://github.com/laverdet/node-fibers) in sync mode to make async code look like sync.
```
baruchvlz:
```suggestion
Every browser/element command is wrapped with [fibers](https://github.com/laverdet/node-fibers) while in sync mode.
```

docs/SyncVsAsync.md
```diff
@@ -0,0 +1,93 @@
...
+
+### Common issues in sync mode
+
+- declaring test function as `async` and not awaiting browser/element functions
+- not awaiting 3rd party librariy promises with `browser.call`
+- using sync mode while `@wdio/sync` package is not installed
+- `fibers` failed to install properly. To fix it see `npm install` errors
```
baruchvlz: Maybe we should add an explanation as to why these are issues.

docs/SyncVsAsync.md
mgrybyk: I'll just remove it

docs/SyncVsAsync.md
mgrybyk: added

mgrybyk:
@baruchvlz thanks a lot for your help! Updated according to your comments

PR #4739, opened 2019-11-06
Pass capabilities to reporter before session started

### LINKED ISSUE #4738: Test fails with v5.16.0 and junit reporter with outputFileFormat parameter

 - **WebdriverIO version:** 5.16.0
 - **Mode:** WDIO Testrunner
 - **Node.js version:** 10.17.0
 - **NPM version:** 6.11.3
 - **Browser name and version:** doesn't matter
 - **Platform name and version:** doesn't matter
 - **Additional wdio packages used (if applicable):** doesn't matter

**Config of WebdriverIO**
```
...
reporters: [
        'spec',
        ['junit', {
            outputDir: './junit-results',
            outputFileFormat: function (options) {
                return `results-${options.cid}.${options.capabilities.browserName}.xml`
            },
        }]
    ],
...
```
**Describe the bug**
Tests will not run with outputFileFormat set. when i comment out this line, test runs fine.
This bug occurs since v5.16.0
**To Reproduce**

see above

**Expected behavior**
test runs
**Log**
https://gist.github.com/pako88/92e7c774befe9f1c675e163793ce1153

**Additional context**
-

mgrybyk:
## Proposed changes

Pass capabilities to reporter before session has started 

fixes #4738

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/packages/wdio-runner/src/index.js
+++ b/packages/wdio-runner/src/index.js
@@ -68,15 +68,15 @@ export default class Runner extends EventEmitter {
         const isMultiremote = this.isMultiremote = !Array.isArray(this.configParser.getCapabilities())
 
         /**
-         * create `browser` stub
+         * create `browser` stub only if `specFiltering` feature is enabled
          */
-        let browser = await this._startSession({
+        let browser = this.config.featureFlags.specFiltering === true ? await this._startSession({
             ...this.config,
             _automationProtocol: this.config.automationProtocol,
             automationProtocol: './protocol-stub'
-        }, caps)
+        }, caps) : undefined
 
-        this.reporter = new BaseReporter(this.config, this.cid, browser.capabilities)
+        this.reporter = new BaseReporter(this.config, this.cid, { ...caps })
         /**
          * initialise framework
          */
--- a/packages/wdio-runner/src/index.js
+++ b/packages/wdio-runner/src/index.js
@@ -68,15 +68,15 @@ export default class Runner extends EventEmitter {
         const isMultiremote = this.isMultiremote = !Array.isArray(this.configParser.getCapabilities())
 
         /**
-         * create `browser` stub only if `specFiltering` feature is enabled
+         * create `browser` stub
          */
-        let browser = this.config.featureFlags.specFiltering === true ? await this._startSession({
+        let browser = await this._startSession({
             ...this.config,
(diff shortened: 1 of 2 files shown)


packages/wdio-runner/src/index.js
```diff
@@ -68,15 +68,15 @@ export default class Runner extends EventEmitter {
         const isMultiremote = this.isMultiremote = !Array.isArray(this.configParser.getCapabilities())
 
         /**
-         * create `browser` stub only if `specFiltering` feature is enabled
+         * create `browser` stub
          */
-        let browser = this.config.featureFlags.specFiltering === true ? await this._startSession({
+        let browser = await this._startSession({
```
christian-bromann: I don't quite understand, are things not dependent on the `specFiltering` flag anymore?

mgrybyk: I screwed up with this:
1. init reporter, session capabilities are not available at this point
2. init test framework
3. then we can create browser session
4. pass capabilities to reporter

before it was:
1. create browser session
2. init reporter
3. run framework

mgrybyk: to deal with it there are 3 ways in mind
- current approach, the easiest. Create browser stub always and pass capabilities to reporter so that it can create log file with capabilities.browserName or whatever
- pass user defined capabilities to reporter
- change test framework init somehow so that reporter can be initialised after test framework. I'm not sure if it's possible at all

mgrybyk: I've chosen to always create browser stub. I takes 5ms to create it on my macbook 2015.

We can also just pass user defined capabilities instead, I think it should work in most of the cases

mgrybyk:
```
// do not change anything here
let browser = this.config.featureFlags.specFiltering === true ? await this._startSession({
    ...this.config,
    _automationProtocol: this.config.automationProtocol,
    automationProtocol: './protocol-stub'
}, caps) : undefined

this.reporter = new BaseReporter(this.config, this.cid, caps) // user defined caps
```

mgrybyk: @christian-bromann I've reverted the change and switch to approach with user defined caps. If you like the previous approach more it's easy to change it back.

Added unit tests!

pako88:
thanks guys - awesome work

PR #4742, opened 2019-11-06
website: Change API header display style to small-caps

ccurtis0:
## Proposed changes

API commands are case-sensitive and the CSS for the website forces names to ALLUPPERCASE, which makes commands like waitForClickable less obvious. The command example given below the title is helpful, but I still find $(compressed).syntax(hardToRead). This changes the "text-transform: uppercase" to "font-variant: small-caps" to preserve the monospace-like nature of capital letters while allowing one to differentiate between capital and lower-case letters when looking at the header.

A font-weight of 200 may also be warranted but this is untested; 300 appears too bold to me.

## Types of changes

- [X] Documentation update

## Checklist

- [V] I have read the [CONTRIBUTING](https://github.com/webdriverio/webdriverio/blob/master/CONTRIBUTING.md) doc

### DIFF
--- a/website/static/styles/_components/_headings.scss
+++ b/website/static/styles/_components/_headings.scss
@@ -1,7 +1,7 @@
 h1 {
     font-size: $h1Size;
     color: $black;
-    text-transform: uppercase;
+    font-variant: small-caps;
     font-weight: 100;
     padding: 0 0 40px;
     line-height: 40px;


mgrybyk:
Can you add screenshot?

mgrybyk:
image

now it looks like this but I still have a feeling that post header title is in uppercase especially if there is only one word in command

image

anyway I don't mind changing it, might be more clear

christian-bromann:
@ccurtis0 mind signing the CLA?

christian-bromann:
ping @ccurtis0

PR #4737, opened 2019-11-06
Build a script that grants TSC members to release @wdio packages

christian-bromann:
There are a lot of `@wdio` packages that need to be released. As per our governance only TSC members are allowed to release packages. To make on-boarding of new members easier and more automatic, let's build a script that calls NPMs API and registers members (maintained in an extra `json` file in the repository).

Also add a sentence in the `Governance.md` to run the script as part of the on-boarding.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -111,7 +111,7 @@ Commits that affect all packages or are not related to any (e.g. changes to NPM 
 
 ## Release New Version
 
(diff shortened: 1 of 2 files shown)


christian-bromann:
It turns out that NPM doesn't really have an API client and that they prefer this via their web interface. Therefor I added a section to the Governance description to add the user to the org which automatically grants publish rights.

PR #4762, opened 2019-11-09
wdio-applitools-service: add typings

ablok:
## Proposed changes

This PR adds typings for the wdio-applitools-service. While adding the typings I noticed that the frame option was written in the documentation, but was not supported in the code, so I added that.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments


packages/wdio-applitools-service/src/index.js
```diff
@@ -51,14 +51,18 @@ export default class ApplitoolsService {
...
+        global.browser.addCommand('takeRegionSnapshot', (title, region, frame) => {
             if (!title) {
                 throw new Error('A title for the Applitools snapshot is missing')
             }
             if (!region || region === null) {
                 throw new Error('A region for the Applitools snapshot is missing')
             }
-            return this.eyes.check(title, Target.region(region))
+            if (!frame) {
+                console.log(frame)
```
christian-bromann: `console.log` should be removed or if you want to log something actual meaningful use the wdio logger

ablok: Done

ablok:
Checking something, please hold off merging

ablok:
Yes, should be good now!

mgrybyk:
@ablok may you please add some very basic test for typings?

mgrybyk:
@ablok add any command to sync.ts in same way other commands are added, ex https://github.com/webdriverio/webdriverio/blob/e30649c9151aebc034a935db1e8399149ddf6078/tests/typings/sync/sync.ts#L97
You have to copy applitools types and add types to tsconfig.json

Copy.js should be here https://github.com/webdriverio/webdriverio/tree/master/tests/typings

tests/typings/webdriverio-applitools/applitools.ts
```diff
@@ -0,0 +1,20 @@
...
+    applitoolsServerUrl: '',
+    applitools: {
+        viewport: {
+            width: 1,
+            height: 1
+        }
+    }
+}
+
+browser.takeSnapshot('title');
```
mgrybyk: Can you move it please to async function and add await?

ablok:
> @ablok add any command to sync.ts in same way other commands are added, ex
> 
> https://github.com/webdriverio/webdriverio/blob/e30649c9151aebc034a935db1e8399149ddf6078/tests/typings/sync/sync.ts#L97
> 
> 
> You have to copy applitools types and add types to tsconfig.json
> Copy.js should be here https://github.com/webdriverio/webdriverio/tree/master/tests/typings

I added tests. But I was having trouble getting the .d.ts file to work properly. I first declared both the sync and async module by doing `declare module "@wdio/syn"` and `declare module "webdriverio"`. Because that is how it's done in all the other packages. But that ddidn't seem to work, so now I used `declare module "WebdriverIO"` and `declare module "WebdriverIOAsync"`.

mgrybyk:
@ablok we hope to deal with it in v6, there should be only webdriverio namespace so it's gonna be easier to add typings

ablok:
@mgrybyk But why does it seem to work for the other packages?

tests/typings/webdriverio-applitools/applitools.ts
ablok: Done

mgrybyk:
It works differently depending on if there are imports of other types or no. We have some packages with types declared like this

ablok:
I just tried using the functions while having the changed package linked locally, but it doesn't seem to work. I get the following error when executing a test:

[chrome  mac os x #0-0] browser.takeSnapshot is not a function
[chrome  mac os x #0-0] TypeError: browser.takeSnapshot is not a function

Any idea?

Issue #4767, opened 2019-11-11
wdio allure reporter does not save stacktrace/screenshot when a test fails on a "before/after" hook using Mocha


luiscspinho:
**Environment (please complete the following information):**
 - **WebdriverIO version:** [5.16.0]
 - **Mode:** [WDIO Testrunner]
 - **If WDIO Testrunner, running sync/async:** [async]
 - **Node.js version:** [10.14.0]
 - **NPM version:** [6.11.3]
 - **Browser name and version:** [Chrome 78]
 - **Platform name and version:** [Mojave 10.14]
 - **Additional wdio packages used (if applicable):** [@wdio/allure-reporter]

**Config of WebdriverIO**
Just add this to the `afterTest` hook:
```js
  afterTest: async test => {
    if (test.error !== undefined) {
      await browser.takeScreenshot();
    }
  },
```

**Describe the bug**
When a test fails on a `before` or `after` hook in mocha, the allure reporter doesn't show the stacktrace or the screenshot available as it would if the test failed on a `it` instruction.
![Screenshot 2019-11-11 at 10 34 56](https://user-images.githubusercontent.com/2779419/68581059-966f9f00-046f-11ea-8a05-b967dd184bf2.png)
![Screenshot 2019-11-11 at 10 35 11](https://user-images.githubusercontent.com/2779419/68581069-9bcce980-046f-11ea-94a0-c18c9b8f4910.png)

**To Reproduce**
Just make a test fail on a `before` and open the report.
I've created a small example on how to reproduce this behaviour.
https://github.com/luiscspinho/wdio-allure-bug

**Expected behavior**
I was expecting to have the stacktrace/screenshot like what happens when a test fails on an `it`

mgrybyk:
Before is not a test, move add screenshot to afterHook as well

luiscspinho:
I've updated the repo to have the `takeScreenshot` in the afterHook but the result is the same.

I reckon that the before is not a test, but shouldn't we register the stacktrace/screenshot in the report? There are some cases were `before` can be used to setup a test and something can go wrong there.

mgrybyk:
May you help investigating it?

luiscspinho:
Yes I'm glad to do it, I'm looking into the code but I need some directions. You can find me on gitter

mgrybyk:
[...]

PR #4406, opened 2019-08-27
allure-reporter: fixed bug with mocha all hooks being treated as tests

Gennadiii:
## Proposed changes

[//]: # Before and After hooks appear in report like tests which breaks statistics of tests. I might have 1 failing test out of 1 and it's 100% fails but with successful hooks it's only 33%.

## Types of changes

[//]: # (What types of changes does your code introduce to WebdriverIO?)
[//]: # (_Put an `x` in the boxes that apply_)

- [ x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

[//]: # (_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._)

- [ x] I have read the [CONTRIBUTING](https://github.com/webdriverio/webdriverio/blob/master/CONTRIBUTING.md) doc
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ x] I have added necessary documentation (if appropriate)

## Further comments

[//]: # (If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...)

PR #4773, opened 2019-11-12
wdio-allure-reporter: add flag to ignore/allow mocha hooks to have stacktrace and screenshots when they fail

luiscspinho:
## Proposed changes

Right now the allure reporter is not saving the stacktrace/screenshot when a mocha hook (`before/after`) fails. So, if we have some setup on our `before` that breaks the test, we're not able to debug it properly. (reference https://github.com/webdriverio/webdriverio/issues/4767 )
This MR adds an option (disableMochaHooks) that needs to be turned on in order to not count the hooks as meaningful for the report, and the default behaviour is, once again, registering all the actions in all the hooks. (basically this PR https://github.com/webdriverio/webdriverio/pull/4406 will have a flag for it to display once again as it was supposed to)

Not sure about marking this as **breaking** though, might be interpreted as a bug fix as well 😅 

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/wdio-allure-reporter/README.md
+++ b/packages/wdio-allure-reporter/README.md
@@ -42,6 +42,7 @@ exports.config = {
 - `disableWebdriverStepsReporting` - optional parameter(`false` by default), in order to log only custom steps to the reporter.
 - `disableWebdriverScreenshotsReporting` - optional parameter(`false` by default), in order to not attach screenshots to the reporter.
 - `useCucumberStepReporter` - optional parameter (`false` by default), set it to true in order to change the report hierarchy when using cucumber. Try it for yourself and see how it looks.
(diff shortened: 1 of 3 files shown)


mgrybyk:
Is it possible to not introduce any option and capture screenshots only for before hook?

luiscspinho:
That is indeed possible and a lot easier to do, but it'll break the functionality introduced here https://github.com/webdriverio/webdriverio/pull/4406 . To have both functionality at the same time I don't think it's possible, unless with a flag to disable that code.

luiscspinho:
BTW, this PR will make the screenshots/stacktrace to be available as default, with no flag. You only turn the flag on if you don't want screenshots on the hooks, which is an exception and will only be used in specific scenarios.

Issue #4539, opened 2019-09-26
[Cucumber JSON Reporter] Failed launching test session: TypeError: args[0].endsWith is not a function


tlee-fm:
**Environment (please complete the following information):**
 - **WebdriverIO version:** 5.11.0
 - **Mode:** WDIO Testrunner
 - **If WDIO Testrunner, running sync/async:** sync
 - **Node.js version:** 10.15.3
 - **NPM version:** 6.4.1
 - **Browser name and version:** Chrome 77.0.3865.90
 - **Platform name and version:** Windows 10
 - **Additional wdio packages used (if applicable):** @wdio/cucumber-framework, @wdio/local-runner, @wdio/sync, and wdio-cucumberjs-json-reporter

**Config of WebdriverIO**
```
{
  capabilities: [
    {
      browserName: "chrome"
    }
  ],
  connectionRetryCount: 3,
  connectionRetryTimeout: 90000,
  cucumberOpts: {
    backtrace: false,
    colors: true,
    compiler: [],
    dryRun: false,
    failFast: false,
    format: ["pretty"],
    ignoreUndefinedDefinitions: false,
    profile: [],
    require: ["steps.js"],
    requireModule: ["@babel/register"],
    snippets: true,
    source: true,
    strict: false,
    tags: [],
    timeout: 60000
  },
  framework: "cucumber",
  hostname: "localhost",
  logLevel: "trace",
  maxInstances: 1,
  outputDir: ".",
  path: "/wd/hub",
  port: 4444,
  reporters: ["cucumberjs-json"],
  runner: "local",
  specs: ["*.feature"],
  waitforTimeout: 10000
};
```

**Describe the bug**
Whenever I try to add in `reporters: ["cucumberjs-json"]` into the `wdio.config.js` file and run the WDIO test runner, the feature files never execute. The WDIO Testrunner exits and fails immediately without outputting the Cucumber JSON Reporter results. Logs are shown below. 

**To Reproduce**
[Gist Code](https://gist.github.com/tlee-fm/97ba6e935378054fbe8587df7c22fc43)
1) Download Zip file of Gist Code above.
2) `npm install`
3) `npm run startSelenium` in one terminal
4) `npm test` in another terminal
5) The terminal that ran `npm test` will output `Running` and then `Failed` without any Cucumber JSON Reporter results. 

**Expected behavior**
The WDIO Testrunner should successfully run all feature files and output the Cucumber JSON Reporter results.

**Log**
```
[...]

PR #4785, opened 2019-11-18
Fix bug where multiple copies of logger breaks logging errors

johnnymo87:
Similar to https://github.com/pimterry/loglevel/issues/117

## Proposed changes

As described in https://github.com/webdriverio/webdriverio/issues/4539#issuecomment-554548185, there's a problem with wdio-logger where it throws
```
TypeError: args[0].endsWith is not a function
```
.. because it assumes `args[0]` is a string. However, there's a bug that causes things to be in `args[0]` that are not a strings.

### How does this happen?
It happens when we have multiple copies of the wdio-logger source code. This is quite common, because wdio-logger is used by many other wdio-* packages.

### Why does this happen?
This happens because the wdio-logger source code uses the [loglevel](https://github.com/pimterry/loglevel) package, and it follows [its advice](https://github.com/pimterry/loglevel#plugins) to extend its `methodFactory` function to decorate it with extra behavior.

This works fine when we only have one copy of the wdio-logger source code in play. When we load our one copy of wdio-logger, it extends loglevel's version of `methodFactory` with its own version of `methodFactory`. When a user of wdio-logger calls `log.info('foo')`, loglevel's and wdio-loggers passes this around in a list, i.e. `['foo']`.

Loglevel's version of `methodFactory` runs first, and puts some metadata ahead of the `'foo'` log line, making the list look like this: `['some metadata', 'foo']`. Then wdio-logger's copy of `methodFactory` runs, and it examines the metadata in `args[0]`, and tweaks it if it finds it to be of a certain format. So the assumption is that `args[0]` is always something from loglevel, and `args[1]` is whatever is being logged, like `'foo'`.

However, when we have another copy of wdio-logger in play, it extends the first copy's version of `methodFactory` **rather than loglevel's version of `methodFactory`**. This causes loglevel's code to never run, causing there to be no metadata. Instead, when wdio-logger's code runs, the list is `['foo']`, and when it reads `args[0]` and finds it to be `'foo'`, it does nothing to it. But if the caller logged an error, e.g.
```js
try {
  // something that throws
} catch(e) {
  log.error(e)
}
```
... then `args[0]` is not `'foo'` but an error object, and wdio-logger's code blows up because it calls `endsWith` on it to decide if it can tweak it, and this method exists on strings, not error objects.

### What the best way to fix it?
Honestly, that's a question I'd like to defer for a moment. As someone who deeply appreciates a good automated test suite, I want to focus on how to capture this bug in a test. I see that we test these wdio packages in relative isolation. So I'm assuming that it's out of the question to set up a test that uses multiple wdio packages at the same time. So the alternative solution I came with is to just copy the wdio-logger source code, and then load the copy in addition to the original in the test. This reproduces the bug. However, it's ugly to see this checked into version control, not to mention it only works if we continue to update the copy every time we update the original.

**So I'd like to pose a question to this PR's reviewers: is there a better way to have a second copy of the source code for the sake of reproducing this bug?**

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments


mgrybyk:
It's difficult to understand what has changed. Also unit tests are failing

mgrybyk:
It's very easy to add simple type check to fix `args[0].endsWith is not a function`.

packages/wdio-logger/tests/node.test.js
```diff
@@ -1,5 +1,6 @@
 import fs from 'fs'
 import nodeLogger from '../src/node'
+import nodeLogger2 from '../src/nodeCopy'
```
johnnymo87: Hi @mgrybyk, sorry that it's not clear what I was asking you. I'm asking if you can think of _a better way_ to capture the bug in a failing test. Because I'd like to have a test that fails because it's capturing the bug before having a bug fix to make it pass.

I have a way to capture the bug using our existing tests (hence why they are failing), but I don't think it's the best way. What I am doing is importing a second copy of `src/node.js` in the test. This is my way of reproducing the bug. This bug is only reproducable when we run multiple copies of the same source code, as happens when multiple wdio packages declare wdio-logger as a dependency.

I've updated the PR description to explain this better.

christian-bromann:
> This is quite common, because wdio-logger is used by many other wdio-* packages.

They all use the same version though. You should only have one version installed. Can you try to update all wdio packages and retry?

johnnymo87:
> > This is quite common, because wdio-logger is used by many other wdio-* packages.
>
> They all use the same version though. You should only have one version installed. Can you try to update all wdio packages and retry?

OK, to sanity check that, I've wiped my `node_modules/` and re-installed. I've got the following number of copies of `@wdio/logger`:
```bash
$ find node_modules/ -wholename '*@wdio/logger/build/node.js'

node_modules/@wdio/cli/node_modules/@wdio/logger/build/node.js
node_modules/@wdio/config/node_modules/@wdio/logger/build/node.js
node_modules/@wdio/cucumber-framework/node_modules/@wdio/logger/build/node.js
node_modules/@wdio/local-runner/node_modules/@wdio/logger/build/node.js
node_modules/@wdio/logger/build/node.js
node_modules/@wdio/mocha-framework/node_modules/@wdio/logger/build/node.js
node_modules/@wdio/runner/node_modules/@wdio/logger/build/node.js
node_modules/@wdio/selenium-standalone-service/node_modules/@wdio/logger/build/node.js
node_modules/@wdio/sync/node_modules/@wdio/logger/build/node.js
node_modules/@wdio/utils/node_modules/@wdio/logger/build/node.js
node_modules/webdriver/node_modules/@wdio/logger/build/node.js
node_modules/webdriverio/node_modules/@wdio/logger/build/node.js
```
And they're all _pretty_ close in version. Here's a quick scan of their version numbers:
```bash
$ grep '@wdio/logger": "' package-lock.json

        "@wdio/logger": "5.16.5",
        "@wdio/logger": "5.16.5",
        "@wdio/logger": "5.16.5",
        "@wdio/logger": "5.16.5",
        "@wdio/logger": "^5.12.1",
            "@wdio/logger": "^5.12.1",
        "@wdio/logger": "5.16.5",
        "@wdio/logger": "5.16.5",
        "@wdio/logger": "5.16.5",
        "@wdio/logger": "5.16.5",
        "@wdio/logger": "^5.11.0",
        "@wdio/logger": "^5.8.0",
        "@wdio/logger": "5.16.5",
        "@wdio/logger": "5.16.5",
```
I've gone into each of these files, and renamed its `methodFactory` function to something unique, and console.logged it so I could see in what order my copies of wdio.logger are running, and whose version of `methodFactory` they're extending. Here's what I saw, in the order that I saw it.

* **@wdio/cli**
   - In `node_modules/@wdio/cli/node_modules/@wdio/logger/build/node.js`, the `methodFactory` that it extends is called `defaultMethodFactory`. This means it's [loglevel's version](https://github.com/pimterry/loglevel/blob/master/lib/loglevel.js#L112).
* **@wdio/config**
   - In `node_modules/@wdio/config/node_modules/@wdio/logger/build/node.js`, the `methodFactory` that it extends is called `methodFactory`. This means it's [loglevel-plugin-prefix's version](https://github.com/kutuluk/loglevel-plugin-prefix/blob/master/lib/loglevel-plugin-prefix.js#L61).
* **@wdio/utils**
   - In `node_modules/@wdio/utils/node_modules/@wdio/logger/build/node.js`, the `methodFactory` that it extends has the unique name from **@wdio/config**.
* **webdriver**
   - In `node_modules/webdriver/node_modules/@wdio/logger/build/node.js`, the `methodFactory` that it extends has the unique name from **@wdio/utils**.
* **webdriverio**
   - In `node_modules/webdriverio/node_modules/@wdio/logger/build/node.js`, the `methodFactory` that it extends has the unique name from **webdriver**.
* **@wdio/local-runner**
   - In `node_modules/@wdio/local-runner/node_modules/@wdio/logger/build/node.js`, the `methodFactory` that it extends has the unique name from **webdriverio**.

So I may be mistaken, but I think there is a bug here. I think that the intention of the code in `wdio-logger` is that it only ever extends the `methodFactory` from the `loglevel` npm packages. The fact that rather than doing this, one copy of `@wdio/logger` ends up extending the `methodFactory` function of another copy of `@wdio/logger`. And this introduces a path in the code that the code is not prepared for, hence `TypeError: args[0].endsWith is not a function`.

I've come up with a cleaner test solution. For the second copy of the code that I load to reproduce the bug, I source it from the `build` dir. No more need for checking in a new copy to version control. And the existing tests now catch the bug:

image

If we're all in agreement that this bug is real and that this method of testing is OK, I'll work on a solution.

johnnymo87:
OK @mgrybyk and @christian-bromann, I think this is ready for review! Please let me know what you both think.

### Testing
I have a test that uses a second copy of the source code, which is the way to reproduce the bug. I get this second copy of the source code from the `build/` directory. I've added this directory to the directories we ignore when scanning for code coverage.

### Documentation
I haven't added any comments to the source code, or any new information to the packages readme. However, I'd be happy to if we feel we need to.

### Solutions
I've come up with three potential solutions. Solution 1 is checked in to this PR, but I can switch to either of the other two as requested. And I can pursue different ones if you all have suggestions.

#### Solution 1
This solution is the smallest non-whitespace change possible. It guards the `log.methodFactory = ` line with a conditional that checks if the `originalMethodFactory` is that of [`loglevel`'s](https://github.com/pimterry/loglevel/blob/fdd97e50f5d3cfdaf73a84001f6326e122303947/lib/loglevel.js#L112-L116). The reason we do this is because we don't want to overwrite `log.methodFactory` on the 2nd, 3rd, nth time we load this source code, because doing so is the root cause of the bug.
```diff
diff --git a/packages/wdio-logger/src/node.js b/packages/wdio-logger/src/node.js
index 564882e9..f9445db8 100644
--- a/packages/wdio-logger/src/node.js
+++ b/packages/wdio-logger/src/node.js
@@ -54,6 +54,7 @@ const logCache = new Set()
 let logFile
 
 const originalFactory = log.methodFactory
+if (originalFactory.name === 'defaultMethodFactory') {
     log.methodFactory = function (methodName, logLevel, loggerName) {
         const rawMethod = originalFactory(methodName, logLevel, loggerName)
         return (...args) => {
@@ -100,6 +101,7 @@ log.methodFactory = function (methodName, logLevel, loggerName) {
             rawMethod(...args)
         }
     }
+}
 
 prefix.apply(log, {
     template: '%t %l %n:',
```

#### Solution 2
This solution attempts to improve on solution 1 by not relying on knowledge that `loglevel`'s name for `methodFactory` is `'defaultMethodFactory'`. Instead it produces a logger that has this information inside it, and compares against that.
```diff
diff --git a/packages/wdio-logger/src/node.js b/packages/wdio-logger/src/node.js
index 564882e9..ef30c961 100644
--- a/packages/wdio-logger/src/node.js
+++ b/packages/wdio-logger/src/node.js
@@ -54,6 +54,7 @@ const logCache = new Set()
 let logFile
 
 const originalFactory = log.methodFactory
+if (originalFactory.name === log.getLogger('_').methodFactory.name) {
     log.methodFactory = function (methodName, logLevel, loggerName) {
         const rawMethod = originalFactory(methodName, logLevel, loggerName)
         return (...args) => {
@@ -100,6 +101,7 @@ log.methodFactory = function (methodName, logLevel, loggerName) {
             rawMethod(...args)
         }
     }
+}
 
 prefix.apply(log, {
     template: '%t %l %n:',
```

#### Solution 3
This solution completely avoids checking the name of the `methodFactory` function. Instead, it avoids overwriting `methodFactory` more than once by only ever overwriting it on the "children" of the root logger, rather than overwriting it on the root logger itself. I had to change its definition of `const loggers =` so that all copies of `wdio-logger` share the same children, otherwise this solution fails.
```diff
diff --git a/packages/wdio-logger/src/node.js b/packages/wdio-logger/src/node.js
index 564882e9..e3abbfd6 100644
--- a/packages/wdio-logger/src/node.js
+++ b/packages/wdio-logger/src/node.js
@@ -48,13 +48,13 @@ const SERIALIZERS = [{
     serialize: (log) => chalk.cyan(log)
 }]
 
-const loggers = {}
+const loggers = log.getLoggers()
 let logLevelsConfig = {}
 const logCache = new Set()
 let logFile
 
 const originalFactory = log.methodFactory
-log.methodFactory = function (methodName, logLevel, loggerName) {
+const wdioLoggerMethodFactory = function (methodName, logLevel, loggerName) {
     const rawMethod = originalFactory(methodName, logLevel, loggerName)
     return (...args) => {
         /**
@@ -101,13 +101,6 @@ log.methodFactory = function (methodName, logLevel, loggerName) {
     }
 }
 
-prefix.apply(log, {
-    template: '%t %l %n:',
-    timestampFormatter: (date) => chalk.gray(date.toISOString()),
-    levelFormatter: (level) => chalk[COLORS[level]](level.toUpperCase()),
-    nameFormatter: (name) => chalk.whiteBright(name)
-})
-
 export default function getLogger (name) {
     /**
      * check if logger was already initiated
@@ -124,6 +117,13 @@ export default function getLogger (name) {
 
     loggers[name] = log.getLogger(name)
     loggers[name].setLevel(logLevel)
+    loggers[name].methodFactory = wdioLoggerMethodFactory
+    prefix.apply(loggers[name], {
+        template: '%t %l %n:',
+        timestampFormatter: (date) => chalk.gray(date.toISOString()),
+        levelFormatter: (level) => chalk[COLORS[level]](level.toUpperCase()),
+        nameFormatter: (name) => chalk.whiteBright(name)
+    })
     return loggers[name]
 }
 /**
```

jlin-namely:
I also see a lof of `TypeError: args[0].endsWith is not a function` error disguising real javascript uncaught error.  I also vote to fix this.

mgrybyk:
I don't mind fixing it but not sure about the approach chosen

johnnymo87:
I've been poking around the issues boards of the github repositories of `loglevel` and `loglevel-plugin-prefix`, and I've found [a conversation](https://github.com/pimterry/loglevel/issues/117) that corroborates what I'm seeing. In short, things can go wrong if we don't take care when prefixing the `methodFactory` function. According to that conversation, the recommended approach is to avoid prefixing the function on the root logger, and to only do it on its children.

When I talk about prefixing, this is what I mean.

Let's say a random wdio packages uses `@wdio/logger` to log an error:
```
const log = logger('test')
log.error(new Error('boom'))
```
Inside `loglevel`, `loglevel-plugin-prefix`, and `@wdio/logger`, the definition of the `error`, `warn`, `info`, `debug`, and `trace` functions are all meta-programmed by a function called `methodFactory`. The `methodFactory` function handles the logged information in a variable named `args` that looks like this:
```
[new Error('boom')]
```
Backing up, when someone imports `@wdio/logger`, it imports `loglevel`, and thus `@wdio/logger` starts out with `loglevel`'s original definition of `methodFactory`.
```
loglevel
```
As `@wdio/logger` continues to initialize itself, it prefixes `loglevel`'s definition of `methodFactory` with its own.
```
@wdio/logger + loglevel
```
And then it prefixes it with that of `loglevel-plugin-prefix`.
```
loglevel-plugin-prefix + @wdio/logger + loglevel
```

This order of prefixing is important because the `@wdio/logger` portion of `methodFactory` assumes that `loglevel-plugin-prefix`'s portion has transformed `args` from `[new Error('boom')]` to `['some string metadata', new Error('boom')]`. In `@wdio/logger`'s portition,  it reaches into what `loglevel-plugin-prefix` has done to `args[0]` and further changes it.

This works well when `@wdio/logger` is loaded only once. However, when it's loaded, say, four times by four different wdio packages, then the prefixing looks like this:
```
@wdio/logger4 + @wdio/logger3 + @wdio/logger2 + loglevel-plugin-prefix + @wdio/logger1 + loglevel
```
This is due to the fact that when the second, third, and fourth copies of `@wdio/logger` load their version of `loglevel`, it no longer has its original definition of `methodFactory`. Rather it has one that has already been prefixed by the first copy of `@wdio/logger`. And these unwanted extra `@wdio/logger` prefixes blow up because they're not preceded by a `loglevel-plugin-prefix` portion.

My goal is to eliminate these unwanted prefixes. Again, according to [this conversation](https://github.com/pimterry/loglevel/issues/117), it looks like the recommended technique is to do prefixing on children rather than the root logger. This works because all three libraries cache their children. And when they're called for, they return them without re-initializing or re-prefixing them in any way.

@mgrybyk, please let me know if this explanation makes any sense, and if you've got any alternative ideas for how to approach this problem.

Kignuf:
This bug has driven me crazy for a couple of days, thanks so much for the deep investigation and the fix @johnnymo87

Issue #4796, opened 2019-11-22
isClickable() returns false on elements inside a shadowRoot


lacell75:
**Environment (please complete the following information):**
 - **WebdriverIO version:** 5.16.7 
 - **Mode:** WDIO
 - **If WDIO Testrunner, running sync/async:** sync
 - **Node.js version:** 12.6.0
 - **NPM version:** 6.9.0
 - **Browser name and version:** ALL (Chrome 78 / firefox /Microsoft Edge 44.17763.771.0)
 - **Platform name and version:**  Windows 10
 - **Additional wdio packages used (if applicable):** "@wdio/cli": "^5.16.0",
    "@wdio/cucumber-framework": "^5.16.6",
    "@wdio/dot-reporter": "^5.16.5",
    "@wdio/local-runner": "^5.16.7",
    "@wdio/mocha-framework": "^5.16.5",
    "@wdio/selenium-standalone-service": "^5.16.5",
    "@wdio/spec-reporter": "^5.16.5",
    "@wdio/sync": "^5.16.5",

**Describe the bug**
isClickable() method returns false even if webdriver can click on it (I use an element's shadowRoot)

**To Reproduce**
```
describe('Try to click button in shadowRoot', () => {
  it('Try to click button in shadowRoot', () => {
      browser.url('http://www.kaidez.com/samples/template-shadowdom-practice/');
  const allbooks = browser.$('#allBooks');
  const button = allbooks.shadow$('#btnPurchase');
  console.log(button.isDisplayed()); // true
  console.log(button.isDisplayedInViewport()); // true
  console.log(button.isEnabled()); // true
  console.log(button.isClickable()); // false
  button.waitForClickable({ timeout: 10000 });
  button.click();
  });
});
```

**Expected behavior**
Should return true as element clearly clickable and webdriver can click on it

**Log**
```
Execution of 1 spec files started at 2019-11-22T11:11:30.481Z

[0-0] RUNNING in chrome - C:\Users\fcellier\Documents\02-Test_Auto\end-to-end-tests-wdio5\features\test.feature
[0-0] true
[0-0] true
[0-0] true
[0-0] false
[0-0] Error in "Try to click button in shadowRoot: Try to click button in shadowRoot: Given Try to click button in shadowRoot"
element ("function() {
      // element has a shadowRoot property
      if (this.shadowRoot) {
        return this.shadowRoot.querySelector('#btnPurchase')
      }
      // fall back to querying the element directly if not
[...]

PR #4799, opened 2019-11-22
Fix/issue/4796

jimmy-collazos:
### package webdriverio
Fix problem with `isElementClickable()` method.

This function use [element.contains()](https://developer.mozilla.org/en/docs/Web/API/Node/contains) to check if element is inside of parent elemente; but this function not found when element (child) is in ShadowRoot (DocumentFragment) node.

[fixes#4796](https://github.com/webdriverio/webdriverio/issues/4796)

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/packages/webdriverio/src/scripts/isElementClickable.js
+++ b/packages/webdriverio/src/scripts/isElementClickable.js
@@ -4,7 +4,19 @@
  * @return {Boolean}           false if element is not overlapped
  */
 export default function isElementClickable (elem) {
-    if (!elem.getBoundingClientRect || !elem.scrollIntoView || !elem.contains || !elem.getClientRects || !document.elementFromPoint) {
+    if (!elem.getBoundingClientRect || !elem.scrollIntoView || !elem.getClientRects || !document.elementFromPoint) {
+        return false
+    }
+
+    // check if the childElement is contained in containerElement with ShadowDom support
+    function nodeContains(containerElement, childElement){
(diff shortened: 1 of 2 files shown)


mgrybyk:
I'm not sure about the approach chosen. It doesn't guarantee that element is clickable but only makes sure it is inside shadowRoot

mgrybyk:
I'll raise a PR that should handle it

mgrybyk:
https://github.com/webdriverio/webdriverio/pull/4800

christian-bromann:
@jimmy-collazos can you take a look into #4800 and provide feedback?

jimmy-collazos:
@christian-bromann sorry; but i'm busy... family time. The proposal of #4800  is correct.

packages/webdriverio/src/scripts/isElementClickable.js
```diff
@@ -4,7 +4,19 @@
  * @return {Boolean}           false if element is not overlapped
  */
 export default function isElementClickable (elem) {
-    if (!elem.getBoundingClientRect || !elem.scrollIntoView || !elem.contains || !elem.getClientRects || !document.elementFromPoint) {
+    if (!elem.getBoundingClientRect || !elem.scrollIntoView || !elem.getClientRects || !document.elementFromPoint) {
+        return false
+    }
+
+    // check if the childElement is contained in containerElement with ShadowDom support
```
mgrybyk: @jimmy-collazos the solution you've proposed works great for Edge because it doesn't support shadow dom, polyfills are used instead like shadydom.
I'd like to use it to fix https://github.com/webdriverio/webdriverio/issues/4804

PR #4802, opened 2019-11-23
Proposal to add element equals command

mgrybyk:
## Proposed changes

Check if two elements are equal (web only)
```
$('foo').equals($('foo')) // true
$('foo').equals($('bar')) // false
```

## Types of changes

- [ ] Documentation update

## Checklist


mgrybyk:
The use case might be element order change or making sure it has changed after previous action. 
I understand that it might be very specific use case and don't mind closing the PR.

1. I'll rename it
2. Will try to make it work on mobile

mgrybyk:
A bit busy right now will come back to it as soon as I can. If anyone wants pick it up feel free!

packages/webdriverio/src/commands/element/isEqual.js
```diff
@@ -0,0 +1,64 @@
...
+    const browser = getBrowserObject(this)
+
+    // mobile native
+    if (browser.isMobile) {
+        const context = await browser.getContext()
+        if (context.toLowerCase().includes('native')) {
+            const currentId = this.elementId
+            // just call any command that would refetch element if needed
+            try {
+                await this.getTagName()
```
mgrybyk: @christian-bromann do you think it's a correct way to do refetch?
I'm afraid that the element was found not with just `$(selector)` but with some custom strategy  or stuff like that

mgrybyk: ignore it, I don't think we need to refetch anything at this point

packages/webdriverio/src/commands/element/isEqual.js
```diff
@@ -0,0 +1,56 @@
+
```
christian-bromann:
```suggestion
```

packages/webdriverio/src/commands/element/isEqual.js
```diff
@@ -0,0 +1,56 @@
...
+/**
+ *
+ * Return true if the selected DOM-element matches with the provided one.
```
christian-bromann: As this is also supported for mobile we shouldn't restrict the description to the DOM API.

PR #4766, opened 2019-11-28
Inconsistency between webdriver and devtools getElementText command

christian-bromann:
 - **WebdriverIO version:** 5.16.5

**Config of WebdriverIO**
`automationProtocom: 'devtools'`

**Describe the bug**
`getText` returns content content that is not visible to user in devtools protocol

**To Reproduce**
```
it('foo', () => {
    browser.url("https://fiddle.jshell.net/xlenz/1rkLjctp/show/")
    $("#result iframe").waitForDisplayed()
    browser.switchToFrame($("#result iframe"))

    const text = $('#t').getText()
    console.log(text)
});
```

ACTUAL result:
``` 

lions,
tigers
and bears

```

**Expected behavior**
`lions, tigers`

**Additional context**
Shall we use `innerText` instead of `textContent` here https://github.com/webdriverio/webdriverio/blob/e30649c9151aebc034a935db1e8399149ddf6078/packages/devtools/src/scripts/getElementText.js#L2 ?

### DIFF
--- a/e2e/protocol.test.js
+++ b/e2e/protocol.test.js
@@ -23,9 +23,8 @@ it('should navigate to a page and get page info', async () => {
     expect(await browser.getUrl()).toContain('http://guinea-pig.webdriver.io')
     expect(await browser.getPageSource()).toContain('WebdriverJS Testpage')
 
-    const elem = await browser.$('#t')
-    const text = await elem.getText()
-    expect(text).toBe('lions, tigers')
+    const elem = await browser.findElement('css selector', '#t')
+    expect(await browser.getElementText(elem[ELEMENT_KEY])).toBe('lions, tigers')
 })
 
 it('should include the hash', async () => {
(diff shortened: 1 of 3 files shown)


christian-bromann:
@mgrybyk updated the `devtools` script and added test for it. Please review!

e2e/protocol.test.js
```diff
@@ -22,6 +22,10 @@ it('should navigate to a page and get page info', async () => {
     expect(await browser.getTitle()).toBe('WebdriverJS Testpage')
     expect(await browser.getUrl()).toContain('http://guinea-pig.webdriver.io')
     expect(await browser.getPageSource()).toContain('WebdriverJS Testpage')
+
+    const elem = await browser.$('#t')
```
mgrybyk: I think we don't have $ here

mgrybyk:
👍 

can't approve my own issue, you may merge if whenever ready

PR #4853, opened 2019-12-04
Sync vs Async test execution equality

### LINKED ISSUE #4849: beforeCommand, afterCommand do not work when use async
Hello,

I'm trying to use beforeCommand and afterCommand hooks in order to catch browser.url command. unfortunately, it does not work.

```
    /**
     * Runs before a WebdriverIO command gets executed.
     * @param {String} commandName hook command name
     * @param {Array} args arguments that command would receive
     */
    beforeCommand: function (commandName, args) {
//something
    },

    /**
     * Runs after a WebdriverIO command gets executed
     * @param {String} commandName hook command name
     * @param {Array} args arguments that command would receive
     * @param {Number} result 0 - command success, 1 - command error
     * @param {Object} error error object if any
     */
     afterCommand: function (commandName, args, result, error) {
//something
     },

```

Did I miss something ?

Thx

### LINKED ISSUE #4851: Access number of retries per test automatically

If someone would be interested how often the retry happened within the test he would have to build his own counter. Given that we have the data we should provide that data automatically.

Add number of retries to the function scope , e.g. `this.retries`

none.

christian-bromann:
## Proposed changes

Currently there is a difference when running tests synchronously (with `@wdio/sync`) and asynchronously (without `@wdio/sync`):

- tests are not being retried
- no `beforeCommand` and `afterCommand` hooks are run

## Types of changes

This patch fixes the behavior to make it the same. It also adds `retries` to the function scope so that user have an idea how many times the test was already run.

- [ ] Documentation update

## Checklist

## Further comments

closes #4851
closes #4849

### DIFF
--- a/docs/Retry.md
+++ b/docs/Retry.md
@@ -7,7 +7,7 @@ You can rerun certain tests with the WebdriverIO testrunner that turn out to be 
 
 ## Rerun suites in Mocha
 
-Since version 3 of Mocha, you can rerun whole test suites (everything inside an `describe` block). If you use Mocha you should favor this retry mechanism instead of the WebdriverIO implementation that only allows you to rerun certain test blocks (everything within an `it` block). In order to use the `this.retries()` method, the suite block `describe` must use an unbound function `function(){}` instead of a fat arrow function `()=>{}`, as described in [Mocha docs](https://mochajs.org/#arrow-functions).
+Since version 3 of Mocha, you can rerun whole test suites (everything inside an `describe` block). If you use Mocha you should favor this retry mechanism instead of the WebdriverIO implementation that only allows you to rerun certain test blocks (everything within an `it` block). In order to use the `this.retries()` method, the suite block `describe` must use an unbound function `function(){}` instead of a fat arrow function `() => {}`, as described in [Mocha docs](https://mochajs.org/#arrow-functions). Using Mocha you can also set a retry count for all specs using `mochaOpts.retries` in your `wdio.conf.js`.
 
 Here is an example:
 
@@ -29,7 +29,7 @@ describe('retries', function() {
 })
 ```
 
-## Rerun single tests in Jasmine or Mocha (`@wdio/sync` only)
+## Rerun single tests in Jasmine or Mocha
 
 To rerun a certain test block you can just apply the number of reruns as last parameter after the test block function:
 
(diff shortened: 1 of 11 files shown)


docs/Retry.md
```diff
@@ -7,8 +7,7 @@ You can rerun certain tests with the WebdriverIO testrunner that turn out to be
 
 ## Rerun suites in Mocha
 
-Since version 3 of Mocha, you can rerun whole test suites (everything inside an `describe` block). If you use Mocha you should favor this retry mechanism instead of the WebdriverIO implementation that only allows you to rerun certain test blocks (everything within an `it` block).  
-In order to use the `this.retries()` method, the suite block `describe` must use an unbound function `function(){}` instead of a fat arrow function `()=>{}`, as described in [Mocha docs](https://mochajs.org/#arrow-functions).
+Since version 3 of Mocha, you can rerun whole test suites (everything inside an `describe` block). If you use Mocha you should favor this retry mechanism instead of the WebdriverIO implementation that only allows you to rerun certain test blocks (everything within an `it` block). In order to use the `this.retries()` method, the suite block `describe` must use an unbound function `function(){}` instead of a fat arrow function `()=>{}`, as described in [Mocha docs](https://mochajs.org/#arrow-functions).
```
erwinheitzman: it's also possible to set it like so if I am not mistaken:
```
mochaOpts: {
  retries: 3
}
```

christian-bromann: That is correct.

christian-bromann: will do

packages/wdio-sync/src/index.js
```diff
@@ -34,7 +35,7 @@ const executeSync = async function (fn, retries, args = []) {
     } catch (e) {
         if (retries.limit > retries.attempts) {
             retries.attempts++
```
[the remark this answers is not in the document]
christian-bromann: it has to use `retries` as it is passed in as a parameter

erwinheitzman:
Last question is if we need to extend the Mocha test context or the WebdriverIO test object for example to have the retries property reflected?

christian-bromann:
Are you talking about the typescript definition for the test context? I don't think we have that nor do we have these for the `it` functions (including the retry object that differs from the original usage). We should create a new ticket for that.

PR #4861, opened 2019-12-06
Fix naming of retry variable

christian-bromann:
## Proposed changes

We still experienced issues with running WebdriverIO in standalone mode. This patch is suppose to fix these once and for all.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments


packages/wdio-sync/src/index.js
```diff
@@ -75,3 +75,9 @@ export {
     executeSync,
     runSync,
 }
+
+export default function sync (testFn) {
```
mgrybyk: why do we need another export?

christian-bromann: given that @wdio/sync is no suppose to be used by end user I thought it would be nice if people can just do import sync from '@wdio/sync' when they want to use it.

jlin-namely:
Got a question.  Is there to add a timeout between retry?  If not, can we add this feature too?

erwinheitzman:
@jlin-namely that would be a seperate feature we would implement, we are currently fixing some regression with this. also why would you want a interval?

jlin-namely:
Retry is usually used when a process is asynchronous but usually, there is still approximate ballpark like 10 secs, 30 secs, 1 min or 5 mins, and etc...   
For example, I know that there is a 95% chance to finish within 30 secs.  I would like to set retry to 3 times with an interval of 10 secs.  Without specifying an interval, I have to do trial and error with a number of retries.  Currently, I have to put a browser.pause() at the beginning of test scenario to accomplish the same thing but I don't want to boilerplate code and I only need timeout if there is retry.

erwinheitzman:
@jlin-namely  I think you are confusing this functionality for something else, the retries have got nothing to do with async code specifically. Let's continue the conversation on Gitter as this is not something to discuss in this PR :)

jlin-namely:
@erwinheitzman we are talking about test retry, right?  Just explain a little better what I am talking about.  For example, I am testing UI triggering an email and I am using email API to retrieve the response.  Since there is no guarantee of how long email will take to send (could be 5-sec to 1 min), I want to retry 5-sec interval up to 60 secs.  I can't really do that right now.  Here is what I like to do:  ```    this.Given(/^some step definition$/, { wrapperOptions: { retry: 2, interval: 5000 } }, () => {
        // ...
    })```

mgrybyk:
@jlin-namely this is not a right place to discuss it anyway

erwinheitzman:
These changes seem to break the hooks for named custom services, can anyone confirm this?

mgrybyk:
Overall LGTM, but let's fix the failing tests

packages/wdio-utils/src/shim.js
```diff
@@ -146,14 +146,21 @@ const executeAsync = async function (fn, retries, args = []) {
...
-    executeSync = wdioSync.executeSync
-    runSync = wdioSync.runSync
+    /**
+     * only require `@wdio/sync` if `WDIO_NO_SYNC_SUPPORT` which allows us to
+     * create a smoke test scenario to test actual absence of the package
+     * (internal use only)
+     */
+    if (!process.env.WDIO_NO_SYNC_SUPPORT) {
+        // eslint-disable-next-line import/no-unresolved
+        const wdioSync = require('@wdio/sync')
```
CrispusDH: should be here `const wdioSync = require('@wdio/sync').default` ?

christian-bromann: This would give me the reference to the [`sync`](https://github.com/webdriverio/webdriverio/blob/cb-retry-fix/packages/wdio-sync/src/index.js#L79-L83) method. I need to exported functions above though.

tests/smoke.runner.js
```diff
@@ -35,7 +35,9 @@ async function runTests (tests) {
         }
     } else {
         // parallel
-        await Promise.all(tests.map(test => test()))
+        await Promise.all(tests.map(test => test().catch(() => {
```
CrispusDH: will it catch just one failed test? Maybe, here might be better to use `Promise.allSettled()`?

christian-bromann: I would prefer to stop running other tests once one smoke test is failing.

christian-bromann:
@CrispusDH can you approve/reject the PR if you have no further comments?

tests/smoke.runner.js
mgrybyk: we cant use allSettled because it was added in nodejs 12 !!

Issue #4880, opened 2019-12-14
Cache list of spec files


gustavohenke:
**Is your feature request related to a problem? Please describe.**
Finding specs in a big monorepo takes ages. Think 130K+ files (sources + tests + assets).
Rough directory structure is this:
```
page1/
	- features/
		page1.feature
	- subpage1/
		- features/
			- subpage1.feature
		- subpage1.js
	- page1.js
	- page1.css
page2/
	<same structure>
ui/shared_component/
	- features/
		- shared_component_in_page1.feature
		- shared_component_in_page2.feature
	- shared_component.js
```

This is a screenshot from a JS CPU profile of the first ~2 minutes of running WDIO on our repo. At that point, it didn't even get to actually running the tests.

<img width="1474" alt="Screen Shot 2019-12-14 at 19 55 39" src="https://user-images.githubusercontent.com/826553/70846260-b994d080-1eab-11ea-9bf8-c316a773398b.png">

Note how consistently [`ConfigParser#getSpecs`](https://github.com/webdriverio/webdriverio/blob/f89319bb91a35f99bda361b9bcdb95e7a7c5d35d/packages/wdio-config/src/lib/ConfigParser.js#L167) appears in the chart, and how much time it consumes.

This massive time is to find just ~250 `.feature` files among the sources.
I also had passed a tag expression that would filter that down to just 4 files, so it's unacceptable that WDIO takes this long.

**Describe the solution you'd like**
`ConfigParser#getSpecs` could be memoized, so that when it takes the same input, it always returns the same value. Should shave a bunch of time from the 

I guess that adding/removing files could become a problem with that solution, in which case it should be coupled with a file system watching solution to manipulate the cache.

**Describe alternatives you've considered**
The `specs` option was too inclusive (`src/**/features/*.feature`), so I tinkered with the `suites` option:

```js
suites: {
	page1: ['src/page1/**/features/*.feature', 'src/ui/**/features/*.feature']
	// and so on
}
```

This didn't help bringing the time to boot up down noticeably, though.

And, as probably obvious from above, keeping a list of `.feature` files in the config that doesn't make use of wildcards is impossible.

**Additional context**
Node.js v10.16.3

`@wdio` packages used:
```json
    "@wdio/cli": "^5.16.10",
    "@wdio/cucumber-framework": "^5.16.10",
    "@wdio/local-runner": "^5.16.10",
    "@wdio/selenium-standalone-service": "^5.16.10",
[...]

PR #4881, opened 2019-12-14
Minor cleanup to avoid `getSpecs` call

christian-bromann:
## Proposed changes

While this is not a solution to #4880 it is a first small step.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

I will try to see if there are more improvements to make.

### DIFF
--- a/packages/wdio-cli/src/interface.js
+++ b/packages/wdio-cli/src/interface.js
@@ -7,7 +7,7 @@ import { getRunnerName } from './utils'
 const log = logger('@wdio/cli')
 
 export default class WDIOCLInterface extends EventEmitter {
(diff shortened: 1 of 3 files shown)


packages/wdio-cli/src/interface.js
```diff
@@ -16,7 +16,6 @@ export default class WDIOCLInterface extends EventEmitter {
          * `FORCE_COLOR=0` - forcibly disable colors
          */
         this.hasAnsiSupport = !!chalk.supportsColor.hasBasic
-        this.specs = specs
```
mgrybyk: We don't need to have specs here?

christian-bromann: nope, `this.specs` is not used anywhere

Issue #4788, opened 2019-11-19
WDIO freezes VS Code debugger upon use of WDIO commands


Khromchenko:
I'm able to use VS Code debugger but with no ability to use WDIO commands. Any use of WDIO apis such as browser.$()... freezes VS Code debugger.

Here is my launch.json

```
{
    "version": "0.2.0",
    "configurations": [
        {
            "type": "node",
            "request": "launch",
            "name": "WDIO",
            "program": "${workspaceFolder}/node_modules/.bin/wdio",
            "port": 5859,
            "autoAttachChildProcesses": true,
            "protocol": "inspector",
            "args": [
                "wdio.conf.js",
                "--spec",
                "spec/login/login-spec.js" // File which you would like to debug
            ],
            "internalConsoleOptions": "openOnSessionStart",
            "cwd": "${workspaceRoot}",
            "env": {
                "DEBUG": "1" 
                // use an environment variable to be able
                // to toggle debug mode on and off
            }
        }
    ]
}
```

wdio.conf.js inludes following:

```
debug: true,
execArgv: ['--inspect-brk=127.0.0.1:5859'],
```

**Is there a way to set up VS Code to be able to use all WDIO apis?** 
Just like we would use browser.debug()

**Environment:**
 - **WebdriverIO version:** 5.16.6
 - **Node.js version:** v12.12.0
 - **NPM version:** 6.13.0
 - **Browser name and version:** Chrome
 - **Platform name and version:** UNIX

Khromchenko:
Just FYI, Looks like it worked in WDIO 4 for this guy:
http://blog.likewise.org/2017/02/debugging-a-javascript-webdriverio-project-in-vscode/

mgrybyk:
You may use browser.debug instead. The behavior you see is how fibers work. I think there is nothing can be done here

christian-bromann:
Being able to use VSCode debugger capabilities should be something that is suppose to work. Any help on this is appreciated.

mgrybyk:
Don't know how to deal with it, let's ask node-fibers team to help https://github.com/laverdet/node-fibers/issues/425

mgrybyk:
closing because @christian-bromann replaces node-fibers with deasync https://github.com/webdriverio/webdriverio/pull/4886

johnnymo87:
Since @christian-bromann closed that PR without merging it, should we open this issue back up @mgrybyk ?

christian-bromann:
[...]

PR #4886, opened 2019-12-15
Replace `node-fibers` with `deasync`

christian-bromann:
Due to issue laverdet/node-fibers#428 it seems to be a good idea to move away from it and use `deasync` instead which seems to be much better maintained and:

> You may tempted to use node-fibers or a module derived from it, but node fibers can only wrap async function call into a sync function inside a fiber. In the case above you cannot assume all callers are inside fibers. On the other hand, if you start a fiber in getData then getData itself will still return immediately without waiting for the async call result. For similar reason ES6 generators introduced in Node v0.11 won't work either.

Given that we would need to handle sync vs async as a setting I think we should move it to v6.


christian-bromann:
> Are you aware of any other alternatives? What I see is that it's generally not recommended to use, same as fibers, I think.

`deasync` is the alternative I found. It seems like a much better implementation of the same idea however I encountered some issues I have to figure out, e.g.:

- `deasync` doesn't work in Node workers (not sure if that is important to us)
- I ran into issues when running smoke tests: `async hook stack has become corrupted (actual: 209, expected: 210)`

Let's hope these issues can be fixed.

CrispusDH:
link with article what can go wrong with using `deasync` - [article](https://joecreager.com/5-reasons-to-avoid-deasync-for-node-js/#return-note-1129-6)

mgrybyk:
Is it possible to debug properly https://github.com/webdriverio/webdriverio/issues/4788 ?

christian-bromann:
> link with article what can go wrong with using `deasync` - [article](https://joecreager.com/5-reasons-to-avoid-deasync-for-node-js/#return-note-1129-6)

Thanks for sharing the article. I looked into the claims the author made and came to the following conclusion:

- re 1) the comment from the original author was in 2015 which is a decade ago in Node.js years. A lot of things have changed since then.
- re 2) I have mixed feelings about the claim that the library is not stable, they [switched](https://github.com/abbr/deasync/pull/102/files) at some point to N-API which provides a stable API for Node internals - therefor they don't use internal Node APIs anymore (which was the main concern of the Node maintainers why it is not safe to use `deasync`) and rely on an official API that has LTS
- re 3) Apart from the fact that we don't run WebdriverIO in application code I tried the example that has been mentioned in the blog posts and it worked for me as expected (did not hang indefinitely)
- re debugging) yes, I was able to use [Kevins video](https://www.youtube.com/watch?v=7ewSTCeI3W0&t=475s) to debug my tests using VSCode: 
Screen Shot 2019-12-15 at 9 59 17 AM

I can't tell for sure if the library is 100% secure and solid but it makes the impression to me. Looking at the PR I believe it makes the code so much simpler and less error prone than Fibers (given Fibers only supports one fibers context). Let's make sure we properly test this across our production test suites and make a decision then.

laverdet:
Hi just stopping by from the other issue. 

> deasync doesn't work in Node workers (not sure if that is important to us)

```
-   uv_run(uv_default_loop(), UV_RUN_ONCE);
+   uv_run(GetCurrentEventLoop(v8::Isolate::GetCurrent()), UV_RUN_ONCE);
```

Applying this change to deasync.cc should fix deasync in worker_threads.

> re 2) I have mixed feelings about the claim that the library is not stable, they switched at some point to N-API which provides a stable API for Node internals - therefor they don't use internal Node APIs anymore (which was the main concern of the Node maintainers why it is not safe to use deasync) and rely on an official API that has LTS

The issue is that `uv_run` is not reentrant. This is documented in libuv:

http://docs.libuv.org/en/v1.x/loop.html

> uv_run() is not reentrant. It must not be called from a callback.

Switching to N-API was just a style choice which doesn't change this invariant. Deasync clearly is still calling uv_run in a way that violates this and could explode at any time: https://github.com/abbr/deasync/blob/master/src/deasync.cc

As someone who has lived in this rabbit hole for 8 years I would definitely advise to avoid both fibers and deasync if at all possible :)

christian-bromann:
@laverdet thanks for your input. I would definitely not use these packages for any application development and users of WebdriverIO still have the choice to not install `@wdio/sync`. That said, would you say that using it in this context of making asynchronous automation commands synchronous to simplify tests is a valid use case for such libraries?

> Applying this change to deasync.cc should fix deasync in worker_threads.

Awesome, I will make a PR in `deasync`.

laverdet:
> Awesome, I will make a PR in deasync.

You'll need keep the old code behind `#if NODE_MODULE_VERSION < 64` to avoid dropping support for nodejs v9 and below.

>  That said, would you say that using it in this context of making asynchronous automation commands synchronous to simplify tests is a valid use case for such libraries?

As a case study the node-sass project recently ran into issues because node-fibers didn't have support for nodejs v13.0.0 right away. Since they couldn't test coverage for the new version their release was delayed until I updated fibers. Only you can make the call of whether or not it's worth it. But do know that when you run `npm install fibers` or `npm install desync` you're making a deal with the devil and he'll come back eventually to reap your soul. Just make sure you have an escape plan in place for the day it all comes burning down.

christian-bromann:
@laverdet thanks for your input, really appreciate it.

Looking at `deasync` they seem to be fairly quick with updating Node.js bundles. For Node v13 (released 10/10/2019) they pushed a release on [November 16th](https://github.com/abbr/deasync/commit/d6b7e51670977391368475f5b0438955f45e4943). Given that WebdriverIO users don't necessarily need the latest Node version for their tests, this limitation seems fine to me.

> Only you can make the call of whether or not it's worth it.

Looking back it didn't felt like it was any disadvantage to transform commands to run synchronously. The opposite is the case. A lot of people from other languages like Java didn't had the pain to handle promises all the time. In many [blog posts](https://moduscreate.com/blog/automated-tests-for-react-js-web-apps-using-webdriverio/) I've seen synchronous commands to be evaluated as very positive and noted as simplification compared to other frameworks.

> you're making a deal with the devil and he'll come back eventually to reap your soul

By now my soul has been reaped many times already. And quite frankly I enjoyed it 😉 in the many years we have been using `node-fibers` it benefited us and made a lot of code and tests simpler. I strongly advocate to continue this devil dance. That said: we should definitely make this clear to users that dance the devil dance too and ensure they understand the possible consequences (e.g. new Node.js version isn't available right after the release). I will make sure to document this properly so that users can take a deliberate decision whether they want to opt-in or not.

Thanks for your feedback!

christian-bromann:
@webdriverio/project-committers this is now ready for review

e2e/transitions.test.js
```diff
@@ -1,6 +1,8 @@
 import DevTools from '../packages/devtools/src/index'
 import { ELEMENT_KEY } from '../packages/devtools/src/constants'
 
+jest.setTimeout(30000)
```
mgrybyk: Does it make sense to set timeout once for every all tests?

packages/wdio-runner/src/index.js
```diff
@@ -382,6 +377,7 @@ export default class Runner extends EventEmitter {
         }
 
         await global.browser.deleteSession()
+        process.env.WDIO_SYNC_ENABLED = false
```
mgrybyk: Is it allowed to use boolean here? I assumed that only strings are allowed

packages/wdio-sync/src/index.js
```diff
@@ -1,83 +1,31 @@
...
-     * synchronously in standalone mode. In this case we neither have
-     * `global.browser` nor `this`
-     */
-    if (global.browser) {
-        delete global.browser._NOT_FIBER
-    }
-    if (this) {
-        this.wdioRetries = retries.attempts
+export default function waitForPromise (promise) {
+    if (!(promise instanceof Promise) || !process.env.WDIO_SYNC_ENABLED) {
```
mgrybyk: If string false is passed to wdio sync enabled the condition won't work as expected

packages/wdio-sync/src/index.js
```diff
@@ -1,83 +1,31 @@
...
-         * sometimes function result is Promise,
-         * we need to await result before proceeding
-         */
-        if (res instanceof Promise) {
-            return await res
-        }
+    promise.then((res) => {
+        result = res
+        isDone = true
+    }, (err) => {
```
mgrybyk: It might be more reliable to use catch instead

packages/wdio-sync/tests/index.test.js
```diff
@@ -1,127 +1,46 @@
...
-        try {
-            await executeSync.call({}, () => { throw new Error('foobar') }, {})
-        } catch (err) {
-            error = err
-        }
-        expect(error.stack).not.toContain('at new Promise (<anonymous>)')
-        expect(global.browser._NOT_FIBER).toBe(undefined)
-    })
+jest.mock('deasync', () => ({
+    loopWhile: jest.fn().mockImplementation((fn) => fn())
```
mgrybyk: Shouldn't we clear the mock in afterEach hook?

packages/webdriverio/src/middlewares.js
```diff
@@ -9,38 +8,44 @@ import implicitWait from './utils/implicitWait'
...
+            if (error.name === 'stale element reference') {
+                const element = await refetchElement(this, commandName)
+                this.elementId = element.elementId
+                this.parent = element.parent
+
+                return commandFn.apply(this, args)
+            }
+            throw error
+        }
+        // })
```
mgrybyk: Missed some commented code

tests/smoke.runner.js
```diff
@@ -50,7 +50,7 @@ const mochaTestrunner = async () => {
         {
             specs: [
                 path.resolve(__dirname, 'mocha', 'test.js'),
-                path.resolve(__dirname, 'mocha', 'test-middleware.js'),
+                // path.resolve(__dirname, 'mocha', 'test-middleware.js'),
```
mgrybyk: Why is it commented?

tests/mocha/test.js
```diff
@@ -60,7 +57,7 @@ describe('Mocha smoke test', () => {
         assert.deepEqual(results, ['https://mymockpage.com', 'https://mymockpage.com'])
     })
 
-    it('should handle waitUntil timeout', () => {
+    it.skip('should handle waitUntil timeout', () => {
```
mgrybyk: Why is it skipped?

mgrybyk:
Missed your comment https://github.com/webdriverio/webdriverio/pull/4886#issuecomment-567280029 

Sorry

packages/wdio-sync/src/index.js
```diff
@@ -1,83 +1,31 @@
...
+    let isDone = false
 
-        /**
-         * sometimes function result is Promise,
-         * we need to await result before proceeding
-         */
-        if (res instanceof Promise) {
-            return await res
-        }
+    promise.then((res) => {
```
mgrybyk: let's use `promise.finally` to set `isDone` to `true`

jcreager:
> link with article what can go wrong with using `deasync` - [article](https://joecreager.com/5-reasons-to-avoid-deasync-for-node-js/#return-note-1129-6)

Hi @christian-bromann, I am the author of the article that was cited earlier in the thread.  A long time ago I made a very small pr against the docs for WebdriverIO.  I was pretty green at the time but you were very patient and helpful.  I really appreciated your feedback.  I am excited that you read my post even if you didn't agree with it.

> Switching to N-API was just a style choice which doesn't change this invariant. Deasync clearly is still calling uv_run in a way that violates this and could explode at any time: https://github.com/abbr/deasync/blob/master/src/deasync.cc
> 
> As someone who has lived in this rabbit hole for 8 years I would definitely advise to avoid both fibers and deasync if at all possible :)

@laverdet articulates my point better than I did.  However, I understand that you are making a pragmatic choice.

> That said: we should definitely make this clear to users that dance the devil dance too and ensure they understand the possible consequences (e.g. new Node.js version isn't available right after the release). I will make sure to document this properly so that users can take a deliberate decision whether they want to opt-in or not.

I am glad you are going to do this, and I think it is a good thing that the sync module is not required to use WebdriverIO.  I agree with what you say about synchronous I/O being very useful and it was a very opinionated choice to enforce asynchronous I/O for the majority of Node.js I/O APIs.  Given `fibers` was in use before `deasync`, I don't see a good way to avoid breaking backwards compatibility.  I do not personally advocate for the use of `deasync`, but if it works for the users of WebdriverIO I cannot argue with that.

> @laverdet thanks for your input. I would definitely not use these packages for any application development and users of WebdriverIO still have the choice to not install `@wdio/sync`.

I think this understates the value of WebdriverIO.  I'd imagine it serves an important role at a lot of organizations.  Thank you for all of the work you have done to build and maintain WebdriverIO.

christian-bromann:
I've experienced more issues than initially expected and put this PR on hold.

packages/wdio-sync/src/index.js
christian-bromann: How? Using the Promise API this way does exactly the same.

packages/wdio-sync/src/index.js
christian-bromann: This is no different from the approach above. I tried the version at it makes it a lot harder to unit test.

tests/smoke.runner.js
christian-bromann: not done yet 😉

tests/mocha/test.js
christian-bromann: not done yet 😉

packages/wdio-sync/tests/index.test.js
christian-bromann: fixed

packages/wdio-sync/src/index.js
christian-bromann: removed that part

packages/wdio-runner/src/index.js
christian-bromann: removed that part

christian-bromann:
After looking into this again I found a problem that makes it hard to implement + `deasync` in WebdriverIO: since you don't define a context where commands should be run synchronously all commands are being run that way which makes the internal usage of commands problematic as they are expected to run async. Let me show an example:

```js
const elem = $('#elem')
elem.click()
```

Technically we should/can only run one sync command at a time which allows me to say: okay I have this flag that is `false` and if it is that way we run a command and switch it to `true` and run the command in sync mode. But following on we run all commands (as they are executed internally within WebdriverIO) asynchronously. Once the command is done we switch the flag and can run the next command synchronously.

This works besides for the following cases:

```js
browser.waitUntil(() => {
  return browser.call(() => {
    const text = $("...").getText()
    return browser.waitUntil(...)
  })
})
```

Here we have a layer of async/sync calls where can't determine which of them should be handled sync or async. For example I want that `browser.call` is being handled synchronously but if the `$` command triggers `findElement` or `waitForExist` (of not found right away) we want these to be handled synchronously.

I will fiddle around with this, maybe there is a solution. Just wanted to update y'all.

christian-bromann:
Ok, I was able to build some sort of sync/async handler that would keep track of nested stack traces and would decide whether or not to execute a command sync or asynchronously based on the idea that there can be only one sync command called by the user. This works great and is passing all smoke tests (besides one). Will check now how to adapt this idea for multiremote.

christian-bromann:
@webdriverio/project-committers this PR is now ready for review. I was able to integrate `deasync` appropriately. While it had its own challenges I think overall this is the better solution to move forward with for multiple reasons:

- commands don't need to be run in a fiber which has caused a lot of confusion for some users running a command in a different eventloop
- above allows us to remove a lot of code that we don't need to maintain anymore
- the C code part is just one simple file which makes it more realistic for us to fix bugs in case they come up
- above makes it also simpler to compile `deasync` for various OS/node combinations (also @abbr seems to be on top of it)
- for the user nothing will changes, it is still required to opt-in

mgrybyk:
I'll take a look. Are you going to increase the coverage?

christian-bromann:
> I'll take a look. Are you going to increase the coverage?

Yes, there is also one `skip` set for one of the smoke tests which I am not sure how to fix right now and if it is just a test problem or an actual issue with `deasync`.

christian-bromann:
While everything runs fine on Node v10 it seems to not work for v12. While `deasync` generally works in that version, it doesn't work in the way we run nested promises. So far there doesn't seem to be an easy fix for it. I also experienced errors like `async hook stack has become corrupted (actual: 209, expected: 210)` which might be related to this. Given that fibers still works I would not make this a blocker for the v6 release.

christian-bromann:
Closing as this seems to be unlikely to land given last observations.

jasonwilliams:
@christian-bromann looks like the author of node-fibers [recommends](https://github.com/laverdet/node-fibers) you avoiding the package from this point onwards, do you have plans for an alternative or is `deasync` still the best option?

christian-bromann:
@jasonwilliams for WebdriverIO `fibers` still makes a lot of sense to use and we will try to maintain it moving forward but there are chances that it just won't be compatible with upcoming Node.js versions anymore. I hope we can either continue to support it or propose something to Node itself. I could see that `deasync` could become an alternative given that [above concerns](https://github.com/webdriverio/webdriverio/pull/4886#issuecomment-574116222) can be resolved.

PR #4542, opened 2019-12-15
Drop Node v8 Support

christian-bromann:
Node v8 LTS support is ending at the end of the year. We should drop the support for it and clean up the fiber integration.

### DIFF
--- a/packages/wdio-cli/package.json
+++ b/packages/wdio-cli/package.json
@@ -7,10 +7,10 @@
   "license": "MIT",
   "main": "./build/index",
   "bin": {
-    "wdio": "./bin/wdio.js"
+    "node": ">=10.0.0"
   },
   "engines": {
-    "node": ">= 8.11.0"
+    "node": ">=10.0.0"
   },
   "scripts": {
     "build": "run-s clean compile copy",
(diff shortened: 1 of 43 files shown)


packages/wdio-cli/package.json
```diff
@@ -7,10 +7,10 @@
   "license": "MIT",
   "main": "./build/index",
   "bin": {
-    "wdio": "./bin/wdio.js"
+    "node": ">=10.0.0"
```
gustavohenke: Oops 😄

packages/wdio-cli/tests/launcher.test.js
```diff
@@ -318,7 +318,7 @@ describe('launcher', () => {
             expect(launcher.getNumberOfSpecsLeft()).toBe(0)
         })
 
-        it('should run as much as maxInstances allows', () => {
+        it.only('should run as much as maxInstances allows', () => {
```
gustavohenke: Should this be `.only`?

christian-bromann:
There is a bug in the test that is related to how Node.js does sorting in v10 vs v12. Currently trying to find a solution for this.

PR #4907, opened 2019-12-19
Set default WebDriver path to `/`

christian-bromann:
## Proposed changes

Having `/wd/hub` as default path is a relict from the old days where people would use Selenium via JsonWireProtocol to automate browser. These days the default path from most of the driver is `/`.

## Types of changes

- [x] Documentation update

## Checklist

## Further comments

Breaking change to land in v6.


packages/webdriver/tests/request.test.js
```diff
@@ -30,7 +30,7 @@ describe('webdriver request', () => {
 
     describe('createOptions', () => {
         it('fails if command requires sessionId but none given', () => {
-            const req = new WebDriverRequest('POST', '/wd/hub/session/:sessionId/element')
```
mgrybyk: Shall we move path to constant in the top of the file to reduce possible refactoring in later on?

packages/wdio-cli/src/templates/wdio.conf.tpl.ejs
```diff
@@ -24,16 +24,6 @@ exports.config = {
     path: '<%= answers.path %>',<% } else if (answers.services.includes('appium')) { %>
     port: 4723,<% }
 
-    if(answers.services.includes('chromedriver')) { %>
```
mgrybyk: Do you think it makes sense now to set default path to wd/hub if selenium standalone is used? At least until v4 is out (no idea when is it going to happen)

packages/webdriverio/tests/__mocks__/request.js
```diff
@@ -49,7 +49,7 @@ const requestMock = jest.fn().mockImplementation((params, cb) => {
     }
 
     switch (params.uri.path) {
-    case '/wd/hub/session':
```
mgrybyk: Having path in constant might help here as well

christian-bromann:
@mgrybyk I added an `if` clause in `request.js` that checks for an HTML body that contains: `Whoops! The URL specified routes to this help page.` which is the case for someone using WebdriverIO v6 but with a Selenium Standalone v3.x. It will notify the user to change the path to `/wd/hub`.

I also set `/wd/hub` as default when user is running tests on cloud provider as they still seem to run on the old path.

erwinheitzman:
Looking good to me. I can test out this branch with some services if you'd like to make sure we don't break anything.

Issue #4914, opened 2019-12-24
@wdio/appium-service not working on Windows


wswebcreation:
**Environment (please complete the following information):**
 - **WebdriverIO version:** `5.18.0`
 - **Mode:** WDIO Testrunner
 - **If WDIO Testrunner, running sync/async:** sync
 - **Node.js version:** `12.0.0`
 - **NPM version:** `6.9.0`
 - **Browser name and version:** -
 - **Platform name and version:** Windows 10
 - **Additional wdio packages used (if applicable):** `@wdio/appium-service`

**Config of WebdriverIO**
See [appium-boilerplate](https://github.com/webdriverio/appium-boilerplate)

**Describe the bug**
The Appium server will not be started on Windows

**To Reproduce**
See [appium-boilerplate](https://github.com/webdriverio/appium-boilerplate)

**Log**
```log
$ ./node_modules/.bin/wdio ./config/wdio.android.app.conf.js --logLevel=debug

Execution of 6 spec files started at 2019-12-24T09:01:51.465Z

2019-12-24T09:01:51.471Z DEBUG @wdio/utils:initialiseServices: initialise wdio service "appium"
2019-12-24T09:01:51.573Z INFO @wdio/cli:launcher: Run onPrepare hook
2019-12-24T09:01:51.574Z DEBUG @wdio/appium-service: Will spawn Appium process: appium
events.js:173
      throw er; // Unhandled 'error' event
      ^

Error: spawn appium ENOENT
    at Process.ChildProcess._handle.onexit (internal/child_process.js:248:19)
    at onErrorNT (internal/child_process.js:431:16)
    at processTicksAndRejections (internal/process/task_queues.js:83:17)
    at process.runNextTicks [as _tickCallback] (internal/process/task_queues.js:58:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:828:11)
    at internal/main/run_main_module.js:17:11
Emitted 'error' event at:
    at Process.ChildProcess._handle.onexit (internal/child_process.js:254:12)
    at onErrorNT (internal/child_process.js:431:16)
    [... lines matching original stack trace ...]
    at internal/main/run_main_module.js:17:11
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

``

wswebcreation:
Found the problem,

For Windows we need to start the server with 

```js
[...]

PR #4915, opened 2019-12-24
fix: fix spawning Appium on Windows

wswebcreation:
## Proposed changes
This fixes spawning Appium on Windows, see https://github.com/webdriverio/webdriverio/issues/4914

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments
Need to add tests for it, first wanted to check if you think this is the best way to do it

### DIFF
--- a/packages/wdio-appium-service/src/launcher.js
+++ b/packages/wdio-appium-service/src/launcher.js
@@ -65,7 +65,7 @@ export default class AppiumLauncher {
         })
 
         // only capture first error to print it in case Appium failed to start.
-        process.stderr.on('data', err => { error = err })
+        process.stderr.once('data', err => { error = err })
 
         process.once('exit', (exitCode) => {
             let errorMessage = `Appium exited before timeout (exit code: ${exitCode})`
--- a/packages/wdio-appium-service/src/launcher.js
+++ b/packages/wdio-appium-service/src/launcher.js
@@ -7,7 +7,6 @@ import getFilePath from './utils/getFilePath'
 
 const log = logger('@wdio/appium-service')
 const DEFAULT_LOG_FILENAME = 'appium.txt'
-const isWindows = process.platform === 'win32'
 
 export default class AppiumLauncher {
     constructor() {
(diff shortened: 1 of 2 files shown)


mgrybyk:
I think we did something similar for selenium standalone service. 
Mind adding unit tests?

wswebcreation:
Sure, I want to add this one

```js
        test('should set correct config properties for Windows', async () => {
            const config = {
                appium: {
                    logPath: './',
                    command: 'path/to/my_custom_appium',
                    args: { foo: 'bar' }
                }
            }
            await launcher.onPrepare(config)

            expect(launcher.command).toBe('cmd')
            expect(launcher.appiumArgs).toEqual(['/c', config.appium.command, '--foo', 'bar'])

        })
```

But I don't know how to mock the `process.env` here so I can hit line 54

packages/wdio-appium-service/src/launcher.js
```diff
@@ -48,8 +49,12 @@ export default class AppiumLauncher {
     }
 
     _startAppium(command, args, waitStartTime, callback) {
-        log.debug(`Will spawn Appium process: ${command} ${args.join(' ')}`)
-        let process = spawn(command, args, { stdio: ['ignore', 'pipe', 'pipe'] })
+        const cmd = isWindows ? 'cmd' : command
```
baruchvlz: Is this the best place to do this? It seems like you'll be overwriting any command passed through the [onPrepare method](https://github.com/webdriverio/webdriverio/blob/master/packages/wdio-appium-service/src/launcher.js#L24-L30).

wswebcreation: I think so, Windows commands need to be started different in comparison to Mac/Linux

christian-bromann: @wswebcreation I think what @baruchvlz was saying is whether we could move this to `onPrepare`

packages/wdio-appium-service/src/launcher.js
```diff
@@ -48,8 +49,12 @@ export default class AppiumLauncher {
...
+        if(isWindows){
+            args.unshift('/c', command)
```
erwinheitzman: I'd prefer to use `path.parse(process.cwd()).root` to identify the operating system's root but it's not mandatory I suppose

packages/wdio-appium-service/src/launcher.js
wswebcreation: Moved it, also gave me the option to mock the `process.platform`

PR #4920, opened 2019-12-26
Add markdown Reporter

carmenmitru:
Hello,

I created a new Reporter that generates a markdown file.


christian-bromann:
@carmenmitru thanks for another great plugin! Would you mind adding it to the CLI configurator too, see https://github.com/webdriverio/webdriverio/blob/master/packages/wdio-cli/src/constants.js#L50-L65

carmenmitru:
@christian-bromann  I added it :)

Issue #4922, opened 2019-12-26
Using Launcher api more then once in the same execution


naorzr:
[//]: # NOTE: This repository only maintains packages that are listed in the [Readme](https://github.com/webdriverio/webdriverio/blob/master/README.md#packages). Please make sure that your issue is directly caused by one of these packages and if not file an issue in the correct 3rd party package repository.

**Environment (please complete the following information):**
 - **WebdriverIO version:** [e.g. 5+]
 - **Mode:** WDIO Testrunner
 - **If WDIO Testrunner, running sync
 - **Node.js version:** [e.g. 12.13.0]
 - **NPM version:** [e.g. 6.13.1]

**Describe the bug**
When running the wdio.conf.js files using the launcher api, if ran more than once, on the consecutive run, wdio launcher will run all the current hooks + the hooks of the previous launch configs, and in the next consecutive run it will run all the current hooks + the hooks of the previous launch and the one before, and so on and so forth. 

**To Reproduce**
Assuming you have the basic setup with a config file and a test file, simply create an index.js file
with the following content:
```
 const Launcher = require('@wdio/cli').default;
 new Launcher('./wdio.conf.js').run().then(() => {
   new Launcher('./wdio.conf2.js').run()
 })
```
and run it.

**Expected behavior**
Only the configuration file of the current runner should be read and invoked.

**Additional Info**
The cause for this issue lies with the DEFAULT_CONFIGS constant under wdio-config/constants.
It is being exported out of the file, but as a constant, in this manner the config file is actually not being redefined on different imports, so the information persists between launchers.
It can be easily solved by wrapping it with a function, which will act as a factory for the default configurations.

christian-bromann:
Fixed in #4923

PR #4923, opened 2019-12-26
fix default configs being exported as a singleton

naorzr:
## Proposed changes

This pr is suppose to fix this bug issue #4922.

## Types of changes

## Checklist

### DIFF
--- a/packages/wdio-config/src/lib/ConfigParser.js
+++ b/packages/wdio-config/src/lib/ConfigParser.js
@@ -14,7 +14,8 @@ const MERGE_OPTIONS = { clone: false }
 
 export default class ConfigParser {
     constructor () {
-        this._config = DEFAULT_CONFIGS
+        this._config = DEFAULT_CONFIGS()
+
         this._capabilities = []
     }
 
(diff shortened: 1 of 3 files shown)


packages/wdio-config/src/lib/ConfigParser.js
```diff
@@ -14,7 +14,8 @@ const MERGE_OPTIONS = { clone: false }
 
 export default class ConfigParser {
     constructor () {
-        this._config = DEFAULT_CONFIGS
+        this._config = DEFAULT_CONFIGS()
```
christian-bromann: Can we keep the configs as object and rather use the spread syntax:
```suggestion
        this._config = { ...DEFAULT_CONFIGS }
```

naorzr: for one, this will not work.
the service hooks will still hold the same reference, since the spread operator merely performs a shallow copy.

second reason I think this should be discarded from design perspective, is that it transfer the responsibility to the developers to understand and not repeat this bug again.. which proved fatal once..

christian-bromann: makes sense, thanks for clarifying

PR #4924, opened 2019-12-27
Remove ff dependency

pmerwin:
## Proposed changes
Move puppeteer-firefox to an optional dependency so, we can use npm install --no-optional to exclude it from out env, where we only want chromium.

## Types of changes

- [ ] Documentation update

I am not sure if this will break other implementations or not, but for us we only need chromium.

## Checklist

- [ x] I have read the [CONTRIBUTING](https://github.com/webdriverio/webdriverio/blob/master/CONTRIBUTING.md) doc


just-boris:
Great PR, but this was not enough, unfortunately. 

The [source code](https://github.com/webdriverio/webdriverio/blob/master/packages/devtools/src/launcher.js#L3) attempts to require the module and fails if it is not installed. We need to make modules lazy-loaded, only when firefox is actually requested. The easiest way to do so is to convert that import statement into `require` and move it into `launchFirefox` function. Then it will not be loaded unless firefox is directly requested.

pmerwin:
@just-boris Awe great catch!  New PR coming...

pmerwin:
@just-boris see:

https://github.com/webdriverio/webdriverio/pull/4927

I made the changes requested, I am perplexed over how not to break this unit test though: ..webdriverio/packages/devtools/tests/launcher.test.js

any suggestions?

We are importing import puppeteerFirefox from 'puppeteer-firefox' to mock it, can we just make a 'dummy' file for this, as the test only mock this call and verify it was NOT called when it should not be, etc

PR #4606, opened 2019-12-28
Remove ability to set service configs on root level

christian-bromann:
Currently services define settings in wrong ways. All attach their configurations on the root of the config file, e.g. statice server service:

```js
 // wdio.conf.js
 export.config = {
    // ...
    staticServerFolders: [
        { mount: '/fixtures', path: './tests/fixtures' },
        { mount: '/dist', path: './dist' },
    ],
    // ...
 };
```

With v5 we allowed to have configurations that are specific to services defined in the `services` array:

```js
 // wdio.conf.js
 export.config = {
    // ...
    services: [
		['static-server', {
    		staticServerFolders: [
        		{ mount: '/fixtures', path: './tests/fixtures' },
        		{ mount: '/dist', path: './dist' }
			]
		}]
    ],
    // ...
 };
```

Here is what this PR does:

- updates all service docs to define example configuration as in the second example above
- ensures that services receive the configuration from the right place
- ensures that all supported configs are documented
- changes some service argument naming conventions to be similar across all service args (e.g. I renamed `seleniumArgs` to just `args` in the standalone service)


packages/wdio-appium-service/src/launcher.js
```diff
@@ -1,46 +1,46 @@
...
 const log = logger('@wdio/appium-service')
 const DEFAULT_LOG_FILENAME = 'appium.txt'
 
 export default class AppiumLauncher {
-    constructor() {
-        this.logPath = null
-        this.command = ''
+    constructor(options, caps, config) {
+        this.options = options
+        this.logPath = options.logPath || config.outputDir
```
baruchvlz: Is there a chance that both of those would be undefined?

packages/wdio-browserstack-service/src/launcher.js
```diff
@@ -1,19 +1,27 @@
...
+
 import BrowserstackLocalLauncher from 'browserstack-local'
 import logger from '@wdio/logger'
 
 const log = logger('@wdio/browserstack-service')
 
 export default class BrowserstackLauncherService {
-    onPrepare(config, capabilities) {
-        if (!config.browserstackLocal) {
+    constructor (options, capabilities, config) {
```
baruchvlz: Seems like `capabilities` is not used, can we remove it from the constructor?

packages/wdio-browserstack-service/src/service.js
```diff
@@ -6,14 +6,14 @@ import { BROWSER_DESCRIPTION } from './constants'
 const log = logger('@wdio/browserstack-service')
 
 export default class BrowserstackService {
-    constructor (config) {
+    constructor (options, caps, config) {
```
baruchvlz: The argument `caps` is not being used, do we still need it in the constructor?

packages/wdio-firefox-profile-service/tests/launcher.test.js
```diff
@@ -4,27 +4,26 @@ import FirefoxProfile from 'firefox-profile'
...
     describe('onPrepare', () => {
         test('should return when no firefoxProfile set in the config', async () => {
-            const config = {}
+            const options = {}
             const capabilities = [{}]
+            const config = {}
 
-            const service = new Launcher()
-            service.onPrepare(config, capabilities)
+            const service = new Launcher(options, capabilities, config)
```
baruchvlz: According to the changes in the service, `FirefoxProfileLauncher` constructor only takes one parameter.

packages/wdio-sauce-service/src/launcher.js
```diff
@@ -1,59 +1,75 @@
...
+            `SC-tunnel-${Math.random().toString().slice(2)}`)
+
         this.sauceConnectOpts = Object.assign({
             username: config.user,
             accessKey: config.key,
-            logger: log.debug
-        }, config.sauceConnectOpts)
+            logger: log.debug,
+            tunnelIdentifier: sauceConnectTunnelIdentifier
+        }, sauceConnectOpts)
```
baruchvlz: Instead of using `Object.assign` can we use spread operator to clone the object?

packages/wdio-sauce-service/src/launcher.js
```diff
@@ -1,59 +1,75 @@
...
+        if (Array.isArray(capabilities)) {
+            capabilities.forEach(capability => {
+                if (capability['sauce:options'] === undefined) {
+                    capability.tunnelIdentifier = capability.tunnelIdentifier || sauceConnectTunnelIdentifier
+                } else {
+                    capability['sauce:options'].tunnelIdentifier = capability['sauce:options'].tunnelIdentifier || sauceConnectTunnelIdentifier
+                }
+            })
+        } else {
+            Object.keys(capabilities).forEach(browser => {
```
baruchvlz: This is just a code style suggestion, but in some other services we're using `for..in` loops and here we're using a `forEach`, maybe we should add a linter to force one or the other?

packages/wdio-sauce-service/tests/launcher.test.js
```diff
@@ -1,61 +1,80 @@
...
             tunnelIdentifier: 'my-tunnel'
         }
     }
-    const service = new SauceServiceLauncher()
+    const caps = [{}]
+    const config = {
+        user: 'foobaruser',
+        key: '12345'
+    }
+    const service = new SauceServiceLauncher(options, caps, config)
```
baruchvlz: Launcher's constructor only takes 1 parameter.

packages/wdio-selenium-standalone-service/src/launcher.js
```diff
@@ -4,34 +4,29 @@ import { promisify } from 'util'
...
         if (!this.skipSeleniumInstall) {
-            await promisify(SeleniumStandalone.install)(this.seleniumInstallArgs)
+            await promisify(SeleniumStandalone.install)(this.installArgs)
         }
 
-        this.process = await promisify(SeleniumStandalone.start)(this.seleniumArgs)
+        this.process = await promisify(SeleniumStandalone.start)(this.args)
 
-        if (typeof this.seleniumLogs === 'string') {
+        if (typeof this.logPath === 'string') {
```
baruchvlz: Seems like `logPath` should always be a string, should we throw an error if it's not?

mgrybyk:
Is it ok to use array indexes instead of object properties? 
```
services: [
    ['service1', { ...options1 }],
    ['service2', { ...options1 }],
    ['service3'],
]
```
VS
```
services: [
    { name: 'service1', options: { ...options1 } },
    { name: 'service2', options: { ...options2 } },
    { name: 'service3' },
]
```

christian-bromann:
> Is it ok to use array indexes instead of object properties?

I would prefer first approach as I've seen it in other frameworks too. The latter just introduces the necessity of knowing the properties for `name` and `options` where it is clear with the first approach.

packages/wdio-appium-service/src/launcher.js
christian-bromann: yes, in this case we would not print the Appium logs into a file.

packages/wdio-browserstack-service/src/launcher.js
christian-bromann: We could rename it to `_` given that it is not important but we use config and therefor need to access the 3rd parameter somehow.

packages/wdio-browserstack-service/src/service.js
christian-bromann: same as above

packages/wdio-sauce-service/src/launcher.js
christian-bromann: replaced it with for..of loop. I don't have a strong opinion on enforcing it. This should be a separated issue/PR then as I think that there are a bunch of forEach loops used in the code base.

packages/wdio-selenium-standalone-service/src/launcher.js
christian-bromann: I would prefer to handle option type checks separated so we can assume in the services that if set it is definitely the expected type.

christian-bromann:
@baruchvlz @mgrybyk I applied feedback and addressed question? Anything left to do? If not, can I get an approval?

mgrybyk:
Do you think it makes sense moving user and key from global level to services? Not sure why these two are global.

Overall LGTM

christian-bromann:
> Do you think it makes sense moving user and key from global level to services?

It would be nice to make running tests on these cloud services simple even without using the wdio testrunner or a service.

erwinheitzman:
One comment though, you do not need the services in order to be able to run in the cloud.
Providing the user / key is enough to get it running, do we want to change this?

erwinheitzman:
And is it still possible to change the root config's properties with these changes?
Like for example in a service do:
```
onPrepare(config) {
  config.exampleProp = 'example value'
}
```

christian-bromann:
> do we want to change this?

No.

> And is it still possible to change the root config's properties with these changes?

Yes. Services do still have access to the full options object.

baruchvlz:
LGTM

PR #4925, opened 2019-12-29
Updated docs for Applitools service

amitzur:
## Proposed changes

This PR intends to replace the docs for the Applitools Eyes service, from the one developed within `@wdio` to the one provided by Applitools. Nature of the change mostly includes changing reference in the main Readme, and a change to the Readme at `wdio-applitools-service`, in order to reflect the new docs in the website.

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ If you're looking for issues to help out with, check out [the issues labelled "g
 ### Services
 
 - [@wdio/appium-service](https://github.com/webdriverio/webdriverio/tree/master/packages/wdio-appium-service) - A WebdriverIO service to start & stop Appium Server
-- [@wdio/applitools-service](https://github.com/webdriverio/webdriverio/tree/master/packages/wdio-applitools-service) - A WebdriverIO service for visual regression testing using Applitools
+- [@applitools/eyes-webdriverio5-service](https://github.com/applitools/eyes-webdriverio5-service) - A WebdriverIO service for visual regression testing using Applitools
 - [@wdio/devtools-service](https://github.com/webdriverio/webdriverio/tree/master/packages/wdio-devtools-service) - A WebdriverIO service that allows you to run Chrome DevTools commands in your tests
 - [@wdio/firefox-profile-service](https://github.com/webdriverio/webdriverio/tree/master/packages/wdio-firefox-profile-service) - A WebdriverIO service that lets you define your Firefox profile in your wdio.conf.js
 - [@wdio/sauce-service](https://github.com/webdriverio/webdriverio/tree/master/packages/wdio-sauce-service) - A WebdriverIO service that provides a better integration into SauceLabs
--- a/docs/ConfigurationFile.md
+++ b/docs/ConfigurationFile.md
@@ -134,7 +134,7 @@ exports.config = {
     // use 'silent' level to disable logger
     logLevels: {
         webdriver: 'info',
-        '@wdio/applitools-service': 'info'
+        '@wdio/devtools-service': 'info'
     },
     //
     // Set directory to store all logs into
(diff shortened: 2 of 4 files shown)


christian-bromann:
Closing. See comment above.

amitzur:
Hey @christian-bromann,

We'd like this PR to be reopened. We've done much work on the service in the past months:
1. Added support for the Ultrafast grid
2. Added detailed documentation (in both [the service](https://www.npmjs.com/package/@applitools/eyes-webdriverio5-service) and the [SDK itself](https://www.npmjs.com/package/@applitools/eyes-webdriverio))
3. Updated internal parameters that help us track usage of different SDK's (specifically it's what we call `agentId`).

We'd like to take ownership of this service - it is well maintained, it now also has great documentation, has a large user base, and the existence of 2 services is creating confusion and support overhead for our customers.

Lastly, this PR is actually not the ideal solution for us - we want to recommend our users to use the SDK package directly (`@applitools/eyes-webdriverio`). They would need to require it as a custom service rather than use the string method. See the relevant section in the docs [here](https://www.npmjs.com/package/@applitools/eyes-webdriverio#using-the-sdk-as-a-service).
Our strategy for moving forward is to support all major versions >=5 in the same package.

The advantages to this approach from our perspective:
1. There's one less thing for our users to know and care about.
2. Users can require any exported object from the SDK directly from tests even if they use it as a service. There are quite a few configuration and utilities they can import.
3. There's no `5` in the name of the package :)

But it doesn't immerse well into the install script. In order to automatically add the service, the script should support more than only strings and be able to automatically add a require at the top of the config file.

I'd love to hear your thoughts and ideas on what we can do to move forward to the optimal solution.

But even if we at least document `@applitools/eyes-webdriverio5-service` it's a step forward for us.

Thanks!

christian-bromann:
Hey 👋

> We'd like to take ownership of this service - it is well maintained

Which service are you referring too? There is the one maintained in this repository (`@wdio/applitools-service`) and one maintained by you.

> the existence of 2 services is creating confusion and support overhead for our customers.

Totally agree, let's find a solution for it.

> we want to recommend our users to use the SDK package directly (`@applitools/eyes-webdriverio`). They would need to require it as a custom service rather than use the string method.

What is the reason behind this decision? WebdriverIO has a name pattern for services that allows to import the service automatically, keeping the config file clean and simple. I would much rather prefer instead of having this setup:

```js
const {EyesService} = require('@applitools/eyes-webdriverio')
 
exports.config = {
  // ...
  services: [[EyesService, {
    // webdriver.io version 6: Eyes service configuration here
  }]],
  // ...
  eyes: {
    // webdriver.io version 5: Eyes service configuration here
  }
}
```
to just do:
```js
exports.config = {
  // ...
  services: [['applitools', {
    // Eyes service configuration here
  }]],
}
```
I was wondering why Applitools preferred to build a custom service rather than contributing to the one we have build in the first place. WebdriverIO has an open governance that allows anyone to participate in the development of individual services of this repository.

> But it doesn't immerse well into the install script. In order to automatically add the service, the script should support more than only strings and be able to automatically add a require at the top of the config file.

As mentioned above, there is no need for 3rd party services to be imported in the config as WebdriverIO automatically imports them. We would like to avoid complexity in configurations and rather avoid unnecessary code. This makes services easier discoverable and consumable.

Are there any arguments against collaborating on the [existing service](https://github.com/webdriverio/webdriverio/tree/master/packages/wdio-applitools-service) together?

amitzur:
Actually we'd love to collaborate on the existing service. We see a lot of value in being immersed into the ecosystem.
There are a few fine points that need to be discussed regarding the workflow, releasing new versions etc. That would make sense to discuss separately. Christian, what's a good way to reach out to you directly?

christian-bromann:
> Christian, what's a good way to reach out to you directly?

Best would be over my company email christian[at]saucelabs.com

PR #4927, opened 2019-12-30
Remove ff dependency

pmerwin:
## Proposed changes

from:

https://github.com/webdriverio/webdriverio/pull/4924

"Great PR, but this was not enough, unfortunately.

The source code attempts to require the module and fails if it is not installed. We need to make modules lazy-loaded, only when firefox is actually requested. The easiest way to do so is to convert that import statement into require and move it into launchFirefox function. Then it will not be loaded unless firefox is directly requested."

I made these changes, I am perplexed over how not to break this unit test though: ..webdriverio/packages/devtools/tests/launcher.test.js

any suggestions?

We are importing import puppeteerFirefox from 'puppeteer-firefox' to mock it, can we just make a 'dummy' file for this, as the test only mock this call and verify it was NOT called when it should not be, etc

## Types of changes

Change import to require as per Bor's suggestion

## Checklist

- [ x] I have added necessary documentation (if appropriate)


packages/devtools/src/launcher.js
```diff
@@ -58,6 +57,7 @@ async function launchChrome (capabilities) {
 }
 
 function launchFirefox (capabilities) {
+    puppeteerFirefox = require('puppeteer-firefox');
```
just-boris: This has to be defined

```suggestion
    const puppeteerFirefox = require('puppeteer-firefox');
```

just-boris:
Regarding the mock in the test: if I understand Jest implementation correctly, jest should handle this case as well, it also mocks the require calls.

packages/devtools/src/launcher.js
pmerwin: oops my bad, ok.

pmerwin:
> Regarding the mock in the test: if I understand Jest implementation correctly, jest should handle this case as well, it also mocks the require calls.

Awe ok great thanks!

packages/devtools/src/launcher.js
christian-bromann: There are still some linter issues:

```
/home/travis/build/webdriverio/webdriverio/packages/devtools/src/launcher.js
  60:5   error  'puppeteerFirefox' is not defined  no-undef
  60:52  error  Extra semicolon                    semi
  62:12  error  'puppeteerFirefox' is not defined  no-undef
```

Please fix by
```suggestion
    const puppeteerFirefox = require('puppeteer-firefox')
```

pmerwin:
I am not sure what I am doing wrong, but how can I get the unit tests to run?

npm install --no-optional

npm run test:
Gets me linter errors and no tests run?
✖ 212 problems (212 errors, 0 warnings)

and:

 npm run ts:devtools

> webdriverio-monorepo@ ts:devtools /Users/pmerwin/Projects/webdriverio
> cd tests/typings/devtools && tsc --incremental

error TS2688: Cannot find type definition file for 'devtools'.

pmerwin:
yeah shoot the unit test is broken because it imports ff-puppeteer too, etc

https://travis-ci.org/webdriverio/webdriverio/jobs/631028363?utm_medium=notification&utm_source=github_status

packages/devtools/src/launcher.js
```diff
@@ -58,6 +57,7 @@ async function launchChrome (capabilities) {
 }
 
 function launchFirefox (capabilities) {
+    const puppeteerFirefox = require('puppeteer-firefox')
```
christian-bromann: To fix unit tests, please modify the [`puppeteer-firefox`](https://github.com/webdriverio/webdriverio/blob/master/packages/devtools/tests/__mocks__/puppeteer-firefox.js) mock to use CommonJS syntax:

```js
module.exports = {
    launch: jest.fn()
}
```

pmerwin: Awe ok will do!

pmerwin:
@christian-bromann Looks that fixe the unit test, thanks!!!

Issue #4715, opened 2019-10-31
Replace args with options object in some methods


mgrybyk:
**Is your feature request related to a problem? Please describe.**
it's not convenient to have `browser.waitUntil(() => $(''foo).isExisting(), null, null, 1000)`

**Describe the solution you'd like**
I'd like to use options instead, ex:
`browser.waitUntil(() => $(''foo).isExisting(), { interval: 1000 })`

**Describe alternatives you've considered**

**Additional context**

affected `browser` methods:
- browser.addCommand (shall we move `attachToSession` to options?)
- newWindow
- react$
- react$$
- waitUntil

affected `element` methods:
- dragAndDrop
- moveTo
- react$
- react$$
- waitForClickable
- waitForDisplayed
- waitForEnabled
- waitForExist
- waitUntil

baruchvlz:
I think `custom$/$$` commands would be affected as well once #4603 is merged.

christian-bromann:
Fixed by #4928

> I think `custom$/$$` commands would be affected as well once #4603 is merged.

@baruchvlz I didn't see a reason to change the interface there as the parameters are not options to the command but actual required parameters.

PR #4928, opened 2019-12-30
Replace args with options object in some methods

christian-bromann:
## Proposed changes

It's not convenient to have `browser.waitUntil(() => $(''foo).isExisting(), null, null, 1000)`

Let's use options instead, ex:
`browser.waitUntil(() => $(''foo).isExisting(), { interval: 1000 })`

affected `browser` methods:
- newWindow
- react$
- react$$
- waitUntil

affected `element` methods:
- dragAndDrop
- moveTo
- react$
- react$$
- scrollIntoView
- waitForClickable
- waitForDisplayed
- waitForEnabled
- waitForExist
- waitUntil

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

See original issue: #4715

### DIFF
--- a/packages/webdriverio/src/commands/element/waitForDisplayed.js
+++ b/packages/webdriverio/src/commands/element/waitForDisplayed.js
@@ -15,35 +15,35 @@
     :waitForVisibleExample.js
     it('should detect when element is visible', () => {
         const elem = $('#elem')
-        elem.waitForDisplayed(3000);
+        elem.waitForDisplayed({ timeout: 3000 });
     });
     it('should detect when element is no longer visible', () => {
         const elem = $('#elem')
         // passing 'undefined' allows us to keep the default timeout value without overwriting it
-        elem.waitForDisplayed(undefined, true);
+        elem.waitForDisplayed({ reverse: true });
     });
  * </example>
  *
  * @alias element.waitForDisplayed
- * @param {Number=}  ms       time in ms (default: 500)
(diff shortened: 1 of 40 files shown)


packages/webdriverio/src/commands/element/waitForDisplayed.js
```diff
@@ -15,35 +15,35 @@
...
-    if (typeof ms !== 'number') {
-        ms = this.options.waitforTimeout
-    }
-
-    const isReversed = reverse ? '' : 'not '
-    const errorMsg = typeof error === 'string' ? error : `element ("${this.selector}") still ${isReversed}displayed after ${ms}ms`
-
-    return this.waitUntil(async () => reverse !== await this.isDisplayed(), ms, errorMsg)
+export default async function waitForDisplayed ({
+    timeout = this.options.waitforTimeout,
```
erwinheitzman: Do not set the defaults on these methods but set the default on the `waitUntil method` instead so that we can maintain it in one place instead of many

christian-bromann:
@erwinheitzman addressed your comment, anything left to do?

erwinheitzman:
The build is failing and there's some conflicts

christian-bromann:
@erwinheitzman resolved the conflicts, can you take another look?

christian-bromann:
@erwinheitzman had to bring back the defaults for individual waitFor commands because otherwise the default `timeoutMsg` wouldn't be correct.

Issue #4936, opened 2020-01-03
Can't fill the Angular Material input with 'Delete' value


IgorSasovets:
**Environment (please complete the following information):**
 - **WebdriverIO version:** [@wdio/cli@5.18.1]
 - **Mode:** [WDIO Testrunner]
 - **If WDIO Testrunner, running sync/async:** [async]
 - **Node.js version:** [v8.11.3]
 - **NPM version:** [6.11.3]
 - **Browser name and version:** [Chrome 79 (79.0.3945.88)]
 - **Platform name and version:** [Ubuntu 17.10]
 - **Additional wdio packages used (if applicable):** [https://github.com/IgorSasovets/wdio-project/blob/master/package.json]

**Config of WebdriverIO**
```
'use strict';
const path = require('path');
global.mocha = require('mocha');

exports.config = {
    runner: 'local',
    path: '/',
    waitforTimeout: 10000,
    connectionRetryTimeout: 90000,
    connectionRetryCount: 3,
    framework: 'mocha',
    mochaOpts: {
        ui: 'bdd',
        timeout: 99999
    },
    capabilities: [
      {
        browserName: 'chrome',
        'goog:chromeOptions': {
            prefs: {'profile.managed_default_content_settings.notifications': 1}
        }
      }
    ],
    baseUrl: 'https://github.com',
    logLevel:  'debug',
    specs: ['tests/*spec.js'],
    services: [
        'chromedriver'
    ],
    before: function(capabilities, specs) {
        browser.maximizeWindow();
    }
};
```

**Describe the bug**
I cannot set the value of angular material input component to `Delete`. If I change some letters or use another word, tests will pass successfully.

**To Reproduce**
1. Clone the [repository](https://github.com/IgorSasovets/wdio-project).
2. Prepare the environment (node, npm version, install modules for the project).
3. Run the tests using `npm run test-samples` command.

To reproduce the bug, run [this suite](https://github.com/IgorSasovets/wdio-project/blob/master/tests/material.spec.js).

**Expected behavior**
Input value successfully changed to `Delete`

**Log**
If applicable, add logs to help explain your problem.
[logs link](https://raw.githubusercontent.com/IgorSasovets/wdio-project/master/logs.log)

**Additional context**
Angular material sample project is located [here](https://stackblitz.com/edit/angular-cbpbdt?file=src%2Fapp%2Fdialog-overview-example.ts). `Delete` text is interpreted as below:
```
DATA { text: '' }
```

[...]

PR #4946, opened 2020-01-08
Added possibility to disable automatic strings to unicode translation

IgorSasovets:
## Proposed changes

This is a fix for wdio automatic translation to [unicode issue](https://github.com/webdriverio/webdriverio/issues/4936). In general, user can't send keys that are predefined in the UNICODE_CHARACTERS [list](https://github.com/webdriverio/webdriverio/blob/master/packages/webdriverio/src/constants.js).

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/webdriverio/src/commands/element/addValue.js
+++ b/packages/webdriverio/src/commands/element/addValue.js
@@ -21,7 +21,8 @@
  *
  * @alias element.addValue
  * @param {string | number | boolean | object | Array<any>}      value     value to be added
- * @param {boolean} translateToUnicode enable translation string to unicode value automatically
+ * @param {AddValueOptions} options                    command options (optional)
+ * @param {boolean}         options.translateToUnicode enable translation string to unicode value automatically
  * @uses protocol/elements, protocol/elementIdValue
  * @type action
  *
--- a/packages/webdriverio/src/commands/element/addValue.js
+++ b/packages/webdriverio/src/commands/element/addValue.js
@@ -4,7 +4,8 @@
  * characters like Left arrow or Back space. WebdriverIO will take care of
  * translating them into unicode characters. You’ll find all supported characters
  * [here](https://w3c.github.io/webdriver/webdriver-spec.html#keyboard-actions).
- * To do that, the value has to correspond to a key from the table.
+ * To do that, the value has to correspond to a key from the table. It can be disabled
+ * by setting `translateToUnicode` optional parameter to false.
  *
  * <example>
     :addValue.js
@@ -20,6 +21,7 @@
  *
  * @alias element.addValue
  * @param {string | number | boolean | object | Array<any>}      value     value to be added
+ * @param {boolean} translateToUnicode enable translation string to unicode value automatically
  * @uses protocol/elements, protocol/elementIdValue
  * @type action
  *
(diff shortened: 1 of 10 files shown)


packages/webdriverio/src/commands/element/addValue.js
```diff
@@ -20,22 +21,24 @@
  *
  * @alias element.addValue
  * @param {string | number | boolean | object | Array<any>}      value     value to be added
+ * @param {boolean} translateToUnicode enable translation string to unicode value automatically
```
christian-bromann: This needs to be formatted as follows so that type definitions are generated properly:

```suggestion
 * @param {AddValueOptions} options                    command options (optional)
 * @param {Boolean}         options.translateToUnicode enable translation string to unicode value automatically
```

Then please add the type to the [definition template](https://github.com/webdriverio/webdriverio/blob/master/scripts/templates/webdriverio.tpl.d.ts#L181):

```ts
interface AddValueOptions {
    translateToUnicode: boolean
}
```

packages/webdriverio/src/commands/element/setValue.js
```diff
@@ -19,12 +20,13 @@
  *
  * @alias element.setValue
  * @param {string | number | boolean | object | Array<any>}      value    Value to be added
+ * @param {boolean} translateToUnicode enable translation string to unicode value automatically
```
christian-bromann: same here

packages/webdriverio/src/utils.js
```diff
@@ -91,7 +91,7 @@ export function transformToCharString (value) {
 
     for (const val of value) {
         if (typeof val === 'string') {
-            ret.push(...checkUnicode(val))
+            (translateToUnicode) ? ret.push(...checkUnicode(val)) : ret.push(...`${val}`.split(''))
```
christian-bromann: no brackets needed

```suggestion
            translateToUnicode ? ret.push(...checkUnicode(val)) : ret.push(...`${val}`.split(''))
```

IgorSasovets: Agreed, I'll add required changes

packages/webdriverio/src/commands/element/addValue.js
IgorSasovets: Fixed

packages/webdriverio/src/commands/element/setValue.js
IgorSasovets: fixed

packages/webdriverio/src/commands/element/addValue.js
IgorSasovets: @christian-bromann , it seems that type should be added to some other files (at least `webdriverio-core-v5.d.ts`), isn't it?

IgorSasovets: Something like this
```
addValue(
            value: string | number | boolean | object | any[],
            translateToUnicode?: AddValueOptions
        ): void;

```
?

christian-bromann: yeah

IgorSasovets:
Sure, are there any examples on how to do it in the right way?

IgorSasovets:
This option (`translateToUnicode`) is not a function and I am not sure how to correctly add it to the [generate-typing-utils](https://github.com/webdriverio/webdriverio/blob/master/scripts/type-generation/generate-typings-utils.js#L13)

christian-bromann:
Add is as follows:

```ts
    case 'Function':
    case 'RegExp':
    case 'AddValueOptions': // <-- this was added
    case 'WaitForOptions':
    case 'Element':
```

IgorSasovets:
Done

christian-bromann:
Release in `5.18.5`

PR #4944, opened 2020-01-15
webdriver.d.ts colliding type definitions for launchApp between Appium and Chromium

christian-bromann:
 - **WebdriverIO version:** 5.18.0
 - **Node.js version:** 13.5.0
 - **NPM version:** 6.13.4

**Describe the bug**
node_modules/webdriver/webdriver.d.ts has the following definition for Appium

```
    /**
         * [appium]
         * Launch an app on device. iOS tests with XCUITest can also use the `mobile: launchApp` method. See detailed [documentation](http://appium.io/docs/en/writing-running-appium/ios/ios-xctest-mobile-apps-management/index.html#mobile-launchapp).
         * http://appium.io/docs/en/commands/device/app/launch-app/
         */
        launchApp(): void;
```

And the following definition for Chromium, which comes later

```
        /**
         * [chromium]
         * Launches a Chrome app by specified id.
         * https://github.com/bayandin/chromedriver/blob/2.45/session_commands.cc#L521-L539
         */
        launchApp(id: string): void;
```

That means I am not able to use Appium's method definition (the one without arguments) directly.

The easiest fix I can think of is making `id` optional.

### DIFF
--- a/packages/wdio-protocols/protocols/chromium.json
+++ b/packages/wdio-protocols/protocols/chromium.json
@@ -3,7 +3,7 @@
     "GET": {
       "command": "isAlertOpen",
       "description": "Whether a simple dialog is currently open.",
-      "ref": "https://github.com/bayandin/chromedriver/blob/2.45/alert_commands.cc#L42-L49",
+      "ref": "https://github.com/bayandin/chromedriver/blob/v2.45/alert_commands.cc#L42-L49",
       "examples": [
         [
           "console.log(browser.isAlertOpen()); // outputs: false",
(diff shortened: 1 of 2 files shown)


christian-bromann:
I renamed the Chrome command to `launchChromeApp` to resolve this conflict. As this will be a breaking change we will move it into the v6 release.

mgrybyk:
@christian-bromann looks like you forgot to generate types, no?

christian-bromann:
@mgrybyk done

PR #4959, opened 2020-01-15
Support fully qualified paths to plugins

### LINKED ISSUE #4867: Support fully qualified paths to plugins
Currently [`initialisePlugin()`](https://github.com/webdriverio/webdriverio/blob/v5.16.15/packages/wdio-utils/src/initialisePlugin.js#L13) does direct module resolution only when the package name starts with `@`.

I want to be able to use fully resolved paths for plugins so that I can build a shared config package with shared dependencies and properly reference the bundled dependencies in a way that the consumer of this package can use them properly.

I would like for this to also do direct module resolution when the package name starts with a slash.

So maybe we could change [`initialisePlugin()`](https://github.com/webdriverio/webdriverio/blob/v5.16.15/packages/wdio-utils/src/initialisePlugin.js#L13) like this:
```diff
- if (name[0] === '@') {
+ if (name[0] === '@' || name[0] === path.sep) {
```

Another approach might be to allow pre-loaded plugins in more situations (e.g., `services` can already take a function). This would unfortunately require loading these plugins during config definition rather than config usage. I suspect that we'd rather continue to lazily load plugins during config usage if possible.

**Additional context**
Here's a more concrete example of what I'd like to see work:
```js
{
  runner: require.resolve("@wdio/local-runner"),
  framework: require.resolve("@wdio/mocha-framework"),
  reporters: [
    require.resolve("@wdio/spec-reporter")
  ]
}
```

christian-bromann:
## Proposed changes

closes #4867

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/packages/wdio-utils/src/initialisePlugin.js
+++ b/packages/wdio-utils/src/initialisePlugin.js
@@ -1,3 +1,4 @@
+import path from 'path'
 import { safeRequire } from './utils'
 
 /**
@@ -8,9 +9,9 @@ import { safeRequire } from './utils'
  */
 export default function initialisePlugin (name, type, target = 'default') {
     /**
-     * directly import packages that are scoped
+     * directly import packages that are scoped or start with an absolute path
      */
-    if (name[0] === '@') {
+    if (name[0] === '@' || name[0] === path.sep) {
         const service = safeRequire(name)
         return service[target]
     }
--- a/packages/wdio-utils/src/initialisePlugin.js
+++ b/packages/wdio-utils/src/initialisePlugin.js
@@ -11,7 +11,7 @@ export default function initialisePlugin (name, type, target = 'default') {
     /**
      * directly import packages that are scoped or start with an absolute path
      */
-    if (name[0] === '@' || name[0] === path.sep) {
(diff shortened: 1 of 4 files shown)


packages/wdio-utils/src/initialisePlugin.js
```diff
@@ -8,9 +9,9 @@ import { safeRequire } from './utils'
  */
 export default function initialisePlugin (name, type, target = 'default') {
     /**
-     * directly import packages that are scoped
+     * directly import packages that are scoped or start with an absolute path
      */
-    if (name[0] === '@') {
+    if (name[0] === '@' || name[0] === path.sep) {
```
mgrybyk: [path.sep](https://nodejs.org/api/path.html#path_path_sep) is either `/` or `\`.

How to handle the following cases:
- `./foo/bar` mentioned in the doc
- `c:\foo\bar`
- or just `foo\bar`

christian-bromann:
@mgrybyk I replaced the check with `path.isAbsolute` which should work for Linux and Windows systems. Also updated the docs to be implicit about using absolute paths.

PR #4626, opened 2020-01-16
Separate launcher from worker services

christian-bromann:
## Problem

Currently we require a service module within the worker process even we don't need it there (e.g. Chromedriver service). This causes quit some memory hog. See also #4623.

Let's allow services to define where they are suppose to be used (in the main launcher process or in the worker process or maybe in both). We need to expose this information so that we don't need to require the whole module in the worker.

### DIFF
--- a/packages/wdio-utils/src/initialiseServices.js
+++ b/packages/wdio-utils/src/initialiseServices.js
@@ -84,21 +84,23 @@ export function initialiseLauncherService (config, caps) {
     const ignoredWorkerServices = []
     const launcherServices = []
 
-    const services = initialiseServices(config.services.map(sanitizeServiceArray))
-    for (const [service, serviceConfig, serviceName] of services) {
-        /**
-         * add object service
-         */
-        if (typeof service === 'object') {
-            return launcherServices.push(service)
-        }
+    try {
+        const services = initialiseServices(config.services.map(sanitizeServiceArray))
+        for (const [service, serviceConfig, serviceName] of services) {
+            /**
+             * add custom services as object or function
+             */
+            if (typeof service === 'object' && !serviceName) {
+                launcherServices.push(service)
+                continue
+            }
 
-        try {
             /**
              * add class service
              */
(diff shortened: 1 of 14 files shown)


christian-bromann:
@webdriverio/project-committers what do you all think about this solution? I didn't want to create a too big backwards incompatibility so I thought it could be a possibility to just check the source code of the plugin if it actually has a named export that is tried to be accessed.

christian-bromann:
I think I've found a better solution where I would just check in the main process if the service also exports a worker class and if so I would send this information along the worker so he can skip importing it. However I need #4926 first as I did some refactorings there that would be useful for this.

christian-bromann:
I think this solution will work. The launcher will always require all services and checks if the service also has a default export. If not it will pass this information along the worker which then can ignore this service and not require it. Missing pieces:

- tests

christian-bromann:
@webdriverio/project-committers this is now ready for review!

packages/wdio-utils/src/initialiseServices.js
```diff
@@ -5,64 +5,158 @@ import initialisePlugin from './initialisePlugin'
...
     const initialisedServices = []
-
-    if (!Array.isArray(config.services)) {
-        return initialisedServices
-    }
-
-    for (let serviceName of config.services) {
-        let serviceConfig = {}
-
+    for (let [serviceName, serviceConfig] of services) {
```
erwinheitzman: I don't think destructuring like this will work when the service is an object instead of an array

erwinheitzman: Or we need to update the example below from:
```
         * services: [{
         *     beforeTest: () => { ... }
         * }]
```
to:
```
         * services: [
                [{
         *         beforeTest: () => { ... }
         *     }]
         * ]
```

christian-bromann:
@erwinheitzman addressed your comments. Anything left to do?

christian-bromann:
Moving forward as mentioned in the project committers channel.

Issue #4949, opened 2020-01-10
`local` runner spawns one transpilation process for each feature file with cucumber framework


joscha:
**Environment (please complete the following information):**
 - **WebdriverIO version:** `5.18.4`
 - **Mode:** not sure what that means - `local`?
 - **If WDIO Testrunner, running sync/async:** `sync`
 - **Node.js version:** `v10.16.0`
 - **NPM version:** `not using npm; yarn 1.21.1`
 - **Browser name and version:** `Chrome`, but does not matter for this problem
 - **Platform name and version:** `OSX Mojave 10.14.6 (18G87)`
 - **Additional wdio packages used (if applicable):**
```
"@wdio/cli": "^5.18.4",
"@wdio/cucumber-framework": "^5.16.15",
"chai": "^4.2.0",
"chai-as-promised": "^7.1.1",
"ts-node": "^8.5.4",
"tsconfig-paths": "^3.9.0",
"typescript": "^3.7.4"
"@wdio/local-runner": "^5.18.4",
"@wdio/mocha-framework": "^5.16.15",
"@wdio/spec-reporter": "^5.16.11",
"@wdio/sync": "^5.17.0",
"chromedriver": "^79.0.0",
"wdio-chromedriver-service": "^5.0.2"
```

**Config of WebdriverIO**

https://github.com/joscha/wdio-ts-no-recompile/blob/master/wdio.conf.ts

EDIT: https://github.com/joscha/wdio-ts-no-recompile/blob/master/wdio.conf.js (removed one layer of Typescript to show the problem without noise)

**Describe the bug**

When using the cucumber framework with the `local` runner and Typescript it turns out that because the `local` runner spawns one subprocess per feature file the project gets transpiled `n` times for `n` feature files.

Our (mono-)repository has ~250 feature files and tens of thousands of Typescript files, which means that the overhead is immense (about 1 minute for each transpilation step; and that's without type-checking enabled).

We are trying to upgrade from wdio 4 (`4.14.4`) to wdio 5 and with the current `local` runner it doesn't seem possible to make this work.

Current thought is to do one of the following:

* Open a PR to the current `local` runner in this repository to teach it to run all feature files in once process if so desired (something something `maxWorkers=1`)
* Writing our own runner (am I thinking of doing this in the right spot or should it be a launcher?); downside of this is that we lose all the nice things the local runner brings (logging, repl, etc.)
* Writing a service to make this work (I am unsure of whether this would work, hints appreciated).

[...]

PR #4974, opened 2020-01-24
[#4949] WIP: feat: batch specs per suite

joscha:
This is a fix for #4949; it uses batching for the spec files, which mean it is possible to run more than one spec file in each test run.
Unfortunately the reporters expect a fixed 1 spec per test run, so reporting is broken.


joscha:
> The new option also needs to be added to [/docs/Options.md@`master`#wdio-options](https://github.com/webdriverio/webdriverio/blob/master/docs/Options.md#wdio-options)

will do, do you have any hints on how to best fix the reporting?

christian-bromann:
@joscha what is the status of this? Anything we can do to help you land this?

joscha:
Hey Christian,

Basically there are two ways to fix this:

* Run more than one spec per child process up to a number of Infinity specs in a child process, effectively allowing the user to decide what the optimum is between time spent transpiling and time spent in sequential spec runs
* Reusing workers and runners between runs, allowing to have a pool of runners that have a transpiled version of the code ready to run.

I tried both approaches and I think the first one is more manageable. The fix in the WIP pull request is doing the first approach and whilst it works perfectly for the runs itself the reporting is broken (and if I understand the did right most likely also some of the rerun logic as it is suite-based, not spec-based at the moment). Initially I thought it was just the spec reporter that was funky but looking at the code it seems like it was never intended to run more than one spec in a worker/runner sub process, hence the callbacks are all incorrect (they count up and down by one for the most part). I currently don't see a good way to split this refactor and I think some of the logic around it is also not covered by tests, yet.
Any recommendations around which part to tackle first or a general approach on how to make the callbacks spec-aware would be welcome!

Cheers,
Joscha

joscha:
@christian-bromann any chance we can iterate on this together, especially after seeing https://www.linkedin.com/posts/christian-bromann-01a91244_sauce-labs-deepens-investment-in-oss-community-activity-6638889939675164674-aX4r :)

christian-bromann:
> any chance we can iterate on this together

Sure thing. I was on PTO last week and will try to have a look this week.

joscha:
That would be great! Happy to jump on a zoom, Skype, Hangout, chat,
email-chain, Tik took ;) any time!

christian-bromann:
How about instead of adding a config `maxSpecsPerSuite` we allow the user more control as to which spec files are run in a single process. So instead of:

```js
// wdio.conf.js
specs: [
	"test1.js",
	"test2.js",
	"test3.js"
]
```

Where all spec files are being run in parallel, a user could define specs as `string | string[]`, e.g.:

```js
// wdio.conf.js
specs: [
	["test1.js", "test2.js", "test3.js"],
    ["test4.js"]
]
```

Here, test1 and test4 would start at the same time but after test1 it would start test2 and test3 sequentially in the same process.

joscha:
Re: string[] | string for specs definition: absolutely that would work as well if that is easier to implement.

joscha:
@christian-bromann was your suggestion in https://github.com/webdriverio/webdriverio/pull/4974#issuecomment-609704220 something that is implemented or planned currently?

christian-bromann:
> was your suggestion in [#4974 (comment)](https://github.com/webdriverio/webdriverio/pull/4974#issuecomment-609704220) something that is implemented or planned currently?

@joscha it is not implemented yet and I think it makes the most sense to specify multiple test suites per worker.

joscha:
Okay. If I implemented your sample API, would you be able to help with the
reporting?

Cheers,
Joscha

christian-bromann:
> Okay. If I implemented your sample API, would you be able to help with the reporting?

Yes!

christian-bromann:
@joscha any updates on this?

joscha:
Hi Christian,

Yes, we planned the change for the next development cycle. Once we
implemented your API we'll need help fixing the reporting.

abubakarafzal:
Is this functionality implemented?

joscha:
it is not (yet) unfortuantely @abubakarafzal

azinchenko91:
@josha Do you have any updates about this request?

joscha:
No, no good news, sorry, we're currently actually discussing abandoning
wdio altogether because of it.

christian-bromann:
We will tackle this effort in a separate issue #6469

PR #4975, opened 2020-01-24
[#4949] WIP: fix: transpile once only

joscha:
This PR exemplifies potential changes to fix the issue described in #4949.

It is a WIP and is there to discuss how this can be made to work in a better way.
The general gist is:

* it removes the child process used for the test execution
* it injects a dependency into all packages that assume to be executed in a child process which use `process.send` to communicate with the parent process that is used instead for communication.
* Reporting works as expected in contrary to #4974

### DIFF
--- a/packages/wdio-cucumber-framework/src/index.js
+++ b/packages/wdio-cucumber-framework/src/index.js
@@ -73,9 +73,14 @@ class CucumberAdapter {
         let result
 
         try {
+            // we need tp ensure that we don't run the registration code
+            // multiple times, otherwise we will get duplicated transpilations
+            // we also need to make sure that we only require spec files once
+            // otherwise they will be transpiled each time they are required
+            if(!global.supportCodeLibrary) {
+                // indentation is deliberately off so the diff is minimal
             this.registerRequiredModules()
             Cucumber.supportCodeLibraryBuilder.reset(this.cwd)
-
             /**
              * wdio hooks should be added before spec files are loaded
              */
@@ -89,7 +94,8 @@ class CucumberAdapter {
              */
             setUserHookNames(Cucumber.supportCodeLibraryBuilder.options)
             Cucumber.setDefaultTimeout(this.cucumberOpts.timeout)
-            const supportCodeLibrary = Cucumber.supportCodeLibraryBuilder.finalize()
+            global.supportCodeLibrary = Cucumber.supportCodeLibraryBuilder.finalize()
+            }
 
             /**
              * gets current step data: `{ uri, feature, scenario, step, sourceLocation }`
@@ -102,7 +108,7 @@ class CucumberAdapter {
             const runtime = new Cucumber.Runtime({
                 eventBroadcaster: this.eventBroadcaster,
                 options: this.cucumberOpts,
-                supportCodeLibrary,
+                supportCodeLibrary: global.supportCodeLibrary,
                 testCases: this.testCases
             })
 
--- a/packages/wdio-local-runner/src/worker.js
+++ b/packages/wdio-local-runner/src/worker.js
@@ -1,10 +1,9 @@
 import path from 'path'
-import child from 'child_process'
 import EventEmitter from 'events'
 
 import logger from '@wdio/logger'
+import Runner from '@wdio/runner'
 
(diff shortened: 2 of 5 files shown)


packages/wdio-cucumber-framework/src/index.js
```diff
@@ -73,9 +73,14 @@ class CucumberAdapter {
         let result
 
         try {
+            // we need tp ensure that we don't run the registration code
```
joscha: this change can potentially go away if we only ever have one `Runner` instance that we reuse across runs.

packages/wdio-cucumber-framework/src/index.js
```diff
@@ -89,7 +94,8 @@ class CucumberAdapter {
              */
             setUserHookNames(Cucumber.supportCodeLibraryBuilder.options)
             Cucumber.setDefaultTimeout(this.cucumberOpts.timeout)
-            const supportCodeLibrary = Cucumber.supportCodeLibraryBuilder.finalize()
+            global.supportCodeLibrary = Cucumber.supportCodeLibraryBuilder.finalize()
+            }
```
baruchvlz: Lines 81-97's scope is missing indentation.

packages/wdio-local-runner/src/worker.js
```diff
@@ -63,24 +62,50 @@ export default class WorkerInstance extends EventEmitter {
...
+                this.runner = new Runner(this)
+                this.worker = worker
+                this.runner.on('error', ({ name, message, stack }) => this.worker._handleError({
+                    origin: 'worker',
+                    name: 'error',
+                    content: { name, message, stack }
+                }))
+                this.runner.on('exit', ::this.worker._handleExit)
+            }
+            send(m) {
```
baruchvlz: Don't use single letter names for arguments, makes it difficult to understand the rest of the code.

packages/wdio-cucumber-framework/src/index.js
joscha: See comment above ;)

packages/wdio-local-runner/src/worker.js
joscha: That code is copied almost verbatim from `run.js`, I didn't touch the car naming - happy to change that though.

joscha:
Thanks for the review @baruchvlz - did you look at the original issue? Do you have some non-stylistic feedback as well by any chance?

baruchvlz:
@joscha I will still want to indent the block, don't want anyone else in the future thinking that this is ok :).

I would a bit more clarity on why you're using the `FakeChildProcess` class, I read the comment but it feels a bit hacky to me.

joscha:
Yes, the FakeChildProcess is a workaround for now. I don't intend to merge this as-is. The pull request is just to discuss whether this approach would be viable in general. I was hoping for comments, discussion and potentially a better approach from the maintainers but it seems to me that this might not happen, so maybe I should clean it up after all and we discuss actual mergeability.

baruchvlz:
Ah, I see. Sorry for the misunderstanding you can ignore my comments. I do have a question then, why can't we just transpile once before running and run the tests from those files instead of having to transpile on every execution?

> but it seems to me that this might not happen

Maybe the lack of comments is, in itself, a way of saying that this isn't the correct approach. And to be honest I'm not entirely convinced that passing this `target` to the `Runner` is the best solution.

> it removes the child process used for the test execution

Correct me if I'm wrong, but wouldn't this affect test execution? What will happen to concurrency if you remove child processes and everything runs on the main thread?

>  but it seems to me that this might not happen, so maybe I should clean it up after all and we discuss actual mergeability

Yea, this would probably be the fastest and best approach. You have to keep in mind that maintainers and committers have other responsibilities and must of us do this on our free time, so for some is hard to jump at every pull request and engage in conversations about it at all times. Completed pull requests will gain the most traction.

christian-bromann:
This approach won't work because it eliminates the ability to run tests in parallel. With Mocha (not sure about Jasmine and Cucumber) it just won't work to create multiple runner instances and have them run in parallel.

joscha:
Okay will close. Thanks for the feedback!

PR #5007, opened 2020-02-10
Use "devtools" automation protocol as fallback

christian-bromann:
## Proposed changes

Given that Puppeteer now supports Chrome, Firefox and Edge (Chromium) without having to install a browser it makes sense to use it as fallback in case a driver doesn't exist. This patch adds:

- support for Edge (Chromium) using the devtools automation protocol
- removal of `puppeteer-firefox` dependency (now lives directly within `puppeteer`)
- driver endpoint check and fallback to `devtools` if endpoint couldn't be reached

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

This is work in progress. It needs to be tested on Linux and Windows.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,6 +68,9 @@ matrix:
       addons:
         chrome: stable
         firefox: latest
+      before_script:
+        - "export DISPLAY=:99.0"
+        - "sh -e /etc/init.d/xvfb start"
       script: npx run-s test:e2e
 
 deploy:
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,14 @@ node_js:
   - 10
   - 12
 
+os:
+  - linux
+  - osx
+  - windows
+
 addons:
   chrome: stable
+  firefox: latest
 
 script:
   - npm run ci
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,7 +68,7 @@ matrix:
       addons:
         chrome: stable
         firefox: latest
-      before_script:
+      before_install:
         - "export DISPLAY=:99.0"
         - "sh -e /etc/init.d/xvfb start"
       script: npx run-s test:e2e
--- a/.travis.yml
+++ b/.travis.yml
@@ -86,6 +86,8 @@ matrix:
       name: 'e2e Tests (Windows)'
       os: windows
 
+allow_failures:
+    - name: 'e2e Tests (Windows)'
 
 deploy:
   provider: script
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,49 @@
 dist: xenial
 language: node_js
 
+cache:
+  directories:
+    - node_modules
+    - packages/devtools/node_modules
+    - packages/eslint-plugin-wdio/node_modules
+    - packages/node_modules/node_modules
+    - packages/wdio-allure-reporter/node_modules
+    - packages/wdio-appium-service/node_modules
+    - packages/wdio-applitools-service/node_modules
+    - packages/wdio-browserstack-service/node_modules
+    - packages/wdio-cli/node_modules
+    - packages/wdio-concise-reporter/node_modules
+    - packages/wdio-config/node_modules
+    - packages/wdio-crossbrowsertesting-service/node_modules
+    - packages/wdio-cucumber-framework/node_modules
+    - packages/wdio-devtools-service/node_modules
+    - packages/wdio-dot-reporter/node_modules
+    - packages/wdio-firefox-profile-service/node_modules
+    - packages/wdio-jasmine-framework/node_modules
+    - packages/wdio-junit-reporter/node_modules
+    - packages/wdio-lambda-runner/node_modules
+    - packages/wdio-local-runner/node_modules
+    - packages/wdio-logger/node_modules
+    - packages/wdio-mocha-framework/node_modules
+    - packages/wdio-protocols/node_modules
+    - packages/wdio-repl/node_modules
+    - packages/wdio-reporter/node_modules
+    - packages/wdio-runner/node_modules
+    - packages/wdio-sauce-service/node_modules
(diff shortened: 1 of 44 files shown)


packages/webdriverio/src/utils.js
```diff
@@ -414,3 +415,33 @@ export const enhanceElementsArray = (elements, parent, selector, foundWith = '$$
  * @param {string} automationProtocol
  */
 export const isStub = (automationProtocol) => automationProtocol === './protocol-stub'
+
+export const getAutomationProtocol = async (config) => {
+    /**
+     * don't modify automation protocol if hostname is set as it is very likely
+     * that "webdriver" will be used
+     */
+    if (config.hostname || (config.user && config.key)) {
```
mgrybyk: should we add config.port as well?

mgrybyk: should we add config.port as well?

packages/webdriverio/src/utils.js
```diff
@@ -414,3 +415,33 @@ export const enhanceElementsArray = (elements, parent, selector, foundWith = '$$
...
+
+    /**
+     * make a head request to check if a driver is available
+     */
+    const driverEndpointHeaders = await new Promise((resolve) => {
+        const req = http.request({
+            method: 'GET',
+            host: 'localhost',
+            port: config.port || 4444,
+            path: '/status'
```
mgrybyk: isn't config.path ignored in such case?

christian-bromann: I will remove `config.port` - if nothing is specified we need to assume that the driver is on `localhost:4444`.

mgrybyk: Yeah, but path /status is not necessarily valid. Can it be /wd/hub/path ?

christian-bromann: `/status` is defined by the WebDriver spec and therefor has to be there

christian-bromann:
@webdriverio/project-committers 

This is finally ready for review, here is the Gist of things that change with this patch:

- I updated the Travis build to run lint, unit and smoke test separated from e2e (given that they require a different setup where browser need to be installed)

Build before:
Screenshot 2020-02-15 at 18 37 54

Build now:
Screenshot 2020-02-15 at 18 37 36

Given that we have 5 concurrent VMs the overall time shouldn't change much.

- I added support for Chromium Edge (currently only Mac and Windows)
- The DevTools automation is now default, that means if WebdriverIO can't find a driver endpoint it defaults to that
- due to a weird bug in Jest I had to change the setup for test coverage which now always includes all files (even the ones that are not test)

.travis.yml
```diff
@@ -1,18 +1,91 @@
 dist: xenial
 language: node_js
 
-node_js:
-  - 10
-  - 12
+cache:
+  directories:
```
mgrybyk: do we need to cache website's node modules here?

mgrybyk:
is it due to Jest' bug there are some index files ignored `/* istanbul ignore file */` ?

packages/devtools/package.json
```diff
@@ -35,12 +35,9 @@
     "@wdio/protocols": "6.0.0-alpha.1",
     "@wdio/utils": "6.0.0-alpha.1",
     "chrome-launcher": "^0.11.1",
-    "puppeteer-core": "^1.18.1"
+    "puppeteer-core": "^2.1.1"
   },
   "devDependencies": {
     "@types/puppeteer": "^1.19.0"
```
mgrybyk: let's update types to 2.0.0 ?

packages/webdriverio/src/constants.js
```diff
@@ -422,3 +421,9 @@ export const APPIUM_CAPABILITES = [
     ...APPIUM_ANDROID_CAPABILITIES,
     ...APPIUM_IOS_CAPABILITIES
 ]
+export const DRIVER_DEFAULT_ENDPOINT = {
+    method: 'GET',
+    host: 'localhost',
+    port: 4444,
+    path: '/status'
```
mgrybyk: path should respect the `path` defined by user in config! For selenium v3 and some cloud providers it should be /wd/hub/status

I've just verified it locally

e2e/testrunner.test.js
```diff
@@ -1,10 +1,28 @@
...
+    const hasPassed = failures === 0
+
+    /**
+     * print log files for debugging if test fails
+     */
+    if (!hasPassed) {
+        const rootPath = path.join(__dirname, 'wdio')
+        const logFiles = fs.readdirSync(rootPath).filter((file) => (
+            // only log files
+            file.endsWith('.log')
```
erwinheitzman: This filter does not work as `file.endsWith('.log')` needs to be returned here.

.travis.yml
```diff
@@ -1,18 +1,93 @@
...
+    - node_modules
+    - packages/devtools/node_modules
```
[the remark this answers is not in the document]
christian-bromann: well we don't want to cache our build files

e2e/testrunner.test.js
christian-bromann: it returns because the surrounding `(` and `)`

christian-bromann: made it more clear

packages/webdriverio/src/constants.js
christian-bromann: by default all drivers should start with path `/` if you change that we should assume you run `webdriver`. I modified the `getAutomationProtocol` method to now take this into consideration:

```js
/**
 * run WebDriver if hostname or port is set
 */
if (config.hostname || config.port || config.path || (config.user && config.key)) {
    return 'webdriver'
}
```

christian-bromann:
@mgrybyk @erwinheitzman addressed your comments

christian-bromann:
Moving forward as mentioned in the project committers channel.

PR #5016, opened 2020-02-13
wdio-cli: Add Experitest Cloud to configuration utility

ArielExperitest:
## Proposed changes

WebdriverIO users will be able to setup a config file that integrates with the Experitest Cloud with the help of wdio-cli.
Experitest users will need only access key from their Cloud.
		
If Experitest integration is chosen we will add to the config file:
- 'hostname' - The hostname of the Cloud (example.ex.com) - User configurable
- 'protocol' - HTTP or HTTPS - User configurable
- 'port' - 80, 443 or other - User configurable
- 'path' - /wd/hub
- capabilities.accessKey - User configurable

## Types of changes

WebdriverIO users can have integration with Experitest Cloud

- [ ] Documentation update

## Checklist

- [x] I have added tests that prove my fix is effective or that my feature works - **Not applicable because the change is only in the configuration utility**
- [x] I have added necessary documentation (if appropriate) - **Documentation is not applicable because the change is only in the configuration utility**
- [x] I have added proper type definitions for new commands (if appropriate) - **Not applicable because the change is only in the configuration utility**

## Further comments

### DIFF
--- a/packages/wdio-cli/src/constants.js
+++ b/packages/wdio-cli/src/constants.js
@@ -99,7 +99,7 @@ export const QUESTIONNAIRE = [{
     message: 'Where is your automation backend located?',
     choices: [
         'On my local machine',
-        'In the Experitest cloud',
+        'In the cloud using Experitest',
         'In the cloud using Sauce Labs',
         'In the cloud using Browserstack or Testingbot or a different service',
         'I have my own Selenium cloud'
@@ -117,43 +117,43 @@ export const QUESTIONNAIRE = [{
     when: /* istanbul ignore next */ (answers) => answers.backend.indexOf('different service') > -1
 }, {
     type: 'input',
-    name: 'exp_env_access_key',
+    name: 'expEnvAccessKey',
     message: 'Access key from Experitest Cloud',
     default: 'EXPERITEST_ACCESS_KEY',
-    when: /* istanbul ignore next */ (answers) => answers.backend === 'In the Experitest cloud'
+    when: /* istanbul ignore next */ (answers) => answers.backend === 'In the cloud using Experitest'
 }, {
     type: 'input',
-    name: 'exp_env_hostname',
+    name: 'expEnvHostname',
     message: 'Environment variable for cloud url',
     default: 'example.experitest.com',
-    when: /* istanbul ignore next */ (answers) => answers.backend === 'In the Experitest cloud'
+    when: /* istanbul ignore next */ (answers) => answers.backend === 'In the cloud using Experitest'
 }, {
     type: 'list',
-    name: 'exp_env_port',
+    name: 'expEnvPort',
     message: 'Choose a port for environment variable',
     default: 443,
     choices: [
-        443,
-        80,
+        '443',
+        '80',
         'Other'
     ],
-    when: /* istanbul ignore next */ (answers) => answers.backend === 'In the Experitest cloud'
+    when: /* istanbul ignore next */ (answers) => answers.backend === 'In the cloud using Experitest'
 }, {
     type: 'input',
-    name: 'exp_env_port',
+    name: 'expEnvPort',
     message: 'Environment variable for other port',
-    default: 'EXPERITEST PORT',
-    when: /* istanbul ignore next */ (answers) => answers.exp_env_port === 'Other'
+    default: 'OTHER_PORT',
+    when: /* istanbul ignore next */ (answers) => answers.expEnvPort === 'Other'
 }, {
     type: 'list',
-    name: 'exp_env_protocol',
+    name: 'expEnvProtocol',
     message: 'Choose a protocol for environment variable',
     default: 'https',
     choices: [
         'https',
         'http'
     ],
-    when: /* istanbul ignore next */ (answers) => answers.backend === 'In the Experitest cloud'
+    when: /* istanbul ignore next */ (answers) => answers.expEnvPort && answers.expEnvPort !== '80' && answers.expEnvPort !== '443'
 }, {
     type: 'input',
     name: 'env_user',
--- a/packages/wdio-cli/src/constants.js
+++ b/packages/wdio-cli/src/constants.js
@@ -99,6 +99,7 @@ export const QUESTIONNAIRE = [{
     message: 'Where is your automation backend located?',
     choices: [
         'On my local machine',
+        'In the Experitest cloud',
         'In the cloud using Sauce Labs',
         'In the cloud using Browserstack or Testingbot or a different service',
         'I have my own Selenium cloud'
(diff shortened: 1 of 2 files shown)


packages/wdio-cli/src/templates/wdio.conf.tpl.ejs
```diff
@@ -7,7 +7,13 @@ exports.config = {
     // WebdriverIO allows it to run your tests in arbitrary locations (e.g. locally or
     // on a remote machine).
     runner: '<%= answers.runner %>',<%
-    if(answers.hostname && answers.port) { %>
```
christian-bromann: this would break existing behavior if someone uses custom port or hostname

packages/wdio-cli/src/templates/wdio.conf.tpl.ejs
```diff
@@ -84,6 +90,7 @@ exports.config = {
     // https://docs.saucelabs.com/reference/platforms-configurator
     //
     capabilities: [{
+    <%if(answers.exp_env_access_key){%> accessKey: '<%= answers.exp_env_access_key %>',<%}%>
```
christian-bromann: this should be added by WebdriverIO as it does already with Sauce Labs or BrowserStack

packages/wdio-cli/src/constants.js
```diff
@@ -114,6 +115,45 @@ export const QUESTIONNAIRE = [{
...
+}, {
+    type: 'input',
+    name: 'exp_env_port',
+    message: 'Environment variable for other port',
+    default: 'EXPERITEST PORT',
+    when: /* istanbul ignore next */ (answers) => answers.exp_env_port === 'Other'
+}, {
+    type: 'list',
+    name: 'exp_env_protocol',
+    message: 'Choose a protocol for environment variable',
```
christian-bromann: I don't think we need to ask this, `https` only!

packages/wdio-cli/src/constants.js
```diff
@@ -114,6 +115,45 @@ export const QUESTIONNAIRE = [{
...
+    choices: [
+        443,
+        80,
+        'Other'
+    ],
+    when: /* istanbul ignore next */ (answers) => answers.backend === 'In the Experitest cloud'
+}, {
+    type: 'input',
+    name: 'exp_env_port',
+    message: 'Environment variable for other port',
```
christian-bromann: Does Experitest allow multiple ports to be set? If not we don't need to ask this.

packages/wdio-cli/src/constants.js
```diff
@@ -114,6 +115,45 @@ export const QUESTIONNAIRE = [{
...
+}, {
+    type: 'input',
+    name: 'exp_env_hostname',
+    message: 'Environment variable for cloud url',
+    default: 'example.experitest.com',
+    when: /* istanbul ignore next */ (answers) => answers.backend === 'In the Experitest cloud'
+}, {
+    type: 'list',
+    name: 'exp_env_port',
+    message: 'Choose a port for environment variable',
```
christian-bromann: Let's only allow SSL ports (443), I don't think there is a scenario where someone prefers an HTTP connection

packages/wdio-cli/src/constants.js
ArielExperitest: Experitest offers an on-premise (non-SASS) selenium hub for our customers. 
As such our customers are those who decide which protocol or port to use.
We cannot limit them to just HTTPS.

packages/wdio-cli/src/constants.js
ArielExperitest: Experitest offers an on-premise (non-SASS) selenium hub for our customers. 
As such our customers are those who decide which protocol or port to use.
We cannot limit them to port 443.

packages/wdio-cli/src/templates/wdio.conf.tpl.ejs
ArielExperitest: No, line 10 moved to line 16.

packages/wdio-cli/src/constants.js
ArielExperitest: Only a single port is allowed but it's can be configurable by the customer to be any port.

ArielExperitest:
> 
> 
> Can we re-use `user` and `key` instead of using custom `exp_env_access_key` variables.

No, exp_env_access_key is a unique key for a user.

ArielExperitest:
> 
> 
> Can we re-use `user` and `key` instead of using custom `exp_env_access_key` variables.

No, exp_env_access_key is a unique key for a user.

christian-bromann:
I don't think we need to add these question as they are not special to Experitest. If someone would pick `In the cloud using Browserstack or Testingbot or a different service` he is also able to pick hostname or possible. It would be better if we can change existing wording to make it more clear that people using Experitest can do the "custom hostname" route.

ArielExperitest:
> 
> 
> I don't think we need to add these question as they are not special to Experitest. If someone would pick `In the cloud using Browserstack or Testingbot or a different service` he is also able to pick hostname or possible. It would be better if we can change existing wording to make it more clear that people using Experitest can do the "custom hostname" route.

Experitest offers for our customers an on-premise (non-SASS) selenium hub in Experitest Cloud.
As such our customers Cloud not necessarily in the experitest.com domain and there decide which protocol or port to use.

In "In the cloud using Browserstack or Testingbot or a different service" option, 
we need only accessKey and not user name variable in our setup. We do need the user to specify the cloud's hostname and protocol (HTTP/HTTPS) because we have customers using both.

Hope this clarifies our PR. Maybe I missed something. can you explain more about your suggestion?

packages/wdio-cli/src/templates/wdio.conf.tpl.ejs
christian-bromann: this is not W3C conform .. is Experitest planning to use vendor capabilities?

packages/wdio-cli/src/constants.js
```diff
@@ -99,6 +99,7 @@ export const QUESTIONNAIRE = [{
     message: 'Where is your automation backend located?',
     choices: [
         'On my local machine',
+        'In the Experitest cloud',
```
christian-bromann: How about we align the wording to 
```suggestion
        'In the cloud using Experitest',
```

packages/wdio-cli/src/constants.js
```diff
@@ -114,6 +115,45 @@ export const QUESTIONNAIRE = [{
     message: 'What is the port on which that service is running?',
     default: '80',
     when: /* istanbul ignore next */ (answers) => answers.backend.indexOf('different service') > -1
+}, {
+    type: 'input',
+    name: 'exp_env_access_key',
```
christian-bromann: Can we use camelCase for property names? I know we haven't been consistent (e.g. `env_user`) but we shouldn't continue with this.

packages/wdio-cli/src/constants.js
christian-bromann: let's auto set the protocol if protocol is either `http` or `https` and only ask this question if port is custom

packages/wdio-cli/src/templates/wdio.conf.tpl.ejs
ArielExperitest: This 'accessKey' is Experitest capability and used for authentication in Experitest Cloud.
please let me know if it's ok so I can push all your suggestions

christian-bromann: @ArielExperitest .. I am assuming you work for Experitest. This key is invalid. According to the spec the following capabilities are allowed: https://w3c.github.io/webdriver/#capabilities. You have to vendor prefix it like `"experitest:accessKey"`.

packages/wdio-cli/src/templates/wdio.conf.tpl.ejs
```diff
@@ -7,7 +7,15 @@ exports.config = {
     // WebdriverIO allows it to run your tests in arbitrary locations (e.g. locally or
     // on a remote machine).
     runner: '<%= answers.runner %>',<%
-    if(answers.hostname && answers.port) { %>
+
+    if(answers.expEnvAccessKey){ %>
+    hostname: '<%= answers.expEnvHostname %>',
+    <% if (answers.expEnvPort === '443'){%>protocol: https,
```
christian-bromann:
```suggestion
    <% if (answers.expEnvPort === '443'){%>protocol: 'https',
```

packages/wdio-cli/src/templates/wdio.conf.tpl.ejs
```diff
@@ -7,7 +7,15 @@ exports.config = {
...
+    <%} else if (answers.expEnvPort === '80'){%>protocol: http,
```
christian-bromann:
```suggestion
    <%} else if (answers.expEnvPort === '80'){%>protocol: 'http',
```

christian-bromann:
@ArielExperitest have you checked if passing `experitest:accessKey` into your capabilities works?

ArielExperitest:
> 
> 
> @ArielExperitest have you checked if passing `experitest:accessKey` into your capabilities works?

The change right now in QA and will publish in the next few days.

christian-bromann:
Awesome, let me know when it is live and we will land this.

ArielExperitest:
> 
> 
> Awesome, let me know when it is live and we will land this.

We did the change on our end, and it will be released as part 20.2 (to be released by the end of February).
We would like to test it before our customers.
Anything else is required to complete the pull-request?

PR #5029, opened 2020-02-16
Remove appium bug workaround

erwinheitzman:
## Proposed changes

Remove workaround for a bug that was around in Appium almost a year ago.

## Checklist


alfonso-presa:
@erwinheitzman @christian-bromann Just fyi.... Removing this workaround broke our test automation using Saucelabs as appium 1.8.0 was being provided by default by such platform. Changing to appium 1.16.0 solved the issue (although I would have prefered not to stick to an specific one). I'm unsure about intermediate versions compatibility.

christian-bromann:
@alfonso-presa so you specified the Appium version to be `1.8.0`? I don't think you should be afraid of version in-compatibility. The Appium team ensures that minor updates are backwards compatible.

alfonso-presa:
@christian-bromann nope. We were no specifying any, and SL was providing 1.8.0 by default. Forcing 1.16.0 from capabilities fixed the issue.

PR #5039, opened 2020-02-18
Implement `performActions` for devtools automation protocol

christian-bromann:
## Proposed changes

These are the last two important protocol commands we have to implement for the devtools automation protocol. Given that there are a few commands (e.g. `keys` or `dragAndDrop`) that use `performActions` this was necessary to do before the v6 release.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/e2e/element.test.js
+++ b/e2e/element.test.js
@@ -122,6 +122,7 @@ describe('elements', () => {
     })
 
     it('should be able to do a drag&drop', async () => {
+        await browser.executeScript('window.scrollTo(0, 0)', [])
         await browser.performActions([{
             type: 'pointer',
             id: 'finger1',
--- a/e2e/element.test.js
+++ b/e2e/element.test.js
@@ -120,6 +120,64 @@ describe('elements', () => {
         const link = await browser.findElement('partial link text', 'new tab')
         expect(await browser.getElementText(link[ELEMENT_KEY])).toBe('open new tab')
     })
+
+    it('should be able to do a drag&drop', async () => {
+        await browser.performActions([{
+            type: 'pointer',
+            id: 'finger1',
+            parameters: {
+                pointerType: 'mouse'
+            },
+            actions: [{
+                type: 'pointerMove',
+                duration: 0,
+                x: 65,
+                y: 544
+            }, {
+                type: 'pointerDown',
+                button: 0
+            }, {
+                type: 'pause',
+                duration: 10
+            }, {
+                type: 'pointerMove',
+                duration: 100,
+                origin: 'pointer',
+                x: 1,
+                y: -251
+            }, {
+                type: 'pointerUp',
+                button: 0
+            }]
+        }])
+
+        const elem = await browser.$('.searchinput')
+        expect(await browser.getElementProperty(elem[ELEMENT_KEY])).toBe('Dropped!')
+    })
+
+    it('should be able to use keys command', async () => {
+        const textarea = await browser.findElement('css selector', 'textarea')
+        await browser.elementClick(textarea[ELEMENT_KEY])
+        await browser.performActions([{
+            type: 'key',
+            id: 'keyboard',
+            actions: [
+                { type: 'keyDown', value: 'f' },
+                { type: 'keyDown', value: 'o' },
+                { type: 'keyDown', value: 'o' },
+                { type: 'keyDown', value: 'b' },
+                { type: 'keyDown', value: 'a' },
+                { type: 'keyDown', value: 'r' },
+                { type: 'keyUp', value: 'f' },
+                { type: 'keyUp', value: 'o' },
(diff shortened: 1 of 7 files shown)


e2e/element.test.js
```diff
@@ -120,6 +120,66 @@ describe('elements', () => {
...
+                button: 0
+            }]
+        }])
+
+        const elem = await browser.findElement('css selector', '.searchinput')
+        expect(await browser.getElementProperty(elem[ELEMENT_KEY], 'value'))
+            .toBe('Dropped!')
+    })
+
+    it('should be able to use keys command', async () => {
```
erwinheitzman: Mind if we also test keys like SHIFT and an emoji?

packages/devtools/src/commands/performActions.js
```diff
@@ -0,0 +1,82 @@
...
+                        x += lastPointer.x
+                        y += lastPointer.y
+                    }
+
+                    lastPointer.x = x
+                    lastPointer.y = y
+                    await keyboardFn(x, y, { steps: 10 })
+                    continue
+                } else {
+                    const pptrButton = (
```
erwinheitzman: I would make 'left' the default.
I would also prefer a `switch` or some `if` statements for readability.

packages/devtools/src/devtoolsdriver.js
```diff
@@ -103,9 +103,11 @@ export default class DevToolsDriver {
             }
 
             this.emit('result', { command, params, retries, result })
-            log.info('RESULT', command.toLowerCase().includes('screenshot')
-                && typeof result === 'string' && result.length > 64
-                ? `${result.substr(0, 61)}...` : result)
+            if (typeof result !== 'undefined') {
```
erwinheitzman: I would like the readability improved here as well by splitting the condition or use variables/functions for naming perhaps

packages/devtools/src/commands/performActions.js
```diff
@@ -0,0 +1,82 @@
...
+    const lastPointer = {}
+
+    /**
+     * see https://github.com/jlipps/simple-wd-spec#input-sources-and-corresponding-actions
+     * for details on the `actions` format
+     */
+    for (const action of actions) {
+        if (action.type === null || action.type === 'null') {
+            for (const singleAction of action.actions) {
+                await sleep(singleAction.duration)
```
erwinheitzman: `singleAction.duration` can be undefined here?

e2e/element.test.js
christian-bromann: I added a test that includes an emoji. It seems that using shift to get characters written as uppercase seems to be not possible. I would maybe create a bug to investigate this later but wouldn't see it as a show stopper.

packages/devtools/src/commands/performActions.js
christian-bromann: I added a comment for better readability and made `left` the default value. Using switch here doesn't seem to make the code more readable IMO.

packages/devtools/src/devtoolsdriver.js
christian-bromann: created an `isScreenshot` variable to make it more readable.

packages/devtools/src/commands/performActions.js
```diff
@@ -0,0 +1,109 @@
...
+
+        if (action.type === 'key') {
+            const skipChars = []
+            for (const singleAction of action.actions) {
+                if (singleAction.type === 'pause') {
+                    await sleep(singleAction.duration)
+                    continue
+                }
+
+                const cmd = singleAction.type.slice(3).toLowerCase()
```
mgrybyk: What if we avoid using constants like 3 or 7 and use action.type length instead or something like that to make it more understandable? Maybe define function to get cmd based on action.type

Same for pointer

packages/devtools/src/commands/performActions.js
```diff
@@ -0,0 +1,109 @@
...
+                const keyboardFn = ::page.keyboard[cmd]
```
mgrybyk: should we check if cmd is one of type/press/down/up/sendCharacter ?

ref https://github.com/puppeteer/puppeteer/blob/v2.1.1/docs/api.md#class-keyboard

erwinheitzman:
I think Mykola has some valid points. Should we address these?

christian-bromann:
> Should we address these?

Absolutely, just didn't have the time to circle back to this.

packages/devtools/src/commands/performActions.js
christian-bromann: I use either `down`/`up` or `sendCharacter` as I thought they would best represent what the protocol is requesting.

christian-bromann:
@erwinheitzman @mgrybyk addressed both comments, anything left to do here?

PR #5062, opened 2020-02-26
webdriver: Add interceptors for HTTP requests and responses [v5]

janizde:
## Proposed changes

This PR introduces two new configuration options for `webdriver` (and `webdriverio`) with which the HTTP communication between `webdriver` and the server can be intercepted. In the config, the functions

* `transformRequest(RequestOptions): RequestOptions` and
* `transformResponse(Response, RequestOptions): Response`

can be specified and are then called in `webdriver/src/request.js`

This can also be seen as a generalization of the options `agent` and `headers`, which I have left in to keep compatibility.

The PR is the result of [a conversation on Gitter](https://gitter.im/webdriverio/webdriverio?at=5e43fb51cd2d737bb072b4fa) where the specific needs for my use case are stated.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

* I based this PR on `v5`, but I'm happy to provide a similar solution (with `got` instead of `request`) for v6 once we agreed on how to proceed with this (there's already a branch, see https://github.com/webdriverio/webdriverio/compare/master...janizde:http-request-interceptors-v6?expand=0)
* Alternatively to passing the `request.Response` to `transformResponse` I thought about only passing the response body, but ended up giving full control over the response object. Happy to discuss this!
* ~~As for the typings I have added `transformRequest: (requestOptions: any): any` and `transformResponse(response: any, requestOptions: any): any` but couldn't get more specific since as soon as `import request from 'request'` the definition file becomes a module instead of a script. Also I couldn't get it to work with `/// <reference types="request"/>`. Any idea how to make this more specific?~~

### DIFF
--- a/packages/webdriver/package.json
+++ b/packages/webdriver/package.json
@@ -36,5 +36,8 @@
     "@wdio/utils": "5.18.6",
     "lodash.merge": "^4.6.1",
     "request": "^2.83.0"
+  },
+  "devDependencies": {
+    "@types/request": "^2.48.4"
   }
 }
--- a/packages/webdriver/src/request.js
+++ b/packages/webdriver/src/request.js
@@ -117,7 +117,7 @@ export default class WebDriverRequest extends EventEmitter {
                 response = transformResponse(response, fullRequestOptions)
             }
 
-            let { body } = response
+            let body = response && response.body
             const error = err || getErrorFromResponseBody(body)
 
             /**
--- a/packages/webdriver/src/request.js
+++ b/packages/webdriver/src/request.js
@@ -168,7 +168,7 @@ export default class WebDriverRequest extends EventEmitter {
             this.emit('retry', { error, retryCount })
             log.warn('Request failed due to', error.message)
             log.info(`Retrying ${retryCount}/${totalRetryCount}`)
-            this._request(fullRequestOptions, totalRetryCount, retryCount, transformResponse).then(resolve, reject)
+            this._request(fullRequestOptions, transformResponse, totalRetryCount, retryCount).then(resolve, reject)
         }))
     }
 }
--- a/packages/webdriver/src/request.js
+++ b/packages/webdriver/src/request.js
@@ -39,7 +39,7 @@ export default class WebDriverRequest extends EventEmitter {
 
     makeRequest (options, sessionId) {
         let fullRequestOptions = Object.assign({}, this.defaultOptions, this._createOptions(options, sessionId))
-        if (options.transformRequest) {
+        if (typeof options.transformRequest === 'function') {
             fullRequestOptions = options.transformRequest(fullRequestOptions)
         }
 
@@ -113,7 +113,7 @@ export default class WebDriverRequest extends EventEmitter {
         }
 
         return new Promise((resolve, reject) => request(fullRequestOptions, (err, response) => {
-            if (transformResponse) {
+            if (typeof transformResponse === 'function') {
                 response = transformResponse(response, fullRequestOptions)
             }
 
@@ -168,7 +168,7 @@ export default class WebDriverRequest extends EventEmitter {
             this.emit('retry', { error, retryCount })
             log.warn('Request failed due to', error.message)
             log.info(`Retrying ${retryCount}/${totalRetryCount}`)
-            this._request(fullRequestOptions, totalRetryCount, retryCount).then(resolve, reject)
+            this._request(fullRequestOptions, totalRetryCount, retryCount, transformResponse).then(resolve, reject)
         }))
     }
 }
--- a/packages/webdriver/src/request.js
+++ b/packages/webdriver/src/request.js
@@ -105,7 +105,7 @@ export default class WebDriverRequest extends EventEmitter {
         return requestOptions
     }
 
-    _request (fullRequestOptions, totalRetryCount = 0, retryCount = 0, transformResponse) {
+    _request (fullRequestOptions, transformResponse, totalRetryCount = 0, retryCount = 0) {
         log.info(`[${fullRequestOptions.method}] ${fullRequestOptions.uri.href}`)
 
         if (fullRequestOptions.body && Object.keys(fullRequestOptions.body).length) {
--- a/packages/webdriver/src/request.js
+++ b/packages/webdriver/src/request.js
@@ -38,9 +38,13 @@ export default class WebDriverRequest extends EventEmitter {
     }
 
     makeRequest (options, sessionId) {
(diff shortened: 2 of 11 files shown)


packages/webdriver/webdriver.d.ts
```diff
@@ -370,6 +370,8 @@ declare namespace WebDriver {
         headers?: {
             [name: string]: string;
         };
+        transformRequest?: (requestOptions: any) => any;
+        transformResponse?: (requestOptions: any, response: any) => any;
```
christian-bromann: `requestOptions` and `response` are pretty much well defined, especially for [response](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/got/index.d.ts#L313-L325). For `requestOptions` it would be great to have at least some decent type support.

janizde: Thanks for looking into it! As described in the summary I've tried to add the typings from `@types/request` / `@types/got` but I can't figure out how I can reference those without an `import`. Is there a way to do it with `/// <reference types="requets" />` ? Or should I just copy over the typings into the definition files of webdriver?

christian-bromann: I think it should be pretty similar to how we do it e.g. in the Cucumber framework where we have the types as [a dependency](https://github.com/webdriverio/webdriverio/blob/master/packages/wdio-cucumber-framework/package.json#L41) and import it in the [`*.d.ts` file](https://github.com/webdriverio/webdriverio/blob/master/packages/wdio-cucumber-framework/cucumber-framework.d.ts#L1)

christian-bromann:
@janizde is there any chance to also make this PR for v6?

janizde:
> I based this PR on v5, but I'm happy to provide a similar solution (with got instead of request) for v6 once we agreed on how to proceed with this (there's already a branch, see https://github.com/webdriverio/webdriverio/compare/master...janizde:http-request-interceptors-v6?expand=0)

I have a branch already for v6, still need to apply the changes with the advanced typings to it but this should be quick. Would you prefer opening the two PRs at once or only create the v6 PR when this one is done?

christian-bromann:
> Would you prefer opening the two PRs at once or only create the v6 PR when this one is done?

Let's have them both ready at once so we can ship the feature for both versions.

janizde:
alright, here it is https://github.com/webdriverio/webdriverio/pull/5082

christian-bromann:
Thanks @janizde 

Please update your v5 base branch and the CI failure should go away.

janizde:
Both are green now 👍

packages/webdriver/src/request.js
```diff
@@ -38,9 +38,13 @@ export default class WebDriverRequest extends EventEmitter {
...
     makeRequest (options, sessionId) {
-        const fullRequestOptions = Object.assign({}, this.defaultOptions, this._createOptions(options, sessionId))
+        let fullRequestOptions = Object.assign({}, this.defaultOptions, this._createOptions(options, sessionId))
+        if (typeof options.transformRequest === 'function') {
+            fullRequestOptions = options.transformRequest(fullRequestOptions)
+        }
+
         this.emit('request', fullRequestOptions)
-        return this._request(fullRequestOptions, options.connectionRetryCount)
+        return this._request(fullRequestOptions, options.connectionRetryCount, 0, options.transformResponse)
```
christian-bromann: Let's change order of arguments here so that we can still work with defaults.

```suggestion
        return this._request(fullRequestOptions, options.transformResponse, options.connectionRetryCount)
```

packages/webdriver/src/request.js
```diff
@@ -101,14 +105,19 @@ export default class WebDriverRequest extends EventEmitter {
         return requestOptions
     }
 
-    _request (fullRequestOptions, totalRetryCount = 0, retryCount = 0) {
+    _request (fullRequestOptions, totalRetryCount = 0, retryCount = 0, transformResponse) {
```
christian-bromann:
```suggestion
    _request (fullRequestOptions, transformResponse, totalRetryCount = 0, retryCount = 0) {
```

packages/webdriver/src/request.js
christian-bromann: The pattern is usually: required parameters first, optional last

packages/webdriver/src/request.js
```diff
@@ -159,7 +168,7 @@ export default class WebDriverRequest extends EventEmitter {
             this.emit('retry', { error, retryCount })
             log.warn('Request failed due to', error.message)
             log.info(`Retrying ${retryCount}/${totalRetryCount}`)
-            this._request(fullRequestOptions, totalRetryCount, retryCount).then(resolve, reject)
+            this._request(fullRequestOptions, totalRetryCount, retryCount, transformResponse).then(resolve, reject)
```
christian-bromann:
```suggestion
            this._request(fullRequestOptions, transformResponse, totalRetryCount, retryCount).then(resolve, reject)
```

tests/typings/copy.js
```diff
@@ -33,6 +33,11 @@ const packages = {
     '@types/jasmine/ts3.1': 'packages/wdio-jasmine-framework/node_modules/@types/jasmine/ts3.1',
     '@wdio/jasmine-framework': 'packages/wdio-jasmine-framework',
 
+    '@types/request': 'packages/webdriver/node_modules/@types/request',
+    '@types/caseless': 'packages/webdriver/node_modules/@types/caseless',
+    '@types/tough-cookie': 'packages/webdriver/node_modules/@types/tough-cookie',
+    'form-data': 'packages/webdriver/node_modules/@types/request/node_modules/form-data',
```
christian-bromann: Can we add a typings test here:

- https://github.com/webdriverio/webdriverio/blob/v5/tests/typings/webdriverio/async.ts#L139
- https://github.com/webdriverio/webdriverio/blob/v5/tests/typings/sync/config.ts

janizde: It works for the sync test but when I'm attaching the config to `async` it doesn't recognize `transformRequest`. Also it does not recognize any of the other config options coming from `webdriver` like `headers` or `agent`. Any hint what has to be changed there?

christian-bromann: I am currently looking into this. I don't want to make this a blocker for this PR.

packages/webdriver/package.json
```diff
@@ -36,5 +36,8 @@
     "@wdio/utils": "5.18.6",
     "lodash.merge": "^4.6.1",
     "request": "^2.83.0"
+  },
+  "devDependencies": {
+    "@types/request": "^2.48.4"
```
dylang: This needs be a `dependency` because it's now used in the types this package exports.

PR #5064, opened 2020-02-27
Fix selector typing

Kignuf:
## Proposed changes

Hello, I noticed there is a glitch in the typings for $ and $$. They both take a string or function as argument, but it actually should be string | function | object since the introduction of android data matcher strategy.

## Types of changes

- [x] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/wdio-sync/webdriverio-core.d.ts
+++ b/packages/wdio-sync/webdriverio-core.d.ts
@@ -194,6 +194,11 @@ declare namespace WebdriverIO {
         reverse?: boolean,
     }
 
+    type DataMatcher = {
+        name: string,
+        args: Array<string>
+    }
+
     type ReactSelectorOptions = {
         props?: object,
         state?: any[] | number | string | object | boolean
@@ -264,7 +269,7 @@ declare namespace WebdriverIO {
          * it from an element scope is that the driver will look within the children of that element.
          */
         $$(
-            selector: string | Function | object
+            selector: string | Function | DataMatcher
         ): ElementArray;
 
         /**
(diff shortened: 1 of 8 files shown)


mgrybyk:
What is android data matcher strategy?

christian-bromann:
> I believe it should only accept a HTMLElement object

What do you mean by that?

Kignuf:
> Thanks for the PR! How about we create an type interface for the data matcher and add that to it. An Object can be anything and the data matcher is actually well defined.

Sure, I can and will do that today :)

Kignuf:
Actually, after having a look at the code, I am not sure how I am supposed to implement this.
I tried as follow, but it fails to generate typings:
```javascript
* @param {String|Function|{name:string, args: Array<string>}} selector  selector or JS Function to fetch a certain element
```

I have searched for similar examples in the codebase but could not find any.
As I am not very familiar with types generation mechanism, could you indicate me if there is an easy solution for this problem ?

Thanks

christian-bromann:
@Kignuf you can define your custom interface in there, e.g.:
```js
* @param {String|Function|DataMatcher} selector  selector or JS Function to fetch a certain element
```
and define the interface in the type definition template, e.g. the same way we did for drag and drop options: https://github.com/webdriverio/webdriverio/blob/master/scripts/templates/webdriverio.tpl.d.ts#L201-L203

Kignuf:
Thanks @christian-bromann 

I've done the changes, and also added type definition for the global var ```driver``` which was missing. I hope that's correct. Let me know if you prefer to have a separate PR for this change.

Issue #5157, opened 2020-03-26
Unable to use in TS project: webdriver.d.ts references outdated @types/got package


adierkens:
**Environment (please complete the following information):**
 - **WebdriverIO version:** 6.0.2
 - **Mode:** Standalone
 - **Node.js version:** 12.16.1
 - **NPM version:** 6.13.4
- **TS version:** 3.8.3

The `webdriver.d.ts` types file [here](https://github.com/webdriverio/webdriverio/blob/master/packages/webdriver/webdriver.d.ts#L13) references the `got` package, using generics for `GotOptions` and `Response`: 

```ts
declare type HTTPRequestOptions = import('got').GotOptions<any>;
declare type HTTPResponse = import('got').Response<any>;
```

These generics come from the `@types/got` package, but the version of `got` requested by `webdriver` already ships with types; and doesn't use generics for either type. This causes a typescript compilation error:

```sh
@proof-ui/browser: $ tsc
@proof-ui/browser: ../../node_modules/webdriver/webdriver.d.ts(13,35): error TS2315: Type 'GotOptions' is not generic.
```

The current suite of tests copies the `@types/got` declaration into the testing folder, and the root `node_modules` contains an older (non-typed) version of `got`, so the types don't conflict, and the tests pass. The [copy](https://github.com/webdriverio/webdriverio/blob/master/tests/typings/copy.js#L54) utility also fails to copy `.d.ts` files nested in a sub-folder (as they are for the `got` package, and many natively typescript packages)

christian-bromann:
Fixed in #5160

PR #5160, opened 2020-03-27
Fix typings for got and its tests

christian-bromann:
## Proposed changes

As described in #5157 we used the wrong set of typings for our tests which caused TypeScript to error out as `got` introduced their own typings what we still checked against `@types/got`. This patch:

- makes the typing checks more robust by linking whole directories instead of copying single files
- fixes the got typings in `webdriver.d.ts`

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

n/a

### DIFF
--- a/tests/typings/sync/config.ts
+++ b/tests/typings/sync/config.ts
@@ -27,7 +27,7 @@ const conf: WebdriverIO.Config = {
     },
     transformResponse: (response, requestOptions) => {
         if (requestOptions.method === 'DELETE' && response.statusCode === 200) {
-            response.body.deleted = true
+            console.log(response.body)
         }
 
         return response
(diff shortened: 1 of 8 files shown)


tests/typings/sync/config.ts
```diff
@@ -27,7 +27,7 @@ const conf: WebdriverIO.Config = {
     },
     transformResponse: (response, requestOptions) => {
         if (requestOptions.method === 'DELETE' && response.statusCode === 200) {
-            response.body.deleted = true
+            console.log(response.body)
```
erwinheitzman: Is this supposed to be a console.log?

christian-bromann: yes, as a check that the response object has access to a body property.

PR #5180, opened 2020-03-31
webdriverio: Add Espresso ViewMatcher strategy

ccharnkij:
## Proposed changes

Per [appium/appium#13747](https://github.com/appium/appium/issues/13747), This is for Espresso's ViewMatcher strategy, similar to DataMatcher.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

I have to make both datamatcher and viewmatcher use 'directly' strategy because both of them have more or less the same object structure, and there is no way to tell from the object passing in if it's for datamatcher or viewmatcher. I have already done integration test using the feature.

### DIFF
--- a/packages/webdriverio/src/utils/findStrategy.js
+++ b/packages/webdriverio/src/utils/findStrategy.js
@@ -2,7 +2,7 @@ import { W3C_SELECTOR_STRATEGIES } from '../constants'
 import isPlainObject from 'lodash.isplainobject'
 
 const DEFAULT_STRATEGY = 'css selector'
-const DIRECT_SELECTOR_REGEXP = /^(id|css selector|xpath|link text|partial link text|name|tag name|class name|-android uiautomator|-android datamatcher|-ios uiautomation|-ios predicate string|-ios class chain|accessibility id):(.+)/
+const DIRECT_SELECTOR_REGEXP = /^(id|css selector|xpath|link text|partial link text|name|tag name|class name|-android uiautomator|-android datamatcher|-android viewmatcher|-ios uiautomation|-ios predicate string|-ios class chain|accessibility id):(.+)/
 const XPATH_SELECTORS_START = [
     '/', '(', '../', './', '*/'
 ]
@@ -24,9 +24,11 @@ const defineStrategy = function (selector) {
     // Condition with checking isPlainObject(selector) should be first because
     // in case of "selector" argument is a plain object then .match() will cause
     // an error like "selector.match is not a function"
-    // Use '-android datamatcher' strategy if selector is a plain object (Android only)
+    // Use '-android datamatcher' or '-android viewmatcher' strategy if selector is a plain object (Android only)
     if (isPlainObject(selector)) {
-        return '-android datamatcher'
+        if (JSON.stringify(selector).indexOf('test.espresso.matcher.ViewMatchers') < 0)
+            return '-android datamatcher'
+        return '-android viewmatcher'
     }
     // Check if user has specified locator strategy directly
     if (selector.match(DIRECT_SELECTOR_REGEXP)) {
@@ -128,6 +130,11 @@ export const findStrategy = function (selector, isW3C, isMobile) {
     }
     case '-android datamatcher': {
         using = '-android datamatcher'
+        value = JSON.stringify(value)
+        break
+    }
+    case '-android viewmatcher': {
+        using = '-android viewmatcher'
         value = JSON.stringify(value)
         break
     }
--- a/packages/webdriverio/webdriverio-core.d.ts
+++ b/packages/webdriverio/webdriverio-core.d.ts
@@ -384,9 +384,10 @@ declare namespace WebdriverIO {
         reverse?: boolean,
     }
 
-    type DataMatcher = {
+    type Matcher = {
         name: string,
-        args: Array<string>
+        args: Array<string | object>
+        class?: string
     }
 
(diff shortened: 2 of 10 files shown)


packages/webdriverio/src/utils/findStrategy.js
```diff
@@ -21,13 +20,6 @@ const XPATH_SELECTOR_REGEXP = [
 ]
 
 const defineStrategy = function (selector) {
-    // Condition with checking isPlainObject(selector) should be first because
-    // in case of "selector" argument is a plain object then .match() will cause
-    // an error like "selector.match is not a function"
-    // Use '-android datamatcher' strategy if selector is a plain object (Android only)
-    if (isPlainObject(selector)) {
```
christian-bromann: Removing this would be a breaking change.

ccharnkij:
@christian-bromann I did as suggested and checked for the ViewMatcher class instead. Let me know how this one goes.

scripts/templates/webdriverio.tpl.d.ts
```diff
@@ -383,6 +383,12 @@ declare namespace WebdriverIO {
         args: Array<string>
     }
 
+    type ViewMatcher = {
+        name: string,
+        args: Array<string | object>
+        class?: string
```
christian-bromann, line 389: Shouldn't be class a required property? Otherwise this would turn into a `DataMatcher` no?

packages/webdriverio/src/utils/findStrategy.js
```diff
@@ -24,9 +24,11 @@ const defineStrategy = function (selector) {
     // Condition with checking isPlainObject(selector) should be first because
     // in case of "selector" argument is a plain object then .match() will cause
     // an error like "selector.match is not a function"
-    // Use '-android datamatcher' strategy if selector is a plain object (Android only)
+    // Use '-android datamatcher' or '-android viewmatcher' strategy if selector is a plain object (Android only)
     if (isPlainObject(selector)) {
-        return '-android datamatcher'
+        if (JSON.stringify(selector).indexOf('test.espresso.matcher.ViewMatchers') < 0)
```
christian-bromann, line 29: How about:

```suggestion
        if (selector.class === 'test.espresso.matcher.ViewMatchers')
```

scripts/templates/webdriverio.tpl.d.ts
ccharnkij: Yes. This is the reason of my breaking change in the 1st commit, both DataMatcher and ViewMatcher objects can look identical. The class field is an optional property on both DataMatcher and ViewMatcher, and similarly, the args field can be an array of either String or Object. The difference is on the Espresso side, DataMatcher calls 'onData' and ViewMatcher calls 'onView'.

packages/webdriverio/src/utils/findStrategy.js
ccharnkij: The ViewMatchers class is not guaranteed to appear in the root object, it can appear in the nested object as well. For ex.
`
{
  name: 'allOf',
  args: [{
     name: 'withId',
     args: 'testId',
     class: 'androidx.test.espresso.matcher.ViewMatchers'
  },
  {
     name: 'withText',
     args: 'testText',
     class: 'androidx.test.espresso.matcher.ViewMatchers'
  }],
  class: 'org.hamcrest.Matcher'
}
`

In Java Espresso - onView(allOf(withId(R.id.my_view), withText("Hello!")));

christian-bromann: Is there a possibility to have e.g.:
```json
{
   "name":"allOf",
   "args":[
      {
         "name":"withId",
         "args":"testId",
         "class":"androidx.test.espresso.matcher.ViewMatchers"
      },
      {
         "name":"withText",
         "args":"testText",
         "class":"androidx.test.espresso.matcher.DataMatchers"
      }
   ],
   "class":"org.hamcrest.Matcher"
}
```
_Note:_ there is a data and view matcher.

packages/webdriverio/webdriverio-core.d.ts
```diff
@@ -389,6 +389,12 @@ declare namespace WebdriverIO {
         args: Array<string>
     }
 
+    type ViewMatcher = {
```
christian-bromann, line 392: Can there be a `class` property in data matcher selectors? If so it might make sense to merge both types.

scripts/templates/webdriverio.tpl.d.ts
christian-bromann: ah ok, so can we merge these types then and give them a more general name?

packages/webdriverio/src/utils/findStrategy.js
ccharnkij: Yes, this is also possible. Well...if what you mean is different classes in the args (I don't think that DataMatchers class exist). Whether or not it's going to find the element is another story.

packages/webdriverio/webdriverio-core.d.ts
ccharnkij: Yes, DataMatcher and ViewMatcher have the same object structure. I can combine the two in the just one Matcher type for TS. What do you think?

christian-bromann: Sounds good!

ccharnkij:
@christian-bromann How about this one?

Issue #5178, opened 2020-03-30
Click $(selector).click({ x, y }) is performed on wrong coordinates


katerina-stepanova:
**Environment:**
 - **WebdriverIO version:** 6.0.1
 - **Mode:** WDIO Testrunner, sync
 - **Node.js version:** v12.16.1
 - **NPM version:** 6.13.4
 - **Browser name and version:** Chrome 80
 - **Platform name and version:** MacOs 10.14.6 (other OSes not checked)
 - **Additional wdio packages used (if applicable):** n/a

**Config of WebdriverIO**
Nothing special, default config.

**Describe the bug**
Click $(selector).click({ x, y }) is performed x horizontal and y vertical pixels away from the top left corner of the page, instead of x horizontal and y vertical pixels away from the location of an element found by selector.

**To Reproduce**
browser.url('https://webdriver.io/')
let size = $('.pluginRowBlock div:first-child').getSize() //size is ~115x40
$('.pluginRowBlock div:first-child').click({x: 25, y: 10}) 

**Expected behavior**
Click should be performed x horizontal pixels and y vertical pixels away from the elements location.
In the example above, 'Get started' button should be clicked 25 pixels horizontally and 10 pixels vertically from its top left corner

christian-bromann:
@katerina-stepanova I assume you were using the `devtools` package as automation layer which indeed differed from the `webdriver` way of doing things. This is being fixed by #5193.

PR #5193, opened 2020-04-01
Recognise element origin in actions

christian-bromann:
## Proposed changes

The devtools and webdriver package had a difference in their `performAction` behavior where the devtools package wouldn't recognise an element origin. This patch fixes this.

Fixes: #5178

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

n/a


christian-bromann:
> Just a check, did you also test this with both instead of only with a UT?

The test is actually an e2e test. For all commands implemented in the DevTools package we run e2e tests to see if they are actually working.

PR #5202, opened 2020-04-02
Shorten screenshot base64 strings from logs

christian-bromann:
## Proposed changes

Currently we log the whole screenshot base64 string in the wdio logs which can be pretty verbose and doesn't make much sense. This patch tries to shorten them to make these logs more readable.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

n/a

### DIFF
--- a/packages/wdio-utils/src/index.js
+++ b/packages/wdio-utils/src/index.js
@@ -5,7 +5,7 @@ import { initialiseWorkerService, initialiseLauncherService } from './initialise
 import webdriverMonad from './monad'
 import {
     commandCallStructure, isValidParameter, getArgumentType, safeRequire,
-    isFunctionAsync
+    isFunctionAsync, transformCommandLogResult
 } from './utils'
 import {
     wrapCommand, runFnInFiberContext, executeHooksWithArgs,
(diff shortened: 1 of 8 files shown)


packages/wdio-utils/src/index.js
```diff
@@ -5,7 +5,7 @@ import { initialiseWorkerService, initialiseLauncherService } from './initialise
 import webdriverMonad from './monad'
 import {
     commandCallStructure, isValidParameter, getArgumentType, safeRequire,
-    isFunctionAsync
+    isFunctionAsync, transformCommandLogResult
```
wswebcreation, line 8: This might be nitpicking, but why do we have the imports on 2 lines and the imports on line 11 on multiple lines?

christian-bromann: I think GitHub made this confusing, if you expand the code the imports look similar:

```js
import {
    commandCallStructure, isValidParameter, getArgumentType, safeRequire,
	isFunctionAsync, transformCommandLogResult
}
```
vs.
```js
import {
    wrapCommand, runFnInFiberContext, executeHooksWithArgs,
    hasWdioSyncSupport, executeSync, executeAsync,
} from './shim'
```

In general I would be fine with including an Eslint rule to enforce a similar coding style.

PR #5219, opened 2020-04-06
Fix/better error logging

alfonso-presa:
## Proposed changes

Log status code on errors running commands (or starting the session). It makes easier to debug problems with not found (404), or access error (401) when configuration is wrong.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/packages/webdriver/src/request.js
+++ b/packages/webdriver/src/request.js
@@ -154,7 +154,7 @@ export default class WebDriverRequest extends EventEmitter {
          * retry, e.g. if sessionId is invalid
          */
         if (retryCount >= totalRetryCount || error.message.includes('invalid session id')) {
-            log.error('Request failed due to', error)
+            log.error(`Request failed with status ${response.statusCode} due to ${error}`)
             this.emit('response', { error })
             error.statusCode = response.statusCode
             error.statusMessage = response.statusMessage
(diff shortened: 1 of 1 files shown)


packages/webdriver/src/request.js
```diff
@@ -154,7 +154,7 @@ export default class WebDriverRequest extends EventEmitter {
          * retry, e.g. if sessionId is invalid
          */
         if (retryCount >= totalRetryCount || error.message.includes('invalid session id')) {
-            log.error('Request failed due to', error)
+            log.error('Request failed with status', response.statusCode, 'due to', error)
```
christian-bromann, line 157: How about:
```suggestion
            log.error(`Request failed with status ${response.statusCode} due to ${error.stack}`)
```

packages/webdriver/src/request.js
```diff
@@ -163,7 +163,7 @@ export default class WebDriverRequest extends EventEmitter {
 
         ++retryCount
         this.emit('retry', { error, retryCount })
-        log.warn('Request failed due to', error.message)
+        log.warn('Request failed with status', response.statusCode, 'due to', error.message)
```
christian-bromann, line 166: same:
```suggestion
        log.warn(`Request failed with status ${response.statusCode} due to ${error.message}`)
```

packages/webdriver/src/request.js
alfonso-presa: Done :-)

packages/webdriver/src/request.js
alfonso-presa: Done too :-)

PR #5237, opened 2020-04-09
Ensure that default connection is not overwritten

christian-bromann:
## Proposed changes

When running a setup where no connection details are defined (no hostname, port etc.) but it finds a driver server running and therefor defaults to WebDriver on `localhost:4444` due to default values of the WebDriver package, the connection values were overwritten to `undefined` again. This patch fixes that.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

Note that my change in the smoke test actually simulates the same condition and therefor would fail without this patch.


erwinheitzman:
Could have add some test(s) for this?

christian-bromann:
@erwinheitzman I tweaked the smoke tests (tests/helpers/config.js) which would fail without this patch.

PR #5259, opened 2020-04-13
Add any non-passing Cucumber test result to failure count

esaari:
## Proposed changes

Addresses https://github.com/webdriverio/webdriverio/issues/5256

[//]: # (Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue.)
The Browserstack API does not always report test status as either 'passed' or 'failed'. Timeout failures may be reported as well: 
`curl -u "xxxx:yyyy" https://api.browserstack.com/automate/sessions/<my_session_id>.json {"automation_session":{"name":"","duration":375,"os":"Windows","os_version":"10","browser_version":"80.0","browser":"chrome","device":null,"status":"timeout", ... }`

This was causing our test execution numbers to fluctuate night by night. 

Also bumping the Browserstack local dep.

## Types of changes
For the Cucumber-js specific `afterScenario()` hook, modify to be in line with the Mocha/Jasmine bloc and add any test not explicitly marked as `passed` to the failure count.
[//]: # (What types of changes does your code introduce to WebdriverIO?)
[//]: # (_Put an `x` in the boxes that apply_)

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## Checklist

[//]: # (_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._)

- [x] I have read the [CONTRIBUTING](https://github.com/webdriverio/webdriverio/blob/master/CONTRIBUTING.md) doc
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)
- [ ] I have added proper type definitions for new commands (if appropriate)

## Further comments

[//]: # (If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...)

Issue #5256, opened 2020-04-13
Browserstack Plugin not counting certain unsuccessful tests as failures


esaari:
[//]: # NOTE: This repository only maintains packages that are listed in the [Readme](https://github.com/webdriverio/webdriverio/blob/master/README.md#packages). Please make sure that your issue is directly caused by one of these packages and if not file an issue in the correct 3rd party package repository.

**Environment (please complete the following information):**
 - **WebdriverIO version:** [6.0.15]
 - **Mode:** [WDIO Testrunner]
 - **If WDIO Testrunner, running sync/async:** [e.g. sync]
 - **Node.js version:** [12.16.1]
 - **NPM version:** [6.13.4]
 - **Browser name and version:** [Chrome 80]
 - **Platform name and version:** [OS X Catalina]
 - **Additional wdio packages used (if applicable):** [e.g. @wdio/spec reporter, @wdio/selenium-standalone service]

**Config of WebdriverIO**
An example of how you configured WebdriverIO config

**Describe the bug**
Browserstack may occasionally return a `timeout` test status in addition to `passed` and `failed`. For Cucumber JS, these types of test failures haven't been recognized as failures in the `afterScenario()` hook. 

**To Reproduce**
Force any test to somehow timeout or not establish connection with Browserstack - here's API output from an example test.  I'll have a PR up shortly.

`curl -u "xxxx:yyyy" https://api.browserstack.com/automate/sessions/<my_session_id>.json
{"automation_session":{"name":"","duration":375,"os":"Windows","os_version":"10","browser_version":"80.0","browser":"chrome","device":null,"status":"timeout", ... }`

esaari:
https://github.com/webdriverio/webdriverio/pull/5259

christian-bromann:
PR with a fix got merged. Closing.

PR #5271, opened 2020-04-15
Rework Browserstack service, improve Cucumber integration

nextlevelbeard:
#### Reworks BrowserStack service way of handling failures
- No longer using a failure counter:
    - directly uses `after` hooks' `result` code to measure failure as it is more accurate
    - exception being `onReload` which has no choice but to update session `status` by evaluating number of failure reasons

#### Mocha / Jasmine
- Initial session name is now the top-level Suite' name (if it exists)
- If there is no top-level Suite, session name is the default `sessionId`
- Keeps old behaviour: at the end of the session, the name is updated to the last test's name

#### Improves Cucumber integration:
- Better reporting of session status on failure by redefining failure
- Includes the reasons for failure
- Makes the status report respect the `strict` option provided in `cucumberOpts`
- In the beginning of the session the name is updated to be the spec / Feature name
- Adds `preferScenarioName` option for service which updates the session name at the end end to the scenario name if only a single scenario ran on that session. Useful for parallel runs with [wdio-cucumber-parallel-execution](https://github.com/SimitTomar/wdio-cucumber-parallel-execution).
- Addresses some concerns of #5256. Based on @esaari's #5259.
---
#### For 2 Sessions, same Cucumber feature, different Scenarios:

`preferScenarioName = false`
image

`preferScenarioName = true`, on session start has feature name
2 sessions 1 scenario each feature on start

`preferScenarioName = true`, on session end has scenario name
2 sessions 1 scenario each feature on end

---

Single reasons for failure (i.e. 1 test failed)
pending scenario reason
custom error reason
fail reason timeout
Reason when multiple tests fail
multiple errors reason

## Types of changes

- [X] Documentation update

## Checklist


packages/wdio-browserstack-service/src/service.js
```diff
@@ -6,10 +6,12 @@ import { BROWSER_DESCRIPTION } from './constants'
 const log = logger('@wdio/browserstack-service')
 
 export default class BrowserstackService {
-    constructor (options, caps, config) {
+    constructor ({ preferScenarioName = false }, caps, config) {
         this.config = config
-        this.failures = 0
         this.sessionBaseUrl = 'https://api.browserstack.com/automate/sessions'
+        this.preferScenarioName = preferScenarioName
```
christian-bromann, line 12: This seems to be a boolean

```suggestion
        this.preferScenarioName = Boolean(preferScenarioName)
```

packages/wdio-browserstack-service/src/service.js
```diff
@@ -6,10 +6,12 @@ import { BROWSER_DESCRIPTION } from './constants'
...
+        this.strict = !!(config.cucumberOpts && config.cucumberOpts.strict)
```
christian-bromann, line 13:
```suggestion
        this.strict = Boolean(config.cucumberOpts && config.cucumberOpts.strict)
```

packages/wdio-browserstack-service/src/service.js
```diff
@@ -36,16 +38,18 @@ export default class BrowserstackService {
             this.sessionBaseUrl = 'https://api-cloud.browserstack.com/app-automate/sessions'
         }
 
+        this.scenariosThatRan = []
+
         return this._printSessionURL()
     }
 
-    afterSuite(suite) {
```
christian-bromann, line 42: We can't remove `afterSuite` as it is needed for other non Cucumber frameworks

packages/wdio-browserstack-service/src/service.js
```diff
@@ -57,32 +61,50 @@ export default class BrowserstackService {
...
-            this.failReason = (test.error && test.error.message ? test.error.message : 'Unknown Error')
+        if (!passed) {
+            this.failReasons.push(error && error.message || 'Unknown Error')
         }
     }
 
-    after() {
-        return this._update(this.sessionId, this._getBody())
+    after(result) {
+        if(this.preferScenarioName && this.scenariosThatRan.length === 1){
```
christian-bromann, line 70: can we add a comment block documenting what this is doing?

packages/wdio-browserstack-service/README.md
```diff
@@ -54,6 +55,12 @@ Set this to true to enable routing connections from Browserstack cloud through y
 Type: `Boolean`<br>
 Default: `false`
 
+### preferScenarioName
+Cucumber only. Set this to true to enable updating the session name to the Scenario name if only a single Scenario was ran. Useful when running in parallel with [wdio-cucumber-parallel-execution](https://github.com/SimitTomar/wdio-cucumber-parallel-execution)
```
christian-bromann, line 59:
```suggestion
Cucumber only. Set this to true to enable updating the session name to the Scenario name if only a single Scenario was ran. Useful when running in parallel with [wdio-cucumber-parallel-execution](https://github.com/SimitTomar/wdio-cucumber-parallel-execution).
```

packages/wdio-browserstack-service/src/service.js
nextlevelbeard: Is it still needed though? `afterSuite` was simply updating the `failures` counter to be used in the `after` hook: this PR now uses `after` hook's `result` for that

christian-bromann: You are right

esaari:
@nextlevelbeard I can provide you an example curl request that will return a session `status:` of `timeout`. I tried DM'ing you on gitter, but here is a partial response object (omitted most of it due to PII) with a `status` of `timeout`: 

`{"automation_session":{"name":"","duration":375,"os":"Windows","os_version":"10","browser_version":"80.0","browser":"chrome","device":null,"status":"timeout","hashed_id":"f360634511258f578d9e5d9cd9ad1563db887d3b","reason":"TIMEOUT","build_name": ... } `

I realize their own documentation says `passed` or `failed` but they're documentation isn't always accurate. I've brought this to their attention.

goatsy:
@nextlevelbeard Is this issue here also related with the problem that all API-Status marked with 

Status	
 Failed
(Marked via REST API : Unknown Error)

image

packages/wdio-browserstack-service/src/service.js
nextlevelbeard: How's this?

nextlevelbeard:
> Can we make these modifications also in the Sauce, Testingbot and Crossbrowsertesting service?

I can but it should be out of scope of this PR, don't you agree?

nextlevelbeard:
> @nextlevelbeard Is this issue here also related with the problem that all API-Status marked with
> 
> Status
> Failed
> (Marked via REST API : Unknown Error)
> 
> image

Not quite sure but this PR is redefining failure, shouldn't report any false failures like that, no.

nextlevelbeard:
@christian-bromann tests run locally just fine, do you have an idea why CI is failing here?

christian-bromann:
> tests run locally just fine, do you have an idea why CI is failing here?

@nextlevelbeard the error is caused by this line: https://github.com/webdriverio/webdriverio/pull/5271/files#diff-540108e09662df242d6404527d1597ddR96

nextlevelbeard:
> @nextlevelbeard the error is caused by this line: https://github.com/webdriverio/webdriverio/pull/5271/files#diff-540108e09662df242d6404527d1597ddR96

Should be taken care of.

packages/wdio-browserstack-service/src/service.js
```diff
@@ -57,32 +67,51 @@ export default class BrowserstackService {
             `${test.parent} - ${test.title}`
         )
 
-        if (!results.passed) {
-            this.failures++
-            this.failReason = (test.error && test.error.message ? test.error.message : 'Unknown Error')
```
nextlevelbeard, line 62: @christian-bromann can you confirm this change for `afterTest`?
`test.error` ->`error` from las parameter object

packages/wdio-browserstack-service/src/service.js
```diff
@@ -36,16 +44,18 @@ export default class BrowserstackService {
...
         return this._printSessionURL()
     }
 
-    afterSuite(suite) {
-        if (Object.prototype.hasOwnProperty.call(suite, 'error')) {
-            this.failures++
-        }
+    beforeFeature(uri, feature) {
+        const oldTitle = this.fullTitle
+        this.fullTitle = [this.fullTitle, feature.document.feature.name].filter(Boolean).join(', ')
```
erwinheitzman, line 54: `.filter(Boolean)` why do this?

erwinheitzman, line 54: using shadow of  `this.fullTitle` make it hard to read, replace with oldTitle where possible

packages/wdio-browserstack-service/src/service.js
```diff
@@ -6,10 +6,18 @@ import { BROWSER_DESCRIPTION } from './constants'
 const log = logger('@wdio/browserstack-service')
 
 export default class BrowserstackService {
-    constructor (options, caps, config) {
+    constructor ({ preferScenarioName = false }, caps, config) {
```
erwinheitzman, line 9: keep the names the way they are so that when other people who work on this in the future will know what the object represents. This does for any destructering in this PR.

packages/wdio-browserstack-service/src/service.js
```diff
@@ -57,32 +67,51 @@ export default class BrowserstackService {
...
+
+            this.failReasons.push(exception)
         }
     }
 
     async onReload(oldSessionId, newSessionId) {
         this.sessionId = newSessionId
-        await this._update(oldSessionId, this._getBody())
-        this.failures = 0
+        await this._update(oldSessionId, { name: this.fullTitle })
```
erwinheitzman, line 111: how is the status updated in this case?

packages/wdio-browserstack-service/src/service.js
```diff
@@ -57,32 +67,51 @@ export default class BrowserstackService {
...
+    after(result) {
+        // For Cucumber: Checks scenarios that ran (i.e. not skipped) on the session
+        // Only 1 Scenario ran and option enabled => Redefine session name to Scenario's name
+        if(this.preferScenarioName && this.scenariosThatRan.length === 1){
+            this.fullTitle = this.scenariosThatRan.pop()
+        }
+
+        const hasReasons = Boolean(this.failReasons.filter(Boolean).length)
+
+        return this._update(this.sessionId, {
```
erwinheitzman, line 84: are the status and reason required here as it's missing in the update of the onReload?

packages/wdio-browserstack-service/src/service.js
christian-bromann: 👍

packages/wdio-browserstack-service/src/service.js
christian-bromann: yes, this is the v6 way of doing things. 👍

packages/wdio-browserstack-service/src/service.js
nextlevelbeard: @erwinheitzman  @christian-bromann
I took a look at the Sauce service:
https://github.com/webdriverio/webdriverio/blob/e30d665b06f9a1b16dcf721471fd2488339d67ea/packages/wdio-sauce-service/src/service.js#L11

apparently the whole raison d'être for the failure counter (which this PR removes) is so that  `onReload` hook can report on the ending session

We don't have any session results exposed on `onReload`, we would ideally use what is later given in `after` hook. Would it make sense to expose some `result`s on the `onReload` hook along with `oldSessionId` and `newSessionId`? Something like 
```js
    onReload: function(oldSessionId, result, newSessionId) {
    ...
    },
```

Illustrating the problem:
```js
it("Test that runs first", () => throw new Error('I fail!')) // sessionId 1 will fail
it("Test that runs second", () => {
    console.log("I am running on session 1")
    browser.reloadSession() // We should update session 1 with status "failed"
    // but there's no session 1 results exposed in `onReload`
    // every service has to count failures so far (with varying degrees of success)
    return true // session 2 passes and `after` will update the status accordingly
})
```

packages/wdio-browserstack-service/src/service.js
nextlevelbeard: Was under the impression we could run multiple features (spec files) on one session and this would correctly set the session name accordingly.


`[undefined, 'Feature 1'].filter(Boolean).join(', ')`
`['Feature 1', 'Feature 2'].filter(Boolean).join(', ')`
`['Feature 1, Feature 2', 'Feature 3'].filter(Boolean).join(', ')`

erwinheitzman: So that means this.fullTitle can be undefined? If so, simply don't add it as it would make it explicit what you are trying to prevent from happening instead of patching it afterwards.

This would also mean that this will update with undefined:
```js
this._update(this.sessionId, { name: this.fullTitle }) // name can be undefined
```

nextlevelbeard:
https://github.com/webdriverio/webdriverio/pull/5271#discussion_r416523913

^ Making this more visible.
Would appreciate some guidance from the devs about what to do here for `onReload`.

packages/wdio-browserstack-service/src/service.js
nextlevelbeard: 1 session equals 1 Feature so I've removed this

packages/wdio-browserstack-service/src/service.js
nextlevelbeard: they are as are in `onReload` which was updated

nextlevelbeard:
Should be good to go, OP updated.

nextlevelbeard:
Addressed one last thing about restructuring parameters.

packages/wdio-browserstack-service/tests/service.test.js
```diff
@@ -200,100 +207,479 @@ describe('before', () => {
...
+                service.strict = true
+
+                await service.beforeFeature({}, { document: { feature: { name: 'Feature1' } } })
+
+                await service.afterScenario({}, {},
+                    { name: 'Can do something skipped 1' }, { status: 'skipped' })
+                await service.afterScenario({}, {},
+                    { name: 'Can do something skipped 2' }, { status: 'skipped' })
+                await service.afterScenario({}, {},
+                    { name: 'Can do something skipped 3' }, { status: 'passed'  })
```
nextlevelbeard, line 503:
```suggestion
                    { name: 'Can do something skipped 3' }, { status: 'skipped'  })
```

Issue #3648, opened 2019-02-28
Add "name" and "reason" to update request to Browserstack API


christian-bromann:
**Is your feature request related to a problem? Please describe.**
It seems that Browserstack now supports modifying the name of the session as well as the reason for the failure in [their API](https://www.browserstack.com/automate/rest-api).

**Describe the solution you'd like**
Let's enhance the `@wdio/browserstack-service` to update the session name as well as give the reason for a test failure.

OriTheMan:
Hi,
I can give a try to handle it (its my first time here).
1. there is way to get the error in the service? (for "reason" field)
2. who decide what is the session name? the user or automatically

christian-bromann:
> 1. there is way to get the error in the service? (for "reason" field)

The afterTest hook should have information about test name and error, please check the sauce service where we already store the test name

PR #5293, opened 2020-04-19
wdio-browserstack-service: do not overwrite name by test title

bcaudan:
## Proposed changes

Session name is overridden at the end of the execution by the last executed spec full title.
Keep this behavior only if session name is not defined.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/wdio-browserstack-service/src/service.js
+++ b/packages/wdio-browserstack-service/src/service.js
@@ -7,6 +7,7 @@ const log = logger('@wdio/browserstack-service')
 
 export default class BrowserstackService {
     constructor (options, caps, config) {
+        this.caps = caps
         this.config = config
         this.failures = 0
         this.sessionBaseUrl = 'https://api.browserstack.com/automate/sessions'
@@ -97,7 +98,7 @@ export default class BrowserstackService {
     _getBody() {
         return {
             status: this.failures === 0 ? 'passed' : 'failed',
-            name: this.fullTitle,
+            name: this.caps.name !== undefined ? this.caps.name : this.fullTitle,
             reason: this.failReason
         }
     }


packages/wdio-browserstack-service/src/service.js
```diff
@@ -97,7 +98,7 @@ export default class BrowserstackService {
     _getBody() {
         return {
             status: this.failures === 0 ? 'passed' : 'failed',
-            name: this.fullTitle,
+            name: this.caps.name !== undefined ? this.caps.name : this.fullTitle,
```
christian-bromann, line 101: `name` is not a valid capability so why would someone define the test name there?

bcaudan: `name` is documented in the browserstack capabilities generator: https://www.browserstack.com/automate/capabilities

esaari: @bcaudan This would also need to be made compatible with Browserstack w3c capabilities, which has slightly different key names and enclosed these in a `'bstack:options'` object: 
https://www.browserstack.com/automate/capabilities?tag=selenium-4
```
var capabilities = {
'bstack:options' : {
"osVersion" : "11",
"deviceName" : "iPhone 8 Plus",
"realMobile" : "true",
"projectName" : "Project",
"buildName" : "Build",
"sessionName" : "Name",
"local" : "false",
"userName" : "ericsaari1",
"accessKey" : "BjCyFyXBycePQTycFMxz",
},
"browserName" : "iPhone",
```

bcaudan: Oh, right! By looking a bit deeper on why the session name is set to the test full title (https://github.com/webdriverio/webdriverio/issues/3648), it seems that the intention is to update the session name with the failing test when the session is failed. However, it seems to update it with the last test when the session is successful. Would it be a better strategy to not update the session when there is no failure?

christian-bromann: @bcaudan can you elaborate what the current issue with the current reporting is? What is missing?

bcaudan: Yes, I have continue to dig and in fact the issue is related to jasmine compatibility.
The service look at [test.passed](https://github.com/webdriverio/webdriverio/blob/master/packages/wdio-browserstack-service/src/service.js#L60) to determine if the test is successful or not.
However, this property does not exist on jasmine which make the service update the test as failed every time.
I can close this PR and open a new one that also check test.failedExpectations in order to be more compatible with jasmine.
What do you think?

christian-bromann: What does `test.passed` has to do with changing the name of the test?

bcaudan: Sorry, I was not clear enough.
From my understanding, the goal of the current implementation is to update the session name with the failed test full title when there is a test failure during the session, cf #3648.
However, it seems to me that there is two issues:
- if there is no failure, the session name is updated with the last test full title
- if jasmine is used, `test.passed` is not defined so the session is always updated as failed

So there is an issue with always changing the name and another one with `test.passed`.

do you concur?

bcaudan: Opened #5323 and #5324 instead

PR #3691, opened 2019-03-06
 browserstack-service: fixes #3648 modifying name and reason

OriTheMan:
## Proposed changes
 Browserstack supports modifying the name of the session as well as the reason for the failure as described in #3648 

[//]: # (Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue.)
I added new parameters for Browserstack update request.
it fixes #3648 and contains small changes.

## Types of changes

[//]: # (What types of changes does your code introduce to WebdriverIO?)
[//]: # (_Put an `x` in the boxes that apply_)

- [x] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

[//]: # (_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._)

- [x] I have read the [CONTRIBUTING](https://github.com/webdriverio/webdriverio/blob/master/CONTRIBUTING.md) doc
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

## Further comments

[//]: # (If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...)

### Reviewers: @webdriverio/technical-committee

PR #5323, opened 2020-04-25
wdio-browserstack-service: do not update session name on success

bcaudan:
## Proposed changes

session name is supposed to be updated with last failing test (#3691).
However if the session is successful, the session name is updated with the last test name.

Changes to not update the session name is the session is successful.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/wdio-browserstack-service/src/service.js
+++ b/packages/wdio-browserstack-service/src/service.js
@@ -95,8 +95,11 @@ export default class BrowserstackService {
     }
 
     _getBody() {
+        if (this.failures === 0) {
+            return { status: 'passed' }
+        }
         return {
-            status: this.failures === 0 ? 'passed' : 'failed',
+            status: 'failed',
             name: this.fullTitle,
             reason: this.failReason
         }
--- a/packages/wdio-browserstack-service/tests/service.test.js
+++ b/packages/wdio-browserstack-service/tests/service.test.js
@@ -292,8 +292,7 @@ describe('_getBody', () => {
 
     it('should return "passed" if no errors', () => {
         service.failures = 0
-        service.fullTitle = 'foo - bar'
 
-        expect(service._getBody()).toEqual({ status: 'passed', name: 'foo - bar', reason: undefined })
+        expect(service._getBody()).toEqual({ status: 'passed' })
     })
 })


packages/wdio-browserstack-service/src/service.js
```diff
@@ -95,8 +95,11 @@ export default class BrowserstackService {
     }
 
     _getBody() {
+        if (this.failures === 0) {
+            return { status: 'passed' }
```
christian-bromann, line 99: Shouldn't the body contain the test name also in success case?

bcaudan: Name has been introduced in the body by #3691 and is not required by [browserstack update API](https://www.browserstack.com/automate/rest-api#rest-api-sessions).
For success case, keeping the name will update the session name with the last test name which is surprising to me.
Is it used for another purpose?

christian-bromann: @bcaudan would you mind changing this behavior? It should update with the initial suite name, similar as in the sauce service where this is defined in `beforeTest`

bcaudan: I had a look but from a couple of tests with jasmine, my understanding is that an implementation [similar to sauce](https://github.com/webdriverio/webdriverio/blob/master/packages/wdio-sauce-service/src/service.js#L47) would not produced the expected behavior:

- `suiteTitle` is never initialized to 'Jasmine__TopLevel__Suite', so  `suiteTitle` is never set with the suite name
- sauce implementation return the test name and not the suite name
- by executing this code on beforeTest, the session name should still be updated with the last test name and not the initial suite
- the failure reason that is used to update the session could then be related to another test that the one that is used to update the session name, it can be confusing. 

What am I missing?

christian-bromann: The current behavior is that the session gets the name from the first test suite to identify the session in the vendor dashboard.

> I am also not sure to understand the need to update the session name with the initial suite

What other the name the session should have in your opinion?

> since a session can have multiple suites.

The idea is to set the test name to the name of the first suite, e.g.

```js
describe("foo", () => {
  describe("bar", () => {
    // ...
  })
})
```

for such test the session name should be `foo`.

bcaudan: > What other the name the session should have in your opinion?

On my current project, we have used crossbrowsertesting and we are now using browserstack. For both of them, user can define the session name through the service configuration. I don't know if it is possible on other services like saucelabs though.

We used it to display info like:

- local [username] - [browser name]
- ci [job number] / [git commit] - [browser name]

For our usage, it is more convenient to see from the vendor dashboard what is responsible for the session execution rather than which suite or test has been executed. 

> since a session can have multiple suites.

I had this case in mind:

```js
describe("foo", () => {
  // ...
})

describe("bar", () => {
  // ...
})
``` 

where displaying `foo` is not that relevant.

christian-bromann: I think these are legit reasons and it would make sense to add some sort of configuration option that allows a user to set the test name based on some information, e.g.

```
services: [
  ["browserstack, {
    testName: (test, capabilities) => {
      return `${process.env.COMMIT_SHA} - ${capabilities.browserName}`
    }
  }]
],
```

What do you think?

bcaudan: That would work, however without behaviors of updating the session name with the failed test name or the initial suite name this can ban handled out of the box with either browserstack capability options:

- `name: xxx` (legacy)
- `'bstack:options': { sessionName: xxx }` (w3c)

what about relying on those by default and add a configuration option to update the session name with the failed test name or the initial suite name?

IMO, it would be less surprising if the configuration generated from [browserstack capabilities](https://www.browserstack.com/automate/capabilities?tag=selenium-4) can be copy/pasted to wdio configuration without any trouble.

christian-bromann: > relying on those by default

This would be a change in the current behavior and might confuse people who like to have the suite name as session name

> add a configuration option to update the session name with the failed test name or the initial suite name

I am not sure why it should depend whether or not your test fails.

bcaudan: > This would be a change in the current behavior and might confuse people who like to have the suite name as session name

Yes, I realize this could be a breaking change.
I let you decide which approach seems the more appropriate for wdio here.

> I am not sure why it should depend whether or not your test fails.

From my understanding, it was the intent of #3691.
If we go with: 

- initial suite for session name
- last test error message for session failure reason

we can have:
```js
describe("foo", () => {
  // ...
})

describe("bar", () => {
  it("qux", () => {
    expect(true).toBe(false)
  })
})
```

updating the session with:

- name: 'foo'
- reason: 'Expected true to be false'
- status: 'failed'

is it the expected behavior?

christian-bromann: > is it the expected behavior?

Yes. The the idea behind it is that you don't have 2 root level suites and rather do:

```js
describe("my feature", () => {
  describe("does XYZ", () => {
    it("fills out form A", () => {
      ...
    })
    ...
  })
})
```

bcaudan: OK, let me know if you want me to implement one of the mentioned strategy.
Note that it seems that [the current sauce implementation does not produced the expected behavior](https://github.com/webdriverio/webdriverio/pull/5323#discussion_r415105220) on my side.

christian-bromann:
@bcaudan sorry for not responding in a while. I would like to propose that we should assume that there are never two root level describe blocks. If there are we take the name of the __last__ one. Does this make sense?

bcaudan:
> @bcaudan sorry for not responding in a while. I would like to propose that we should assume that there are never two root level describe blocks. If there are we take the name of the **last** one. Does this make sense?

That makes sense to me.
I won't have time to quickly update this PR accordingly, but I could have a look at it in the following weeks.

SlaveWilson:
Can we just do the below code inside service.js? So we can assume the root describe block title is the session name? Otherwise, it quite confusing to implement the test case for the session name at the moment. 

``` js
afterSuite(suite) {
    this.fullTitle = suite.title;

    this._update(this.sessionId, {
      name: this.fullTitle
    });
  }
```

christian-bromann:
@SlaveWilson sounds like a reasonable solution to me.

christian-bromann:
@bcaudan are you still interested moving this forward?

christian-bromann:
Closing due to inactivity. @bcaudan please let me know if I can do anything to help land this.

PR #5360, opened 2020-05-05
Move docs deployment to GitHub workflow

christian-bromann:
## Proposed changes

Currently we run a step in our TravisCI pipeline to deploy the docs when a tag is pushed. Even though this step is skipped when a commit push happens the job is still being scheduled and takes at least 1:30min. By moving this step to GitHub we can save some time in our pipeline and speed up the checks.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

n/a


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,23 @@
...
+name: Docs Deploy
+
+on:
+  create:
+    tags:
+      - v*
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
```
mgrybyk, line 12: Do you think it's worst specifying exact os version, ex [ubuntu-18.04](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on)

christian-bromann: I don't see a reason why we wouldn't want to always run on latest

mgrybyk: Unless some new versions of ubuntu can break something for us we should not worry. I still prefer using latest LTS however there should be no impact for us, I believe. Just wanted to make sure, thanks

PR #5417, opened 2020-05-19
Update SauceLabs package and remove sauce-connect-launcher

christian-bromann:
## Proposed changes

With https://github.com/saucelabs/node-saucelabs/releases/tag/v4.1.0 we have embedded Sauce Connect support into the node saucelabs package. With that we can remove `sauce-connect-launcher` as dependency.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

I removed the `scRealy` option as it will be deprecated with the next Sauce Connect Proxy release.

### DIFF
--- a/packages/wdio-sauce-service/src/launcher.js
+++ b/packages/wdio-sauce-service/src/launcher.js
@@ -1,19 +1,19 @@
-import { promisify } from 'util'
 import { performance, PerformanceObserver } from 'perf_hooks'
 
 import logger from '@wdio/logger'
-import SauceConnectLauncher from 'sauce-connect-launcher-update'
+import SauceLabs from 'saucelabs'
 
-function modifyObject (root, propertiesObject) {
-    for (const [k, v] of Object.entries(propertiesObject)) {
-        root[k] = v
-    }
-}
+const SC_RELAY_DEPCRECATION_WARNING = [
+    'The "scRelay" option is depcrecated and will be removed',
+    'with the upcoming versions of @wdio/sauce-service. Please',
+    'remove the option as tests should work identically without it.'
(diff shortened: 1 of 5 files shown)


packages/wdio-sauce-service/src/launcher.js
```diff
@@ -1,19 +1,19 @@
...
 import logger from '@wdio/logger'
-import SauceConnectLauncher from 'sauce-connect-launcher-update'
+import SauceLabs from 'saucelabs'
 
-function modifyObject (root, propertiesObject) {
-    for (const [k, v] of Object.entries(propertiesObject)) {
-        root[k] = v
-    }
-}
+const SC_RELAY_DEPCRECATION_WARNING = [
```
erwinheitzman, line 6: Add newlines at the end?

packages/wdio-sauce-service/src/launcher.js
```diff
@@ -103,6 +103,6 @@ export default class SauceLauncher {
             return
         }
 
-        return new Promise(resolve => this.sauceConnectProcess.close(resolve))
+        return this.sauceConnectProcess.close()
```
erwinheitzman, line 106: The resolve was previously passed to the close method, this method no longer expects this value?

christian-bromann: no, the close method returns a promise so the hook will wait until the tunnel has closed.

packages/wdio-sauce-service/src/launcher.js
christian-bromann: this will be logged into the log file or stdout and doesn't need a new line.

PR #5438, opened 2020-05-27
Setup Babel and TypeScript if config files are found

christian-bromann:
## Proposed changes

With this patch the CLI config wizard detects if TypeScript or Babel is setup in the project (by checking existence of `babel.config.js` or `tsconfig.json` in the working dir) and adds a question if these compilers should be integrated into the config file.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

n/a

### DIFF
--- a/packages/wdio-cli/src/templates/wdio.conf.tpl.ejs
+++ b/packages/wdio-cli/src/templates/wdio.conf.tpl.ejs
@@ -170,16 +170,15 @@ exports.config = {
     // Whether or not retried specfiles should be retried immediately or deferred to the end of the queue
     // specFileRetriesDeferred: false,
     //
-    <%- include('reporters', { reporters: answers.reporters }) %> <%
-    if(answers.framework === 'mocha') { %>
+    <%- include('reporters', { reporters: answers.reporters }) %>
+    <% if(answers.framework === 'mocha') { %>
     //
     // Options to be passed to Mocha.
     // See the full list at http://mochajs.org/
     mochaOpts: {
         <% if(answers.isUsingTypeScript) {
         %>// TypeScript setup
-        require: 'ts-node/register',
-        compilers: ['tsconfig-paths/register'],
+        require: ['ts-node/register'],
         <% } else if(answers.isUsingBabel) {
         %>// Babel setup
         require: ['@babel/register'],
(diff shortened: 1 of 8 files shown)


mgrybyk:
Looks really good!

Here are couple of ideas:
- Can we use typed configurations for both Babel and TypeScript? Users often want to use `import` keyword in their config files.
- Can we automatically generate tsconfig/babel.config if there are no such files and user wants to use babel/ts?
- Can we automatically add types to tsconfig?

I guess that most of them may be implemented later on.

mgrybyk:
Have very minor styling question. I see that indentation is missing between reporters and cucumberOpts properties in config. Not sure if it's caused by the PR changes or not

image

packages/wdio-cli/src/templates/wdio.conf.tpl.ejs
```diff
@@ -184,15 +176,30 @@ exports.config = {
     // Options to be passed to Mocha.
     // See the full list at http://mochajs.org/
     mochaOpts: {
-        ui: 'bdd',
+        <% if(answers.isUsingTypeScript) {
+        %>// TypeScript setup
+        require: 'ts-node/register',
+        compilers: ['tsconfig-paths/register'],
```
mgrybyk, line 182: `compilers` is deprecated, let's use `require` instead.
`require` accepts an array (not sure if string works at all)

mgrybyk:
@christian-bromann will you remove tsconfig-paths for every test framework?

christian-bromann:
@mgrybyk I removed the `tsconfig-paths/register` integration. Regarding your comments above: I think these are all great ideas and some of them came to my mind too but I would rather put them into separate PRs.

Anything left to do here?

christian-bromann:
I created a ticket for one of your comments, for the others:

> * Can we automatically generate tsconfig/babel.config if there are no such files and user wants to use babel/ts?

I feel like this would be a great idea for an init command where a new project is being generated similar to `vue create my-project`. But I assume most of the user add WebdriverIO to an existing project.

> * Can we automatically add types to tsconfig?

We can but we run into a lot problems doing this. If there is something that makes this easy I am up for it but challenges I see is: keeping the format of the file, adding types to existing types, how to ensure to not surprise the user from this modification.

PR #5443, opened 2020-05-28
Couldn't find plugin "" runner

christian-bromann:
With #5437 we introduced this regression bug.

```
npm init -y
npm i @wdio/cli
npx wdio config (mash enter)
// add a test file under test/specs
npx wdio
```


christian-bromann:
Will fast track this as it is a regression bug and user currently experience it.

PR #5477, opened 2020-06-09
Implement Network Primitives

christian-bromann:
Implementation of the proposal for network primitives.


mgrybyk:
We can use `network` commands only with `devtools` automation protocol?

packages/webdriverio/src/utils/getNetworkObject.js
```diff
@@ -0,0 +1,111 @@
...
+import logger from '@wdio/logger'
+import { webdriverMonad, wrapCommand, commandCallStructure } from '@wdio/utils'
+
+const log = logger('webdriverio')
+
+import { FF_REMOTE_DEBUG_ARG } from '../constants'
+import { getBrowserObject, getPrototype as getWDIOPrototype } from '../utils'
+
+/**
+ * transforms and findElement response into a WDIO element
```
mgrybyk, line 12: Is the description valid?

christian-bromann:
> We can use `network` commands only with `devtools` automation protocol?

No. These will be available in all tests running WebdriverIO. I will also implement them to work on sessions running at Sauce Labs.

packages/webdriverio/src/utils/getNetworkObject.js
```diff
@@ -0,0 +1,111 @@
...
+                browserURL: `http://localhost:${rdPort}`,
+                defaultViewport: null
+            })
+            return this.puppeteer
+        }
+    }
+
+    throw new Error(
+        'Network primitives aren\'t available for this session. ' +
+        'This feature is only support for local Chrome, Firefox and Edge testing.'
```
mgrybyk, line 109: This feature is only support_ed_ for local Chrome, Firefox, and Edge testing.

christian-bromann:
> Can you please add a smoke / e2e test for network command?

This PR doesn't contain any command that could be tested with a smoke test. I will leave this to the subsequent PRs.

christian-bromann:
Please don't merge this PR. I will merge the rest of the network features into here to ship it as a whole.

PR #5472, opened 2020-06-09
Implement Network Throttling

christian-bromann:
Throttling is a useful feature that can help emulate certain scenarios (e.g. user looses internet connection).

```js
/**
 * throttle network
 * a) via static string
 * b) via custom values
 */
browser.network.throttle('Regular 3G')
browser.network.throttle({
    'offline': false,
    'downloadThroughput': 200 * 1024 / 8,
    'uploadThroughput': 200 * 1024 / 8,
    'latency': 20
})
```

### DIFF
--- a/packages/webdriverio/src/commands/network/throttle.js
+++ b/packages/webdriverio/src/commands/network/throttle.js
@@ -1,8 +1,129 @@
 /**
- * some description
+ * Throttle the network capabilities of the browser. This can help to
+ * emulate certain scenarios where a user looses the internet connection
+ * and your app needs to address that.
+ *
+ * <example>
+    :throttle.js
+    it('should throttle the network', () => {
+        // via static string
+        browser.network.throttle('Regular 3G')
+
+        // via custom values
+        browser.network.throttle({
+            'offline': false,
+            'downloadThroughput': 200 * 1024 / 8,
+            'uploadThroughput': 200 * 1024 / 8,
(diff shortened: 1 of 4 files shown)


packages/webdriverio/src/commands/network/throttle.js
```diff
@@ -1,3 +1,129 @@
...
+     * use WebDriver extension if connected with cloud service
+     */
+    if (this.isSauce) {
+        const browser = getBrowserObject(this)
+        await browser.throttleNetwork(params)
+        return null
+    }
+
+    // Connect to Chrome DevTools
+    const pages = await this.puppeteer.pages()
```
mgrybyk, line 117: I guess we need to add some verification to make sure devtools automation protocol is used otherwise will get an error "cannot read pages of undefined" ?

christian-bromann: If Puppeteer wasn't able to connect to the browser it will [throw an error](https://github.com/webdriverio/webdriverio/blob/cb-network-scope/packages/webdriverio/src/utils/getNetworkObject.js#L67-L108) it would not allow to even execute this command.

christian-bromann:
> Will you be able to add types for the new command? It should be easier to choose one of network presets with help of autocompletion

@mgrybyk is it ok for you of we have this as separated PR in #5480 ?

PR #5471, opened 2020-06-11
Implement Network Mocking

christian-bromann:
Interacting with the network is already straight forward in Puppeteer and requires not much sugar coding. Something we might want to abstract is matching urls based on globbing (instead of registering an event and look into the url manually) as well as the ability to mock out every incoming request no matter which page someone is on.

1. Creating a mock is done with a simple command that returns a mock object people can use to modify the request:

    ```js
    // a simple mock based on url filter
    const mock = browser.network.mock('/foo/**/bar')

    // a mock with url and extra filters based on HTTP method or headers, e.g.
	// only post requests of that url
    const mock = browser.network.mock('/foo/**/bar', {
		method: 'post'
    })
    // only requests with a specific header
    const mock = browser.network.mock('/foo/**/bar', {
		header: {
			'Content-Type': 'application/json'
		}
    })
    ```

1. I believe the most common usage is to just mock out a certain url and provide a response body instead:

    ```js
    /**
     * mock an endpoint with a fixed fixture
     */
    mock.respond([{
        title: 'Injected Todo',
        order: null,
        completed: false,
        url: "http://todo-backend-express-knex.herokuapp.com/916"
    }])

	/**
	 * respond with different status code or header
	 */
	mock.respond([{
        title: 'Injected Todo',
        order: null,
        completed: false,
        url: "http://todo-backend-express-knex.herokuapp.com/916"
    }], {
		statusCode: 404,
		headers: {
			'x-custom-header': 'foobar'
		}
	})

	/**
	 * replace response with content of local file
	 */
	mock.respond('./node_modules/file.js')

	/**
	 * replace response with other web resource
	 */
	mock.respond('https://test.com/some-image.png')

    // open ToDo app
    browser.url('https://todobackend.com/client/index.html?https://todo-backend-express-knex.herokuapp.com/')

    $('#todo-list li').waitForExist()
    console.log($$('#todo-list li').length) // 1
    console.log($('#todo-list li label').getText()) // Injected Todo
    ```

1. Multiple individual responses:

	```js
    mock.respondOnce([{
        title: 'Injected Todo'
    }])
    mock.respondOnce([{
        title: 'Another Injected Todo'
    }])

    browser.url('https://todobackend.com/client/index.html?https://todo-backend-express-knex.herokuapp.com/')
    console.log($('#todo-list li label').getText()) // Injected Todo
    browser.refresh()
    console.log($('#todo-list li label').getText()) // Another Injected Todo
	```

1. Dynamic Mocks allow to set responses based on the original response payload:
    ```js
    mock.respond((req) => {
        return req.body.map((item, i) => ({ ...item, title: i }))
    })

    browser.url('https://todobackend.com/client/index.html?https://todo-backend-express-knex.herokuapp.com/')

    $('#todo-list li').waitForExist()
    console.log($$('#todo-list li').length) // 1
    console.log($$('#todo-list li label').map((el) => el.getText()))
	// returns 
    // [
    //   '0',  '1',  '2',  '19', '20',
    //   '21', '3',  '4',  '5',  '6',
    //   '7',  '8',  '9',  '10', '11',
    //   '12', '13', '14', '15', '16',
    //   '17', '18', '22'
    // ]
    ```

1. Abort requests:

	```js
	mock.abortOnce('NameNotResolved')
    mock.abort('ConnectionFailed')
	```

1. Restore Mocks: to restore the mocked resources and have it as it was before we can just call:

    ```js
    // Resets all information stored in the `mock.calls` set.
    mock.clear()
    // Does everything that `mock.clear()` does, and also
	// removes any mocked return values or implementations.
    mock.restore()
    ```


christian-bromann:
I modified the interface slightly to allow to work with mocks as variables. This simplifies the implementation of assertion helpers and we could do things like:

```js
const mock = browser.network.mock('/foo/**/bar')
expect(mock).toBeCalledTimes(1)
expect(mock).toBeCalledWithResponse({ foo: 'bar' })
// etc.
```

christian-bromann:
I am working on implementing this behavior on the Sauce platform which will make it work the same way as locally. Firefox is not yet supporting the `Fetch` domain yet but they have it on their [immediate roadmap](https://bugzilla.mozilla.org/show_bug.cgi?id=1591389).

christian-bromann:
Missing in this PR:

- allow to also change the status code or header of a response
- allow to mock by
	- local file (e.g. replace an asset with a local asset)
	- redirect resource (by applying a web resource)

christian-bromann:
This PR is ready for review now 🎉

packages/webdriverio/src/utils/index.js
```diff
@@ -510,3 +510,36 @@ export const updateCapabilities = async (params, automationProtocol) => {
...
+ * @param {String} file file to check access for
+ * @return              true if file can be accessed
+ */
+export const canAccess = (file) => {
+    if (!file) {
+        return false
+    }
+
+    try {
+        fs.accessSync(file)
```
mgrybyk, line 540: Can we avoid using sync functions in Promise?

canAccess is called in async function above

christian-bromann: What is the harm of using the synchronous version here? We use it for other `fs` functions to and there is no significant burden to it.

mgrybyk: Ok, let's do this way

packages/webdriverio/src/utils/interception/devtools.js
```diff
@@ -0,0 +1,182 @@
...
+
+                const { errorReason, overwrite, params } = mock.respondOverwrites[0].sticky
+                    ? mock.respondOverwrites[0]
+                    : mock.respondOverwrites.shift()
+
+                /**
+                 * when response is modified
+                 */
+                if (overwrite) {
+                    let body = overwrite
```
mgrybyk, line 73: there is also `body` variable declared at line 47. I know they are not conflicting but might be confusing.

packages/webdriverio/src/utils/interception/devtools.js
```diff
@@ -0,0 +1,182 @@
...
+                        ...Object.entries(params.headers || {}).map(([key, value]) => { key, value })
+                    ]
+
+                    /**
+                     * check if local file and load it
+                     */
+                    const responseFilePath = path.isAbsolute(body) ? body : path.join(process.cwd(), body)
+                    if (fs.existsSync(responseFilePath) && canAccess(responseFilePath)) {
+                        body = fs.readFileSync(responseFilePath).toString()
+                    } else if (body.startsWith('http')) {
```
mgrybyk, line 94: can you please help understanding use case of this one?

packages/webdriverio/src/utils/interception/devtools.js
```diff
@@ -0,0 +1,182 @@
...
+                }
+
+                const { errorReason, overwrite, params } = mock.respondOverwrites[0].sticky
+                    ? mock.respondOverwrites[0]
+                    : mock.respondOverwrites.shift()
+
+                /**
+                 * when response is modified
+                 */
+                if (overwrite) {
```
mgrybyk, line 72:
```suggestion
                if (overwrite !== undefined) {
```

mgrybyk:
@christian-bromann what do you think about changing signature to 
```
mock.respond({
        body: { foo: 'bar' },
        statusCode: 200,
        headers: { 'Content-Type': 'application/json' }
    })
```
?

It looks for me more readable in comparison to 
```
mock.respond({ foo: 'bar' }, {
        statusCode: 200,
        headers: { 'Content-Type': 'application/json' }
    })
```

packages/webdriverio/src/utils/index.js
```diff
@@ -510,3 +510,18 @@ export const updateCapabilities = async (params, automationProtocol) => {
...
 }
+
+/**
+ * compare if an object (`base`) contains the same values as another object (`match`)
+ * @param {object} base  object to compare to
+ * @param {object} match object that needs to match thebase
+ */
+export const containsObject = (base, match) => {
+    for (const [key, value] of Object.entries(match)) {
+        if (typeof base[key] === 'undefined' || base[key] !== value) {
```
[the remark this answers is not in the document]
christian-bromann: this method is to check whether the header filter which you can apply if you define a mock, e.g.

```js
const mock = browser.network.mock('/foo/**/bar', {
	header: {
		'Content-Type': 'application/json'
	}
})
```

matches the headers in the request. There is no case where header properties are undefined. I renamed the method to give it a more explicit name.

packages/webdriverio/src/utils/interception/devtools.js
christian-bromann: This allows to just replace a web resource with another web resource. This is helpful if your mock (e.g. a JS or CSS file) is already hosted somewhere on the internet, e.g.

```js
mock.respond('https://test.com/some-image.png')
```

christian-bromann:
> what do you think about changing signature to

Mhm .. I think in most cases you don't apply any custom params and only do:

```js
mock.respond({ foo: 'bar' })
```

which would look cumbersome if you had to set it like this:

```js
mock.respond({ body: { foo: 'bar' } })
```

So I prefer to stick with the current interface and see what feedback people provide.

christian-bromann:
> * Can you please add types browser.network?

These are implemented in #5480

> * I'm not able to use empty string as a response body - `mock.respond('')`. In such case mocking doesn't work. Can you fix this?

Done!

PR #5488, opened 2020-06-13
Print cid in logger and not in the stream transform.

### LINKED ISSUE #5479: Add option to not print cid
I am running my tests always with a single worker instance so all of the logs are prepended with the [0-0] or [0-1] etc. As I checked the sources this is called 'cid' and is prepended in RunnerTransformStream class. However, for unknown reason sometimes the cid is not prepended and the log looks like this:

```
[0-1] 2020-06-10T06:51:40.670Z INFO E2E: ✓ ------------------------------------------------------------
2020-06-10T06:51:40.670Z INFO E2E: ✓ Running: Assure Upload settings are correct
[0-1] 2020-06-10T06:51:40.671Z INFO E2E: ✓ ------------------------------------------------------------
```

I'd like to have an option to not print these values as they have no meaning for me and only make the logs bigger and less readable.

I tried to set the JEST_WORKER_ID environment variable but that has other consequences.

**Additional context**
The missing cid happens for me on Windows 10 and also on Ubuntu 16.

RafalSkorka:
## Proposed changes

[//]: # This change will print the worker cid in the logger and will not use the transform stream for that.
This should resolve #5479 
## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/wdio-logger/src/node.js
+++ b/packages/wdio-logger/src/node.js
@@ -119,7 +119,7 @@ export default function getLogger (name) {
     loggers[name].setLevel(logLevel)
     loggers[name].methodFactory = wdioLoggerMethodFactory
     prefix.apply(loggers[name], {
-        template: `[${process.env.cid}] %t %l %n:`,
+        template: `${process.env.cid ? '[' + process.env.cid + '] ' : ''}%t %l %n:`,
         timestampFormatter: (date) => chalk.gray(date.toISOString()),
         levelFormatter: (level) => chalk[COLORS[level]](level.toUpperCase()),
         nameFormatter: (name) => chalk.whiteBright(name)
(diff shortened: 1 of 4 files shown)


packages/wdio-logger/src/node.js
```diff
@@ -119,7 +119,7 @@ export default function getLogger (name) {
     loggers[name].setLevel(logLevel)
     loggers[name].methodFactory = wdioLoggerMethodFactory
     prefix.apply(loggers[name], {
-        template: '%t %l %n:',
+        template: `[${process.env.cid}] %t %l %n:`,
```
christian-bromann, line 122: This would create an empty space for every log where @wdio/logger is not used with the testrunner.

RafalSkorka:
> 
> 
> Instead of applying this change to the global logger, how about we check in the [interface class](https://github.com/webdriverio/webdriverio/blob/master/packages/wdio-cli/src/interface.js#L165) whether there is line break and attach the cid there.

I tried that, but that would only print cid for some of cli logs, not for anything that is logged from the test itself.

RafalSkorka:
I realized I would need to modify the log line even more, so I'll not use the wdio.logger.

PR #5506, opened 2020-06-17
Add mochaOpts.grep documentation

pjcalvo:
## Proposed changes

This is a simple PR to add instructions on how we can use the mochaOpts flag to filter specific tests cases using the `.grep` flag.

## Types of changes

- [X ] Documentation update

## Checklist

- [X ] I have read the [CONTRIBUTING](https://github.com/webdriverio/webdriverio/blob/master/CONTRIBUTING.md) doc
- [ X] I have added necessary documentation (if appropriate)

## Further comments

I have seen people looking for this feature and just throwing the towel because there is no explicit instructions.

### DIFF
--- a/docs/OrganizingTestSuites.md
+++ b/docs/OrganizingTestSuites.md
@@ -174,6 +174,17 @@ grep -r -l --include "*.js" "myText" | wdio wdio.conf.js
 
 _**Note:** This will_ not _override the `--spec` flag for running a single spec._
 
+## Running Specific Tests with MochaOpts
+
+You can also filter which specific `suite|describe` and/or `it|test` you want to run by passing a mocha specific argument: `--mochaOpts.grep` to the wdio CLI.
+
+```sh
+wdio wdio.conf.js --mochaOpts.grep myText
+wdio wdio.conf.js --mochaOpts.grep "Text with spaces"
+```
+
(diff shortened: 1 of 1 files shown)


docs/OrganizingTestSuites.md
```diff
@@ -174,6 +174,17 @@ grep -r -l --include "*.js" "myText" | wdio wdio.conf.js
...
+## Running Specific Tests with MochaOpts
+
+You can also filter which specific `suite|describe` and/or `it|test` you want to run by passing a mocha specific argument: `--mochaOpts.grep` to the wdio CLI.
+
+```sh
+wdio wdio.conf.js --mochaOpts.grep myText
+wdio wdio.conf.js --mochaOpts.grep "Text with spaces"
+```
+
+_**Note:** Mocha will filter the tests after the WDIO test runner creates the instances, so you might see several instances being spawned but not actually executed.
```
christian-bromann, line 186:
```suggestion
**Note:** Mocha will filter the tests after the WDIO test runner creates the instances, so you might see several instances being spawned but not actually executed.
```

pjcalvo: Actually I was missing the _ at the end, to make it italic style.

PR #5528, opened 2020-06-22
Check for nightly channel to handle devtools

mathew-jithin:
## Proposed changes

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/devtools/src/launcher.js
+++ b/packages/devtools/src/launcher.js
@@ -120,6 +120,8 @@ function launchBrowser (capabilities, product) {
 
     if (!executablePath) {
         throw new Error('Couldn\'t find executable for browser')
+    } else if (!executablePath.includes('Nightly')) {
+        throw new Error('Only Nightly release channel is supported in Devtools for Firefox')
     }
 
     log.info(`Launch ${executablePath} with config: ${JSON.stringify(puppeteerOptions)}`)


mathew-jithin:
@christian-bromann  Added a very basic handling for this here. Some doubts actually, if you can please help here:

- Do we need to handle binary name changes (Asking this to check whether we need to add specific handling in this case)? As this is a quick fix for a feature that will be fixed in the near future, do we need to add this? Current change which I did only checks for Nightly in the binary path, or else will fail. If someone explicitly changes the name of the installed firefox app, this fix will fail for all versions.

- Is a format similar to this needed? https://www.selenium.dev/selenium/docs/api/javascript/module/selenium-webdriver/firefox_exports_Channel.html

sethuster:
Just to be clear, the 'channel' here is determined by the executable path where FireFox is installed in.  If the path contains nightly, this version of FireFox was installed from that channel.  

Is that correct?

mathew-jithin:
@sethuster Nope. Actually Firefox follows 4 release channels like @christian-bromann had mentioned:

Please refer to these docs. I think these docs would be better than me explaining :-) :

https://wiki.mozilla.org/Release_Management/Release_Process#Channels.2FRepositories
https://www.selenium.dev/selenium/docs/api/javascript/module/selenium-webdriver/firefox_exports_Channel.html

packages/devtools/src/launcher.js
```diff
@@ -120,6 +120,8 @@ function launchBrowser (capabilities, product) {
 
     if (!executablePath) {
         throw new Error('Couldn\'t find executable for browser')
+    } else if (!executablePath.toLowerCase().includes(CHANNEL_FIREFOX_NIGHTLY)) {
```
christian-bromann, line 123: This check should be skipped if we don't want to start Firefox.

packages/devtools/src/launcher.js
```diff
@@ -120,6 +120,8 @@ function launchBrowser (capabilities, product) {
...
+        throw new Error('Only Nightly release channel is supported in Devtools for Firefox')
```
christian-bromann, line 124: Let's add a longer error message explaining where you could download FF nightly and how to run it (setting firefox options).

packages/devtools/src/launcher.js
mathew-jithin: Haa! Yes. I just missed that Edge also uses the same function. My bad! Will fix this. Thanks!

packages/devtools/src/launcher.js
mathew-jithin: Sure. Will add it. And will fix the tests and add the unit tests too. Thanks!

mathew-jithin:
@christian-bromann I have made the review changes and added/changed the corresponding unit tests.

Actually I am not sure how to fix this e2e test. So, the same is breaking. It's due to the same nightly consideration:

- should allow to run multiple browser at once

packages/devtools/src/constants.js
```diff
@@ -93,8 +99,27 @@ export const ERROR_MESSAGES = {
...
+    firefoxNightly: `Only Nightly release channel is supported in Devtools/Puppeteer for Firefox. Refer to the following issue:
+        https://bugzilla.mozilla.org/show_bug.cgi?id=1606604
+
+        You can use the following link to download Firefox Nightly edition:
+        https://www.mozilla.org/en-US/firefox/channel/desktop/
+
+        Adding the following binary capability in Firefox Options is mandatory to run with Nightly edition:
+
+        'moz:firefoxOptions': {
+         binary: '/path/to/firefox'
```
christian-bromann, line 112:
```suggestion
            binary: '/path/to/firefox'
```

packages/devtools/src/constants.js
```diff
@@ -93,8 +99,27 @@ export const ERROR_MESSAGES = {
...
+        }
+
+        Note: "Nightly" as a term should be present in the "Firefox Application Name" across all OS's in binary path mentioned above for this to work.
+        `
```
christian-bromann, line 116:
```suggestion
        Note: "Nightly" as a term should be present in the "Firefox Application Name" across all OS's in binary path mentioned above for this to work.`
```

packages/devtools/tests/launcher.test.js
```diff
@@ -151,14 +171,33 @@ test('throws if browser is unknown', async () => {
...
+            browserName: 'firefox',
+            'moz:firefoxOptions': {
+                binary: '/foo/firefox',
+                headless: true
+            },
+            ignoreDefaultArgs: true
+        })
+    } catch (err) {
+        expect(err.message).toContain('Only Nightly release channel is supported')
+    }
```
christian-bromann, line 198: This test would also pass if `launch` would not throw an error here. You can either use `expect.assertions(1)` or what I prefer:

```suggestion
    const err = await launch({
        browserName: 'firefox',
        'moz:firefoxOptions': {
            binary: '/foo/firefox',
            headless: true
        },
        ignoreDefaultArgs: true
    }).catch(err => err)
    expect(err.message).toContain('Only Nightly release channel is supported')
```

christian-bromann:
> Actually I am not sure how to fix the below e2e test

Which e2e test are you referring to?

mathew-jithin:
> > Actually I am not sure how to fix the below e2e test
> 
> Which e2e test are you referring to?

@christian-bromann This one:

**should allow to run multiple browser at once** -- I think this runs on devtools as a fallback?

christian-bromann:
> I think this runs on devtools as a fallback?

Exactly. It should pass though because in Travis [we've defined](https://github.com/webdriverio/webdriverio/blob/master/.travis.yml#L5-L11) Firefox Nightly as the only browser on the system.

mathew-jithin:
@christian-bromann Oh! 👍

packages/devtools/tests/launcher.test.js
mathew-jithin: Handled it with `expect.assertions(1)`

mathew-jithin:
Have made all the suggested fixes.

Issue #5336, opened 2020-04-28
Proposal: Adding network primitives to the browser object


christian-bromann:
# Proposal

As we have better cross browser support for Puppeteer we should move forward and provide better network primitives for WebdriverIO to help user access the network layer. In addition to that these features are likely to be supported once WebDriver moves to a BiDi protocol (which is a new version of WebDriver that is currently designed by the W3C working group and browser vendors). Here are some of my ideas:

## Goal

The goal of this proposal is to simplify network primitives for users without exposing them to too much detail. Commands should allow to do the majority over normal use cases with as less coding as possible while always providing the ability to access Puppeteer native APIs.
Additionally we should create useful commands that help with the testing flow, e.g.: integrate it into our assertion library.

## Compatibility

These features will be only supported in Chrome, Firefox (beta) and Edge when running locally as well as on Sauce Labs (using extended debugging capabilities). With release of Selenium 4 this can be also supported if someone has an internal grid or uses the selenium standalone service.

## State

The state of these mocked network resources will stay for the live time of the session. So if the user switches between windows, opens and closes pages, the mocked resources stays as defined. Adding the ability to reset a certain mocked resources makes sense here. Again a power user will be always able to use Puppeteer directly if he desires more granular access to CDP.

### Network Mocking

See proposal: https://github.com/webdriverio/webdriverio/pull/5471

### Throttling

See proposal: https://github.com/webdriverio/webdriverio/pull/5472

### Events

See (rejected) proposal: https://github.com/webdriverio/webdriverio/issues/5473

Not considered for implementation anymore.

### Wait Commands

See proposal: https://github.com/webdriverio/webdriverio/issues/5474

### Overwriting User Agent

See (rejected) proposal: https://github.com/webdriverio/webdriverio/issues/5475

Not considered for implementation anymore.

### Recording

See proposal: https://github.com/webdriverio/webdriverio/issues/5476

erwinheitzman:
I love this idea and am wondering how the state of these mocks work, are they lost after the test, do they stay and have to be cleaned or reset?

christian-bromann:
@erwinheitzman added a section to the state of mocks

erwinheitzman:
Would it be possible to extend the given examples further like so?
```js
browser.network.waitForResponse({
  url: '/foo/**/bar',
  method: 'post',
  statusCode: 400,
  // maybe a specific body that equals or matches, etc
  // etc..
}, 4000)
```

[...]

PR #5480, opened 2020-06-26
Implement TypeScript Support

christian-bromann:
All new commands proposed in #5336 require TypeScript support. Let's implement this as part of this issue.

### DIFF
--- a/packages/wdio-sync/webdriverio-core.d.ts
+++ b/packages/wdio-sync/webdriverio-core.d.ts
@@ -835,7 +835,7 @@ declare namespace WebdriverIO {
          */
         mock(
             url: string,
-            filterOptions: MockFilterOptions
+            filterOptions?: MockFilterOptions
         ): Mock;
 
         /**
@@ -870,7 +870,7 @@ declare namespace WebdriverIO {
          */
         respond(
             overwrites: object | string,
-            params: MockResponseParams
+            params?: MockResponseParams
         ): void;
 
         /**
@@ -878,7 +878,7 @@ declare namespace WebdriverIO {
          */
         respondOnce(
             overwrites: object | string,
-            params: MockResponseParams
+            params?: MockResponseParams
         ): void;
 
         /**
--- a/packages/wdio-sync/webdriverio-core.d.ts
+++ b/packages/wdio-sync/webdriverio-core.d.ts
@@ -435,6 +435,18 @@ declare namespace WebdriverIO {
         y: number
     }
 
+    type MockResponseParams = {
+        statusCode?: number,
+        headers?: object
+    }
+
+    type MockFilterOptions = {
+        method?: string,
+        headers?: object
+    }
+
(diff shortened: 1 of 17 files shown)


packages/wdio-sync/webdriverio-core.d.ts
```diff
@@ -435,6 +435,18 @@ declare namespace WebdriverIO {
...
     }
 
+    type MockResponseParams = {
+        statusCode?: number,
+        headers?: object
+    }
+
+    type MockFilterOptions = {
+        method?: string,
+        headers?: object
```
mgrybyk, line 445:
```suggestion
        headers?: Record<string, string>
```

packages/wdio-sync/webdriverio-core.d.ts
```diff
@@ -813,6 +825,68 @@ declare namespace WebdriverIO {
...
+        /**
+         * Resets all information stored in the `mock.calls` array.
+         */
+        clear(): void;
+
+        /**
+         * Always respond with same overwrite.
+         */
+        respond(
+            overwrites: object | string,
```
mgrybyk, line 872: I guess we can change it to any? Or at least Record<string, unknown>, Array<Record<string, unknown>>, Function, string.

packages/webdriverio/src/commands/network/throttle.js
```diff
@@ -1,4 +1,8 @@
-/* istanbul ignore file */
+/**
+ * some description
```
mgrybyk, line 2: are you going to add description?

packages/webdriverio/webdriverio-core.d.ts
```diff
@@ -435,6 +435,18 @@ declare namespace WebdriverIO {
         y: number
     }
 
+    type MockResponseParams = {
+        statusCode?: number,
+        headers?: object
```
mgrybyk, line 440:
```suggestion
        headers?: Record<string, string>
```

packages/webdriverio/webdriverio-core.d.ts
```diff
@@ -435,6 +435,18 @@ declare namespace WebdriverIO {
...
+    }
+
+    type MockFilterOptions = {
+        method?: string,
+        headers?: object
```
mgrybyk, line 445:
```suggestion
        headers?: Record<string, string>
```

packages/webdriverio/src/commands/network/throttle.js
christian-bromann: Yes, it is part of this PR: https://github.com/webdriverio/webdriverio/pull/5472/files#diff-0a3a23f19504d4292082bcf209d07d31R1-R26

packages/wdio-sync/webdriverio-core.d.ts
christian-bromann: actually this could be a function too, so all usages are as follows:

```js
mock.respond([{
    title: 'Injected Todo',
    order: null,
    completed: false,
    url: "http://todo-backend-express-knex.herokuapp.com/916"
}])
mock.respond('./node_modules/file.js')
mock.respond('https://test.com/some-image.png')
mock.respond((req) => {
    return req.body.map((item, i) => ({ ...item, title: i }))
})
```

So I will update it to `object | Function | string`

PR #5691, opened 2020-08-05
Catch exception caused if deleteSession is called during test execution

mikesalvia:
## Proposed changes

This PR will trap and discard exceptions causes by `deleteSession` at the end of the test runner life cycle.

My use case started as I wanted to inspect the network logs available from Browserstack to verify traffic was being sent and received properly. I wanted to fail the test if we were missing any of these items. In order to download the network logs, it is required to have closed the current session. I call `browser.deleteSession()` I am able to capture network logs but then at this stage (at the end of the runner lifecycle), there is a bunch of warnings and errors displayed, however, these are for a certain part known.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/wdio-runner/src/index.js
+++ b/packages/wdio-runner/src/index.js
@@ -350,7 +350,7 @@ export default class Runner extends EventEmitter {
         try {
             await global.browser.deleteSession()
         } catch (err) {
-            console.log('Failed or delete session, it has either timed out or has been ended prior to test runner completion.')
+            console.log('Failed to delete session, it has either timed out or has been ended prior to test runner completion.')
         }
         
         /**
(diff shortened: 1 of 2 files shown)


mikesalvia:
@christian-bromann Maybe this should be put in `deleteSession()` since the trapping is now done in 2 places (`reloadSession`, and `endSession`)

christian-bromann:
@mikesalvia I don't think this is the right approach. A session should not be closed prematurely. I recommend to build a service that checks the network logs in the `onComplete` hook and throws a [`SevereServiceError`](https://webdriver.io/docs/customservices.html) if they are not as expected. Suppressing the error can cause undesired side effects where browser linger around without being noticed by the user.

mikesalvia:
@christian-bromann Would that be able to create a "pass/fail" for the test though?

PR #5710, opened 2020-08-10
devtools: Fix `ignoreDefaultArgs` capability option

koggdal:
## Proposed changes

The value of `ignoreDefaultArgs` was previously passed directly to `launch` from the package `chrome-launcher`. However, that function doesn't take the same parameter, [it takes an `ignoreDefaultFlags` boolean parameter](https://github.com/GoogleChrome/chrome-launcher/blob/v0.13.4/src/chrome-launcher.ts#L40). It also doesn't do any filtering of arguments, it's either just ignoring all default flags or using all of them.

This changes code in `devtools` to always send `ignoreDefaultFlags: true`. This should be fine, since we have the default arguments in a list in this project, and passing all default arguments, together with custom arguments, onwards to `chrome-launcher`.

To not break previous behavior, I'm also adding these arguments as default arguments, as they were used from within the default arguments in `chrome-launcher`, but not in `devtools`:
* `--no-default-browser-check`
* `--force-fieldtrials=*BackgroundTracing/default/`

## Types of changes

- [ ] Documentation update

## Checklist


packages/devtools/src/launcher.js
```diff
@@ -79,7 +79,7 @@ async function launchChrome (capabilities) {
 
     const chrome = await launchChromeBrowser({
         chromePath: chromeOptions.binary,
-        ignoreDefaultArgs,
+        ignoreDefaultFlags: true,
         chromeFlags
```
SrinivasanTarget, line 83: @koggdal Can you also simplify the condition [here](https://github.com/koggdal/webdriverio/blob/devtools-fix-ignore-chrome-flags/packages/devtools/src/launcher.js#L55)?

packages/devtools/src/constants.js
```diff
@@ -23,17 +23,20 @@ export const DEFAULT_FLAGS = [
     '--mute-audio',
     // Skip first run wizards
     '--no-first-run',
+    '--no-default-browser-check',
     // chromedriver flags
     '--disable-hang-monitor',
     '--disable-prompt-on-repost',
     '--disable-client-side-phishing-detection',
+    '--disable-features=TranslateUI',
```
SrinivasanTarget, line 31: Should be removed as UItranslation is already disabled [here](https://github.com/koggdal/webdriverio/blob/devtools-fix-ignore-chrome-flags/packages/devtools/src/constants.js#L45)

packages/devtools/tests/__snapshots__/launcher.test.js.snap
```diff
@@ -107,16 +107,19 @@ Array [
...
         "--disable-client-side-phishing-detection",
+        "--disable-features=TranslateUI",
         "--password-store=basic",
         "--use-mock-keychain",
         "--disable-component-extensions-with-background-pages",
         "--disable-breakpad",
         "--disable-dev-shm-usage",
         "--disable-ipc-flooding-protection",
         "--disable-renderer-backgrounding",
+        "--force-fieldtrials=*BackgroundTracing/default/",
```
SrinivasanTarget, line 122: Do we need to disable background tracing by default? Will this have any impact on trace gatherer? @christian-bromann

packages/devtools/src/constants.js
koggdal: Oh good catch! Thanks!

packages/devtools/src/launcher.js
koggdal: That line should be unaffected by these changes. Could you clarify what you'd like me to simplify? The line you link to should still have the same logic, as we need to still support the filtering logic, and the boolean value. The line you commented on here changes because that's the API of `chrome-launcher` and we're doing all the filtering and default handling here before passing it to `chrome-launcher`.

packages/devtools/tests/__snapshots__/launcher.test.js.snap
koggdal: From looking at the [implementation in `chrome-launcher`](https://github.com/GoogleChrome/chrome-launcher/blob/v0.13.4/src/chrome-launcher.ts#L156-L174) it looks to me like it will always use `DEFAULT_FLAGS` if `ignoreDefaultFlags` is falsy. `ignoreDefaultFlags` should always have been falsy, since we weren't passing it (we passed `ignoreDefaultArgs`). [`DEFAULT_FLAGS`](https://github.com/GoogleChrome/chrome-launcher/blob/v0.13.4/src/flags.ts#L8) contains `--force-fieldtrials=*BackgroundTracing/default/`. So it seems to me like we've always gotten the argument, so this PR shouldn't have any impact on this, as from what I can tell, nothing changes in this regard. Of course, if you'd like to actually stop using the argument, let me know!

packages/devtools/src/launcher.js
SrinivasanTarget: Got it thanks.

koggdal:
Thanks for the review @SrinivasanTarget! 🙏 Seems like we need someone with write access. @christian-bromann? Thanks!

christian-bromann (approved):
Thanks for the review @SrinivasanTarget 

LGTM

PR #5754, opened 2020-08-24
@wdio/devtools-service udpate to 6.4.0 causes timeout at browser.relo…

joshskumar:
…adSession()

## Proposed changes

[//]: # change to the function passed in the map into an async function and have an await in the definition of the function.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

I would like the reviewer to review it in the angle of performance as well as I am not sure of the complexity involved in the execution of the hook function..

I have ran the tests and confirmed that the code flows into the If condition and it look good

### DIFF
--- a/packages/webdriverio/src/commands/browser/reloadSession.js
+++ b/packages/webdriverio/src/commands/browser/reloadSession.js
@@ -43,7 +43,7 @@ export default async function reloadSession () {
     await ProtocolDriver.reloadSession(this)
 
     if (Array.isArray(this.options.onReload) && this.options.onReload.length) {
-        await Promise.all(this.options.onReload.map((hook) => hook(oldSessionId, this.sessionId)))
+        await Promise.all(this.options.onReload.map(async (hook) => await hook(oldSessionId, this.sessionId)))
     }
 
     return this.sessionId


christian-bromann (changes requested):
Can you reference the issue you are trying to fix here?

I think 
```js
await Promise.all(this.options.onReload.map((hook) => await hook(oldSessionId, this.sessionId)))
```
and
```js
await Promise.all(this.options.onReload.map(async (hook) => await hook(oldSessionId, this.sessionId)))
```

do exactly the same. So I don't believe that this patch helps with anything.

joshskumar:
@christian-bromann Yes I agree it is doing the same

would you mind telling me where to look for the definition of the the hook method so I can give the right fix

christian-bromann:
> of the the hook method

which hook method?

joshskumar:
hi @christian-bromann  sorry I get it from where hook comes from after some research..

my basic idea of adding await there was to make sure the array of functions are mapped one at a time and then await for the entire Promise completion.. rather than starting off together and waiting for them all to complete..

but if you feel if this does not make sense I will close my PR

christian-bromann:
It is intentional that all service hooks run at the same time. No service should take precedence over another service.

joshskumar:
thanks for clarification @christian-bromann

PR #5770, opened 2020-08-28
git followTags

mgrybyk:
## Proposed changes

I'd like to set `push.followTags = true` for the repo so that other users don't need to set it manually.

Users who have disabled `postinstall` scripts can run the command manually.

I believe this is only required for the users who can release webdriverio.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,8 @@
     "ts:webdriverio-reporter": "cd tests/typings/webdriverio-reporter && tsc --incremental",
     "watch": "node ./scripts/watch",
     "watch:pkg": "lerna exec 'npm run compile -- --watch' --scope",
-    "version": "npm run changelog && git add CHANGELOG.md"
+    "version": "npm run changelog && git add CHANGELOG.md",
+    "postinstall": "git config --local include.path '../.gitconfig'"
   },
   "devDependencies": {
     "@babel/cli": "^7.6.2",
--- a/.gitconfig
+++ b/.gitconfig
@@ -1,0 +1,2 @@
+[push]
+followTags = true


package.json
```diff
@@ -58,7 +58,8 @@
     "ts:webdriverio-reporter": "cd tests/typings/webdriverio-reporter && tsc --incremental",
     "watch": "node ./scripts/watch",
     "watch:pkg": "lerna exec 'npm run compile -- --watch' --scope",
-    "version": "npm run changelog && git add CHANGELOG.md"
+    "version": "npm run changelog && git add CHANGELOG.md",
+    "postinstall": "git config --local include.path '../.gitconfig'"
```
christian-bromann, line 62:
```suggestion
    "postinstall": "git config --local include.path './.gitconfig'"
```

christian-bromann (commented):
Shouldn't it be this?

christian-bromann:
Also interestingly: Lerna is already [adding](https://github.com/lerna/lerna/blob/master/commands/version/lib/git-push.js#L12) a `--followTags` when a release is made.

package.json
mgrybyk: It should be .. because the path is relative to . git folder in the repo. We should go up to be in the project root.

mgrybyk:
Maybe the problem we facing is with the way github merges branches to master?

package.json
mgrybyk: Yes, the . gitconfig is in the repo root same as .git but the base path for any config path is relative to projectRoot/.git.
So we should go up to the project root from .git folder

mgrybyk: Yes, the . gitconfig is in the repo root same as .git but the base path for any config path is relative to projectRoot/.git.
So we should go up to the project root from .git folder

christian-bromann:
Closing in favor of #5794

christian-bromann:
It turns out that having this option set also doesn't force Lerna to push annotated tags.

PR #5659, opened 2020-08-31
docs: Don't have types being linked to their own pages

christian-bromann:
**Description of the improvement / report**
Our website is generated by Docusaurus (v1) and it is likely that it transforms the API markdown files so that such occurrences create links. We either need to modify the api template (in `/scripts/templates`) or need to find a way to create the markdown so that no such type link is created.

Before:
Screenshot 2020-08-31 at 11 18 51

After:
Screenshot 2020-08-31 at 11 18 43

### DIFF
--- a/scripts/templates/api.tpl.ejs
+++ b/scripts/templates/api.tpl.ejs
@@ -29,7 +29,7 @@ title: <?= method.command ?>
 | ---- | ---- | ------- |
 <? method.paramTags.forEach((paramTag) => {
     var paramKey = paramTag.types
-        ? paramTag.types.map((type) => `<code>${type}</code>`).join('|')
+        ? paramTag.types.map((type) => `<code>${type.replace(/>/g, "&gt;").replace(/</g, "&lt;")}</code>`).join('|')
         : paramTag.type
     ?>| <code><var><?= paramTag.name ?></var></code><? if ((!paramTag.required && typeof paramTag.optional === 'undefined') || paramTag.optional) { ?><br><span class="label labelWarning">optional</span><? }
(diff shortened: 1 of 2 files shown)


website/i18n/en.json
```diff
@@ -98,6 +98,9 @@
       "wdio-reportportal-service": {
         "title": "Report Portal Service"
       },
+      "wdio-rerun-service": {
+        "title": "Re-run Service Service"
```
mgrybyk, line 102: I guess you've not rebased with the latest master because it should be only one `Service` word here.

I can update en.json in my TypeScript PR

scripts/templates/api.tpl.ejs
```diff
@@ -28,7 +28,9 @@ title: <?= method.command ?>
 | Name | Type | Details |
 | ---- | ---- | ------- |
 <? method.paramTags.forEach((paramTag) => {
-    var paramKey = (paramTag.types && paramTag.typesDescription) ? paramTag.typesDescription : paramTag.type
+    var paramKey = paramTag.types
+        ? paramTag.types.map((type) => `<code>${type.replace(/>/g, "&gt;").replace(/</g, "&lt;")}</code>`).join('|')
```
mgrybyk, line 32: Is it possible to get value without html tags from `type`?
Sorry to say I don't know what is `type` Object.

christian-bromann: type could be `Promise.<any>` and I am replacing the `<` and `>` here to not quit the `</code>` to early.

mgrybyk (approved):

PR #5788, opened 2020-09-01
fix: add null check for scenario outline

BlueMona:
## Proposed changes

Fix the following error, which I encountered after the first case of a Scenario Outline: 

ERROR @wdio/runner: TypeError: Cannot read property 'indexOf' of undefined
at /.../test-project/node_modules/@wdio/cucumber-framework/build/utils.js:86:20
at Array.forEach (<anonymous>)

## Types of changes

- [ ] Documentation update

## Checklist

- [ x] I have read the [CONTRIBUTING](https://github.com/webdriverio/webdriverio/blob/master/CONTRIBUTING.md) doc

## Further comments

### DIFF
--- a/packages/wdio-cucumber-framework/src/utils.js
+++ b/packages/wdio-cucumber-framework/src/utils.js
@@ -69,7 +69,7 @@ export function getUniqueIdentifier (target, sourceLocation) {
         if (Array.isArray(target.examples)) {
             target.examples.forEach((example) => {
                 example.tableHeader.cells.forEach((header, idx) => {
-                    if (name.indexOf('<' + header.value + '>') === -1) {
+                    if (!name || name.indexOf('<' + header.value + '>') === -1) {
                         return
                     }
 


mgrybyk:
Can you please provide with a reproducible scenario?

mgrybyk:
Can you please provide with a reproducible scenario?

christian-bromann:
@BlueMona you can find our smoke tests [here](https://github.com/webdriverio/webdriverio/tree/master/tests/cucumber). Currently we expect that
```js
let name = target.name || target.text
```
always results in a name and is never `undefined`. It would be useful to understand why that is in your case.

BlueMona:
In my own test repo I was trying to run a cucumber Scenario Outline like this one. I was able to watch the test complete the first case, apparently correctly in the browser, then error out. I included the full stack trace at the bottom.  
I haven't looked deep enough yet into the wdio repo to write a test case and I am not sure I will have time to do it but I'll try to keep an eye on this issue.

Scenario Outline:
    When I make a booking with <card> card
    Then the booking is created  

Examples: 
    | card       |
    | Visa       | 
    | Mastercard | 

```[0-0] 2020-09-01T20:56:06.597Z ERROR @wdio/runner: TypeError: Cannot read property 'indexOf' of undefined
    at /...../git/test-project/node_modules/@wdio/cucumber-framework/build/utils.js:85:20
    at Array.forEach (<anonymous>)
    at /...../git/test-project/node_modules/@wdio/cucumber-framework/build/utils.js:84:35
    at Array.forEach (<anonymous>)
    at getUniqueIdentifier (/...../git/test-project/node_modules/@wdio/cucumber-framework/build/utils.js:83:23)
    at CucumberReporter.handleAfterScenario (/...../git/test-project/node_modules/@wdio/cucumber-framework/build/reporter.js:175:43)
    at CucumberEventListener.emit (events.js:310:20)
    at CucumberEventListener.EventEmitter.emit (domain.js:482:12)
    at CucumberEventListener.onTestCaseFinished (/...../git/test-project/node_modules/@wdio/cucumber-framework/build/cucumberEventListener.js:135:10)
    at EventEmitter.emit (events.js:310:20)
    at EventEmitter.emit (domain.js:482:12)
    at TestCaseRunner.emit (/...../git/test-project/node_modules/cucumber/lib/runtime/test_case_runner.js:92:27)
    at TestCaseRunner.run (/...../git/test-project/node_modules/cucumber/lib/runtime/test_case_runner.js:227:12)
    at async Runtime.runTestCase (/...../git/test-project/node_modules/cucumber/lib/runtime/index.js:76:28)
[0-0]  Error:  Cannot read property 'indexOf' of undefined
[0-0] 2020-09-01T20:56:06.598Z INFO webdriver: COMMAND deleteSession()
2020-09-01T20:56:06.598Z INFO webdriver: [DELETE] http://localhost:4444/wd/hub/session/42febe7e-5152-8849-839c-5d37dc4ac9fb
[0-0] FAILED in firefox - /src/featureFiles/CustomerWebPortal/Booking/square.feature```

**Comment by osmolyar** [comment]
@BlueMona I wonder if the issue is that there is no reference in your steps or step data to the data in the Examples table, so it's not really seeing a Scenario Outline but just a Scenario.  Meaning, if your example above is not just pseudocode, then it should be something like `When I make a booking with <card>`, with the angle brackets to indicate parametrization with the card column in the Examples table.  Or if the step defines a table argument, then <card> could go in the step data.  But as it is there are not multiple scenarios in the outline.

**Comment by mgrybyk** [comment]
The more details we have the better!

@BlueMona it would be great if you can share your repository or provide with more accurate example.

@osmolyar even if there is a mistake in the feature file we should improve the error message because `Cannot read property 'indexOf' of undefined` is not informative enough.

**Comment by BlueMona** [comment]
I don't think I can share my repository but I can assure you that I have a step definition as follows: 

```When(/^I make a booking with (.*) card$/, (card) => {
   // do stuff
});```

It works when I have defined each Scenario individually (instead of using Scenario Outline), but since I have about 10 cards to test, this is a lot of repetition and unnecessary. 

It also works when I have modified the code in the node_modules as in this PR.

**Comment by BlueMona** [comment]
Sorry, it looks like I was missing the < > in the scenario outline, but I have tested again that this does not work with the current code but works with my fix: 

Scenario Outline:
    When I make a booking with <card> card
    Then the booking is created  

Examples: 
    | card              |
    | Visa              | 
    | Mastercard        | 
    | MastercardEU      |
    | Discover          |
    | DiscoverEU        | 
    | DinersClub        | 
    | JCB               | 
    | AmericanExpress   | 
    | AmericanExpressEU |
    | SquareGiftCard    |

**Comment by mgrybyk** [comment]
@BlueMona just add some name for your `Scenario Outline`, ex:
```
Scenario Outline: ADD SOME TITLE HERE
    When I make a booking with <card> card
    Then the booking is created  

Examples: 
    | card              |
    | Visa              | 
    | Mastercard        | 
```

File: packages/wdio-cucumber-framework/src/utils.js [hunk 1]
```diff
@@ -69,7 +69,7 @@ export function getUniqueIdentifier (target, sourceLocation) {
         if (Array.isArray(target.examples)) {
             target.examples.forEach((example) => {
                 example.tableHeader.cells.forEach((header, idx) => {
-                    if (name.indexOf('<' + header.value + '>') === -1) {
+                    if (!name || name.indexOf('<' + header.value + '>') === -1) {
```
mgrybyk, line 72:
```suggestion
                    if (name.indexOf('<' + header.value + '>') === -1) {
```

packages/wdio-cucumber-framework/src/utils.js
```diff
@@ -69,7 +69,7 @@ export function getUniqueIdentifier (target, sourceLocation) {
         if (Array.isArray(target.examples)) {
```
mgrybyk, line 69:
```suggestion
        if (!name) {
            name = path.basename(sourceLocation.uri)
        } else if (Array.isArray(target.examples)) {
```

mgrybyk (changes requested):
If we add `!name` to the condition the `name` will be `undefined`.
I'd suggest following the naming convention used for `'Hook'`.

Please also add either [unit](https://github.com/webdriverio/webdriverio/blob/master/packages/wdio-cucumber-framework/tests/utils.test.js#L286) or [smoke](https://github.com/webdriverio/webdriverio/blob/master/tests/cucumber/test.feature) test (or both!).

osmolyar:
@mgrybyk  @BlueMona just a note that if a Scenario Outline name is not parametrized with a header value from the Examples table, all the Scenario instances will be treated as reruns of the same test by at least some reporters (e.g. allure-reporter) and only one result will be reported for all.  So would think that the original suggestion of a better error message if Scenario Outline name is missing would be preferable to populating it with a string that does not depend on example data.

packages/wdio-cucumber-framework/src/utils.js
erwinheitzman: if (!name.includes('<' + header.value + '>')) {

mgrybyk:
let me add at least some warning message

Issue #4582, opened 2019-10-07
root level tests are not shown in reporters in Jasmine


mgrybyk:
**Environment (please complete the following information):**
 - **WebdriverIO version:** 5.14.4

**Config of WebdriverIO**
```
framework: 'jasmine',
reporters: ['spec]'
```

**Describe the bug**
test added in root scope is ignored in reporters

**To Reproduce**
create file with `it('root level test', () => {})`

**Expected behavior**
root level test should be shown in reporter

**Log**
`Cannot read property 'tests' of undefined`

christian-bromann:
see #5790

PR #5790, opened 2020-09-02
log warning if jasmine tests have no root describe

christian-bromann:
## Proposed changes

fixes: #4582

If users don't start with a root describe block it might have negative effects on the reporting. We can't insert a describe nor mock the events therefor with this patch I suggest to just log a warning.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

n/a

### DIFF
--- a/packages/wdio-jasmine-framework/src/reporter.js
+++ b/packages/wdio-jasmine-framework/src/reporter.js
@@ -36,7 +36,7 @@ export default class JasmineReporter {
         this.testStart = new Date()
         test.type = 'test'
         test.start = new Date()
-        let parentSuite = this.parent[this.parent.length - 1]
+        const parentSuite = this.parent[this.parent.length - 1]
 
         /**
          * if jasmine test has no root describe block, create one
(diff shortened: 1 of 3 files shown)


packages/wdio-jasmine-framework/src/reporter.js
```diff
@@ -32,7 +36,20 @@ export default class JasmineReporter {
...
+        /**
+         * if jasmine test has no root describe block, create one
+         */
+        if (!parentSuite) {
+            log.warn(
+                'No root suite was defined! This can cause reporters to malfunction. ' +
+                'Please always start a spec file with describe("...", () => { ... }).'
+            )
+        } else {
+            parentSuite.tests++
```
mgrybyk, line 50: I guess we should always call this one to keep existing behavior

mgrybyk (changes requested):

packages/wdio-jasmine-framework/src/reporter.js
christian-bromann: @mgrybyk the existing behavior is an error being thrown as mentioned in your issue:

> Cannot read property 'tests' of undefined

mgrybyk (approved):
:+1:

PR #5804, opened 2020-09-07
Get rid of Babel and rewrite @wdio/repl

### LINKED ISSUE #5424: Setup TypeScript for the project
As the code base is growing and the ecosystem leaning more and more towards TypeScript as type safe JavaScript flavor it makes sense to introduce it into the project as well. It help us grow the code base while at the same time reduce the amount of errors due to type issues.

Let's get rid of Babel as compiler and replace it with TypeScript to start type safe programming. As we can't just rewrite the whole code base we should introduce a setup that allows us to continue to use our ESNext code while start writing new code in TypeScript.

Alternatively I considered just getting rid of Babel and convert everything to ES Modules (see #5401). However this can still be achieved with TypeScript.

**Additional context**
This discussion was triggered by #5399 and was discussed with the @webdriverio/technical-steering-committee. Everyone agreed that switching to TypeScript is a good idea. The reason why (particular myself) was hesitating towards this was that adding TypeScript to the code base increases the barrier for contributing to the project. However the code base increased in size which already creates a high barrier for people that aren't experienced enough.

### LINKED ISSUE #5795: Remove Babel from the project
In #5446 (part of #5424) Babel was replaced with TypeScript. However, Babel was not yet removed which still needs to be done.

Remove Babel from the project.

**Additional context**
If Babel is still used somewhere, please share your findings here so that we can discuss the solutions.

christian-bromann:
## Proposed changes

closes #5795 
closes #5424

I took a stab at rewriting the smallest module to see how easy it is. Some things that need to get figured out:

- for some reasons unit test fail because the import [here](https://github.com/webdriverio/webdriverio/blob/master/packages/wdio-cli/src/commands/config.js#L9-L12) resolves in undefined for all
- to access private properties in tests I had to use `repl['_somePrivateIdentifier']` .. not sure if there is a better way to do this
- in our unit tests TypeScript things we use `expect-webdriverio` rather than the actual `expect` framework

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

n/a


packages/wdio-repl/tests/repl.test.ts
```diff
@@ -1,10 +1,15 @@
 import vm from 'vm'
 import replMock from 'repl'
+import expect from 'expect'
```
christian-bromann, line 3: Any idea how I can get rid of this? Without this TypeScript believes we use `expect-webdriverio`. The problem importing `expect` is that it would be redundant and Jest would mock dependencies away that we have defined in this project.

erwinheitzman: I feel like we should exclude the type in that case since apparently it is installed I think?

tsconfig.json
```diff
@@ -1,18 +1,19 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "es2017",
```
erwinheitzman, line 3: This would cause us not able to run browsers like IE11 anymore, is this really what we want?

erwinheitzman: If you want to use newer JavaScript then you need to define a lib instead of changing the output (target)

christian-bromann: The WebdriverIO code only runs on Node.js. We have some scripts in the webdriverio package that we inject into the browser but they should not be compiled.

packages/wdio-repl/tests/repl.test.ts
christian-bromann: How would you exclude it?

packages/webdriverio/tests/utils/interception/devtools.test.js
```diff
@@ -78,7 +78,7 @@ test('abort request', async () => {
         responseHeaders: [{ name: 'Content-Type', value: 'application/json' }]
     }
     const mock = new NetworkInterception('**/foobar/**')
-    expect(::mock.abort).toThrow()
+    expect(mock.abort.bind(mock)).toThrow()
```
erwinheitzman, line 81: I feel like these binds are overkill, especially in a test. Is it not?

christian-bromann: They are necessary otherwise we would loose the scope and the function would not work.

mgrybyk (changes requested):
Let's avoid using ts ignore if possible

packages/wdio-repl/src/index.ts
```diff
@@ -73,71 +81,79 @@ export default class WDIORepl {
...
+            }
         }
 
-        this.replServer = repl.start(this.config)
+        this._replServer = repl.start(this._config)
 
-        return new Promise(
-            (resolve) => this.replServer.on('exit', resolve))
+        return new Promise((resolve) => {
+            if (!this._replServer) {
```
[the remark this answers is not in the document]
christian-bromann: yeah, it can't be defined at initialisation because the interface only starts it when the start method is being executed.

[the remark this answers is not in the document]
christian-bromann: Usually this is not necessary but TypeScript was not happy without checking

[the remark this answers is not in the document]
mgrybyk: This is because replServer can be undefined according to line 53.
Anyway, it's also possible to use 'as'
(this._replServer as ReplServerTypeHere).on(...

christian-bromann:
@mgrybyk @erwinheitzman I addressed all issues and tests should pass now

> * for some reasons unit test fail because the import [here](https://github.com/webdriverio/webdriverio/blob/master/packages/wdio-cli/src/commands/config.js?rgh-link-date=2020-09-07T21%3A53%3A59Z#L9-L12) resolves in undefined for all

This was because of a circular dependency. Good to see that TS catches this.

> * in our unit tests TypeScript things we use `expect-webdriverio` rather than the actual `expect` framework

This problem resolved by itself.

Overall building the project and watching all files has become a bit slower and more CPU intensive. I would like to explore what we can do to improve this.

Please review @webdriverio/technical-steering-committee , if this lands we can release the first TS compiled WebdriverIO version.

christian-bromann:
Some further improvements:

- added a `tsconfig.prod.json` that compiles without source maps
- modified build script: instead of running 38 TS compiler processes it now runs only one with the parameter `-p @wdio/cli @wdio/utils ...`
- moved build and watch step into one file
- removed `.npmrc` file from all packages as we don't use it
- added tsconfig to .npmignore

@mgrybyk mind taking another review?

christian-bromann:
💯 please review @webdriverio/technical-steering-committee

tsconfig.prod.json
```diff
@@ -0,0 +1,23 @@
+{
+  "compilerOptions": {
```
mgrybyk, line 2: Shall sourceMap be set to false here?

christian-bromann: no need, it is false by default

christian-bromann: but happy to add it to make it more clear

.eslintignore
```diff
@@ -2,6 +2,5 @@
 /packages/**/node_modules
 /packages/node_modules
```
erwinheitzman, line 3: I think that `/packages/**/node_modules` already covers `/packages/node_modules`

.eslintrc.js
```diff
@@ -1,44 +1,44 @@
...
+            skipBlankLines: false,
+            ignoreComments: false
+        }],
+        quotes: ['error', 'single', { avoidEscape: true }],
+        'unicode-bom': ['error', 'never'],
+        'object-curly-spacing': ['error', 'always'],
+        'require-atomic-updates': 0
+    },
+    overrides: [{
+        files: ['*.ts', '*.tsx'],
```
erwinheitzman, line 38: we will never use tsx files in this project, will we?

.gitignore
```diff
@@ -66,8 +66,8 @@ tests/typings/**/dist
 /*.js
 allure-reporter.d.ts
 devtools.d.ts
-!babel.config.js
 !.eslintrc.js
+!jest.config.js
```
erwinheitzman, line 70: Why is this done?

CONTRIBUTING.md
```diff
@@ -77,13 +77,13 @@ If you only work on a single package you can watch only for that one by calling:
...
 # e.g. `$ npm run watch:pkg @wdio/runner`
-$ npm run watch:pkg <package-name>
+$ npm run watch <package-name>
 ```
 
 It is also a good idea to run jest in watch mode while developing on a single package to see if changes affect any tests:
 
 ```sh
-$ npx jest ./packages/<package-name>/tests --watch
+$ npx jest ./packages/<package-name>/tests --watch --coverageReporters lcov
```
erwinheitzman, line 86: define the coverageReporters in the config file instead?

scripts/build.js
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env node
```
erwinheitzman, line 1: why is this done over using the package.json files? for new people this adds complexity over using a simple setup that just uses lerna and the package.json files.

.eslintignore
christian-bromann: I assume I added it because it did not. 🤷

.gitignore
christian-bromann: I moved the Jest configurations from the package.json into a `jest.config.js`, given that we ignore all JS files in the root I had to ignore this new config file.

CONTRIBUTING.md
christian-bromann: We do. This just changes the coverage reporter so you don't get to see this coverage list anymore and instead see the test results.

scripts/build.js
christian-bromann: Running 1 TS compiler is much less CPU effective than 38. I tested both variants and this way it just works more efficient. At the end of the day it is the same command for the developer.

.eslintrc.js
christian-bromann: I can remove the `tsx` entry.

scripts/build.js
erwinheitzman: Nice to know, thanks!

mgrybyk (approved):

PR #5805, opened 2020-09-08
Fix Unified Platform and Sauce Connect bug

wswebcreation:
## Proposed changes
Sauce Labs Unified platform is not W3C ready. When using a SC tunnel (or started from the service or using an existing tunnel) the `tunnelIdentifier` was put into `sauce:options`. This PR fixes that issue so customer can use WDIO with UP and SC

## Types of changes
- [ ] Documentation update

## Checklist

### DIFF
--- a/packages/wdio-sauce-service/src/utils.js
+++ b/packages/wdio-sauce-service/src/utils.js
@@ -59,20 +59,21 @@ export function makeCapabilityFactory(tunnelIdentifier, options) {
             !capability['sauce:options']
         )
 
-        if (!capability['sauce:options'] && !isLegacy) {
+        // Unified Platform is currently not W3C ready, so the tunnel needs to be on the cap level
+        if (!capability['sauce:options'] && !isLegacy && !isUnifiedPlatform(capability)) {
             capability['sauce:options'] = {}
         }
 
         Object.assign(capability, options)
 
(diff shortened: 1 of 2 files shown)


christian-bromann (commented):
This test looks odd .. you pass in caps which we check after calling `onPrepare` and in the check the caps still look the same. Even though this is not related to your change but while we are at this piece of code, could you check if this test makes sense?

wswebcreation:
> This test looks odd .. you pass in caps which we check after calling `onPrepare` and in the check the caps still look the same. Even though this is not related to your change but while we are at this piece of code, could you check if this test makes sense?

You're totally correct, looks like all tests that check the caps don't make any sense because we don't have access to the config caps anymore. Should I remove them all?

christian-bromann:
> looks like all tests that check the caps don't make any sense because we don't have access to the config caps anymore.

We do, if you pass in `caps` like this:

```js
await service.onPrepare(config, caps)
```

and modify the passed parameter it will be reflected afterwards as you pass in the variable by reference. What I meant is that the `onPreapre` hook is actually changing these passed in caps but this is not reflected in the test as both values (before/after) are the same.

wswebcreation:
@christian-bromann 

I was focusing on the wrong UT. I've fixed it now and added the device caps tunnel check now to (hopefully) the correct UT

christian-bromann (approved):
Nice 👍

PR #5806, opened 2020-09-08
Check array input parameter using Array.isArray

zabil:
Use Array.isArray for checking so that this API can work
with runners that run in new context of a node vm

Signed-off-by: Zabil Cheriya Maliackal <zabilcm@gmail.com>

## Proposed changes

While using webdriver with other runners like [Gauge](https://github.com/getgauge/gauge-js/issues/387#issuecomment-688772025) the `keys` API fails while checking the instance of an Array parameter. Gauge uses a [Virtual Machine Context](https://nodejs.org/api/vm.html#vm_vm_executing_javascript) and this could be an issue while integrating webdriver with other runners too.

The recommended way around this as per https://github.com/nodejs/node-v0.x-archive/issues/1277#issuecomment-1507839 is to check for if the value is an instance of an Array using `Array.isArray`

This pull request makes that change.
A quick grep shows that this is the only place using this check 

```
git grep -i "instanceof Array" master
master:packages/webdriverio/src/commands/browser/keys.js:    } else if (value instanceof Array) {
```

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/webdriverio/src/commands/browser/keys.js
+++ b/packages/webdriverio/src/commands/browser/keys.js
@@ -35,7 +35,7 @@ export default function keys (value) {
      */
     if (typeof value === 'string') {
         keySequence = checkUnicode(value, this.isDevTools)
-    } else if (value instanceof Array) {
+    } else if (Array.isArray(value)) {
         for (const charSet of value) {
             keySequence = keySequence.concat(checkUnicode(charSet, this.isDevTools))
         }


zabil:
A quick way to check this behavior.

```
Welcome to Node.js v14.7.0.
Type ".help" for more information.
> vm.runInNewContext("a instanceof Array", {a:[]})
false
> vm.runInNewContext("Array.isArray(a)", {a:[]})
true
>
```

christian-bromann (approved):
For a small fix this PR is excellent documented! Very well done 👍 Thank you!

Issue #5796, opened 2020-09-04
browser.mock request payload filter


mgrybyk:
**Is your feature request related to a problem? Please describe.**
I'd like to filter requests by request payload

**Describe the solution you'd like**
```
const exactObject = browser.mock('**/post', {
  request: { 'foo': 'bar' }
})

const exactString = browser.mock('**/post', {
  request: '0123456789'
})

const partial = browser.mock('**/post', {
  request: (payload) => { payload.includes('1234567') }
})
```

**Additional context**
Let's update the [doc](https://webdriver.io/docs/api/browser/mock.html#parameters) as well

PR #5821, opened 2020-09-12
Filter postData, responseHeaders, statusCode

mgrybyk:
## Proposed changes

- add `responseHeaders` to `mock` for verification purposes
- add `requestHeaders` similar to `request` filter that stands for `responseHeaders`
- add `statusCode` filter
- add `postData` filter (exact match, function comparator)

```js
browser.mock('**/api/*', {
  requestHeaders: { foo: 'bar' },
  postData: 'foobar' // exact match
})

browser.mock('**/api/*', {
  method: 'POST',
  postData: data => data && data.includes('foo')
})
```
#5796

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/packages/webdriverio/src/utils/interception/devtools.js
+++ b/packages/webdriverio/src/utils/interception/devtools.js
@@ -15,7 +15,7 @@ export default class DevtoolsInterception extends Interception {
                 headers[name] = value
                 return headers
             }, {})
-            const { requestId, request } = event
+            const { requestId, request, responseStatusCode } = event
 
             for (const mock of mocks) {
                 /**
@@ -25,6 +25,8 @@ export default class DevtoolsInterception extends Interception {
                     continue
                 }
 
+                request.statusCode = responseStatusCode
+
                 /**
                  * match filter options
                  */
@@ -32,7 +34,8 @@ export default class DevtoolsInterception extends Interception {
                     filterMethod(request.method, mock.filterOptions.method) ||
                     filterHeaders(request.headers, mock.filterOptions.requestHeaders) ||
                     filterHeaders(responseHeaders, mock.filterOptions.headers) ||
-                    filterRequest(request.postData, mock.filterOptions.postData)
+                    filterRequest(request.postData, mock.filterOptions.postData) ||
+                    filterStatusCode(responseStatusCode, mock.filterOptions.statusCode)
                 ) {
                     continue
                 }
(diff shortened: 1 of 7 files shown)


packages/webdriverio/src/utils/interception/devtools.js
```diff
@@ -43,6 +42,7 @@ export default class DevtoolsInterception extends Interception {
                     { requestId }
                 ).catch(() => ({}))
 
+                mock.responseHeaders = { ...responseHeaders }
```
mgrybyk, line 45: @christian-bromann is it ok to add `responseHeaders` to `mock` object?

packages/webdriverio/src/utils/interception/devtools.js
```diff
@@ -29,11 +29,10 @@ export default class DevtoolsInterception extends Interception {
                  * match filter options
                  */
                 if (
-                    /**
-                     * HTTP method
-                     */
-                    (mock.filterOptions.method && mock.filterOptions.method.toLowerCase() !== request.method.toLowerCase()) ||
-                    (mock.filterOptions.headers && containsHeaderObject(responseHeaders, mock.filterOptions.headers))
+                    filterMethod(request.method, mock.filterOptions.method) ||
+                    filterHeaders(request.headers, mock.filterOptions.requestHeaders) ||
```
mgrybyk, line 33: I guess I should update `expect().toBeRequestedWith` to accept `requestHeaders` and `responseHeaders ` instead of just `headers`.
I was also thinking about merging headers like this `{ ...requestHeaders, ... responseHeaders }` but I'm afraid it may cause some unexpected issues.
@christian-bromann what do you think?

packages/webdriverio/src/utils/interception/devtools.js
christian-bromann: makes sense to me

packages/webdriverio/src/utils/interception/devtools.js
christian-bromann: > I guess I should update expect().toBeRequestedWith to accept requestHeaders and responseHeaders instead of just headers.

Sounds good to me.

> I was also thinking about merging headers like this { ...requestHeaders, ... responseHeaders } but I'm afraid it may cause some unexpected issues.

Why merging?

mgrybyk: https://github.com/webdriverio/expect-webdriverio/issues/205

packages/webdriverio/src/utils/interception/devtools.js
```diff
@@ -25,15 +25,17 @@ export default class DevtoolsInterception extends Interception {
                     continue
                 }
 
+                request.statusCode = responseStatusCode
```
mgrybyk, line 28: @christian-bromann we can use this one to have `statusCode` check in the `expect`.

christian-bromann: sounds good to me

mgrybyk:
UPD: Added `statusCode` filter

christian-bromann (approved):
👍

PR #5873, opened 2020-09-17
Adds mocha test retry hook to reporter

RimantasDob:
## Proposed changes
Add extra reporter hook to be called if mocha test fails and will be rerun.
Currently, if test is set to have a retry and fails, no `wdio-reporter` hook is executed, leaving user without any knowledge why test fails. Error could be logged in wdio hook afterTest, but it is not called if test timeouts.

## Types of changes

## Checklist

## Further comments

I tested this for mocha only, not sure if it effects Jasmine or Cucumber users.


packages/wdio-reporter/tests/reporter.listeners.test.js
```diff
@@ -540,6 +541,58 @@ describe('WDIOReporter Listeners', () => {
...
+            }
+            testEndEvent = {
+                uid: 'testid'
+            }
+        })
+
+        it('should call test retry with the test stat', () => {
+            reporter.onTestEnd = jest.fn()
+            emitEvents()
+            expect(reporter.onTestEnd).toHaveBeenCalledWith(reporter.tests[testStartEvent.uid])
```
christian-bromann, line 592: shouldn't this check `onTestRetry`? Also could we check if `reporter.retries` was changed too?

christian-bromann (changes requested):
Thanks for the contribution! It would be good to strengthen the unit test a bit.

christian-bromann:
Also would you mind to sign the CLA? Thanks!

packages/wdio-reporter/tests/reporter.listeners.test.js
RimantasDob: Updated unit tests. Added tests for `retries` count, `passes` and `failures` count before and after retry.

RimantasDob:
@christian-bromann I thought I signed CLA, didn't I.
Screenshot 2020-09-23 at 14 12 33

christian-bromann (approved):
Great addition, thanks 👍

PR #5840, opened 2020-09-25
Tracking doc: rewrite `@wdio/logger` package into TypeScript

christian-bromann:

- if you are interested converting source files from `/packages/wdio-logger` into TypeScript please comment on this thread and mention the files you want to convert to avoid duplicate work
- if you make a PR, label it with `TypeScript` and the package name `wdio-logger` so we can easier find all PRs

Happy rewriting 😅


package.json
```diff
@@ -63,7 +63,9 @@
   },
   "devDependencies": {
     "@octokit/rest": "^18.0.0",
+    "@tsconfig/node10": "^1.0.7",
```
erwinheitzman, line 66: A whole dependency for something that seems as simple as the following seems rather overkill to me:
```
{
  "compilerOptions": {
    "lib": ["es2018"],
    "module": "commonjs",
    "target": "es2018",
    "strict": true,
    "esModuleInterop": true,
    "skipLibCheck": true,
    "forceConsistentCasingInFileNames": true
  }  
}
```

packages/wdio-logger/package.json
```diff
@@ -21,6 +21,8 @@
   "bugs": {
     "url": "https://github.com/webdriverio/webdriverio/issues"
   },
+  "types": "./build/node.d.ts",
+  "typeScriptVersion": "3.7.3",
```
erwinheitzman, line 25: why is `typeScriptVersion` needed here?

packages/wdio-cli/src/constants.js
```diff
@@ -1,7 +1,8 @@
-import { version } from '../package.json'
 import { validateServiceAnswers, hasFile } from './utils'
 
-export const CLI_EPILOGUE = `Documentation: https://webdriver.io\n@wdio/cli (v${version})`
+const pkg = require('../package.json')
```
erwinheitzman, line 3: Import wouldn't work here?

packages/wdio-logger/src/node.ts
```diff
@@ -26,37 +26,37 @@ const SERIALIZERS = [{
     /**
      * display error stack
      */
-    matches: (err) => err instanceof Error,
-    serialize: (err) => err.stack
+    matches: (err: any) => err instanceof Error,
```
erwinheitzman, line 29:
```suggestion
    matches: (err: Error) => err instanceof Error,
```

erwinheitzman: Not sure if we can do this but I feel like it might be possible to assign the Error type here

packages/wdio-cli/src/constants.js
christian-bromann: nope, TS error saying that `package.json` is not part of the source files.

packages/wdio-logger/package.json
christian-bromann: it's I think the minimal TS version we support.

package.json
christian-bromann: it seems like this is some sort of TS best practice 🤷

packages/wdio-cli/src/constants.js
mgrybyk: can we overcome it by allowing resolution of json modules? `resolveJsonModule`

packages/wdio-logger/package.json
mgrybyk: btw, it's possible to have type definitions for multiple typescript versions!
ex
```
  "typesVersions": {
    "<=3.4": {
      "*": [
        "ts3.4/*"
      ]
    },
    "<=3.6": {
      "*": [
        "ts3.6/*"
      ]
    }
  },
```
just fyi

packages/wdio-logger/src/node.ts
```diff
@@ -70,7 +70,7 @@ const wdioLoggerMethodFactory = function (methodName, logLevel, loggerName) {
          */
         const match = Object.values(matches).filter(x => args[0].endsWith(`: ${x}`))[0]
         if (match) {
-            const prefixStr = args.shift().slice(0, -match.length - 1)
+            const prefixStr = args.shift()!.slice(0, -match.length - 1)
```
mgrybyk, line 73: how can we be sure `args.shift()` is not undefined?

packages/wdio-logger/src/node.ts
```diff
@@ -131,21 +131,22 @@ export default function getLogger (name) {
  * Calling this prevents part of the logs in the very env to be lost.
  */
 getLogger.waitForBuffer = async () => new Promise(resolve => {
+    // @ts-ignore
```
mgrybyk, line 134: is it really required to have this one?

packages/wdio-logger/tests/node.test.ts
```diff
@@ -1,6 +1,7 @@
 import fs from 'fs'
 import nodeLogger from '../src/node'
 import nodeLogger2 from '../build/node'
+import log from 'loglevel'
```
mgrybyk, line 4: is it possible to import types only?

tsconfig.prod.json
```diff
@@ -1,23 +1,24 @@
...
     "allowJs": true,
     "declaration": true,
     "declarationMap": true,
     "resolveJsonModule": true,
     "removeComments": true,
-    "strict": true,
-    "esModuleInterop": true,
-    "skipLibCheck": true,
-    "forceConsistentCasingInFileNames": true
+    "lib": ["dom"],
```
mgrybyk, line 10: do we really need to have this one? We may miss some errors while trying to interact with some DOM specific API in NodeJS context

mgrybyk (changes requested):

packages/wdio-cli/src/constants.js
christian-bromann: the problem is that TS expects all files you import to be within the files you define in `src`. I don't want to set it to `src: "."` because then we get a folder structure like this:

```
.
└── my-project/
    ├── node_modules/
    │   ├── ...
    │   └── lodash.clonedeep/
    │       ├── package.json
    │       └── index.js
    └── packages/
        └── .../
            └── my-sub-package/
                ├── build/
                │   ├── package.json
                │   └── src/
                │       └── file.js
                ├── src/
                │   └── file.ts
                ├── package.json
                └── ...
```
note the double package.jsons which also causes other TS issues.

packages/wdio-logger/src/node.ts
christian-bromann: yeah, I couldn't find a `writableBuffer` property anywhere in node. It might be a private property and not suppose to be used but we should clean this in a separate PR.

tsconfig.prod.json
christian-bromann: Yes, we need it where we use URL due to a TypeScript bug. People said this would be a workaround until the TS team has fixed the issue.

packages/wdio-logger/src/node.ts
christian-bromann: because `logLevel` always has at least one argument containing the date (even if you do `log.info()`)

packages/wdio-logger/src/node.ts
```diff
@@ -83,7 +83,7 @@ const wdioLoggerMethodFactory = function (methodName, logLevel, loggerName) {
             return arg
         })
 
-        const logText = ansiStrip(`${util.format.apply(this, args)}\n`)
+        const logText = ansiStrip(`${util.format.apply(this, args as [string])}\n`)
```
erwinheitzman, line 86: As discussed before, I think the `as [string]` can go because in reality we pass the entire array

packages/wdio-logger/src/node.ts
```diff
@@ -167,9 +168,9 @@ getLogger.setLogLevelsConfig = (logLevels = {}, wdioLogLevel = DEFAULT_LEVEL) =>
...
          * either apply log level from logLevels object or use global logLevel
          */
-        const logLevel = typeof logLevelsConfig[logLevelName] !== 'undefined' ? logLevelsConfig[logLevelName] : process.env.WDIO_LOG_LEVEL
+        const logLevel = typeof logLevelsConfig[logLevelName] !== 'undefined' ? logLevelsConfig[logLevelName] : process.env.WDIO_LOG_LEVEL as log.LogLevelDesc
 
         loggers[logName].setLevel(logLevel)
     })
 }
-const getLogLevelName = logName => logName.split(':').shift()
+const getLogLevelName = (logName: string): log.LogLevelDesc => logName.split(':').shift() as log.LogLevelDesc
```
erwinheitzman, line 176: The return type isn't needed since it already knows it is going to be of the type log.LogLevelDesc

packages/wdio-logger/src/node.ts
```diff
@@ -26,37 +26,37 @@ const SERIALIZERS = [{
...
 const logCache = new Set()
-let logFile
+let logFile: fs.WriteStream
 
 const originalFactory = log.methodFactory
-const wdioLoggerMethodFactory = function (methodName, logLevel, loggerName) {
-    const rawMethod = originalFactory(methodName, logLevel, loggerName)
-    return (...args) => {
+const wdioLoggerMethodFactory = function (this: log.Logger, methodName: string, logLevel: log.LogLevelDesc, loggerName: string) {
+    const rawMethod = originalFactory(methodName, logLevel as log.LogLevelNumbers, loggerName)
```
erwinheitzman, line 58: I see you casting the logLevel values a lot, I feel like we might need to expand the type of what the logLevel can be.

packages/webdriver/tests/command.test.ts
```diff
@@ -130,24 +158,27 @@ describe('command wrapper result log', () => {
...
-        log.info.mockClear()
+        ;(log.info as jest.Mock).mockClear()
 
         return resultFunction
     }
 
     const takeScreenshotCmd = {
-        command: 'takeScreenshot',
-        endpoint: '/foobar',
+        path: '/foobar',
```
erwinheitzman, line 174: Have we changed functionality here?

tsconfig.prod.json
```diff
@@ -1,23 +1,24 @@
 {
+  "extends": "@tsconfig/node10/tsconfig.json",
```
erwinheitzman, line 2: If this is going to be the code that people are going to use, let's specify a target of `es5` so that WebdriverIO will keep working with older browsers as well 👍

PS: I believe it's `es5` by default but I think it's good to make it explicit for visibility and in order to prevent issues if the default ever changes

scripts/type-generation/webdriver-generate-typings.js
```diff
@@ -51,13 +51,15 @@ for (const [protocolName, definition] of Object.entries(PROTOCOLS)) {
...
 }
 
 const template = fs.readFileSync(TEMPLATE_PATH, 'utf8')
-const outputFile = path.join(__dirname, '..', '..', 'packages', 'webdriver', 'webdriver.d.ts')
-const generatedTypings = EDIT_WARNING + template.replace('// ... insert here ...', lines.join('\n'))
-fs.writeFileSync(outputFile, generatedTypings, { encoding: 'utf-8' })
+const outputFile = path.join(__dirname, '..', '..', 'packages', 'webdriver', 'build', 'types.d.ts')
+const generatedTypings = template.replace('// ... insert here ...', lines.join('\n'))
+
+const origTypings = fs.readFileSync(outputFile, 'utf8')
```
erwinheitzman, line 61:
```suggestion
const origTypings = fs.readFileSync(outputFile, { encoding: 'utf8' })
```

packages/webdriver/src/types.ts
```diff
@@ -240,7 +240,7 @@ export interface Capabilities {
 
 export interface W3CCapabilities {
     alwaysMatch: Capabilities;
-    firstMatch: Capabilities;
+    firstMatch: Capabilities[];
```
erwinheitzman, line 243: Is this a list of a list of capabilities?

packages/webdriver/src/index.ts
```diff
@@ -87,7 +86,7 @@ export default class WebDriver {
      * @param   {Object} instance  the object we get from a new browser session.
      * @returns {string}           the new session id of the browser
     */
-    static async reloadSession (instance: BaseClient) {
+    static async reloadSession (instance: Client): Promise<string> {
```
erwinheitzman, line 89: Return type is not needed as it should recognise the return type automatically?

packages/webdriver/src/command.ts
```diff
@@ -93,7 +92,7 @@ export default function (
         this.emit('command', { method, endpoint, body })
         log.info('COMMAND', commandCallStructure(command, args))
         return request.makeRequest(this.options, this.sessionId).then((result) => {
-            if (result.value != null) {
+            if (typeof result.value !== 'undefined') {
```
erwinheitzman, line 95: Are we sure we don't need to check for `null` anymore?

packages/wdio-repl/package.json
```diff
@@ -22,6 +22,8 @@
   "bugs": {
     "url": "https://github.com/webdriverio/webdriverio/issues"
   },
+  "types": "./build/index.d.ts",
```
erwinheitzman, line 25: Is this build dir also what is available when the package is installed? Because if not, then that part has to be removed from the path

packages/wdio-logger/tests/node.test.ts
```diff
@@ -130,12 +140,14 @@ describe('wdio-logger node', () => {
         const logInfoSpy = jest.spyOn(fs, 'createWriteStream')
         const logCacheAddSpy = jest.spyOn(Set.prototype, 'add')
         const logCacheForEachSpy = jest.spyOn(Set.prototype, 'forEach')
-        let writableBuffer = null
-        logInfoSpy.mockImplementation((path) => ({
-            path,
+        let writableBuffer: any = null
+        logInfoSpy.mockImplementation((path: fs.PathLike): fs.WriteStream => ({
```
erwinheitzman, line 144: Why first assign fs.PathLike and then cast it to a string type when it's only used here?

erwinheitzman (dismissed):
Sorry for all the questions Christian!

jest.config.js
```diff
@@ -36,7 +36,7 @@ module.exports = {
         global: {
             branches: 95,
             functions: 98,
-            lines: 99,
+            lines: 98,
```
mgrybyk, line 39: are you going to increase test coverage with this PR?

packages/wdio-logger/src/node.ts
mgrybyk: do you think it worst adding `args.lengths > 0` to line 72?

packages/wdio-logger/src/node.ts
mgrybyk: tried to overcome it myself, and didn't find any solution. Makes sense!

packages/webdriver/tests/command.test.ts
christian-bromann: nope, just changed the structure a bit

tsconfig.prod.json
christian-bromann: > so that WebdriverIO will keep working with older browsers as well 👍

This code is not run in the browser 😛  WebdriverIO will continue to work given that automation is not run by JS.

scripts/type-generation/webdriver-generate-typings.js
christian-bromann: If options is passed as a string it is automatically used to specify encoding (see https://nodejs.org/docs/latest-v12.x/api/fs.html#fs_fs_readfile_path_options_callback)

packages/webdriver/src/types.ts
christian-bromann: yes

packages/wdio-repl/package.json
christian-bromann: Yes, the NPM package will only contain a `build` dir.

packages/wdio-logger/tests/node.test.ts
christian-bromann: this is how types are defined in Node for some reason, e.g.
```ts
function createWriteStream(path: fs.PathLike, ....)
```
Whereas `WriteStream` is defined as
```ts
    class WriteStream extends stream.Writable {
        close(): void;
        bytesWritten: number;
        path: string | Buffer;
```

tsconfig.prod.json
erwinheitzman: But some commands are injected into the browser and if we output es6 or above it will break unless I am missing something

christian-bromann: None of the scripts have been targeted to ES5 when [Babel was around](https://github.com/webdriverio/webdriverio/blob/v6.3.7/babel.config.js). None of them use some specific ESNext functionality. I feel confident that they continue to work even if testing on IE6.

christian-bromann:
@mgrybyk @erwinheitzman thanks for review! Anything left that I should address?

mgrybyk:
Can you please merge the latest master? Let's make sure there are no conflicts eith shared store  service because there is logger stub.

Can you please remove it?

Let's wait for tests to pass as well

christian-bromann:
> Can you please merge the latest master?

Done

> Can you please remove it?

Done

packages/wdio-logger/src/node.ts
christian-bromann: makes sense!

christian-bromann: 🤔 mhm even after changing it to:

```ts
        if (match && args[0]) {
            const prefixStr = args.shift().slice(0, -match.length - 1)
            args.unshift(prefixStr, match)
        }
```

it still fails.

mgrybyk: sorry for confusing, didn't expect it's not that smart, let's leave it with `!` then
```
if (match) {
            const prefixStr = args.shift()!.slice(0, -match.length - 1)
```

mgrybyk (approved):
👍 
let's merge once tests are passing.
Thank you!

christian-bromann:
@mgrybyk can you check the CI/CD errors?

```
Summary of all failing tests
FAIL packages/wdio-shared-store-service/tests/service.test.ts
  ● Test suite failed to run
    packages/wdio-shared-store-service/src/service.ts:17:48 - error TS2339: Property 'browser' does not exist on type 'Global'.
    17                 value: (key: string) => global.browser.call(() => getValue(key))
                                                      ~~~~~~~
    packages/wdio-shared-store-service/src/service.ts:20:111 - error TS2339: Property 'browser' does not exist on type 'Global'.
    20                 value: (key: string, value: WebdriverIO.JsonCompatible | WebdriverIO.JsonPrimitive) => global.browser.call(() => setValue(key, value))
                                                                                                                     ~~~~~~~
    packages/wdio-shared-store-service/src/service.ts:24:16 - error TS2339: Property 'browser' does not exist on type 'Global'.
    24         global.browser.sharedStore = sharedStore
```

Seems to be related with the rewrite in `@wdio/shared-store-service` any idea why the namespace overwrite can't be found?

mgrybyk:
@christian-bromann I'm getting lots of errors locally in your branch

here is only some of them

christian-bromann:
> tsconfig.prod.json(2,14): error TS6053: File '@tsconfig/node10/tsconfig.json' not found.

I think you missed to install NPM depdendencies.

mgrybyk:
@christian-bromann you are right, I still get one error though
```
packages/wdio-cli/src/constants.js(3,21): error TS6059: File '/Users/m/sources/wdio/webdriverio/packages/wdio-cli/package.json' is not under 'rootDir' '/Users/m/sources/wdio/webdriverio/packages/wdio-cli/src'. 'rootDir' is expected to contain all source files.
```

christian-bromann:
🎉

PR #5903, opened 2020-09-30
Fix double click

mgrybyk:
## Proposed changes

Issue: if there is a need to open an alert with a double click, the alert closes immediately because `releaseActions` fails due to "unexpected alert" error.

Do not call release actions after double click.

Calling `releaseActions` command is only needed to release previously pressed buttons, in case of double click, it's not needed because we do it within `performAction` actions.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

updated "type generate" script.


christian-bromann (approved):
👍

PR #5905, opened 2020-09-30
browser mock comparator functions

mgrybyk:
## Proposed changes

- all browser filter options support comparator function
- ignore case of `Content-Type` header

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

mention: @rickschubert

### DIFF
--- a/packages/webdriverio/tests/utils/interception/devtools.test.js
+++ b/packages/webdriverio/tests/utils/interception/devtools.test.js
@@ -156,7 +156,7 @@ test('decodes base64 responses', async () => {
     const mock = new NetworkInterception('**/foobar/**')
     await NetworkInterception.handleRequestInterception(cdpClient, [mock])({
         request: { url: 'http://test.com/foobar/test.html' },
-        responseHeaders: [{ name: 'Content-Type', value: 'application/json' }]
+        responseHeaders: [{ name: 'content-Type', value: 'application/json' }]
     })
     await NetworkInterception.handleRequestInterception(cdpClient, [mock])({
         request: { url: 'http://test.com/foobar/test.html' },
--- a/packages/webdriverio/tests/utils/interception/devtools.test.js
+++ b/packages/webdriverio/tests/utils/interception/devtools.test.js
@@ -26,7 +26,7 @@ test('allows to access network calls', async () => {
     expect(mock.calls.length).toBe(1)
 })
 
-describe('filter by header', () => {
+describe('filter network calls by header', () => {
     const mockWithCall = async (filter) => {
         const mock = new NetworkInterception('**/foobar/**', filter)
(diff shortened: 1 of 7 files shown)


packages/webdriverio/tests/utils/interception/devtools.test.js
```diff
@@ -156,7 +156,7 @@ test('decodes base64 responses', async () => {
     const mock = new NetworkInterception('**/foobar/**')
     await NetworkInterception.handleRequestInterception(cdpClient, [mock])({
         request: { url: 'http://test.com/foobar/test.html' },
-        responseHeaders: [{ name: 'Content-Type', value: 'application/json' }]
+        responseHeaders: [{ name: 'content-Type', value: 'application/json' }]
```
rickschubert, line 159: I noticed that you changed the capitalisation of the header here. Given that all headers are not case-sensitive, should this be `content-type` then?

rickschubert (approved):
Wow, so fast, this is amazing!

packages/webdriverio/tests/utils/interception/devtools.test.js
rickschubert, line 159: I assume this should be `content-type`, given that the case changed in the other place as well?

rickschubert:
[Looking at these lines](https://github.com/webdriverio/webdriverio/blob/64b24e6e77e34e3b2af99231c38701015c4c99bb/packages/webdriverio/src/utils/interception/devtools.js#L38), it seems to me like only one of the comparator functions needs to return true. Could we change it so that all the comparator functions the user adds need to return true?

Looking at this block I am also wondering why it works, given that all the comparator functions return `false` when there is no filter expectation given. Or am I having a thinking error in my mind? I would have expected that it needs to look like below:

```js
                if (
                    !filterMethod(request.method, mock.filterOptions.method) ||
                    !filterHeaders(request.headers, mock.filterOptions.requestHeaders) ||
                    !filterHeaders(responseHeaders, mock.filterOptions.headers) ||
                    !filterRequest(request.postData, mock.filterOptions.postData) ||
                    !filterStatusCode(responseStatusCode, mock.filterOptions.statusCode)
                ) {
                    continue
                }
```

Additionally, I think that each of the filter functions should return `true` if no expectation is given.

packages/webdriverio/tests/utils/interception/devtools.test.js
mgrybyk: It's test, wanted to have this one in mixed case

christian-bromann (approved):
👍

packages/webdriverio/tests/utils/interception/devtools.test.js
rickschubert: Oh, of course, thank you!

Issue #5844, opened 2020-09-14
Tracking doc: rewrite `@wdio/selenium-standalone-service` package into TypeScript


christian-bromann:
This is a tracking issue for organising the effort of re-writing the code base from vanilla JS into TypeScript. If you are reading this and are interested getting involved, this is a great chance to do so. Here are the rules:

- if you are interested converting source files from `/packages/wdio-selenium-standalone-service` into TypeScript please comment on this thread and mention the files you want to convert to avoid duplicate work
- there is no reason to convert the whole package at once - smaller PRs make it easier to review the code
- if you make a PR, reference this issue so we can keep track of the progress
- if you make a PR, label it with `TypeScript` and the package name `wdio-selenium-standalone-service` so we can easier find all PRs
- if you make a PR, assign it to the `TypeScript Rewrite` project

Happy rewriting 😅

suniljaiswal01:
@christian-bromann  I would like to give a try in "wdio-selenium-standalone-service" package. Can you provide me with some details on what all files has to be changed, will there be any config changes also requried and how can we test if the changes are working fine or not.

Also, can you point me if any of the package is already converted in typescript, that i can use for reference

christian-bromann:
> I would like to give a try

Awesome, thanks!

> Can you provide me with some details on what all files has to be changed

All JS files in `/packages/wdio-selenium-standalone-service` (source and test files)

> how can we test if the changes are working fine or not.

If all tests pass we should be good to go.

> Also, can you point me if any of the package is already converted in typescript, that i can use for reference

Let's wait until https://github.com/webdriverio/webdriverio/pull/5804 is merged. It will have the full TS setup for WebdriverIO and also I rewrote a package in there (`@wdio/repl`).

christian-bromann:
@suniljaiswal01 #5804 has been merged feel free to start working on this.

suniljaiswal01:
@christian-bromann  I am almost completed with the migration, just have few blockers and doubt.
1. @wdio/logger dependency is present in this package, and i can see that there is already a PR raised for logger package. So should i wait for that package to be merge and raise PR for my code?
2. Once the migration is completed, how can i test if the script is properly migrated?. Will running the test present inside the packages will be enough
3. For the below line i am geeting issue and not able to fix it

`await promisify(SeleniumStandalone.install)(this.installArgs)`

Compile Error that i am facing:

[...]

PR #5913, opened 2020-10-01
Tracking doc: rewrite `@wdio/selenium-standalone-service` package into TypeScriptTs selenium standlone service

suniljaiswal01:
## Proposed changes

if you are interested converting source files from /packages/wdio-selenium-standalone-service into TypeScript please comment on this thread and mention the files you want to convert to avoid duplicate work
there is no reason to convert the whole package at once - smaller PRs make it easier to review the code
if you make a PR, reference this issue so we can keep track of the progress
if you make a PR, label it with TypeScript and the package name wdio-selenium-standalone-service so we can easier find all PRs
if you make a PR, assign it to the TypeScript Rewrite project

#5844 

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

There are still some issue which i need some help in resolving


packages/wdio-selenium-standalone-service/src/launcher.ts
```diff
@@ -48,11 +48,11 @@ export default class SeleniumStandaloneLauncher {
         this.skipSeleniumInstall = Boolean(options.skipSeleniumInstall)
     }
 
-    async onPrepare(config: Config) :Promise<void> {
+    async onPrepare(config: Config): Promise<void> {
         this.watchMode = Boolean(config.watch)
 
         if (!this.skipSeleniumInstall) {
-            await promisify(SeleniumStandalone.install)(this.installArgs)
+            await promisify(SeleniumStandalone.install.bind(this.installArgs))
```
suniljaiswal01, line 55: could someone help me on what should be used here.

mgrybyk: I think it should not be changed

suniljaiswal01: If not changed, then i am getting the below error:

`(property) SeleniumStandaloneLauncher.installArgs: Partial<SeleniumStandalone.InstallOpts> Expected 0 arguments, but got 1.ts(2554)`

mgrybyk: you can define `installSeleniumStandalone = promisify(SeleniumStandalone.install)`
it should have the same type as `SeleniumStandalone.install` but result should be wrapped with `Promise<>`

packages/wdio-selenium-standalone-service/src/launcher.ts
```diff
@@ -1,5 +1,8 @@
 import logger from '@wdio/logger'
 
+declare module 'util' {
+    function promisify<T1, TResult>(fn: (arg1: T1, callback: (err: any, result: TResult) => void) => void): (arg1: T1) => Promise<TResult>;
+}
```
christian-bromann, line 5: Why is this necessary?

packages/wdio-selenium-standalone-service/src/launcher.ts
```diff
@@ -16,16 +19,36 @@ const DEFAULT_CONNECTION = {
...
+    outputDir?: string,
+    watch?: boolean,
+}
+
+export interface SeleniumStandaloneOptions {
+    logPath?: string;
+    installArgs?: any;
+    args?: any;
+    skipSeleniumInstall?: boolean;
+}
```
christian-bromann, line 32: add new line
```suggestion
}

```

packages/wdio-selenium-standalone-service/src/launcher.ts
```diff
@@ -16,16 +19,36 @@ const DEFAULT_CONNECTION = {
...
 export default class SeleniumStandaloneLauncher {
-    constructor (options, capabilities, config) {
+
```
christian-bromann, line 34:
```suggestion
```

packages/wdio-selenium-standalone-service/src/launcher.ts
```diff
@@ -16,16 +19,36 @@ const DEFAULT_CONNECTION = {
...
+    capabilities: any
```
christian-bromann, line 35: We should use `Capabilities` from the webdriver package

christian-bromann (changes requested):
Some comments

packages/wdio-selenium-standalone-service/src/launcher.ts
suniljaiswal01: There seems to be some issue with promisify overloading , similar issue is mentioned here :
https://github.com/DefinitelyTyped/DefinitelyTyped/issues/20497

By using the above code, we are specifying which method to overload instead of the generic one

mgrybyk:
@suniljaiswal01 please see https://github.com/webdriverio/webdriverio/compare/ts-selenium-standlone-service-fix-types

Based on my comments I've applied some fixes

packages/wdio-selenium-standalone-service/package.json
```diff
@@ -42,6 +41,8 @@
   },
   "types": "selenium-standalone-service.d.ts",
   "devDependencies": {
-    "@types/fs-extra": "^9.0.1"
+    "@types/fs-extra": "^9.0.1",
+    "webdriver": "^6.6.2",
+    "webdriverio": "^6.6.2"
```
christian-bromann, line 46:
```suggestion
    "webdriver": "6.6.2",
    "webdriverio": "6.6.2"
```

packages/wdio-selenium-standalone-service/src/@types/webdriverio.d.ts
```diff
@@ -0,0 +1,5 @@
+declare module WebdriverIO {
+    interface Config {
+        watch?: boolean
```
christian-bromann, line 3: Can we move this to the original typings?

packages/wdio-selenium-standalone-service/package.json
christian-bromann: Also if we only import types from these packages we don't need to declare them as dev dependencies. Given the project setup, TS will find the packages.

packages/wdio-selenium-standalone-service/src/@types/webdriverio.d.ts
suniljaiswal01: @mgrybyk  added this in his commit. Should i remove it?

christian-bromann: The `watch` property in the config isn't something that is specific to the service, so I would say yes.

mgrybyk:
Don't put watch to Config interface of webdriverio!
We should not allow to set watch option in the config.
@christian-bromann we can have a fix for it in another PR, I'd leave it as is for now.

packages/wdio-selenium-standalone-service/package.json
mgrybyk: Yes! Let's remove both webdriver and webdriverio from here, nice catch!

mgrybyk (changes requested):

packages/wdio-selenium-standalone-service/src/@types/webdriverio.d.ts
mgrybyk: We should think on how to keep it internal only and don't expose to the user.
At least disallow to define it in the config

mgrybyk: Otherwise it might confuse a user that the option does nothing
https://github.com/webdriverio/webdriverio/blob/e34108325ef1e0d2b8e2306fe407b5d934169777/packages/wdio-config/src/lib/ConfigParser.js#L73

mgrybyk: We can also define it in webdriverio tpl config like this `readonly watch?: never` and add a comment that it's used for internal purposes only.
@christian-bromann

packages/wdio-selenium-standalone-service/.eslintrc.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+    globals: {
+        WebdriverIO: true,
+        WebDriver: true,
+    }
+}
```
christian-bromann, line 6: This file can be removed

packages/wdio-selenium-standalone-service/src/@types/webdriverio.d.ts
christian-bromann: sounds good to me

christian-bromann:
if the last 2 outstanding issues are resolved, this will be good to go 👍

suniljaiswal01:
> if the last 2 outstanding issues are resolved, this will be good to go 👍

one is removing the .eslintrc.js file, which is the second? this one:
https://github.com/webdriverio/webdriverio/pull/5913#discussion_r503263687

packages/wdio-selenium-standalone-service/.eslintrc.js
suniljaiswal01: Done.

suniljaiswal01: can you tell me what does this file do?

mgrybyk:
If you are not getting any errors with eslint file removed - go ahead. Added it because I had some issues with linter

packages/wdio-selenium-standalone-service/src/@types/webdriverio.d.ts
mgrybyk: Second one, I guess

christian-bromann: Yes, this is an internal flag and only set if you run:

```sh
$ npx wdio wdio.conf.js --watch
```

as the `watch` flag will be merged into the config.

christian-bromann:
@suniljaiswal01 correct, instead of defining custom WebdriverIO types, add the `watch` property to the WebdriverIO.Config definition.

suniljaiswal01:
Thank you for the help @christian-bromann @mgrybyk . Let me know if anything else need to be change

scripts/templates/webdriverio.tpl.d.ts
```diff
@@ -633,7 +633,9 @@ declare namespace WebdriverIO {
         // ... browser commands ...
     }
 
-    interface Config extends Options, Omit<WebDriver.Options, "capabilities">, Hooks {}
+    interface Config extends Options, Omit<WebDriver.Options, "capabilities">, Hooks {
+        watch?: boolean
```
mgrybyk, line 637:
```suggestion
         /**
         * internal usage only. To run in watch mode see https://webdriver.io/docs/watcher.html
         */
        watch?: never;
```

mgrybyk (changes requested):
Got one minor comment, other than that LGTM.

Thanks a lot for your hard work and patience!

scripts/templates/webdriverio.tpl.d.ts
mgrybyk: User should not be able to set it in `wdio.conf.js`
image

suniljaiswal01: using never type for watch property is failing the execution, we are using `await launcher.onPrepare({ watch: true })` in test file

suniljaiswal01: changing `await launcher.onPrepare({ watch: true })`  to `await launcher.onPrepare({ watch: true } as never)` is fixing the failing test. is this ok?

mgrybyk: Fine for me

christian-bromann (approved):
👍 thank you!

@mgrybyk any outstanding issues?

mgrybyk (approved):

christian-bromann:
Great job @suniljaiswal01 !

PR #5919, opened 2020-10-02
Improve documentation around WDIO options/configs

jrouly:
## Proposed changes

Improve documentation around WDIO configs and options.

## Types of changes

- [x] Documentation update

## Checklist

## Further comments

This change was inspired by confusion around the use of `browser.config` and `browser.options` (the latter of which I could not find documented anywhere).

Here's the behavior I've observed when a `browser.config` and `browser.option` key conflict in name:

| set in `wdio.conf.js` | set on CLI | `wdio.config` value | `wdio.options` value |
| --------------------- | ---------- | ------------------- | -------------------- |
| :white_check_mark: | :white_check_mark: | cli value | cli value |
| :white_check_mark: | :x: | `wdio.conf.js` value | `wdio.conf.js` value |
| :x: | :white_check_mark: | cli value | cli value |
| :x: | :x: | `undefined` | `undefined` |

which is why I added the sentence describing precedence.

Additionally, I couldn't find `browser.options` documented anywhere, so I added it on the Options page.

I feel these are most likely uncontentious improvements, but I'm very open to feedback.

### DIFF
--- a/docs/BrowserObject.md
+++ b/docs/BrowserObject.md
@@ -52,9 +52,9 @@ console.log(browser.requestedCapabilities)
  */
 ```
 
-## Get Config Options
+## Custom Configurations
 
-If using the WDIO testrunner you can always define custom options within your WDIO config:
+If using the WDIO testrunner you can always define custom configurations within your WDIO config:
 
 ```js
 // wdio.conf.js
@@ -89,6 +89,18 @@ console.log(browser.config)
 
 console.log(browser.config.fakeUser) // outputs: "maxmustermann"
 ```
+
+### Versus Options
+
+Custom configurations should not be confused with [Options](Options.md), which are documented separately:
+
+```js
+browser.config.baseUrl; // Custom configuration
+browser.options.baseUrl; // WebdriverIO option
+```
+
+When using the WDIO testrunner, if keys conflict in name (e.g. `baseUrl` above) the option value will take precedence and overwrite the config value.
+
 
 ## Mobile Flags
 
--- a/docs/BrowserObject.md
+++ b/docs/BrowserObject.md
@@ -101,7 +101,6 @@ browser.options.baseUrl; // WebdriverIO option
 
 When using the WDIO testrunner, if keys conflict in name (e.g. `baseUrl` above) the option value will take precedence and overwrite the config value.
 
-
 ## Mobile Flags
 
 If you need to modify your test based on whether or not your session runs on a mobile device, you can access the mobile flags to check.
--- a/docs/BrowserObject.md
+++ b/docs/BrowserObject.md
@@ -62,6 +62,7 @@ exports.config = {
     // ...
     fakeUser: 'maxmustermann',
     fakePassword: 'foobar',
+    baseUrl: 'example.com',
     // ...
 }
 ```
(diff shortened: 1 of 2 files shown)


CMDadabo (approved):
👏

sonamoo (approved):

docs/BrowserObject.md
```diff
@@ -52,9 +52,9 @@ console.log(browser.requestedCapabilities)
  */
 ```
 
-## Get Config Options
+## Custom Configurations
```
jrouly, line 55: If changing the name is an issue due to breaking existing anchor links, I can revert this change. I think it's valuable, though, because `Config Options` just _adds_ to the confusion between Options and Configurations.

docs/BrowserObject.md
```diff
@@ -90,6 +90,18 @@ console.log(browser.config)
...
+Custom configurations should not be confused with [Options](Options.md), which are documented separately:
+
+```js
+browser.config.baseUrl; // Custom configuration
+browser.options.baseUrl; // WebdriverIO option
+```
+
+When using the WDIO testrunner, if keys conflict in name (e.g. `baseUrl` above) the option value will take precedence and overwrite the config value.
+
+
```
christian-bromann, line 104:
```suggestion
```

docs/BrowserObject.md
christian-bromann: I think this should be fine, can you find any references to this heading?

docs/BrowserObject.md
```diff
@@ -90,6 +90,18 @@ console.log(browser.config)
 console.log(browser.config.fakeUser) // outputs: "maxmustermann"
 ```
 
+### Versus Options
+
+Custom configurations should not be confused with [Options](Options.md), which are documented separately:
+
+```js
+browser.config.baseUrl; // Custom configuration
+browser.options.baseUrl; // WebdriverIO option
```
christian-bromann, line 99: This code example doesn't really make it clear for me, can you enhance it with some actual values and a snippet of a `wdio.conf.js`?

christian-bromann (changes requested):

docs/BrowserObject.md
jrouly: Yep absolutely.

docs/BrowserObject.md
jrouly: Nope, not in the docs. I'd just be concerned about _external_ references to this anchor (e.g. bookmarks etc.)

docs/BrowserObject.md
jrouly: I went ahead and added a few code snippets and clarified some of the language.

docs/BrowserObject.md
```diff
@@ -82,14 +83,51 @@ console.log(browser.config)
...
+```
+
+```bash
+# testrunner invocation
+$ npm wdio wdio.conf.js --baseUrl foobar.com
+```
+
+```js
+console.log(browser.config.baseUrl) // 'foobar.com'
+console.log(browser.options.baseUrl) // 'foobar.com'
```
christian-bromann, line 128: shouldn't these values be different given that the one "will take precedence and overwrite the config value"?

jrouly: This is the result of that behavior.

In the previous two snippets, I set one to `example.com` and one to `foobar.com`. But both resolve to `foobar.com` during test execution.

christian-bromann: > Maybe I should make that more clear.

My initial question was, how would I get the value that was passed in as wdio parameter

docs/BrowserObject.md
```diff
@@ -82,14 +83,51 @@ console.log(browser.config)
         connectionRetryTimeout: 120000,
         connectionRetryCount: 3,
         specs: [ 'err.js' ],
-        fakeUser: 'maxmustermann', // <-- custom option
-        fakePassword: 'foobar', // <-- custom option
+        fakeUser: 'maxmustermann', // <-- custom configuration
+        fakePassword: 'foobar', // <-- custom configuration
+        baseUrl: 'example.com', // <-- custom configuration
```
christian-bromann, line 88: `baseUrl` is already defined a bit more up and also is not a custom configuration.

christian-bromann (commented):
just discovered this one

docs/BrowserObject.md
jrouly: Ah good catch I missed that `baseUrl`, will remove the duplicate.

> is not a custom configuration

Well, is it not? I understand it's a "reserved" key in `browser.options`, so it's not "custom" in that sense. But it's still defined alongside other configurations in `wdio.conf.js` with a custom value. And to me, it's confusing when to reference `browser.options` and when to reference `browser.config` when the same key is defined in both.

christian-bromann: To be honest, I am not quite sure why we still have both. It is ambiguous. I will create an issue getting rid of `browser.config`.

christian-bromann (commented):
Let's remove the duplicate `baseUrl` key and move forward. This issue has shown that having two properties like this is ambiguous and we should get rid of one.

jrouly:
@christian-bromann thanks so much for the thoughtful review. Agree it's ambiguous, looking forward to that improvement. Hopefully these docs make it _slightly_ more explicit in the interim? That's my goal at least.

christian-bromann (approved):
Thank you for extended these docs.

LGTM 👍

Issue #5915, opened 2020-10-01
Unable to run Puppeteer from latest Webdriver.io v6.5.2


jamieni:
Reopening issue #5697 on Webdriver.io v6.5.2

I used the same example here: https://github.com/webdriverio/webdriverio/blob/master/examples/devtools/intercept.js
and received this error: `puppeteerBrowser.pages() is not a function`

pmerwin:
I am getting this with the latest:

2020-10-03T19:31:53.810Z ERROR @wdio/utils:initialiseServices: Error: Couldn't initialise "@wdio/devtools-service".
TypeError: Cannot read property 'requestLatencyMs' of undefined

mgrybyk:
@jamieni please change https://github.com/webdriverio/webdriverio/blob/3867103c9c5af0c83ee933b0c75ae8d9af0c73e4/examples/devtools/intercept.js#L8 to `const { remote } = require('webdriverio');`

I'll update the example

PR #5926, opened 2020-10-03
Fix intercept example

mgrybyk:
## Proposed changes

Fix intercept example #5915

- fixed require string
- removed `devtools` as far as `webdriverio` has `devtools` in dependencies.

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/examples/devtools/intercept.js
+++ b/examples/devtools/intercept.js
@@ -1,11 +1,11 @@
 /**
- * in order to run this file make sure you have `webdriverio` and `devtools`
+ * in order to run this file make sure you have `webdriverio`
  * installed using NPM before running it:
  *
- *   $ npm install devtools webdriverio
+ *   $ npm install webdriverio
  *
  */
-const { remote } = require('../../packages/webdriverio')
+const { remote } = require('webdriverio')
 
 let browser;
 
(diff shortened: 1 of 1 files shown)


examples/devtools/intercept.js
```diff
@@ -1,11 +1,11 @@
 /**
- * in order to run this file make sure you have `webdriverio` and `devtools`
```
christian-bromann, line 2: This example should also work without using the devtools protocol to run the test in. With adding the network interceptor functionalities we now initialise Puppeteer also when running local Chromedriver tests.

mgrybyk: Let's create another example for webdriver protocol?

christian-bromann: it would be the same file just with Chromedriver running in the background. I just tested it and found that we should also change:

```patch
+ const puppeteerBrowser = await browser.getPuppeteer()
- const puppeteerBrowser = browser.getPuppeteer()
```

as well as:

```patch
    browser = await remote({
-       automationProtocol: 'devtools',
        capabilities: {
            browserName: 'chrome'
        }
    })
```

mgrybyk: applied the suggested changes

christian-bromann (approved):
👍

PR #5927, opened 2020-10-05
Tracking doc: rewrite `@wdio/shared-store-service` package into TypeScript

mgrybyk:
## Proposed changes

rewrite `@wdio/shared-store-service` package into TypeScript #5845

## Types of changes

- [x] internal

## Checklist


packages/wdio-shared-store-service/src/server.ts
```diff
@@ -1,7 +1,19 @@
-const polka = require('polka')
+// polka doesn't have up to date types
+// and the project is not maintained for 2 years
+const polka: Polka = require('polka')
 const { json } = require('@polka/parse')
```
christian-bromann, line 6: can we use import here?

mgrybyk:
ready for review

packages/wdio-shared-store-service/src/server.ts
mgrybyk: fixed

christian-bromann (approved):
👍

PR #5928, opened 2020-10-05
Added possibility to disable automatic strings to unicode translation

IgorSasovets:
## Proposed changes
This is a fix for wdio automatic translation to [unicode issue](https://github.com/webdriverio/webdriverio/issues/4936). In general, user can't send keys that are predefined in the UNICODE_CHARACTERS [list](https://github.com/webdriverio/webdriverio/blob/master/packages/webdriverio/src/constants.js).

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/scripts/templates/webdriverio.tpl.d.ts
+++ b/scripts/templates/webdriverio.tpl.d.ts
@@ -636,6 +636,6 @@ declare namespace WebdriverIO {
     interface Config extends Options, Omit<WebDriver.Options, "capabilities">, Hooks {}
 
     interface AddValueOptions {
-        translateToUnicode: boolean
+        translateToUnicode?: boolean
     }
 }
--- a/scripts/templates/webdriverio.tpl.d.ts
+++ b/scripts/templates/webdriverio.tpl.d.ts
@@ -634,4 +634,8 @@ declare namespace WebdriverIO {
     }
 
     interface Config extends Options, Omit<WebDriver.Options, "capabilities">, Hooks {}
+
+    interface AddValueOptions {
+        translateToUnicode: boolean
+    }
 }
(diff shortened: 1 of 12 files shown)


christian-bromann (changes requested):
`AddValueOptions` needs to be added [here](https://github.com/webdriverio/webdriverio/blob/master/scripts/type-generation/constants.js)

scripts/templates/webdriverio.tpl.d.ts
```diff
@@ -634,4 +634,8 @@ declare namespace WebdriverIO {
     }
 
     interface Config extends Options, Omit<WebDriver.Options, "capabilities">, Hooks {}
+
+    interface AddValueOptions {
+        translateToUnicode: boolean
```
mgrybyk, line 639: Should be optional

mgrybyk (dismissed):
Please run types tests locally and commit generated d.ts files

scripts/templates/webdriverio.tpl.d.ts
IgorSasovets: Fixed

IgorSasovets:
@christian-bromann , @mgrybyk , added requested changes to the PR

christian-bromann (commented):
One last thing I've just realised: we are using `AddValueOptions` for `setValue` which seems a bit mis-worded. Can we have a generalised name for it that fits to set and addValue?

IgorSasovets:
@christian-bromann , I think yes but we should take into an account that `setValue` command uses `addValue` under the hood. So, does it really make sense to rename it? Lastly, it is hard to pick up some general name for this parameter. I would name it `changeValueOptions` because in both cases we change the value of a selected element. Please share your thoughts and suggestions

christian-bromann (approved):
> I think yes but we should take into an account that setValue command uses addValue under the hood. So, does it really make sense to rename it?

Very good point, let's keep as is. Thank you!

LGTM 👍

IgorSasovets:
@christian-bromann , thanks a lot!

PR #5929, opened 2020-10-05
Added a new param inject (function) to url command

AutomationReddy:
## Proposed changes

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments


packages/webdriverio/tests/commands/browser/url.test.js
```diff
@@ -27,14 +27,39 @@ describe('url', () => {
...
             expect(e.message).toContain('command needs to be type of string')
         }
     })
 
+    it('should throw an exception when a non-function value passed in inject', async () => {
+        try {
+            await browser.url('/', { inject: 123 })
+        } catch (e) {
+            expect(e.message).toContain('command needs to be type of function')
+        }
```
christian-bromann, line 43: better:

```suggestion
        const err = await browser.url('/', { inject: 123 }).catch((err) => err)
        expect(e.message).toContain('command needs to be type of function')
```

packages/webdriverio/tests/commands/browser/url.test.js
```diff
@@ -27,14 +27,39 @@ describe('url', () => {
...
+    })
+
+    it('should continue the navigation if inject parameter is not passed', async () => {
```
christian-bromann, line 46: this test needs a check that puppeteer wasn't used

packages/webdriverio/tests/commands/browser/url.test.js
```diff
@@ -27,14 +27,39 @@ describe('url', () => {
...
+        await browser.url('/')
+        expect(got.mock.calls[0][1].json).toEqual({ url: 'http://foobar.com/' })
+    })
+
+    it('should accept an arrow function as the inject parameter', async () => {
+        const arrowFunction = jest.fn(() => console.log('foo-bar'))
+        await browser.url('/', arrowFunction)
```
christian-bromann, line 53:
```suggestion
        await browser.url('/', { inject: arrowFunction })
```

packages/webdriverio/tests/commands/browser/url.test.js
```diff
@@ -27,14 +27,39 @@ describe('url', () => {
...
+        expect(got.mock.calls[0][1].json).toEqual({ url: 'http://foobar.com/' })
```
christian-bromann, line 54: let's also add a test that we used puppeteer to inject the method

packages/webdriverio/tests/commands/browser/url.test.js
```diff
@@ -27,14 +27,39 @@ describe('url', () => {
...
+    })
+
+    it('should accept a normal function as the inject parameter', async () => {
+        const arrowFunction = jest.fn(function () { console.log('foo-bar') })
+        await browser.url('/', arrowFunction)
```
christian-bromann, line 59: same

christian-bromann (changes requested):
some comments on typescript generation and unit tests

packages/webdriverio/tests/commands/browser/url.test.js
AutomationReddy: `expect(e.message).toContain('command needs to be type of function')` to `        expect(err.message).toContain('command needs to be type of function')
`
I will add this

packages/webdriverio/tests/commands/browser/url.test.js
AutomationReddy: @christian-bromann I believe I need help in this. Help in how to approach is fine.

packages/webdriverio/tests/commands/browser/url.test.js
AutomationReddy: Same here - What would be the way to check this. Sorry, new to unit tests.

packages/webdriverio/tests/commands/browser/url.test.js
```diff
@@ -27,14 +27,39 @@ describe('url', () => {
...
+
+    it('should accept an arrow function as the inject parameter', async () => {
+        const arrowFunction = jest.fn(() => console.log('foo-bar'))
+        await browser.url('/', { inject: arrowFunction })
+        expect(got.mock.calls[0][1].json).toEqual({ url: 'http://foobar.com/' })
+    })
+
+    it('should accept a normal function as the inject parameter', async () => {
+        const arrowFunction = jest.fn(function () { console.log('foo-bar') })
+        await browser.url('/', arrowFunction)
```
erwinheitzman, line 59: Since the arrowFunction isn't used at all for any assertions, simply do the following instead:
```suggestion
        await browser.url('/', () => {})
```

packages/webdriverio/tests/commands/browser/url.test.js
```diff
@@ -27,14 +27,39 @@ describe('url', () => {
...
-    it('should throw an exception when a non-string value passed in', async () => {
+    it('should throw an exception when a non-string value passed in path', async () => {
         try {
             await browser.url(true)
         } catch (e) {
             expect(e.message).toContain('command needs to be type of string')
         }
     })
 
+    it('should throw an exception when a non-function value passed in inject', async () => {
```
erwinheitzman, line 38: This test doesn't fail if no error was thrown, instead use `expect(X).reject.toThrowError(Y)`

packages/webdriverio/tests/commands/browser/url.test.js
```diff
@@ -27,14 +27,39 @@ describe('url', () => {
             .toEqual({ url: 'http://foobar.com/foobar' })
     })
 
-    it('should throw an exception when a non-string value passed in', async () => {
+    it('should throw an exception when a non-string value passed in path', async () => {
```
erwinheitzman, line 30: This test doesn't fail if no error was thrown, instead use expect(X).reject.toThrowError(Y)

packages/webdriverio/src/commands/browser/url.js
```diff
@@ -25,6 +25,8 @@
  * </example>
  *
  * @param {String=} url  the URL to navigate to
```
erwinheitzman, line 27: Everywhere this is called path, please adjust the naming here as well ;)

erwinheitzman (commented):
Left some comments 👍

packages/webdriverio/tests/commands/browser/url.test.js
AutomationReddy: As we are passing a wrong type, it will throw an error and that's what we are doing here.

```
it('should throw an exception when a non-function value passed in inject', async () => {
try {
            await browser.url(``, {inject: 123}) // inject only accepts function and rest will return error and we are validating it in the catch block
        } catch (e) {
            expect(e.message).toContain('command needs to be type of string')
        }
    })
```

packages/webdriverio/tests/commands/browser/url.test.js
AutomationReddy: Same as above test

packages/webdriverio/tests/commands/browser/url.test.js
AutomationReddy: @erwinheitzman I got that. However, if you see @christian-bromann comments, I am lagging on how to verify whether puppeteer was used to inject the function.

Is there a way, where I can verify puppeteer instance is created? The above test will create the puppeteer instance and I can add a step in the above function inject tests to check instance is created or not.

packages/webdriverio/src/commands/browser/url.js
AutomationReddy: Modified

packages/webdriverio/src/commands/browser/url.js
```diff
@@ -25,6 +25,7 @@
...
+ * @param {Function=} inject pass script as a function
```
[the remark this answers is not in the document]
AutomationReddy:
```type UrlOptions = {
        inject: Function
    }
``` Is this fine?

File: packages/webdriverio/tests/commands/browser/url.test.js (hunk 7 above)
Comment (erwinheitzman) [reply packages/webdriverio/tests/commands/browser/url.test.js:38]: What I mean is that it is a false positive if (for whatever reason) no error is thrown.

File: packages/webdriverio/tests/commands/browser/url.test.js (hunk 6 above)
Comment (erwinheitzman) [reply packages/webdriverio/tests/commands/browser/url.test.js:59]: It will not use the webdriver protocol so I would say that you can assert the mock calls

File: packages/webdriverio/src/commands/browser/url.js [hunk 11]
```diff
@@ -35,9 +37,30 @@
...
+            const allPages = await browser.pages()
+            for(let page of allPages) {
+                const state = await page.evaluate(() => document.visibilityState) // eslint-disable-line
+                if(state === 'visible') {
+                    await page.evaluate(inject)
+                    break
+                }
+            }
+        }
+    }
```
christian-bromann, line 60:
```suggestion
    if(inject) {
        if(typeof inject !== 'function') {
            throw new Error('Parameter "inject" for url command needs to be type of function')
        }

        const browser = await this.getPuppeteer()
        const allPages = await browser.pages()
        for(let page of allPages) {
            const state = await page.evaluate(() => document.visibilityState) // eslint-disable-line
            if(state === 'visible') {
                await page.evaluate(inject)
                break
            }
        }
    }
```

packages/webdriverio/tests/commands/browser/url.test.js
christian-bromann: I think what @erwinheitzman is saying is that if `browser.url(true)` doesn't fail the test passes even though it should fail. The solution here is to add

```js
expect.assertions(1)
```

to say we at least expect 1 assertion. This hasn't to do anything with your patch but we should fix it while we at it.

packages/webdriverio/tests/commands/browser/url.test.js
christian-bromann: The solution here is to add

```js
expect.assertions(1)
```

to say we at least expect 1 assertion. This hasn't to do anything with your patch but we should fix it while we at it.

packages/webdriverio/tests/commands/browser/url.test.js
christian-bromann: you need to mock out Puppeteer first:
```js
const pageMock = { evaluate: jest.fn() }
browser.getPuppeteer = jest.fn().mockReturnValue({
    pages: [pageMock]
})
```

then assert if the page mock was called twice:

```js
expect(pageMock.evaluate).toBeCalledTimes(2)
```

packages/webdriverio/tests/commands/browser/url.test.js
christian-bromann: mock puppeteer similar as I did above and then check if evaluate of page has been called with that property:

```js
expect(pageMock).toBeCalledWith(arrowFunction)
```

packages/webdriverio/tests/commands/browser/url.test.js
christian-bromann: I think it is useful to pass in a parameter so that the assertion is easier:

```js
const urlParams = { inject: jest.fn() }
await browser.url('/', urlParams)
...
expect(pageMock.evaluate).toBeCalledWith(urlParams.inject)
```

AutomationReddy:
> 
> 
> @AutomationReddy please let me know if I can help moving this forward

@christian-bromann Sorry for the delay in this. Just tied up with releases. I will work on this ASAP. Thanks for helping out

packages/webdriverio/src/commands/browser/url.js
```diff
@@ -35,9 +37,29 @@
...
+        }
+        const browser = await this.getPuppeteer()
+        const allPages = await browser.pages()
+        for (let page of allPages) {
+            const state = await page.evaluate(() => document.visibilityState) // eslint-disable-line
+            if (state === 'visible') {
+                await page.evaluate(inject)
+                break
+            }
+        }
```
christian-bromann, line 58: 1. let's make sure to only inject it in the page that is currently active
2. We need to use this here: https://pptr.dev/#?product=Puppeteer&version=v5.4.1&show=api-pageevaluateonnewdocumentpagefunction-args in order to have the script executed before the page loads

christian-bromann (changes requested):

christian-bromann:
ping @AutomationReddy , can I help getting this landed?

christian-bromann:
ping @AutomationReddy

packages/webdriverio/src/commands/browser/url.js
```diff
@@ -35,9 +37,29 @@
...
+    /**
+     * If inject is a function, then we attach puppeteer to the existing browser instance and
+     * injects the passed function on the active tab
+     *
+     */
+    if (inject) {
+        if (typeof inject !== 'function') {
+            throw new Error('Parameter "inject" for url command needs to be type of function')
+        }
+        const browser = await this.getPuppeteer()
```
emilyrohrbough, line 50: Assume puppeteer should be attached seems pretty limiting for this feature. Are we sure this assumption is desirable?

For example, one might want to inject axe-core onto the page for accessibility testing, but doesn't use puppeteer.

emilyrohrbough: If attaching Puppeteer is the desired default behavior, would it make sense to create an opt-out option ?

christian-bromann: This doesn't attach Puppeteer to anything. We need Puppeteer to inject a script here. You will be able to use axe-core without having to deal with Puppeteer at all.

emilyrohrbough: Ah this comment might need updated then:
https://github.com/webdriverio/webdriverio/pull/5929/files/fdeff0e094490e4cc12ecdfbcc7454856317f216#diff-e8517450f670f58b95cd042559d95b4938983d60caf1aa97b6567caaf10ceb8dR42-R43

christian-bromann:
@AutomationReddy this PR seems pretty far ahead, do you think you can pick it up eventually?

AutomationReddy:
> 
> 
> @AutomationReddy this PR seems pretty far ahead, do you think you can pick it up eventually?

@christian-bromann I will work on this shortly

christian-bromann:
@AutomationReddy I will close this issue due to lack of activity. Also the codebase has changed since all the commits were made and it might be easier anyway to pick up this work on a clean and updated code base.

PR #5930, opened 2020-10-05
Fix server mocking issues

mgrybyk:
## Proposed changes

- fix headers issue `Invalid parameters responseHeaders.9.name: string value expected`
- fix an attempt to use an empty string as a local file
- fix an issue when user function returns `undefined`
- fix buffer issue causing images to be corrupted
- remove sync methods causing a slowdown when there are thousands of requests
- suppress exceptions on timeout, session close, etc

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/packages/webdriverio/src/utils/interception/devtools.js
+++ b/packages/webdriverio/src/utils/interception/devtools.js
@@ -3,9 +3,12 @@ import path from 'path'
 import atob from 'atob'
 import minimatch from 'minimatch'
 
+import logger from '@wdio/logger'
 import Interception from './'
 import { containsHeaderObject } from '..'
 import { ERROR_REASON } from '../../constants'
+
+const log = logger('webdriverio')
 
 export default class DevtoolsInterception extends Interception {
     static handleRequestInterception (client, mocks) {
@@ -113,7 +116,7 @@ export default class DevtoolsInterception extends Interception {
                         responseHeaders,
                         /** do not mock body if it's undefined */
                         body: isBodyUndefined ? undefined : Buffer.from(newBody, 'binary').toString('base64')
-                    }).catch(/* istanbul ignore next */() => {})
+                    }).catch(/* istanbul ignore next */logFetchError)
                 }
 
                 /**
@@ -123,11 +126,11 @@ export default class DevtoolsInterception extends Interception {
                     return client.send('Fetch.failRequest', {
                         requestId,
                         errorReason
-                    }).catch(/* istanbul ignore next */() => {})
+                    }).catch(/* istanbul ignore next */logFetchError)
                 }
             }
 
-            return client.send('Fetch.continueRequest', { requestId }).catch(/* istanbul ignore next */() => {})
+            return client.send('Fetch.continueRequest', { requestId }).catch(/* istanbul ignore next */logFetchError)
         }
     }
 
@@ -252,4 +255,9 @@ const tryParseJson = (body) => {
     } catch {
         return body
     }
+}
+
+const logFetchError = (err) => {
+    /* istanbul ignore next */
+    log.debug(err && err.message ? err.message : err)
 }
--- a/packages/webdriverio/src/utils/interception/devtools.js
+++ b/packages/webdriverio/src/utils/interception/devtools.js
@@ -1,8 +1,7 @@
-import fs from 'fs'
+import fse from 'fs-extra'
 import path from 'path'
 import atob from 'atob'
 import minimatch from 'minimatch'
-import { canAccess } from '@wdio/utils'
 
 import Interception from './'
 import { containsHeaderObject } from '..'
@@ -76,7 +75,7 @@ export default class DevtoolsInterception extends Interception {
                         newBody = await overwrite(request, client)
                     }
 
-                    if (typeof newBody === undefined) {
+                    if (typeof newBody === 'undefined') {
                         newBody = ''
                     }
 
@@ -94,8 +93,8 @@ export default class DevtoolsInterception extends Interception {
                      * check if local file and load it
                      */
                     const responseFilePath = path.isAbsolute(newBody) ? newBody : path.join(process.cwd(), newBody)
-                    if (responseFilePath.length > 0 && fs.existsSync(responseFilePath) && canAccess(responseFilePath)) {
-                        newBody = fs.readFileSync(responseFilePath).toString()
+                    if (newBody.length > 0 && await fse.pathExists(responseFilePath) && await canAccess(responseFilePath)) {
+                        newBody = (await fse.readFile(responseFilePath)).toString()
                     } else if (newBody.startsWith('http')) {
                         responseCode = 301
                         /**
@@ -229,4 +228,18 @@ const filterStatusCode = (statusCode, expected) => {
         return expected(statusCode) !== true
     }
     return statusCode !== expected
+}
+
+/**
+ * Helper utility to check file access
+ * @param {String} file file to check access for
+ * @return              Promise<true> if file can be accessed
+ */
+const canAccess = async (filepath) => {
+    try {
+        await fse.access(filepath)
+        return true
+    } catch {
+        return false
+    }
 }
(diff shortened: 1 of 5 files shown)


packages/webdriverio/src/utils/interception/devtools.js
```diff
@@ -76,21 +76,25 @@ export default class DevtoolsInterception extends Interception {
                         newBody = await overwrite(request, client)
                     }
 
+                    if (typeof newBody === undefined) {
+                        newBody = ''
+                    }
```
christian-bromann, line 81: I don't think this is necessary because `newBody` will never be undefined, see

```js
                if (overwrite !== undefined) {
                   let newBody = overwrite
```

mgrybyk: if `overwrite` is a function user can return `undefined` in line 76

mgrybyk:
@christian-bromann fixed all issues. Will work on tests a bit later

packages/webdriverio/src/utils/interception/devtools.js
```diff
@@ -107,7 +111,7 @@ export default class DevtoolsInterception extends Interception {
                         requestId,
                         responseCode,
                         responseHeaders,
-                        body: Buffer.from(newBody).toString('base64')
+                        body: newBody ? Buffer.from(newBody, 'binary').toString('base64') : undefined
```
christian-bromann, line 114: What happens if `newBody` is a string here?

mgrybyk: should be `undefined` https://github.com/puppeteer/puppeteer/blob/950ae334ca9c4c37b5a34dc61596a89596fd2bc6/src/common/HTTPRequest.ts#L381

packages/webdriverio/src/utils/interception/devtools.js
```diff
@@ -76,21 +76,25 @@ export default class DevtoolsInterception extends Interception {
                         newBody = await overwrite(request, client)
                     }
 
+                    if (typeof newBody === undefined) {
```
mgrybyk, line 79: typeof newBody === 'undefined'

packages/webdriverio/src/utils/interception/devtools.js
```diff
@@ -76,21 +76,25 @@ export default class DevtoolsInterception extends Interception {
...
-                        ...Object.entries(params.headers || {}).map(([key, value]) => { key, value })
+                        ...Object.entries(params.headers || {}).map(([name, value]) => ({ name, value }))
                     ]
 
                     /**
                      * check if local file and load it
                      */
                     const responseFilePath = path.isAbsolute(newBody) ? newBody : path.join(process.cwd(), newBody)
-                    if (fs.existsSync(responseFilePath) && canAccess(responseFilePath)) {
+                    if (responseFilePath.length > 0 && fs.existsSync(responseFilePath) && canAccess(responseFilePath)) {
```
mgrybyk, line 97: newBody.length

packages/webdriverio/src/utils/interception/devtools.js
mgrybyk: @christian-bromann got your point, I think.
There is no difference between an empty string and undefined in our case.
Let think fix that

mgrybyk: fixed

packages/webdriverio/src/utils/interception/devtools.js
```diff
@@ -76,22 +75,27 @@ export default class DevtoolsInterception extends Interception {
...
                         ...event.responseHeaders,
-                        ...Object.entries(params.headers || {}).map(([key, value]) => { key, value })
+                        ...Object.entries(params.headers || {}).map(([name, value]) => ({ name, value }))
                     ]
 
                     /**
                      * check if local file and load it
                      */
                     const responseFilePath = path.isAbsolute(newBody) ? newBody : path.join(process.cwd(), newBody)
-                    if (fs.existsSync(responseFilePath) && canAccess(responseFilePath)) {
```
christian-bromann, line 93: Why do we need `fs-extra` as extra dependency if `fs.existsSync(responseFilePath)` would do the same thing?

packages/webdriverio/src/utils/interception/devtools.js
```diff
@@ -76,22 +75,27 @@ export default class DevtoolsInterception extends Interception {
...
-                        newBody = fs.readFileSync(responseFilePath).toString()
+                    if (newBody.length > 0 && await fse.pathExists(responseFilePath) && await canAccess(responseFilePath)) {
+                        newBody = (await fse.readFile(responseFilePath)).toString()
```
christian-bromann, line 98: What are the advantages between both?

packages/webdriverio/src/utils/interception/devtools.js
mgrybyk: `fs-extra` exists anyway as a dependency of some packages.
`existsSync` - is sync method, while `pathExists` is async method.

packages/webdriverio/src/utils/interception/devtools.js
mgrybyk: is the question about fs/fs-extra ?

packages/webdriverio/src/utils/interception/devtools.js
christian-bromann: > existsSync - is sync method, while pathExists is async method.

What is the advantage?

packages/webdriverio/src/utils/interception/devtools.js
christian-bromann: yeah

packages/webdriverio/src/utils/interception/devtools.js
mgrybyk: having one of sync methods here causing some issues to mocking
- pages loads slower if there are > 1000 requests
- I'm getting errors regarding invalid InterceptionId

packages/webdriverio/tests/utils/interception/devtools.test.js
```diff
@@ -1,7 +1,19 @@
-import { canAccess } from '@wdio/utils'
-
 import NetworkInterception from '../../../src/utils/interception/devtools'
 
+const fse = jest.requireActual('fs-extra')
```
christian-bromann, line 3: Why do we import the actual package? I think it would be better to use a mock version throughout the whole test.

packages/webdriverio/src/utils/interception/devtools.js
christian-bromann: I would say it is simpler to promisify the `fs` methods here, this would also not require changes in the unit tests.

packages/webdriverio/tests/utils/interception/devtools.test.js
mgrybyk: I need this one to read the file

mgrybyk:
answering to https://github.com/webdriverio/webdriverio/pull/5930#discussion_r500235408

It would now cause some refactoring.
I'd have to import both `promisify` and `fs` in two places, wrap several functions...

What is the issue of having `fs-extra`? Don't think a couple of additional Kb of code will have any impact on the existing 15 Mb.
```
before: `15776`
after: `15984`
delta: `208`
```
Also, changing [one line of code](https://github.com/webdriverio/webdriverio/pull/5930/files#diff-32ea12dd2dcebc878ebb03bae767f45f) doesn't look to me like a refactoring of unit tests.

Anyway, if you insist I can do that, it looks to me like I'm missing something

christian-bromann:
Given that we use `fs-extra` in other packages too I am not too much concerned about it.

The only request I would like to make is from [this comment](https://github.com/webdriverio/webdriverio/pull/5930#discussion_r500234775). Let's not do any actual file reads in unit test and use a mocked version of `fs-extra`. The unit test is not concerned about what is written in the file so we can just return a random string or JSON blob.

mgrybyk:
will do.
Found some other issues, going to fix them as well

christian-bromann (approved):
All these

```js
.catch(/* istanbul ignore next */() => {})
```

worry me a bit, can we at least log the error for visibility?

The rest looks fine to me 👍

mgrybyk:
Done

christian-bromann (approved):
👍

PR #5943, opened 2020-10-06
Suppress warning when running tests in Edge and update docs

mgrybyk:
## Proposed changes

- suppress `Warning: unknown product name edge. Falling back to chrome.` warning when running tests in Edge
- update devtools docs

## Types of changes

- [x] Polish
- [x] Documentation update

## Checklist


packages/devtools/src/launcher.js
```diff
@@ -108,7 +108,8 @@ async function launchChrome (capabilities) {
     return browser
 }
 
-function launchBrowser (capabilities, product) {
+function launchBrowser (capabilities, browserType) {
+    const product = browserType === BROWSER_TYPE.firefox ? BROWSER_TYPE.firefox : BROWSER_TYPE.chrome
```
mgrybyk, line 112: puppeteer only supports firefox and chrome.

christian-bromann: Chromium Edge as well

packages/devtools/src/constants.js
```diff
@@ -125,7 +125,6 @@ export const BROWSER_ERROR_MESSAGES = {
 export const VENDOR_PREFIX = {
     chrome: 'goog:chromeOptions',
     firefox: 'moz:firefoxOptions',
-    edge: 'ms:edgeOptions'
```
christian-bromann, line 128: Chromium Edge is supported too!

packages/devtools/src/launcher.js
mgrybyk: Sorry for not being clear, `product` option support only `chrome` and `firefox`.

packages/devtools/src/constants.js
mgrybyk: let's talk about it in one place https://github.com/webdriverio/webdriverio/pull/5943#discussion_r500648714

packages/devtools/src/launcher.js
mgrybyk: https://github.com/puppeteer/puppeteer/blob/950ae334ca9c4c37b5a34dc61596a89596fd2bc6/src/node/Launcher.ts#L803

mgrybyk: I'm able to run tests in both headless and headful modes in the Edge browser, of course.
https://github.com/mgrybyk/webdriverio-devtools/blob/master/config/wdio.EDGE.conf.ts#L12

packages/devtools/src/constants.js
christian-bromann: We still need this to properly generated [client capabilities](https://github.com/webdriverio/webdriverio/blob/master/packages/devtools/src/index.js#L40-L59)

mgrybyk: Assuming that user won't ever use "ms:edgeOptions" with devtools why would we need it?
Was not able to figure out the use case

mgrybyk: @christian-bromann reverted this one, however still don't understand why we need it.

christian-bromann (approved):
👍

PR #5957, opened 2020-10-08
Refactor webdriverio utils interception index to TS

Fabianopb:
## Proposed changes

Refactoring `webdriverio` to TypeScript as part of https://github.com/webdriverio/webdriverio/issues/5854

## Types of changes

- [ ] Refactoring to TypeScript, starting with a small part of `utils`

## Checklist

## Further comments

I imagine that at this point, the very start, it is ok to have a few `any`s and `unknown`s here and there until types start to hook up with other classes higher in the hiearchy.

### DIFF
--- a/packages/webdriverio/src/index.js
+++ b/packages/webdriverio/src/index.js
@@ -1,3 +1,4 @@
+/// <reference types="webdriverio/webdriverio"/>
 import path from 'path'
 import logger from '@wdio/logger'
 import WebDriver, { DEFAULTS } from 'webdriver'
--- a/packages/webdriverio/src/index.js
+++ b/packages/webdriverio/src/index.js
@@ -1,4 +1,3 @@
-/// <reference types="webdriverio/webdriverio"/>
 import path from 'path'
 import logger from '@wdio/logger'
 import WebDriver, { DEFAULTS } from 'webdriver'
--- a/packages/webdriverio/src/utils/interception/index.ts
+++ b/packages/webdriverio/src/utils/interception/index.ts
@@ -31,7 +31,7 @@ export default class Interception {
         errorReason: string;
     }[];
     matches: any[];
-    calls: any[];
+    calls?: any[];
 
     constructor (url: string, filterOptions = {}, browser: Browser) {
         this.url = url
@@ -39,7 +39,6 @@ export default class Interception {
         this.browser = browser
         this.respondOverwrites = []
         this.matches = []
-        this.calls = []
     }
 
     waitForResponse ({
@@ -58,7 +57,7 @@ export default class Interception {
             interval = this.browser.options.waitforInterval
         }
 
-        const fn = () => this.calls.length > 0
+        const fn = () => this.calls && this.calls.length > 0
         const timer = new Timer(interval, timeout, fn, true) as unknown as Promise<void>
 
         return this.browser.call(() => timer.catch((e) => {
--- a/packages/webdriverio/src/utils/interception/index.ts
+++ b/packages/webdriverio/src/utils/interception/index.ts
@@ -1,19 +1,47 @@
 import Timer from '../Timer'
 
+type FilterOptions = {
+    method?: string;
+    header?: object;
+    responseHeaders?: object;
+};
+
+type Browser = {
+    options: {
+        waitforTimeout: number;
+        waitforInterval: number;
+    };
+    call: (cb: () => void) => void;
+};
+
+type WaitForResponseFn = {
+    timeout?: number;
+    interval?: number;
+    timeoutMsg?: string;
+}
+
 export default class Interception {
-    constructor (url, filterOptions = {}, browser) {
+    url: string;
+    filterOptions: FilterOptions;
+    browser: Browser;
+    respondOverwrites: any[];
+    matches: any[];
+    calls: any[];
+
+    constructor (url: string, filterOptions = {}, browser: Browser) {
         this.url = url
         this.filterOptions = filterOptions
         this.browser = browser
         this.respondOverwrites = []
         this.matches = []
+        this.calls = []
     }
 
     waitForResponse ({
         timeout = this.browser.options.waitforTimeout,
         interval = this.browser.options.waitforInterval,
-        timeoutMsg
-    } = {}) {
+        timeoutMsg,
+    }: WaitForResponseFn = {}) {
         /*!
          * ensure that timeout and interval are set properly
          */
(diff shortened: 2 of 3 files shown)


packages/webdriverio/src/utils/interception/index.ts
```diff
@@ -1,7 +1,39 @@
 import Timer from '../Timer'
 
+type FilterOptions = {
+    method?: string;
+    header?: object;
+    responseHeaders?: object;
+};
+
+type Browser = {
```
christian-bromann, line 9: You should get this from [`WebdriverIO.BrowserObject`](https://github.com/webdriverio/webdriverio/blob/master/packages/webdriverio/webdriverio.d.ts#L50)

christian-bromann (commented):
One comment

packages/webdriverio/src/utils/interception/index.ts
Fabianopb: Hummm... got it, although it doesn't seem to be a public namespace and it's not defined in a module... How do you recommend to import it?

Fabianopb: I have referenced the declarations path to use that namespace (https://github.com/webdriverio/webdriverio/pull/5957/commits/1b3931fbb3ac04511bbad2f31def778ce9e01ee1), this could also be done in `src/index` of this module and I guess the namespace should be available all across the module, what do you think?

packages/webdriverio/src/utils/interception/index.ts
```diff
@@ -1,7 +1,32 @@
...
+    timeoutMsg?: string;
+}
+
 export default class Interception {
-    constructor (url, filterOptions = {}, browser) {
+    url: string;
+    filterOptions: FilterOptions;
+    browser: WebdriverIO.Browser;
+    respondOverwrites: {
+        overwrite: any;
```
christian-bromann, line 21: this needs a type definition

packages/webdriverio/src/utils/interception/index.ts
```diff
@@ -1,7 +1,32 @@
...
+        params: object;
+        sticky: boolean;
+        errorReason: string;
+    }[];
+    matches: any[];
+    calls?: any[];
```
christian-bromann, line 27: same

christian-bromann (changes requested):
Some missing type defintions, let me know if you need help with them

packages/webdriverio/src/utils/interception/index.ts
Fabianopb: yes, I know I know! Just opened as a draft because I wanted to know how to structure it before going deeper into the types! So I guess it's ok to use `<reference types="webdriverio/webdriverio"/>` here?

Fabianopb: I assumed it's fine, updated! :)

packages/webdriverio/src/utils/interception/index.ts
Fabianopb: Fixed too!

Fabianopb:
> Some missing type defintions, let me know if you need help with them

Went ahead and explored a bit more this package's declarations, I hope they are right now! I've marked one to-do item as it depends on another class that is not currently typed. Let me know if those are fine!

packages/webdriverio/src/index.js
```diff
@@ -1,3 +1,4 @@
+/// <reference types="webdriverio/webdriverio"/>
```
christian-bromann, line 1: this is a js file, no need for a ts reference statement

packages/webdriverio/src/utils/interception/index.ts
```diff
@@ -25,8 +37,9 @@ export default class Interception {
             interval = this.browser.options.waitforInterval
         }
 
-        const fn = () => this.calls.length > 0
-        const timer = new Timer(interval, timeout, fn, true)
+        const fn = () => this.calls && this.calls.length > 0
+        // TODO: fix unknown type when Timer is typed
+        const timer = new Timer(interval, timeout, fn, true) as unknown as Promise<void>
```
christian-bromann, line 42: no need to add the todo given that timer hasn't been translated yet, also:
```suggestion
        // TODO: fix unknown type when Timer is typed
        const timer = new Timer(interval, timeout, fn, true) as unknown as Promise<boolean>
```

Fabianopb: Alright, removed

packages/webdriverio/src/index.js
Fabianopb: I think it does, this will be interpreted by the package as global type definitions even though this is a JS file. To show this off I removed it and linter fails because it cannot find the namespace anymore. You can see the same error in the editor:
image

My suggestion is that we keep this reference in this file.

packages/webdriverio/src/utils/interception/index.ts
Fabianopb: Removed!

packages/webdriverio/src/index.js
Fabianopb: So @christian-bromann, what do you say, should we keep this reference here? 🙂 

I think it will make refactoring of the module easier, especially if these types are intended to be globally available in its scope. Otherwise, any suggestions?

christian-bromann: I guess we should. Still, weird that the TS compiler even notices this in a js file.

Fabianopb: Ok, found a more elegant solution: https://github.com/webdriverio/webdriverio/pull/5957/commits/dd0d8669c0802e66e509b859fe3b50c8b5c23553

christian-bromann (approved):
Nice 👌

packages/webdriverio/tsconfig.json
```diff
@@ -3,7 +3,8 @@
   "compilerOptions": {
     "baseUrl": ".",
     "outDir": "./build",
-    "rootDir": "./src"
+    "rootDir": "./src",
+    "types": ["webdriverio"]
```
mgrybyk, line 7: @Fabianopb FYI: you have to update `tsconfig.prod.json` as well.

Fixed it in https://github.com/webdriverio/webdriverio/commit/903a25575cb0e9f9a553265249fcb9c8f50c18ae

Fabianopb: Good point, thanks for the fix!

PR #5961, opened 2020-10-09
retry if SC fails to start

enriquegh:
## Proposed changes

Someone reported an issue of Sauce Connect launcher failing with ENOENT which is normally impossible because we download Sauce Connect before we run it. It seems that there is a race condition where checking if the Sauce Connect binary was downloaded fails when looking for the version. The execution happens as follows:

    check if the binary was downloaded
    if no -> download
    if yes continue
    check if version is the one that has been downloaded
    if no -> download expected version
    if yes finish

The race condition seem to happen when the check if the binary exists was successful but then the consecutive version check fails if in between someone cleaned the binary which can happen in an CI/CD environment where dependencies are shared.

This patch introduces a retry to see if this race condition can be caught.

See:
https://github.com/karma-runner/karma-sauce-launcher/pull/219
and
https://github.com/saucelabs/node-saucelabs/issues/86

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/wdio-sauce-service/src/launcher.js
+++ b/packages/wdio-sauce-service/src/launcher.js
@@ -10,12 +10,14 @@ const SC_RELAY_DEPCRECATION_WARNING = [
     'with the upcoming versions of @wdio/sauce-service. Please',
     'remove the option as tests should work identically without it.'
 ].join(' ')
+const MAX_SC_START_TRIALS = 3
 
 const log = logger('@wdio/sauce-service')
 export default class SauceLauncher {
     constructor (options, capabilities, config) {
         this.options = options
         this.api = new SauceLabs(config)
+        this.scStartTrials = 0
     }
 
     /**
@@ -75,7 +77,28 @@ export default class SauceLauncher {
         obs.observe({ entryTypes: ['measure'], buffered: false })
 
         performance.mark('sauceConnectStart')
-        this.sauceConnectProcess = await this.api.startSauceConnect(this.sauceConnectOpts)
+
+        try {
+            this.sauceConnectProcess = await this.api.startSauceConnect(this.sauceConnectOpts)
+        } catch (err) {
+            ++this.scStartTrials
+            /**
+             * fail starting Sauce Connect eventually
+             */
+            if (
+                /**
+                 * only fail for ENOENT errors due to racing condition
+                 * see: https://github.com/saucelabs/node-saucelabs/issues/86
+                 */
+                !err.message.includes('ENOENT') ||
+                /**
+                 * or if we reached the maximum rety count
+                 */
+                this.scStartTrials >= MAX_SC_START_TRIALS
+            ) {
+                throw err
+            }
+        }
         performance.mark('sauceConnectEnd')
         performance.measure('bootTime', 'sauceConnectStart', 'sauceConnectEnd')
     }
--- a/packages/wdio-sauce-service/src/launcher.js
+++ b/packages/wdio-sauce-service/src/launcher.js
@@ -17,7 +17,6 @@ export default class SauceLauncher {
     constructor (options, capabilities, config) {
         this.options = options
         this.api = new SauceLabs(config)
-        this.scStartTrials = 0
     }
 
     /**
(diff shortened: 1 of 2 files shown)


christian-bromann (changes requested):
Your solution would only retry starting Sauce Connect Proxy once. You can move the whole part into a new function to allow to invoke it recursively.

Also it would be good if we could add tests for this.

enriquegh:
@christian-bromann did a change to this that I think is what you mentioned (I get a little lost with async stuff).
Testing this manually starts the tunnel as expected.
I tried to test the failure manually but didn't have too much luck as it is mostly caused by a race condition.
I tried deleting the binary after the sc binary is downloaded but before the sc --version is called but no luck.

Any ideas on how to try and test this?

packages/wdio-sauce-service/src/launcher.js
```diff
@@ -75,11 +77,39 @@ export default class SauceLauncher {
...
+                 * or if we reached the maximum rety count
+                 */
+                this.scStartTrials >= MAX_SC_START_TRIALS
+            ) {
+                throw err
+            }
+            log.debug(`Failed to start Sauce Connect Proxy due to ${err.stack}`)
+            log.debug(`Retrying ${this.scStartTrials}/${MAX_SC_START_TRIALS}`)
+            this.sauceConnectProcess = await this.startTunnel()
+        }
```
christian-bromann, line 104: I suggest to move this whole try/catch block into `startTunnel` and make sure we return at the end and because right now this will be only retried once.

christian-bromann (changes requested):

packages/wdio-sauce-service/src/launcher.js
```diff
@@ -75,11 +76,38 @@ export default class SauceLauncher {
         obs.observe({ entryTypes: ['measure'], buffered: false })
 
         performance.mark('sauceConnectStart')
-        this.sauceConnectProcess = await this.api.startSauceConnect(this.sauceConnectOpts)
+        await this.startTunnel()
```
christian-bromann, line 79:
```suggestion
        this.sauceConnectProcess = await this.startTunnel()
```

packages/wdio-sauce-service/src/launcher.js
```diff
@@ -75,11 +76,38 @@ export default class SauceLauncher {
...
         performance.mark('sauceConnectEnd')
         performance.measure('bootTime', 'sauceConnectStart', 'sauceConnectEnd')
     }
 
+    async startTunnel(retryCount = 0) {
+        try {
+            this.sauceConnectProcess = await this.api.startSauceConnect(this.sauceConnectOpts)
```
christian-bromann, line 86:
```suggestion
            return this.api.startSauceConnect(this.sauceConnectOpts)
```

packages/wdio-sauce-service/src/launcher.js
```diff
@@ -75,11 +76,38 @@ export default class SauceLauncher {
...
+                /**
+                 * or if we reached the maximum rety count
+                 */
+                retryCount >= MAX_SC_START_TRIALS
+            ) {
+                throw err
+            }
+            log.debug(`Failed to start Sauce Connect Proxy due to ${err.stack}`)
+            log.debug(`Retrying ${retryCount}/${MAX_SC_START_TRIALS}`)
+            this.sauceConnectProcess = await this.startTunnel(retryCount)
```
christian-bromann, line 107:
```suggestion
            return this.startTunnel(retryCount)
```

packages/wdio-sauce-service/src/launcher.js
```diff
@@ -75,11 +76,38 @@ export default class SauceLauncher {
         obs.observe({ entryTypes: ['measure'], buffered: false })
 
         performance.mark('sauceConnectStart')
-        this.sauceConnectProcess = await this.api.startSauceConnect(this.sauceConnectOpts)
+        await this.startTunnel()
         performance.mark('sauceConnectEnd')
         performance.measure('bootTime', 'sauceConnectStart', 'sauceConnectEnd')
     }
 
+    async startTunnel(retryCount = 0) {
```
christian-bromann, line 84:
```suggestion
    startTunnel(retryCount = 0) {
```

christian-bromann (changes requested):
Found a little optimization in the code. If we return the sauce connect process we can actually remove the `async/await` completely. It also makes more sense to rather return a value to assign it to a variable when you call it rather than modifying the scope properties within the function because you never can keep up which properties you will change.

enriquegh:
Testing this looks like the tests are now failing.
I see the the ENOENT is happening but is not getting caught by jest and/or the package.

Need to look into this a bit more unless you got some ideas.

christian-bromann (approved):
👍 , thanks Enrique!

PR #5962, opened 2020-10-09
Tracking doc: rewrite `@wdio/dot-reporter` package into TypeScript

suniljaiswal01:
## Proposed changes

if you are interested converting source files from /packages/wdio-dot-reporter into TypeScript please comment on this thread and mention the files you want to convert to avoid duplicate work
there is no reason to convert the whole package at once - smaller PRs make it easier to review the code
if you make a PR, reference this issue so we can keep track of the progress
if you make a PR, label it with TypeScript and the package name wdio-dot-reporter so we can easier find all PRs
if you make a PR, assign it to the TypeScript Rewrite project

#5835


packages/wdio-dot-reporter/src/index.ts
```diff
@@ -5,7 +5,10 @@ import WDIOReporter from '@wdio/reporter'
  * Initialize a new `Dot` matrix test reporter.
  */
 export default class DotReporter extends WDIOReporter {
-    constructor (options) {
+    options!: WDIOReporter.Options;
+    outputStream: any;
```
christian-bromann, line 9:
```suggestion
    outputStream: fs.WriteStream;
```

christian-bromann (changes requested):
One comment, also there are some trailing space linting errors

packages/wdio-dot-reporter/tests/index.test.ts
```diff
@@ -7,18 +7,19 @@ describe('Dot Reporter', () => {
         const logFile = tmp.fileSync()
         const reporter = new DotReporter({ logFile: logFile.name })
         reporter.write = jest.fn()
+        reporter.outputStream.write = jest.fn()
 
         reporter.onTestSkip()
-        expect(reporter.write.mock.calls[0][0]).toBe('cyanBright .')
-        reporter.write.mockClear()
+        expect((reporter.write as jest.Mock).mock.calls[0][0]).toBe('cyanBright .');
```
mgrybyk, line 13: Can you defined a variable for `reporter.write as jest.Mock` to avoid using `as` that many times

packages/wdio-reporter/reporter.d.ts
```diff
@@ -26,9 +26,9 @@ declare namespace WDIOReporter {
     }
 
     interface Options {
-        configFile: string;
-        logFile: string;
-        logLevel: string;
+        configFile?: string;
```
mgrybyk, line 29: Are you sure these should be optional?

mgrybyk (changes requested):

packages/wdio-reporter/reporter.d.ts
suniljaiswal01: Yes. Because there are tests where we are checking if parameter are not passed, then default value should be selected, and if those are not optional then , it is giving compilation issue

packages/wdio-dot-reporter/tests/index.test.ts
suniljaiswal01: Changes done

packages/wdio-reporter/reporter.d.ts
mgrybyk: So, as a user I can omit passing these params and be sure that some default values are used?
If so, ok for me.
However, if we need the change just to test something let's revert

suniljaiswal01: > So, as a user I can omit passing these params and be sure that some default values are used?
Yes, as per my understanding, that is what happening in the script at below line:

https://github.com/webdriverio/webdriverio/blob/e016d1071be05ebec6126814f4ffc6ad5fa8beb8/packages/wdio-dot-reporter/src/index.ts#L16 

> If so, ok for me.
> However, if we need the change just to test something let's revert

mgrybyk: According to that line it looks like they should be required?

suniljaiswal01: are you sure about that? because it is mentioned in the comment as well that 

**make dot reporter to write to output stream by default**
        options = Object.assign({ stdout: true }, options)

mgrybyk: I think I'm missing something, let's keep it as is for now then

suniljaiswal01: how can i handle these test, if we revert the changes

https://github.com/suniljaiswal01/webdriverio/blob/typescript-dot-reporter/packages/wdio-dot-reporter/tests/index.test.ts#L25-L28

There are giving compile error

packages/wdio-dot-reporter/src/index.ts
```diff
@@ -1,11 +1,15 @@
 import chalk from 'chalk'
 import WDIOReporter from '@wdio/reporter'
+import fs from 'fs'
```
christian-bromann, line 3:
```suggestion
import type fs from 'fs'
```

packages/wdio-reporter/reporter.d.ts
christian-bromann: @mgrybyk I don't see `configFile` ever been used in any of the reporters. I would even say we should remove this from the interface.

WDYT?

christian-bromann (commented):
Minor comment otherwise LGTM

packages/wdio-dot-reporter/src/index.ts
suniljaiswal01: Done

suniljaiswal01:
> @suniljaiswal01 would you mind to remove `configFile` from https://github.com/webdriverio/webdriverio/pull/5962/files#diff-5fa12e644ce8c8987116ac7d1a4e1ba73e0cf847d8589c90bc615a78f015eed1R29 ?

should i remove it from https://github.com/suniljaiswal01/webdriverio/blob/typescript-dot-reporter/tests/typings/webdriverio-reporter/reporter.ts#L37 here as well?

mgrybyk:
Can it be that some 3rd party reporters still rely on that property?

suniljaiswal01:
> Can it be that some 3rd party reporters still rely on that property?

@christian-bromann  any suggestion here?

christian-bromann:
> Can it be that some 3rd party reporters still rely on that property?

I highly doubt it. Especially since this property was only defined in the typings. This should be cleaned up.

christian-bromann:
To fix the tests, remove https://github.com/webdriverio/webdriverio/blob/master/tests/typings/webdriverio-reporter/reporter.ts#L37

christian-bromann (approved):
👍 

any objections @mgrybyk ?

mgrybyk (approved):

PR #5974, opened 2020-10-12
Fix "Error in hook: undefined" error message

hieuxlu:
## Proposed changes

Fix error loggin in runLauncherHook, which might result in confusing error message: "Error in hook: undefined"

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/packages/wdio-cli/src/utils.js
+++ b/packages/wdio-cli/src/utils.js
@@ -48,7 +48,7 @@ export async function runServiceHook(launcher, hookName, ...args) {
  * @param {Object} capabilities
  */
 export async function runLauncherHook(hook, ...args) {
-    const catchFn = (e) => log.error(`Error in hook: ${e.stack}`)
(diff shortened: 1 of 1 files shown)


mgrybyk:
When does this happen?

christian-bromann:
I think this happens if you return a rejected promise passing in a string instead of an error:

```js
onPrepare () {
	return Promise.reject('foo')
}
```

This is rather a user error given that rejected promises should have an error object and not a string.

martinfrancois:
@hieuxlu maybe you can add a test which reproduces the error you've been getting? (Also to make sure we don't run into regressions later).

hieuxlu:
This is indeed an user error. In my case, it's when you request an API during `onPrepare` and reject the Promise using the error response. I closed this because this can be delegated to user for handling these kind of errors, and throw an Error object correctly. Thanks.

PR #5992, opened 2020-10-15
Rewrite WebdriverIO browser API commands

christian-bromann:
## Proposed changes

Rewriting all files in `/packages/webdriverio/src/commands/browser` as part of effort #5861 

__Note:__ I used `any` in some places because there were no types for them available. This should be cleaned up in the following PRs.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

n/a


packages/webdriverio/src/commands/browser/getPuppeteer.ts
```diff
@@ -34,15 +34,26 @@ export default async function getPuppeteer () {
     /**
      * attach to Firefox debugger session
      */
-    if (this.capabilities.browserName.toLowerCase() === 'firefox') {
-        const majorVersion = parseInt(this.capabilities.browserVersion.split('.').shift(), 10)
+    if (this.capabilities.browserName && this.capabilities.browserName.toLowerCase() === 'firefox') {
```
HananArgov, line 37: this can be set to 
```
this.capabilities.browserName?.toLowerCase() === 'firefox'
```
and have the same effect

packages/webdriver/src/types.ts
```diff
@@ -163,12 +163,13 @@ export interface FirefoxLogObject {
...
 export interface FirefoxOptions {
-    binary?: string,
-    args?: string[],
-    profile?: string,
-    log?: FirefoxLogObject,
+    debuggerAddress: string
+    binary?: string
+    args?: string[]
+    profile?: string
+    log?: FirefoxLogObject
```
emilyrohrbough, line 170: why drop the commas? does the linker expect one way or the other?

christian-bromann: I don't think there is any separator needed when you define an interface. Semicolons are optional and commas will be ignored.

packages/webdriverio/src/commands/browser/mock.ts
```diff
@@ -126,7 +127,7 @@ export default async function mock (url, filterOptions) {
     SESSION_MOCKS.add(networkInterception)
 
     if (this.isSauce) {
-        await networkInterception.init()
+        await (networkInterception as WebDriverNetworkInterception).init()
```
martinfrancois, line 130: you could probably do `const networkInterception: WebDriverNetworkInterception = new Net...` on line 126 and then you don't need the `as` here (or if not at least move the `as` there)

packages/webdriverio/src/commands/browser/newWindow.ts
```diff
@@ -56,6 +56,10 @@ export default async function newWindow (url, { windowName = 'New Window', windo
     const tabs = await this.getWindowHandles()
     const newTab = tabs.pop()
 
+    if (!newTab) {
+        throw new Error('No window handle was found to switcht to')
```
martinfrancois, line 60:
```suggestion
        throw new Error('No window handle was found to switch to')
```

packages/webdriverio/src/commands/browser/react$$.ts
```diff
@@ -29,13 +29,18 @@ import fs from 'fs'
...
-export default async function react$$ (selector, { props = {}, state = {} } = {}) {
+export default async function react$$ (
+    this: WebdriverIO.BrowserObject,
+    selector: string,
+    { props = {}, state = {} }: WebdriverIO.ReactSelectorOptions = {}
+) {
     await this.executeScript(resqScript.toString(), [])
     await this.execute(waitToLoadReact)
-    const res = await this.execute(react$$Script, selector, props, state)
+    const res = await this.execute(react$$Script, selector, props, state) as ElementReference[]
```
martinfrancois, line 43: if `execute` returns either `ElementReference` or `ElementReference[]`, I'd suggest adding `ElementReference | ElementReference[]` as the return type for `execute` and then using a type check here

packages/webdriverio/src/commands/browser/react$.ts
```diff
@@ -38,13 +38,18 @@
...
-export default async function react$ (selector, { props = {}, state = {} } = {}) {
+export default async function react$ (
+    this: WebdriverIO.BrowserObject,
+    selector: string,
+    { props = {}, state = {} }: WebdriverIO.ReactSelectorOptions = {}
+) {
     await this.executeScript(resqScript.toString(), [])
     await this.execute(waitToLoadReact)
-    const res = await this.execute(react$Script, selector, props, state)
+    const res = await this.execute(react$Script, selector, props, state) as ElementReference
```
martinfrancois, line 52: see previous comment

packages/webdriverio/src/commands/browser/reloadSession.ts
```diff
@@ -39,7 +39,7 @@ export default async function reloadSession () {
         log.warn(`Suppressing error closing the session: ${err.stack}`)
     }
 
-    const ProtocolDriver = require(this.options.automationProtocol).default
+    const ProtocolDriver = require(this.options.automationProtocol!).default
```
martinfrancois, line 42: is that `!` intentional?

packages/webdriverio/src/commands/browser/setTimeout.ts
```diff
@@ -43,19 +43,19 @@ export default async function setTimeout(timeouts) {
...
 
     /**
      * JsonWireProtocol action
      */
     if (!this.isW3C) {
         return Promise.all([
-            isFinite(implicit) && this.setTimeouts('implicit', implicit),
-            isFinite(pageLoad) && this.setTimeouts('page load', pageLoad),
-            isFinite(script) && this.setTimeouts('script', script),
+            isFinite(implicit) && (this.setTimeouts as any)('implicit', implicit),
```
martinfrancois, line 54: consider refactoring `(this.setTimeouts as any)` into a const to reduce duplication or even better if it's possible, change the type of `this.setTimeouts`

packages/webdriverio/src/commands/browser/url.ts
```diff
@@ -44,5 +44,5 @@ export default function url (path) {
         path = nodeUrl.resolve(this.options.baseUrl, path)
     }
 
-    return this.navigateTo(validateUrl(path))
+    return this.navigateTo(validateUrl(path, null))
```
martinfrancois, line 47: we could consider changing `this.navigateTo` to accept `undefined` as the second parameter, then we could leave this as it was

packages/webdriverio/src/utils/getElementObject.ts
```diff
@@ -87,18 +88,18 @@ export const getElements = function getElements(selector, elemResponse, isReactE
...
                  */
                 if (this.isW3C) {
-                    client[ELEMENT_KEY] = elementId
+                    (client as any)[ELEMENT_KEY] = elementId
                 } else {
                     client.ELEMENT = elementId
                 }
             } else {
-                client.error = res
+                client.error = res as unknown as Error
```
martinfrancois, line 96: it would be better to type  `res` as `... | Error`, then you don't need the cast to `as unknown` first

martinfrancois (changes requested):
Looks good in general, a lot better than what it was before. Most of the functions are still missing return types, would be great to also have them! Feel free to implement whatever you feel appropriate for this PR and any leftovers in following PR's as you mentioned.

packages/wdio-repl/src/index.ts
```diff
@@ -52,15 +52,15 @@ export default class WDIORepl {
...
+    constructor(config?: ReplConfig) {
         this._config = Object.assign(
             DEFAULT_CONFIG,
             { eval: this.eval.bind(this) },
             config
         )
     }
 
-    eval (cmd: string, context: vm.Context, filename: string, callback: (err: Error | null, result: any) => void) {
+    eval (cmd: string, context: vm.Context, filename: string | undefined, callback: (err: Error | null, result: any) => void) {
```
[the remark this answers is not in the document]
christian-bromann: I think this didn't work because it complaint that there can only be optional parameter at the end of the parameter chain. Given that `callback` is required here I had to put `string | undefined` instead of using the optional notation.

packages/webdriver/webdriver.d.ts
```diff
@@ -626,6 +627,8 @@ declare namespace WebDriver {
         isAndroid: boolean;
         isMobile: boolean;
         isIOS: boolean;
+        isSauce: boolean;
```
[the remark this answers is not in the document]
christian-bromann: The parameter is used not only in the Sauce service.

packages/webdriverio/src/commands/browser/getCookies.ts
```diff
@@ -30,18 +30,17 @@
...
 
+    const namesList = !Array.isArray(names) ? [names] : names
+
     if (namesList.every(obj => typeof obj !== 'string')) {
         throw new Error('Invalid input (see https://webdriver.io/docs/api/browser/getCookies.html for documentation.')
     }
 
-    const allCookies = await this.getAllCookies()
-
+    const allCookies: WebDriver.Cookie[] = await this.getAllCookies() as unknown as WebDriver.Cookie[]
```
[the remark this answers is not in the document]
christian-bromann: `getAllCookies` comes from the protocol commands which are not typed. We would need to enhance the type generation for these first. If you follow the link you will get to the json wire protocol definiton:

```
        /**
         * [jsonwp]
         *
         * https://github.com/SeleniumHQ/selenium/wiki/JsonWireProtocol#get-sessionsessionidcookie
         */
        getAllCookies(): object[];
```

packages/webdriverio/src/commands/browser/getPuppeteer.ts
```diff
@@ -34,15 +34,26 @@ export default async function getPuppeteer () {
...
+        const majorVersion = parseInt(this.capabilities.browserVersion.split('.').shift() || '', 10)
         if (majorVersion >= 79) {
             const ffOptions = this.capabilities['moz:firefoxOptions']
-            const ffArgs = this.requestedCapabilities['moz:firefoxOptions'].args
+            const ffArgs = this.requestedCapabilities['moz:firefoxOptions']?.args
 
             const rdPort = ffOptions && ffOptions.debuggerAddress
                 ? ffOptions.debuggerAddress
-                : ffArgs[ffArgs.findIndex((arg) => arg === FF_REMOTE_DEBUG_ARG) + 1]
+                : ffArgs
```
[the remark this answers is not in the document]
christian-bromann: What does `?? null` mean at the end?

packages/webdriverio/src/commands/browser/mock.ts
```diff
@@ -117,7 +117,8 @@ export default async function mock (url, filterOptions) {
         })
         client.on(
             'Fetch.requestPaused',
-            NetworkInterception.handleRequestInterception(client, SESSION_MOCKS)
+            (NetworkInterception as unknown as typeof DevtoolsNetworkInterception)
```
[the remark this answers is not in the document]
christian-bromann: it would still complain

packages/webdriverio/src/commands/browser/reloadSession.ts
christian-bromann: Yes, automationProtocol is an optional value. In a different PR I will make these options required and replace usages where options is used as input with `Partial<Options>`

packages/webdriverio/src/commands/browser/url.ts
christian-bromann: the parameter is part of `validateUrl` which needs to get typed first

packages/webdriverio/package.json
```diff
@@ -47,7 +47,10 @@
   "types": "./webdriverio.d.ts",
   "typeScriptVersion": "3.7.3",
   "dependencies": {
+    "@types/archiver": "^3.1.1",
+    "@types/lodash.clonedeep": "^4.5.6",
```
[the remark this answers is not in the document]
christian-bromann: I would keep it as is for now and see that as an optimization for later

christian-bromann:
@martinfrancois thanks for the thorough review, I tried to address all your comments.

packages/webdriverio/src/commands/browser/$.ts
```diff
@@ -57,7 +57,7 @@ export default async function $ (this: WebdriverIO.BrowserObject, selector: stri
      * convert protocol result into WebdriverIO element
      * e.g. when element was fetched with `getActiveElement`
      */
-    if (selector && typeof (selector as ElementReference)[ELEMENT_KEY] === 'string') {
+    if (typeof selector !== 'string' && typeof (selector as ElementReference)[ELEMENT_KEY] === 'string') {
```
martinfrancois, line 60: Note that `typeof selector !== 'string'` only applies if `selector` is the primitive data type `string`, if you also want to check for all objects of type `String`, you would need to do:
```
!(typeof selector === 'string' || selector instanceof String)
```
(but at this point I would probably extract it to a `util` function)

martinfrancois (approved):
@christian-bromann you're welcome, thanks for taking the time to take all of my suggestions into account, well done! Looks good to me! 👍

tests/typings/webdriverio/config.ts
```diff
@@ -79,7 +79,7 @@ const config: WebdriverIO.Config = {
             logName: 'test.log',
             videoName: 'test.mp4'
         }
-    }],
+    }] as WebDriver.DesiredCapabilities[],
```
mgrybyk, line 82: why do we need to do this now?

tests/typings/webdriverio/async.ts
```diff
@@ -86,7 +86,7 @@ async function bar() {
 
     const executeResult = await browser.execute(function (x: number) {
         return x
-    }, 4)
+    }, 4) as number
```
mgrybyk, line 89: the additional cast should not be required!
image

scripts/templates/devtools.tpl.d.ts
```diff
@@ -2,10 +2,6 @@
 /// <reference types="webdriver"/>
 
 declare namespace WebDriver {
-    interface ClientOptions {
-        isDevTools: boolean;
```
mgrybyk, line 6: why have you deleted this one? It's devtools protocol specific, a way to identify which automation protocol is currently used, webdriver or devtools

packages/webdriverio/webdriverio.d.ts
```diff
@@ -25,7 +31,7 @@ declare namespace WebdriverIO {
          * The executed script is assumed to be synchronous and the result of evaluating the script is returned to
          * the client.
          */
-        execute: <T>(script: string | ((...arguments: any[]) => T), ...arguments: any[]) => Promise<T>;
+        execute: <T>(script: string | ExecuteFunction, ...arguments: any[]) => Promise<T>;
```
mgrybyk, line 34: ExecuteFunction should accept type otherwise browser.execute return value won't be set properly

packages/webdriverio/src/utils/getElementObject.ts
```diff
@@ -87,18 +88,18 @@ export const getElements = function getElements(selector, elemResponse, isReactE
...
             } else {
-                client.error = res
+                client.error = res as Error
             }
 
             client.selector = selector
             client.parent = this
             client.index = i
-            client.emit = this.emit.bind(this)
+            client.emit = this.emit?.bind(this)
```
mgrybyk, line 102: if `emit` always exist in `this` is it possible to explicitly define it?

packages/webdriverio/src/commands/browser/waitUntil.ts
```diff
@@ -68,7 +69,7 @@ export default function (
     const fn = condition.bind(this)
     let timer = new Timer(interval, timeout, fn, true)
 
-    return timer.catch((e) => {
+    return (timer as any).catch((e: Error) => {
```
mgrybyk, line 72: it might be better to define proper types for Timer

packages/webdriverio/src/commands/browser/uploadFile.ts
```diff
@@ -41,20 +41,15 @@ export default async function uploadFile (localPath) {
...
         archiver('zip')
-            .on('error', (err) => reject(err))
-            .on('data', (data) => zipData.push(data))
+            .on('error', (err: Error) => reject(err))
+            .on('data', (data: Uint8Array) => zipData.push(data))
             .on('end', () => this.file(Buffer.concat(zipData).toString('base64')).then(resolve, reject))
             .append(source, { name: path.basename(localPath) })
-            .finalize((err) => {
-                /* istanbul ignore next */
-                if (err) {
```
mgrybyk, line 55: is it ok that errors will be ignored?

packages/webdriverio/src/commands/browser/setTimeout.ts
```diff
@@ -43,19 +43,20 @@ export default async function setTimeout(timeouts) {
 
     const implicit = timeouts.implicit
     // Previously also known as `page load` with JsonWireProtocol
-    const pageLoad = timeouts['page load'] || timeouts.pageLoad
+    const pageLoad = (timeouts as any)['page load'] || timeouts.pageLoad
     const script = timeouts.script
+    const setTimeouts: any = this.setTimeouts.bind(this)
```
mgrybyk, line 48: should setTimeout and timeouts exist in the browser object?

packages/webdriverio/src/commands/browser/custom$.ts
```diff
@@ -43,7 +44,7 @@ async function custom$ (strategyName, strategyArguments) {
     }
 
     if (res && typeof res[ELEMENT_KEY] === 'string') {
-        return await getElement.call(this, strategy, res)
+        return await getElement.call(this, strategy.toString(), res)
```
mgrybyk, line 47: shouldn't strategy be a string already?
(same for custom$$)

mgrybyk (changes requested):

tests/typings/webdriverio/config.ts
christian-bromann: because `capabilities` are defined as:

```ts
capabilities?: WebDriver.DesiredCapabilities[] | MultiRemoteCapabilities;
```

so this says that it is the first type and not multi remote caps.

packages/webdriverio/src/commands/browser/custom$.ts
christian-bromann: nope, it is from type function as we receive it from a stored selector strategy (see line 29)

packages/webdriverio/src/commands/browser/setTimeout.ts
christian-bromann: implementations between JSONWireProtocol and WebDriver are different which is impossible with the current setup to differentiate

packages/webdriverio/src/commands/browser/uploadFile.ts
christian-bromann: yes, TS has shown that `finalize` doesn't take callbacks anymore, so this is a overdue clean up

packages/webdriverio/src/commands/browser/waitUntil.ts
christian-bromann: definitely, in a separate PR though

mgrybyk: Ok

scripts/templates/devtools.tpl.d.ts
christian-bromann: it is already defined in `webdriver.d.ts`

christian-bromann:
@mgrybyk addressed your comments, please have another review, thanks!

packages/webdriver/src/types.ts
```diff
@@ -314,6 +315,13 @@ export interface VendorExtensions extends EdgeCapabilities {
 
     'goog:chromeOptions'?: ChromeOptions;
     'moz:firefoxOptions'?: FirefoxOptions;
+    'ms:edgeOptions'?: MicrosoftEdgeOptions;
+    'ms:edgeChromium'?: MicrosoftEdgeOptions;
+
+    // Safari specific
+    'safari.options'?: {
```
erwinheitzman, line 322:
```suggestion
    'safari.options'?: object
```

packages/webdriverio/src/commands/browser/newWindow.ts
```diff
@@ -56,6 +56,10 @@ export default async function newWindow (url, { windowName = 'New Window', windo
     const tabs = await this.getWindowHandles()
     const newTab = tabs.pop()
 
+    if (!newTab) {
```
erwinheitzman, line 59: If no new tab is created, it would always return the current tab right? So I think this check would have to be changed to the following.

```js
const originalTabs = await this.getWindowHandles()

await this.execute(newWindowHelper, url, windowName, windowFeatures)

const newTabs = await this.getWindowHandles()

if (newTabs.length !== originalTabs.length + 1) {
    throw new Error('No window handle was found to switch to')
}
```

packages/webdriverio/src/commands/browser/setTimeout.ts
```diff
@@ -43,19 +43,20 @@ export default async function setTimeout(timeouts) {
 
     const implicit = timeouts.implicit
     // Previously also known as `page load` with JsonWireProtocol
-    const pageLoad = timeouts['page load'] || timeouts.pageLoad
+    const pageLoad = (timeouts as any)['page load'] || timeouts.pageLoad
```
erwinheitzman, line 46: Shouldn't we extend the WebDriver.Timeout with our own types?

packages/webdriverio/src/commands/browser/touchAction.ts
```diff
@@ -59,6 +59,6 @@
 
 import { touchAction as touchActionCommand } from '../constant'
 
-export default function touchAction (...args) {
-    return touchActionCommand.apply(this, args)
+export default function touchAction (this: WebdriverIO.BrowserObject, ...args: WebdriverIO.TouchAction[]) {
```
erwinheitzman, line 62: Shouldn't this be called `WebdriverIO.TouchActionArgs` instead of `WebdriverIO.TouchAction`?

packages/webdriverio/src/types.ts
```diff
@@ -0,0 +1,35 @@
+export type ElementReferenceId = 'element-6066-11e4-a52e-4f735466cecf'
+
+export type ElementReference = Record<ElementReferenceId, string>
+
+export interface ElementObject extends ElementReference, WebdriverIO.BrowserObject {
+    /**
+     * WebDriver element reference
+     */
+    elementId: string
```
erwinheitzman, line 9: wasn't one of these the JSON Wire reference?

packages/webdriverio/src/commands/browser/newWindow.ts
christian-bromann: This error check is making sure that there is at least a tab to also make TS happy. I don't want to add any additional checks within this PR.

packages/webdriverio/src/commands/browser/setTimeout.ts
christian-bromann: `page load` is a JSONWireProtocol relict which I would not like to add to the typings given that we soon get rid of it anyway

packages/webdriverio/src/commands/browser/touchAction.ts
christian-bromann: I guess it should be called `TouchActionOptions`, also something I would not like to change as part of this PR

packages/webdriverio/src/types.ts
christian-bromann: `ELEMENT` is, yeah!

christian-bromann:
Any further quesions or concerns?

packages/webdriverio/src/utils/getElementObject.ts
```diff
@@ -87,12 +88,12 @@ export const getElements = function getElements(selector, elemResponse, isReactE
                  * set element id with proper key so element can be passed into execute commands
                  */
                 if (this.isW3C) {
-                    client[ELEMENT_KEY] = elementId
+                    (client as any)[ELEMENT_KEY] = elementId
```
erwinheitzman, line 91: I'd say changing the ELEMENT_KEY in the constants file could be part of this

packages/webdriverio/tests/commands/browser/call.test.js
```diff
@@ -20,6 +20,12 @@ describe('call command', () => {
     })
 
     it('should not fail if nothing is applied', () => {
-        browser.call()
+        expect(() => browser.call())
```
erwinheitzman, line 23: should not fail, but you expect it to throw

erwinheitzman (dismissed):
Some remarks, looking solid overall 👍

packages/webdriverio/tests/commands/browser/call.test.js
christian-bromann: if you don't pass a function in it is indeed suppose to fail

christian-bromann:
@erwinheitzman thanks for the review, I tried to address all your comments

packages/webdriverio/src/commands/browser/$.ts
mgrybyk: it doesn't matter in this case

packages/webdriverio/src/commands/browser/custom$.ts
mgrybyk: anyway, casting it to string doesn't make much sense. `strategy` should be a string otherwise we should throw an error, imo @christian-bromann

packages/webdriverio/src/commands/browser/debug.ts
```diff
@@ -33,14 +32,14 @@
...
-export default function debug(commandTimeout = 5000) {
+export default function debug(this: WebdriverIO.BrowserObject, commandTimeout = 5000) {
     const repl = new WDIORepl()
     const { introMessage } = WDIORepl
 
     /**
      * run repl in standalone mode
      */
-    if (!process.env.WDIO_WORKER) {
+    if (!process.env.WDIO_WORKER || typeof process.send !== 'function') {
```
mgrybyk, line 42: is it intentional? Didn't quite get the logic

mgrybyk (approved):
LGTM
please don't forget to deal with other review comments first ;)

packages/webdriverio/src/commands/browser/debug.ts
christian-bromann: `process.send` is not available in parent processes which got recognised by TS so I had to add this little check.

christian-bromann:
> please don't forget to deal with other review comments first ;)

Not sure what you mean here. I addressed all comments afaik.

PR #5999, opened 2020-10-17
Refactor utils interception webdriver to typescript

Fabianopb:
## Proposed changes

Refactoring `webdriverio/utils/interception/webdriver.js` to TypeScript.

## Types of changes

- [x] Refactor from JS to TS

## Checklist

### DIFF
--- a/packages/webdriver/webdriver.d.ts
+++ b/packages/webdriver/webdriver.d.ts
@@ -2504,21 +2504,21 @@ declare namespace WebDriver {
          * Receive request information about requests that match the mocked resource.
          * https://wiki.saucelabs.com/display/DOCS/Custom+Sauce+Labs+WebDriver+Extensions+for+Network+and+Log+Commands
          */
-        getMockCalls(mockId: string): ProtocolCommandResponse;
+        getMockCalls(mockId: number): ProtocolCommandResponse;
 
         /**
          * [saucelabs]
          * Clear list of mock calls.
          * https://wiki.saucelabs.com/display/DOCS/Custom+Sauce+Labs+WebDriver+Extensions+for+Network+and+Log+Commands
          */
-        clearMockCalls(mockId: string, restore: boolean): void;
+        clearMockCalls(mockId: number, restore: boolean): void;
 
         /**
          * [saucelabs]
          * Respond if mock matches a specific resource.
          * https://wiki.saucelabs.com/display/DOCS/Custom+Sauce+Labs+WebDriver+Extensions+for+Network+and+Log+Commands
          */
-        respondMock(mockId: string, payload: object): void;
+        respondMock(mockId: number, payload: object): void;
     }
 
     // selenium types
--- a/packages/webdriver/webdriver.d.ts
+++ b/packages/webdriver/webdriver.d.ts
@@ -2511,7 +2511,7 @@ declare namespace WebDriver {
          * Clear list of mock calls.
          * https://wiki.saucelabs.com/display/DOCS/Custom+Sauce+Labs+WebDriver+Extensions+for+Network+and+Log+Commands
          */
-        clearMockCalls(mockId: number, restore: boolean): void;
+        clearMockCalls(mockId: number, restore?: boolean): void;
 
         /**
          * [saucelabs]
--- a/packages/webdriver/webdriver.d.ts
+++ b/packages/webdriver/webdriver.d.ts
@@ -2504,21 +2504,21 @@ declare namespace WebDriver {
          * Receive request information about requests that match the mocked resource.
          * https://wiki.saucelabs.com/display/DOCS/Custom+Sauce+Labs+WebDriver+Extensions+for+Network+and+Log+Commands
          */
-        getMockCalls(mockId: number): ProtocolCommandResponse;
+        getMockCalls(mockId: string): ProtocolCommandResponse;
 
         /**
          * [saucelabs]
          * Clear list of mock calls.
          * https://wiki.saucelabs.com/display/DOCS/Custom+Sauce+Labs+WebDriver+Extensions+for+Network+and+Log+Commands
          */
-        clearMockCalls(mockId: number, restore?: boolean): void;
+        clearMockCalls(mockId: string, restore?: boolean): void;
 
         /**
          * [saucelabs]
          * Respond if mock matches a specific resource.
          * https://wiki.saucelabs.com/display/DOCS/Custom+Sauce+Labs+WebDriver+Extensions+for+Network+and+Log+Commands
          */
-        respondMock(mockId: number, payload: object): void;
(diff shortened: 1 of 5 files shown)


packages/webdriverio/src/utils/interception/webdriver.ts
```diff
@@ -8,6 +8,8 @@ import { ERROR_REASON } from '../../constants'
  * compliant backend.
  */
 export default class WebDriverInterception extends Interception {
+    mockId: number = NaN;
```
Fabianopb, line 11: Should this be initialized somehow here or then should we check that `mockId !== undefined` before calling the methods that require mockId?

packages/webdriverio/src/utils/interception/webdriver.ts
```diff
@@ -16,6 +18,7 @@ export default class WebDriverInterception extends Interception {
     /**
      * allows access to all requests made with given pattern
      */
+    // @ts-ignore
```
Fabianopb, line 21: Couldn't figure out yet how to match with the `calls` type from the class `Interception` (should it match?)

Also this gives the error:
```
'calls' is defined as a property in class 'Interception', but is overridden here in 'WebDriverInterception' as an accessor.ts(2611)
```

packages/webdriverio/webdriverio.d.ts
```diff
@@ -37,6 +44,12 @@ declare namespace WebdriverIO {
          * to this callback will be returned to the client.
          */
         executeAsync: (script: string | ((...arguments: any[]) => void), ...arguments: any[]) => Promise<any>;
+
+        mockCalls(mockId: number): Promise<void>;
+
+        mockRequest(url: string, filterOptions: MockFilterOptions): { mockId: number };
+
+        respondMock(mockId: number, payload: object): Promise<void>;
```
Fabianopb, line 52: Is this the right way to go, adding the missing types here? If yes, any tips how to document them?

Fabianopb: Hummm... looks like this is not the right approach? Some tests failed in another module... 🤔

Fabianopb:
@christian-bromann if you could please provide a little guidance in this one! 🙏

packages/webdriver/webdriver.d.ts
```diff
@@ -2504,21 +2504,21 @@ declare namespace WebDriver {
          * Receive request information about requests that match the mocked resource.
          * https://wiki.saucelabs.com/display/DOCS/Custom+Sauce+Labs+WebDriver+Extensions+for+Network+and+Log+Commands
          */
-        getMockCalls(mockId: string): ProtocolCommandResponse;
+        getMockCalls(mockId: number): ProtocolCommandResponse;
```
Fabianopb, line 2507: Tests require `number` so should this be changed?

packages/webdriverio/webdriverio.d.ts
Fabianopb: Outdated, figured out already

Fabianopb:
I can't add reviewers to this PR, perhaps @mgrybyk ?

packages/webdriver/webdriver.d.ts
christian-bromann: it should be a string if these commands have been defined as strings before

packages/webdriverio/src/utils/interception/webdriver.ts
christian-bromann: it should be defined as optional

packages/webdriver/webdriver.d.ts
Fabianopb: Then I guess all tests here have to be fixed to assert agains string, right?
https://github.com/webdriverio/webdriverio/blob/master/packages/webdriverio/tests/utils/interception/webdriver.test.js#L32

packages/webdriverio/src/utils/interception/webdriver.ts
Fabianopb: Then how do you suggest dealing with those?
image

packages/webdriver/webdriver.d.ts
Fabianopb: Apparently not, I might have made some confusion while I was developing, changed to string now and everything worked

packages/webdriverio/src/utils/interception/webdriver.ts
Fabianopb: I could cast, maybe

erwinheitzman:
Thank you for your contribution.

I prefer to fully migrate packages, not parts of packages.

The story is also to migrate the webdriverio package, so I'm not sure what to think of this PR at this moment.
I think migrating only parts of packages will make it chaotic and less obvious on what is and is not yet migrated, and it will make it hard for others to start working on things.

christian-bromann:
@erwinheitzman it is actually the way I recommended it. Porting the whole package just creates a too large PR to review. It makes sense to me to find little portions to do it step by step. The interception logic seems like a good start.

Fabianopb:
> @erwinheitzman it is actually the way I recommended it. Porting the whole package just creates a too large PR to review. It makes sense to me to find little portions to do it step by step. The interception logic seems like a good start.

I actually second this approach, I'm currently working in a huge repo where a few dozens of devs contribute and we're migrating it to TS. The only feasible way so far has been to do it step by step! :)

Also, as someone who's still getting acquainted with webdriverio project, I feel more confident to do it in small steps!

christian-bromann (changes requested):
The implementation of this file is not quite correct (it is actually also not yet supported). All `browser.mockCalls(...)` calls need to be `browser.getMockCalls(...)`. The command is implemented in the Sauce Labs WebDriver protocol.

erwinheitzman:
@Fabianopb first of all my apologies if it felt like I was criticizing you, I am very glad to have you contribute to the project! ❤️

I thought it might come with some risks but seems like it's actually rather easy to migrate like this so I'm all for it, thanks 👍

Fabianopb:
> @Fabianopb first of all my apologies if it felt like I was criticizing you, I am very glad to have you contribute to the project! ❤️
> 
> I thought it might come with some risks but seems like it's actually rather easy to migrate like this so I'm all for it, thanks 👍

No biggie! I just wanted to share my two cents about own experience in a similar case! :)

Fabianopb:
> The implementation of this file is not quite correct (it is actually also not yet supported). All `browser.mockCalls(...)` calls need to be `browser.getMockCalls(...)`. The command is implemented in the Sauce Labs WebDriver protocol.

Understood, I have now fixed the implementation and related test.

christian-bromann (approved):
Would be nice to find a way to remove the `@ts-ignore` at some point but let's move forward, the rest looks good 👍

Issue #5843, opened 2020-09-14
Tracking doc: rewrite `@wdio/sauce-service` package into TypeScript


christian-bromann:
This is a tracking issue for organising the effort of re-writing the code base from vanilla JS into TypeScript. If you are reading this and are interested getting involved, this is a great chance to do so. Here are the rules:

- if you are interested converting source files from `/packages/wdio-sauce-service` into TypeScript please comment on this thread and mention the files you want to convert to avoid duplicate work
- there is no reason to convert the whole package at once - smaller PRs make it easier to review the code
- if you make a PR, reference this issue so we can keep track of the progress
- if you make a PR, label it with `TypeScript` and the package name `wdio-sauce-service` so we can easier find all PRs
- if you make a PR, assign it to the `TypeScript Rewrite` project

Happy rewriting 😅

martinfrancois:
Hey @christian-bromann mind if I start working on this one?

christian-bromann:
Assigned you already to `@wdio/sync` 😉

martinfrancois:
@christian-bromann I just noticed that in the `README` for the sauce service, it mentions an option `noAutoconnect` for the  `sauceConnectOpts` and even though `noAutoconnect` is not in the types file or on https://github.com/bermi/sauce-connect-launcher, it is set inside `onPrepare` in `launcher.js`.

Can you please check whether `noAutoconnect` was removed (and thus should also be removed from `onPrepare` and the `README`) or if it was forgotten to be added to the types and to https://github.com/bermi/sauce-connect-launcher ?

Thanks!

martinfrancois:
I also just noticed that `sePort` is used in `launcher.js` on `sauceConnectOpts`, but I couldn't find it either in the `README` or on the linked github repo in the comment above, can you please also check if that one still exists?

martinfrancois:
Actually nevermind - I just found the `saucelabs` dependency already provides `SauceConnectOptions`, I'm removing the types from the package and replace them with the ones from the dependency, since they seem to be more up to date and I also don't think it makes sense to duplicate them.

martinfrancois:
@christian-bromann in the `node-saucelabs` dependency (xxx)  `noAutodetect` in the interface `SauceConnectOptions` is defined as a `string`:
```
  /**
   * Disable the autodetection of proxy settings.
   */
  noAutodetect?: string;
```
[...]

PR #6001, opened 2020-10-17
Tracking doc: rewrite `@wdio/sauce-service` package into TypeScript

martinfrancois:
## Proposed changes

Rewrite `@wdio/sauce-service` package into TypeScript, see #5843

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments


packages/wdio-sauce-service/src/launcher.ts
```diff
@@ -0,0 +1,121 @@
...
+        this._options = value
+    }
+
+    set api(value: SauceLabs) {
+        this._api = value
+    }
+
+    set sauceConnectProcess(value: SauceConnectInstance | undefined) {
+        this._sauceConnectProcess = value
+    }
```
christian-bromann, line 120: What is the reason for using getters/setters instead of public variables?

packages/wdio-sauce-service/tests/service.test.ts
```diff
@@ -60,13 +63,13 @@ jest.mock('../src/utils', () => {
...
 test('before should call isUnifiedPlatform', () => {
-    const service = new SauceService()
+    const service: any = new SauceService()
     service.before()
     expect(isUnifiedPlatform).toBeCalledTimes(1)
 })
 
 test('beforeSuite', () => {
-    const service = new SauceService()
+    const service: any = new SauceService()
```
christian-bromann, line 72: Why do we use `any` here instead of instance of SauceService?

packages/wdio-utils/package.json
```diff
@@ -27,5 +27,6 @@
   },
   "publishConfig": {
     "access": "public"
-  }
+  },
+  "types": "utils.d.ts"
```
christian-bromann, line 31: Can't we just point to `./build/index.d.ts`?

packages/wdio-utils/utils.d.ts
```diff
@@ -0,0 +1 @@
+export * from './src/index'
```
christian-bromann, line 1: we don't push `src` files to NPM

packages/wdio-sauce-service/src/launcher.ts
martinfrancois: Good question, I think that's highly debatable, in languages like Java and C# they ate considered to be best practice, which is why they were added to JS in the first place. Used like this there isn't really a difference, but it could allow for example to return something in a different format, to return a copy of an object to keep encapsulation etc. For the sake of converting it to TS I just added them as is to not have breaking changes, but it could allow further customization for future properties.

christian-bromann: Got it. Given that we don't do this across the code base, let's remove the getters/setters and use public properties instead.

martinfrancois: Alright!

packages/wdio-utils/package.json
martinfrancois: Unfortunately not no, we then get the following error during `npm run setup-full` on `npx tsc -b ...`:
```
error TS2688: Cannot find type definition file for '@wdio/utils'.
```

packages/wdio-utils/utils.d.ts
martinfrancois: I don't quite understand what you mean... I had to do this because else it was always complaining about missing type declarations... Maybe you have a better suggestion here?

packages/wdio-utils/package.json
christian-bromann: we just moved this package over to typescript and it has types defined now

martinfrancois: Ah great, let me rebase my branch then.

packages/wdio-utils/utils.d.ts
christian-bromann: you should have types now if you rebase on master

packages/wdio-utils/package.json
martinfrancois: still getting the error:
```
error TS2688: Cannot find type definition file for '@wdio/utils'.
```
(see builds here)

martinfrancois:
After rebasing I still get the error:

```
error TS2688: Cannot find type definition file for '@wdio/utils'.
```
(see builds here)

Any idea @christian-bromann ?

martinfrancois:
@christian-bromann I think it could be because of the order in which we build the typescript files here, as generated by `./scripts/build`:
```
npx tsc -b packages/devtools/tsconfig.json packages/wdio-allure-reporter/tsconfig.json packages/wdio-appium-service/tsconfig.json packages/wdio-applitools-service/tsconfig.json packages/wdio-browserstack-service/tsconfig.json packages/wdio-cli/tsconfig.json packages/wdio-concise-reporter/tsconfig.json packages/wdio-config/tsconfig.json packages/wdio-crossbrowsertesting-service/tsconfig.json packages/wdio-cucumber-framework/tsconfig.json packages/wdio-devtools-service/tsconfig.json packages/wdio-dot-reporter/tsconfig.json packages/wdio-firefox-profile-service/tsconfig.json packages/wdio-jasmine-framework/tsconfig.json packages/wdio-junit-reporter/tsconfig.json packages/wdio-local-runner/tsconfig.json packages/wdio-logger/tsconfig.json packages/wdio-mocha-framework/tsconfig.json packages/wdio-repl/tsconfig.json packages/wdio-reporter/tsconfig.json packages/wdio-runner/tsconfig.json packages/wdio-sauce-service/tsconfig.json packages/wdio-selenium-standalone-service/tsconfig.json packages/wdio-shared-store-service/tsconfig.json packages/wdio-smoke-test-reporter/tsconfig.json packages/wdio-smoke-test-service/tsconfig.json packages/wdio-spec-reporter/tsconfig.json packages/wdio-static-server-service/tsconfig.json packages/wdio-sumologic-reporter/tsconfig.json packages/wdio-sync/tsconfig.json packages/wdio-testingbot-service/tsconfig.json packages/wdio-utils/tsconfig.json packages/wdio-webdriver-mock-service/tsconfig.json packages/webdriver/tsconfig.json packages/webdriverio/tsconfig.json
```
At least I noticed if I run the command `npm run setup-full` first, then when it fails I run `node ./scripts/build` I don't get the error anymore. So I suspect it could be that since `wdio-utils` is built AFTER `wdio-sauce-service` is built, but since `wdio-sauce-service` depends on `wdio-utils`, it throws an error since it doesn't have the types.

martinfrancois:
@christian-bromann sorry for the spam, but I now added a commit (https://github.com/webdriverio/webdriverio/pull/6001/commits/79201d30084d3a2763ad09a3cf6d75b054585117) where I made it possible to define packages that should be built first, to avoid the issue I described in https://github.com/webdriverio/webdriverio/pull/6001#issuecomment-713911010

Should I open up a separate PR for that or should we keep it in this one?

packages/wdio-sauce-service/src/utils.ts
```diff
@@ -66,12 +68,14 @@ export function makeCapabilityFactory(tunnelIdentifier, options) {
...
-            tunnelIdentifier
-        )
+        const sauceOptions: SauceConnectOptions | undefined = !isLegacy && !isUnifiedPlatform(capability) ? capability['sauce:options'] : capability
+        if (sauceOptions) {
+            sauceOptions.tunnelIdentifier = (
+                capability.tunnelIdentifier ||
+                sauceOptions.tunnelIdentifier ||
+                tunnelIdentifier
+            )
+        }
```
christian-bromann, line 78: Why do we check for `sauceOptions`? Seems that `capability['sauce:options']` can be undefined which is handled a bit above. We should remove the check otherwise there is a chance that `tunnelIdentifier` is not being applied.

packages/wdio-sauce-service/tests/service.test.ts
christian-bromann: any comment?

christian-bromann:
Good catch with the compile order. This could complain some other side issues I experienced.

packages/wdio-sauce-service/src/launcher.ts
```diff
@@ -13,15 +15,19 @@ const SC_RELAY_DEPCRECATION_WARNING = [
...
         this.options = options
-        this.api = new SauceLabs(config)
+        this.api = new SauceLabs(config as SauceLabsOptions)
     }
 
     /**
      * modify config and launch sauce connect
      */
-    async onPrepare (config, capabilities) {
+    async onPrepare (config: SauceLabsOptions, capabilities: DesiredCapabilities | Array<DesiredCapabilities>) {
```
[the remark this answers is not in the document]
martinfrancois: `DesiredCapabilities | Array<DesiredCapabilities>` is equivalent to `DesiredCapabilities | DesiredCapabilities[]`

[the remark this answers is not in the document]
christian-bromann: Sorry, I wasn't clear. My question was more about why it is either an array or not an array, it should be either the one or the other, no?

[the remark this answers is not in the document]
martinfrancois: Ahhh now I understand, you're right, it should be an array!

packages/wdio-sauce-service/tests/service.test.ts
martinfrancois: Good catch, I probably accidentally set that one here, I just changed it.

martinfrancois:
@christian-bromann just resolved your comments, thanks!

packages/wdio-sauce-service/src/service.ts
```diff
@@ -1,13 +1,78 @@
-import SauceLabs from 'saucelabs'
+import SauceLabs, { Job, SauceConnectOptions, SauceLabsOptions } from 'saucelabs'
 import logger from '@wdio/logger'
 import { isUnifiedPlatform } from './utils'
+import SuiteStats from '../../wdio-reporter/build/stats/suite'
```
christian-bromann, line 4: We can't access modules like that, it would fail when installed through NPM

```suggestion
import { SuiteStats } from '@wdio/reporter'
```

packages/wdio-sauce-service/src/service.ts
```diff
@@ -1,13 +1,78 @@
...
+import { Suite as ReporterSuite }  from '../../wdio-reporter/build/stats/suite'
```
christian-bromann, line 5: same

packages/wdio-sauce-service/sauce-service.d.ts
```diff
@@ -1,151 +0,0 @@
-declare module WebdriverIO {
```
christian-bromann, line 1: we can't remove this file as the package.json points to it https://github.com/webdriverio/webdriverio/pull/6001/files#diff-1e974f46805437badbcd51209a7a5b716c7c08ec4b9c1554ffcbc9ca9fa48ef3R45

packages/wdio-sauce-service/src/service.ts
```diff
@@ -166,52 +230,57 @@ export default class SauceService {
...
+        const browserName = this.multiRemoteBrowser.instances.filter(
+            (browserName: string) => (global.browser[browserName as keyof BrowserObject] as BrowserObject).sessionId === newSessionId)[0]
         log.info(`Update (reloaded) multiremote job for browser "${browserName}" and sessionId ${oldSessionId}, ${status}`)
         return this.updateJob(oldSessionId, this.failures, true, browserName)
     }
 
-    async updateJob (sessionId, failures, calledOnReload = false, browserName) {
-        if (this.isRDC) {
+    async updateJob (sessionId: string, failures: number, calledOnReload = false, browserName?: string) {
+        if (this.api && this.isRDC) {
```
christian-bromann, line 266:
```suggestion
        if (!this.api) {
            return
        }

        if (this.isRDC) {
```

packages/wdio-sauce-service/src/service.ts
```diff
@@ -166,52 +230,57 @@ export default class SauceService {
...
             await this.api.updateTest(sessionId, { passed: failures === 0 })
             this.failures = 0
             return
         }
 
         const body = this.getBody(failures, calledOnReload, browserName)
-        await this.api.updateJob(this.config.user, sessionId, body)
+        if (this.api && this.config) {
```
christian-bromann, line 273: It feels weird to check about this given that it will be always set. Maybe we can just ignore type checking here and have

```js
await this.api.updateJob(this.config?.user, sessionId, body as Job)
```

```suggestion
        if (this.config) {
```

packages/wdio-sauce-service/sauce-service.d.ts
martinfrancois: I didn't remove it though - only its contents, and I did so since they are outdated (and now included in `node-saucelabs`, do we need to do anything differently here?

packages/wdio-sauce-service/src/service.ts
martinfrancois: I wanted to do this as well, but since `updateJob` has the following signature: `updateJob(username: string, id: string, body: Job)` it will tell me that `this.conifg?.user` can be `undefined`, but the signature doesn't accept `undefined`, which means I would need to define some sort of a default, which would be even worse...

martinfrancois:
@christian-bromann can you have a look again please? Seems like only one question from you that I answered is left.

packages/wdio-sauce-service/package.json
```diff
@@ -23,13 +23,23 @@
...
   },
+  "devDependencies": {
+    "webdriverio": "6.7.0",
+    "webdriver": "6.7.0",
+    "@wdio/cucumber-framework": "6.7.0",
+    "@wdio/jasmine-framework": "6.7.0",
+    "@wdio/reporter": "6.7.0",
+    "@types/cucumber": "6.0.1",
+    "cucumber": "6.0.5"
+  },
```
christian-bromann, line 42: This will be problematic if we release. As we are in a production environment it won't install dev dependencies and therefor the compile step will fail. I would suggest:

- remove `cucumber`
- move `@types/cucumber` into the dependencies
- reference all wdio types through tsconfig.json and use them through their namespace, e.g. WebDriver.DesiredCapabilities rather than importing them.

WDYT?

martinfrancois: Sounds good, I'll try it and see if it works and report back if I have issues.

packages/wdio-sauce-service/src/service.ts
```diff
@@ -166,52 +231,61 @@ export default class SauceService {
...
-        if (!global.browser.isMultiremote) {
+        if (!this.multiRemoteBrowser.isMultiremote) {
             log.info(`Update (reloaded) job with sessionId ${oldSessionId}, ${status}`)
             return this.updateJob(oldSessionId, this.failures, true)
         }
 
-        const browserName = global.browser.instances.filter(
-            (browserName) => global.browser[browserName].sessionId === newSessionId)[0]
+        const browserName = this.multiRemoteBrowser.instances.filter(
+            (browserName: string) => (global.browser[browserName as keyof BrowserObject] as BrowserObject).sessionId === newSessionId)[0]
```
hieuxlu, line 261: I would say this as quite hacky @christian-bromann. `browserName` is definitely not in the list of `keyof BrowserObject`

packages/wdio-sauce-service/src/service.ts
```diff
@@ -1,13 +1,79 @@
...
+     */
+    setJobNameInBeforeSuite?: boolean;
+}
+
+export type SauceConfig = SauceLabsOptions & WebdriverIO.Config;
+
+interface MultiRemoteBrowser extends WebdriverIO.BrowserObject {
+    instances: string[]
+    isMultiremote: boolean
+}
```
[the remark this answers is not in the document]
martinfrancois: do they have the same kind of type annotation? If so, we could think about extracting it somewhere more global, but anyways we could always move this in a later PR into a different package when there is the need for it.

packages/wdio-sauce-service/src/service.ts
martinfrancois: Since the type of `global.browser` is `BrowserObject`, and you are trying to access a key from `global.browser` here, what else could it be than a key of `BrowserObject`?

packages/wdio-sauce-service/package.json
martinfrancois: @christian-bromann by `reference all wdio types through tsconfig.json` you did mean to do it like that, right?:
```
"types": ["webdriverio", "webdriver", "@wdio/jasmine-framework" ,"@wdio/cucumber-framework", "@wdio/mocha-framework", "@wdio/utils"]
```
Or did you mean something else?

Because I'm getting an error, see `launcher.ts:21`:
```
TS2503: Cannot find namespace 'Webdriver'.
TS4063: Parameter 'capabilities' of constructor from exported class has or is using private name 'Webdriver'.
```

christian-bromann:
@martinfrancois there still seem to be some legit compiling errors:

```
Error: packages/wdio-sauce-service/src/launcher.ts(21,62): error TS2503: Cannot find namespace 'Webdriver'.
Error: packages/wdio-sauce-service/src/launcher.ts(21,100): error TS2503: Cannot find namespace 'Webdriver'.
Error: packages/wdio-sauce-service/src/launcher.ts(29,68): error TS2304: Cannot find name 'DesiredCapabilities'.
```

martinfrancois:
@christian-bromann yes, I mentioned the reason for them here: https://github.com/webdriverio/webdriverio/pull/6001#discussion_r513028705

packages/wdio-sauce-service/sauce-service.d.ts
christian-bromann: let's keep this file for now because it is still being referred to in the package.json (`"types": "sauce-service.d.ts"`)

packages/wdio-sauce-service/package.json
christian-bromann: ok, I guess this solution doesn't work then and we have to use references like [Jest](https://github.com/facebook/jest/blob/master/packages/jest-diff/tsconfig.json#L7-L12) does.

martinfrancois: I googled a bit and came to the same conclusion, I also found this:
https://medium.com/@NiGhTTraX/how-to-set-up-a-typescript-monorepo-with-lerna-c6acda7d4559
https://github.com/NiGhTTraX/ts-monorepo

In this case, I guess we need to change the root `tsconfig.json` to the following, right?:
```
{
    "extends": "./tsconfig.prod.json",
    "compilerOptions": {
        "inlineSourceMap": true,
        "baseUrl": ".",
        "paths": {
            "@wdio/*": ["packages/wdio-*/src"],
            "devtools": ["packages/devtools/src"],
            "eslint-plugin-wdio": ["packages/eslint-plugin-wdio/src"],
            "webdriver": ["packages/webdriver/src"],
            "werbdriverio": ["packages/webdriverio/src"]
        }
    }
}
```
If the `build` command runs fine, does this indicate it will work in production as well?

packages/wdio-sauce-service/src/service.ts
hieuxlu: I think this is the case of accessing something like `Record<string, BrowserObject>`. If you use `<keyOf>`, it would be only declared properties of `BrowserObject`, so the typecheck would be meaning less, because `browserName` is not in the declared properties.

packages/wdio-sauce-service/src/service.ts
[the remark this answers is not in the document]
hieuxlu: Yes I agree it could be a separate issue. But we should think of some solutions now, nonetheless, or our repos will litter with duplicate types like this. I saw some similar problems when you want to cross reference types from other packages such as mocha framework, which extends Webdriverio types.

packages/wdio-sauce-service/package.json
hieuxlu: Did you try specifying `types` in `package.json`? I hope this can solve typescript monorepo problem
https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#including-declarations-in-your-npm-package

martinfrancois: @hieuxlu the source code is right there for you to check yourself, but yes I did so ;)

packages/wdio-sauce-service/src/service.ts
martinfrancois: > If you use <keyOf>, the property will be casted to any, so the typecheck would provide no value.

That's actually incorrect, have a look at the following example:
```ts
const myObj = {
  Hello: "world"
};
type ObjKey = keyof typeof myObj
const foo: ObjKey = "foo";
```
If what you are saying was correct, that `keyof` would cast it to `any`, then I would expect this example to compile without any errors (since `ObjKey` would be `any`). However, this results in the following error:
```
Initializer type "foo" is not assignable to variable type ObjKey
```
If I change the last line to be `const foo: ObjKey = "Hello";`, it compiles fine without any errors.
So clearly TypeScript preserves the types when using `keyof`.

But you are right in a way that by using `browserName as keyof BrowserObject` we are implicitly telling TypeScript to _assume_ that `browserName` is a key of `BrowserObject`, even though it comes from `this.multiRemoteBrowser.instances` (which is a plain array of strings, with no guarantee of those strings being `keyof BrowserObject`). To be 100% correct we would need to either type `instance` as `(keyof BrowserObject)[]` and or run checks on runtime to ensure that `instances` are indeed keys of `BrowserObject`. But this would limit the instances that can be used, which could get annoying to maintain (as the code needs to be changed for each new type of instance) and it would most likely be overkill here and the risk of breaking things would be much higher than the benefit of the type safety we would gain in this situation, in my opinion.

Considering the code has been running fine previously, I'm quite hesitant with adding any more checks here, we could always consider improving type safety here in the future, but probably this would be a bigger undertaking. Of course I could be totally missing something, in which case I would be glad if you could point it out or make suggestions on how to improve the situation.

packages/wdio-sauce-service/src/service.ts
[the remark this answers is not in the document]
martinfrancois: I personally like not to prematurely optimize things - my preferred approach is usually to adapt things to be as generic as they need to be, once it's needed. So I would rather start with having this `interface` here (as it's the only place it's used right now) and refactoring it to a different place as soon as the need for it arises. And where to put it then will be something to look at specifically in the context. Especially considering we are not exporting this type, so we don't run risk of any breaking changes when we move this to a different place.

christian-bromann:
@martinfrancois can we update the branch as have added a bunch of more rewrites to `master`.

I think the following way of handling external types works best: add `webdriverio` or any other sub package to the `"types"` properties in the subpackage tsconfig:

e.g. in `packages/devtools/tsconfig.json`:

```json
{
    "extends": "../../tsconfig",
    "compilerOptions": {
        "baseUrl": ".",
        "outDir": "./build",
        "rootDir": "./src",
        "types": ["webdriverio", "@wdio/protocols"],
    },
    "include": [
        "src/**/*"
    ]
}
```

Then you can use the types in all files via, e.g. in `packages/devtools/src/index.ts`

```ts
static async newSession (options: WebDriver.Options = {}, modifier?: Function, userPrototype = {}, customCommandWrapper?: Function) {
        const params: WebDriver.Options = validateConfig(DEFAULTS, options)
```

__Note:__ the difference between `Webdriver` (wrong namespace) and `WebDriver` (correct namespace). 

Happy to help in any way to get this landed.

christian-bromann:
@martinfrancois do you plan to continue working on this?

martinfrancois:
@christian-bromann excuse me, didn't notice your update here: https://github.com/webdriverio/webdriverio/pull/6001#issuecomment-727911281

Yes, I'll try your suggestion and will get back to you!

christian-bromann:
@martinfrancois we are aiming for a v7 release by end of January. Do you think you can push this over the line until then? If not, no problem. I am happy to pick this one up.

christian-bromann:
I will go ahead and take this over. I will use what you have done so far.

christian-bromann:
Closing in favor of #6274

Issue #6004, opened 2020-10-19
browser.mock respond - do not fetch response body


mgrybyk:
**Is your feature request related to a problem? Please describe.**
Add an option to not fetch the response body and set mocked data immediately.

Ideally, the backend should not be reached if possible.
Otherwise, we should not wait for a backend response.

**Describe the solution you'd like**
```js
const mock = browser.mock('url')
mock.respond('[]', { fetchResponse: false }) // true by default
```

mgrybyk:
@christian-bromann I thing that `fetchResponse` should be false by default.

PR #6006, opened 2020-10-19
webdriverio: mock respond in request stage

mgrybyk:
## Proposed changes

**Added**
Allow sending a mocked response in the Request stage. 
In other words, add an option to not wait for a backend to respond before sending a mocked response.

image

#6004

**Changed**
`abort` and `abortOnce` commands are now aborting a **request** and not waiting for a response.

## Types of changes

## Checklist

## Further comments

still have to add tests...

### DIFF
--- a/packages/webdriverio/src/commands/browser/mock.js
+++ b/packages/webdriverio/src/commands/browser/mock.js
@@ -84,14 +84,14 @@
  * </example>
  *
  * @alias browser.mock
- * @param {String}             url                            url to mock
- * @param {MockFilterOptions=} filterOptions                  filter mock resource by additional options
- * @param {String=}            filterOptions.method           filter resource by HTTP method
- * @param {Object=}            filterOptions.headers          filter resource by specific request headers
- * @param {Object=}            filterOptions.responseHeaders  filter resource by specific response headers
- * @param {Object=}            filterOptions.postData         filter resource by request postData
- * @param {Object=}            filterOptions.statusCode       filter resource by response statusCode
- * @return {Mock}                                             a mock object to modify the response
+ * @param {String}              url                             url to mock
+ * @param {MockFilterOptions=}  filterOptions                   filter mock resource by additional options
+ * @param {String|Function=}    filterOptions.method            filter resource by HTTP method
+ * @param {Object|Function=}    filterOptions.headers           filter resource by specific request headers
+ * @param {Object|Function=}    filterOptions.responseHeaders   filter resource by specific response headers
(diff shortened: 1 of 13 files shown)


packages/webdriverio/src/commands/browser/mock.js
```diff
@@ -89,6 +89,8 @@
  * @param {String=}            filterOptions.method           filter resource by HTTP method
  * @param {Object=}            filterOptions.headers          filter resource by specific request headers
  * @param {Object=}            filterOptions.responseHeaders  filter resource by specific response headers
+ * @param {Object=}            filterOptions.postData         filter resource by request postData
+ * @param {Object=}            filterOptions.statusCode       filter resource by response statusCode
```
mgrybyk, line 93: should be number or function

packages/webdriverio/src/utils/interception/devtools.js
```diff
@@ -13,13 +13,30 @@ const log = logger('webdriverio')
...
+                 * skip response mocks in Request stage
+                 */
+                if (isRequest && (
+                    mock.respondOverwrites.length === 0 || // nothing to do in Request stage
+                    (!mock.respondOverwrites[0].errorReason && // skip if not going to abort a request
+                    // or want to fetch response
+                    mock.respondOverwrites[0].params &&
+                    mock.respondOverwrites[0].params.fetchResponse !== false)
+                )) {
+                    continue
```
christian-bromann, line 37: If we continue here we don't continue the request with the mocked response we have set for the resource, no=

christian-bromann (commented):
If we skip doesn't it mean `'Fetch.continueRequest'`

packages/webdriverio/src/utils/interception/devtools.js
mgrybyk: Not now because of https://github.com/webdriverio/webdriverio/pull/6006/files#diff-c8a4111f05935473915df3949aab097fd9113d5daed85624222ab01bdc9eb90aR118

Listening for 2 events for each request:
1. when request is about to be sent
2. when response received (at this point we have response status code and headers, like before)

mgrybyk: There are two calls in the test as well https://github.com/webdriverio/webdriverio/pull/6006/files#diff-60adb4b9fe864361906b1b1169a2874bca2ec19cce0fc4dcfcfe1dac00f9b9fbR37

christian-bromann (approved):
I see, makes sense. Thanks! 👍

PR #6013, opened 2020-10-20
Changed workflow setup for docs

erwinheitzman:
## Proposed changes

Trigger the documents deployment workflow when a release has been made.

Note that I have never used the workflows like this so if anyone has more knowledge on this that would be great for the review.


christian-bromann (approved):
👍

PR #6020, opened 2020-10-22
allure-reporter: attach screenshot to 'all' hooks on failure

mgrybyk:
## Proposed changes

Add a screenshot to "before all" / "after all"  hooks on failure.

## Types of changes

## Checklist

## Further comments

not sure if it's a bug fix or not but for me it looks like a bug because failures in beforeEach and afterEach hooks are attached to the report.

Steps to reproduce:
1. add `disableMochaHooks: true,` to allure reporter config
2. trigger failure in before all / after all hook, ex:
```
  before('open login page', () => {
    $('body').waitForExist({ reverse: true, timeout: 100 })
  })
```


christian-bromann (approved):
👍

PR #6021, opened 2020-10-22
Refactor webdriverio utils interception devtools to TS

Fabianopb:
## Proposed changes

Refactor webdriverio utils interception devtools to TS

## Types of changes

- [x] TypeScript refactor

## Checklist

### DIFF
--- a/packages/webdriverio/src/utils/interception/devtools.ts
+++ b/packages/webdriverio/src/utils/interception/devtools.ts
@@ -87,15 +87,12 @@ export default class DevtoolsInterception extends Interception {
                     continue
                 }
 
-                try {
-                    const { body, base64Encoded = undefined } = isRequest ? { body: '' } : await client.send(
-                        'Fetch.getResponseBody',
-                        { requestId }
-                    )
-                    request.body = base64Encoded ? atob(body) : body
-                } catch (error) {
-                    // istanbul ignore next
-                }
+                const { body, base64Encoded = undefined } = isRequest ? { body: '' } : await client.send(
+                    'Fetch.getResponseBody',
+                    { requestId }
+                ).catch(/* istanbul ignore next */() => ({} as any))
+
+                request.body = base64Encoded ? atob(body) : body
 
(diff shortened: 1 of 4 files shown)


packages/webdriverio/src/utils/interception/devtools.ts
```diff
@@ -266,15 +298,15 @@ const canAccess = async (filepath) => {
...
         return JSON.parse(body) || body
     } catch {
         return body
     }
 }
 
-const logFetchError = (err) => {
+const logFetchError = (err?: Error) => {
     /* istanbul ignore next */
     log.debug(err && err.message ? err.message : err)
```
christian-bromann, line 311:
```suggestion
    log.debug(err?.message)
```

packages/webdriverio/src/utils/interception/devtools.ts
```diff
@@ -207,12 +239,12 @@ export default class DevtoolsInterception extends Interception {
...
      * @param {string} errorReason  error code of the response
      */
-    abortOnce (errorReason) {
+    abortOnce (errorReason: string) {
         this.abort(errorReason, false)
     }
 }
 
-const filterMethod = (method, expected) => {
+const filterMethod = (method: string, expected?: ((method: string) => boolean) | string) => {
```
christian-bromann, line 247: We can create an expected type that looks like this:

```ts
type ExpectParameter<T> = ((param: T) => boolean) | T
```

and then use it throughout the file like this:
```suggestion
const filterMethod = (method: string, expected?: ExpectParameter<string>) => {
```

christian-bromann (changes requested):

packages/webdriverio/src/utils/interception/devtools.ts
Fabianopb: Yeah, better like that, changed!

packages/webdriverio/src/utils/interception/devtools.ts
Fabianopb: Good suggestion! 👍 Changed!

christian-bromann (approved):
👍

PR #6023, opened 2020-10-22
Webdriverio refactor some lower hanging fruits in webdriverio to TS

Fabianopb:
## Proposed changes

Refactors `findStrategy`, `implicitWait`, `refetchElement`, `SevereServiceError` to TypeScript (files within `packages/webdriverio/src/utils`)

## Types of changes

- [x] TypeScript refactoring

## Checklist


Fabianopb:
Continuing https://github.com/webdriverio/webdriverio/issues/5854, I've done a batch this time, these seemed to be the easier ones in webdriverio utils.

Fabianopb:
Ops, I broke something 🤔 
```
FAIL packages/wdio-reporter/tests/test.stats.test.ts (1915 MB heap size)
  ● TestStats › when there's an end date, duration should be end time - start time

    expect(received).toBeGreaterThanOrEqual(expected)

    Expected: >= 11
    Received:    10

      107 |             expect(stat.end).toBeInstanceOf(Date)
      108 |             expect(stat.end!.getTime()).toBeGreaterThanOrEqual(stat.start!.getTime() + timeout)
    > 109 |             expect(stat.duration).toBeGreaterThanOrEqual(currentDuration + timeout)
          |                                   ^
      110 |             done()
      111 |         }, timeout)
      112 |     })

      at Timeout._onTimeout (packages/wdio-reporter/tests/test.stats.test.ts:109:35)
```

Fabianopb:
I'm a bit puzzled here why this test didn't pass ☝️

christian-bromann:
it's a flaky test, I will clean this up in a new PR.

Fabianopb:
Thanks @christian-bromann, rebased and pushed, everything looks good now!

packages/webdriverio/src/utils/findStrategy.ts
```diff
@@ -168,15 +168,22 @@ export const findStrategy = function (selector, isW3C, isMobile) {
     }
     case 'name': {
         if (isMobile || !isW3C) {
+            const match = selector.match(/^\[name=("|')([a-zA-z0-9\-_.@=[\] ']+)("|')]$/)
+            if (!match) {
+                throw new Error('InvalidSelectorStrategy') // ToDo: move error to wdio-error package
```
christian-bromann, line 173: Can we have a more explicit error message here? Also I don't think we need to move this error anywhere, at least I would not know where.

packages/webdriverio/src/utils/findStrategy.ts
```diff
@@ -168,15 +168,22 @@ export const findStrategy = function (selector, isW3C, isMobile) {
...
+            value = match[2]
         }
         break
     }
     case 'xpath extended': {
         using = 'xpath'
         const match = selector.match(new RegExp(XPATH_SELECTOR_REGEXP.map(rx => rx.source).join('')))
-        const PREFIX_NAME = { '.': 'class', '#': 'id' }
+        if (!match) {
+            throw new Error('InvalidSelectorStrategy') // ToDo: move error to wdio-error package
```
christian-bromann, line 184: same

packages/webdriverio/src/utils/implicitWait.ts
```diff
@@ -1,15 +1,22 @@
 import logger from '@wdio/logger'
 const log = logger('webdriverio')
 
+export type CurrentElement = {
```
christian-bromann, line 4: Can we create an issue for this to clean this up later?

christian-bromann (changes requested):
some comments

packages/webdriverio/src/utils/findStrategy.ts
Fabianopb: Sorry for the "todo", blind copy-paste effect.

I think "invalid selector strategy" describes well this, because if `match` is `null` it means that the selector didn't match any strategy, right? What do you think?

packages/webdriverio/src/utils/implicitWait.ts
Fabianopb: What's the idea, move this to some declaration file?

packages/webdriverio/src/utils/findStrategy.ts
christian-bromann: some more information about the selector that was used and is invalid would be good

packages/webdriverio/src/utils/implicitWait.ts
christian-bromann: the problem is that we generate the interface properties for the webdriverio element to attach its methods compiled from the protocol json files .. not exactly sure how this should look like atm.

packages/webdriverio/src/utils/findStrategy.ts
Fabianopb: I've pushed some changes including more info on those errors!

packages/webdriverio/src/utils/findStrategy.ts
Fabianopb: Taken care in another thread: https://github.com/webdriverio/webdriverio/pull/6023#discussion_r510838614

packages/webdriverio/src/utils/implicitWait.ts
Fabianopb: This is not fully clear to me yet, do you want to create the issue? Or I could create a draft of it to help! 🙂

Fabianopb:
@christian-bromann just a heads up, I've rebased due to some small conflicts.

packages/webdriverio/package.json
```diff
@@ -71,5 +71,10 @@
     "rgb2hex": "^0.2.0",
     "serialize-error": "^7.0.0",
     "webdriver": "6.7.0"
+  },
+  "devDependencies": {
+    "@types/atob": "^2.1.2",
+    "@types/fs-extra": "^9.0.2",
+    "@types/lodash.isplainobject": "^4.0.6"
```
christian-bromann, line 78: We need to move these into the dependencies otherwise they won't be installed when releasing the package

christian-bromann (commented):
one last thing afterwards it seems good to me

packages/webdriverio/package.json
Fabianopb: Alright, moved to the right place now!

christian-bromann (approved):
👍 thanks!

PR #6026, opened 2020-10-22
Minor cleanups after @wdio/reporter migration

christian-bromann:
## Proposed changes

Some things had to be cleaned up after migrating `@wdio/reporter`.

cc @ablok


packages/wdio-sync/webdriverio-core.d.ts
```diff
@@ -572,6 +572,7 @@ declare namespace WebdriverIO {
     type MockFilterOptions = {
         method?: string | ((method: string) => boolean),
         headers?: Record<string, string> | ((headers: Record<string, string>) => boolean),
+        requestHeaders?: Record<string, string> | ((headers: Record<string, string>) => boolean),
```
mgrybyk, line 575: I see changes in webdriverio d ts, but why don't you modify the webdriverio tpl d ts file?

mgrybyk: Is it because someone forgot to generate these files?

christian-bromann: Yes, seems like it.

mgrybyk (approved):

PR #6032, opened 2020-10-25
Tracking doc: rewrite `@wdio/testingbot-service` package into TypeScript

hieuxlu:
## Proposed changes
(#5850) Rewrite `@wdio/testingbot-service` package into TypeScript
## Types of changes

- [X] Documentation update

## Checklist

## Further comments

- Firstly, I'm not proud of this. I used `ts-migrate` to spice things up. There's a lot of `any` because I have no idea how to cross reference a type of a package that hasn't migrated to Typescript yet. `@wdio/testingbot-service` references `wdio-mocha-framework` types, and need to extend `BrowserObject` for multiremote typings. I feel like we're heading blindly into migrating Typesript, without organizing what should be done first.


packages/wdio-testingbot-service/src/launcher.ts
```diff
@@ -7,11 +7,14 @@ import logger from '@wdio/logger'
 const log = logger('@wdio/sauce-service')
 
 export default class TestingBotLauncher {
-    constructor (options) {
+    options: TestingbotOptions;
+    tbTunnelOpts!: TunnelLauncherOptions;
+    tunnel: TestingbotTunnel | undefined;
```
christian-bromann, line 12:
```suggestion
    tunnel?: TestingbotTunnel;
```

packages/wdio-testingbot-service/src/service.ts
```diff
@@ -5,6 +5,14 @@ const log = logger('@wdio/testingbot-service')
...
 
 export default class TestingBotService {
+    capabilities: any;
+    config: any;
+    failures: any;
+    isServiceEnabled: any;
+    suiteTitle: any;
+    tbSecret: any;
+    tbUser: any;
+    testCnt: any;
```
christian-bromann, line 15: Can we define these?

christian-bromann (changes requested):

packages/wdio-testingbot-service/src/service.ts
hieuxlu: Done

christian-bromann:
Still see a lot of `any` let me know if this is ready for review.

hieuxlu:
@christian-bromann This is the best that I can do for now. The remaining any is because of:
- `global.browser.config.mochaOpts` I can't reference wdio-mocha-framework in this package
- `global.browser[name]` Multiremote browser typing might cause compile error in other packages

Any idea how I could solve above issues?

christian-bromann:
@hieuxlu can you check in #6001 how @martinfrancois has accomplished this?

hieuxlu:
> @hieuxlu can you check in #6001 how @martinfrancois has accomplished this?

A bit magic, I add a comment on that PR.

package.json
```diff
@@ -65,7 +65,7 @@
     "@octokit/rest": "^18.0.0",
     "@tsconfig/node10": "^1.0.7",
     "@types/jest": "^26.0.0",
-    "@types/node": "^10.17.35",
+    "@types/node": "^12.19.3",
```
hieuxlu, line 68: This is needed because `@types/node@^10` does not support globalThis, which is needed to use `global.browser`

packages/wdio-applitools-service/src/index.ts
```diff
@@ -8,14 +8,6 @@ const DEFAULT_VIEWPORT = {
     height: 900
 }
 
-declare global {
```
hieuxlu, line 11: This is not needed after upgrading `@types/node` to `^12`, typings for `globalThis` will support extending global object

packages/wdio-shared-store-service/shared-store-service.d.ts
```diff
@@ -1,5 +1,5 @@
 declare namespace WebdriverIO {
-    interface BrowserObject {
+    interface Browser {
```
hieuxlu, line 2: As a rule of thumb, if we need to extend core properties, we should extend `Browser` interface. This is because `BrowserObject` can extends `ClientSync` or `ClientAsync`, so if we modify `BrowserObject`, there will be duplicate codes everywhere

packages/wdio-sync/webdriverio-core.d.ts
```diff
@@ -1333,6 +1333,18 @@ declare namespace WebdriverIO {
...
+        /**
+         * multiremote browser instance names
+         */
+        instances: string[];
+        /**
+         * flag to indicate multiremote browser session
+         */
+        isMultiremote: true;
+    }
+    type MultiRemoteBrowserObject = MultiRemoteBrowser & MultiRemoteBrowserReference
```
hieuxlu, line 1347: Union type is the only way I've found to define typings for multiremote browser access such as `global.browser[browserName]`. See https://stackoverflow.com/a/57932353/3439960 for similar issue and https://www.typescriptlang.org/docs/handbook/interfaces.html for why extending Dictionary types does not work

packages/wdio-sync/webdriverio.d.ts
```diff
@@ -27,9 +27,10 @@ declare namespace WebdriverIO {
     }
 
     interface BrowserObject extends WebDriver.ClientOptions, WebDriver.Client, WebdriverIO.Browser { }
+    interface MultiRemoteBrowser extends WebDriver.ClientOptions, WebDriver.Client, WebdriverIO.Browser { }
 }
 
-declare var browser: WebdriverIO.BrowserObject;
+declare var browser: WebdriverIO.BrowserObject | WebdriverIO.MultiRemoteBrowserObject;
```
hieuxlu, line 33: Typescript is smart enough to conditionally detect child type of union type base on property value difference.
In this case, the property is `isMultiremote`.
In `BrowserObject`, its type is `isMultiremote?: false`
In `MultiRemoteBrowserObject`, its type is `isMultiremote: true`.
So when you check conditions like below, correct type will be detected
```
if(!global.browser.isMultiremote) {
   const a = global.browser // type is detected as WebdriverIO.BrowserObject
} else {
   const b = global.browser // type is detected as WebdriverIO.MultibrowserObject
}
```

packages/webdriver/src/types.ts
```diff
@@ -260,12 +260,13 @@ export interface W3CCapabilities {
...
 }
 
-export interface DesiredCapabilities extends Capabilities, SauceLabsCapabilities, SeleniumRCCapabilities, AppiumIOSCapabilities, GeckodriverCapabilities, IECapabilities, AppiumAndroidCapabilities, AppiumCapabilities, VendorExtensions, GridCapabilities, ChromeCapabilities {
+export interface DesiredCapabilities extends Capabilities, SauceLabsCapabilities, TestingbotCapabilities, SeleniumRCCapabilities, AppiumIOSCapabilities, GeckodriverCapabilities, IECapabilities, AppiumAndroidCapabilities, AppiumCapabilities, VendorExtensions, GridCapabilities, ChromeCapabilities {
     // Read-only capabilities
     cssSelectorsEnabled?: boolean;
     handlesAlerts?: boolean;
     version?: string;
     platform?: string;
+    public?: any;
```
hieuxlu, line 269: `SauceLabsCapabilities` and `TestingbotCapabilities` has conflicting types of property `public`, so extending pattern will not work here. Casting public to any is the only workaround for now. I think we should use a different typing pattern in the future.

tests/typings/webdriverio/types/async.d.ts
```diff
@@ -4,12 +4,6 @@
 
 // module should be "webdriverio" if used within `ts` file instead of `d.ts`
 declare module WebdriverIO {
-    interface BrowserObject {
```
hieuxlu, line 7: We now have proper typings for `MultiremoteBrowserObject`, so this is not necessary any more

packages/wdio-testingbot-service/src/service.ts
```diff
@@ -151,13 +160,14 @@ export default class TestingBotService {
             return this.updateJob(global.browser.sessionId, failures)
         }
 
+        let browser = global.browser
```
hieuxlu, line 163: @martinfrancois This is needed so Typescript will detect type of multiremote browser object correctly
```
        if (!global.browser.isMultiremote) {
            log.info(`Update job with sessionId ${global.browser.sessionId}, ${status}`)
            return this.updateJob(global.browser.sessionId, failures)
        }

        let browser = global.browser
```

packages/wdio-testingbot-service/src/service.ts
```diff
@@ -13,30 +22,30 @@ export default class TestingBotService {
...
-        this.isServiceEnabled = this.tbUser && this.tbSecret
+        this.isServiceEnabled = Boolean(this.tbUser && this.tbSecret)
     }
 
     /**
      * Before suite
      * @param {Object} suite Suite
     */
-    beforeSuite (suite) {
+    beforeSuite (suite: any) {
```
christian-bromann, line 40: Do we want to keep them untyped?

hieuxlu: No, this was an oversight. Will fix it.

christian-bromann: there are a bunch of `any` still the file which is why I commented https://github.com/webdriverio/webdriverio/pull/6032#issuecomment-716670117

hieuxlu:
I know, I just happen to solve the most headache with multiremote and mocha
typings first, so I requested your reviews

packages/wdio-testingbot-service/src/service.ts
hieuxlu: @christian-bromann Fixed most of the any. The remaining are:
- Cucumber types, which does not export any typing yet so I can't use types for `after/beforeFeature`
- Unit test, which sometimes must cast to any because explicit cast does not work

packages/wdio-testingbot-service/tsconfig.prod.json
```diff
@@ -3,7 +3,8 @@
   "compilerOptions": {
     "baseUrl": ".",
     "outDir": "./build",
-    "rootDir": "./src"
+    "rootDir": "./src",
+    "types": ["webdriverio", "@wdio/logger", "jest", "@wdio/mocha-framework"]
```
christian-bromann, line 7: These should be the same as in `packages/wdio-testingbot-service/tsconfig.json`

packages/webdriver/src/types.ts
christian-bromann: Let's make Sauce Labs property a boolean too and get rid of this.

christian-bromann (changes requested):
some last comments after that we should be able to merge this.

packages/webdriver/src/types.ts
hieuxlu: But it is indeed string, possible values are public, public restricted, share and team
https://wiki.saucelabs.com/display/DOCS/Test+Configuration+Options

tests/typings/webdriverio/tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "outDir": "dist",
     "noImplicitAny": true,
-    "lib": ["es2018"],
+    "lib": ["es2018", "dom"],
```
hieuxlu, line 5: Puppeteer typings now will need dom lib to use `Document` type, possibly related to upgrading NodeJS 12 typings

christian-bromann (approved):
Awesome, this will also fix #6070

Issue #4547, opened 2019-09-29
Control interface output


Krinkle:
**Environment:**
 - **WebdriverIO version:** 5.13.2
 - **If WDIO Testrunner, running mode:** sync
 - **Node.js:** 10.15.2 (with npm 6.5.0)
 - **Browser name and version:** Chromium 71
 - **Platform name and version:** Debian 9 Stretch

```
    "@wdio/cli": "5.13.2",
    "@wdio/devtools-service": "5.13.2",
    "@wdio/dot-reporter": "5.13.2",
    "@wdio/junit-reporter": "5.13.2",
    "@wdio/local-runner": "5.13.2",
    "@wdio/mocha-framework": "5.13.2",
    "@wdio/sauce-service": "5.13.2",
    "@wdio/sync": "5.13.2",
    "chromedriver": "73.0.0",
    …
    "wdio-chromedriver-service": "5.0.2",
    "webdriverio": "5.13.2"
```

**Config of WebdriverIO**
```
	logLevel: 'error',
	framework: 'mocha',
	reporters: [
		'dot',
		[ 'junit', {
			outputDir: logPath
		} ]
	],
```

Full config at [gerrit.wikimedia.org/g/mediawiki/core/+/refs/changes/41/539441/9](https://gerrit.wikimedia.org/g/mediawiki/core/+/refs/changes/41/539441/9/tests/selenium/wdio.conf.js).

**Describe the bug**

There is some other unknown reporter active and it appears to be preventing the "dot" reporter from reporting anything until everything is done, at which point a  `"dot" Reporter` section appears with all the output printed at once.

```
$ npm run selenium
> wdio ./tests/selenium/wdio.conf.js

Execution of 6 spec files started at 2019-09-29T03:24:45.248Z

Starting ChromeDriver 73.0.3683.20 (8e2b610813e167eee3619ac4ce6e42e3ec622017) on port 4444
[0-1] RUNNING in chrome - /tests/selenium/specs/page.js
[0-1] PASSED in chrome - /tests/selenium/specs/page.js
[0-2] RUNNING in chrome - /tests/selenium/specs/rollback.js
[0-2] PASSED in chrome - /tests/selenium/specs/rollback.js
[0-3] RUNNING in chrome - /tests/selenium/specs/specialrecentchanges.js
[0-3] PASSED in chrome - /tests/selenium/specs/specialrecentchanges.js
[...]

PR #6034, opened 2020-10-26
@wdio/cli should log the entire stack trace if available

Photonios:
## Proposed changes
`@wdio/cli` logs the execution of every test suite and logs when a test fails. When it does fail with an error, it only logs the message. For example:

```
[0-0] TypeError in "Bla bla my test"
TypeError: Cannot read property 'test' of undefined
```

This does not tell the user where the error came from. This change logs the stack trace if available, thus changing the output to be more like this:

```
[0-0] TypeError in "Bla bla my test"
TypeError: Cannot read property 'test' of undefined
    at Function.selectValueFromFilter (/Users/swen/Code/somerepo/frontend/e2e/test/desktopPages/pages/filters.js:136:27)
    at Context.<anonymous> (/Users/swen/Code/somerepo/frontend/e2e/test/specs/project/filters/search_page/verifySomething.desktop.test.js:34:17)
    at Context.sync (/Users/swen/Code/project/frontend/node_modules/mocha-steps/lib/step.js:29:24)
```

If no stack trace is available, it'll still just log the error's message.

## Types of changes
- [ ] Documentation update

## Checklist

## Further comments
An alternative solution is for users to add a custom reporter that only logs stack traces:

```
class ErrorReporter extends WDIOReporter {
    onTestFail(test) {
        if (test.error.stack) {
            console.error(test.error.stack);
        }
    }
}
```

Unfortunately, this results in the message being printed twice because there is no way to turn off the behaviour in `@wdi/cli`:

```
[0-0] TypeError in "Bla bla my test"
TypeError: Cannot read property 'test' of undefined
[0-0] TypeError: Cannot read property 'test' of undefined
    at Function.selectValueFromFilter (/Users/swen/Code/somerepo/frontend/e2e/test/desktopPages/pages/filters.js:136:27)
    at Context.<anonymous> (/Users/swen/Code/somerepo/frontend/e2e/test/specs/project/filters/search_page/verifySomething.desktop.test.js:34:17)
    at Context.sync (/Users/swen/Code/project/frontend/node_modules/mocha-steps/lib/step.js:29:24)
```

### DIFF
--- a/packages/wdio-cli/src/interface.js
+++ b/packages/wdio-cli/src/interface.js
@@ -102,12 +102,14 @@ export default class WDIOCLInterface extends EventEmitter {
     }
 
     onTestError(payload) {
-        let error = { type: 'Error', message: typeof payload.error === 'string' ? payload.error : 'Unknown error.' }
+        let error = { type: 'Error', message: typeof payload.error === 'string' ? payload.error : 'Unknown error.', stack: null }
         if (payload.error) {
             error.type = payload.error.type || error.type
             error.message = payload.error.message || error.message
+            error.stack = payload.error.stack || error.stack
         }
-        return this.log(`[${payload.cid}]`, `${chalk.red(error.type)} in "${payload.fullTitle}"\n${chalk.red(error.message)}`)
+
+        return this.log(`[${payload.cid}]`, `${chalk.red(error.type)} in "${payload.fullTitle}"\n${chalk.red(error.stack || error.message)}`)
     }
 
     getFilenames(specs = []) {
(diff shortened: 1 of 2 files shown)


mgrybyk:
Not really sure if it's a proper place to print out stack.
The full stack or whatever should be handled by the reporter. I'd do it as a part of 
https://github.com/webdriverio/webdriverio/issues/4547

What do you think?

Photonios:
Yes, being able to control the output of `@wdio/cli`'s interface would also resolve this problem. That would be nicer. It would require significant changes yes.

As I demonstrated, logging the stack trace in a reporter would duplicate the message in the console because `@wdio/cli` already prints the message.

An alternative solution would be for `@wdio/cli` to not print the error message at all and leave the logging of error messages entirely to the reporter. Would that be satisfactory?

christian-bromann (approved):
Showing more data is definitely not a bad idea. I agree with @mgrybyk though that we ultimately want to have better control over the CLI interface and output.

PR #6049, opened 2020-10-28
Do not disable webdriver steps reporting when using cucumber step reporter

### LINKED ISSUE #6048: Allow webdriver steps reporting in allure when using cucumber step reporting.
When using the allure with cucumber. When I set `useCucumberStepReporter` I lose the nice webdriver steps reporting.

I would like the cucumber step reporting to be controlled only by the existing option `disableWebdriverStepsReporting`.

Not use `useCucumberStepReporter` and live with each scenario step being a test.

gambrose:
## Proposed changes

Resolves #6048

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

I have marked this as a breaking change as it will affect existing users that have `useCucumberStepReporter` turned on. If they want to preserve the existing behaviour then would need to also enable `disableWebdriverStepsReporting`

### DIFF
--- a/packages/wdio-allure-reporter/src/index.js
+++ b/packages/wdio-allure-reporter/src/index.js
@@ -193,7 +193,9 @@ class AllureReporter extends WDIOReporter {
             return
         }
 
-        if (this.options.disableWebdriverStepsReporting || this.options.useCucumberStepReporter || this.isMultiremote) {
+        const { disableWebdriverStepsReporting } = this.options
+
+        if (disableWebdriverStepsReporting || this.isMultiremote) {
             return
         }
 
@@ -209,7 +211,7 @@ class AllureReporter extends WDIOReporter {
     }
 
     onAfterCommand(command) {
-        const { disableWebdriverStepsReporting, disableWebdriverScreenshotsReporting, useCucumberStepReporter } = this.options
+        const { disableWebdriverStepsReporting, disableWebdriverScreenshotsReporting } = this.options
         if (this.isScreenshotCommand(command) && command.result.value) {
             if (!disableWebdriverScreenshotsReporting) {
                 this.lastScreenshot = command.result.value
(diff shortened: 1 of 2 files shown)


christian-bromann:
@gambrose can you please sign the CLA with the email you committed the PR with?

@BorisOsipov I would need your input here if this makes sense 🤔

packages/wdio-allure-reporter/src/index.js
```diff
@@ -193,7 +193,9 @@ class AllureReporter extends WDIOReporter {
             return
         }
 
-        if (this.options.disableWebdriverStepsReporting || this.options.useCucumberStepReporter || this.isMultiremote) {
```
christian-bromann, line 196: It seems that this patch disables whatever `useCucumberStepReporter` was suppose to be for, doesn't seem to me like a right solution.

gambrose: From my understanding, `useCucumberStepReporter` setting is used to change the report hierarchy when using cucumber. It goes this by using steps instead of tests (see line 103).

It appears it also disabled the command logging but as this is controlled using a  separate flag and I would like to use both features as per #6049 I think this is the correct approach.

BorisOsipov:
@gambrose @christian-bromann 
I am `useCucumberStepReporter` option author. I don't remember why I disable webdriver steps reporting. If it works and report doesn't look like a mess I don't mind

gambrose:
> If it works and report doesn't look like a mess I don't mind

The report obviously has more steps with web driver reporting enabled but these extra steps appear as sub-steps to the steps with the gerkin text. So it is still readable and you can then drill down to see what commands where run.

You can also easily get the original behaviour by also enabling `disableWebdriverStepsReporting`.

christian-bromann:
@gambrose can you provide a screenshot how the behavior was before and how it is now?

gambrose:
Original behaviour 
original

After change
after

And as I said you regain original behaviour if you also specify `disableWebdriverStepsReporting`

* example is taken from https://github.com/webdriverio/cucumber-boilerplate

christian-bromann (approved):
Sounds good! Thanks for looking into this 👍

PR #6061, opened 2020-11-02
allure-reporter: fix mime type

### LINKED ISSUE #6053: [wdio-allure-reporter] Unable to attach buffer content type as a 'image/png'  to allure report
https://github.com/webdriverio/webdriverio/blob/63519058a7fab1bdc424fc95c14f97207d496627/packages/wdio-allure-reporter/src/index.js#L538-L540

All kind of buffer content types (like screenshots created in wdio) are handed as a  'text/plain'  due to changes made in #6022

mgrybyk:
## Proposed changes

fix #6053

- allows user to override the default types
- assume a Buffer is an image unless the type is explicitly provided
- added tests to cover all the scenarios

works like this: `AllureReporter.addAttachment('scr', Buffer.from(browser.takeScreenshot(), 'base64'))`

## Types of changes

## Checklist

### DIFF
--- a/packages/wdio-allure-reporter/src/index.js
+++ b/packages/wdio-allure-reporter/src/index.js
@@ -444,7 +444,8 @@ class AllureReporter extends WDIOReporter {
     }
 
     dumpJSON(name, json) {
-        this.allure.addAttachment(name, JSON.stringify(json, null, 2), 'application/json')
+        const content = JSON.stringify(json, null, 2)
+        this.allure.addAttachment(name, typeof content === 'string' ? content : `${content}`, 'application/json')
     }
 
     attachScreenshot() {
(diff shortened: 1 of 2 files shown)


packages/wdio-allure-reporter/src/index.js
```diff
@@ -444,7 +444,8 @@ class AllureReporter extends WDIOReporter {
     }
 
     dumpJSON(name, json) {
-        this.allure.addAttachment(name, JSON.stringify(json, null, 2), 'application/json')
+        const content = JSON.stringify(json, null, 2)
+        this.allure.addAttachment(name, typeof content === 'string' ? content : `${content}`, 'application/json')
```
christian-bromann, line 448: Why do we need to check if `content` is a string if we use `JSON.stringify` above? That method always returns a string.

mgrybyk: It doesn't, it can return undefined that causes failures

christian-bromann: This would mean we would add `undefined` as content. How about we have a:

```js
if (!json) {
    return
}
```
instead?

mgrybyk: not sure, I'd always add whatever the user wants, even if it's undefined.
I'd like to see undefined in attachments if I passed it

christian-bromann: but undefined is not really saying anything, and it is also not a `'application/json'` mime type. Can we have something like "no content" or something like that?

mgrybyk: yeah, I'll set it to text plain

mgrybyk: but I'd not change undefined to anything, because it can be what user has passed, ex
```
let a
if (b) { a = b }
addAttachment('a', a) // I expect to see either undefined or b
```

christian-bromann (approved):
LGTM 👍

PR #6062, opened 2020-11-02
browser.mock use utf8 by default

mgrybyk:
## Proposed changes

Defect: Chars like `ö` in the mocked body are converted to `�`.

let's use utf8 encoding by default and do not convert file buffer to string and back.

## Types of changes

## Checklist


christian-bromann (approved):
👍

PR #5833, opened 2020-11-04
Tracking doc: rewrite `@wdio/cucumber-framework` package into TypeScript

christian-bromann:

- if you are interested converting source files from `/packages/wdio-cucumber-framework` into TypeScript please comment on this thread and mention the files you want to convert to avoid duplicate work
- if you make a PR, label it with `TypeScript` and the package name `wdio-cucumber-framework` so we can easier find all PRs

Happy rewriting 😅

### DIFF
--- a/packages/wdio-cucumber-framework/package.json
+++ b/packages/wdio-cucumber-framework/package.json
@@ -22,6 +22,9 @@
     "url": "https://github.com/webdriverio/webdriverio/issues"
   },
   "dependencies": {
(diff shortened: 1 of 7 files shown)


christian-bromann:
Cucumber seems to provide much better TS support with v7 which is not yet released. I will hold off for a moment as this will be a breaking change also for WebdriverIO users to update it. Will do this package last before releasing WebdriverIO v7.

christian-bromann:
Closing in favor of #6263

Issue #5828, opened 2020-09-14
Tracking doc: rewrite `@wdio/browserstack-service` package into TypeScript


christian-bromann:
This is a tracking issue for organising the effort of re-writing the code base from vanilla JS into TypeScript. If you are reading this and are interested getting involved, this is a great chance to do so. Here are the rules:

- if you are interested converting source files from `/packages/wdio-browserstack-service` into TypeScript please comment on this thread and mention the files you want to convert to avoid duplicate work
- there is no reason to convert the whole package at once - smaller PRs make it easier to review the code
- if you make a PR, reference this issue so we can keep track of the progress
- if you make a PR, label it with `TypeScript` and the package name `wdio-browserstack-service` so we can easier find all PRs
- if you make a PR, assign it to the `TypeScript Rewrite` project

Happy rewriting 😅

goveanick:
@christian-bromann I'll like to give this one a try.

christian-bromann:
@goveanick how are things going? Is there anything we can help with?

christian-bromann:
@goveanick given I haven't heard back I will un-assign you and open it up for others.

smarigowda:
I am interested in contributing, can I get help on how do I get started ?

christian-bromann:
@smarigowda I have assigned the ticket already to @Fabianopb (maybe there is a way to collaboratively work together on this). If you haven't done any work on the project I recommend to look into our [contribution docs](https://webdriver.io/docs/contribute.html). Please let me know if they are unclear.

Fabianopb:
@christian-bromann as a Member should I be able to push branches directly to this repo or should I continue with a fork? I have some initial work for this one already, from where @smarigowda can probably continue.

christian-bromann:
> or should I continue with a fork?

Please continue with your fork. Once you reached 10 qualifying PRs and become a project committer you can work on branches.

christian-bromann:
Closing in favor of #6172

PR #6100, opened 2020-11-12
Refactor/webdriverio timer ts

Fabianopb:
## Proposed changes

Refactor webdriverio utils Timer to TypeScript.

## Types of changes

- [x] Refactor to TypeScript

## Checklist

## Further comments

Continuing with https://github.com/webdriverio/webdriverio/issues/5854

### DIFF
--- a/packages/webdriverio/src/commands/browser/waitUntil.ts
+++ b/packages/webdriverio/src/commands/browser/waitUntil.ts
@@ -44,7 +44,7 @@ import Timer from '../../utils/Timer'
 
 export default function waitUntil(
     this: WebdriverIO.BrowserObject,
-    condition: () => boolean | Promise<Boolean>,
+    condition: () => boolean | Promise<boolean>,
     {
         timeout = this.options.waitforTimeout,
         interval = this.options.waitforInterval,
@@ -64,6 +64,10 @@ export default function waitUntil(
 
     if (typeof interval !== 'number') {
         interval = this.options.waitforInterval
+    }
+
+    if (interval === undefined || timeout === undefined) {
+        throw new Error('interval or timeout are not set')
     }
 
     const fn = condition.bind(this)
--- a/packages/webdriverio/src/commands/browser/waitUntil.ts
+++ b/packages/webdriverio/src/commands/browser/waitUntil.ts
@@ -59,15 +59,11 @@ export default function waitUntil(
      * ensure that timeout and interval are set properly
      */
     if (typeof timeout !== 'number') {
-        timeout = this.options.waitforTimeout
+        timeout = this.options.waitforTimeout as number
     }
 
     if (typeof interval !== 'number') {
-        interval = this.options.waitforInterval
-    }
-
-    if (interval === undefined || timeout === undefined) {
-        throw new Error('interval or timeout are not set')
+        interval = this.options.waitforInterval as number
     }
 
     const fn = condition.bind(this)
--- a/packages/webdriverio/src/utils/Timer.ts
+++ b/packages/webdriverio/src/utils/Timer.ts
@@ -2,17 +2,21 @@ import { hasWdioSyncSupport, runFnInFiberContext } from '@wdio/utils'
 
 const TIMEOUT_ERROR = 'timeout'
 
-/**
- * Promise-based Timer. Execute fn every tick.
- * When fn is resolved — timer will stop
- * @param {Number} delay - delay between ticks
- * @param {Number} timeout - after that time timer will stop
- * @param {Function} fn - function that returns promise. will execute every tick
- * @param {Boolean} leading - should be function invoked on start
- * @return {promise} Promise-based Timer.
- */
 class Timer {
-    constructor (delay, timeout, fn, leading) {
+    private _delay: number;
+    private _timeout: number;
+    private _fn: () => Promise<boolean> | boolean;
+    private _leading: boolean;
+    private _conditionExecutedCnt: number;
+    private _start?: number;
+    private _ticks?: number;
+    private _timeoutId?: NodeJS.Timeout | null;
+    private _mainTimeoutId?: NodeJS.Timeout | null;
+    lastError?: Error | null;
+    private _resolve?: (value?: PromiseLike<Timer>) => void;
+    private _reject?: (reason?: any) => void;
+
+    constructor (delay: number, timeout: number, fn: () => Promise<boolean> | boolean, leading: boolean) {
         this._delay = delay
         this._timeout = timeout
         this._fn = fn
@@ -29,14 +33,14 @@ class Timer {
             this._fn = () => runFnInFiberContext(fn)()
         }
 
-        const retPromise = new Promise((resolve, reject) => {
+        const retPromise = new Promise<Timer>((resolve, reject) => {
             this._resolve = resolve
             this._reject = reject
         })
 
         this.start()
 
-        return retPromise
+        return { ...this, retPromise }
     }
 
     start () {
@@ -59,7 +63,9 @@ class Timer {
 
             emitTimerEvent({ id: this._start, timeout: true })
             const reason = this.lastError || new Error(TIMEOUT_ERROR)
-            this._reject(reason)
+            if (this._reject) {
+                this._reject(reason)
+            }
             this.stop()
         }, this._timeout)
     }
@@ -73,13 +79,15 @@ class Timer {
 
     stopMain () {
         emitTimerEvent({ id: this._start })
-        clearTimeout(this._mainTimeoutId)
+        if (this._mainTimeoutId) {
+            clearTimeout(this._mainTimeoutId)
+        }
     }
 
     tick () {
         const result = this._fn()
 
-        if (typeof result.then !== 'function') {
+        if (typeof result === 'boolean') {
             if (!result) {
                 return this.checkCondition(new Error('return value was never truthy'))
             }
@@ -93,15 +101,19 @@ class Timer {
         )
     }
 
-    checkCondition (err, res) {
+    checkCondition (err: Error | null, res?: any) {
         ++this._conditionExecutedCnt
         this.lastError = err
 
         // resolve timer only on truthy values
-        if (res) {
+        if (res && this._resolve) {
             this._resolve(res)
             this.stop()
             this.stopMain()
+            return
+        }
+
+        if (!this._start || !this._ticks) {
             return
         }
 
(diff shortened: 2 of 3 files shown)


packages/webdriverio/src/utils/Timer.ts
```diff
@@ -29,14 +33,14 @@ class Timer {
...
-        const retPromise = new Promise((resolve, reject) => {
+        const retPromise = new Promise<Timer>((resolve, reject) => {
             this._resolve = resolve
             this._reject = reject
         })
 
         this.start()
 
-        return retPromise
+        return { ...this, retPromise }
```
Fabianopb, line 43: Think it's ok to spread the context here? Otherwise types will complain that `retPromise` is missing several properties from the class (which makes sense)

packages/webdriverio/src/utils/interception/index.ts
```diff
@@ -37,8 +37,12 @@ export default class Interception {
...
         }
 
-        const fn = () => this.calls && this.calls.length > 0
-        const timer = new Timer(interval, timeout, fn, true) as unknown as Promise<void>
+        if (interval === undefined || timeout === undefined) {
+            throw new Error('interval or timeout are not set')
+        }
+
+        const fn = () => this.calls && this.calls.length > 0 || false
+        const timer = new Timer(interval, timeout, fn, true) as unknown as Promise<Timer>
```
Fabianopb, line 45: Still need to figure out how to get rid of this `unknown`, any ideas anyone?

packages/webdriverio/src/commands/browser/waitUntil.ts
```diff
@@ -59,11 +59,11 @@ export default function waitUntil(
      * ensure that timeout and interval are set properly
      */
     if (typeof timeout !== 'number') {
-        timeout = this.options.waitforTimeout
+        timeout = this.options.waitforTimeout as number
     }
 
     if (typeof interval !== 'number') {
-        interval = this.options.waitforInterval
+        interval = this.options.waitforInterval as number
```
christian-bromann, line 66: Do we really need to cast them?

packages/webdriverio/src/utils/Timer.ts
```diff
@@ -2,17 +2,21 @@ import { hasWdioSyncSupport, runFnInFiberContext } from '@wdio/utils'
...
+    private _fn: () => Promise<boolean> | boolean;
+    private _leading: boolean;
+    private _conditionExecutedCnt: number;
+    private _start?: number;
+    private _ticks?: number;
+    private _timeoutId?: NodeJS.Timeout | null;
+    private _mainTimeoutId?: NodeJS.Timeout | null;
+    lastError?: Error | null;
+    private _resolve?: (value?: PromiseLike<Timer>) => void;
+    private _reject?: (reason?: any) => void;
```
christian-bromann, line 17:
```suggestion
    private _reject?: (reason?: Error) => void;
```

Fabianopb:
Reverted some changes to fix broken tests, but still gotta figure out those `any`s...

packages/webdriverio/src/utils/Timer.ts
```diff
@@ -2,17 +2,21 @@ import { hasWdioSyncSupport, runFnInFiberContext } from '@wdio/utils'
...
+    private _delay: number;
+    private _timeout: number;
+    private _fn: () => Promise<boolean> | boolean;
+    private _leading: boolean;
+    private _conditionExecutedCnt: number;
+    private _start?: number;
+    private _ticks?: number;
+    private _timeoutId?: NodeJS.Timeout | null;
+    private _mainTimeoutId?: NodeJS.Timeout | null;
+    lastError?: Error | null;
```
christian-bromann, line 15: Do we access the variable somewhere else? If not we can make it private too, no?
```suggestion
    private lastError?: Error;
```

packages/webdriverio/src/commands/browser/waitUntil.ts
Fabianopb: I couldn't think of a better approach here so far... we have a few other options:
1 - try to type guard against `undefined` before passing these args to Timer (this is what broke tests before).
2 - Timer would accept `undefined` for `interval` and `timeout` (doesn't feel right unless Timer would have defaults for them)
3 - WebdriverIO.Browser.options would declare `waitforTimeout` and `waitforInterval` as mandatory (also doesn't sound like a good option 🤔 )
```
/**
  * Default timeout for all `waitFor*` commands. (Note the lowercase f in the option name.)
  * This timeout only affects commands starting with `waitFor*` and their default wait time.
  */
waitforTimeout?: number;
```

packages/webdriverio/src/utils/Timer.ts
Fabianopb: Apparently we don't access anywhere so I have the same question. I didn't make it `private` at first as _it seemed to me_ that the private variables were declared starting with an __underscore_

But making it private also doesn't seem to cause any problem!

packages/webdriverio/src/utils/Timer.ts
Fabianopb: This was wrong, nevermind

packages/webdriverio/src/utils/Timer.ts
Fabianopb: Oops, ok it seems it can't be private, broke some other tests.

christian-bromann: @Fabianopb to access the property in the unit tests just do `timer['_lastError']`

packages/webdriverio/src/commands/browser/waitUntil.ts
christian-bromann: At some point the testrunner will apply default values to it (I think at the point where we validate the options). I wasn't aware that these can be also undefined. Therefor I would argue that we could leave this as is.

packages/webdriverio/src/utils/Timer.ts
```diff
@@ -2,17 +2,21 @@ import { hasWdioSyncSupport, runFnInFiberContext } from '@wdio/utils'
...
+    private _delay: number;
+    private _timeout: number;
+    private _fn: () => Promise<boolean> | boolean;
+    private _leading: boolean;
+    private _conditionExecutedCnt: number;
+    private _start?: number;
+    private _ticks?: number;
+    private _timeoutId?: NodeJS.Timeout | null;
+    private _mainTimeoutId?: NodeJS.Timeout | null;
+    private lastError?: Error;
```
christian-bromann, line 15: Let's stay compliant with the theme of having a `_` before all private properties
```suggestion
    private _lastError?: Error;
```

packages/webdriverio/src/utils/Timer.ts
```diff
@@ -29,14 +33,14 @@ class Timer {
...
-        const retPromise = new Promise((resolve, reject) => {
+        const retPromise = new Promise<Timer>((resolve, reject) => {
             this._resolve = resolve
             this._reject = reject
         })
 
         this.start()
 
-        return retPromise
+        return retPromise as any
```
christian-bromann, line 43: shouldn't this be `Promise<boolean>`?

packages/webdriverio/src/utils/Timer.ts
```diff
@@ -88,23 +96,27 @@ class Timer {
         }
 
         result.then(
-            (res) => this.checkCondition(null, res),
-            (err) => this.checkCondition(err)
+            (res: any) => this.checkCondition(null, res),
+            (err: any) => this.checkCondition(err)
```
christian-bromann, line 100:
```suggestion
            (err: Error) => this.checkCondition(err)
```

packages/webdriverio/src/utils/Timer.ts
```diff
@@ -88,23 +96,27 @@ class Timer {
         }
 
         result.then(
-            (res) => this.checkCondition(null, res),
-            (err) => this.checkCondition(err)
+            (res: any) => this.checkCondition(null, res),
```
christian-bromann, line 99:
```suggestion
            (res: boolean) => this.checkCondition(null, res),
```

packages/webdriverio/src/utils/Timer.ts
```diff
@@ -2,17 +2,21 @@ import { hasWdioSyncSupport, runFnInFiberContext } from '@wdio/utils'
...
+    private _fn: () => Promise<boolean> | boolean;
+    private _leading: boolean;
+    private _conditionExecutedCnt: number;
+    private _start?: number;
+    private _ticks?: number;
+    private _timeoutId?: NodeJS.Timeout | null;
+    private _mainTimeoutId?: NodeJS.Timeout | null;
+    private lastError?: Error;
+    private _resolve?: (value?: PromiseLike<Timer>) => void;
+    private _reject?: (reason?: Error) => void;
```
christian-bromann, line 17: Given that we definitely assign these two values in the constructor I would suggest to expect them to exist:
```suggestion
    private _resolve: (value?: PromiseLike<Timer>) => void;
    private _reject: (reason?: Error) => void;
```

If the TS compiler doesn't like it because we assign them within the Promise callback I suggest to have this at the top of the constructor:

```js
const NOOP = () => { throw new Error('not assigned') }

// ...
this._resolve = NOOP
this._reject = NOOP
```

christian-bromann (commented):
Some comments

packages/webdriverio/src/utils/Timer.ts
Fabianopb: Here actually if we get the type and the condition right in:
```ts
const result = this._fn() as any

if (typeof result.then !== 'function') { //...
```
Then these two `any`'s will be correctly inferred

packages/webdriverio/src/utils/Timer.ts
Fabianopb: I still didn't figure out how to handle this return type, it complains that is missing a few properties, which are all the private properties that are not optional and the class methods. This is correct if you think that a class constructor usually returns the class itself, however a class can also return an object, which seems to be the case, but I couldn't yet find a way to handle this with TS.

packages/webdriverio/src/utils/Timer.ts
Fabianopb: Fixed that along with other related changes

packages/webdriverio/src/commands/browser/waitUntil.ts
Fabianopb: > I would argue that we could leave this as is.

Sorry, you mean with the cast?

Fabianopb:
Coverage failed, which makes sense as type guards add branches for calculating coverage.

packages/webdriverio/src/utils/Timer.ts
Fabianopb: Not being able to specify return type of constructors seems to be an old discussion: https://github.com/microsoft/TypeScript/issues/11588

Fabianopb: One not so trivial solution, as it probably requires some good refactor in all the use cases, is to use a static factory https://2ality.com/2019/11/creating-class-instances.html#solution%3A-static-factory-method

christian-bromann:
@Fabianopb thanks for this work. I incorporated it into #5854. Closing this.

Fabianopb:
> @Fabianopb thanks for this work. I incorporated it into #5854. Closing this.

Alright, no worries, thanks for taking it forward and sorry for getting stuck with it!

christian-bromann:
@Fabianopb no worries at all! Would you be interested taking a look at https://github.com/webdriverio/webdriverio/issues/5828?

Fabianopb:
> @Fabianopb no worries at all! Would you be interested taking a look at #5828?

Sure, I'll take a look!

PR #6105, opened 2020-11-13
Try to fix @wdio/testingbot-service not using the Tunnel

testingbot:
## Proposed changes

The latest versions of @wdio/testingbot-service are no longer forwarding requests through http://localhost:4445
As a result, webdriverio tests will no longer use the TestingBot Tunnel.

This change includes a `tunnelIdentifier` instead, which should be passed through as a capability.
I tried to use `tb:options` for this, but for some reason this gets stripped out further down by webdriverio.

This PR does not fix the problem yet. If someone could guide me in how to pass this through the capabilities, then I can adjust this PR.

Many thanks

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## Checklist

[//]: # (_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._)

- [x] I have read the [CONTRIBUTING](https://github.com/webdriverio/webdriverio/blob/master/CONTRIBUTING.md) doc
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)
- [ ] I have added proper type definitions for new commands (if appropriate)

PR #6109, opened 2020-11-14
Fix @wdio/testingbot-service not using the Tunnel because of wrong import type

hieuxlu:
## Proposed changes

## Types of changes
- Fix #6105

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/wdio-testingbot-service/src/launcher.ts
+++ b/packages/wdio-testingbot-service/src/launcher.ts
@@ -1,7 +1,7 @@
 import { performance, PerformanceObserver } from 'perf_hooks'
 import { promisify } from 'util'
 
-import testingbotTunnel from 'testingbot-tunnel-launcher'
+import testingbotTunnel = require('testingbot-tunnel-launcher')
 import logger from '@wdio/logger'
 
 const log = logger('@wdio/testingbot-service')
@@ -14,24 +14,20 @@ export default class TestingBotLauncher {
         this.options = options
     }
 
-    async onPrepare (config: WebdriverIO.Config, capabilities: WebDriver.Capabilities) {
+    async onPrepare (config: WebdriverIO.Config) {
         if (!this.options.tbTunnel || !config.user || !config.key) {
             return
         }
 
-        const tbTunnelIdentifier = `TB-tunnel-${Math.random().toString().slice(2)}`
-
         this.tbTunnelOpts = Object.assign({
             apiKey: config.user,
-            apiSecret: config.key,
-            'tunnel-identifier': tbTunnelIdentifier,
+            apiSecret: config.key
         }, this.options.tbTunnelOpts)
 
-        if (!capabilities['tb:options']) {
-            capabilities['tb:options'] = {}
-        }
-
-        capabilities['tb:options'].tunnelIdentifier = tbTunnelIdentifier
+        config.protocol = 'http'
+        config.hostname = 'localhost'
+        config.port = 4445
+        config.path = '/wd/hub'
 
         /**
          * measure TestingBot tunnel boot time
--- a/packages/wdio-testingbot-service/src/launcher.ts
+++ b/packages/wdio-testingbot-service/src/launcher.ts
@@ -4,7 +4,7 @@ import { promisify } from 'util'
 import testingbotTunnel from 'testingbot-tunnel-launcher'
 import logger from '@wdio/logger'
 
-const log = logger('@wdio/sauce-service')
+const log = logger('@wdio/testingbot-service')
 
 export default class TestingBotLauncher {
     options: TestingbotOptions;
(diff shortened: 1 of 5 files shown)


packages/wdio-testingbot-service/src/launcher.ts
```diff
@@ -1,10 +1,10 @@
 import { performance, PerformanceObserver } from 'perf_hooks'
 import { promisify } from 'util'
 
-import testingbotTunnel from 'testingbot-tunnel-launcher'
+import testingbotTunnel = require('testingbot-tunnel-launcher')
```
christian-bromann, line 4: Why should we switch to commonjs import just for this package?

christian-bromann (changes requested):
Let's have ESM imports for TypeScript files.

packages/wdio-testingbot-service/src/launcher.ts
hieuxlu: As specified in the docs, it's the only way to work with `export =`
https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require

hieuxlu:
> Let's have ESM imports for TypeScript files.

It's the only way. Other types of import will be transpiled as `import default`, which caused the original bug because the library used `export =`
https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require

packages/wdio-testingbot-service/src/launcher.ts
christian-bromann: Why do we make the change in `testingbot-tunnel-launcher.d.ts` then as well as in the mock? The [original package](https://github.com/testingbot/testingbot-tunnel-launcher/blob/master/index.js) has a normal commonjs export:

```js
module.exports = require('./build/tunnel-launcher')
```

which can be imported using

```js
import TunnelLauncher from 'testingbot-tunnel-launcher'
```

So we should try to mimic this in our TS files and mocks.

christian-bromann (changes requested):
I still don't see any reason why we would need to change the way we import/export the tunnel launcher package.

hieuxlu:
> I still don't see any reason why we would need to change the way we import/export the tunnel launcher package.

I don't really understand. The migration of this package to Typescript caused a breaking change because Typescript by - default transpile  `import from` as `import default`. This does not work with packages using `export =` so we need to use correct `import require` for this package.

christian-bromann:
> This does not work with packages using `export =`

The `testingbot-tunnel-launcher` does not use this kind of export. If you look into the build code of that package you see at the end of the file:

```js
module.exports = downloadAndRun;
module.exports.kill = killTunnel;
```

This means we can just import it like any other package via

```js
import TunnelLauncher from 'testingbot-tunnel-launcher'
```

Who is using `export =`?

hieuxlu:
Isn't that the same? The exported result is a function and not es6 module.

christian-bromann:
> Isn't that the same?

No, `export = ...` is different from `module.exports`. A lot of packages this project depends on exports their modules this way and we also don't need to treat them differently. Also if you note, none of the proposed changes in #6105 was related to import/export issues of the tunnel package.

hieuxlu:
> > Isn't that the same?
> 
> No, `export = ...` is different from `module.exports`. A lot of packages this project depends on exports their modules this way and we also don't need to treat them differently. Also if you note, none of the proposed changes in #6105 was related to import/export issues of the tunnel package.

I will have a look. The changes proposed there only happen because they can not use the tunnel library at all, so they had to generate tunnel id randomly, and remove the tunnel code entirely

hieuxlu:
@christian-bromann You're probably right, I can't reproduce the original issue any more. I don't know what happened, but I'll reopen the original PR

hieuxlu:
I still don't understand why I can't reproduce the issue. It's very clear to me from the docs that `import require` is the only way that'll work with `module.exports =` or `export =`

https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require
```
The export = syntax specifies a single object that is exported from the module. This can be a class, interface, namespace, function, or enum.

When exporting a module using export =, TypeScript-specific import module = require("module") must be used to import the module.
```

christian-bromann:
@hieuxlu can you please tell me where you see `export =` being used??? I don't see this at all, neither in the compiled code of the `testingbot-tunnel-launcher` nor in any of our sub packages.

hieuxlu:
> @hieuxlu can you please tell me where you see `export =` being used??? I don't see this at all, neither in the compiled code of the `testingbot-tunnel-launcher` nor in any of our sub packages.

It's a Typescript syntax that will be compiled into `module.exports =`

hieuxlu:
Hey, sorry the the heated discussion, but I have no idea why I can't reproduce this any more. In addition, the original issue I think was a misconfiguration with wdio v6 because of root-level config breaking changes, so there were no bug, just a wrong type definition. So once testingbot confirmed, I will correct the import require. Thanks.

christian-bromann:
Given we merged #6105 I think this got obsolete now.

packages/wdio-testingbot-service/tests/__mocks__/testingbot-tunnel-launcher.ts
```diff
@@ -1,4 +1,4 @@
-export default (opts: any, cb: any) => {
+export = (opts: any, cb: any) => {
     cb(null, {
```
hieuxlu, line 2: @christian-bromann This was what made me confused. The mock for Jest uses `export default`, so I was confused.

hieuxlu: So if it's OK, I'd propose to fix the unit test mocking for consistency in this PR. I'll revert unrelated codes.

christian-bromann: Why does the mock need to be changed? `export default` is correct here as it mimics the way how testingbot exports its service: `module.exports = require('./build/tunnel-launcher')`

hieuxlu:
Typescript is not treating them the same. But I guess it's fine to leave it
as is. Two wrong is one right, the definition file use export default too
so nothing breaks. I'll leave it here.

PR #6127, opened 2020-11-18
Switch window query strings

### LINKED ISSUE #6012: Better deal with strings in `switchWindow`
Currently the [`switchWindow`](https://github.com/webdriverio/webdriverio/blob/master/packages/webdriverio/src/commands/browser/switchWindow.js) command iterates over all window handles and calls matches on url and title to see if the input matches it. This can be problematic if users use strings that contain a regexp character:

```js
// works perfectly if you match like this
browser.switchWindow('google.com')

// doesn't work if you have a `?` character
browser.switchWindow('google.com?foobar=barfoo') // does not match the url e.g. https://google.com?foobar=barfoo
```

Let's check if the command input is a string and use [`includes`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes) if so otherwise use [`match`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match) if it is a RegExp.

n/a

**Additional context**
This comes from a conversation I had with someone in Gitter stumbling over this.

christon88:
## Proposed changes

Fix #6012 
SwitchWindow was set to use `String.match(url)` for both Regex and string arguments, which will fail when called with a string containing regex characters.
Created a type based matching function that uses `String.includes()` for string type arguments.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/webdriverio/src/commands/browser/switchWindow.ts
+++ b/packages/webdriverio/src/commands/browser/switchWindow.ts
@@ -37,6 +37,14 @@ export default async function switchWindow (this: WebdriverIO.BrowserObject, url
 
     const tabs = await this.getWindowHandles()
 
+    const matchesTarget = (target: string): boolean => {
+        if (typeof urlOrTitleToMatch ==='string') {
+            return target.includes(urlOrTitleToMatch)
+        }
+        return !!target.match(urlOrTitleToMatch)
+
+    }
+
     for (const tab of tabs) {
         await this.switchToWindow(tab)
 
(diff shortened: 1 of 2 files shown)


packages/webdriverio/src/commands/browser/switchWindow.ts
```diff
@@ -37,22 +37,30 @@ export default async function switchWindow (this: WebdriverIO.BrowserObject, url
 
     const tabs = await this.getWindowHandles()
 
+    const matchesTarget = (target: string): boolean => {
+        if (typeof urlOrTitleToMatch ==='string') {
+            return target.includes(urlOrTitleToMatch)
+        }
+        return !!target.match(urlOrTitleToMatch)
+
```
christian-bromann, line 45:
```suggestion
```

christian-bromann (approved):
👍

Issue #5831, opened 2020-09-14
Tracking doc: rewrite `@wdio/config` package into TypeScript


christian-bromann:
This is a tracking issue for organising the effort of re-writing the code base from vanilla JS into TypeScript. If you are reading this and are interested getting involved, this is a great chance to do so. Here are the rules:

- if you are interested converting source files from `/packages/wdio-config` into TypeScript please comment on this thread and mention the files you want to convert to avoid duplicate work
- there is no reason to convert the whole package at once - smaller PRs make it easier to review the code
- if you make a PR, reference this issue so we can keep track of the progress
- if you make a PR, label it with `TypeScript` and the package name `wdio-config` so we can easier find all PRs
- if you make a PR, assign it to the `TypeScript Rewrite` project

Happy rewriting 😅

Y4suyuki:
@christian-bromann Hi, I would like to work on this issue!
This is my first time contributing to open source projects but I checked out related issues and PRs and I think I'm good to go.

christian-bromann:
@Y4suyuki awesome, let me know if you need any help. Feel free to rather make PRs for individual files rather than the whole package. Makes reviewing of PRs easier.

christian-bromann:
@Y4suyuki do you have any updates or need any help?`

Y4suyuki:
@christian-bromann 
Hi, sorry for not updating for long time. I'm working on wdio-config/src/utils.js (https://github.com/webdriverio/webdriverio/blob/master/packages/wdio-config/src/utils.js).
And yes I need some help.
I started working on utils.js, spent some time to catch up code base and I have problems typing `validateConfig` and `detectBackend` in utils.js.
- For detectBackend, I need [ConfigParser._config](https://github.com/webdriverio/webdriverio/blob/4d01a5b7efae9f0677c90722ce175a5d148a1575/packages/wdio-config/src/lib/ConfigParser.js#L17) to be typed. First I have bit stucked wondering how this config is typed and used. I searched in the source code for the spec and usage about config and found [ConfigurationFile.md](https://github.com/webdriverio/webdriverio/blob/master/docs/ConfigurationFile.md) (just few minutes ago :sweat_smile: ). With ConfigurationFile.md I think I can type config and move on. Do you think ConfigurationFile.md have sufficient information to type config? And also Do you think that there is anything I have to check out for typing config?
[...]

PR #6135, opened 2020-11-19
Rewrite ConfigParser.ts of wdio-config in typescript

abdatta:
## Proposed changes
This PR rewrites [ConfigParser.js](https://github.com/webdriverio/webdriverio/blob/b54edafa6551158a64eda24837a8f42f8196d175/packages/wdio-config/src/lib/ConfigParser.js) files of `wdio-config` package to Typescript.
The complete ts migration of wdio-config is being tracked by #5831 .

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/wdio-config/src/lib/ConfigParser.ts
+++ b/packages/wdio-config/src/lib/ConfigParser.ts
@@ -109,10 +109,10 @@ export default class ConfigParser {
          * run single spec file only, regardless of multiple-spec specification
          */
         if (spec.length > 0) {
-            this._config.specs = [...this.setFilePathToFilterOptions(spec, this._config.specs!)]
+            this._config.specs = this.setFilePathToFilterOptions(spec, this._config.specs!)
         }
         if (exclude.length > 0) {
-            this._config.exclude = [...this.setFilePathToFilterOptions(exclude, this._config.exclude!)]
+            this._config.exclude = this.setFilePathToFilterOptions(exclude, this._config.exclude!)
         }
 
         /**
@@ -188,7 +188,7 @@ export default class ConfigParser {
             for (let suiteName of suites) {
                 let suite = this._config.suites?.[suiteName]
                 if (!suite) {
-                    console.warn(`No suite was found with name "${suiteName}"`)
+                    log.warn(`No suite was found with name "${suiteName}"`)
                 }
                 if (Array.isArray(suite)) {
                     suiteSpecs = suiteSpecs.concat(ConfigParser.getFilePaths(suite))
--- a/packages/wdio-config/src/lib/ConfigParser.ts
+++ b/packages/wdio-config/src/lib/ConfigParser.ts
@@ -2,28 +2,24 @@ import fs from 'fs'
 import path from 'path'
 import glob from 'glob'
 import merge from 'deepmerge'
-
 import logger from '@wdio/logger'
 
 import { detectBackend, removeLineNumbers, isCucumberFeatureWithLineNumber, validObjectOrArray } from '../utils'
-
 import { DEFAULT_CONFIGS, SUPPORTED_HOOKS, SUPPORTED_FILE_EXTENSIONS } from '../constants'
+import type { Capabilities, ConfigOptions, Hooks } from '../types'
 
 const log = logger('@wdio/config:ConfigParser')
 const MERGE_OPTIONS = { clone: false }
 
 export default class ConfigParser {
-    constructor () {
-        this._config = DEFAULT_CONFIGS()
-
-        this._capabilities = []
-    }
+    private _config: ConfigOptions = DEFAULT_CONFIGS()
+    private _capabilities: Capabilities = [];
 
     /**
      * merges config file with default values
      * @param {String} filename path of file relative to current directory
      */
-    addConfigFile (filename) {
+    addConfigFile (filename: string) {
         if (typeof filename !== 'string') {
             throw new Error('addConfigFile requires filepath')
         }
@@ -34,13 +30,13 @@ export default class ConfigParser {
             /**
              * clone the original config
              */
-            const fileConfig = merge(require(filePath).config, {}, MERGE_OPTIONS)
+            const fileConfig = merge<ConfigOptions>(require(filePath).config, {}, MERGE_OPTIONS)
 
             /**
              * merge capabilities
              */
-            const defaultTo = Array.isArray(this._capabilities) ? [] : {}
-            this._capabilities = merge(this._capabilities, fileConfig.capabilities || defaultTo, MERGE_OPTIONS)
+            const defaultTo: Capabilities = Array.isArray(this._capabilities) ? [] : {}
+            this._capabilities = merge<Capabilities>(this._capabilities, fileConfig.capabilities || defaultTo, MERGE_OPTIONS)
             delete fileConfig.capabilities
 
             /**
@@ -81,7 +77,7 @@ export default class ConfigParser {
      * merge external object with config object
      * @param  {Object} object  desired object to merge into the config object
      */
-    merge (object = {}) {
+    merge (object = {} as ConfigOptions) {
         const spec = Array.isArray(object.spec) ? object.spec : []
         const exclude = Array.isArray(object.exclude) ? object.exclude : []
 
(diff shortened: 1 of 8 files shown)


packages/wdio-config/src/lib/ConfigParser.ts
```diff
@@ -144,18 +140,31 @@ export default class ConfigParser {
...
-            } else if (Array.isArray(service[hookName])) {
-                for (let hook of service[hookName]) {
-                    if (typeof hook === 'function') {
-                        this._config[hookName].push(hook.bind(service))
+            if (typeof hooksToBeAdded === 'function') {
+                addHook(hookName, hooksToBeAdded)
+            } else if (Array.isArray(hooksToBeAdded)) {
+                for (const hookToAdd of hooksToBeAdded) {
+                    if (typeof hookToAdd === 'function') {
+                        addHook(hookName, hookToAdd)
```
christian-bromann, line 167: How about we simplify this to:

```suggestion
                for (const hookToAdd of hooksToBeAdded.filter((hookToAdd) => typeof hookToAdd === 'function')) {
                    addHook(hookName, hookToAdd)
```

packages/wdio-config/src/lib/ConfigParser.ts
```diff
@@ -165,22 +174,23 @@ export default class ConfigParser {
...
-            let suiteSpecs = []
+            let suiteSpecs: string[] = []
             for (let suiteName of suites) {
-                // TODO: log warning if suite was not found
-                let suite = this._config.suites[suiteName]
-
-                if (suite && Array.isArray(suite)) {
+                let suite = this._config.suites?.[suiteName]
+                if (!suite) {
+                    console.warn(`No suite was found with name "${suiteName}"`)
```
christian-bromann, line 191: Let's use WebdriverIOs logger utility for this
```suggestion
                    log.warn(`No suite was found with name "${suiteName}"`)
```

packages/wdio-config/src/lib/ConfigParser.ts
```diff
@@ -247,7 +257,7 @@ export default class ConfigParser {
         if (filesToFilter.size === 0) {
             throw new Error(`spec file(s) ${cliArgFileList.join(', ')} not found`)
         }
-        return filesToFilter
+        return [...filesToFilter]
```
christian-bromann, line 260: Any reason why we transform it back to an array here? I have no strong opinion whether to use a Set or array but if we transform here we shouldn't do it again at places where `setFilePathToFilterOptions` is used.

packages/wdio-config/src/types.ts
```diff
@@ -6,16 +6,19 @@ export type Hooks = {
 export type Capabilities = (WebDriver.DesiredCapabilities | WebDriver.W3CCapabilities)[] | WebdriverIO.MultiRemoteCapabilities;
 
 export interface ConfigOptions extends Omit<WebdriverIO.Config, 'capabilities' | keyof WebdriverIO.Hooks>, Hooks {
+    spec?: string[];
+    suite?: string[];
     capabilities?: Capabilities;
     specFileRetryAttempts?: number;
+    cucumberFeaturesWithLineNumbers?: string[];
```
christian-bromann, line 13: It seems we have missed to add this to the docs. Could we add this to the [testrunner options](https://github.com/webdriverio/webdriverio/blob/master/docs/Options.md#testrunner-options)?

christian-bromann (changes requested):
A couple of comments. Great job!

packages/wdio-config/src/lib/ConfigParser.ts
abdatta: The reason I did not simplify this to that, is that Typescript fails to narrow down the type of `hookToAdd` when used with a `filter()` function. You can check that in this [playground link](https://www.typescriptlang.org/play?#code/MYewdgzgLgBAFiEBrCAVEAhApgQQCZ5Z4BcMAFAGICuYwUAluDAD4zW0PgDaAugJS8YAXhi8A3ACgJAMxp1GYGAEMCACURIyCZADklAWyyloAJ3pgA5gBp4G9PhJs5nMHxgBvCTBgB6HzAA6IN8fCQBfKVBIWG0kPUNhGAByWIBaMAMsJMkZEBNyKOhbZHsCGBBpYpR0bAciN09vekqyKABPAAcsCqrSvGEhESTZDgUkhq9vZTUNLQ14rBtYvr5JbwiI3PyyQpi7EAdyyti0TFwCIgDpegAbKCwTMjmSg4I3IQA+GHaunuXX-qDIYjeTgcYTbwqPDqZDPOKZJb7ByrcISIA).

packages/wdio-config/src/lib/ConfigParser.ts
abdatta: Ok I'll update this.

packages/wdio-config/src/lib/ConfigParser.ts
abdatta: I do not have a strong opinion on this either. The reasons I chose to convert it into strings are:
1. The JSDoc comment above states the return type to be `String[]`.
2. It semantically feels a bit weird to me that a function that takes an array of filenames to filter, would return a `Set` type. The uniqueness property is an implementation-specific thing, and the function return type shouldn't be implementation-specific, I think. That's why I changed it to `string[]`. I'm okay to change it back if you insist.

Btw, you're right about not to transform again where `setFilePathToFilterOptions` is used. I will make that change.

packages/wdio-config/src/types.ts
abdatta: OK, sure I'll add that to the doc. Let me know what should be the one-line description for the same.

packages/wdio-config/src/lib/ConfigParser.ts
christian-bromann: We can assume here that `hooksToBeAdded` is from type `Function[]` which makes this work: https://www.typescriptlang.org/play?#code/MYewdgzgLgBAFiEBrCAVEAhApgQQCZ5Z4BcMAFAGICuYwUAluDAD4zW0PgDaAugJS8YAXhi8A3ACgJAMxp1GYGAEMCACURIyCZADklAWyyloAJ3pgA5gBp4G9PhJs5nMHxgBvCTBgB6HzAA6IN8fCQBfKVBIWG0kPUNhGAByWIBaMAMsJMkZEBNyKOhbZHsCGBBpYpR0bAciN09vekqyKABPAAcsCqrSvGEhESTZDgUkhq9vZTUNLQ14rBtYvr5JbwiI3PyyQpi7EAdylti0TFwCImUIJ1HufgDpegAbKCwTMjmSg4I3IQA+GDtLo9ZbffqDIYjeTgcYTbwqPDqZCfOKZJb7ByrcISIA

packages/wdio-config/src/lib/ConfigParser.ts
christian-bromann: If we return with an array let's define `const filesToFilter: string[] = []` then

abdatta: Oh I thought we want to avoid duplicates, that's why we are using Set internally in that function

packages/wdio-config/src/lib/ConfigParser.ts
abdatta: Okay, I'll make that change then.

packages/wdio-config/src/lib/ConfigParser.ts
abdatta: I'm getting this TS error:
```
This expression is not callable.
  Each member of the union type '{ <S extends (config: Config, capabilities: DesiredCapabilities[]) => void>(predicate: (value: (config: Config, capabilities: DesiredCapabilities[]) => void, index: number, array: ((config: Config, capabilities: DesiredCapabilities[]) => void)[]) => value is S, thisArg?: any): S[]; (predicate: (value: (config: Confi...' has signatures, but none of those signatures are compatible with each other.ts(2349)
```

packages/wdio-config/src/lib/ConfigParser.ts
abdatta: I'm getting the TS error when I'm trying to use `filter`:
```
This expression is not callable.
  Each member of the union type '{ <S extends (config: Config, capabilities: DesiredCapabilities[]) => void>(predicate: (value: (config: Config, capabilities: DesiredCapabilities[]) => void, index: number, array: ((config: Config, capabilities: DesiredCapabilities[]) => void)[]) => value is S, thisArg?: any): S[]; (predicate: (value: (config: Confi...' has signatures, but none of those signatures are compatible with each other.ts(2349)
```

packages/wdio-config/src/lib/ConfigParser.ts
christian-bromann: ah makes sense, nevermind then

abdatta:
@christian-bromann  I have pushed the requested changes (except the `filter` one, which is failing in Typescript). Please review once, then I think this can be merged.

christian-bromann (approved):
LGTM 👍 thank you!

PR #6139, opened 2020-11-20
Rewrite `webdriverio` element commands

christian-bromann:
Part of #5854 which includes all element commands.


mgrybyk (changes requested):

christian-bromann:
@mgrybyk thanks for the review! I addressed the comments. Anything left to do?

mgrybyk (approved):

PR #6143, opened 2020-11-23
Don't use global.browser in services

### LINKED ISSUE #6025: do not rely on global browser instance
Some packages like spectron run webdriverio programmatically and suppose that a browser instance created in a programmatic way is the only entry point (which I believe is correct)

I'd like to remove all the occurrences of `global.browser` from the project

christian-bromann:
## Proposed changes

fixes #6025

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

n/a


mgrybyk:
overall looks good to me. I guess you want to proceed with the remaining services?

christian-bromann:
> I guess you want to proceed with the remaining services?

@mgrybyk which remaining services? I fixed all `global` usages from services we maintain and updated the docs so 3rd party services can update or keep the way they use it right now.

mgrybyk:
Shared store service for instance

christian-bromann:
@mgrybyk good catch! I adjusted the missing services. Good to go now?

mgrybyk:
How about failing tests?

packages/wdio-config/src/lib/ConfigParser.ts
```diff
@@ -115,24 +115,6 @@ export default class ConfigParser {
             this._config.exclude = this.setFilePathToFilterOptions(exclude, this._config.exclude!)
         }
 
-        /**
```
mgrybyk, line 118: Don't remember what this code do, I guess you removed it for a reason

mgrybyk (commented):
Do you still need to run npm run generate typings?

packages/wdio-config/src/lib/ConfigParser.ts
christian-bromann: yes, needed to increase test coverage and this wasn't covered by any tests and the bug that was fixed with it doesn't rely on this piece of code. Haven't seen any reason to keep it around.

docs/CustomServices.md
```diff
@@ -28,22 +26,67 @@ export default class CustomService {
...
+import CustomLauncherService from './launcher'
+import CustomWorkerService from './service'
+
+export default CustomWorkerService
+export const launcher = CustomLauncherService
+```
+
+## Service Error Handling
+
+An Error thrown during a service hook will be logged while the runner continues. If a hook in your launcher service is critical to the setup or teardown of the test runner, the `SevereServiceError` exposed from the `webdriverio` package can be used to stop the runner.
```
erwinheitzman, line 72: Should we add an example on how to stop the runner from within a server?

packages/wdio-devtools-service/src/index.ts
```diff
@@ -55,6 +57,15 @@ export default class DevToolsService implements WebdriverIO.HookFunctions {
     private _traceGatherer?: TraceGatherer
     private _devtoolsGatherer?: DevtoolsGatherer
 
+    constructor (
+        serviceOptions: ServiceOptions,
+        caps: WebDriver.Capabilities,
+        config: WebdriverIO.Config,
+        private _browser: WebdriverIO.BrowserObject | WebdriverIO.MultiRemoteBrowserObject
+    ) {
+        this._browser = _browser
```
erwinheitzman, line 66: this can be removed as `_browser` is already assigned and you are overwriting it with the same thing

packages/wdio-devtools-service/src/commands.ts
```diff
@@ -16,21 +16,23 @@ const log = logger('@wdio/devtools-service:CommandHandler')
...
     private _traceEvents?: TraceEvent[]
 
-    constructor (session: CDPSession, page: Page) {
-        this._session = session
-        this._page = page
-        this._networkHandler = new NetworkHandler(session)
+    constructor (
+        private _session: CDPSession,
+        private _page: Page,
+        private _browser: WebdriverIO.BrowserObject | WebdriverIO.MultiRemoteBrowserObject
```
erwinheitzman, line 24: why the change to _browser here?

packages/wdio-shared-store-service/src/service.ts
```diff
@@ -1,9 +1,18 @@
...
+interface ServiceOptions {}
 
 export default class SharedStoreService {
+    constructor (
+        serviceOptions: ServiceOptions,
+        caps: WebDriver.Capabilities[],
+        config: WebdriverIO.Config,
+        private _browser: WebdriverIO.BrowserObject | WebdriverIO.MultiRemoteBrowserObject
+    ) {
+        this._browser = _browser
```
erwinheitzman, line 13: can be removed

packages/wdio-webdriver-mock-service/src/index.ts
```diff
@@ -10,17 +10,25 @@ const ELEMENT_ID = '401c0039-3306-6a46-a98d-f5939870a249'
...
-    constructor() {
-        this.mock = new WebDriverMock()
+
+    constructor (
+        serviceOptions: ServiceOptions,
+        caps: WebDriver.Capabilities[],
+        config: WebdriverIO.Config,
+        private _browser: WebdriverIO.BrowserObject | WebdriverIO.MultiRemoteBrowserObject
+    ) {
+        this._browser = _browser
```
erwinheitzman, line 26: can be removed

erwinheitzman (changes requested):
Added some remarks

docs/CustomServices.md
christian-bromann: What do you mean with `from within a server`?

christian-bromann:
@mgrybyk @erwinheitzman thanks for the review. In order to make this work I had to add the following code:

```js
            /**
             * as services still use the original stub given that services are initiated before
             * a session was started we need to attach properties and commands from the actual
             * browser session to the stub as well
             */
            Object.entries(browser).forEach(([key, value]) => browserStub[key] = value)
            browser.commandList
                .filter((command) => typeof browser[command] === 'function')
                .forEach((command) => browserStub[command] = browser[command].bind(browser))
```

to `_initSession` which we call when initiating the actual session. The problem is that when services are created, we don't have a browser instance yet. So what we currently (with this PR) pass in is the stub. In order to allow users to use that stub in all hooks we have to overwrite its properties to mimic the actual browser instance (see code above).

With that we essentially maintain two browser instances: the stub in all services and the real browser instance. This might add to our memory consumption given we have a second object with 200 something properties and individual context attached to it. But I can't really think of a better solution here.

An alternative would be to pass in the browser instance in the `before` hook for the service to consume. What do you all think?

christian-bromann:
Ok, I modified the PR to now apply the browser to the `before` hook in all our services. Please review.

mgrybyk:
The solution with before hook looks good

christian-bromann:
@mgrybyk good to merge then?

docs/CustomServices.md
erwinheitzman: I meant from within a service, it was late haha

erwinheitzman:
Looks good. One question regarding your explanation, can we assign the properties of the stub to the browser object and then destroy the stub after the properties have been assigned?

erwinheitzman (approved):

mgrybyk (approved):

christian-bromann:
> and then destroy the stub after the properties have been assigned?

The stub should get garbage collected afaik unless someone holds a reference to it.

PR #6172, opened 2020-11-27
Start refactoring wdio-browserstack to TypeScript

Fabianopb:
## Proposed changes

Package refactoring to TypeScript as part of https://github.com/webdriverio/webdriverio/issues/5828
Starting with `index`, `constants` and `utils`

## Types of changes

- [x] TypeScript refactoring

## Checklist


packages/wdio-browserstack-service/src/util.ts
```diff
@@ -0,0 +1,64 @@
+import { BROWSER_DESCRIPTION } from './constants'
+
+type GenericBrowserCaps = {
+    [key: string]: { capabilities: BrowserCapabilities };
+};
+
+type BrowserCapabilities = GenericBrowserCaps & {
```
christian-bromann, line 7: We should use `WebDriver.Capabilities` types here instead of defining something custom

christian-bromann (changes requested):

packages/wdio-browserstack-service/src/util.ts
Fabianopb: Thanks for the tip! I gave it another round using `WebDriver.Capabilities` and `WebdriverIO.BrowserObject`, however two things I'm still not sure:

1. Not all keys from `BROWSER_DESCRIPTION` are within `Webdriver.Capabilities` so I had to included that in the local type.

2. `getBrowserCapabilities` receives a `string` `browserName` used to index `browser` and `caps` to get a `Webdriver.Capabilities` object. The way I found to solve it was through the union with 
```
{
    [key: string]: { capabilities: Webdriver.Capabilities };
}
```

What do you think?

packages/wdio-browserstack-service/src/util.ts
```diff
@@ -1,10 +1,17 @@
-import { BROWSER_DESCRIPTION } from './constants'
+import { BROWSER_DESCRIPTION, BrowserDescription } from './constants'
 
+type Capabilities = WebDriver.Capabilities & BrowserDescription & {
+    [key: string]: { capabilities: WebDriver.Capabilities };
+};
+
+type Browser = WebdriverIO.BrowserObject & {
```
christian-bromann, line 7: Just use `WebdriverIO.BrowserObject | WebdriverIO.MultibrowserObject`

packages/wdio-browserstack-service/src/util.ts
```diff
@@ -1,10 +1,17 @@
-import { BROWSER_DESCRIPTION } from './constants'
+import { BROWSER_DESCRIPTION, BrowserDescription } from './constants'
 
+type Capabilities = WebDriver.Capabilities & BrowserDescription & {
```
christian-bromann, line 3: same, use `WebDriver.Capabilities | WebdriverIO.MultiremoteBrowserCapabilities`

christian-bromann (commented):
Tests are missing. Let me know if I can help.

christian-bromann:
FYI: I closed #6015 as there seem to be no activity. So we can rewrite the whole package with this PR.

Fabianopb:
> FYI: I closed #6015 as there seem to be no activity. So we can rewrite the whole package with this PR.

Alright, so one PR to rule them all!

Fabianopb:
> Tests are missing. Let me know if I can help.

You mean converting the tests to TS, right?

christian-bromann:
> You mean converting the tests to TS, right?

Yes

packages/wdio-browserstack-service/src/util.ts
Fabianopb: Thanks, fixed!

packages/wdio-browserstack-service/src/util.ts
Fabianopb: Thanks, fixed, but apparently this has to be intersection instead of union! :)

Fabianopb:
Converted utils tests today and I'm unsure about one thing. This function has one argument only in the signature, but there's one test using a second arg, probably a mistake that wasn't caught before?

image

Fabianopb:
Strange new error breaking tests 🤔 
```
Error: packages/devtools/src/index.ts(2,10): error TS2616: 'UAParser' can only be imported by using 'import UAParser = require("../node_modules/@types/ua-parser-js")' or a default import.
```

christian-bromann:
> Strange new error breaking tests 🤔 Seems unrelated to this PR

@Fabianopb I just fixed in on master

christian-bromann:
> a mistake that wasn't caught before?

Correct, just remove the argument.

Fabianopb:
> I just fixed in on master

Thanks, I have rebased the branch

> Correct, just remove the argument.

Removed!

I'll continue refactoring the other files...

Fabianopb:
Started to give a try to `launcher.test.ts` and there were a few issues that we have to solve:

The launcher's `onComplete` doesn't accept an argument, but there are tests passing arguments:
image

Same for `addArgs`, this class method doesn't exist in `browserstack-local` but there are tests relying on mocking it:
image

How do you want to proceed? Do we remove those?

christian-bromann:
> The launcher's `onComplete` doesn't accept an argument, but there are tests passing argument

We can remove the arguments here.

> Same for `addArgs`, this class method doesn't exist in `browserstack-local` but there are tests relying on mocking it:

Do we use `addArgs` in our code somewhere? If so we should continue to test it and it seems some of the typings for BS local are wrong.

Fabianopb:
> Do we use `addArgs` in our code somewhere? If so we should continue to test it and it seems some of the typings for BS local are wrong.

It seems we don't use it anywhere, a global search shows results only in this test file!
image

packages/wdio-browserstack-service/src/launcher.ts
```diff
@@ -6,13 +6,27 @@ import logger from '@wdio/logger'
...
+    stop(callback: (err?: any) => void): void;
+}
+
+type ExtendedCapabilities = WebDriver.Capabilities & {
+    'browserstack.local'?: boolean;
+}
+
+type Capabilities = (ExtendedCapabilities | ExtendedCapabilities[])
+
 export default class BrowserstackLauncherService {
```
christian-bromann, line 20: Let's ensure we implement according to the service standard:
```suggestion
export default class BrowserstackLauncherService implements WebdriverIO.ServiceInstance {
```

packages/wdio-browserstack-service/src/launcher.ts
```diff
@@ -72,13 +86,13 @@ export default class BrowserstackLauncherService {
...
-            return process.kill(this.browserstackLocal.pid)
+            return process.kill(this.browserstackLocal.pid as number)
         }
 
-        let timer
+        let timer: NodeJS.Timeout
         return Promise.race([
             new Promise((resolve, reject) => {
-                this.browserstackLocal.stop((err) => {
+                this.browserstackLocal?.stop((err: any) => {
```
christian-bromann, line 95:
```suggestion
                this.browserstackLocal?.stop((err: Error) => {
```

packages/wdio-browserstack-service/src/util.ts
```diff
@@ -21,10 +24,11 @@ export function getBrowserDescription(cap) {
...
 /**
  * get correct browser capabilities object in both multiremote and normal setups
- * @param {*} caps browser capbilities object. In case of multiremote, the object itself should have a property named 'capabilities'
- * @param {*} browserName browser name in case of multiremote
+ * @param browser browser object
+ * @param caps browser capbilities object. In case of multiremote, the object itself should have a property named 'capabilities'
+ * @param browserName browser name in case of multiremote
  */
-export function getBrowserCapabilities(browser, caps, browserName) {
+export function getBrowserCapabilities(browser: Browser, caps: Capabilities = {}, browserName: string = '') {
```
christian-bromann, line 31: We should not default to empty string
```suggestion
export function getBrowserCapabilities(
    browser: Browser,
    caps: Capabilities = {},
    browserName?: string
) {
```

packages/wdio-browserstack-service/tsconfig.json
```diff
@@ -3,7 +3,8 @@
   "compilerOptions": {
     "baseUrl": ".",
     "outDir": "./build",
-    "rootDir": "./src"
+    "rootDir": "./src",
+    "types": ["../webdriver", "../webdriverio", "./"]
```
christian-bromann, line 7:
```suggestion
    "types": ["webdriverio", "./"]
```

christian-bromann:
@Fabianopb let me know if this is ready for merge or if you plan more changes. Thanks for all the work so far 👏

packages/wdio-browserstack-service/src/util.ts
Fabianopb: Done and fixed types

Fabianopb:
> @Fabianopb let me know if this is ready for merge or if you plan more changes. Thanks for all the work so far 👏

Fixed according to suggestions! There's still `service.js` to be converted, wanna do it in a separate PR?

I have noticed that `utils.test.ts` have type errors because the `browser` mock lacks properties from the type for `Browser` we defined in `utils.ts`. Should we create a mock or do you have any other suggestion?

I noticed also this new error:
```
Error: packages/webdriverio/src/utils/index.ts(165,30): error TS2349: This expression is not callable.
```

christian-bromann:
> There's still `service.js` to be converted, wanna do it in a separate PR?

We can have the whole package as one PR.

> I have noticed that `utils.test.ts` have type errors because the `browser` mock lacks properties from the type for `Browser` we defined in `utils.ts`. Should we create a mock or do you have any other suggestion?

Can you point out which test?

Fabianopb:
> Can you point out which test?

Basically all tests in `utils.test.ts`, this one for example is missing lots of properties in the `browser` mock object passed to the function, which contains only `{ capabilities: { browser: 'browser' } }`

image

christian-bromann:
@Fabianopb in tests we don't necessarily need to have total type safety. It is absolutely fine to pretend objects to be from a certain type to keep tests readable and simple (e.g. check out [those tests](https://github.com/webdriverio/webdriverio/blob/cb-rewrite-runner/packages/wdio-runner/tests/index.test.ts#L68-L75) I've just worked on).

packages/wdio-browserstack-service/src/service.ts
```diff
@@ -1,12 +1,72 @@
...
+
+type Pickle = {
+    name: string;
+}
+
+type MultiRemoteAction = (sessionId: string, browserName?: string) => ReturnType<GotRequestFunction> | Promise<void>;
+
+type SessionResponse = {
+    // eslint-disable-next-line camelcase
+    automation_session: {
```
Fabianopb, line 52: @christian-bromann I have created types here just to simulate what would be needed to solve all the errors in this file, however I'm not sure where they should come from, I'd need some help here! :)

packages/wdio-browserstack-service/src/service.ts
```diff
@@ -1,12 +1,72 @@
...
+
+type Feature = {
+    document: {
+        feature: {
+            name: string;
+        }
+    }
+};
+
+type Results = {
```
christian-bromann, line 37: use `WebdriverIO.TestResult` instead

packages/wdio-browserstack-service/src/service.ts
```diff
@@ -1,12 +1,72 @@
...
+    config: Config;
+    sessionBaseUrl: string;
+    failReasons: string[];
+    scenariosThatRan: string[];
+    preferScenarioName: boolean;
+    strict: boolean;
+    failureStatuses: string[];
+    caps: Capabilities;
+    browser?: Browser;
+    fullTitle?: string;
```
christian-bromann, line 68: Can we make them private starting with an `_`?

```suggestion
    private _config: Config;
    private _sessionBaseUrl: string;
    private _failReasons: string[];
    private _scenariosThatRan: string[];
    private _preferScenarioName: boolean;
    private _strict: boolean;
    private _failureStatuses: string[];
    private _caps: Capabilities;
    private _browser?: Browser;
    private _fullTitle?: string;
```

packages/wdio-browserstack-service/src/service.ts
```diff
@@ -1,12 +1,72 @@
...
+    constructor (options: BrowserstackConfig = {}, caps: Capabilities, config: Config) {
```
christian-bromann, line 69: defining like this allows us to remove `this.config = config` in the constructor
```suggestion
    constructor (options: BrowserstackConfig = {}, caps: Capabilities, private _config: Config) {
```

packages/wdio-browserstack-service/src/service.ts
```diff
@@ -1,12 +1,72 @@
...
         this.config = config
         this.sessionBaseUrl = 'https://api.browserstack.com/automate/sessions'
```
christian-bromann, line 71: let's define defaults in the class head

packages/wdio-browserstack-service/src/service.ts
```diff
@@ -1,12 +1,72 @@
...
+
+type SessionResponse = {
+    // eslint-disable-next-line camelcase
+    automation_session: {
+        // eslint-disable-next-line camelcase
+        browser_url: string
+    }
+}
+
 export default class BrowserstackService {
```
christian-bromann, line 58:
```suggestion
export default class BrowserstackService implements WebdriverIO.ServiceInstance {
```

packages/wdio-browserstack-service/src/service.ts
```diff
@@ -1,12 +1,72 @@
...
 
 const log = logger('@wdio/browserstack-service')
 
+type Capabilities = WebDriver.Capabilities & WebdriverIO.MultiRemoteCapabilities;
+
+type Browser = WebdriverIO.BrowserObject & WebdriverIO.MultiRemoteBrowserObject;
+
+type Config = WebdriverIO.Config & { cucumberOpts?: { strict: boolean } };
+
+type Suite = {
```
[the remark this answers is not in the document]
Fabianopb: I had actually tried this, but got errors:
image

The `Suite` type used here, seems to come from `jasmine-framework.d.ts`: https://github.com/webdriverio/webdriverio/blob/master/packages/wdio-jasmine-framework/jasmine-framework.d.ts#L7-L11

packages/wdio-browserstack-service/src/service.ts
```diff
@@ -1,12 +1,72 @@
...
+    title: string;
+}
+
+type Test = {
```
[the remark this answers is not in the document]
Fabianopb: Same deal here, the type that should be used seems to be the one coming from here: https://github.com/webdriverio/webdriverio/blob/master/packages/wdio-jasmine-framework/jasmine-framework.d.ts#L13-L16

packages/wdio-browserstack-service/src/service.ts
```diff
@@ -1,12 +1,72 @@
...
+    fullName: string,
+    parent: string,
+    title: string
+};
+
+type Context = any;
+
+type Specs = any;
+
+type Feature = {
```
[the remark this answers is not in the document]
Fabianopb: Hummm... It seems we don't have a type called `Feature` in `WebdriverIO` 🤔

packages/wdio-browserstack-service/src/service.ts
Fabianopb: Same deal, `TestResult` doesn't include the properties we use here, perhaps the one we're looking for is: https://github.com/webdriverio/webdriverio/blob/master/packages/wdio-cucumber-framework/cucumber-framework.d.ts#L15-L17

packages/wdio-browserstack-service/src/service.ts
Fabianopb: Done and fixed all references throughout the file!

packages/wdio-browserstack-service/src/service.ts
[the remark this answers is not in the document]
christian-bromann: both Mocha and Jasmine suite interfaces have `title` properties.

packages/wdio-browserstack-service/src/service.ts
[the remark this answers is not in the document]
christian-bromann: Let's use `WebdriverIO.Test` for now like in other services and clean this up later.

packages/wdio-browserstack-service/src/service.ts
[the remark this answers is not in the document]
christian-bromann: right, nevermind then

packages/wdio-browserstack-service/src/service.ts
christian-bromann: [this](https://github.com/webdriverio/webdriverio/blob/master/packages/webdriverio/webdriverio-core.d.ts#L257-L263) interface should be used

Fabianopb: I think it needs some augmentation as it lacks `exception` and `status`, I'll add that!

packages/wdio-browserstack-service/src/service.ts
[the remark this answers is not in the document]
Fabianopb: I've augmented the interfaces in `browserstack-service.d.ts`

christian-bromann:
@Fabianopb anything I can do to push this over the line?

Fabianopb:
Sorry @christian-bromann, I ended stalling this as holidays came! I think some types in tests were still broken if you want to take a look, otherwise I can return to this after New Year!

Fabianopb:
Happy New Year @christian-bromann ! 🎉 
I'm back to biz, just finished converting the last test file, please take a careful look to see if we have all in place now!

packages/wdio-browserstack-service/src/service.ts
```diff
@@ -0,0 +1,240 @@
...
+    private _browser?: Browser;
+    private _fullTitle?: string;
+    constructor (options: BrowserstackConfig = {}, caps: Capabilities, private _config: WebdriverIO.Config) {
+        // Cucumber specific
+        this._preferScenarioName = Boolean(options.preferScenarioName)
+        this._strict = Boolean(_config.cucumberOpts && _config.cucumberOpts.strict)
+        // See https://github.com/cucumber/cucumber-js/blob/master/src/runtime/index.ts#L136
+        this._strict && this._failureStatuses.push('pending')
+        this._caps = caps
+    }
```
christian-bromann, line 54: some improvements here:
```suggestion
    constructor (private _options: BrowserstackConfig = {}, private _caps: Capabilities, private _config: WebdriverIO.Config) {
        if (this._strict) {
            this._failureStatuses.push('pending')
        }
    }
```

`_strict` is nowhere else used so we can get rid of it and by making the whole options object a class property we don't need to specifically define `_preferScenarioName`. The comment pointed to line of code that doesn't really indicate any information as to why this was implemented like it.

packages/wdio-browserstack-service/src/service.ts
```diff
@@ -0,0 +1,240 @@
...
+
+        if (!passed) {
+            this._failReasons.push((error && error.message) || 'Unknown Error')
+        }
+    }
+
+    after(result: number) {
+        // For Cucumber: Checks scenarios that ran (i.e. not skipped) on the session
+        // Only 1 Scenario ran and option enabled => Redefine session name to Scenario's name
+        if (this._preferScenarioName && this._scenariosThatRan.length === 1){
```
christian-bromann, line 119: based on the suggestion above:
```suggestion
        if (this._options._preferScenarioName && this._scenariosThatRan.length === 1){
```

packages/wdio-browserstack-service/src/service.ts
```diff
@@ -0,0 +1,240 @@
...
+        browser_url: string
+    }
+}
+
+export default class BrowserstackService implements WebdriverIO.ServiceInstance {
+    private _sessionBaseUrl: string = 'https://api.browserstack.com/automate/sessions';
+    private _failReasons: string[] = [];
+    private _scenariosThatRan: string[] = [];
+    private _preferScenarioName: boolean;
+    private _strict: boolean;
```
christian-bromann, line 42: With that, these aren't needed anymore
```suggestion
```

packages/wdio-browserstack-service/src/service.ts
```diff
@@ -0,0 +1,240 @@
...
+    private _failureStatuses: string[] = ['failed', 'ambiguous', 'undefined', 'unknown'];
+    private _caps: Capabilities;
```
christian-bromann, line 44: with that this gets defined in the constructor.

packages/wdio-browserstack-service/src/service.ts
```diff
@@ -0,0 +1,240 @@
...
+            exception = exception || (status === 'pending'
+                ? `Some steps/hooks are pending for scenario "${pickle.name}"`
+                : 'Unknown Error')
+
+            this._failReasons.push(exception)
+        }
+    }
+
+    async onReload(oldSessionId: string, newSessionId: string) {
+        const { _browser } = this
```
christian-bromann, line 153: I would prefer to just use `this._browser`

packages/wdio-browserstack-service/src/service.ts
```diff
@@ -0,0 +1,240 @@
...
+        log.debug(`Updating Browserstack session at ${sessionUrl} with request body: `, requestBody)
+        return got.put(sessionUrl, {
+            json: requestBody,
+            username: this._config.user,
+            password: this._config.key
+        })
+    }
+
+    async _printSessionURL() {
+        const { _browser } = this
```
christian-bromann, line 222: same

packages/wdio-browserstack-service/src/util.ts
```diff
@@ -0,0 +1,47 @@
+import { BROWSER_DESCRIPTION } from './constants'
+
+type Capabilities = WebDriver.Capabilities & WebdriverIO.MultiRemoteCapabilities;
+
+type Browser = WebdriverIO.BrowserObject & WebdriverIO.MultiRemoteBrowserObject;
```
christian-bromann, line 5: we have defined these in `service.ts` already. I suggest to move those into a `types.ts` and import them here.

christian-bromann (commented):
Thank you so much for all your work 😊 I added some comments, mostly optimisations that are not much related to the rewrite.

packages/wdio-browserstack-service/src/launcher.ts
```diff
@@ -6,13 +6,27 @@ import logger from '@wdio/logger'
...
 
-export default class BrowserstackLauncherService {
-    constructor (options, capabilities, config) {
+type BrowserstackLocal = BrowserstackLocalLauncher.Local & {
+    pid?: number;
+    stop(callback: (err?: any) => void): void;
+}
+
+type ExtendedCapabilities = WebDriver.Capabilities & {
+    'browserstack.local'?: boolean;
```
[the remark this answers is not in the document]
Fabianopb: Done!

packages/wdio-browserstack-service/src/service.ts
```diff
@@ -0,0 +1,240 @@
+import logger from '@wdio/logger'
+import got, { GotRequestFunction } from 'got'
+import WebDriver from 'webdriver'
```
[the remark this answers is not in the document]
Fabianopb: Removed WebDriver import

packages/wdio-browserstack-service/src/service.ts
```diff
@@ -0,0 +1,240 @@
...
+
+type MultiRemoteAction = (sessionId: string, browserName?: string) => ReturnType<GotRequestFunction> | Promise<void>;
+
+type SessionResponse = {
+    // eslint-disable-next-line camelcase
+    automation_session: {
+        // eslint-disable-next-line camelcase
+        browser_url: string
+    }
+}
```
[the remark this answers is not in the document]
Fabianopb: Moved!

packages/wdio-browserstack-service/src/service.ts
Fabianopb: Removing these will break some tests:
```ts
this._preferScenarioName = Boolean(_options.preferScenarioName)
this._strict = Boolean(_config.cucumberOpts && _config.cucumberOpts.strict)
```

christian-bromann: the tests should be updated then.

packages/wdio-browserstack-service/src/service.ts
Fabianopb: Done! (I didn't commit this suggestion as "preferScenarioName" shouldn't be suffixed by underscore)

packages/wdio-browserstack-service/src/service.ts
Fabianopb: Removed from here and from tests

packages/wdio-browserstack-service/src/service.ts
Fabianopb: 👍

packages/wdio-browserstack-service/src/service.ts
Fabianopb: 👍

packages/wdio-browserstack-service/src/service.ts
Fabianopb: 👍

packages/wdio-browserstack-service/src/service.ts
```diff
@@ -0,0 +1,207 @@
...
+    private _browser?: Browser;
+    private _fullTitle?: string;
+    constructor (private _options: BrowserstackConfig = {}, private _caps: Capabilities, private _config: WebdriverIO.Config) {
+        // Cucumber specific
+        const strict = Boolean(_config.cucumberOpts && _config.cucumberOpts.strict)
+        // See https://github.com/cucumber/cucumber-js/blob/master/src/runtime/index.ts#L136
+        if (strict) {
+            this._failureStatuses.push('pending')
+        }
+    }
```
Fabianopb, line 23: @christian-bromann after the last changes one question remains: is this how this strict check should still happen?

packages/wdio-browserstack-service/src/service.ts
Fabianopb: Did some changes, but I still have a question, follow up in this comment: https://github.com/webdriverio/webdriverio/pull/6172#discussion_r553558912

packages/wdio-browserstack-service/src/service.ts
christian-bromann: I think so, what concerns do you have?

Fabianopb: None, I think I had the impression from this suggestion (https://github.com/webdriverio/webdriverio/pull/6172#discussion_r551799101) that this line under "Cucumber specific" comment was to be removed.
```ts
// Cucumber specific
const strict = Boolean(_config.cucumberOpts && _config.cucumberOpts.strict)
```

Fabianopb: So, this should be good to go then? 🙂

christian-bromann (approved):
Yes, this looks good! Thank you so much 🙏

Fabianopb:
> Yes, this looks good! Thank you so much 🙏

Yay! That was a long stretch! Thanks for the reviews! 🎉

reidj32:
I realize this has been merged, but after doing a package upgrade today I found an issue which might be related to this and wanted to confirm before opening a new issue. I have both @wdio/browserstack-service (6.12.0) and @wdio/cucumber-framework (6.11.1) installed and they are using TypeScript. After taking the most recent upgrades, there appears to be a type collision in the WebdriverIO module. Here is the error I receive when compiling:
image

After some digging, I found that the same interface extension occurs in both of these packages which is causing the issue. If I open `node_modules/@wdio/browserstack-service/browserstack-service.d.ts` and make the following modifications, I am able to compile without issue.
```typescript
declare module WebdriverIO {
    interface ServiceOption extends BrowserstackConfig {}

    // interface Config extends CucumberOptsConfig {

    // }

    interface Suite {
        title: string;
        fullName: string;
        file: string;
    }

    interface Test extends Suite {
        parent: string;
        passed: boolean;
    }

    interface TestResult {
        exception: string;
        status: string;
    }
}

interface BrowserstackConfig {
    /**
     * Set this to true to enable routing connections from Browserstack cloud through your computer.
     * You will also need to set `browserstack.local` to true in browser capabilities.
     */
    browserstackLocal?: boolean;
    /**
     * Cucumber only. Set this to true to enable updating the session name to the Scenario name if only
     * a single Scenario was ran. Useful when running in parallel
     * with [wdio-cucumber-parallel-execution](https://github.com/SimitTomar/wdio-cucumber-parallel-execution).
     */
    preferScenarioName?: boolean;
    /**
     * Set this to true to kill the browserstack process on complete, without waiting for the
     * browserstack stop callback to be called. This is experimental and should not be used by all.
     */
    forcedStop?: boolean;
    /**
     * Specified optional will be passed down to BrowserstackLocal. See this list for details:
     * https://stackoverflow.com/questions/39040108/import-class-in-definition-file-d-ts
     */
    opts?: Partial<import('browserstack-local').Options>
}

// interface CucumberOptsConfig {
//     cucumberOpts?: CucumberOpts
// }

// interface CucumberOpts {
//     /**
//      * Fail if there are any undefined or pending steps
//      * @default false
//      */
//     strict?: boolean
// }
```

I am curious why the browserstack service cares about the cucumber types?

christian-bromann:
> I am curious why the browserstack service cares about the cucumber types?

Because it handles reporting job status differently depending whether [cucumberOpts are set](https://github.com/webdriverio/webdriverio/blob/master/packages/wdio-browserstack-service/src/service.ts#L18) to strict. Let's report this or also PRs would be appreciated 🙏

PR #6176, opened 2020-11-29
selenium-standalone: fix driver keys condition

mgrybyk:
## Proposed changes

Fix drivers keys condition. There should be at least one driver provided, not two

## Types of changes

## Checklist

### DIFF
--- a/packages/wdio-selenium-standalone-service/src/launcher.ts
+++ b/packages/wdio-selenium-standalone-service/src/launcher.ts
@@ -53,7 +53,7 @@ export default class SeleniumStandaloneLauncher {
         this.skipSeleniumInstall = Boolean(options.skipSeleniumInstall)
 
         // simplified mode
-        if (options.drivers && Object.keys(options.drivers).length > 0) {
+        if (this.isSimplifiedMode(options)) {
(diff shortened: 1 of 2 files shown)


packages/wdio-selenium-standalone-service/src/launcher.ts
```diff
@@ -53,7 +53,7 @@ export default class SeleniumStandaloneLauncher {
         this.skipSeleniumInstall = Boolean(options.skipSeleniumInstall)
 
         // simplified mode
-        if (options.drivers && Object.keys(options.drivers).length > 1) {
+        if (options.drivers && Object.keys(options.drivers).length > 0) {
```
christian-bromann, line 56: Maybe we move `0` to a const to name it property which helps to understand its meaning.

christian-bromann (commented):
Do you think it makes sense to add a unit test for this?

packages/wdio-selenium-standalone-service/src/launcher.ts
mgrybyk: I'll add a test and add function with meaningful name

mgrybyk:
Let's release this one

christian-bromann (approved):
👍

PR #5836, opened 2020-12-02
Tracking doc: rewrite `@wdio/firefox-profile-service` package into TypeScript

christian-bromann:

- if you are interested converting source files from `/packages/wdio-firefox-profile-service` into TypeScript please comment on this thread and mention the files you want to convert to avoid duplicate work
- if you make a PR, label it with `TypeScript` and the package name `wdio-firefox-profile-service` so we can easier find all PRs

Happy rewriting 😅

### DIFF
--- a/packages/wdio-firefox-profile-service/tests/launcher.test.ts
+++ b/packages/wdio-firefox-profile-service/tests/launcher.test.ts
@@ -1,5 +1,7 @@
 import Launcher from '../src/launcher'
 import FirefoxProfile from 'firefox-profile'
+import type WebDriver from 'webdriver'
+import type WebdriverIO from 'webdriverio'
 
 describe('Firefox profile service', () => {
     describe('onPrepare', () => {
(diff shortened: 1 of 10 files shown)


packages/wdio-firefox-profile-service/tests/launcher.test.ts
```diff
@@ -176,5 +178,24 @@ describe('Firefox profile service', () => {
...
+                profileDirectory: '/tmp/firefox-profile'
+            }
+            const capabilities: WebDriver.DesiredCapabilities[] = [{
+                browserName : 'firefox',
+            }]
+
+            ;(FirefoxProfile.copy as any as jest.Mock).mockImplementationOnce((profileDirectory: string, cb: Function) => cb())
+            const service = new Launcher(options)
+            await service.onPrepare({}, capabilities)
+        })
```
klamping, line 193: Should these tests have assertions? It looks like the code is being run, but I don't see an assertion for this one or the next test that was added.

christian-bromann: no assertion needed because the test would fail if we wouldn't return early in the onPrepare hook. Maybe I can make this more clear?

PR #5829, opened 2020-12-04
Tracking doc: rewrite `@wdio/cli` package into TypeScript

christian-bromann:

- if you are interested converting source files from `/packages/wdio-cli` into TypeScript please comment on this thread and mention the files you want to convert to avoid duplicate work
- if you make a PR, label it with `TypeScript` and the package name `wdio-cli` so we can easier find all PRs

Happy rewriting 😅


packages/wdio-cli/src/commands/config.ts
```diff
@@ -27,31 +29,34 @@ export const cmdArgs = {
...
     const answers = await getAnswers(yes)
 
     const packageAnswers = ['reporters', 'runner', 'services', 'framework']
 
-    Object.keys(answers).forEach((key) => {
+    Object.keys(answers).forEach((key: keyof Questionnair) => {
         if (packageAnswers.includes(key)) {
             if (Array.isArray(answers[key])) {
-                answers[key] = answers[key].map(answer => convertPackageHashToObject(answer))
+                // @ts-ignore
```
mgrybyk, line 50: is it possible to remove ts ignore?

packages/wdio-cli/src/interface.ts
```diff
@@ -101,8 +144,12 @@ export default class WDIOCLInterface extends EventEmitter {
         return _logger(...details)
     }
 
-    onTestError(payload) {
-        let error = { type: 'Error', message: typeof payload.error === 'string' ? payload.error : 'Unknown error.', stack: null }
+    onTestError (payload: CLIInterfaceEvent) {
+        const error: TestError = {
```
mgrybyk, line 148: `stack` is missing here

packages/wdio-cli/src/types.ts
```diff
@@ -0,0 +1,107 @@
+import { BACKEND_CHOICES, MODE_OPTIONS, REGION_OPTION, COMPILER_OPTION_ANSWERS } from './constants'
```
mgrybyk, line 1: shouldn't the extension of this file be `.d.ts`?

packages/wdio-cli/src/utils.ts
```diff
@@ -135,21 +147,21 @@ export function findInConfig(config, type) {
     return config.match(regex)
 }
 
-export function replaceConfig(config, type, name) {
+export function replaceConfig (config: string, type: string, name: string) {
+    if (type === 'framework') {
```
mgrybyk, line 151: is it ok that the condition order changed?

mgrybyk (changes requested):

packages/wdio-cli/src/types.ts
christian-bromann: There is no reason why it should (or are there any advantages I am missing?). Before the v7 release I will have all types properly be exposed from the `index.d.ts` file.

packages/wdio-cli/src/utils.ts
christian-bromann: that should be irrelevant

packages/wdio-cli/src/types.ts
mgrybyk: Not quite sure, I supposed `.d.ts` extension is used for type definitions.

mgrybyk (approved):

PR #5841, opened 2020-12-07
Tracking doc: rewrite `@wdio/mocha-framework` package into TypeScript

christian-bromann:

- if you are interested converting source files from `/packages/wdio-mocha-framework` into TypeScript please comment on this thread and mention the files you want to convert to avoid duplicate work
- if you make a PR, label it with `TypeScript` and the package name `wdio-mocha-framework` so we can easier find all PRs

Happy rewriting 😅


packages/wdio-mocha-framework/src/utils.ts
```diff
@@ -1,5 +1,8 @@
-export function loadModule (name, context) {
+import type { MochaContext } from './types'
+
+export function loadModule (name: string, context: MochaContext) {
     try {
+        // @ts-ignore
         module.context = context
```
mgrybyk, line 6: [_not relations to the PR_] Looks like some dark magic :)

mgrybyk (changes requested):

packages/wdio-mocha-framework/src/index.ts
```diff
@@ -26,32 +28,37 @@ const DEFAULT_INTERFACE_TYPE = 'bdd'
...
-        this.specs = specs
-        this.config = Object.assign({
+    private _mocha?: Mocha
+    private _runner?: Runner
+    private _specLoadError?: Error
+
+    private _level = 0
+    private _hasTests = true
+    private _suiteIds: string[] = ['0']
+    private _suiteCnt: Map<string, number> = new Map()
```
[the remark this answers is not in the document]
christian-bromann: I looked into it and apparently this is desired for the same reason we defined `_suiteIds` as string (see e.g. [here](https://github.com/webdriverio/webdriverio/blob/master/packages/wdio-mocha-framework/src/index.js#L324)). How this plays out you can see in [this test](https://github.com/webdriverio/webdriverio/blob/master/packages/wdio-mocha-framework/tests/adapter.test.js#L360-L412). Definitely worth improving eventually but not as part of this PR.

packages/wdio-mocha-framework/src/utils.ts
christian-bromann: true, let's keep it as is and eventually clean this up.

mgrybyk: agree, no idea how it works

mgrybyk (approved):

PR #6216, opened 2020-12-14
Tracking doc: rewrite `@wdio/runner` package into TypeScript

christian-bromann:
As we move the code base to TypeScript we also should rewrite `@wdio/runner`.


packages/wdio-runner/src/index.ts
```diff
@@ -5,19 +5,49 @@ import EventEmitter from 'events'
...
+type RunParams = {
+    cid: string
+    args: Args
+    specs: string[]
+    caps: Capability
+    configFile: string
+    retries: number
+}
+
+interface Framework {
```
mgrybyk, line 28: maybe it can be renamed to test framework or similar because Framework looks too generic

packages/wdio-runner/src/index.ts
```diff
@@ -90,47 +123,49 @@ export default class Runner extends EventEmitter {
...
             isMultiremote,
             sessionId: browser.sessionId,
             capabilities: isMultiremote
-                ? browser.instances.reduce((caps, browserName) => {
-                    caps[browserName] = browser[browserName].capabilities
-                    caps[browserName].sessionId = browser[browserName].sessionId
+                ? multiRemoteBrowser.instances.reduce((caps: WebdriverIO.MultiRemoteCapabilities, browserName) => {
+                    // @ts-ignore loosly typed multiremote caps
+                    caps[browserName] = multiRemoteBrowser[browserName].capabilities
+                    // @ts-ignore this might not be needed
```
mgrybyk, line 154: Looks like it should be required according to how shutdown function works

packages/wdio-runner/src/index.ts
```diff
@@ -29,59 +59,62 @@ export default class Runner extends EventEmitter {
...
+        this._framework = initialisePlugin(this._config.framework as string, 'framework').default as unknown as Framework
+        this._framework = await this._framework.init(cid, this._config, specs, caps, this._reporter)
+        process.send!({ name: 'testFrameworkInit', content: { cid, caps, specs, hasTests: this._framework.hasTests() } })
+        if (!this._framework.hasTests()) {
             return this._shutdown(0, retries)
         }
 
-        browser = await this._initSession(this.config, this.caps, browser)
-        this.inWatchMode = Boolean(this.config.watch)
+        browser = await this._initSession(this._config as SingleConfigOption, this._caps, browser as WebdriverIO.BrowserObject) as WebdriverIO.BrowserObject
```
mgrybyk, line 117: What if you declare `browser` as `WebdriverIO.BrowserObject | null` at line 89 `let browser = await this._startSession({`.

Maybe in such a case, the additional `as` can be removed.

tests/smoke.runner.js
```diff
@@ -27,13 +27,24 @@ async function runTests (tests) {
...
-    if (process.env.CI || testFilter) {
+    if (process.env.CI || (testFilter && !testFilter.startsWith('--'))) {
         // sequential
         const testsFiltered = testFilter ? tests.filter(test => test.name === testFilter) : tests
         for (let test of testsFiltered) {
             await test()
         }
     } else {
+        // run sequentially
+        if ([...process.argv].includes('--sequential')) {
```
mgrybyk, line 38: this is some additional way to run tests? What is the difference between passing CI=true ?

christian-bromann: yeah, it runs the smoke tests sequentially rather than in parallel which helps to better identify which one fails. I will add some more docs to it.

christian-bromann: ah actually it is the same 🙈

christian-bromann:
@mgrybyk addressed comments, anything left to do?

PR #5939, opened 2020-12-15
Remove `browser.options` across the code base

christian-bromann:
This issue is based on conversations we had in https://github.com/webdriverio/webdriverio/pull/5919

Basically we currently have `browser.config` and `browser.options` which both represent a similar data structure. It seems there is a lot of ambiguity. I would say let's get rid of one of them.

### DIFF
--- a/docs/BrowserObject.md
+++ b/docs/BrowserObject.md
@@ -91,42 +91,6 @@ console.log(browser.config)
 console.log(browser.config.fakeUser) // outputs: "maxmustermann"
 ```
 
-### Configurations versus Options
-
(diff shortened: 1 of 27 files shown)


erwinheitzman:
When checking the types and the following documentation: Custom configurations should not be confused with Options.

I get the idea that the config holds the options but also any custom properties that the user has added to the config, if this is true we do need both. Can you clarify if this is now no longer the case?

christian-bromann:
Closing. We might can consider this for the next major release. Currently the only difference is that `browser.config` contains also custom configurations/properties while `browser.options` contains only the ones that are supported. If we remove one of them we need to provide the ability to set custom options somehow.

PR #6245, opened 2020-12-18
fix: Add config parameter for before session hook

denis-a1qa:
## Proposed changes
Before session hook is missing "config" as first parameter. This causes issues for example in sauce-service because user and key are incorrectly parsed
The issue appeared after this commit:
https://github.com/webdriverio/webdriverio/pull/6216/files#diff-84608493ac5a71c7f4e0d78854aaa5920e875a89b3c6c88dad8f72621ea40ce3R104

``` 
beforeSession (config, capabilities) {
        this.config = config
        this.capabilities = capabilities
        this.api = new SauceLabs(this.config)
        this.isRDC = 'testobject_api_key' in this.capabilities // this always returns undefined because instead of capabilities code is reading specs
        this.isServiceEnabled = true

        /**
         * if no user and key is specified even though a sauce service was
         * provided set user and key with values so that the session request
         * will fail (not for RDC tho due to other auth mechansim)
         */
        if (!this.isRDC && !config.user) {
            this.isServiceEnabled = false
            config.user = 'unknown_user'
        }
        if (!this.isRDC && !config.key) {
            this.isServiceEnabled = false
            config.key = 'unknown_key'
        }
    }
```
## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/wdio-runner/src/index.ts
+++ b/packages/wdio-runner/src/index.ts
@@ -101,7 +101,7 @@ export default class Runner extends EventEmitter {
             caps as unknown as WebDriver.DesiredCapabilities,
             args.ignoredWorkerServices
         ).map(this._configParser.addService.bind(this._configParser))
-        await executeHooksWithArgs('beforeSession', this._config.beforeSession, [this._caps, this._specs])
+        await executeHooksWithArgs('beforeSession', this._config.beforeSession, [this._config, this._caps, this._specs])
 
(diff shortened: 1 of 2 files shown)


denis-a1qa:
I am not sure but there is also might be similar issue in after session hook

christian-bromann (approved):
Thanks 👍

PR #3583, opened 2019-02-20
Adding outputDir to standalone

Gilad-WT:
## Proposed changes

[//]: # (Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue.)
Added outputDir option for standalone (no testrunner) 
## Types of changes

[//]: # (What types of changes does your code introduce to WebdriverIO?)
[//]: # (_Put an `x` in the boxes that apply_)

- [ ] Bugfix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

[//]: # (_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._)

- [X] I have read the [CONTRIBUTING](https://github.com/webdriverio/webdriverio/blob/master/CONTRIBUTING.md) doc
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

## Further comments

[//]: # (If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...)

PR #6270, opened 2020-12-28
webdriverio: Fix log path in standalone mode

koggdal:
## Proposed changes

When running in standalone mode (directly using the `remote` function), the logs started going to stdout instead of to a file in version 6.10.6. The issue started in this commit: e78ecef891b2de4a2f7c198117d35db8debf1c68

That commit moves the setting of `process.env.WDIO_LOG_PATH` into the `webdriver` package, but it still has to remain in the `webdriverio` package as well for standalone mode to work. Looks like this feature was introduced in PR #3583.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/packages/devtools/src/constants.ts
+++ b/packages/devtools/src/constants.ts
@@ -80,6 +80,12 @@ export const DEFAULTS: DefaultOptions<WebDriver.Options> = {
         match: /(trace|debug|info|warn|error|silent)/
     },
     /**
+     * directory for log files
+     */
+    outputDir: {
+        type: 'string'
+    },
+    /**
      * maxConnectionRetries in chrome-launcher
      */
     connectionRetryCount: {
--- a/packages/devtools/src/index.ts
+++ b/packages/devtools/src/index.ts
@@ -1,4 +1,5 @@
 import os from 'os'
+import path from 'path'
 import UAParser from 'ua-parser-js'
 import { v4 as uuidv4 } from 'uuid'
 
@@ -27,6 +28,13 @@ export default class DevTools {
 
         if (params.logLevel && (!options.logLevels || !(options.logLevels as any)['devtools'])) {
             logger.setLevel('devtools', params.logLevel)
+        }
+
+        /**
+         * Store all log events in a file
+         */
+        if (params.outputDir && !process.env.WDIO_LOG_PATH) {
+            process.env.WDIO_LOG_PATH = path.join(params.outputDir, 'wdio.log')
         }
 
         const browser = await launch(params.capabilities as WebDriver.Capabilities)
--- a/packages/devtools/src/index.ts
+++ b/packages/devtools/src/index.ts
@@ -37,6 +37,8 @@ export default class DevTools {
             process.env.WDIO_LOG_PATH = path.join(params.outputDir, 'wdio.log')
         }
 
+        log.info('Initiate new session using the DevTools protocol')
+
         const browser = await launch(params.capabilities as WebDriver.Capabilities)
         const pages = await browser.pages()
         const driver = new DevToolsDriver(browser, pages)
--- a/packages/webdriver/src/constants.ts
+++ b/packages/webdriver/src/constants.ts
@@ -63,6 +63,12 @@ export const DEFAULTS: DefaultOptions<Options> = {
         match: /(trace|debug|info|warn|error|silent)/
     },
     /**
+     * directory for log files
+     */
+    outputDir: {
+        type: 'string'
+    },
+    /**
      * Timeout for any WebDriver request to a driver or grid
      */
     connectionRetryTimeout: {
--- a/packages/webdriver/src/index.ts
+++ b/packages/webdriver/src/index.ts
@@ -27,6 +27,13 @@ export default class WebDriver {
         }
 
         /**
+         * Store all log events in a file
+         */
+        if (params.outputDir && !process.env.WDIO_LOG_PATH) {
+            process.env.WDIO_LOG_PATH = path.join(params.outputDir, 'wdio.log')
+        }
+
+        /**
          * if the server responded with direct connect information, update the
          * params to speak directly to the appium host instead of a load
          * balancer (see https://github.com/appium/python-client#direct-connect-urls
@@ -41,13 +48,6 @@ export default class WebDriver {
             params.hostname = directConnectHost
             params.port = directConnectPort
             params.path = directConnectPath
-        }
-
-        /**
-         * Store all log events in a file
-         */
-        if (params.outputDir) {
-            process.env.WDIO_LOG_PATH = path.join(params.outputDir, 'wdio.log')
         }
 
         const { sessionId, capabilities } = await startWebDriverSession(params)
--- a/packages/webdriver/src/index.ts
+++ b/packages/webdriver/src/index.ts
@@ -33,6 +33,8 @@ export default class WebDriver {
             process.env.WDIO_LOG_PATH = path.join(params.outputDir, 'wdio.log')
         }
 
+        log.info('Initiate new session using the WebDriver protocol')
+
         /**
          * if the server responded with direct connect information, update the
          * params to speak directly to the appium host instead of a load
--- a/packages/devtools/tests/__mocks__/puppeteer-core.ts
+++ b/packages/devtools/tests/__mocks__/puppeteer-core.ts
@@ -45,6 +45,7 @@ class PageMock {
     url = jest.fn().mockReturnValue('about:blank')
     emulate = jest.fn()
     setViewport = jest.fn()
+    setDefaultTimeout = jest.fn()
     target = jest.fn().mockReturnValue(target)
 }
 const page = new PageMock()
@@ -58,6 +59,7 @@ class PuppeteerMock {
     waitForTarget = jest.fn().mockImplementation(() => target)
     getActivePage = jest.fn().mockImplementation(() => page)
     pages = jest.fn().mockReturnValue(Promise.resolve([page, page2]))
+    userAgent = jest.fn().mockImplementation(() => 'MOCK USER AGENT')
     _connection = { _transport: { _ws: { addEventListener: jest.fn() } } }
 }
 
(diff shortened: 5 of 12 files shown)


christian-bromann:
> but it still has to remain in the `webdriverio` package as well for standalone mode to work.

I don't think it needs to. I just tried to run this script:

```js
const { remote } = require('webdriverio')

;(async () => {
    const browser = await remote({ capabilities: { browserName: 'chrome' }})

    await browser.url('http://json.org')
    await browser.deleteSession()
})()
```
with WebDriver and DevTools and both would store the log file if I would run it via:

```sh
WDIO_LOG_PATH=/tmp/wdiorepo/wdio.log node index.js
```

I also don't think the other change is needed. If I pass in an empty options object the script fails with:

```js
(node:59762) UnhandledPromiseRejectionWarning: Error: Required option "capabilities" is missing
```

Will close this. I might have miss understood something. Please provide more information when you experience this.

koggdal:
@christian-bromann What I meant with that it has to remain is that the feature was introduced in #3583 and removing it would be a breaking change (something for v7), not something that changes in a patch release. In 6.10.5 I could run without setting any environment variable on my own, but in 6.10.6 this stops working. Yes, I can make it work by adding the environment variable myself, but this PR is a fix for what I would consider a bug (unless introduced as a breaking change in v7).

The change around capabilities was because of the TypeScript types, where `capabilities` is optional, but it’s treated in the code as not optional (tests were failing if I didn’t pass it). Maybe the types are wrong?

christian-bromann:
Ah, now I understand. Thanks for clarifying. We moved this line of code intentionally into the `webdriver` package so that this functionality would be available there too. The problem is that we missed to move the [`outputDir`](https://github.com/webdriverio/webdriverio/blob/master/packages/webdriverio/src/constants.ts#L262-L268) option into the [WebDriver package](https://github.com/webdriverio/webdriverio/blob/master/packages/webdriver/src/constants.ts#L7). Would you mind making this adjustment?

> Maybe the types are wrong?

That is weird because our pipeline is passing.

koggdal:
Ah thanks! :) I think I'm understanding it all a bit better now. I removed my previous commit and pushed up three new commits. Check the commit messages for more details, but in short:

* Moving `outputDir` from `webdriverio` constants to `webdriver` constants, but changing default value to `undefined` as that's what's needed to get the functionality that is documented on the website.
* Add `outputDir` support to `devtools` to align behavior of where logs go, regardless of protocol.
* For `webdriverio`, `webdriver`, and `devtools`: Make sure that logs are not happening before the log path has been set up.

On the topic of `capabilities`: I removed the change for now. For reference though, what I'm referring to is that in `packages/webdriverio/src/types.ts` the `capabilities` property is typed as `capabilities?: WebDriver.DesiredCapabilities | WebDriver.W3CCapabilities`, which means it's optional. If I do `await webdriverio.remote({})` I get no TypeScript error, but I do get a runtime error because of the extra validation that's happening at runtime. My tests don't depend on the capabilities, so I think it would be fine to avoid setting it, but then I get this in my test:

```
    TypeError: Cannot read property 'browserName' of undefined

      533 |      * (this will be ignored if not supported)
      534 |      */
    > 535 |     if (automationProtocol === 'webdriver' && caps.browserName === 'firefox') {
          |                                                    ^
      536 |         if (!caps['moz:firefoxOptions']) {
      537 |             caps['moz:firefoxOptions'] = {}
      538 |         }

      at Object.<anonymous>.exports.updateCapabilities (packages/webdriverio/src/utils/index.ts:535:52)
      at Object.<anonymous>.exports.remote (packages/webdriverio/src/index.ts:50:11)
```

So it seems the code requires `capabilities` to be set, but it's typed as optional. I'm not sure what the right behavior is, but in my opinion, the types should help avoid runtime errors, so if the runtime behavior is correct, maybe `capabilities` shouldn't be marked as optional in the type? If the runtime behavior is not correct (if it would be fine to skip passing capabilities), then changing the code as I did before could be the way (to default `params.capabilities` to `{}`). Anyway, I'm fine with keeping it the way it is for now, but this was my experience on how this tripped me up when writing tests.

packages/devtools/src/index.ts
```diff
@@ -29,6 +30,15 @@ export default class DevTools {
...
         }
 
+        /**
+         * Store all log events in a file
+         */
+        if (params.outputDir) {
+            process.env.WDIO_LOG_PATH = path.join(params.outputDir, 'wdio.log')
+        }
+
+        log.info('Initiate new session using the devtools protocol')
```
christian-bromann, line 40:
```suggestion
        log.info('Initiate new session using the DevTools protocol')
```

packages/webdriver/src/index.ts
```diff
@@ -26,6 +26,15 @@ export default class WebDriver {
...
         }
 
+        /**
+         * Store all log events in a file
+         */
+        if (params.outputDir) {
+            process.env.WDIO_LOG_PATH = path.join(params.outputDir, 'wdio.log')
+        }
+
+        log.info('Initiate new session using the webdriver protocol')
```
christian-bromann, line 36:
```suggestion
        log.info('Initiate new session using the WebDriver protocol')
```

packages/webdriver/src/index.ts
```diff
@@ -26,6 +26,15 @@ export default class WebDriver {
             logger.setLevel('webdriver', params.logLevel!)
         }
 
+        /**
+         * Store all log events in a file
+         */
+        if (params.outputDir) {
```
christian-bromann, line 32:
```suggestion
        if (params.outputDir && !process.env.WDIO_LOG_PATH) {
```

packages/devtools/src/index.ts
```diff
@@ -29,6 +30,15 @@ export default class DevTools {
             logger.setLevel('devtools', params.logLevel)
         }
 
+        /**
+         * Store all log events in a file
+         */
+        if (params.outputDir) {
```
christian-bromann, line 36:
```suggestion
        if (params.outputDir && !process.env.WDIO_LOG_PATH) {
```

christian-bromann (commented):
Some comments, mainly: we should not set the global if it is already set by the user

packages/devtools/src/constants.ts
```diff
@@ -79,6 +79,13 @@ export const DEFAULTS: DefaultOptions<WebDriver.Options> = {
         default: 'info',
         match: /(trace|debug|info|warn|error|silent)/
     },
+    /**
+     * directory for log files
+     */
+    outputDir: {
+        type: 'string',
+        default: undefined
```
christian-bromann, line 87: if the default value is not defined it automatically defaults to undefined

packages/webdriver/src/constants.ts
```diff
@@ -62,6 +62,13 @@ export const DEFAULTS: DefaultOptions<Options> = {
         default: 'info',
         match: /(trace|debug|info|warn|error|silent)/
     },
+    /**
+     * directory for log files
+     */
+    outputDir: {
+        type: 'string',
+        default: undefined
```
christian-bromann, line 70: same

christian-bromann:
> * but changing default value to `undefined` as that's what's needed to get the functionality that is documented on the website.

I would rather update the docs than changing the existing behavior.

> Anyway, I'm fine with keeping it the way it is for now, but this was my experience on how this tripped me up when writing tests.

We still have some inconsistencies between the types we had manually defined and the ones that are now generated by the TS compiler as we rewrite the codebase to TS. This will be cleaned up at a later stage. So we can keep it for now.

koggdal:
@christian-bromann Thanks! Could you clarify what behavior you would like? What the [documentation says](https://webdriver.io/docs/options.html#outputdir) is this:

> Directory to store all testrunner log files (including reporter logs and `wdio` logs). If not set, all logs are streamed to `stdout`. Since most reporters are made to log to `stdout`, it is recommended to only use this option for specific reporters where it makes more sense to push report into a file (like the `junit` reporter, for example).
>
> When running in standalone mode, the only log generated by WebdriverIO will be the `wdio` log.
>
> Type: `String`
> Default: `null`

That behavior is something we're relying on in our project, to set `outputDir` so that we get logs to a file instead of `stdout`. I could imagine other people depending on the reverse behavior if they'd like the logs the `stdout` (when not setting `outputDir`).

What I noticed was that in 6.10.5 and earlier there was this: https://github.com/webdriverio/webdriverio/blob/v6.10.5/packages/webdriverio/src/index.js#L49 That was checking `params.outputDir` instead of `config.outputDir`, where `params` is the direct input to the `remote` function, and `config` is the parsed input with default values. So, in 6.10.5 and earlier, in the case where you didn't pass `outputDir`, the `WDIO_LOG_PATH` env var would not be set, causing logs to go to `stdout`. If you did pass `outputDir`, `params.outputDir` would have that value and `WDIO_LOG_PATH` would get set, and logs would go to file. This is according to the docs.

In e78ecef891b2de4a2f7c198117d35db8debf1c68 (part of 6.10.6) this check was moved into `packages/webdriver/src/index.ts`, where the parsed config is actually put into a `params` variable, so the `params.outputDir` check then works properly. However, there are two issues there still. In https://github.com/webdriverio/webdriverio/blob/v6.10.5/packages/webdriverio/src/index.js#L58 `params` is passed instead of `config`, which means that when `outputDir` isn't passed to `remote`, not even the default `outputDir` ends up in the options for the `webdriver` package. And the second issue there is that the `outputDir` property was missing from `packages/webdriver/src/constants.ts`, so even if the right data was passed down, it wouldn't be parsed correctly here.

So, all in all, from what I can see, the documented behavior is the behavior we've seen in 6.10.5 and below and then it broke with the TypeScript refactor.

I'll fix up the other comments you had! 👍

koggdal:
Fixed up the commits with the changes you commented about. See diff since last round here: https://gist.github.com/koggdal/a45bee03577d37fb584e1f2b18a45ad3

christian-bromann (approved):
This LGTM, thanks so much for taking a stab at it 👍

Issue #5522, opened 2020-06-21
Move docs to Docosaurus v2


christian-bromann:
The tool for generating the docs (Docosaurus) has publish a [v2](https://v2.docusaurus.io/) of their framework. We should transition over to use the new version.

JapneetSingh5:
Can I help out with this? I have no past experience with docusaurus, but reading the guide for migration I think I can handle this, with some guidance perhaps.

christian-bromann:
> Can I help out with this?

Absolutely. I will assign this to you. Let me know if you have any questions along the way.

JapneetSingh5:
how do I start off? v2 changes the directory structure

christian-bromann:
> how do I start off? v2 changes the directory structure

It would be good if we could keep the `docs` directory and the `website` stuff separated. Feel free to do any necessary changes to transition to v2.

christian-bromann:
@JapneetSingh5 any updates? Is there anything I can do to help?

JapneetSingh5:
Sorry for not updating, I couldn't figure out a way to keep the directory structure the same. How do I start? On a forked repository?

christian-bromann:
> How do I start? On a forked repository?

Yes. You fork this repository and eventually make a pull request to this one. Please have a read to out [contributing guidelines](https://github.com/webdriverio/webdriverio/blob/master/CONTRIBUTING.md). If it is not possible to keep the folder structure than feel free to change it.

JapneetSingh5:
Alright, thanks a lot. I will start working on the migration right away.

JapneetSingh5:
Hi, I was working on the migration and found this thread on Github.

https://github.com/facebook/docusaurus/issues/3035

JapneetSingh5:
Should I try out the CLI?

christian-bromann:
> Should I try out the CLI?

yeah, why not?

slorber:
hi,

https://github.com/facebook/docusaurus/pull/3015

the cli is not released yet, but maybe you can get in touch with @anshulrgoyal to see if automated migration works fine on this site

anshulrgoyal:
Hi,
we are currently working on migration cli. I can help u, just tell us how your site is built.

anshulrgoyal:
> Sorry for not updating, I couldn't figure out a way to keep the directory structure the same. How do I start? On a forked repository?

I can help with that. You can change default path of docs and blogs in the new version. 

```
presets: [
    [
[...]

Issue #5364, opened 2020-05-05
Allow support for fibers v4 and v5


christian-bromann:
**Is your feature request related to a problem? Please describe.**
Fibers got updated to v5 to support Node.js v14. Updating it would break support for Node.js v10. Also the fibers maintainer has announced to not further maintain the package which leads to the question how we want to further provide support for synchronous execution.

Any help or feedback is appreciated.

Abel0916:
Encountered this problem when I upgrade Fibers to V5. Then WDIO sync way doesn't work.

pmerwin:
Yes, I think we ran into an issue in AWS Linux 2 lambda, fibers vs 5, and sync mode exits with a code 1 immediately on test execution but only in AWS lambda, could not reproduce locally on OSX.

christian-bromann:
@Abel0916 don't update to Fibers v5. It seems that Fibers in v4 still works with the latest Node v14. Therefor there is no need of action right now.

pathmapper:
> It seems that Fibers in v4 still works with the latest Node v14

@christian-bromann on `linux` and `macos` it seems to work, but on `windows` it's failing: 

https://github.com/maputnik/editor/pull/695

https://github.com/maputnik/editor/pull/695/checks?check_run_id=715807535

christian-bromann:
@pathmapper thanks for the feedback, I will prioritise this then.

pathmapper:
@christian-bromann our failing test was related to `node-sass` (https://github.com/maputnik/editor/pull/695#issuecomment-635209469). Sorry for the noise.

christian-bromann:
ahh ok, nevermind then 😆

hexsprite:
is it recommended to use async then for new projects? or you intend to support sync regardless of fibers support?

christian-bromann:
We are trying to continue supporting sync mode but we can't promise it.

thomaseitler:
Hi,

sorry if this is not related at all - I can create a new issue if necessary. However the error mentions the fibers module.

I just updated to node ```v15.2.1``` and cannot complete the initial config ``` npx wdio config -y ```

Infor from the debug log:

```
...
2016 verbose stack Error: fibers@4.0.3 install: `node build.js || nodejs build.js`
2016 verbose stack spawn ENOENT
2016 verbose stack     at ChildProcess.<anonymous> (/usr/lib/node_modules/npm/node_modules/npm-lifecycle/lib/spawn.js:48:18)
[...]

PR #5901, opened 2021-01-04
WebdriverIO v7 Tracking Doc

christian-bromann:
I propose to have another major version release once we have finished the [TypeScript rewrite effort](https://github.com/webdriverio/webdriverio/projects/14). While we can continue to rewrite every package and release the code compiled by the TS compiler we have to replace our current generated type definitions by the ones we have in our .ts files. We can't guarantee that this transition won't be without regression which is why I propose a major version bump. Here is what v7 should entail:

- a complete rewrite of the code base in typescript
- all type definitions are from package files and are only enhance by a script that reads commands from the protocol json files as well as from the JSDocs
- remove all occurrences of the JSONWireProtocol: given that Selenium also releases v4 that is WebDriver only we should get rid of all docs and implementations of the JSONWireProtocol. The reason for this is that I am seeing a lot of users that still get confused by the old docs, it also reduces the memory as not so many functions are registered to the scope anymore.
- clean ups (unnecessary configs and commands)
- update Fibers and drop support for Node.js v10 (#5364)
- A new refurbished website (#5522 as well as #6212)


christian-bromann:
Merged into new `main` branch.

PR #3407, opened 2021-01-05
Automatically run worker with Babel if setup

christian-bromann:

Currently it is rather difficult to have wdio.conf files in ESNext.

It would be great if the wdio-runner could detect if babel is setup in the project and automatically compile the files.

Have proper documentation of how to do this.

### DIFF
--- a/docs/Babel.md
+++ b/docs/Babel.md
@@ -27,46 +27,4 @@ module.exports = {
 }
 ```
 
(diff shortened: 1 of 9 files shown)


mohanraj-r:
@christian-bromann Does this mean that if the project is already using babel, we don't need to add ts-node as a dependency?

christian-bromann:
You will always need `ts-node` as dependencies but there won't be any configurations necessary anymore to make it work.

PR #6297, opened 2021-01-08
Fixes #6294

### LINKED ISSUE #6294: Inconsistent behavior using chrome args in DevTools
With the WebDriver protocol Chrome flags don't need to be prefixed with `--`. With DevTools this won't work and will result in Chrome opening a tab for every flag used, e.g.

image

Let's make this consistent and ensure that in DevTools all args are prefixed with `--` [when launching the browser](https://github.com/webdriverio/webdriverio/blob/master/packages/devtools/src/launcher.ts#L92).

rkeshwani:
## Proposed changes

This resolves the issue described in  #6294

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments
I was unsure on how to write a test for this.
browser.pages() always seems to return two pages regardless of how many parameters are added to chromeOptions.args.
I was hoping to use the returned length of browser.pages as a determinant but that doesn't seem to work like that even though documentation states that it returns all pages within all contexts.
 

### DIFF
--- a/packages/devtools/src/launcher.ts
+++ b/packages/devtools/src/launcher.ts
@@ -80,10 +80,6 @@ async function launchChrome (capabilities: ExtendedCapabilities) {
     const defaultFlags = Array.isArray(ignoreDefaultArgs) ? DEFAULT_FLAGS.filter(flag => !ignoreDefaultArgs.includes(flag)) : (!ignoreDefaultArgs) ? DEFAULT_FLAGS : []
     const deviceMetrics = mobileEmulation.deviceMetrics || {}
 
-    let chromeOptionsArgs = chromeOptions.args
-    if (typeof chromeOptions.args !== 'undefined') {
-        chromeOptionsArgs = chromeOptions.args.map(x=>x.indexOf('--')==-1?'--'+x:x)
-    }
     const chromeFlags = [
         ...defaultFlags,
         ...[
@@ -94,7 +90,7 @@ async function launchChrome (capabilities: ExtendedCapabilities) {
             '--headless',
             '--no-sandbox'
         ] : []),
-        ...(chromeOptionsArgs || [])
+        ...(chromeOptions.args || []).map((a) => a.startsWith('--') ? a : `--${a}`)
     ]
 
     if (typeof deviceMetrics.pixelRatio === 'number') {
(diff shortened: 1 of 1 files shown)


packages/devtools/src/launcher.ts
```diff
@@ -89,7 +94,7 @@ async function launchChrome (capabilities: ExtendedCapabilities) {
             '--headless',
             '--no-sandbox'
         ] : []),
-        ...(chromeOptions.args || [])
+        ...(chromeOptionsArgs || [])
```
christian-bromann, line 97: Instead of the 3 lines above we can just do:

```suggestion
        ...(chromeOptions.args || []).map((a) => a.startsWith('--') ? a : `--${a}`)
```

christian-bromann (changes requested):
Thanks for taking a stab at this. Would you mind adding a unit test for it?

rkeshwani:
> Thanks for taking a stab at this. Would you mind adding a unit test for it?

As I mentioned above. I tried writing a unit test for this but I cannot recreate the original issue.
Doing something like this returns the same results with and without the changes as above. Based on your original issue I would have expected that the # of pages would have increased but it seems like there are only two that open with or without that code. Adding additional chromeOptions does not increase the # of pages.
I added the following to launcher.test.ts within the 'launch chrome with chrome arguments' test: `expect((await browser.pages()).length).toEqual(2);`

erwinheitzman:
This solution will cause issues when passing `-someflag`, since using Firefox a single dash is a valid option, it can be confusing. Do we want to solve this with this change?

christian-bromann:
Good catch @erwinheitzman , @rkeshwani would you mind addressing this?

rkeshwani:
> This solution will cause issues when passing `-someflag`, Firefox allows the use of a single dash so it can be confusing. Do we want to solve this with this change?

So this specific feature was to only be implemented within the scope of the launchChrome method from my understanding. Is the concern that a user could confuse the arguments or input arguments that only contain a single '-'?

christian-bromann:
> Is the concern that a user could confuse the arguments or input arguments that only contain a single '-'?

@rkeshwani people using Microsoft Edge could run into the same issues given it is the same underlying browser engine (Blink) underneath. So I suggest to make the same change to `launchBrowser` too. Also would you mind signing the CLA and add some tests?

christian-bromann:
any updates @rkeshwani ?

christian-bromann:
Closing due to inactivity.

Issue #6301, opened 2021-01-11
Unable to provide custom sauce-job-name in after hook


AutomationReddy:
[//]: # NOTE: This repository only maintains packages that are listed in the [Readme](https://github.com/webdriverio/webdriverio/blob/master/README.md#packages). Please make sure that your issue is directly caused by one of these packages and if not file an issue in the correct 3rd party package repository.

**Environment (please complete the following information):**
 - **WebdriverIO version:** [6.5]
 - **Mode:** [WDIO Testrunner]
 - **Node.js version:** [>12]
 - **NPM version:** [>6]
 - **Browser name and version:** [Chrome]
 - **Platform name and version:** [Windows 10]
 - **Additional wdio packages used (if applicable):** [@wdio/sauce-service]

**Describe the bug**
Sauce service provides option to update the job name using `sauceOptions: setJobNameInBeforeSuite(boolean)`. However, my requirement is to update job name with a custom name and I am doing it in the after hook: `browser.execute('sauce:job-name=' + customName)`. But when I see the sauce labs results board, the job name is overriding with test suite title

**To Reproduce**
- Open a test in sauce labs
- Update the job name in afterhook or anywhere in before and after using `browser.execute('sauce:job-name=' + customName)`

**Sometimes, it updates the job name if there are any failures/no failures in the test** (Updates the custom name in 1/100th execution)
**Expected behavior**
Job name in SauceLabs should be customName and not the suite title

**Additional context**
After checking the sauce-service library, I am assuming, ```await this._api.updateJob(this._config.user as string, sessionId, body as Job)``` in `this.updateJob()` is updating the jobname back to the test suite name which is calling in after hook of sauce service.

in getBody method --> `body.name = this._suiteTitle`

```
 if (!this._browser.isMultiremote) {
            log.info(`Update job with sessionId ${this._browser.sessionId}, ${status}`)
            return this._isUP ? this.updateUP(failures) : this.updateJob(this._browser.sessionId, failures)
        }
```

christian-bromann:
Any suggestions how this could be handled?

wswebcreation:
@AutomationReddy 

Can you also explain when you see this behaviour? Is this with Sauce Desktop VMs or with Sauce Labs real devices?

AutomationReddy:
[...]

PR #6310, opened 2021-01-11
Change sauce update jobname order

wswebcreation:
## Proposed changes
This PR changes the order of where the jobname is updated, see also this issue https://github.com/webdriverio/webdriverio/issues/6301
It also changes the following:
- remove `setJobNameInBeforeSuite`, because it's now set by default in the before hook for all frameworks
- only update job name in the after hook/reload for retries

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/packages/wdio-sauce-service/README.md
+++ b/packages/wdio-sauce-service/README.md
@@ -6,7 +6,7 @@ WebdriverIO Sauce Service
 > - the Sauce Labs Real Device cloud (iOS and Android) and can **ONLY** update the job status to passed / failed
 
 > - By default the Sauce Service will update the 'name' of the job when the job starts. This will give you the option to update the name at any given point in time.
->The only time when you can't update the name of the job is when you execute a retry or a session reload.
+> - The only time when you can't update the name of the job is when you execute a retry or a session reload.
 
 ## Installation
 
--- a/packages/wdio-sauce-service/README.md
+++ b/packages/wdio-sauce-service/README.md
@@ -5,7 +5,7 @@ WebdriverIO Sauce Service
 > - the Sauce Labs virtual machine cloud (desktop web and em/simulators) and can update the job metadata ('name'*, 'passed', 'tags', 'public', 'build', 'custom-data') and runs Sauce Connect if desired.
 > - the Sauce Labs Real Device cloud (iOS and Android) and can **ONLY** update the job status to passed / failed
 
-> \* By default the Sauce Service will update the 'name' of the job when the job starts. This will give you the option to update the name at any given point in time.
+> - By default the Sauce Service will update the 'name' of the job when the job starts. This will give you the option to update the name at any given point in time.
 >The only time when you can't update the name of the job is when you execute a retry or a session reload.
 
 ## Installation
--- a/packages/wdio-sauce-service/README.md
+++ b/packages/wdio-sauce-service/README.md
@@ -2,8 +2,11 @@ WebdriverIO Sauce Service
 =========================
 
 > WebdriverIO service that provides a better integration into Sauce Labs. This service can be used for:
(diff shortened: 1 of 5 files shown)


packages/wdio-sauce-service/README.md
```diff
@@ -2,9 +2,12 @@ WebdriverIO Sauce Service
 =========================
 
 > WebdriverIO service that provides a better integration into Sauce Labs. This service can be used for:
-> - the Sauce Labs virtual machine cloud (desktop web and em/simulators) and can update the job metadata ('name', 'passed', 'tags', 'public', 'build', 'custom-data') and runs Sauce Connect if desired.
+> - the Sauce Labs virtual machine cloud (desktop web and em/simulators) and can update the job metadata ('name'*, 'passed', 'tags', 'public', 'build', 'custom-data') and runs Sauce Connect if desired.
 > - the Sauce Labs Real Device cloud (iOS and Android) and can **ONLY** update the job status to passed / failed
 
+> \* By default the Sauce Service will update the 'name' of the job when the job starts. This will give you the option to update the name at any given point in time.
```
christian-bromann, line 8:
```suggestion
> - By default the Sauce Service will update the 'name' of the job when the job starts. This will give you the option to update the name at any given point in time.
```

packages/wdio-sauce-service/README.md
```diff
@@ -2,9 +2,12 @@ WebdriverIO Sauce Service
...
+>The only time when you can't update the name of the job is when you execute a retry or a session reload.
```
christian-bromann, line 9:
```suggestion
> - The only time when you can't update the name of the job is when you execute a retry or a session reload.
```

christian-bromann: Should we just have normal paragraphs rather than quotes here?

packages/wdio-sauce-service/src/service.ts
```diff
@@ -86,6 +84,12 @@ export default class SauceService implements WebdriverIO.ServiceInstance {
             this._suiteTitle = test.fullName.slice(0, test.fullName.indexOf(test.description) - 1)
         }
 
+        /* istanbul ignore if */
+        if (this._browser && !this._isUP && !this._isJobNameSet) {
+            this._browser.execute('sauce:job-name=' + this._suiteTitle)
```
christian-bromann, line 89: We can't do this here because the suite title never changes between tests. You probably want to put this in the before hook?

christian-bromann (changes requested):

packages/wdio-sauce-service/src/service.ts
wswebcreation: AFAIK the before hook doesn't have the title for Jasmine, that's why I've put it there. It also is only going to executed once

christian-bromann: ah ok, we should remove `/* istanbul ignore if */` and have a test for this

wswebcreation: Ahh, chips

It needed to be ignore else. There are 2 tests already added

wswebcreation: Changed it to else

packages/wdio-sauce-service/src/service.ts
```diff
@@ -86,6 +84,12 @@ export default class SauceService implements WebdriverIO.ServiceInstance {
             this._suiteTitle = test.fullName.slice(0, test.fullName.indexOf(test.description) - 1)
         }
 
+        /* istanbul ignore else */
```
christian-bromann, line 87:
```suggestion
```

christian-bromann (approved):
👍

PR #6327, opened 2021-01-19
Sauce Service: Push error stack to Sauce

wswebcreation:
## Proposed changes
Sauce Labs provides good insights for test failures, but never knows the context of an error. This PR adds the feature to push the error stack to Sauce Labs so users can also debug the test from the commands tab of Sauce Labs, see the example
image

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/packages/wdio-sauce-service/src/service.ts
+++ b/packages/wdio-sauce-service/src/service.ts
@@ -110,6 +110,17 @@ export default class SauceService implements WebdriverIO.ServiceInstance {
 
     afterTest (test: any, context: any, results: any) {
         /**
+         * If the test failed push the stack to Sauce Labs in separate lines
+         * This should not be done for UP because it's not supported yet and
+         * should be removed when UP supports `sauce:context`
+         */
+        const { error } = results
+        if (error && !this._isUP){
+            const lines = error.stack.split(/\r?\n/)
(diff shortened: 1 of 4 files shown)


packages/wdio-sauce-service/src/service.ts
```diff
@@ -109,6 +109,17 @@ export default class SauceService implements WebdriverIO.ServiceInstance {
...
 
     afterTest (test: any, context: any, results: any) {
+        /**
+         * If the test failed push the stack to Sauce Labs in separate lines
+         * This should not be done for UP because it's not supported yet and
+         * should be removed when UP supports `sauce:context`
+         */
+        const { error } = results
+        if (error && !this._isUP){
+            const lines = error.stack.split(/\r?\n/)
```
christian-bromann, line 119: I wonder if we should limit the amount of lines to like 5 to ensure we don't run crazy many context calls. Also what would happen if we pass in HTML tags? Are they getting interpreted?

```suggestion
            const lines = error.stack.split(/\r?\n/).slice(0, MAX_ERROR_STACK_LENGTH)
```

Let's define `MAX_ERROR_STACK_LENGTH` somewhere at the top of the file.

christian-bromann (approved):
LGTM 👍

PR #6275, opened 2020-12-29
TypeScript auto compiling

christian-bromann:
## Proposed changes

Currently it is not quite straight forward to setup TypeScript with WDIO. You need to modify your config to ensure the TS compiler is triggered. We can help here a bit and automatically trigger it if found.

Also by adding `ts-node` to your dev dependencies it makes it easier to start with a TS setup right away.

## Types of changes

[//]: # (What types of changes does your code introduce to WebdriverIO?)
[//]: # (_Put an `x` in the boxes that apply_)

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## Checklist

[//]: # (_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._)

- [x] I have read the [CONTRIBUTING](https://github.com/webdriverio/webdriverio/blob/master/CONTRIBUTING.md) doc
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)
- [ ] I have added proper type definitions for new commands (if appropriate)

## Further comments

This might have an impact on the load performance but some e2e testing showed that it is marginal and doesn't really makes a difference.

PR #6368, opened 2021-01-30
Add tsNodeOpts to enable configuring auto loaded ts-node

r4j4h:
## Proposed changes

https://github.com/webdriverio/webdriverio/pull/6275 introduced automatic loading of ts-node. I believe that this was accidentally a breaking change for users who need to configure ts-node differently than the automatic loader. The proposed workaround is to remove custom loading code and rely on the automatic loader, which may not work for all users. Since webdriverio worked for them before the change, it should work them after.

I think we need the loadTypeScriptCompiler mechanism to look through the config file for something like a tsNodeOpts block and spread that with its defaults. The defaults can be the current value so that users who do not define it get the current behavior. Users who need to modify it could now modify their wdio.conf.js file as neede, remove their custom loading code, have everything working well.

There are more details described in my response to pull 6275: https://github.com/webdriverio/webdriverio/pull/6275#issuecomment-770051932

My change ended up being a little more expansive, in order to provide the config I had to reorder where loadTypeScriptCompiler is called in ConfigParser. I also updated the types.

## Types of changes

- [x] Documentation update

## Checklist

I am a little hesitant about the documentation part, otherwise I would love to add some. I tried to add it to the examples.

Please let me know if I missed any areas.

## Further comments

- I refactored the non-loading logic from  ConfigParser's addConfigFile into addConfigEntry so that I could spread a modified config with tsConfigOpts in the test. This was not necessary, but would have required me adding another config file to the fixtures solely for this purpose and I did not want to muddy the waters especially in my first commit. Besides I do think this separates the logic a little more cleanly.

- I used ts-node's RegisterOptions type in packages/wdio-config/src/utils.ts but this could just as well be any. I call this out incase it might unknowingly make ts-node's types a dependency. If so let me know and I will set it to any.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -140,7 +140,9 @@ $ npm install
 $ npm start
 ```
 
-This will set up everything needed to run the page on [`localhost:3000`](http://localhost:3000/). You can now modify the content of the [`/docs`](https://github.com/webdriverio/webdriverio/tree/main/docs) files as well as change styles and templates. The page will be automatically updated. If you add documentation in other places, you have to rerun the `npm start` script to re-generate the docs.
+This will set up everything needed to run the page on [`localhost:3000`](http://localhost:3000/). If you need to run on a different host or port, pass them as additional arguments to npm start, like `-- --host 0.0.0.0`.
+
+You can now modify the content of the [`/docs`](https://github.com/webdriverio/webdriverio/tree/main/docs) files as well as change styles and templates. The page will be automatically updated. If you add documentation in other places, you have to rerun the `npm start` script to re-generate the docs.
 
 Every time a new release is pushed to GitHub the WebdriverIO docs are automatically generated and pushed to the project's S3 bucket. The process is defined in a GitHub Actions [pipeline](https://github.com/webdriverio/webdriverio/blob/main/.github/workflows/deploy.yml) and does not need to be done manually.
 
--- a/examples/wdio.conf.js
+++ b/examples/wdio.conf.js
@@ -182,6 +182,21 @@ exports.config = {
         ignoreUndefinedDefinitions: false, // <boolean> Enable this config to treat undefined definitions as warnings.
     },
     //
+    // Auto-compilation configuration
+    autoCompileOpts: {
+        // Enable/disable auto-compilation (enabled by default)
+        autoCompile: true,
+
+        // Configure how ts-node is automatically included when present
+        tsNodeOpts: {
+            transpileOnly: true,
+            project: 'tsconfig.json'
+        },
+
+        // Configure how @babel/register is automatically included when present (and ts-node isn't)
+        babelOpts: {}
+    },
+    //
     // =====
     // Hooks
     // =====
--- a/packages/wdio-cli/src/launcher.ts
+++ b/packages/wdio-cli/src/launcher.ts
@@ -56,6 +56,9 @@ class Launcher {
         private _isWatchMode = false
     ) {
         this.configParser = new ConfigParser()
+        if ( this.configParser.autoCompile ) {
+            this.configParser.autoCompile() // autocompile before parsing configs so we support ES6 features in configs
+        }
         this.configParser.addConfigFile(_configFilePath)
         this.configParser.merge(_args)
 
@@ -374,7 +377,7 @@ class Launcher {
             cid: runnerId,
             command: 'run',
             configFile: this._configFilePath,
-            args: this._args,
+            args: { ...this._args, ...(config?.autoCompileOpts ? { autoCompileOpts: config.autoCompileOpts } : {}) },
             caps,
             specs,
             execArgv,
--- a/packages/wdio-cli/tests/watcher.test.ts
+++ b/packages/wdio-cli/tests/watcher.test.ts
@@ -13,6 +13,9 @@ jest.mock('../src/launcher', () => {
         interface: any
 
         constructor (configFile, args) {
+            if ( this.configParser.autoCompile ) {
+                this.configParser.autoCompile()
+            }
             this.configParser.addConfigFile(configFile)
             this.configParser.merge(args)
             this.isMultiremote = args.isMultiremote
--- a/packages/wdio-config/package.json
+++ b/packages/wdio-config/package.json
@@ -30,5 +30,8 @@
   },
   "publishConfig": {
     "access": "public"
+  },
+  "devDependencies": {
+    "minimatch": "^3.0.4"
   }
 }
--- a/packages/wdio-config/src/constants.ts
+++ b/packages/wdio-config/src/constants.ts
@@ -88,3 +88,11 @@ export const SUPPORTED_HOOKS: (keyof Services.Hooks)[] = [
 export const SUPPORTED_FILE_EXTENSIONS = [
     '.js', '.mjs', '.es6', '.ts', '.feature', '.coffee', '.cjs'
 ]
+
+export const DEFAULT_AUTOCOMPILE_CONFIGS: () => Options.AutoCompileConfig = () => ({
+    autoCompile: true,
+    tsNodeOpts: {
+        transpileOnly: true
+    },
+    babelOpts: {}
+})
--- a/packages/wdio-config/src/lib/ConfigParser.ts
+++ b/packages/wdio-config/src/lib/ConfigParser.ts
@@ -1,16 +1,21 @@
-import fs from 'fs'
-import path from 'path'
-import glob from 'glob'
 import merge from 'deepmerge'
 import logger from '@wdio/logger'
 import type { Capabilities, Options, Services } from '@wdio/types'
 
 import {
     detectBackend, removeLineNumbers, isCucumberFeatureWithLineNumber, validObjectOrArray,
-    loadTypeScriptCompiler, loadBabelCompiler
+    loadAutoCompilers, ModuleRequireService
 } from '../utils'
-import { DEFAULT_CONFIGS, SUPPORTED_HOOKS, SUPPORTED_FILE_EXTENSIONS } from '../constants'
-
+import {
+    SUPPORTED_HOOKS,
+    SUPPORTED_FILE_EXTENSIONS,
+    DEFAULT_AUTOCOMPILE_CONFIGS
+} from '../constants'
+import {
+    DEFAULT_CONFIGS
+} from '../'
+import FileSystemPathService from './FileSystemPathService'
+import RequireLibrary from './RequireLibrary'
 const log = logger('@wdio/config:ConfigParser')
 const MERGE_OPTIONS = { clone: false }
 
@@ -26,9 +31,54 @@ interface MergeConfig extends Omit<Partial<TestrunnerOptionsWithParameters>, 'sp
     exclude?: string | string[]
 }
 
+// Get current working directory
+interface CurrentPathFinder {
+    getcwd(): string
+}
+
+// Require a .js/.json/dotfile config file
+interface LoadConfigFile {
+    loadFile<T>(path: string): T
+}
+
+// Detect if a file is present
+interface IsFileDetector {
+    isFile(path: string): boolean
+}
+
+interface DeterminesAbsolutePath {
+    ensureAbsolutePath(path: string): string
+}
+
+// Glob to find file paths matching a pattern
+interface Globber {
+    glob(pattern: string): string[];
+}
+
+export interface PathService extends CurrentPathFinder, LoadConfigFile, IsFileDetector, Globber, DeterminesAbsolutePath {}
+
 export default class ConfigParser {
+    private _autocompileConfig : Options.AutoCompileConfig = DEFAULT_AUTOCOMPILE_CONFIGS()
     private _config: TestrunnerOptionsWithParameters = DEFAULT_CONFIGS()
     private _capabilities: Capabilities.RemoteCapabilities = [];
+    private _pathService: PathService;
+    private _moduleRequireService: ModuleRequireService;
+
+    constructor(pathService: PathService = new FileSystemPathService(), moduleRequireService:ModuleRequireService = new RequireLibrary()) {
+        this._pathService = pathService
+        this._moduleRequireService = moduleRequireService
+    }
+
+    autoCompile() {
+        /**
+         * on launcher compile files if Babel or TypeScript are installed using our defaults
+         */
+        if (!loadAutoCompilers(this._autocompileConfig, this._moduleRequireService)) {
+            if ( this._autocompileConfig.autoCompile ) {
+                log.debug('No compiler found, continue without compiling files')
+            }
+        }
+    }
 
     /**
      * merges config file with default values
@@ -39,20 +89,27 @@ export default class ConfigParser {
             throw new Error('addConfigFile requires filepath')
         }
 
-        const filePath = path.resolve(process.cwd(), filename)
+        const filePath = this._pathService.ensureAbsolutePath(filename)
 
         try {
+            const config = this._pathService.loadFile<{config: TestrunnerOptionsWithParameters}>(filePath).config
+
+            if (typeof config !== 'object') {
+                throw new Error('addConfigEntry requires config key')
+            }
+
             /**
-             * compile files if Babel or TypeScript are installed
+             * on launcher compile files if Babel or TypeScript are installed using our defaults
              */
-            if (!loadTypeScriptCompiler() && !loadBabelCompiler()) {
-                log.debug('No compiler found, continue without compiling files')
+            // Merge in config over defaults
+            if ( config.autoCompileOpts ) {
+                this._autocompileConfig = merge(this._autocompileConfig, config.autoCompileOpts, MERGE_OPTIONS)
             }
 
             /**
              * clone the original config
              */
-            const fileConfig = merge<Omit<Options.Testrunner, 'capabilities'> & { capabilities?: Capabilities.RemoteCapabilities }>(require(filePath).config, {}, MERGE_OPTIONS)
+            const fileConfig = merge<Omit<Options.Testrunner, 'capabilities'> & { capabilities?: Capabilities.RemoteCapabilities }>(config, {}, MERGE_OPTIONS)
 
             /**
              * merge capabilities
@@ -95,6 +152,11 @@ export default class ConfigParser {
         const spec = Array.isArray(object.spec) ? object.spec : []
         const exclude = Array.isArray(object.exclude) ? object.exclude : []
         this._config = merge(this._config, object, MERGE_OPTIONS) as TestrunnerOptionsWithParameters
+
+        // Merge in config over defaults
+        if ( object.autoCompileOpts ) {
+            this._autocompileConfig = merge(this._autocompileConfig, object.autoCompileOpts, MERGE_OPTIONS)
+        }
 
         /**
          * overwrite config specs that got piped into the wdio command
@@ -172,9 +234,9 @@ export default class ConfigParser {
      * get excluded files from config pattern
      */
     getSpecs (capSpecs?: string[], capExclude?: string[]) {
-        let specs = ConfigParser.getFilePaths(this._config.specs!)
+        let specs = ConfigParser.getFilePaths(this._config.specs!, undefined, this._pathService)
         let spec  = Array.isArray(this._config.spec) ? this._config.spec : []
-        let exclude = ConfigParser.getFilePaths(this._config.exclude!)
+        let exclude = ConfigParser.getFilePaths(this._config.exclude!, undefined, this._pathService)
         let suites = Array.isArray(this._config.suite) ? this._config.suite : []
 
         /**
@@ -188,7 +250,7 @@ export default class ConfigParser {
                     log.warn(`No suite was found with name "${suiteName}"`)
                 }
                 if (Array.isArray(suite)) {
-                    suiteSpecs = suiteSpecs.concat(ConfigParser.getFilePaths(suite))
+                    suiteSpecs = suiteSpecs.concat(ConfigParser.getFilePaths(suite, undefined, this._pathService))
                 }
             }
 
@@ -202,11 +264,11 @@ export default class ConfigParser {
             let tmpSpecs = spec.length > 0 ? [...specs, ...suiteSpecs] : suiteSpecs
 
             if (Array.isArray(capSpecs)) {
-                tmpSpecs = ConfigParser.getFilePaths(capSpecs)
+                tmpSpecs = ConfigParser.getFilePaths(capSpecs, undefined, this._pathService)
             }
 
             if (Array.isArray(capExclude)) {
-                exclude = ConfigParser.getFilePaths(capExclude)
+                exclude = ConfigParser.getFilePaths(capExclude, undefined, this._pathService)
             }
 
             specs = [...new Set(tmpSpecs)]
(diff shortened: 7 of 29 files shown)


packages/wdio-config/src/lib/ConfigParser.ts
```diff
@@ -77,12 +87,19 @@ export default class ConfigParser {
              */
             this._config = merge(detectBackend(this._config), this._config, MERGE_OPTIONS)
 
+            /**
+             * compile files if Babel or TypeScript are installed
+             */
+            if (!loadTypeScriptCompiler(this._config.tsNodeOpts) && !loadBabelCompiler()) {
```
christian-bromann, line 93: This should happen before we call `addConfigEntry` otherwise any imports or TS code in the config can't be compiled, no?

packages/wdio-types/src/index.ts
```diff
@@ -21,6 +21,7 @@ declare global {
         interface MochaOpts { [key: string]: any }
         interface JasmineOpts { [key: string]: any }
         interface CucumberOpts { [key: string]: any }
+        interface TsNodeOpts { [key: string]: any }
```
christian-bromann, line 24: We can move this into the options.ts file and import it in the files where we need it directly. These options are only defined here so we can overwrite them in case someone has `@wdio/cucumber-framework` etc installed where we have access to these types.
```suggestion
```

packages/wdio-types/src/Options.ts
```diff
@@ -403,6 +403,7 @@ export interface Testrunner extends Hooks, Omit<WebdriverIO, 'capabilities'>, We
     mochaOpts?: WebdriverIO.MochaOpts
     jasmineOpts?: WebdriverIO.JasmineOpts
     cucumberOpts?: WebdriverIO.CucumberOpts
+    tsNodeOpts?: WebdriverIO.TsNodeOpts
```
christian-bromann, line 406:
```suggestion
    tsNodeOpts?: { [key: string]: any }
```

packages/wdio-config/src/utils.ts
```diff
@@ -220,10 +221,10 @@ export function validateConfig<T>(defaults: Options.Definition<T>, options: T, k
     return params
 }
 
-export function loadTypeScriptCompiler () {
+export function loadTypeScriptCompiler (tsNodeOpts?: WebdriverIO.TsNodeOpts) {
```
christian-bromann, line 224:
```suggestion
export function loadTypeScriptCompiler (tsNodeOpts?: RegisterOptions) {
```

christian-bromann (commented):
Great job! I was already thinking about this the other day. This is very needed.

packages/wdio-config/src/lib/ConfigParser.ts
r4j4h: Great question! I moved it here so that we could have the options parsed to hand to ts-node. Is there a better spot off-hand we could put it, or do we need to minimally open the config to get the options and then re-parse after?

christian-bromann: Ah yes, we can't read the `tsNodeOptions` before we actually have compiled it for which we need the `tsNodeOptions`. A typical chicken&egg problem 😄 It seems to be that the compiler options can be passed in as environment variables. I would suggest to keep things as they are and instead add some documentation as to how to modify these options.

We have received a lot of comments about how user can use imports in their config file and this solution should make this very easy. So I would love to be able to use TypeScript in the config without having to set up a special way to call wdio.

packages/wdio-types/src/index.ts
r4j4h: Great suggestion, that is good to know 👍

packages/wdio-config/src/utils.ts
r4j4h: Great call, was also able to get rid of that "as any" 💃

packages/wdio-config/src/lib/ConfigParser.ts
r4j4h: Oooh, using environment variables is a great idea! Confirmed it works on our end too. 🎉 

OK I want to make sure I understand the new direction, are we saying:

1) Remove the new tsNodeOpts config stuff and add docs instructing to use environment variables for any specific config? https://github.com/TypeStrong/ts-node#cli-and-programmatic-options seems to give one for each so this should be sufficient.

Or are we saying:

2) Keep the new tsNodeOpts config stuff and also add docs letting people know about both options.

BorisOsipov: Environment variables can be pain in daily tests development routines. I don't like the idea when I need set env variable before I write\debug new test when I run `yarn run wdio --spec` every day.
It could sound crazy but I would like to have another config file e.g. `.wdiocompile` to control wdio autocompile behavior

CONTRIBUTING.md
```diff
@@ -140,7 +140,9 @@ $ npm install
 $ npm start
 ```
 
-This will set up everything needed to run the page on [`localhost:3000`](http://localhost:3000/). You can now modify the content of the [`/docs`](https://github.com/webdriverio/webdriverio/tree/main/docs) files as well as change styles and templates. The page will be automatically updated. If you add documentation in other places, you have to rerun the `npm start` script to re-generate the docs.
+This will set up everything needed to run the page on [`localhost:3000`](http://localhost:3000/). If you need to run on a different host or port, pass them as additional arguments to npm start, like `-- --host 0.0.0.0`.
```
r4j4h, line 143: I added a line about changing host or port. I can remove if you don't like it. I run as much of my development environment in containers and so the default localhost didn't work and I needed to dig into docusaurus to learn how to configure it and figured this might save other people that same time

packages/wdio-config/src/lib/ConfigParser.ts
r4j4h: Thanks for joining in! :)

.wdiocompile might be a good idea to also solve the chicken-and-egg config parsing problem too.

Another option would be to nested tsNodeOpts under autoCompileOpts, but I like how a dedicated config file would solve the parsing issue of the rest of the config, just wanted to mention it as an option though

christian-bromann: I think we should advocate both options. The important piece is that people can write TS in their config file and have the ability to add type safety to it. So maybe the tsconfig options are not really necessary for just compiling the config but when running the tests. To give more context: the config is read 2x:

1. in the launcher (main process) to read services and run the onPrepare and onComplete hooks
2. within the worker process to establish the framework, reporters and everything else

@r4j4h if you could find out whether you need the ts-node options also for just compiling the config, that would be great. Otherwise we have to be clear in the docs that only environment variables can be used if you have the need to set these configs.

r4j4h: @christian-bromann Great distinction. I would hazard a guess that it is only when running the tests, as it is after chromedriver is initialized:

```
2021-02-01T18:52:17.624Z DEBUG @wdio/config:utils: Found 'ts-node' package, auto-compiling TypeScript files

Execution of 1 spec files started at 2021-02-01T18:52:17.831Z

Starting ChromeDriver 87.0.4280.88 (89e2380a3e36c3464b5dd1302349b1382549290d-refs/branch-heads/4280@{#1761}) on port 9515
Only local connections are allowed.
Please see https://chromedriver.chromium.org/security-considerations for suggestions on keeping ChromeDriver safe.
ChromeDriver was started successfully.
[1612205537.997][SEVERE]: bind() failed: Cannot assign requested address (99)
[0-0] RUNNING in chrome - /features/testing.feature
```

I am making pretty good progress toward .wdiocompile as well, for what that is worth

christian-bromann: Great! So we can keep the `tsNodeOpts` in our WebdriverIO Testrunner config and add to the docs that if people are interested adding specific compile options for the config they can through environment variables.

christian-bromann (commented):
So I think two parts are missing here:

- move the transpiler call (`!loadTypeScriptCompiler(this._config.tsNodeOpts) && !loadBabelCompiler()`) above the line where we initially require the config
- add some documentation how to apply transpiler option through environment variables

Let me know if I can help with any of these.

r4j4h:
Thanks, that seems quite simple. While adding the .wdiocompile file support, I've been adding to the test suite and increasing the testability of the ConfigParser, but I can probably split those out into separate MRs to get this one in quicker. Do you have a preference?

r4j4h:
Also, by

> move the transpiler call (!loadTypeScriptCompiler(this._config.tsNodeOpts) && !loadBabelCompiler()) above the line where we initially require the config

Do you mean before we require the config file, or immediately after requiring it? I'm just a bit confused if this doesn't still break that situation that someone is using typescript in their config file? If this is fine to not customize ts-node here b/c it won't affect when it runs the tests, don't we still need to have it customizable somewhere before it runs the tests with it?

christian-bromann:
> then parse them into as args into the worker

It might be necessary here to parse the ts-node config into the worker process as args and then be read by the `ConfigParser` before actually requiring the config file again.

christian-bromann:
Yeah, the idea is to read the tsnode parameters when we read the config file in the testrunner process. Then pass these information along when the worker is being created so that we can apply them when we read the config in the work process. The config file is unlikely to have complex code in it and should be able to be compiled without a `tsconfig.json` so initially the testrunner reads these values and applies them in the worker.

r4j4h:
OK, do let me know if there was already a way to distinguish between the launcher/runner context, but this works locally! Still need to update the docs

r4j4h:
Have everything working with tests and even the .wdiocompile although we can remove that if you don't like it. Haven't pushed those parts because they refactor and would add noise.

Running into an error in the runner tests that I'm not sure of the direct cause of. At one point I had the enum in types, but seemed to be able to move it into config, but perhaps this is why I originally moved it to types and forgot?

```
    TypeError: Cannot read property 'Runner' of undefined

      62 | 
      63 | export default class Runner extends EventEmitter {
    > 64 |     private _configParser = new ConfigParser(ConfigEnvironmentContext.Runner)
```

What's odd is it is JUST the tests. I can build them fine and they work in the project fine. If you have any ideas it would be a big help!

r4j4h:
Noticed the pipeline gave a much more helpful error:

```
Error: packages/wdio-config/src/lib/ConfigParser.ts(67,59): error TS2339: Property 'tsNodeOpts' does not exist on type 'Pick<Testrunner, "onPrepare" | "onWorkerStart" | "onComplete" | "onReload" | "before" | "beforeCommand" | "beforeHook" | "beforeSession" | "beforeSuite" | "beforeTest" | ... 56 more ... | "autoCompileOpts">'.
```

I think I missed one of the changes with the types so I went on and pushed everything up, sorry if it's a lot to look at

r4j4h:
Hmm apparently it didn't give that error in the pipelines. The linter got mad at a test line I added though 😊

r4j4h:
Awesome, well I'm exhausted for the night but it looks like it got way further, now the cli tests are yelling about the same thing. Will pick back up tomorrow

r4j4h:
Nope, I don't think it got further, it's just doing the same thing as on my stack

packages/wdio-config/src/constants.ts
```diff
@@ -88,3 +88,18 @@ export const SUPPORTED_HOOKS: (keyof Services.Hooks)[] = [
...
     '.js', '.mjs', '.es6', '.ts', '.feature', '.coffee', '.cjs'
 ]
+
+export enum ConfigEnvironmentContext {
+    Unknown = 'unknown',
+    Launcher = 'launcher',
+    Runner = 'runner'
+}
+
+export const AUTOCOMPILE_CONFIG_FILENAME = '.wdiocompile'
```
christian-bromann, line 98: I don't think such a file is needed. If we can compile the config in the runner process and only need the compiler options in the worker we can just pass along the options from the runner process into the worker. Here is how this would go:

- the CLI launcher reads the config initially (https://github.com/webdriverio/webdriverio/blob/main/packages/wdio-cli/src/launcher.ts#L58-L60), we can allow users to pass in some CLI args as you have implemented it already to allow modify compile options for that part. My assumption tho is that we don't need them normally there so compiling the config file should be fine
- we can now access the compile options defined in the config and pass them into the worker as [args](https://github.com/webdriverio/webdriverio/blob/main/packages/wdio-cli/src/launcher.ts#L377)
- these args are then available in the [runner](https://github.com/webdriverio/webdriverio/blob/main/packages/wdio-runner/src/index.ts#L86) and we can pass the compile options along when we [read the config file](https://github.com/webdriverio/webdriverio/blob/main/packages/wdio-runner/src/index.ts#L95)

That should make this a bit easer, what do you think?

christian-bromann: This would also make the `ConfigEnvironmentContext` obsolete.

packages/wdio-types/src/Options.ts
```diff
@@ -403,6 +403,15 @@ export interface Testrunner extends Hooks, Omit<WebdriverIO, 'capabilities'>, We
     mochaOpts?: WebdriverIO.MochaOpts
     jasmineOpts?: WebdriverIO.JasmineOpts
     cucumberOpts?: WebdriverIO.CucumberOpts
+    /**
+     * autocompile options
+     */
+    autoCompileOpts?: AutoCompileConfig
+}
+
+export interface AutoCompileConfig {
```
christian-bromann, line 412: I like this idea of putting Babel and TS options into one property. It can be also nicely extended for other purposes in the future 👍

christian-bromann (commented):
Great progress 👍  Added some comments with optimisation ideas. Let me know if I can help with this. Feel free to book a [Open Office Hour](https://webdriver.io/blog/2020/07/01/office-hours.html) slot if you want to sync.

packages/wdio-config/package.json
```diff
@@ -30,5 +30,8 @@
   },
   "publishConfig": {
     "access": "public"
+  },
+  "devDependencies": {
+    "minimatch": "^3.0.4"
```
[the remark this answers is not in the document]
r4j4h: I used it in configparser.test.ts to simulate the glob behavior

packages/wdio-config/src/constants.ts
r4j4h: Good call. Are you sure we need to pass along through the args since it re-loads the file anyway? I have it currently doing that and it seems to work fine, but I'm totally open to what you are asking just curious if we even need to at this point :)

r4j4h:
I thought my changes to the tests would make things faster and easier to maintain but apparently I've caused a memory leak or something and the smoke tests are running out of memory :(

packages/wdio-types/src/Options.ts
r4j4h: I'm glad you like this idea 🎉

website/docs/AutoCompile.md
```diff
@@ -0,0 +1,36 @@
+---
+id: autocompile
+title: Auto-compiling
```
christian-bromann, line 3: I wonder if it makes more sense to just mention these auto compile options in the Babel/TypeScript pages rather than having a dedicated page for it.

packages/wdio-config/src/lib/ConfigParser.ts
```diff
@@ -26,9 +32,75 @@ interface MergeConfig extends Omit<Partial<TestrunnerOptionsWithParameters>, 'sp
...
+
+class RequireLibrary implements ModuleRequireService {
+    require<T>(module: string): T {
+        return require(module) as T
+    }
+
+    resolve(request: string, options: { paths?: string[] }): string {
+        return require.resolve(request, options)
+    }
+}
```
christian-bromann, line 91: What is the purpose of those? If it is for testing purposes we could just use the mock features of jest, no? It seems they create a lot of overhead, especially when calling:

```js
exclude = ConfigParser.getFilePaths(capExclude, undefined, this._pathService)
```

christian-bromann (commented):
mhm .. interesting error. Are the smoke test working for you locally?

r4j4h:
Yeah, locally it worked fine for me with the launcher settings being passed because that prevented ts-node from being loaded twice, but because I didn't quite follow the args approach and loaded ts-node twice in practice it works but the smoke tests blow up. The only thing messed up with the launcher settings is the TypeScript enum doesn't seem to carry over in the tests, even though it as well also works in practice.

I will look more into using the args as you described instead of my launcher settings as that should also work, even if it's just moving the launcher setting to the arg..

packages/wdio-config/src/lib/ConfigParser.ts
r4j4h: Yes they are for treating the file system as repositories. I desired testing different config files without needing to have more in the fixtures directory when all that changes is one property in them. In an earlier test I read the current fixture file into memory and modified it afterward, but it was feeling hacky.

As far as overhead I am not quite sure how that would be, it is passing a reference around to an object yes, but that object is running literally the same code as before. 

I think the mock features of jest would also work. I wanted to avoid coupling further with the implementation and mocking the process objects and felt capturing the abstract logic was cleaner. Both should absolutely work though, the big difference is where the hook for the tests happens: in application code as an interface instead of calling require, or in Node.js require infrastructure code providing mocked require responses for the code under test.

r4j4h:
Glad we have the smoke tests. https://github.com/TypeStrong/ts-node/issues/409 indicates double loading ts-node is fraught with problems. That thread also makes me want to add a global killswitch in autoCompileOpts that just opts out of all of it for users who want that. What do you think?

packages/wdio-config/src/constants.ts
r4j4h: Running into a problem with the config lifecycle where the current autocompile happens during `addConfigFile` calls and the args come in through `merge` calls which currently always happen sequentially after addConfigFile calls. These only seem to be separate because the runner shuts down if addConfigFile throws.

If we could combine them then we could check to see if the args has a autoCompileOpts and use that in the same way `ConfigEnvironmentContext` was being used to prevent double auto-compiler loads.

The reason `ConfigEnvironmentContext` worked before was it allowed addConfigFile to run ts-node ASAP in the launcher and to not run at all in the runner.

Merging them would allow us to have only one autoCompile check and call, where with `ConfigEnvironmentContext` before I had two of them, but it changes the API surface around that try/catch a little.

TL;DR; addConfigFile wants to run ts-node ASAP, so without altering the code there is no way to prevent a double-load in the runner via the args because addConfigFile will have already loaded ts-node with defaults by the time the merge code could load it using the args. :(

r4j4h: Not merging them is also an option, but would require adding a flag if not to ConfigParser to the function calls, effectively bringing back `ConfigEnvironmentContext` just maybe in another name :/

website/docs/AutoCompile.md
r4j4h: Since it's dedicated behavior I felt a sub-page under Test Runner section was a good fit. Especially in light of https://github.com/TypeStrong/ts-node/issues/409, but seeing as WebdriverIO only supports Babel and TypeScript and there are pages for them already I see it makes sense to consolidate the information in both of those pages and remove this one which just effectively will point to either of them.

packages/wdio-config/src/constants.ts
r4j4h: Ah, I am moving towards a solution of exposing a third function on ConfigParser to trigger auto-compilation, and triggering it before feeding files or merging in the launcher, and triggering it after those steps in the runner. Seems to do the trick!

r4j4h:
Similarly to when I tried to add the enum for context, the other projects' tests do not see the autoCompile function I added. :(

```
this.configParser.autoCompile is not a function
```

r4j4h:
Wasn't able to solve the type problem, but I was able to debug it further. In each case it was getting a Mocked ConfigParser that did not have those functions on it. But when I checked the node_modules folders manually it was there. Somehow jest or something got confused or there must be a type copy step I missed, but since this change is orthogonal to all of those adding duck-typing worked to get the tests happy :)

packages/wdio-cli/src/launcher.ts
```diff
@@ -56,6 +56,9 @@ class Launcher {
         private _isWatchMode = false
     ) {
         this.configParser = new ConfigParser()
+        if ( this.configParser && this.configParser.autoCompile ) {
```
christian-bromann, line 59: Would that ever be undefined?
```suggestion
        if (this.configParser.autoCompile) {
```

packages/wdio-cli/tests/watcher.test.ts
```diff
@@ -13,6 +13,9 @@ jest.mock('../src/launcher', () => {
         interface: any
 
         constructor (configFile, args) {
+            if ( this.configParser && this.configParser.autoCompile ) {
```
christian-bromann, line 16: same
```suggestion
            if (this.configParser.autoCompile) {
```

packages/wdio-runner/src/index.ts
```diff
@@ -101,6 +101,10 @@ export default class Runner extends EventEmitter {
          * merge cli arguments into config
          */
         this._configParser.merge(args)
+        // autocompile after parsing configs so we support ES6 features in tests with config driven by users
+        if ( this._configParser && this._configParser.autoCompile ) {
```
christian-bromann, line 105: same
```suggestion
        if (this._configParser.autoCompile) {
```

packages/wdio-config/package.json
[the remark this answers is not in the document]
christian-bromann: Are these tests pushed? I don't see them 🤔

christian-bromann (commented):
Some last comments

examples/wdio.conf.js
```diff
@@ -182,6 +182,15 @@ exports.config = {
         ignoreUndefinedDefinitions: false, // <boolean> Enable this config to treat undefined definitions as warnings.
     },
     //
+    // Auto-compilation control hook
+    // See TODO link to where the docs page will end up
```
r4j4h, line 186: Ooopps, I missed this

packages/wdio-config/src/lib/ConfigParser.ts
```diff
@@ -39,20 +122,27 @@ export default class ConfigParser {
             throw new Error('addConfigFile requires filepath')
         }
 
-        const filePath = path.resolve(process.cwd(), filename)
+        const filePath = path.isAbsolute(filename) ? path.normalize(filename) : path.resolve(this._pathService.getcwd(), filename)
```
r4j4h, line 125: I actually meant to leave this out, as this is a change, but I noticed it was used elsewhere and seemed like it might be safer. Wanted to highlight it to make sure you saw :)

r4j4h:
I think this is ready to go. It's a lot of commits though, would you like me to rebase it?

christian-bromann (approved):
This is excellent. I haven't seen such an high quality first time contribution for a long time. Thank you so much for all your effort and time 👍 

Do you have anything left to do here or can I go ahead and merge it?

packages/wdio-config/src/lib/ConfigParser.ts
christian-bromann: 👍

r4j4h:
Thank you for working with me to get it such a nice state. It was my pleasure!

Nope I think we're ready to merge if you are!

r4j4h:
Hah, I am so weird now that I type that I want to rebase it lol, give me a few more moments please it shouldn't take long

r4j4h:
OK Now we should be ready to merge! :) sorry about that

christian-bromann:
Thank you @r4j4h .. hope to see more of these excellent PRs in the future 😊

PR #6392, opened 2021-02-06
Add sauce sharable report links

wswebcreation:
## Proposed changes

By default Sauce Labs is providing a report link which can only be viewed by a team member. This PR enables a sharable link so the reports can be seen by all teams.
This PR also fixes the multiRemote UT's that were not validating proper outcomes

## Types of changes

- [x] Documentation update

## Checklist

## Further comments
I had some challenges with TypeScript. Example

According to the types `user`, `key`,  `sessionId` can be option on the `config`
I use them all in the `getTestLink`
Then I get this
image

I _fixed_ it by casting the props.

### DIFF
--- a/packages/wdio-spec-reporter/src/index.ts
+++ b/packages/wdio-spec-reporter/src/index.ts
@@ -27,7 +27,7 @@ export default class SpecReporter extends WDIOReporter {
         failed: '✖'
     }
 
-    private _sauceLabsSharableLinks:boolean = true
+    private _sauceLabsSharableLinks = true
 
     constructor (options: SpecReporterOptions) {
         /**
--- a/packages/wdio-spec-reporter/src/index.ts
+++ b/packages/wdio-spec-reporter/src/index.ts
@@ -2,7 +2,7 @@ import WDIOReporter, { SuiteStats, HookStats, RunnerStats, TestStats, Argument }
 import { Capabilities } from '@wdio/types'
 import chalk, { Chalk } from 'chalk'
 import prettyMs from 'pretty-ms'
-import { buildTableData, printTable, getFormattedRows } from './utils'
+import { buildTableData, printTable, getFormattedRows, sauceAuthenticationToken } from './utils'
 import type { StateCount, Symbols, SpecReporterOptions, TestLink } from './types'
 
 const DEFAULT_INDENT = '   '
@@ -27,12 +27,17 @@ export default class SpecReporter extends WDIOReporter {
         failed: '✖'
     }
 
+    private _sauceLabsSharableLinks:boolean = true
+
     constructor (options: SpecReporterOptions) {
         /**
          * make spec reporter to write to output stream by default
          */
         super(Object.assign({ stdout: true }, options))
         this._symbols = { ...this._symbols, ...this.options.symbols || {} }
+        this._sauceLabsSharableLinks = <boolean>('sauceLabsSharableLinks' in options
+            ? options.sauceLabsSharableLinks
+            : this._sauceLabsSharableLinks)
     }
 
     onSuiteStart (suite: SuiteStats) {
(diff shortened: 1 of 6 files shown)


packages/wdio-spec-reporter/src/index.ts
```diff
@@ -27,12 +27,17 @@ export default class SpecReporter extends WDIOReporter {
         failed: '✖'
     }
 
+    private _sauceLabsSharableLinks:boolean = true
```
christian-bromann, line 30: It will be automatically a boolean if you have a default value like this
```suggestion
    private _sauceLabsSharableLinks = true
```

packages/wdio-spec-reporter/src/index.ts
```diff
@@ -27,12 +27,17 @@ export default class SpecReporter extends WDIOReporter {
...
+
     constructor (options: SpecReporterOptions) {
         /**
          * make spec reporter to write to output stream by default
          */
         super(Object.assign({ stdout: true }, options))
         this._symbols = { ...this._symbols, ...this.options.symbols || {} }
+        this._sauceLabsSharableLinks = <boolean>('sauceLabsSharableLinks' in options
+            ? options.sauceLabsSharableLinks
+            : this._sauceLabsSharableLinks)
```
christian-bromann, line 40: I don't think casting is needed here, both should be booleans
```suggestion
        this._sauceLabsSharableLinks = 'sauceLabsSharableLinks' in options
            ? options.sauceLabsSharableLinks
            : this._sauceLabsSharableLinks
```

packages/wdio-spec-reporter/src/index.ts
```diff
@@ -140,7 +145,10 @@ export default class SpecReporter extends WDIOReporter {
                 ? '.us-east-1'
                 : ['eu', 'eu-central-1'].includes(config.region || '') ? '.eu-central-1' : ''
             const multiremoteNote = isMultiremote ? ` ${instanceName}` : ''
-            return [`Check out${multiremoteNote} job at https://app${dc}.saucelabs.com/tests/${sessionId}`]
+            const sauceLabsSharableLinks = this._sauceLabsSharableLinks
+                ? sauceAuthenticationToken( <string>config.user, <string>config.key, <string>sessionId )
```
christian-bromann, line 149: in order to avoid these castings I would suggest to do this:

```js
if (isSauceJob && config.user && config.key && sessionId) {
```

Because they could technically be undefined.

packages/wdio-spec-reporter/src/types.ts
```diff
@@ -14,6 +14,7 @@ export interface Symbols {
 }
 
 export interface SpecReporterOptions {
+    sauceLabsSharableLinks?: boolean
```
christian-bromann, line 17: can we add some docs in here, similar to e.g. https://github.com/webdriverio/webdriverio/blob/main/packages/wdio-junit-reporter/src/types.ts

christian-bromann (changes requested):
some comments

packages/wdio-spec-reporter/src/index.ts
wswebcreation: 👍

wswebcreation: I needed to cast them otherwise I need to accept `boolean` and `undefined` , what do you want

christian-bromann: How about this then:

```suggestion
        this._sauceLabsSharableLinks = 'sauceLabsSharableLinks' in options
            ? options.sauceLabsSharableLinks as boolean
            : this._sauceLabsSharableLinks
```

wswebcreation: Yes, this will work to, will fix it, thanks for the suggestion

packages/wdio-spec-reporter/src/index.ts
wswebcreation: Yep, that worked, will update it

packages/wdio-spec-reporter/src/types.ts
wswebcreation: Done

christian-bromann (approved):
👍 thank you very much!

wswebcreation:
You're welcome, also thanks for your support

Issue #6384, opened 2021-02-03
Detecting the backend is happening without reading the Cli arguments


wswebcreation:
**Environment (please complete the following information):**
 - **WebdriverIO version:** 6
 - **Mode:** Testrunner
 - **If WDIO Testrunner, running sync/async:** both
 - **Node.js version:** all
 - **NPM version:** all
 - **Browser name and version:** all
 - **Platform name and version:** all
 - **Additional wdio packages used (if applicable):** 

**Config of WebdriverIO**
```js
exports.config = {
  // ====================
  // Runner Configuration
  // ====================
  runner: 'local',
  // ==================
  // Specify Test Files
  // ==================
  specs: [
    './test/e2e/specs/**/*.js'
  ],
  // ============
  // Capabilities
  // ============
  maxInstances: 20,
  capabilities: [
    {
        browserName: 'chrome',
        platformName: 'Windows 10',
        browserVersion: 'latest',
        'sauce:options': {
        },
    },
  ],
  // ===================
  // Test Configurations
  // ===================
  logLevel: 'silent',
  bail: 0,
  baseUrl: 'http://localhost:3000',
  waitforTimeout: 10000,
  connectionRetryTimeout: 90000,
  connectionRetryCount: 3,
  framework: 'jasmine',
  reporters: ['spec'],
  jasmineNodeOpts: {
    defaultTimeoutInterval: 60000,
    helpers: [require.resolve('@babel/register')],
  },
  services: ['sauce'],
};
```

**Describe the bug**
When you want to provide your `user` and `key` through the command line WDIO now defaults back to the default backend which is `http://localhost:4444/`. It looks like the order of reading the CLI args and detecting the backend is not the correct one

**To Reproduce**
Use the above config and this command 

`yarn test.e2e.sauce.eu --spec=./test/e2e/specs/checkout.complete.spec.js --user=username --key###`

**Expected behavior**
The `detectBackend` method will be used with the CLI args and will start a session in Sauce Labs

christian-bromann:
Let's continue convo in #6408

PR #6408, opened 2021-02-11
Refactor `detectBackend` and move it to the `webdriverio` package

christian-bromann:
## Proposed changes

We are using our `detectBackend` helper multiple times in the process of creating a session with the test runner. There is a big room for improvement here by moving this utility into the `webdriverio` package and only call it once before creating the session. This works now compared as to the past because also `user` and `key` options have been moved to the webdriverio package.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

fixes: #6384


wswebcreation (commented):
Hi @christian-bromann 

It always looks that easy, but a lot of files are changed, thanks for doing this

I have 2 small remarks:
- There are a lot of errors, form GitHub, regarding the typings. I don't know if that's part of this cope, but should they also be fixed?
- are the UTs giving a good sense of what might break or should this change also be tested through _E2E_?

christian-bromann:
> * but should they also be fixed?

Yes, builds often fail because TypeScript is behind and some problems are only unveiled when building the project from a clean pull. Working on fixing these.

> * are the UTs giving a good sense of what might break or should this change also be tested through _E2E_?

I think UT are sufficient here or better said: e2e tests would be too expensive that it wouldn't justify the value add.

> * Will this change also be migrated to V6?

Adding a `backport-requested` label to note that this feature should be backported. Not sure when I get to it though. Feel free to take a stab at that.

wswebcreation (approved):
LGTM then, merge and ship it 😉

PR #6422, opened 2021-02-13
Update Cucumber import

wswebcreation:
## Proposed changes
The docs are not clear that the import changed, this makes it more clear

## Types of changes

- [x] Documentation update

## Checklist


christian-bromann:
We have clearly stated it in our release blog post. I am happy to add it in the changelog as well but I don't think keeping it as part of the docs is a good idea. At some point the docs will be full with such references. I rather would have the changelog file documenting this.

wswebcreation:
I understand, but people don't read the blog post / changelogs, they just upgrade. This might prevent questions, but if you want to close it feel free to do so 😄

christian-bromann:
People also don't read docs 😛 let's see how many inquiries we get, if it becomes unbearable it would definitely makes sense to address that.

Issue #6438, opened 2021-02-15
wdio-allure-reporter: `TypeError: Cannot read property 'includes' of undefined` and `TypeError: Cannot read property 'split' of undefined` errors


rickschubert:
Please find below the full error stack trace for `Cannot read property 'includes' of undefined`: 

```
2021-02-15T11:53:41.988Z ERROR @wdio/runner: TypeError: Cannot read property 'includes' of undefined
    at /Users/rickschubert/Documents/github/app/e2e/cucumber-browser-tests/node_modules/@wdio/allure-reporter/build/utils.js:48:85
    at Array.some (<anonymous>)
    at Object.exports.isMochaEachHooks (/Users/rickschubert/Documents/github/app/e2e/cucumber-browser-tests/node_modules/@wdio/allure-reporter/build/utils.js:48:66)
    at AllureReporter.onHookStart (/Users/rickschubert/Documents/github/app/e2e/cucumber-browser-tests/node_modules/@wdio/allure-reporter/build/index.js:237:56)
    at AllureReporter.<anonymous> (/Users/rickschubert/Documents/github/app/e2e/cucumber-browser-tests/node_modules/@wdio/allure-reporter/node_modules/@wdio/reporter/build/index.js:72:18)
    at AllureReporter.emit (events.js:311:20)
    at AllureReporter.EventEmitter.emit (domain.js:482:12)
    at /Users/rickschubert/Documents/github/app/e2e/cucumber-browser-tests/node_modules/@wdio/runner/build/reporter.js:36:56
    at Array.forEach (<anonymous>)
    at BaseReporter.emit (/Users/rickschubert/Documents/github/app/e2e/cucumber-browser-tests/node_modules/@wdio/runner/build/reporter.js:36:25)
    at CucumberReporter.emit (/Users/rickschubert/Documents/github/app/e2e/cucumber-browser-tests/node_modules/@wdio/cucumber-framework/build/reporter.js:194:24)
    at CucumberReporter.handleBeforeStep (/Users/rickschubert/Documents/github/app/e2e/cucumber-browser-tests/node_modules/@wdio/cucumber-framework/build/reporter.js:57:14)
    at CucumberEventListener.emit (events.js:311:20)
    at CucumberEventListener.EventEmitter.emit (domain.js:482:12)
[...]

PR #6441, opened 2021-02-15
Ensure that 'title' property is passed down by cucumber framework to reporters

rickschubert:
This resolves the problem which was reported earlier at #6438 . The problem was that the cucumber-framework did create an undefined `title` property for hooks as these do not possess a `step.text` property.

### DIFF
--- a/packages/wdio-cucumber-framework/src/utils.ts
+++ b/packages/wdio-cucumber-framework/src/utils.ts
@@ -88,13 +88,6 @@ export function getFeatureId (uri: string, feature: messages.GherkinDocument.IFe
     return `${path.basename(uri)}:${feature.location?.line}:${feature.location?.column}`
 }
 
-function createStepTitle(step: ExtendedPickleStep): string {
-    if (isStepTypeHook(step)) {
-        return `hook-${step.hookId}`
-    }
-    return step.text
-}
-
 /**
  * build payload for test/hook event
  */
@@ -102,7 +95,7 @@ export function buildStepPayload(
     uri: string,
     feature: messages.GherkinDocument.IFeature,
     scenario: messages.IPickle,
-    step: messages.Pickle.IPickleStep,
+    step: ExtendedPickleStep,
     params: {
         type: string
         state?: messages.TestStepFinished.TestStepResult.Status | string | null
@@ -115,7 +108,7 @@ export function buildStepPayload(
 ) {
     return {
         uid: step.id,
-        title: createStepTitle(step),
+        title: isStepTypeHook(step) ? `hook-${step.hookId}` : step.text,
         parent: scenario.id,
         argument: createStepArgument(step),
         file: uri,
--- a/packages/wdio-cucumber-framework/src/utils.ts
+++ b/packages/wdio-cucumber-framework/src/utils.ts
@@ -64,15 +64,20 @@ export function formatMessage ({ payload = {} }: any) {
 }
 
 function isStepTypeHook(step: messages.TestCase.ITestStep): boolean {
-    return step.hookId !== undefined && step.hookId !== null && step.hookId !== false
+    return step.hookId !== undefined && step.hookId !== null
+}
+
+enum StepType {
+    hook = 'hook',
+    test = 'test'
 }
 
 /**
  * Get step type
  * @param {string} type `Step` or `Hook`
  */
-export function getStepType (step: messages.TestCase.ITestStep) {
-    return isStepTypeHook(step) ? 'hook' : 'test'
+export function getStepType (step: messages.TestCase.ITestStep): StepType[keyof StepType] {
+    return isStepTypeHook(step) ? StepType.hook : StepType.test
 }
 
 export function getFeatureId (uri: string, feature: messages.GherkinDocument.IFeature) {
--- a/packages/wdio-cucumber-framework/src/utils.ts
+++ b/packages/wdio-cucumber-framework/src/utils.ts
@@ -63,17 +63,9 @@ export function formatMessage ({ payload = {} }: any) {
     return content
 }
 
-function isStepTypeHook(step: messages.TestCase.ITestStep): boolean {
-    return step.hookId !== undefined && step.hookId !== null
-}
-
 enum StepType {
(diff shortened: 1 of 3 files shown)


rickschubert:
This is my first contribution to the webdriverio project; please let me know if there are things that I should change.

wswebcreation (changes requested):
Hi @rickschubert 

Thanks for your first contribution. Can you also add some Unit tests for these changes?

packages/wdio-cucumber-framework/src/utils.ts
```diff
@@ -95,7 +115,7 @@ export function buildStepPayload(
 ) {
     return {
         uid: step.id,
-        title: step.text,
+        title: createStepTitle(step),
```
christian-bromann, line 118: Let's safe one more function call

```suggestion
        title: isStepTypeHook(step) ? `hook-${step.hookId}` : step.text,
```

christian-bromann (commented):
As @wswebcreation mentioned a unit test for this would be great!

rickschubert:
> Hi @rickschubert
> 
> Thanks for your first contribution. Can you also add some Unit tests for these changes?

Okay, I will have a look.

packages/wdio-cucumber-framework/src/utils.ts
rickschubert: Thanks, I will apply this.

packages/wdio-cucumber-framework/src/utils.ts
```diff
@@ -63,18 +63,38 @@ export function formatMessage ({ payload = {} }: any) {
...
+    return isStepTypeHook(step) ? StepType.hook : StepType.test
 }
 
 export function getFeatureId (uri: string, feature: messages.GherkinDocument.IFeature) {
     return `${path.basename(uri)}:${feature.location?.line}:${feature.location?.column}`
 }
 
+function createStepTitle(step: ExtendedPickleStep): string {
+    if (isStepTypeHook(step)) {
+        return `hook-${step.hookId}`
```
wswebcreation, line 93: The challenge here is that Cucumber changed their event system and structure, see also https://github.com/cucumber/cucumber/blob/master/messages/messages.md
This means that its not easy to detect the title and keyword. For example

We used to have this

```js
{
  type: 'Hook',
  location: {
    line: 191,
    column: 0,
    uri: 'node_modules/@wdio/cucumber-framework/build/index.js'
  },
  keyword: 'Hook',
  text: ''
}
```

And it is now this

```js
{
  "id": "33",
  "hookId": "25"
}
```

for now I would suggest to return the word `Hook` for the title if the `step.text===undefined`, what do you think @christian-bromann?

christian-bromann: sounds good to me

rickschubert:
@wswebcreation and @christian-bromann I have applied the code review feedback. Could you take another look please?

packages/wdio-cucumber-framework/src/utils.ts
rickschubert: Does this require other changes to this pull request or is it just something related?

packages/wdio-cucumber-framework/src/utils.ts
```diff
@@ -115,7 +108,7 @@ export function buildStepPayload(
 ) {
     return {
         uid: step.id,
-        title: createStepTitle(step),
+        title: isStepTypeHook(step) ? `hook-${step.hookId}` : step.text,
```
wswebcreation, line 111: I would set this to `Hook` here's an example of a hook in V6

```js
{
  uid: 'index.js187',
  title: 'Hook',
  parent: 'LoginPage: As a user I want to be able to login',
  argument: undefined,
  file: 'test/features/login.feature',
  tags: [ { type: 'Tag', location: [Object], name: '@happy-flow' } ],
  keyword: 'Hook',
  featureName: 'LoginPage',
  scenarioName: 'As a user I want to be able to login',
  type: 'hook'
}
```

and a Step in V6

```js
{
  uid: 'I load the Swag Labs demo website5',
  title: 'Given I load the Swag Labs demo website',
  parent: 'LoginPage: As a user I want to be able to login',
  argument: undefined,
  file: 'test/features/login.feature',
  tags: [ { type: 'Tag', location: [Object], name: '@happy-flow' } ],
  keyword: 'Given ',
  featureName: 'LoginPage',
  scenarioName: 'As a user I want to be able to login',
  type: 'step',
  state: 'pass',
  error: undefined,
  duration: 221,
  passed: true
}
```

rickschubert: Okay, I will amend this.

packages/wdio-cucumber-framework/src/utils.ts
```diff
@@ -108,7 +108,7 @@ export function buildStepPayload(
 ) {
     return {
         uid: step.id,
-        title: isStepTypeHook(step) ? `hook-${step.hookId}` : step.text,
+        title: isStepTypeHook(step) ? 'Hook' : step.text,
```
wswebcreation, line 111: Hi @rickschubert 

@christian-bromann and I took a look at it and we need to change more code to prevent Allure and other reporters for Cucumber to work with the never version of CucumberJS (we're also missing tags, keywords, datatables and so on).
I'll dive into this next week, but for now we think the fix can be made a bit simpler with the following

Change 

`title: isStepTypeHook(step) ? 'Hook' : step.text,`

to

`title: step.text || 'Hook',`

So it should work for your situation. That means you can remove the rest of your code. Can you update that so we can approve it and release it?

rickschubert: Okay, I can amend this.

christian-bromann (approved):
LGTM 👍 

@wswebcreation thoughts?

wswebcreation (approved):
LGTM, thanks!

christian-bromann:
Seems like a flake in the unit test, will go ahead and merge.

rickschubert:
Thanks for the support with my first contribution here; happy to have helped out.

PR #6472, opened 2021-02-18
Rename `jasmineNodeOpts` into `jasmineOpts`

christian-bromann:
## Proposed changes

There is currently a bug where TypeScript expects a `jasmineOpts` in the config while the framework adapter expects `jasmineNodeOpts`. Given that `mochaOpts` and `cucumberOpts` don't have `Node` in their property name I think it is time to consolidate this.

Note: I made sure that we still accept a `jasmineNodeOpts` property in the framework adapter so this change is still backwards compatible.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

n/a

### DIFF
--- a/examples/wdio.conf.js
+++ b/examples/wdio.conf.js
@@ -146,8 +146,8 @@ exports.config = {
     },
     //
     // Options to be passed to Jasmine.
-    // See also: https://github.com/webdriverio/webdriverio/tree/main/packages/wdio-jasmine-framework#jasminenodeopts-options
-    jasmineNodeOpts: {
+    // See also: https://github.com/webdriverio/webdriverio/tree/main/packages/wdio-jasmine-framework#jasmineopts-options
+    jasmineOpts: {
         //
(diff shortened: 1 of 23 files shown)


wswebcreation:
@christian-bromann 

I think we can change it, but this will be a **_breaking_** change for people. 
Maybe also make this clear in a blogpost or in the runner docs itself that this is a breaking change if they migrate from V6 to V7

christian-bromann:
> but this will be a **_breaking_** change for people.

It will be breaking only for people using `jasmineNodeOpts` as cli parameter which not many people do. The types are already expecting (wrongly) you to write `jasmineOpts`. Please note my comment:

> I made sure that we still accept a jasmineNodeOpts property in the framework adapter so this change is still backwards compatible.

wswebcreation:
Ok, cool, thanks, then I think it's good to go 🚀

wswebcreation (approved):

PR #6507, opened 2021-02-24
Fix unable to use nativeWebTap in capabilities

lukyth:
## Proposed changes

Adding `nativeWebTap` to `capabilities` will result in an error:

```
Invalid or unsupported WebDriver capabilities found ("nativeWebTap"). Ensure to only use valid W3C WebDriver capabilities (see https://w3c.github.io/webdriver/#capabilities).
```

This is because `nativeWebTap` is not in the valid [capabilities list](https://github.com/lukyth/webdriverio/blob/cf06843fb1add1645d98b4dd4a1c50c35ec8d8ad/packages/webdriver/src/constants.ts#L153-L157) which is used in this [capabilities validation](https://github.com/lukyth/webdriverio/blob/01f045635ad191db89b6ba4117549edaa8081b00/packages/webdriver/src/utils.ts#L28-L48).

It could be fixed by adding `nativeWebTap` to the list. But seeing this [table of standard capabilities](https://w3c.github.io/webdriver/#capabilities), I think is not appropriate to do so.

So I decided to fix this by adding type definition for `appium:nativeWebTap`, which will pass the capablities validation but its type was missing before. And remove type definition for `nativeWebTap` since it'll cause an error when used.

I tested using BrowserStack with iOS capabilities locally, with something like:
```ts
await browser.url(urlWithATag)
const currentHandles = await browser.getWindowHandles()
await browser.click("a") // <a href="otherpage.html" target="_blank" rel="noreferrer noopener">Open</a>
const newHandles = await browser.getWindowHandles()
expect(newHandles.length).toBeGreaterThan(currentHandles.length)
```
This will fail without `nativeWebTap`. And with this I confirmed that `appium:nativeWebTap` and `nativeWebTap` can be used interchangeably.

I'm not sure how should I add a test for this here. Please let me know if you want me to add a test for this anywhere.

I could also add `appium:` prefix to `AppiumAndroidCapabilities` and `AppiumIOSCapabilities` (since using any other capabilities in there will probably cause the same error), but I only tested `nativeWebTap` so far, so I only updated that. Please let me know if want me to update those twos as well, or if you think I should update the validation logic to allow non-prefix capabilities instead. But personally, I think having a prefix is better since it's more clear what it is.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/wdio-types/src/Capabilities.ts
+++ b/packages/wdio-types/src/Capabilities.ts
@@ -439,7 +439,6 @@ export interface AppiumIOSCapabilities {
     autoAcceptAlerts?: boolean;
     autoDismissAlerts?: boolean;
     nativeInstrumentsLib?: boolean;
-    nativeWebTap?: boolean;
     safariInitialUrl?: string;
     safariAllowPopups?: boolean;
     safariIgnoreFraudWarning?: boolean;
@@ -460,6 +459,7 @@ export interface AppiumIOSCapabilities {
     enableAsyncExecuteFromHttps?: boolean;
     skipLogCapture?: boolean;
     webkitDebugProxyPort?: number;
+    'appium:nativeWebTap'?: boolean;
 }
 
 // IE specific
--- a/packages/wdio-types/src/Capabilities.ts
+++ b/packages/wdio-types/src/Capabilities.ts
@@ -356,6 +356,7 @@ export interface AppiumW3CCapabilities {
     'appium:eventTimings'?: boolean;
     'appium:enablePerformanceLogging'?: boolean;
     'appium:printPageSourceOnFindFailure'?: boolean;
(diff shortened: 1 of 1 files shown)


christian-bromann (changes requested):
Thanks for raising the PR. I suggest a [better interface](https://github.com/webdriverio/webdriverio/blob/main/packages/wdio-types/src/Capabilities.ts#L342-L359) to add this too.

lukyth:
@christian-bromann Sorry for the late update. I moved the `appium:nativeWebTap` to `AppiumW3CCapabilities` as you suggested.

I also just realized that the capabilities validation will allow `nativeWebTap` if it's not being used with other extension capabilities, so I reverted the change that removes its type as well 🙇

christian-bromann (approved):
👍  thanks!

Issue #6515, opened 2021-02-26
[corner case] exception when calling `click` if the window is closed after the click


link89:
One of our products has a detached window that can be put back by clicking an `attach back` button. By doing this the window will be closed. We found that the following error will be thown.

`ERROR webdriver: Request failed with status 404 due to no such window: no such window: window was already closed
  (Session info: chrome=88.0.4324.192)`

After some investigation, I found that `webdirver.io` implements the` click` method by `ActionChain`. And it will call `Release Actions` after `Perform Actions` is executed. And since the window is no longer exists after calling the `Perform Action`, the `Release Actions` will always fail.  We can't even try-catch this exception because it is an async call.

We are now working around this problem by simulating the click action via `execute script`. But I hope the bug could be fixed in `webdriver.io`.

Webdriver.io version: 6.6.6

christian-bromann:
@link89 thanks for filing an issue. This is indeed an interesting corner case. I suggest to wrap the `releaseAction` command in our [click implementation](https://github.com/webdriverio/webdriverio/blob/main/packages/webdriverio/src/commands/element/click.ts#L133) within a try/catch and only re-throw the error if it doesn't contain `no such window`. I think this is a nice first ticket to work on, you want to take a stab at it?

christian-bromann:
Let's continue conversation in #6535

PR #6535, opened 2021-03-03
catch the 'no such window' error for a corner case

link89:
fix https://github.com/webdriverio/webdriverio/issues/6515

### DIFF
--- a/packages/webdriverio/src/commands/element/click.ts
+++ b/packages/webdriverio/src/commands/element/click.ts
@@ -129,8 +129,14 @@ export default async function click (
                 button
             }]
         }])
-
-        return this.releaseActions()
+        try {
+            return await this.releaseActions()
+        } catch (e) {
+            if (e.name === 'no such window') {
+                return
+            }
+            throw e
+        }
     }
 
     const { width, height } = await this.getElementSize(this.elementId)


packages/webdriverio/src/commands/element/click.ts
```diff
@@ -129,8 +129,14 @@ export default async function click (
                 button
             }]
         }])
-
-        return this.releaseActions()
+        try {
```
christian-bromann, line 132: Can we add a comment as to why we catch this here and ignore the `no such window` error?

christian-bromann (changes requested):

packages/webdriverio/src/commands/element/click.ts
link89: > Can we add a unit test for this?

Any guidelines for the unit tests? I guess I need to mock the action chains request in `got.ts` but not sure how to write it.

christian-bromann: Have a look at the [existing unit tests](https://github.com/webdriverio/webdriverio/blob/main/packages/webdriverio/tests/commands/element/click.test.ts). Everything is already mocked and you can just add a test for your use case in a similar fashion. Let me know if anything is unclear.

christian-bromann:
@link89 anything I can do to help?

christian-bromann:
ping @link89

christian-bromann:
Closing due to inactivity.

PR #6555, opened 2021-03-05
Added overload typings to getLocation function

### LINKED ISSUE #6549: Improve typings by using function overloads
The Typescript signatures are sometimes too loose and unnecessarily hard to use. Take `getLocation` as an example:
```
export default function getLocation(this: WebdriverIO.Element, prop?: 'x' | 'y'): Promise<number | {
    x: number;
    y: number;
}>;
```
This is hard to use because the return type is not bound to the input type. I have to do a type cast which should not be needed.

This could be easily solved by using overload signatures like this:

```
export function getLocation(this: WebdriverIO.Element): Promise<{
    x: number;
    y: number;
}>
export function getLocation(this: WebdriverIO.Element, prop: 'x' | 'y'): Promise<number>
function getLocation(this: WebdriverIO.Element, prop?: 'x' | 'y'): Promise<number | {
    x: number;
    y: number;
}> {
   // Implementation does not need to change ...
}
```

**Additional context**

This is just an example that - together with `getSize` - jumped me right in the face while updating our codebase to WebdriverIO 7.

azygis:
## Proposed changes

Add more explicit typings where possible. Currently - `getLocation` function. Potentially resolves #6549.

## Types of changes

- [ ] Documentation update
- [x] Improvement for development (_well... It's kinda none of the above_)

## Checklist

The checklist **will be** completed.

## Further comments

Does anyone have any specific pointers on which functions/files/etc could be changed according to this as well? One function might be _a little too little_. Or shall I just glance over a bunch of files to check what else could be changed?

Edit:
I've noticed now that the build fails because of `no-redeclare` rule. Any suggestions on that? Do we override the rule in some way? It looks like `@typescript-eslint/no-redeclare` should do the trick instead of the generic one.

### DIFF
--- a/packages/webdriverio/src/commands/element/getLocation.ts
+++ b/packages/webdriverio/src/commands/element/getLocation.ts
@@ -1,5 +1,10 @@
 import { getElementRect } from '../../utils'
 
+type Location = {
+    x: number,
+    y: number
+}
+
 /**
  *
  * Determine an element’s location on the page. The point (0, 0) refers to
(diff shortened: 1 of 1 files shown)


azygis:
Another person started the work on this.

PR #6558, opened 2021-03-06
docs(typings): #6549 improve typings using overloads - getLocation, getSize

iamkenos:
## Proposed changes

- function overloads as suggested [here](https://github.com/webdriverio/webdriverio/issues/6549#issue-822350507)

the idea is to have something like this as a result:
Screenshot 2021-03-06 at 11 14 45 PM

## Types of changes

- [x] Documentation update (typings)

## Checklist

## Further comments

1. The [issue](https://github.com/webdriverio/webdriverio/issues/6549) we're aiming to fix gave `getLocation` and `getSize` as examples. I skimmed other browser and element commands under the webdriverio package that may need the same change and couldn't find others. I could be mistaken though so would appreciate if someone can check that.

2. ~~Not really sure if I would need to add tests for this either, for now I did without.~~

3. There's a completely unrelated [commit](https://github.com/webdriverio/webdriverio/pull/6558/commits/413b4e2465752f01df182a7126995ac83a05faed) in this PR; i had to add this in because of [this recent change](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/51520/files) that broke the tests 

### DIFF
--- a/packages/webdriverio/src/commands/element/getLocation.ts
+++ b/packages/webdriverio/src/commands/element/getLocation.ts
@@ -1,13 +1,11 @@
 import type { RectReturn } from '@wdio/protocols'
 import { getElementRect } from '../../utils'
 
-export type Coordinates = Pick<RectReturn, 'x' | 'y'>;
+export type Location = Pick<RectReturn, 'x' | 'y'>;
 
-export type Location<T> = T extends keyof Coordinates ? number : Coordinates
+function getLocation (this: WebdriverIO.Element): Promise<Location>
 
-function getLocation<T extends undefined> (this: WebdriverIO.Element): Promise<Location<T>>
-
-function getLocation<T extends keyof Coordinates> (this: WebdriverIO.Element, prop: T): Promise<Location<T>>
+function getLocation (this: WebdriverIO.Element, prop: keyof Location): Promise<number>
 
 /**
  *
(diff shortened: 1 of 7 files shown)


packages/webdriverio/src/commands/element/getLocation.ts
```diff
@@ -27,16 +36,11 @@ import { getElementRect } from '../../utils'
  * @uses protocol/elementIdLocation
  * @type property
  */
-export default async function getLocation (
+export async function getLocation<T extends keyof Coordinates> (
```
erwinheitzman, line 39: why export twice this method twice?

erwinheitzman (commented):
Thank you for the contribution, love the idea. Left a quick question regarding the double export

erwinheitzman:
We do have tests for our types and we would love to see these changes covered as well =]

packages/webdriverio/src/commands/element/getLocation.ts
iamkenos: Hi! By twice, were you referring to the default export below or the named exports above?

iamkenos:
> We do have tests for our types and we would love to see these changes covered as well =]

Apologies and thanks in advance for the patience, I may need some hand holding here 😅

Would you be able to point out a file I can refer to for some samples on typings ut please? Can't promise I can do it by i'll give it a try!

erwinheitzman:
No worries, here you go: https://github.com/webdriverio/webdriverio/tree/main/tests/typings

iamkenos:
> No worries, here you go: https://github.com/webdriverio/webdriverio/tree/main/tests/typings

thanks! added the tests too

iamkenos:
> Thank you for the contribution, love the idea. Left a quick question regarding the double export

Removed redundant exports, thanks for highlighting! I ended up removing the generics, was getting the some weird issues with the return types and it turns out it was because of the default export.

christian-bromann (approved):
Great patch, thanks! 👍

PR #6574, opened 2021-03-10
docs(typings): #6458 use a more suitable return type for waitUntil

### LINKED ISSUE #6458: Stricter type definition for `waitUntil`
from #6412

> I also noticed waitUntil is hardcoded to return Promise<any> - Maybe this should return Promise<void> because either it waits or it throws an error, we shouldn't care about what it returns. If we do, then maybe a generic for the return value? This was throwing me off because I thought the any was related to the original issue. https://github.com/webdriverio/webdriverio/blob/main/packages/webdriverio/src/commands/browser/waitUntil.ts#L71

iamkenos:
## Proposed changes

- fixes #6458 
- basically just define the return type 😬 

## Types of changes

- [x] Documentation update - typings

## Checklist

## Further comments

### DIFF
--- a/packages/webdriverio/src/commands/browser/waitUntil.ts
+++ b/packages/webdriverio/src/commands/browser/waitUntil.ts
@@ -50,7 +50,7 @@ export default function waitUntil(
         interval = this.options.waitforInterval,
         timeoutMsg
     }: Partial<WaitUntilOptions> = {}
-) {
+): Partial<true | void> {
     if (typeof condition !== 'function') {
         throw new Error('Condition is not a function')
     }
--- a/packages/webdriverio/src/commands/browser/waitUntil.ts
+++ b/packages/webdriverio/src/commands/browser/waitUntil.ts
@@ -50,7 +50,7 @@ export default function waitUntil(
         interval = this.options.waitforInterval,
         timeoutMsg
     }: Partial<WaitUntilOptions> = {}
-): true | void {
+): Promise<true | void> {
     if (typeof condition !== 'function') {
         throw new Error('Condition is not a function')
     }
--- a/packages/webdriverio/src/commands/browser/waitUntil.ts
+++ b/packages/webdriverio/src/commands/browser/waitUntil.ts
@@ -50,7 +50,7 @@ export default function waitUntil(
         interval = this.options.waitforInterval,
         timeoutMsg
     }: Partial<WaitUntilOptions> = {}
-): Partial<true | void> {
+): true | void {
     if (typeof condition !== 'function') {
         throw new Error('Condition is not a function')
     }
--- a/packages/webdriverio/tests/commands/browser/waitUntil.test.ts
+++ b/packages/webdriverio/tests/commands/browser/waitUntil.test.ts
@@ -18,11 +18,12 @@ describe('waitUntil', () => {
 
     it('Should throw an error if an invalid condition is used', async () => {
         let error
+        let val
         // @ts-ignore uses expect-webdriverio
-        expect.assertions(1)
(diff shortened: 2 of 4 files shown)


christian-bromann:
> Can i request for an example where it would return a value other than true please?

Sure, I think the comments are wrong then, I double checked:

```js
› browser.waitUntil(() => 'foobar')
'foobar'
› browser.waitUntil(() => 123)
123
```

iamkenos:
> > Can i request for an example where it would return a value other than true please?
> 
> Sure, I think the comments are wrong then, I double checked:
> 
> ```js
> › browser.waitUntil(() => 'foobar')
> 'foobar'
> › browser.waitUntil(() => 123)
> 123
> ```

i see, thanks! 

this is actually what i meant by - `...the check for that is already imposed by the argument type.`

image

one thing we can do is change the arg type to truthy instead of boolean, then combine it with generics to use the same arg type as the return type. it's [not gonna be straightforward](https://stackoverflow.com/questions/51390276/type-for-non-falsy-aka-truthy) but it will be more accurate.

personally, i've mixed feelings about this...thoughts?

packages/webdriverio/src/commands/browser/waitUntil.ts
```diff
@@ -50,7 +50,7 @@ export default function waitUntil(
         interval = this.options.waitforInterval,
         timeoutMsg
     }: Partial<WaitUntilOptions> = {}
-) {
+): true | void {
```
christian-bromann, line 53:
```suggestion
): Promise<true | void> {
```

christian-bromann:
> one thing we can do is change the arg type to truthy instead of boolean, then combine it with generics to use the same arg type as the return type. it's [not gonna be straightforward](https://stackoverflow.com/questions/51390276/type-for-non-falsy-aka-truthy) but it will be more accurate.

That would be my preference given that this is how it actually works, regardless of the types. Also I can see that user would like to use the value returned by `waitUntil`

iamkenos:
> > one thing we can do is change the arg type to truthy instead of boolean, then combine it with generics to use the same arg type as the return type. it's [not gonna be straightforward](https://stackoverflow.com/questions/51390276/type-for-non-falsy-aka-truthy) but it will be more accurate.
> 
> That would be my preference given that this is how it actually works, regardless of the types. Also I can see that user would like to use the value returned by `waitUntil`

Noted, thanks! Lemme see what i can do!

iamkenos:
> > > one thing we can do is change the arg type to truthy instead of boolean, then combine it with generics to use the same arg type as the return type. it's [not gonna be straightforward](https://stackoverflow.com/questions/51390276/type-for-non-falsy-aka-truthy) but it will be more accurate.
> > 
> > 
> > That would be my preference given that this is how it actually works, regardless of the types. Also I can see that user would like to use the value returned by `waitUntil`
> 
> Noted, thanks! Lemme see what i can do!

hi @christian-bromann 

i tried [this](https://stackoverflow.com/a/51390763/2285470) approach and bumped into the limitation that was highlighted -- `Note how it won't allow anything which it thinks might be falsy...`

samples: 
```
browser.waitUntil(() => '') // type error, expected
browser.waitUntil(() => 123) // ok
browser.waitUntil(() => 'false') //ok
browser.waitUntil(() => true) // ok
browser.waitUntil(() => false) // type error, expected
browser.waitUntil(() => el.isClickable()) // type error, should be allowed but might be falsy so nope
```

for now i kept the changes minimal, as per last requests. thoughts?

christian-bromann (approved):
LGTM 👍  thank you so much!

PR #6099, opened 2021-03-12
Upload WDIO logs to Sauce Labs via Sauce Service

christian-bromann:
It would be useful for Sauce Labs user to have their WDIO logs uploaded to the platform so in case of an issue support people have the ability to get more insights about what is going on in the test.

### DIFF
--- a/packages/wdio-sauce-service/README.md
+++ b/packages/wdio-sauce-service/README.md
@@ -160,7 +160,7 @@ Default: `{ noAutodetect: true }`
 
 ### uploadLogs
 
-If `true` this options uploads all WebdriverIO log files to the Sauce Labs platform for further inspection.
+If `true` this options uploads all WebdriverIO log files to the Sauce Labs platform for further inspection. Make sure you have [`outputDir`](https://webdriver.io/docs/options#outputdir) set in your wdio config to write logs into files, otherwise data will be streamed to stdout and can't get uploaded.
 
 Type: `Boolean`<br />
 Default: `true`
--- a/packages/wdio-sauce-service/README.md
+++ b/packages/wdio-sauce-service/README.md
@@ -148,15 +148,22 @@ Default: `false`
 *(only for vm and or em/simulators)*
 
 ### sauceConnectOpts
-Apply Sauce Connect options (e.g. to change port number or logFile settings). See [this list](https://wiki.saucelabs.com/display/DOCS/Sauce+Connect+Proxy+Command-Line+Quick+Reference+Guide) for more information. Per default the service disables SC proxy auto-detection as via `noAutodetect` as this can be unreliable for some machines.  
+Apply Sauce Connect options (e.g. to change port number or logFile settings). See [this list](https://wiki.saucelabs.com/display/DOCS/Sauce+Connect+Proxy+Command-Line+Quick+Reference+Guide) for more information. Per default the service disables SC proxy auto-detection as via `noAutodetect` as this can be unreliable for some machines.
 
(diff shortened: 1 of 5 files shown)


packages/wdio-sauce-service/src/types.ts
```diff
@@ -24,13 +24,21 @@ export interface SauceServiceConfig {
...
      * Apply Sauce Connect options (e.g. to change port number or logFile settings). See this
      * list for more information: https://github.com/bermi/sauce-connect-launcher#advanced-usage
      *
      * @default {}
      */
     sauceConnectOpts?: SauceConnectOptions
+
+    /**
+     * Upload WebdriverIO logs to the Sauce Labs platform.
+     * @default false
```
wswebcreation, line 38: Does this really slow down the execution if we just enable it by default? I think it would be really helpful to always have this information in Sauce

packages/wdio-sauce-service/src/service.ts
```diff
@@ -216,17 +219,39 @@ export default class SauceService implements Services.ServiceInstance {
...
+     * upload files to Sauce Labs platform
+     * @param jobId id of the job
+     * @returns a promise that is resolved once all files got uploaded
+     */
+    private async _uploadLogs (jobId: string) {
+        if (!this._options.uploadLogs || !this._config.outputDir) {
+            return
+        }
+
+        const files = (await fs.promises.readdir(this._config.outputDir))
```
wswebcreation, line 245: Am I reading this correctly that this is connected to also writing the logs to your local disk in order for this to work?
If so, I'm not seeing that back in the docs, secondly, can we read this from the stream by default and push it then to Sauce, so the local disk won't be "polluted" with extra files?

wswebcreation (commented):
Hi @christian-bromann 

I really love the feature and maybe even love to see it as a default feature. Please see my feedback in the files. Last comment I have is more a question. Will the documentation be automatically be updated with this PT because I only see the types being updated and not the readme

packages/wdio-sauce-service/src/types.ts
christian-bromann: I am not too much concerned about execution time but rather about whether or not people actually look into the data. We can have it default to `true` and have a dashboard for this to check if people actually use it or not.

packages/wdio-sauce-service/src/service.ts
christian-bromann: Good point, let me look into this. My hunch is that it would only work for logs we store in files because a stream never keeps data in memory.

christian-bromann: Updated readme to add option description there. Unfortunately it is not possible to get data from a stream once written to the target (here stdout).

christian-bromann:
@wswebcreation addressed comments, please have another look. Thanks!

packages/wdio-sauce-service/README.md
```diff
@@ -148,16 +148,23 @@ Default: `false`
...
 It can also be turned into camelCase (e.g. `shared-tunnel` or `sharedTunnel`).
 
 Type: `Object`<br />
 Default: `{ noAutodetect: true }`
 
 *(only for vm and or em/simulators)*
 
+### uploadLogs
+
+If `true` this options uploads all WebdriverIO log files to the Sauce Labs platform for further inspection.
```
wswebcreation, line 163: Just one last remark, this will only work if the logs are written to a file, but I don't see it in the docs here. If you would add it here then it's good to go

PR #6594, opened 2021-03-17
Better expose config handler

christian-bromann:
In order to allow CLI tools to better integrate with WebdriverIO it would be useful if `@wdio/cli` would expose the config wizard. In addition to that it would be useful which packages the config wizard did install.

I will assign this to me as I am working on the Angular CLI integration and found need for this.


WillBrock (approved):
LGTM

PR #6606, opened 2021-03-19
webdriverio: fix setvalue typings

erwinheitzman:
## Proposed changes

Someone reported that the `setValue` only allows `string` values, even though it should accept more.

Because we change the type for things that can be passed, this will break this command for users that are passing objects or boolean values.

## Types of changes

- [ ] Documentation update

## Checklist


christian-bromann (commented):
Does `object` and `boolean` make sense here? Also rather than `Array<any>` I would say it should be `string[]`, right?

kevin-longmate:
@christian-bromann I mentioned this to @erwinheitzman on gitter.
As long as string and number are there then I'm happy, the rest of the types were taken from https://webdriver.io/docs/api/element/setValue/
Feel free to change or remove them as you like

erwinheitzman:
I have taken the types from addValue because they depend on each other. If we feel that the types are incorrect then we need to update both commands and their documentation.

christian-bromann:
@erwinheitzman I would suggest to update in both.

erwinheitzman:
@christian-bromann in that case I feel that the only valid value would be a string. Do you see any use for for example passing an array of strings to a single input element?

christian-bromann:
> Do you see any use for for example passing an array of strings to a single input element?

Yes, sending a string array allows to send a combination of different key strokes. e.g. to copy&paste you would do:

```js
$('#source').addValue(['CTRL', 'a', 'c']) // copy all content
$('#target').setValue(['CTRL', 'v'])
```

I would also allow numbers for simplicity, so:

```js
$('#source').addValue([1, 2, 3])
```

would be the same as:

```js
$('#source').addValue('123')
```

I can see that sometimes it would be easier to pass in a number variable rather than transforming it into a string.

That said, my suggestion would be to type it as `(number|string)[]`. What do you think?

erwinheitzman:
I'll update the types accordingly and I'd love to discuss on Hitter how we can improve what I shared above

christian-bromann:
@erwinheitzman found the issue: https://github.com/webdriverio/webdriverio/issues/4065 , with TS now it would totally make sense to allow something like this:

```ts
import { Keyboard } from 'webdriverio'
browser.setValue([Keyboard.CTRL, 'a'])
```

What do you think?

erwinheitzman:
I'd love that

packages/webdriverio/src/commands/element/setValue.ts
```diff
@@ -1,5 +1,7 @@
 import type { AddValueOptions } from './addValue'
 
+type Value = string | number
```
christian-bromann, line 3: Let's import this from `addValue`
```suggestion
import type { AddValueOptions, ValueType } from './addValue'
```

christian-bromann (commented):
Can we add some typings test that ensures only the right types can be set, e.g.

```ts
browser.setValue('foobar')
browser.setValue(['f', 'o', 'o', 'b', 'a', 'r'])
browser.setValue(123)
browser.setValue([1, 'o', 2, 'b', 3, 'r'])
// @ts-expect-error
browser.setValue(true)
```

erwinheitzman:
I have done some extra work but I think it was worth diving a bit deeper and solving this properly

packages/webdriverio/src/commands/element/addValue.ts
```diff
@@ -1,9 +1,22 @@
...
 
-export type AddValueOptions = {
+export type CommandOptions = {
     translateToUnicode?: boolean
 }
 
+export type Value = string | number
+
+function isValidType(value: unknown) {
+    const isNumberOrString = (_: unknown) => typeof _ === 'string' || typeof _ === 'number'
```
christian-bromann, line 10: let's give the parameter a proper name
```suggestion
    const isNumberOrString = (value: unknown) => typeof value === 'string' || typeof value === 'number'
```

packages/webdriverio/src/commands/element/addValue.ts
```diff
@@ -1,9 +1,22 @@
...
+
+    if (isNumberOrString(value)) {
+        return true
+    } else if (Array.isArray(value) && value.every((item) => isNumberOrString(item))) {
+        return true
+    }
+    return false
```
christian-bromann, line 17: How about
```suggestion
    return (
        isNumberOrString(value)) ||
        Array.isArray(value) && value.every((item) => isNumberOrString(item))
    )
```

packages/webdriverio/src/utils/index.ts
```diff
@@ -101,11 +101,11 @@ export function getBrowserObject (elem: WebdriverIO.Element | WebdriverIO.Browse
 /**
  * transform whatever value is into an array of char strings
  */
-export function transformToCharString (value: any, translateToUnicode = true) {
+export function toUnicodeCharacterArray (value: string | number | (string|number)[], translateToUnicode = true) {
     const ret: string[] = []
 
     if (!Array.isArray(value)) {
-        value = [value]
+        value = [value as (number | number)]
```
christian-bromann, line 108: seems to be a duplicate
```suggestion
        value = [value as string | number]
```

packages/webdriverio/src/utils/index.ts
```diff
@@ -116,13 +116,6 @@ export function transformToCharString (value: any, translateToUnicode = true) {
         } else if (typeof val === 'number') {
             const entry = `${val}`.split('')
             ret.push(...entry)
-        } else if (val && typeof val === 'object') {
-            try {
-                ret.push(...JSON.stringify(val).split(''))
-            } catch (e) { /* ignore */ }
-        } else if (typeof val === 'boolean') {
-            const entry = val ? 'true'.split('') : 'false'.split('')
-            ret.push(...entry)
```
christian-bromann, line 125: mhm 🤔 I am wondering if anyone did use this. I would keep it around to not make this a breaking change for folks that do use this.

packages/webdriverio/src/utils/index.ts
erwinheitzman: this is actually correct it's either string or number and needs to be surrounded by () or it won't accept it correctly

packages/webdriverio/src/utils/index.ts
erwinheitzman: This was only used by `addValue` in the codebase, hence I felt confident it could be removed. Do you strongly feel we should keep this around?

It's not part of our API and I therefor I feel that we shouldn't take into account that someone might have used some of our internal utils.

erwinheitzman: Also, this change in itself is already a breaking one by changing the allowed types/values

packages/webdriverio/src/utils/index.ts
christian-bromann: @erwinheitzman what you propose is `value = [value as (number | number)]` which seems not right? It says cast it to number or number?

packages/webdriverio/src/utils/index.ts
christian-bromann: How about we add a deprecation warning, via:

```js
log.warn("...")
```

And remove it eventually at some point later. While it technically is a breaking change I assume that not the majority of users use TS. Changing this functionality actually does impact everyone.

packages/webdriverio/src/utils/index.ts
erwinheitzman: The value can be a string, number or an array containing numbers and/or strings. If the value is not already an array, which is needed in this case, we wrap the value in an array. In order to do that you need to tell the compiler that the type of the value that we want to wrap is of type string or number instead of type Array<string | number> because this is the type we want to end up with.

All this does is: if the value isn't already an array (thus either a number or string), wrap it in an array (which can also contain strings and/or numbers).

erwinheitzman: I'm going to revert this and add a warning like you suggested though.

Although I can't imagine anyone doing things like `setValue([1, '2', [1, '2'], {}, true])` and such because you'd just end up with an output of `'12[12]{}true'`

erwinheitzman:
I got my fingers crossed haha but if you have any feedback, bring it! :D

erwinheitzman:
After taking a look I think we might want to change the warning a little bit as it feel a bit forced, let me know what you think. It's pretty late and I worked quite a lot today so sorry that this takes some iterations

packages/webdriverio/src/commands/element/addValue.ts
```diff
@@ -1,9 +1,21 @@
...
+export type CommandOptions = {
     translateToUnicode?: boolean
 }
 
+export type Value = string | number
+
+function isValidType(value: unknown) {
+    const isNumberOrString = (input: unknown) => typeof input === 'string' || typeof input === 'number'
+
+    return (
```
christian-bromann, line 13: Let's define this function only once.

```suggestion
export type Value = string | number
const isNumberOrString = (input: unknown) => typeof input === 'string' || typeof input === 'number'

function isValidType(value: unknown) {
    return (
```

packages/webdriverio/src/commands/element/addValue.ts
```diff
@@ -26,18 +38,22 @@ export type AddValueOptions = {
...
  */
 export default function addValue (
     this: WebdriverIO.Element,
-    value: string | number | boolean | object | Array<any>,
-    { translateToUnicode = true }: AddValueOptions = {}
+    value: Value | Value[],
+    { translateToUnicode = true }: CommandOptions = {}
 ) {
+    if (!isValidType(value)) {
+        new DeprecatedWarning('Support for values that are not of type "string", "number" or "Array<string | number>" will soon be dropped')
```
christian-bromann, line 52: Do you want to log or throw here? Right now I assume it doesn't do anything.

packages/webdriverio/tests/commands/element/addValue.test.ts
```diff
@@ -1,13 +1,24 @@
 // @ts-ignore mocked (original defined in webdriver package)
 import gotMock from 'got'
 import { remote } from '../../../src'
+import { DeprecatedWarning } from '../../../../webdriverio/src/utils/DeprecatedWarning'
+
+jest.mock('../../../../webdriverio/src/utils/DeprecatedWarning')
```
christian-bromann, line 6:
```suggestion
import { DeprecatedWarning } from '../../../src/utils/DeprecatedWarning'

jest.mock('../../../src/utils/DeprecatedWarning')
```

packages/webdriverio/src/commands/element/addValue.ts
christian-bromann: Ah now I have seen the class for it. Is there any advantage of wrapping this within a class to create an instance that is immediately remove from GC? Wouldn't a simple:
```suggestion
        log.warn('@DeprecatedWarning: Support for values that are not of type "string", "number" or "Array<string | number>" will soon be dropped')
```
do the same?

christian-bromann (commented):
A comment on the deprecation approach and two little clean ups.

christian-bromann (approved):
👍  thank you!

Issue #6622, opened 2021-03-25
Incorrect typing for browser.call


giacomomagini:
[//]: # NOTE: This repository only maintains packages that are listed in the [Readme](https://github.com/webdriverio/webdriverio/blob/main/README.md#packages). Please make sure that your issue is directly caused by one of these packages and if not file an issue in the correct 3rd party package repository.

**Environment (please complete the following information):**
 - **WebdriverIO version: 7.2.3
 - **Mode: standalone
 - **If WDIO Testrunner, running sync/async: sync
 - **Node.js version: 14.15.1
 - **NPM version: 6.14.8
 - **Browser name and version:
 - **Platform name and version: MacOS Catalina

**Config of WebdriverIO**
An example of how you configured WebdriverIO config

**Describe the bug**
Typing for `browser.call` doesn't seem correct
![image](https://user-images.githubusercontent.com/16667545/112338724-50f7ee80-8cbf-11eb-98c2-f7549efbb2f9.png)

**To Reproduce**

Follow the typescript setup.
I use VScode and `typescript 4.0.5`.

**Expected behavior**
The return type of `call` is the return type of the function passed as argument.

christian-bromann:
This works for the async types, e.g.:

![Screenshot 2021-03-25 at 12 04 53](https://user-images.githubusercontent.com/731337/112463388-7427aa00-8d62-11eb-83c6-726705e3420f.png)

but since we transform the sync types from the async types all generics get lost. The solution to this is to overwrite the derived types here: https://github.com/webdriverio/webdriverio/blob/main/packages/webdriverio/src/types.ts#L15-L26

Any contributions would be appreciated.

giacomomagini:
Hi @christian-bromann, I'll try open a PR in the next days. Thanks for confirming the issue and pointing out the relevant code!

giacomomagini:
Opened a PR https://github.com/webdriverio/webdriverio/pull/6625

christian-bromann:
Let's continue conversation in #6625

PR #6625, opened 2021-03-27
fix sync browser type for "call" method

giacomomagini:
## Proposed changes

Types for the sync `browser` object are set to `unknown`.
Copy types for the sync `browser` interface.

Issue: https://github.com/webdriverio/webdriverio/issues/6622

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/webdriverio/src/types.ts
+++ b/packages/webdriverio/src/types.ts
@@ -13,16 +13,17 @@ import type { Size } from './commands/element/getSize'
 
 export type BrowserCommandsType = typeof BrowserCommands
 export type BrowserCommandsTypeSync = {
-    [K in keyof Omit<BrowserCommandsType, 'execute'>]: (...args: Parameters<BrowserCommandsType[K]>) => ThenArg<ReturnType<BrowserCommandsType[K]>>
+    [K in keyof Omit<BrowserCommandsType, 'execute' | 'call'>]: (...args: Parameters<BrowserCommandsType[K]>) => ThenArg<ReturnType<BrowserCommandsType[K]>>
 } & {
     /**
      * we need to copy type definitions for execute and executeAsync as we can't copy over
      * generics with method used above
      */
+    call: <T>(fn: () => Promise<T>) => T,
     execute: <ReturnValue, InnerArguments extends any[] = any[], OuterArguments extends InnerArguments = any>(
         script: string | ((...innerArgs: OuterArguments) => ReturnValue),
         ...args: InnerArguments
-    ) => ReturnValue
+    ) => ReturnValue,
 }
 export type ElementCommandsType = typeof ElementCommands
 export type ElementCommandsTypeSync = {


christian-bromann (changes requested):
I think most of them we don't need as we copy them over via:

```ts
{
    [K in keyof Omit<BrowserCommandsType, 'execute'>]: (...args: Parameters<BrowserCommandsType[K]>) => ThenArg<ReturnType<BrowserCommandsType[K]>>
}
```

So only commands that work with generics as we can't propagate that. Let's avoid duplicate definition as much as possible and only keep `call` here.

giacomomagini:
Makes sense, I got threw off by some `any` in the signature while trying the type but they seem coming from the original function. Same for some missing return type.

christian-bromann (commented):
Can we add `call` to the exclude list before?
```suggestion
export type BrowserCommandsTypeSync = {
-    [K in keyof Omit<BrowserCommandsType, 'execute'>]: (...args: Parameters<BrowserCommandsType[K]>) => ThenArg<ReturnType<BrowserCommandsType[K]>>
+    [K in keyof Omit<BrowserCommandsType, 'execute' | 'call'>]: (...args: Parameters<BrowserCommandsType[K]>) => ThenArg<ReturnType<BrowserCommandsType[K]>>
} & {
```

Our define the list of commands that we overwrite in an extra type and use keyof, e.g.

```suggestion
export type BrowserCommandsTypeSync = {
-    [K in keyof Omit<BrowserCommandsType, 'execute'>]: (...args: Parameters<BrowserCommandsType[K]>) => ThenArg<ReturnType<BrowserCommandsType[K]>>
+    [K in keyof Omit<BrowserCommandsType, keyOf CommandsWithGenerics>]: (...args: Parameters<BrowserCommandsType[K]>) => ThenArg<ReturnType<BrowserCommandsType[K]>>
} & {
```

packages/webdriverio/src/types.ts
```diff
@@ -13,16 +13,17 @@ import type { Size } from './commands/element/getSize'
...
     /**
      * we need to copy type definitions for execute and executeAsync as we can't copy over
      * generics with method used above
      */
+    call: <T>(fn: () => T) => T,
     execute: <ReturnValue, InnerArguments extends any[] = any[], OuterArguments extends InnerArguments = any>(
         script: string | ((...innerArgs: OuterArguments) => ReturnValue),
         ...args: InnerArguments
-    ) => ReturnValue
+    ) => ReturnValue,
```
christian-bromann, line 26:
```suggestion
    ) => ReturnValue
```

christian-bromann (approved):
👍

christian-bromann:
It seems that one to of our typing tests fails: https://github.com/webdriverio/webdriverio/blob/main/tests/typings/sync/sync.ts#L70-L73

giacomomagini:
@christian-bromann I checked the failing test but I don't know where that global `browser` var comes from.
If I try to use the type in the file where i made the changes I can see the same line (`<number>browser.call(...)`) working fine.
Any heads up?

christian-bromann:
> but I don't know where that global `browser` var comes from.

It is just a test file that expects the browser to be registered in global scope. This is the case if wdio types are registered in [tsconfig.json](https://github.com/webdriverio/webdriverio/blob/main/tests/typings/sync/tsconfig.json#L6-L22).

I believe the following two files need to be updated:
- tests/typings/webdriverio/async.ts
- tests/typings/sync/sync.ts

Both currently have similar examples for the call command:
```ts
const callResult = <number>browser.call(() =>
    new Promise(resolve => setTimeout(() => resolve(4), 1))
)
callResult.toFixed(2)
```
I think with this PR we should be able to remove the conversion and only have:
```ts
const callResult = browser.call(() =>
    new Promise(resolve => setTimeout(() => resolve(4), 1))
)
callResult.toFixed(2)
```

right?

giacomomagini:
`I think with this PR we should be able to remove the conversion and only have` -> yes, but it should work also specifying the generic.

The type was wrong, the correct function return type is `Promise<T>` instead of `T`.

It's fixed now! :)

christian-bromann (approved):
👍  thank you so much!

christian-bromann:
@giacomomagini would you mind signing the CLA?

giacomomagini:
I've signed it 🤔 I can't see the button anymore when i open the link.

giacomomagini:
Sorry, my global git config had my work email. I re-pushed with my private email in the commit.
Thanks for helping 😄

christian-bromann:
Perfect, thanks!

PR #6664, opened 2021-04-06
Fix Devtools hang on Android

### LINKED ISSUE #6663: Devtools service hanging when used with Android

 - **WebdriverIO version:** 7.3.0
 - **Mode:** WDIO Testrunner
 - **Node.js version:** 14.15.1
 - **NPM version:** 6.14.8
 - **Browser name and version:** Chrome 89.0.4389.105
 - **Platform name and version:** Android 10
   - @wdio/cli
   - @wdio/devtools-service
   - @wdio/local-runner
   - @wdio/mocha-framework
   - @wdio/sync
   - @wdio/wdio-spec-reporter

**Config of WebdriverIO**

```js
exports.config = {
  runner: "local",
  path: "/wd/hub",
  maxInstances: 10,
  maxInstancesPerCapability: 10,
  logLevel: "debug",
  bail: 1,
  baseUrl: "http://localhost",
  waitforTimeout: 20000,
  filesToWatch: [],
  connectionRetryTimeout: 15000,
  connectionRetryCount: 0,
  framework: "mocha",
  reporters: ["spec"],
  services: [
    "devtools",
  ],
  mochaOpts: {
    ui: "bdd",
    timeout: 60000,
    bail: true,
  },
};
```

**Describe the bug**

The Devtools service hangs at [this line](https://github.com/webdriverio/webdriverio/blob/v7.3.0/packages/wdio-devtools-service/src/index.ts#L231) when used with Android.

**To Reproduce**

We have prepared a repository that demonstrates the issue as precisely as possible :
https://github.com/nicolasbouffard/wdio-android-devtools

**Expected behavior**

We expect to be able to run the Devtools service with an Android device running Chrome juste like we can with a desktop Chrome browser.

**Log**

See the repository to reproduce the issue, as it contains more precise information than what could be provided with logs alone :
https://github.com/nicolasbouffard/wdio-android-devtools

nicolasbouffard:
## Proposed changes

This PR should close #6663. Please see the issue and the mentioned repository to reproduce the issue for more information.

**Disclaimer :** We do not believe that we have enough understanding of WebdriverIO's ins and outs to guarantee that the proposed changes are valid and don't introduce breaking changes. We simply wanted to share our researches with the community, in hope that a proper fix for this bug (or at least what looks like a bug to us) could be merged to WebdriverIO's codebase soon.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/packages/wdio-devtools-service/src/index.ts
+++ b/packages/wdio-devtools-service/src/index.ts
@@ -222,8 +222,7 @@ export default class DevToolsService implements Services.ServiceInstance {
 
         this._target = await this._puppeteer.waitForTarget(
             /* istanbul ignore next */
-            // @ts-ignore
-            (t) => t._targetInfo.browserContextId != null)
+            (t) => t['_targetInfo'].browserContextId)
         /* istanbul ignore next */
         if (!this._target) {
             throw new Error('No page target found')
--- a/packages/wdio-devtools-service/src/index.ts
+++ b/packages/wdio-devtools-service/src/index.ts
@@ -222,7 +222,7 @@ export default class DevToolsService implements Services.ServiceInstance {
 
         this._target = await this._puppeteer.waitForTarget(
             /* istanbul ignore next */
-            (t) => t['_targetInfo'].browserContextId)
+            (t) => t.type() === 'page' || t['_targetInfo'].browserContextId)
         /* istanbul ignore next */
         if (!this._target) {
             throw new Error('No page target found')
--- a/packages/wdio-devtools-service/src/index.ts
+++ b/packages/wdio-devtools-service/src/index.ts
@@ -222,7 +222,8 @@ export default class DevToolsService implements Services.ServiceInstance {
 
         this._target = await this._puppeteer.waitForTarget(
             /* istanbul ignore next */
-            (t) => t.type() === 'page')
+            // @ts-ignore
+            (t) => t._targetInfo.browserContextId != null)
         /* istanbul ignore next */
         if (!this._target) {
             throw new Error('No page target found')


packages/wdio-devtools-service/src/index.ts
```diff
@@ -222,7 +222,8 @@ export default class DevToolsService implements Services.ServiceInstance {
 
         this._target = await this._puppeteer.waitForTarget(
             /* istanbul ignore next */
-            (t) => t.type() === 'page')
+            // @ts-ignore
+            (t) => t._targetInfo.browserContextId != null)
```
christian-bromann, line 226: Would this allow us to remove the `@ts-ignore`?
```suggestion
            (t) => t['_targetInfo'].browserContextId)
```

christian-bromann (commented):
Thanks for your investigations into this issue. I have some concerns about the implications of this change. For one we access a private property (`_targetInfo`) that can change anytime, the other is whether this has the same behavior for desktop browser.

2 Questions:
- is it possible to do something like this:
  ```js
  (t) => t.type() === 'page' || t['_targetInfo'].browserContextId)
  ```
- is there a way to check for context id without accessing a private property?

nicolasbouffard:
> I have some concerns about the implications of this change.

I understand your concerns. Like I said, we aren't exactly sure of the implications of this change either, we just found that it fixed our issues and went with it, but we'd be much more confident if the community could look at this in more details.

> is there a way to check for context id without accessing a private property?

The `Target` class in Puppeteer's code seems to have two ways to access the browser context info :
1. Through the [`_targetInfo` property](https://github.com/puppeteer/puppeteer/blob/e34a6d53183c3e1f63a375ba6a26bee0dcfcf542/src/common/Target.ts#L28), which is the initial proposed solution. But there doesn't seem to be any public getter for it ;
2. Through the [`_browserContext` property](https://github.com/puppeteer/puppeteer/blob/e34a6d53183c3e1f63a375ba6a26bee0dcfcf542/src/common/Target.ts#L29), which has a [public getter](https://github.com/puppeteer/puppeteer/blob/e34a6d53183c3e1f63a375ba6a26bee0dcfcf542/src/common/Target.ts#L194). From there, we thought the [`_id` property](https://github.com/puppeteer/puppeteer/blob/e34a6d53183c3e1f63a375ba6a26bee0dcfcf542/src/common/Browser.ts#L614) would be equal to the `_targetInfo.browserContextId` property that we used before, but it doesn't seem to be the case since, according to our testings, when `_targetInfo.browserContextId` is different than `null`, `t.browserContext()._id` is still `null`.

> the other is whether this has the same behavior for desktop browser

For our use cases at least, this change didn't seem to impact the behavior of desktop browsers (we use WebdriverIO for both smartphone and desktop testing).

> is it possible to do something like this:
> ```js
> (t) => t.type() === 'page' || t['_targetInfo'].browserContextId)
> ```

I wasn't very optimistic about this because I thought the `t.type() === 'page'` part would match before the `t['_targetInfo'].browserContextId`, causing the issue to come back, but after one attempt it seems to work the same.

packages/wdio-devtools-service/src/index.ts
nicolasbouffard: This works just as well indeed.

christian-bromann:
> but after one attempt it seems to work the same.

So should we add my suggested change to the PR?

christian-bromann:
ping @nicolasbouffard

nicolasbouffard:
> ping @nicolasbouffard

Sorry for the delay, I just pushed the changes.

christian-bromann (approved):
👍  thank you!

Issue #6652, opened 2021-04-02
Multiremote config errors due to capabilities values - Illegal key values seen in w3c capabilities: [hostname, path, port, protocol]


amandaflagg:
**Environment (please complete the following information):**
 - **WebdriverIO version:** 7.3.0
 - **Mode:** WDIO Testrunner
 - **If WDIO Testrunner, running sync/async:** sync
 - **Node.js version:** 12.18.3
 - **NPM version:** 6.14.6
 - **Browser name and version:** Chrome 89.0.4389.114
 - **Platform name and version:** Mac OS Catalina 10.15.7
 - **Additional wdio packages used (if applicable):** 
 - @wdio/spec reporter, @wdio/selenium-standalone service, @wdio/mocha-framework, @wdio/sync, @wdio/local-runner

**Config of WebdriverIO**
```exports.config = {
    runner: 'local',
    specs: [
        './webdriver/tests/**.ts'
    ],
    exclude: [
        // 'path/to/excluded/files'
    ],
    autoCompileOpts: {
        autoCompile: true,
        // see https://github.com/TypeStrong/ts-node#cli-and-programmatic-options
        // for all available options
        tsNodeOpts: {
            transpileOnly: true,
            project: './webdriver/tsconfig.json'
        },
        // tsconfig-paths is only used if "tsConfigPathsOpts" are provided, if you
        // do please make sure "tsconfig-paths" is installed as dependency
        tsConfigPathsOpts: {
            baseUrl: './'
        }
    },
    maxInstances: 10,
    automationProtocol: 'webdriver',
    path: '/wd/hub',
    capabilities: 
    {
        chromeBrowser: {
            capabilities: {
                browserName: 'chrome'
            }
        },
        firefoxBrowser: {
            capabilities: {
                browserName: 'firefox'
            }
        }
    },
    logLevel: 'info',
    bail: 0,
    baseUrl: 'http://www.google.com/',
    waitforTimeout: 10000,
    connectionRetryTimeout: 120000,
    connectionRetryCount: 3,
    services: [
        ['selenium-standalone', { drivers: { firefox: '0.28.0', chrome: true, chromiumedge: 'latest' } }]
    ],
    framework: 'mocha',
    reporters: ['spec'],
    mochaOpts: {
        ui: 'bdd',
        timeout: 60000
    }
}
```

**Describe the bug**
[...]

PR #6681, opened 2021-04-09
Fix connection details setting for multiremote in selenium server

christian-bromann:
## Proposed changes

Fixes: #6652

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments


packages/wdio-selenium-standalone-service/src/launcher.ts
```diff
@@ -82,10 +82,35 @@ export default class SeleniumStandaloneLauncher {
...
+             * ```js
+             * capabilities: {
+             *   myBrowser: {
+             *     hostname: 'localhost',
+             *     capabilities: { browserName: 'chrome' }
+             *   }
+             * }
+             */
+            const remoteCapabilities = (cap as Capabilities.W3CCapabilities).alwaysMatch || cap
+            const objectToApplyConnectionDetails = Array.isArray(this._capabilities)
```
erwinheitzman, line 104: what bothers me a bit is that we already did a check for the type of capabilities at line 82, what do you think about something like this?
```
Array.isArray(this._capabilities) ? this._capabilities : Object.values(this._capabilities))

if (Array.isArray(this._capabilities)) {
  // logic for standard mode
} else {
  // logic for multiremote mode
}
```
It would prevent us from having multiple checks and is easy to follow when reading

christian-bromann:
@erwinheitzman I refined the types. I need to check twice because in the first time I need to decided whether to get the capabilities from a multiremote object vs. a set of caps and second I need to decide to attach the new options to the caps directly or to the multiremote options. I don't see a better way to do this.

wswebcreation (approved):
LGTM

PR #6682, opened 2021-04-09
Enhance docs for custom locator strategy

christian-bromann:
## Proposed changes

Clarify how to use the scoped element when using custom selector strategy.

## Types of changes

- [x] Documentation update

## Checklist

## Further comments

n/a

### DIFF
--- a/website/docs/Selectors.md
+++ b/website/docs/Selectors.md
@@ -506,7 +506,12 @@ If your app requires a specific way to fetch elements you can define yourself a 
 
 ```js
 browser.addLocatorStrategy('myCustomStrategy', (selector, root) => {
-    return root.querySelectorAll(selector)
+    /**
+     * scope should be document if called on browser object
+     * and `root` if called on an element object
+     */
+    const scope = root ? root : document
+    return scope.querySelectorAll(selector)
 })
 ```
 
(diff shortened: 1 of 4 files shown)


website/docs/Selectors.md
```diff
@@ -505,17 +505,28 @@ browser.react$$('MyComponent') // returns the WebdriverIO Elements for the array
 If your app requires a specific way to fetch elements you can define yourself a custom selector strategy that you can use with `custom$` and `custom$$`. For that register your strategy once in the beginning of the test:
 
 ```js
-browser.addLocatorStrategy('myCustomStrategy', (selector) => {
-    return document.querySelectorAll(selector)
+browser.addLocatorStrategy('myCustomStrategy', (selector, root) => {
```
erwinheitzman, line 508: Can't find anywhere in the code that this should work and when running this I also get a confirmation about this saying that root is undefined. Why is this change made?

erwinheitzman (changes requested):

website/docs/Selectors.md
erwinheitzman: Also, if this is indeed true. We should also update the documentation at all other places and also our types to reflect this as well

christian-bromann: > Can't find anywhere in the code that this should work and when running this I also get a confirmation about this saying that root is undefined. Why is this change made?

It only works when calling the custom locator strategy on an element object. The reference implementation is here: https://github.com/webdriverio/webdriverio/blob/main/packages/webdriverio/src/commands/element/custom$.ts#L37-L51

christian-bromann:
@erwinheitzman I improved the types and added a test.

> Also, if it would be true that root can be used here, we should also update the documentation at all other places and also our types to reflect this as well

I haven't seen an example where we called a custom locator strategy on an element in other places.

erwinheitzman (approved):

PR #6720, opened 2021-04-16
fix: Only log when the entry is not a string or array

seanpoulter:
## Proposed changes

There is a bug in the logic that causes us to always log instead of only when the `file` is not a string or array.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

Based on the size of the change I have not created an Issue. Please let me know if you'd like one.


christian-bromann (approved):
👍

PR #6709, opened 2021-04-16
Implement `>>>` (deep) selector for WebdriverIO

christian-bromann:
I just stumbled upon this amazing library again: https://github.com/Georgegriff/query-selector-shadow-dom and thing it is worth while to natively integrate it into WebdriverIO. My idea would be to add a "deep" selector that would work like this:

```js
$('>>>.btn-in-shadow-dom')
```

What do you think @Georgegriff ?

### DIFF
--- a/@types/query-selector-shadow-dom.d.ts
+++ b/@types/query-selector-shadow-dom.d.ts
@@ -1,0 +1,30 @@
+const ExecuteMethod = (element: Element | Document, selector: string) => Element | null;
+
+/**
+ * Contains two functions `queryOne` and `queryAll` that can
+ * be {@link Puppeteer.registerCustomQueryHandler | registered}
+ * as alternative querying strategies. The functions `queryOne` and `queryAll`
+ * are executed in the page context.  `queryOne` should take an `Element` and a
+ * selector string as argument and return a single `Element` or `null` if no
+ * element is found. `queryAll` takes the same arguments but should instead
+ * return a `NodeListOf<Element>` or `Array<Element>` with all the elements
+ * that match the given query selector.
+ * @public
+ */
+interface CustomQueryHandler {
+    queryOne?: ExecuteMethod
+    queryAll?: (element: Element | Document, selector: string) => Element[] | NodeListOf<Element>;
+}
+
+declare const querySelectorShadowDom: {
+    QueryHandler: CustomQueryHandler
+    locatorStrategy: ExecuteMethod
+}
+
+declare module "query-selector-shadow-dom/plugins/puppeteer" {
+    export = querySelectorShadowDom
+}
+
+declare module "query-selector-shadow-dom/plugins/webdriverio" {
+    export = querySelectorShadowDom
+}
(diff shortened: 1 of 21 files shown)


christian-bromann:
@Georgegriff took a stab at the implementation of it. What do you think? The script example is now:
```js
const { remote } = require('./packages/webdriverio')

;(async () => {
        capabilities: {
            browserName: 'chrome'
        }
    })

    await browser.url('chrome://downloads')
    const moreActions = await browser.$('>>>#moreActions')
    await moreActions.click()
    const span = await browser.$('>>>.dropdown-item:not([hidden])')
    const text = await span.getText()
    // prints `Open downloads folder`
    console.log(text)

    await browser.deleteSession()
})().catch((e) => console.error(e))
```

I wonder if a `>>>` prefix makes sense or a different one is better, what are your thoughts? I am happy to update the readme in your project to simplify the example and clarify that this feature is built in.

Great work on the plugin btw., it works really well and was super easy to integrate!

erwinheitzman:
that's awesome!

I do see a possible improvement when removing the `>>>` prefix and instead assigging it's own command so that it can be chained like so to allow for more flexibility `$('root-elem').$('sub-elem').shadow('.shadow-elem')` and maybe chain even further with more usage of `$`, `$$` or `shadow`. What do you think?

gkushang:
Indeed, a much needed support ❤️. This library works great for us. I recently have committed the approach of using @Georgegriff [query-selector-shadow-dom](https://github.com/Georgegriff/query-selector-shadow-dom/tree/d0a6384615896f9dac9bc6c6d5aa5d81ef61811e) in CodeceptJS with ****WebdriverIO**, through CustomLocatoryStrategy.** 

We, @salesforce, building apps in WebComponents (LWC) and this solution was must required. 

More info: https://engineering.salesforce.com/automation-made-easy-for-web-components-shadow-roots-b23016779ce8

Working Solution: https://github.com/salesforce/codeceptjs-bdd/tree/develop/examples/webcomponents-playwright-webdriver-example#webcomponents-playwright--webdriver-e2e-example

@christian-bromann : There are some [limitations](https://github.com/Georgegriff/query-selector-shadow-dom/tree/d0a6384615896f9dac9bc6c6d5aa5d81ef61811e#known-issues) with this Library, and would be great if WebdriverIO overcomes with it thru the Native approach,

- Safari does not work (the major limitations)

christian-bromann:
> I do see a possible improvement when removing the `>>>` prefix and instead assigning it's own command so that it can be chained like so to allow for more flexibility `$('root-elem').$('sub-elem').shadow('.shadow-elem')` and maybe chain even further with more usage of `$`, `$$` or `shadow`. What do you think?

I don't see a reason why we need to introduce a new selector command for this. You can achieve the same chaining with integrating it into our main selector command, e.g. `$('root-elem').$('sub-elem').$('>>>.shadow-elem')`

> * Safari does not work (the major limitations)

I can take a look but doubt that we can overcome this limitation by integrating the lib into WebdriverIO.

christian-bromann:
@webdriverio/project-committers added tests and updated docs, this is ready for review now.

erwinheitzman (approved):

Issue #6724, opened 2021-04-19
Bug DevToolsDriver removes all cached elements when page loads new frames


kyryloonufriiev:
**Environment (please complete the following information):**
 - **WebdriverIO version:** 7.4.6
 - **Mode:** standalone
 - **Node.js version:** 12.20.2
 - **NPM version:** 6.14.11
 - **Browser name and version:** Chrome 89
 - **Platform name and version:** MacOS 11.2.3
 - **Additional wdio packages used (if applicable):** devtools

**Config of WebdriverIO**
`automationPrococol: 'devtools'`

**Describe the bug**
I have an issue with `devtools` package. It creates elements cache with fake ids and saves elements in an element store. But in order to clear unused references, a wrong strategy was chosen. The implementation listens to `framenavigated` event from page object and clear element store. But this event will be fired when any of the iframes on the page will be reloaded or a new frame will be added. In my opinion, this is wrong behavior, if you want to clear element references when the page itself is reloaded/navigated then before clearing (in `framenavigatedHandler` method) you should check that navigated frame (it will be the first argument of this method) is the main one (e.g. `!navigatedFrame.parentFrame()`). The superior solution will store elements together with the frame object reference, and remove elements only for the specific frame that was navigated.

**To Reproduce**
1. Find element
2. Run script like this
```js
browser.execute(() => document.head.append(document.createElement('iframe')))
```
3. Try to use the element found in step 1. (e.g. `browser.execute(el => el, el_from_step_1)`)

**Expected behavior**
The element should not be removed from the cache.

christian-bromann:
Thanks for reporting @kyryloonufriiev . Any contributions would be appreciated.

christian-bromann:
fixed in #6724

PR #6725, opened 2021-04-19
Fix DevToolsDriver elements storing

kyryloonufriiev:
## Proposed changes

In this PR I propose a fix for a bug in `DevToolsDriver` that was described in this [issue](https://github.com/webdriverio/webdriverio/issues/6724).
I propose to store frames map of type `Map<Frame, string>` where value is generated element ids. And remove the only elements of a specific frame if this frame was reloaded/navigated, instead of completely clear the whole elements store.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/packages/devtools/src/devtoolsdriver.ts
+++ b/packages/devtools/src/devtoolsdriver.ts
@@ -152,11 +152,7 @@ export default class DevToolsDriver {
 
     framenavigatedHandler(frame: Frame) {
         this.currentFrameUrl = frame.url()
-        if (!frame.parentFrame()) {
-            this.elementStore.clear()
-        } else {
-            this.elementStore.clear(frame)
-        }
+        this.elementStore.clear(frame.parentFrame() ? frame : undefined)
     }
 
     setTimeouts(implicit?: number, pageLoad?: number, script?: number) {
--- a/packages/devtools/src/devtoolsdriver.ts
+++ b/packages/devtools/src/devtoolsdriver.ts
@@ -7,6 +7,7 @@ import type { Browser } from 'puppeteer-core/lib/cjs/puppeteer/common/Browser'
 import type { Dialog } from 'puppeteer-core/lib/cjs/puppeteer/common/Dialog'
 import type { Page } from 'puppeteer-core/lib/cjs/puppeteer/common/Page'
 import type { CommandEndpoint } from '@wdio/protocols'
+import type { Frame } from 'puppeteer-core/lib/cjs/puppeteer/common/FrameManager'
 
 import ElementStore from './elementstore'
 import { validate, sanitizeError } from './utils'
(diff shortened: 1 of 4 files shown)


packages/devtools/src/devtoolsdriver.ts
```diff
@@ -149,9 +150,13 @@ export default class DevToolsDriver {
...
 
-    framenavigatedHandler(frame: Page) {
+    framenavigatedHandler(frame: Frame) {
         this.currentFrameUrl = frame.url()
-        this.elementStore.clear()
+        if (!frame.parentFrame()) {
+            this.elementStore.clear()
+        } else {
+            this.elementStore.clear(frame)
+        }
```
christian-bromann, line 159: How about simplifying it into
```suggestion
        this.elementStore.clear(frame.parentFrame() ? frame : undefined)
```

packages/devtools/src/elementstore.ts
```diff
@@ -1,12 +1,23 @@
...
 export default class ElementStore {
     private _index = 0
     private _elementMap: Map<string, ElementHandle> = new Map()
+    private _frameMap: Map<Frame, Set<string>> = new Map()
 
     set (elementHandle: ElementHandle) {
         const index = `ELEMENT-${++this._index}`
         this._elementMap.set(index, elementHandle)
+        const frame = elementHandle.executionContext().frame()
+        if (frame) {
```
christian-bromann, line 13: We usually always expect a frame right? What would be a scenario where frame is null here? I wonder if it makes sense to better throw in this case rather than silently ignoring this.

packages/devtools/src/elementstore.ts
```diff
@@ -25,7 +36,16 @@ export default class ElementStore {
...
+        if (!frame) {
+            this._elementMap.clear()
+            this._frameMap.clear()
+        } else {
+            const elementIndexes = this._frameMap.get(frame)
+            if (elementIndexes) {
+                elementIndexes.forEach((elementIndex) => this._elementMap.delete(elementIndex))
+                this._frameMap.delete(frame)
+            }
+        }
```
christian-bromann, line 49: This is just personal preference, feel free to ignore but
```suggestion
        if (!frame) {
            this._elementMap.clear()
            this._frameMap.clear()
            return
        }

        const elementIndexes = this._frameMap.get(frame)
        if (elementIndexes) {
            elementIndexes.forEach((elementIndex) => this._elementMap.delete(elementIndex))
            this._frameMap.delete(frame)
        }
```

packages/devtools/src/elementstore.ts
kyryloonufriiev: Yeah, a frame should be there in all of the cases. We are extracting the `ExecutionContext` of an element, which is always an `ExecutionContext` of the element owner frame. `ExecutionContex` of the frame will be always associated with this frame [pptr doc](https://pptr.dev/#?product=Puppeteer&version=v8.0.0&show=api-executioncontextframe). So I'm not sure which kind of error I could possibly throw here, because even if an element will not have a frame for some reason it just will be preserved in store until the whole page will be reloaded/navigated.

christian-bromann (approved):
Thanks for the quick fix!

Issue #6760, opened 2021-04-26
Can't change window size for the multiremote mode


kongory:
**Environment (please complete the following information):**
 - **WebdriverIO version:** 7.7.
 - **Mode:** WDIO Testrunner
 - **If WDIO Testrunner, running sync/async:** async
 - **Node.js version:** 14.16.0
 - **NPM version:** 6.14.11
 - **Browser name and version:** Chrome 90
 - **Platform name and version:** Windows 10
 - **Additional wdio packages used (if applicable):** 

**Config of WebdriverIO**

**Describe the bug**
 Can't change sizes of multiremote instances in test.

**To Reproduce**
Have such configuring in the 'it':
```
const browser = await multiremote({
      userA: {
        capabilities: {
          browserName: "chrome",
          chromeOptions: {
            args: ['window-size=1920,1080']
        }
        },
      },
      userB: {
        capabilities: {
          browserName: "chrome",
          chromeOptions: {
            args: ['window-size=1920,1080']
        }
        },
      },
    });
```
When I use
`await browser["userA"].setWindowSize(1920, 1080);`
have an error: _"Cannot assign to read only property '_NOT_FIBER' of object '#<MultiRemoteDriver>'"_, but with
```
await browser["userA"].setWindowSize(1920, 1080);
await browser["userB"].setWindowSize(1920, 1080);
```
nothing happens.

**Expected behavior**
Sizes are changed.

**Log**
[Logs](https://gist.github.com/kongory/9af17a86ad2c08df5dd8a15a51a00bc5.js)

christian-bromann:
It seems that when calling the `setWindowSize` we loose the scope of the browser instance. This error likely happens somewhere where we [attach the browser to the main instance](https://github.com/webdriverio/webdriverio/blob/main/packages/webdriverio/src/multiremote.ts#L55). Any contributions on this would be appreciated.

elaichenkov:
Hey @kongory 

Can you please provide a fully reproducible example and logs?

christian-bromann:
@elaichenkov using the [multiremote example](https://github.com/webdriverio/webdriverio/tree/main/examples/wdio/multiremote) allowed me to reproduce it by replacing everything with:
```js
describe('multiremote example', () => {
    it('should open chat application', () => {
        browser.url('https://socketio-chat-h9jt.herokuapp.com/')
        browser["browserA"].setWindowSize(1920, 1080)
        browser.pause(5000)
    })
})
```

[...]

PR #6761, opened 2021-04-26
fix methods for the multiremote

elaichenkov:
## Proposed changes

bugfix: #6760 

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/packages/webdriverio/src/multiremote.ts
+++ b/packages/webdriverio/src/multiremote.ts
@@ -37,7 +37,8 @@ export default class MultiRemote {
         for (const commandName of wrapperClient.commandList) {
             propertiesObject[commandName] = {
                 value: this.commandWrapper(commandName),
-                configurable: true
+                configurable: true,
+                writable: true
             }
         }
 
--- a/tests/multiremote/test.js
+++ b/tests/multiremote/test.js
@@ -9,6 +9,12 @@ describe('smoke test multiremote', () => {
         assert.equal(browser.browserA.getTitle(), 'Mock Page Title')
     })
 
+    it('should return async value', async () => {
+        assert.deepEqual(await browser.getUrl(), ['https://mymockpage.com', 'https://mymockpage.com'])
+        assert.equal(await browser.browserA.getUrl(), 'https://mymockpage.com')
+        assert.equal(await browser.browserB.getUrl(), 'https://mymockpage.com')
+    })
+
     describe('add customCommands', () => {
         it('should respect promises', () => {
             browser.addCommand('customFn', () => {


christian-bromann:
Thanks for raising the PR 👏 

I suggest to add a smoke test to https://github.com/webdriverio/webdriverio/blob/main/tests/multiremote/test.js where a command is called via `browser.<browserName>.<command>()` so we can ensure it works properly. You can read more about our smoke tests in the [contribution guidelines](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#run-e2e-experience-with-smoke-tests). Let me know if you have any questions.

elaichenkov:
I've just pushed one test, but again I'm not sure that's what we actually need for that case :(
So, feel free to modify it :)

christian-bromann:
@elaichenkov the smoke test the way you put it makes sense. However when I reviewed the change as a whole it turns out that it already got fixed by this patch: #6759. Can you double check? In this case I would suggest to not merge given that adding `writeable` is not really necessary.

elaichenkov:
I've just tested it with the patch you've mentioned, but unfortunately, the problem still exists with the same message: we cannot assign it to read-only property. After adding the `writable` property the issue seems to be disappeared.

christian-bromann:
@elaichenkov you are right, the only problem is that the smoke test also passes without the change. So running the example using the example wdio script fails but running the multiremote smoke test passes just fine.

christian-bromann:
A bit of investigation showed that `_NOT_FIBER` is where the error is thrown in the passing case `undefined` and in the failing case a `[Function: wrapCommandFn]`

elaichenkov:
> @elaichenkov you are right, the only problem is that the smoke test also passes without the change. So running the example using the example wdio script fails but running the multiremote smoke test passes just fine.

Totally agree with you.


> A bit of investigation showed that `_NOT_FIBER` is where the error is thrown in the passing case `undefined` and in the failing case a `[Function: wrapCommandFn]`

Well, sorry for the silly question but what does it mean for us?

christian-bromann:
> Well, sorry for the silly question but what does it mean for us?

Sorry for not being clear, I did a `console.log(this._NOT_FIBER)` right before the error is thrown and when running with the example files (where it fails) I get `[Function: wrapCommandFn]` and when running using the smoke tests (and where it passes) I get `undefined`. This confuses me because both things run with the same setup (wdio testrunner + `@wdio/sync`). While I am not too much concerned about this it would be great if the multiremote test could help us validate this fix.

elaichenkov:
Oh, I see. 
Thank you for explaining. 
I'll try to explore it and conduct a root cause.

konstantin-gorbachev-Simpligov:
New updates. Also when I try use `await browser.url("/")` in async mode. Have the error: **Cannot assign to read only property '_NOT_FIBER' of object '#<MultiRemoteDriver>'**

christian-bromann:
What is the status of this? Anything I can do to help move this forward?

elaichenkov:
Sorry, I haven't gotten a chance to take a look at this yet.

christian-bromann:
Closing due to inactivity. I believe this is fixed now.

PR #6773, opened 2021-04-28
Fix incorrect typing of $(this:...) function

### LINKED ISSUE #6772: Incorrect typing of $(this:...) function
**Describe the bug**
You cannot create correct typings for custom wrappers like `testID$(selector:string, context: WebdriverIO.Browser | WebdriverIO.Element)` and use them through `browser.addCommand(...)`.

**To Reproduce**
Example `testID` helper function:
```
export const testID$ = async (
    testID: string,
    context: WebdriverIO.Browser | WebdriverIO.Element = browser
): Promise<WebdriverIO.Element> => await context.$(`~${testID}`);
```

See https://github.com/webdriverio/webdriverio/blob/main/packages/webdriverio/src/commands/browser/%24.ts#L61

It should have same types as 
https://github.com/webdriverio/webdriverio/blob/main/packages/webdriverio/src/commands/browser/%24%24.ts#L53

**Expected behavior**
`$` to have same type for `this` as `$$`

vitalyiegorov:
closes #6772

### DIFF
--- a/packages/webdriverio/src/commands/browser/$.ts
+++ b/packages/webdriverio/src/commands/browser/$.ts
@@ -58,7 +58,7 @@ import type { Selector } from '../../types'
  *
  */
 export default async function $ (
-    this: WebdriverIO.Browser,
+    this: WebdriverIO.Browser | WebdriverIO.Element,
     selector: Selector
 ) {
     /**


christian-bromann:
I don't think this is the correct way of solving this. We should consolidate the commands similar how we did this with $$. It seems that the implementations for $ for the browser as well as the element are the same.

christian-bromann:
fixed in #6810

Issue #6765, opened 2021-04-26
Wdio CLI Wizard Improvement


Askarov5:
**Is your feature request related to a problem? Please describe.**
1. We(in the company) would like to create several webdriverio projects with the same configurations(to keep standards). Each time when a QA creates a new wdio project, he needs to ask which options to select on wdio CLI.

2. Building UI on top of webdriverio project, and I would like to be able to create a new project by selecting options(between available reporters and services, local or cloud etc) and by clicking on the create button.

**Describe the solution you'd like**
Improve wdio CLI Installation wizard by:
   1. Adding parameter such as --file where we can set all needed parameters then create the project based on that. It will be easy to share between the team members. For example: `npx wdio config --file /project-settings.js` where project-settings file(js ro json) contains object like following: `{
   location: 'local',
   framework: 'mocha',
   sync: true,
   featureFileLocation: '/features',
   autoGenerateTests: true,
   usePom: true,
  useCompiler: false,
  reporters:['spec', 'HtmlReporter'],
  services: ['chromedriver'],
  baseUrl: 'https://github.com/webdriverio/webdriverio/issues'
   }`

  2.  adding parameters for each option mentioned above. For example: `npx wdio config --framework mocha --location local` etc

erwinheitzman:
When using your suggestion, mocha would be installed but you'd have no control over the version installed. Isn't this something you'd want to manage in say, a private NPM package instead of the suggestion above so that you not only have control over the configuration but also the package versions?

Don't get me wrong, I think it's a good feature to implement but I don't see how it will help your usecase specifically so I'd like to hear from you if you do care about the version that is being installed or not.

Nonetheless we can definitely implement this

Askarov5:
For now, we don't need npm package version control. It will be enough if it works the same way as installation CLI Q-A. As far as I know, when we go through them, it will install latest/compatible versions. 
I believe that Simple solution will be better for QA team members(including less technical and new QA engineers) and It will be easy to share in the team without any additional actions like registration to npm or GitHub website or giving them access. 

Perfect solution will be having a template(like package.json) file with all dependencies, scripts, wdio specific options such as "usePOM", "useCompiler", "localOrCloud" etc. where you can create a new project identical to the previous one.

christian-bromann:
What I've seen in many companies is that they have their own configuration package that comes with smart logic around company specific infrastructure, e.g.:

```ts
import companyFooConfigHelper from '@company/wdio-config'
[...]

PR #6778, opened 2021-04-28
Wdio cli add new feature 6765

erwinheitzman:
## Proposed changes

Implement new feature #6765 

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/wdio-cli/src/commands/config.ts
+++ b/packages/wdio-cli/src/commands/config.ts
@@ -11,7 +11,7 @@ import {
     addServiceDeps, convertPackageHashToObject, renderConfigurationFile,
     hasFile, generateTestFiles, getAnswers, getPathForFileGeneration
 } from '../utils'
-import { ConfigCommandArguments, ParsedAnswers } from '../types'
+import { ConfigCommandArguments, Framework, ParsedAnswers } from '../types'
 import yargs from 'yargs'
 
 const pkg = require('../../package.json')
@@ -40,9 +40,9 @@ export const builder = (yargs: yargs.Argv) => {
         .help()
 }
 
-const runConfig = async function (useYarn: boolean, yes: boolean, exit = false) {
+const runConfig = async function (useYarn: boolean, yes: boolean, framework: Framework, exit = false) {
     console.log(CONFIG_HELPER_INTRO)
-    const answers = await getAnswers(yes)
+    const answers = await getAnswers(yes, framework)
     const frameworkPackage = convertPackageHashToObject(answers.framework)
     const runnerPackage = convertPackageHashToObject(answers.runner || SUPPORTED_PACKAGES.runner[0].value)
     const servicePackages = answers.services.map((service) => convertPackageHashToObject(service))
@@ -192,7 +192,11 @@ const runConfig = async function (useYarn: boolean, yes: boolean, exit = false) 
 }
 
 export function handler(argv: ConfigCommandArguments) {
-    return runConfig(argv.yarn, argv.yes)
+    if (argv.framework && !SUPPORTED_PACKAGES.framework.find(({ name }) => name === argv.framework)) {
+        throw new Error(`InvalidArgumentError: Framework ${argv.framework} is not supported.`)
+    }
+
+    return runConfig(argv.yarn, argv.yes, argv.framework)
 }
 
 /**
--- a/packages/wdio-cli/src/types.ts
+++ b/packages/wdio-cli/src/types.ts
@@ -90,9 +90,12 @@ export interface InstallCommandArguments {
     name: string
 }
 
+export type Framework = 'mocha' | 'jasmine' | 'cucumber'
+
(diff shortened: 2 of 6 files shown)


packages/wdio-cli/src/types.ts
```diff
@@ -90,9 +90,12 @@ export interface InstallCommandArguments {
     name: string
 }
 
+export type Framework = 'mocha' | 'jasmine' | 'cucumber'
```
christian-bromann, line 93: This should be moved into `@wdio/types` and applied in other areas where we type `framework`

packages/wdio-cli/src/commands/config.ts
```diff
@@ -192,7 +192,11 @@ const runConfig = async function (useYarn: boolean, yes: boolean, exit = false)
 }
 
 export function handler(argv: ConfigCommandArguments) {
-    return runConfig(argv.yarn, argv.yes)
+    if (argv.framework && !SUPPORTED_PACKAGES.framework.find(({ name }) => name === argv.framework)) {
+        throw new Error(`InvalidArgumentError: Framework ${argv.framework} is not supported.`)
+    }
+
+    return runConfig(argv.yarn, argv.yes, argv.framework)
```
christian-bromann, line 199: Given that we essentially want to allow to provide answers or all questions I suggest to pass in an object rather than defining individual answers. I believe we can re-use the `QUESTIONNAIRE` and and loop over this.

christian-bromann (changes requested):
Rather then defining the behavior for one parameter I would like to have this generalised based on the `QUESTIONNAIRE` we pass into `inquirer`. Otherwise this would be difficult to maintain.

packages/wdio-cli/src/commands/config.ts
```diff
@@ -191,8 +192,41 @@ const runConfig = async function (useYarn: boolean, yes: boolean, exit = false)
...
+        if (typeof port === 'undefined') {
+            return true
+        }
+
+        const num = Number(port)
+        if (Number.isInteger(num) && num > 0 && num < 65535) {
+            return true
+        }
+
+        return false
```
christian-bromann, line 218: Let's move these into a utils directory/file

packages/wdio-types/src/Services.ts
```diff
@@ -249,3 +249,10 @@ export interface HookFunctions {
         specs: string[]
     ): void;
 }
+
+export type Service = 'chromedriver' | 'sauce' | 'testingbot' | 'selenium-standalone' |
+    'devtools' | 'applitools' | 'browserstack' | 'appium' | 'firefox-profile' |
+    'crossbrowsertesting' | 'lambdatest' | 'zafira-listener' | 'reportportal' | 'docker' |
+    'wdio-ui5' | 'wiremock' | 'ng-apimock' | 'slack' | 'intercept' | 'docker' |
+    'visual-regression-testing' | 'novus-visual-regression' | 'rerun' | 'winappdriver' |
+    'ywinappdriver' | 'performancetotal' | 'aws-device-farm' | 'ocr-native-apps' | 'wait-for'
```
christian-bromann, line 258: I am not sure if it is a good idea to have such strict type definitions for services given that we have to clearly communicate it when people build their services. I would say for the purpose of this PR this is not necessary.

packages/wdio-types/src/Reporters.ts
```diff
@@ -50,3 +50,7 @@ export type ReporterEntry = (
      */
     [ReporterClass, WebdriverIO.ReporterOption]
 )
+
+export type Reporter = 'spec' | 'dot' | 'junit' | 'allure' | 'sumologic' |
```
christian-bromann, line 54: same

packages/wdio-types/src/Frameworks.ts
```diff
@@ -52,3 +52,5 @@ export interface World {
         message?: string
     }
 }
+
+export type Framework = 'mocha' | 'jasmine' | 'cucumber'
```
christian-bromann, line 56: same

christian-bromann (changes requested):
Some comments, also tests are still failing.

packages/wdio-types/src/Services.ts
erwinheitzman: That is only the case when they want to add it to the cli though, is it really a bad thing to have types so that it is correctly added to the CLI? I'd say if anything, it would make it more clear.

christian-bromann: typing things is generally good but this is not maintainable, the list will get quickly out of sync. Therefor I rather would but `string` as service.

christian-bromann (commented):
Anything I can do to help move this forward?

christian-bromann:
ping @erwinheitzman

christian-bromann:
Given this has been stale for over a year now I will close it as a lot of code has changed since then and it would make more sense to re-open the same PR if desired on the new code base.

PR #6788, opened 2021-05-04
Document package exports for `webdriverio`, `webdriver` and `devtools` package

christian-bromann:
The `webdriverio`, `webdriver` and `devtools` packages all export functions to start a session, attach to one and start a multiremote session. These functions are documented nowhere. I suggest to extend https://webdriver.io/docs/setuptypes to include some information on parameter types and their behavior.


TuHuynhVan (approved):
lgtm!

PR #6837, opened 2021-05-07
Added parent reference to wdio reporter suite and test stats

abhinaba-ghosh:
To build a custom reporter, often it is necessary to track the parent references for nested test structure. This helps in the logical grouping or filtration of reporting stats.


christian-bromann (approved):
👍

PR #6848, opened 2021-05-11
Fix loop infinite loop with $$(....)[xx].$(...) when the element is not found

### LINKED ISSUE #6798: Infinite loop and memory crash with $$(....)[xx].$(...) when the element is not found
*Environment (please complete the following information):**
 - **WebdriverIO version:** 7.5.2
 - **Mode:** WDIO
 - **Node.js version:** 14.16
 - **NPM version:** 6.14.11
 - **Browser name and version:** Any
 - **Platform name and version:**  Windows 10
 - **Additional wdio packages used (if applicable):** "@wdio/cli": "^7.5.2",
    "@wdio/cucumber-framework": "^7.5.2",
    "@wdio/dot-reporter": "^7.5.2",
    "@wdio/local-runner": "^7.5.2",
    "@wdio/mocha-framework": "^7.5.2",
    "@wdio/spec-reporter": "^7.5.2",
    "@wdio/sync": "^7.5.2",
    "@wdio/allure-reporter": "^7.5.2",
     "expect": "^26.6.2",
    "expect-webdriverio": "^2.0.1",

**Describe the bug**
When I try to select an element in a list of elements, if the element is not found and its position is not at the beginning of the list. I have a infinite loop and a memory crash

**To Reproduce**
https://webdriver.io/docs/api/browser/$$

```
it('should return false', () => {
  browser.url('https://webdriver.io/docs/api');
  const list = $$('.menu__list-item');
  console.log(list.length); // 107
  console.log(list[50].$('[icon="me"]').isDisplayed()); // Expected false
});
```

if I select the second element of the list, that works:
```
it('should return false', () => {
  browser.url('https://webdriver.io/docs/api');
  const list = $$('.menu__list-item');
  console.log(list.length); // 107
  console.log(list[2].$('[icon="me"]').isDisplayed()); // false is displayed
});
```

**Expected behavior**
The result should be displayed 

**wdio.config file:**
```
const expect = require('expect');
require('expect-webdriverio').setOptions({ wait: 5000 });
waitforTimeout: 5000,
 waitforInterval: 5000,
  //
  // Default timeout in milliseconds for request
  // if Selenium Grid doesn't send response
  connectionRetryTimeout: 240000,
  //
  // Default request retries count
  connectionRetryCount: 0,

 framework: 'cucumber',

  reporters: ['spec',
    ['allure', {
      outputDir: 'report/allure-results',
      disableWebdriverStepsReporting: false,
      disableWebdriverScreenshotsReporting: false,
      useCucumberStepReporter: true,
    }]],
  cucumberOpts: {
    // <boolean> show full backtrace for errors
    backtrace: false,
    // <string[]> module used for processing required features
    requireModule: ['@babel/register'],
    // <boolean< Treat ambiguous definitions as errors
    failAmbiguousDefinitions: true,
    // <boolean> invoke formatters without executing steps
    // dryRun: false,
    // <boolean> abort the run on first failure
    failFast: false,
    // <boolean> Enable this config to treat undefined definitions as
    // warnings
    ignoreUndefinedDefinitions: false,
    // <string[]> ("extension:module") require files with the given
    // EXTENSION after requiring MODULE (repeatable)
    name: [],
    // <boolean> hide step definition snippets for pending steps
    snippets: true,
    // <boolean> hide source uris
    source: true,
    // <string[]> (name) specify the profile to use
    profile: [],
    // <string[]> (file/dir) require files before executing features
    require: [
      './src/steps/given.js',
      './src/steps/then.js',
      './src/steps/when.js',
      './src/steps/after.js',
      // Or search a (sub)folder for JS files with a wildcard
      // works since version 1.1 of the wdio-cucumber-framework
      // './src/**/*.js',
    ],
    // <string> specify a custom snippet syntax
    snippetSyntax: undefined,
    // <boolean> fail if there are any undefined or pending steps
    strict: true,
    // <string> (expression) only execute the features or scenarios with
    // tags matching the expression, see
    // https://docs.cucumber.io/tag-expressions/
    tagExpression: 'not @Pending',
    // <boolean> add cucumber tags to feature or scenario name
    tagsInTitle: true,
    // <number> timeout for step definitions
    timeout: 99999999,
  },
```

L0tso:
## Proposed changes
Fix #6798 

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/webdriverio/src/middlewares.ts
+++ b/packages/webdriverio/src/middlewares.ts
@@ -2,6 +2,7 @@
 import refetchElement from './utils/refetchElement'
 import implicitWait from './utils/implicitWait'
 import { ELEMENT_KEY } from './constants'
+import { getBrowserObject } from './utils'
 import type { Capabilities } from '@wdio/types'
 
 /**
@@ -25,12 +26,8 @@ export const elementErrorHandler = (fn: Function) => (commandName: string, comma
                  * as `stale element reference`
                  */
                 if (result && result.error === 'no such element') {
-                    let currentElement = element
-                    //Crawl back to the browser object
-                    while (currentElement.elementId && currentElement.parent) {
-                        currentElement = currentElement.parent as WebdriverIO.Element
-                    }
-                    const capabilities = currentElement?.capabilities as Capabilities.Capabilities
+                    const browser = getBrowserObject(element)
+                    const capabilities = browser?.capabilities as Capabilities.Capabilities
                     if (capabilities?.browserName?.includes('safari')) {
                         const err = new Error()
                         err.name = 'stale element reference'
--- a/packages/webdriverio/src/middlewares.ts
+++ b/packages/webdriverio/src/middlewares.ts
@@ -24,9 +24,16 @@ export const elementErrorHandler = (fn: Function) => (commandName: string, comma
                  * as `stale element reference`
                  */
                 if (result && result.error === 'no such element') {
-                    const err = new Error()
-                    err.name = 'stale element reference'
-                    throw err
+                    let currentElement = element
+                    //Crawl back to the browser object
+                    while (currentElement.elementId && currentElement.parent) {
+                        currentElement = currentElement.parent as WebdriverIO.Element
+                    }
+                    if (currentElement?.capabilities?.browserName?.includes('safari')) {
+                        const err = new Error()
+                        err.name = 'stale element reference'
+                        throw err
+                    }
                 }
 
                 return result
--- a/packages/webdriverio/src/middlewares.ts
+++ b/packages/webdriverio/src/middlewares.ts
@@ -2,6 +2,7 @@
 import refetchElement from './utils/refetchElement'
 import implicitWait from './utils/implicitWait'
 import { ELEMENT_KEY } from './constants'
+import type { Capabilities } from '@wdio/types'
 
 /**
  * This method is an command wrapper for elements that checks if a command is called
@@ -29,7 +30,8 @@ export const elementErrorHandler = (fn: Function) => (commandName: string, comma
                     while (currentElement.elementId && currentElement.parent) {
                         currentElement = currentElement.parent as WebdriverIO.Element
                     }
-                    if (currentElement?.capabilities?.browserName?.includes('safari')) {
+                    const capabilities = currentElement?.capabilities as Capabilities.Capabilities
+                    if (capabilities?.browserName?.includes('safari')) {
                         const err = new Error()
                         err.name = 'stale element reference'
                         throw err
--- a/packages/webdriverio/tests/middleware.test.ts
+++ b/packages/webdriverio/tests/middleware.test.ts
@@ -72,6 +72,13 @@ describe('middleware', () => {
     })
 
     it('should successfully getAttribute of an element that falls stale after being re-found in Safari', async () => {
+        browser = await remote({
+            capabilities: {
+                browserName: 'safari',
+                // @ts-ignore mock feature
+                keepBrowserName: true
+            }
+        })
         const elem = await browser.$('#foo')
         elem.selector = '#nonexisting'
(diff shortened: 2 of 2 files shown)


packages/webdriverio/src/middlewares.ts
```diff
@@ -24,9 +25,17 @@ export const elementErrorHandler = (fn: Function) => (commandName: string, comma
...
                  */
                 if (result && result.error === 'no such element') {
-                    const err = new Error()
-                    err.name = 'stale element reference'
-                    throw err
+                    let currentElement = element
+                    //Crawl back to the browser object
+                    while (currentElement.elementId && currentElement.parent) {
+                        currentElement = currentElement.parent as WebdriverIO.Element
+                    }
```
christian-bromann, line 32: There is already a utility method for that: https://github.com/webdriverio/webdriverio/blob/main/packages/webdriverio/src/utils/index.ts#L94-L100

packages/webdriverio/src/middlewares.ts
```diff
@@ -24,9 +25,17 @@ export const elementErrorHandler = (fn: Function) => (commandName: string, comma
...
+                    const capabilities = currentElement?.capabilities as Capabilities.Capabilities
+                    if (capabilities?.browserName?.includes('safari')) {
```
christian-bromann, line 34: Why do we only do this for Safari browser? We should rather look into the protocol response and see how it differs from other browser to avoid have browser specific logic everywhere.

christian-bromann:
> Do you have any suggestions how to solve this problem properly?

I don't think this was a Safari specific hack. If this infinite loop only happens in Safari we should check how the WebDriver responses differ from a Safari session vs a Chrome session and modify/apply patches to specific protocol events rather than depending on a browser name.

packages/webdriverio/src/middlewares.ts
L0tso: Code that rewrites error name was added here https://github.com/webdriverio/webdriverio/pull/3701#discussion_r263853285 for Safari related bug https://github.com/SeleniumHQ/selenium/issues/8376. Before my changes it worked for all browsers instead of working only for Safari.
The bug is still open, and there is no difference between Safari's `not found element` and Safari's `stale element exception.`

christian-bromann: It seems like throwing the error here is causing the whole issue in the first place. It is independent from the browser. It might make sense to see why this was put here in the first place and whether we want to consider removing this piece afterall.

L0tso: Yes, current issue is independent from the browser. But it's causing by code that was added for handling  [the issue in  SafariDriver](https://bugs.webkit.org/show_bug.cgi?id=216479)
If we remove it we won't be able to handle `stale element exception` (that looks like `not found element`) in Safari anymore, .It can break some tests.

christian-bromann:
@L0tso anything I can do to help to move this forward?

L0tso:
@christian-bromann We should decide. Left it as is, or remove this [part](https://github.com/webdriverio/webdriverio/pull/6848/files#diff-5e6684aa42dc1adaf0b76b3ed32caaed7d0b1b69cbf3da87b06243887591fd89L27-L29). If we remove it we won't be able to handle stale element exception (that looks like not found element in Safari) anymore. It can break some tests.

christian-bromann:
I would like to understand how Safari is behaving different here? Can you show where the protocol response differentiates between e.g. Chrome and Safari? Technically it should behave the same. The reason why I am asking is that we can modify responses given a certain browser name to make it more compatible.

L0tso:
@christian-bromann https://bugs.webkit.org/show_bug.cgi?id=216479

christian-bromann:
I will close this given that this seems to be an issue with Safari that can't be fixed within WebdriverIO. Thanks a lot @L0tso for investigating.

PR #6954, opened 2021-06-01
Implement new async API

christian-bromann:
## Proposed changes

In #6702 we've evaluated what is the best path forward after it was clear that fibers support would not continue with Node.js v16 and up. As a first step this patch improves the async API by minimizing the amount of `await` necessary. Today our async API has no support for element chaining, requiring users to do things like that:

```js
const tagName = await (await $('foobar')).getTagName()
```

This patch adds a much nicer handling of these chained calls and allows things as follows:

```js
const tagName = await $('body').$$('div').get(2).getTagName()
const elem = await $('body').$('header').$$('div').find(
    async (elem) => {
        const loc = await elem.getLocation()
        return loc.x === 30
    }
)
console.log((await elem.getLocation()).x) // 30
console.log(await $('body').$$('div').length) // 5
```

This change should be backwards compatible as the behavior stays the same when an element promise is resolved before calling a command on it. It will be also possible for users to have sync code with async code mixed, e.g.:

```js
describe('webdriver.io page', () => {
    it('should run async', async () => {
        console.log('async result', await $('div').$('div').getTagName())
    })

    it('should run sync', () => {
        browser.url('https://webdriver.io')
        console.log('sync result', $('div').$('div').getTagName())
    })
})
```

This also simplifies transitioning between both modes.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

This patch requires some more work and I will also add some additional documentation changes to have all code examples be `async` by default with a `sync` option to switch.


packages/wdio-utils/src/shim.ts
```diff
@@ -99,6 +146,121 @@ let wrapCommand = function wrapCommand<T>(commandName: string, fn: Function): (.
 
         return commandResult
     }
+
+    function wrapElementFn (promise: Promise<Clients.Browser>, cmd: Function, args: any[], prevInnerArgs?: { prop: string, args: any[] }): any {
+        return new Proxy(
+            Promise.resolve(promise).then((ctx: Clients.Browser) => cmd.call(ctx, ...args)),
```
erwinheitzman, line 152: Would this work as well?
```suggestion
            promise.then((ctx: Clients.Browser) => cmd.call(ctx, ...args)),
```

packages/wdio-utils/src/shim.ts
```diff
@@ -99,6 +146,121 @@ let wrapCommand = function wrapCommand<T>(commandName: string, fn: Function): (.
...
+            {
+                get: (target, prop: string) => {
+                    /**
+                     * if we call a query method on a resolve promise, e.g.:
+                     * ```js
+                     * const elem = await $('foo')
+                     * await elem.$('bar')
```
erwinheitzman, line 159: Can be simplified to the following right?
```suggestion
                     * await $('foo').$('bar')
```

packages/wdio-utils/src/shim.ts
```diff
@@ -99,6 +146,121 @@ let wrapCommand = function wrapCommand<T>(commandName: string, fn: Function): (.
...
+                    /**
+                     * if we call "get" on an element array promise, e.g.:
+                     * ```js
+                     * const elems = await $$('foo').get(2)
+                     * ```
+                     */
+                    if (commandName.endsWith('$$') && prop === 'get') {
+                        return (index: number) => wrapElementFn(
+                            target,
+                            function (this: any, index: number) {
```
erwinheitzman, line 175: what is `this` referring to in this case?

packages/wdio-utils/src/shim.ts
```diff
@@ -99,6 +146,121 @@ let wrapCommand = function wrapCommand<T>(commandName: string, fn: Function): (.
...
+                            function (this: never, mapIterator: Function): any {
+                                // @ts-ignore
+                                return iterators[prop](this, mapIterator)
+                            },
+                            [mapIterator]
+                        )
+                    }
+
+                    /**
+                     * allow to grab the length of fetch element set, e.g.:
```
erwinheitzman, line 202: not just the length but any element property it seems

packages/wdio-utils/src/shim.ts
```diff
@@ -99,6 +146,121 @@ let wrapCommand = function wrapCommand<T>(commandName: string, fn: Function): (.
...
+                     * ```js
+                     * const elemAmount = await $$('foo').length
+                     * ```
+                     */
+                    if (ELEMENT_PROPS.includes(prop)) {
+                        return target.then((res) => res[prop])
+                    }
+
+                    /**
+                     * allow to resolve an chained element query, e.g.:
```
erwinheitzman, line 212:
```suggestion
                     * allow to resolve a chained element query, e.g.:
```

packages/wdio-utils/src/shim.ts
christian-bromann: nope, because we don't know if `promise` is actually a promise, so to be sure we wrap it this way

packages/wdio-utils/src/shim.ts
```diff
@@ -99,6 +146,121 @@ let wrapCommand = function wrapCommand<T>(commandName: string, fn: Function): (.
...
+         * - @wdio/sync package is installed and can be resolved
+         * - we are in a fiber context (flag is set when outer function is wrapped into fibers context)
+         *
+         * also if we run command asynchronous and the command suppose to return an element, we
+         * apply `chainElementQuery` to allow chaining of these promises.
+         */
+        const command = hasWdioSyncSupport && wdioSync && !runAsync
+            ? wdioSync!.wrapCommand(commandName, fn)
+            : ELEMENT_QUERY_COMMANDS.includes(commandName)
+                ? function chainElementQuery(this: Promise<Clients.Browser>, ...args: any[]): any {
```
erwinheitzman, line 257: I suggest to either define the function outside of this or to create an arrow function instead

packages/wdio-utils/src/shim.ts
```diff
@@ -160,27 +324,51 @@ async function executeAsync(this: any, fn: Function, retries: Retries, args: any
...
+        runAsync = false
+        const result = fn.apply(this, args)
+
+        if (typeof result.finally === 'function') {
+            return result.finally(() => (runAsync = switchFlag))
+        }
+
+        if (typeof result === 'function') {
+            return function (this: any, ...args: any[]) {
+                const switchFlag = runAsync
```
erwinheitzman, line 345: switchFlag is a shadow of the switchFlag defined above. it would be better to use another name here to prevent confusion

packages/wdio-utils/src/shim.ts
```diff
@@ -160,27 +324,51 @@ async function executeAsync(this: any, fn: Function, retries: Retries, args: any
...
-        executeSync = wdioSync.executeSync
-        runSync = wdioSync.runSync
+export function switchSyncFlag (fn: Function) {
+    return function (this: unknown, ...args: any[]) {
+        const switchFlag = runAsync
+        runAsync = false
+        const result = fn.apply(this, args)
+
+        if (typeof result.finally === 'function') {
+            return result.finally(() => (runAsync = switchFlag))
```
erwinheitzman, line 340: no need to return here right? if so:
```suggestion
            return result.finally(() => { runAsync = switchFlag })
```

packages/wdio-utils/src/shim.ts
```diff
@@ -160,27 +324,51 @@ async function executeAsync(this: any, fn: Function, retries: Retries, args: any
...
+        }
+
+        if (typeof result === 'function') {
+            return function (this: any, ...args: any[]) {
+                const switchFlag = runAsync
+                const res = result.apply(this, args)
+                if (typeof result.finally === 'function') {
+                    return result.finally(() => (runAsync = switchFlag))
```
erwinheitzman, line 348: no need to return here right?
```suggestion
                    return result.finally(() =>  { runAsync = switchFlag })
```

erwinheitzman (changes requested):
Added some comments, great changes to see!

mgrybyk:
maybe it makes sense to completely get rid of everything related to wdio-sync and implemented Chainable Promise API as the only standard?

christian-bromann:
@mgrybyk as much as I would like to get rid of it we have the majority of users that use sync mode and also like it. Many probably wouldn't like to rewrite their whole test suite so we will have to stick with it until deprecation of Node v16

mgrybyk:
Sorry to say you are right even though it's easier to maintain single official design.
I guess we are stuck till 2023-04-30

klamping:
Anything I can do to help move this PR forward? Is more testing needed?

christian-bromann:
@klamping I was a bit stuck on the TS implementation but @mgrybyk helped me out so I can move forward with this soonish.

klamping:
@christian-bromann Good to hear! Let me know if any testing is needed and I'll do my best to help out.

mgrybyk:
Is it possible to access elements array by index instead of `find`, ex `await $('foo').$$('bar')[2].getTagName()` ?

christian-bromann:
> Is it possible to access elements array by index instead of `find`, ex `await $('foo').$$('bar')[2].getTagName()` ?

I don't think so, would that even be possible in general? Given that accessing an index is not calling a function that you can chain I am not sure. This is why this patch proposes a `get` method to do:

```js
await $('foo').$$('bar').get(2).getTagName()
```

mgrybyk:
@christian-bromann yes, it's possible with `Proxy` :)

christian-bromann:
Cool, let me try that. I definitely prefer using indexes rather than introducing a `get` command.

alexmi256:
Is this fully implemented in 7.9.0? 
I'm getting:
```
TypeError: elem[prop] is not a function
    at /.../node_modules/@wdio/utils/build/shim.js:197:38
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
   ...
    at Context.executeAsync (/.../node_modules/@wdio/utils/build/shim.js:265:16)
    at Context.testFrameworkFnWrapper (/.../node_modules/@wdio/utils/build/test-framework/testFnWrapper.js:52:18)
```
When using something like:
```
const elem = await $('foo').$$('bar').get(0);
```
Maybe I'm using this the wrong way.

christian-bromann:
> const elem = await $('foo').$$('bar').get(0);

We moved away from the `get` command, do this instead:

```js
const elem = await $('foo').$$('bar')[0];
```

PR #6987, opened 2021-06-07
Update docs to prefer new async API

christian-bromann:
## Proposed changes

Given that we have to deprecate sync mode our documentation should not advertise to use it anymore. That said this patch:

- replaces sync with async mode for all examples within the guide section
- makes sure that there is an sync and async example for all API commands

Given that sync and async interface will be the same just with an `await` in front of the statement, I was able to just make all command examples async and use `replace` to generate a sync examples. This is how it now looks like:

new docs

## Types of changes

- [x] Documentation update

## Checklist

## Further comments


packages/webdriverio/src/commands/browser/call.ts
```diff
@@ -3,19 +3,24 @@
...
  *
+ * This command helps to run asynchronous code within a synchronous context. With
+ * WebdriverIO depcrecating synchronous usage (see [RFC](https://github.com/webdriverio/webdriverio/discussions/6702))
+ * this command is not very useful anymore.
+ *
  * <example>
     :call.js
-    it('some testing here', () => {
-        browser.url('http://google.com')
+    it('some testing here', async () => {
```
erwinheitzman, line 12: can this command be removed when everything is async anyway?

website/docs/SetupTypes.md
```diff
@@ -267,13 +240,13 @@ Here is the same example from above, written as a test spec and executed by WDIO
...
-        $('#search_form_input_homepage').setValue('WebdriverIO')
-        $('#search_button_homepage').click()
+        await $('#search_form_input_homepage').setValue('WebdriverIO')
+        await $('#search_button_homepage').click()
 
-        const title = browser.getTitle()
+        const title = await browser.getTitle()
         console.log('Title is: ' + title)
         // outputs: "Title is: WebdriverIO (Software) at DuckDuckGo"
     })
```
klamping, line 252: The paragraph below this needs to be updated or removed

packages/webdriverio/src/commands/browser/call.ts
klamping: That's what I would assume as well

christian-bromann: Yes and no 😄  as we continue to support sync until the end of Node.js v16 we can't get rid of the command just yet

website/docs/SetupTypes.md
christian-bromann: good catch!

website/docs/AutomationProtocols.md
```diff
@@ -123,7 +121,7 @@ const { remote } = require('webdriverio')
```
klamping, line 121: Should the 'We recommend wrapping your Puppeteer calls within the `call` command, so that all calls are executed before WebdriverIO continues with the next WebDriver command' paragraph be removed?

redonkulus:
@christian-bromann when is it necessary to add `await` before `expect`, I see it applied inconsistently in the examples.

DMazona:
> @christian-bromann when is it necessary to add `await` before `expect`, I see it applied inconsistently in the examples.

I know this wasn't for me but i've come from a jasmine/protractor world and have given this some thought before. I like not applying await before expect because it stands out more in the it blocks. so if you have
```
await el.click()
await el2.click()
expect(await el.getText()).toEqual('42)
```
you can more easily see which line the assert comes from

redonkulus:
I've have seen issues locally where not adding `await` before expect will cause an assertion failure to get skipped. I'm curious what the overall approach should be, I'm fine with or without, just want consistency.

For example:

```
// does not fail when I expect it to
const el = await page.elementFoo; // assume getter that returns element obj
expect(el).toHaveElementClassContaining('foo');

// successfully fails when "foo" does not exist
await expect(page.elementFoo).toHaveElementClassContaining('foo');
```

christian-bromann:
@redonkulus @DMazona in an async environment you always have to use `await` when using an `expect-webdriverio` matcher, e.g.:

```js
await expect(browser).toHaveTitle("foobar")
```

However if you like to use assertions without `await` you can use the synchronous matchers, e.g.:
```js
const title = await browser.getTitle()
expect(title).toBe("foobar")
```

website/docs/SyncVsAsync.md
```diff
@@ -25,6 +25,12 @@ The package will be automatically detected by the framework and the environment
 
 If you're using [`@wdio/sync`](https://www.npmjs.com/package/@wdio/sync) then you can avoid awaiting for command calls. It is still required to deal with Promises from 3rd-party libraries, you should use [browser.call](api/browser/call) for this to wrap them and make them synchronous too.
 
+:::caution
+
+Running commands synchronously is depcrecated and will stop working with Node.js v16 and above.
```
erwinheitzman, line 30: I would use the same code snippet you have used in the other documentation pages:
```suggestion
:::caution
Synchronous Mode will depcrecated with Node.js v16. With an update to the
underlying Chromium version it became technically impossible to provide the
same synchronous behavior. We recommend to start transition to asynchronous
command execution. For more information, see our <a href="https://github.com/webdriverio/webdriverio/discussions/6702">RFC</a>.
:::
```

website/docs/MocksAndSpies.md
```diff
@@ -54,36 +37,12 @@ Once you have defined a mock you can define custom responses for it. Those custo
...
-// outputs: "[ 'Injected (non) completed Todo', 'Injected completed Todo' ]"
-```
-<!--Async-->
 ```js
 const mock = await browser.mock('https://todo-backend-express-knex.herokuapp.com/', {
     method: 'get'
 })
 
-await mock.respond([{
+mock.respond([{
```
erwinheitzman, line 45: mock.respond doesn't need the await here?

erwinheitzman (commented):
Some small remarks

website/docs/MocksAndSpies.md
christian-bromann: nope, respond is a synchronous operation

website/docs/SyncVsAsync.md
christian-bromann: 👍

PR #7000, opened 2021-07-20
TypeScript support for new Async API

christian-bromann:
With getting rid of fibers we want to support a better async API, e.g.:
```js
const tagName = await $('foo').$$('bar').get(2).getTagName()
const elem = await $('body').$('header').$$('div').find(
    async (elem) => {
        const loc = await elem.getLocation()
        return loc.x === 30
    }
)
console.log((await elem.getLocation()).x) // 30
console.log(await $('foo').$$('bar').length) // 5
```

This API needs to have proper type support.


packages/webdriverio/src/types.ts
```diff
@@ -11,9 +11,78 @@ import type DevtoolsInterception from './utils/interception/devtools'
...
+interface ChainablePromiseArray<T> extends Promise<T> {
+    /**
+     * Amount of element fetched.
+     */
+    length: Promise<number>
+
+    /**
+     * Unwrap the nth element of the element list.
+     */
+    get: (i: number) => T
```
mgrybyk, line 63: I'd replace `get` method with accessing element by index in the elements array

packages/webdriverio/src/types.ts
```diff
@@ -11,9 +11,78 @@ import type DevtoolsInterception from './utils/interception/devtools'
...
+    forEach: (iterator: (elem: T) => void) => ChainablePromiseArray<T>
+    forEachSeries: (iterator: (elem: T) => void) => ChainablePromiseArray<T>
```
mgrybyk, line 65: What do *Series methods stand for?

packages/webdriverio/src/types.ts
```diff
@@ -11,9 +11,78 @@ import type DevtoolsInterception from './utils/interception/devtools'
...
+    findSeries: (iterator: (elem: T) => boolean) => ChainablePromiseArray<T>
+    findIndex: (iterator: (elem: T) => boolean) => ChainablePromiseArray<T>
+    findIndexSeries: (iterator: (elem: T) => boolean) => ChainablePromiseArray<T>
+    some: (iterator: (elem: T) => boolean) => ChainablePromiseArray<T>
+    someSeries: (iterator: (elem: T) => boolean) => ChainablePromiseArray<T>
+    every: (iterator: (elem: T) => boolean) => ChainablePromiseArray<T>
+    everySeries: (iterator: (elem: T) => boolean) => ChainablePromiseArray<T>
+    filter: (iterator: (elem: T) => boolean) => ChainablePromiseArray<T>
+    filterSeries: (iterator: (elem: T) => boolean) => ChainablePromiseArray<T>
+    reduce: (iterator: (initialValue: any, elem: T) => any, initialValue: any) => any
```
mgrybyk, line 78: Origina Array's `reduce` method definition looks like this 
```
reduce<U>(callbackfn: (previousValue: U, currentValue: T, currentIndex: number, array: T[]) => U, initialValue: U): U;
```
Is it possible to adjust our wrapper accordingly?

packages/webdriverio/src/types.ts
christian-bromann: same as the companion command but runs only one callback at a time.

packages/webdriverio/src/types.ts
christian-bromann: 👍  let me take a look

christian-bromann:
I finished the typing definition and refactored the typing tests to actually check for proper types. This is ready for review now.

PR #6999, opened 2021-06-10
Use the Ky request library in a web browser context instead of Got

jlipps:
## Proposed changes

Some time ago, WebdriverIO standardized on Got as the request library. Unfortunately this broke downstream projects like Web2Driver which need to be able to bundle the webdriver package for the web. This PR adds in support for the Ky browser-based request library, and selects it in a non-Node.js environment.

## Types of changes

[//]: # (What types of changes does your code introduce to WebdriverIO?)
[//]: # (_Put an `x` in the boxes that apply_)

- [ ] Bugfix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## Checklist

[//]: # (_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._)

- [X] I have read the [CONTRIBUTING](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md) doc
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [X] I have added proper type definitions for new commands (if appropriate)

## Further comments

There are currently no tests for the browser-based client, since I think that would be a pretty big update to the whole test environment. I'm open to hearing how to approach testing (beyond using it in my own project).

PR #7169, opened 2021-07-22
feat: allow use of the ky request library in browser contexts

christian-bromann:
See original PR https://github.com/webdriverio/webdriverio/pull/6999


jlipps:
awesome, thanks @christian-bromann!

Issue #7284, opened 2021-08-16
Configuration typings incorrect for capabilities "goog:chromeOptions"


waynevanson:
**Environment (please complete the following information):**
 - **WebdriverIO version:**  "^7.9.1"
 - **Mode:** Standalone 
 - **If WDIO Testrunner, running sync/async:** N/A
 - **Node.js version:** 14.7.3
 - **NPM version:** 6.14.14
 - **Browser name and version:** N/A
 - **Platform name and version:** MacOS and not sure (new laptop)
 - **Additional wdio packages used (if applicable):**N/A

**Config of WebdriverIO**

```ts
import { RemoteOptions, remote } from "webdriverio";

const configuration: RemoteOptions = {
  capabilities: {
    
    // Type 'string' is not assignable to type 'WebdriverIO'.ts(2322)
    // (property) Capabilities.browserName?: string | undefined
    browserName: "chrome",
    
    // Property 'capabilities' is missing in type '{ headless: true; }' but required in type 'WebdriverIO'.ts(2741)
    // Options.d.ts(251, 5): 'capabilities' is declared here.
    // (property) VendorExtensions['goog:chromeOptions']?: ChromeOptions | undefined
    "goog:chromeOptions": {
        headless: true,
    }
  }
}

const browser = remote(configuration) 
```

**Describe the bug**
In Capabilities, it runs the following correctly but incorrectly throws and error during the type checking phase (typescript)

The above configuration is valid.

**To Reproduce**

type config into an IDE with TS support

**Expected behavior**
No compile time errors in this case.

**Log**
Mentioned in config

christian-bromann:
Thanks for reporting the issue. Your typings are correct however, `capabilities` are not specified in your example. It accepts a raw W3C object as well as normal caps and multiremote caps. For your example the following example would work:

```ts
import { RemoteOptions, remote } from "webdriverio";
import type { Capabilities } from '@wdio/types'

const configuration: RemoteOptions = {
  hostname: 'foo',
  capabilities: {
    // Type 'string' is not assignable to type 'WebdriverIO'.ts(2322)
    // (property) Capabilities.browserName?: string | undefined
    browserName: 'chrome',
    
    // Property 'capabilities' is missing in type '{ headless: true; }' but required in type 'WebdriverIO'.ts(2741)
    // Options.d.ts(251, 5): 'capabilities' is declared here.
    // (property) VendorExtensions['goog:chromeOptions']?: ChromeOptions | undefined
[...]

PR #7287, opened 2021-08-20
Document WebdriverIO TypeScript definitions

christian-bromann:
Within our `@wdio/types` package we define a bunch of types that TypeScript users might want to use. We should add documentations to these types. I suggest to look out for configurations or capability definition examples within our docs and add a TypeScript example to it (as e.g. a TypeScript tab).

See also #7284


waynevanson (commented):
Looks good to me!

PR #7349, opened 2021-08-27
Further improvements between sync and async execution

christian-bromann:
## Proposed changes

As more people transition between sync and async we need to ensure that both APIs can be used interchangeably to allow a fluent transition. As mentioned in #7289 by @RossVertizan there are still some issues. In this case using async code within synchronous `waitUntil` and `call` functions, e.g.:

```js
browser.call(async () => {
    const isExisting = await $("foo").$("bar").isExisting()
    return isExisting
})
```

This example currently doesn't work. This patch tries to fix this.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

work in progress

### DIFF
--- a/packages/wdio-utils/src/shim.ts
+++ b/packages/wdio-utils/src/shim.ts
@@ -32,6 +32,7 @@ declare global {
 const ELEMENT_QUERY_COMMANDS = ['$', '$$', 'custom$', 'custom$$', 'shadow$', 'shadow$$', 'react$', 'react$$']
 const ELEMENT_PROPS = ['elementId', 'error', 'selector', 'parent', 'index', 'isReactElement', 'length']
 const PROMISE_METHODS = ['then', 'catch', 'finally']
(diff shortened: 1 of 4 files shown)


christian-bromann:
Here is a smoke test that can reproduce the issue:
__tests/mocha/test.js__
```
    it('allow async call within sync environment', () => {
        expect(browser.call(() => Promise.resolve('foo'))).toBe('foo')
        expect(browser.call(async () => 'bar')).toBe('bar')

        browser.isExistingScenario()
        expect(browser.call(async () => {
            const body = await browser.$('body')
            return body.$('.selector-1').isExisting()
        })).toBeNaN(true)
    })
```

christian-bromann:
As people move on porting tests to async I doubt it makes sense to spend much more time on making sync/async more compatible. Please check [our guidelines](https://webdriver.io/docs/async-migration) on how to port tests over and reach out if there are any questions. Cheers!

PR #7352, opened 2021-08-27
chore: make return value optional in executeAsync function

Joozty:
## Proposed changes

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/webdriverio/tests/__mocks__/got.ts
+++ b/packages/webdriverio/tests/__mocks__/got.ts
@@ -218,7 +218,7 @@ const requestMock: any = jest.fn().mockImplementation((uri, params) => {
         const script = Function(params.json.script)
         let result
         script.call(this, ...params.json.args, (_result: any) => result = _result)
-        value = result
+        value = result ?? {}
         break
     } case `${path}/${sessionId}/element/${genericElementId}/elements`:
         value = [
--- a/packages/webdriverio/tests/commands/browser/executeAsync.test.ts
+++ b/packages/webdriverio/tests/commands/browser/executeAsync.test.ts
@@ -46,18 +46,4 @@ describe('isEnabled test', () => {
         // @ts-ignore test invalid parameter
         expect(() => browser.executeAsync(1234)).toThrow()
     })
-
-    it('should allow empty done call', async () => {
-        const browser: WebdriverIO.BrowserObject = await remote({
-            baseUrl: 'http://foobar.com',
-            capabilities: {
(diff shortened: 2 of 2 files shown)


christian-bromann:
@Joozty thanks for taking a stab at this. All unit tests for WebdriverIO commands are stubbed. You can find the stub and the reason why it is returning `{}` [here](https://github.com/webdriverio/webdriverio/blob/main/packages/webdriverio/tests/__mocks__/got.ts#L217-L222). You can update the return value to return undefined rather than `{}`. Given this is a typing change I suggest to also add an example in the project [typing tests](https://github.com/webdriverio/webdriverio/blob/main/tests/typings/webdriverio/async.ts#L25).

packages/webdriverio/tests/__mocks__/got.ts
```diff
@@ -218,7 +218,7 @@ const requestMock: any = jest.fn().mockImplementation((uri, params) => {
         const script = Function(params.json.script)
         let result
         script.call(this, ...params.json.args, (_result: any) => result = _result)
-        value = result ?? {}
+        value = result
```
christian-bromann, line 221: I believe this change causes the test to fail. It it really necessary?

Joozty: it is because this is a bug. Value cannot be undefined or null but the truth is that real call would return these values.

christian-bromann: then we need to adjust the unit test

Joozty: I don't know why it is failing and have trouble running tests locally. If you know how to solve it let me know otherwise I will close the PR and report this as an issue rather.

christian-bromann:
@Joozty I recommend to just revert all changes in

- `packages/webdriverio/tests/__mocks__/got.ts`
- `packages/webdriverio/tests/commands/browser/executeAsync.test.ts`

Given that you are only changing types you should not be concerned about unit tests that check functionality.

Joozty:
@christian-bromann I don't agree since the unit test is written wrong. However, I cannot afford spending much time on this atm so I reverted those changes. Please have a look :)

christian-bromann:
> I don't agree since the unit test is written wrong

It is a unit test which mocks the WebDriver request. There might an improvement we can make in that particular one but since you only changed types here there is no need for it now.

christian-bromann (approved):
Thanks 👍

PR #7401, opened 2021-09-08
fix(webdriver): browser request should use btoa for basic auth, not atob

jlipps:
Oops :-) I thought "atob" meant "ascii to base64", but it does not!

### DIFF
--- a/packages/webdriver/src/request/browser.ts
+++ b/packages/webdriver/src/request/browser.ts
@@ -38,7 +38,7 @@ export default class BrowserRequest extends WebDriverRequest {
         }
 
         if (options.username && options.password) {
-            const encodedAuth = atob(`${options.username}:${options.password}`)
+            const encodedAuth = btoa(`${options.username}:${options.password}`)
             kyOptions.headers = {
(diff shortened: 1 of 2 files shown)


christian-bromann (commented):
Can we add a test for this?

jlipps:
Hmm, yes, I think we can add a test for this, but only one that mocks `atob` in the first place, since it's not a Node function, and we don't have a test harness that bundles and runs code in a browser context.

Is a test that mocks `atob` any better than no test at all? I suppose the most it would do is make sure we don't change it to some other function name in the future?

christian-bromann:
@jlipps you can run Jest tests within a browser environment using jsdom, similar we do this e.g. with the [`isFocused`](https://github.com/webdriverio/webdriverio/blob/main/packages/webdriverio/tests/scripts/isFocused.test.ts) test.

jlipps:
Oh, sweet! Done.

jlipps:
the windows CI failure seems unrelated.

christian-bromann (approved):
This looks good, thanks 👍

Issue #7468, opened 2021-09-23
Sample Typescript project get generated without tsconfig file


praveendvd:
**Environment (please complete the following information):**
 - **WebdriverIO version:** [e.g. 4.13.2] :  7.13.1
 - **Mode:** [Standalone mode or WDIO Testrunner] : Test runner
 - **If WDIO Testrunner, running sync/async:** [e.g. sync/async] : Async
 - **Additional wdio packages used (if applicable):** [e.g. @wdio/spec reporter, @wdio/selenium-standalone service] : @wdio/cli

**Describe the bug**

On creating sample typescri[t project with @wdio/cli , it does not create tsconfig.json file automatically. This results in type error for global objects like browser , $ etc .

**To Reproduce**

  run `npx wdio`
  generate sample jasmine-typescript (or other typescript ) project

**Expected behavior**

basic tsconfig.json file should be created and the project should run without any type error

**Log**

![image](https://user-images.githubusercontent.com/45095911/134598566-eb008ae1-ce93-449c-8b66-99816d175b54.png)

**Required tsconfig**

```
{
    "compilerOptions": {
        "types": ["node", "webdriverio/async","@wdio/jasmine-framework"],
        "target": "ES6"
    }
    
}
```

christian-bromann:
I suggest to check if a `tsconfig.json` is existing and if not create a minimal required one as you suggested there.

praveendvd:
> I suggest to check if a `tsconfig.json` is existing and if not create a minimal required one as you suggested there. The code for this needs to be written [here](https://github.com/webdriverio/webdriverio/blob/main/packages/wdio-cli/src/commands/config.ts). If there are any questions, please let me know.

Thank you so much  @christian-bromann  will pick this up , then .

christian-bromann:
Thanks for taking a stab at this. Let's continue conversation in #7471

PR #7471, opened 2021-09-26
Check for tsconfig file and create if it doesn't exists for type script project

praveendvd:
## Proposed changes

This is to create tsconfig.json for auto generated typescript project . This is fix for Sample Typescript project get generated without tsconfig file #7468

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue) : x
- [ ] Documentation update

## Checklist

## Further comments

Not sure how to write the tests


praveendvd:
@christian-bromann  I used the same fs.promises.writefile from babel step , i am not sure why the test fails could you have a look .

packages/wdio-cli/src/commands/config.ts
```diff
@@ -65,6 +65,16 @@ const runConfig = async function (useYarn: boolean, yes: boolean, exit = false)
         if (!hasPackage('ts-node')) {
             packagesToInstall.push('ts-node', 'typescript')
         }
+        if (!hasFile('tsconfig.json')){
+            const config = {
+                'compilerOptions': {
+                    'types': ['node', 'webdriverio/async', frameworkPackage.package],
+                    'target': 'ES6'
+                }
```
christian-bromann, line 73:
```suggestion
                compilerOptions: {
                    types: ['node', 'webdriverio/async', frameworkPackage.package],
                    target: 'ES2018'
                }
```

christian-bromann (commented):
@praveendvd thanks for taking a stab at this. Please note that our unit tests mostly mock dependencies such as `fs` so that we don't actually write files but are able to assert the call for it. Please see an example [here](https://github.com/webdriverio/webdriverio/blob/main/packages/wdio-cli/tests/commands/config.test.ts#L182-L183). Let's add a similar unit test that ensures that we write a tsconfig if none is given.

packages/wdio-cli/src/commands/config.ts
```diff
@@ -65,6 +65,16 @@ const runConfig = async function (useYarn: boolean, yes: boolean, exit = false)
         if (!hasPackage('ts-node')) {
             packagesToInstall.push('ts-node', 'typescript')
         }
+        if (!hasFile('tsconfig.json')){
+            const config = {
+                'compilerOptions': {
+                    'types': ['node', 'webdriverio/async', frameworkPackage.package],
+                    'target': 'ES6'
```
erwinheitzman: Let's keep es5 as a default please, having es6 will cause issues when writing unsupported code inside `execute` commands for Internet Explorer for example and the output doesn't matter other then compatibility.

packages/wdio-cli/src/commands/config.ts
```diff
@@ -65,6 +65,16 @@ const runConfig = async function (useYarn: boolean, yes: boolean, exit = false)
...
+                }
+            }
+
+            await fs.promises.writeFile(path.join(process.cwd(), 'tsconfig.json'), JSON.stringify(config, null, 2))
```
erwinheitzman, line 76: An indentation of 4 is used for the Babel config, any reason to change that for Typescript?

packages/wdio-cli/src/commands/config.ts
praveendvd: @erwinheitzman  i believe browsers support only es5 right ? if i am not wrong

christian-bromann: I have some doubts how stable this is. Babel doesn't know that the piece of function is executed in the browser. It mostly defines helper variables outside of it which won't be available when run in the browser.

packages/wdio-cli/src/commands/config.ts
```diff
@@ -65,6 +65,16 @@ const runConfig = async function (useYarn: boolean, yes: boolean, exit = false)
         if (!hasPackage('ts-node')) {
             packagesToInstall.push('ts-node', 'typescript')
         }
+        if (!hasFile('tsconfig.json')){
+            const config = {
+                'compilerOptions': {
+                    'types': ['node', 'webdriverio/async', frameworkPackage.package],
```
erwinheitzman: I think we might need `expect-webdriverio` as well for the default assertions

packages/wdio-cli/src/commands/config.ts
praveendvd: I will change to 4 , thank you so much @erwinheitzman  for the review 🎉

erwinheitzman (commented):
Thank you for this PR! Left some feedback :)

packages/wdio-cli/src/commands/config.ts
erwinheitzman: Babel? This is Typescript which will change your arrow functions for example to regular function (if your target is es5 instead of es6) which is needed for IE (specifically for `execute` command only).

praveendvd:
@christian-bromann  and @erwinheitzman 

at the end of setup there is message saying , now you can execute the project as 

`npx wdio run wdio.config.js`

but this is also possible to be run as

packages/wdio-cli/src/commands/config.ts
erwinheitzman: @praveendvd up to date browsers support higher then es5 but I can't find all the supported versions for each browser this minute

erwinheitzman: all in all, this transforms the user code from typescript syntax to the chosen javascript output syntax.

christian-bromann:
> should we update the fnal message to add ts or js

yes, based on whether or not the user picked TS as compiler we should set the file name to `wdio.conf.ts`

praveendvd:
> > should we update the fnal message to add ts or js
> 
> yes, based on whether or not the user picked TS as compiler we should set the file name to `wdio.conf.ts`

i am not sure from whic hfile this message comes from could you help me with this

christian-bromann:
> i am not sure from whic hfile this message comes from could you help me with this

https://github.com/webdriverio/webdriverio/blob/main/packages/wdio-cli/src/constants.ts#L21-L25

packages/wdio-cli/src/commands/config.ts
erwinheitzman: I see you changed it to es2018 which is even newer so that doesn't help :P

erwinheitzman: to give you an idea `es6` === `es2015` as far as I am aware there's no `es2XXX` variant for `es5` ;)

erwinheitzman: `lib` is used to allow the user to write specific syntax versions
`target` is used to compile to a specific javascript version for compatibility reasons

packages/wdio-cli/src/constants.ts
```diff
@@ -24,6 +24,12 @@ To run your tests, execute:
 $ npx wdio run wdio.conf.js
 `
 
+export const CONFIG_HELPER_SUCCESS_MESSAGE_TS = `
+Configuration file was created successfully!
+To run your tests, execute:
+$ npx wdio run wdio.conf.ts
```
christian-bromann, line 30: No need to have an extra variable for this, just do:
```js
export const CONFIG_HELPER_SUCCESS_MESSAGE_TS = `
Configuration file was created successfully!
To run your tests, execute:
$ npx wdio run wdio.conf.%s
`

and us it via `util.format(CONFIG_HELPER_SUCCESS_MESSAGE, 'ts' /* or "js" */)`

**Review by christian-bromann — COMMENTED** [review]
Can we add an assertion that we used `writeFile` to write the tsconfig file?

**Comment by praveendvd** [comment]
> Can we add an assertion that we used `writeFile` to write the tsconfig file?

sure @christian-bromann  , how to mock hasfile returns true use case ?

**Comment by christian-bromann** [comment]
> how to mock hasfile returns true use case ?

See Jest docs: https://jestjs.io/docs/mock-functions#mock-return-values

**Comment by christian-bromann** [comment]
@praveendvd can we add an additional check to see if we actually write a tsconfig file?

**Comment by praveendvd** [comment]
Sure @christian-bromann I will check that , I am trying to update the snapshot now

**Comment by erwinheitzman** [comment]
all you need it to add `--updateSnapshot` to the test command like so `npx jest ./packages/webdriverio/test --updateSnapshot` :)

File: packages/wdio-cli/tests/commands/config.test.ts [hunk 6]
```diff
@@ -104,28 +104,44 @@ describe('install compliant NPM tag packages', () => {
     test('it should install tagged version if cli is tagged with beta', async () => {
         setFetchSpec('beta');
         (inquirer.prompt as any as jest.Mock).mockReturnValue(Promise.resolve(args))
+        // @ts-expect-error
+        fs.promises = { writeFile: jest.fn()
```
erwinheitzman, line 108: is these are all the same then you can define it once on top or place it in the beforeEach instead unless the other tests would be affected by it

erwinheitzman:
Looking good but could you add a test which verifies that both `js` and `ts` are supported? 👍

packages/wdio-cli/tests/commands/__snapshots__/config.test.ts.snap
```diff
@@ -39,7 +39,7 @@ For for information on TypeScript integration check out: https://webdriver.io/do
     "
 Configuration file was created successfully!
 To run your tests, execute:
-$ npx wdio run wdio.conf.js
+$ npx wdio run wdio.conf.ts
```
erwinheitzman, line 42: why are these now `.ts`? is this because the project has a `tsconfig.json` file? if so we should remove that dependency

erwinheitzman:
Noticed that the default for the tests is now `.ts`, left some feedback on that as well

packages/wdio-cli/tests/commands/__snapshots__/config.test.ts.snap
christian-bromann: No it is because the unit test is simulating a configuration with TypeScript as compiler.

christian-bromann: @praveendvd however I don't believe they should be all TypeScript. Let's mock `hasFile` so that it only returns `true` when running our TypeScript unit test.

praveendvd: @christian-bromann  Thats the success message , there i am not checking if tsconfig.json is there.  Typescript auto generated project doesn't have wdio.config.js . should i change message to ` npx wdio run wdio.conf.ts or  npx wdio run wdio.conf.js`

packages/wdio-cli/tests/commands/config.test.ts
christian-bromann: Agree, let's move this up to the top of the file and call `fs.promises.writeFile.mockClear()` in `beforeEach`.

packages/wdio-cli/tests/commands/__snapshots__/config.test.ts.snap
praveendvd: @christian-bromann  i will set hsFIle to return true for actual tests.

christian-bromann: Nevermind this, seems the unit test all are already like this.

christian-bromann (commented):
Two last requests:
- https://github.com/webdriverio/webdriverio/pull/7471/files#r716621185
- add an assertion that `fs.promises.writeFile` was called with proper parameters

praveendvd:
> Looking good but could you add a test which verifies that both `js` and `ts` are supported? 👍

@erwinheitzman 

yes i tested with this and it works , thank you so much 

@christian-bromann

packages/wdio-cli/tests/commands/config.test.ts
praveendvd: WIll look into this

praveendvd:
@christian-bromann  and @erwinheitzman  added the assertion could you have a look 

Also tried the other use case if hasFIle return true:

```
 test('it should not install tagged version if cli is tagged with a specific version', async () => {
        setFetchSpec('7.0.8');

        (inquirer.prompt as any as jest.Mock).mockReturnValue(Promise.resolve(args))
        // @ts-expect-error
        fs.promises = { writeFile: jest.fn()
            .mockReturnValue(Promise.resolve('')) }
               
        jest.mock('../../src/utils', () => ({
            hasFile:jest.fn().mockReturnValue(true),
        }))
        await handler({} as any)

        expect(fs.promises.writeFile).not.toBeCalled()
        expect(consoleLogSpy.mock.calls).toMatchSnapshot()
    })

```

but this always executes fs.promise.writeFile no matter what

christian-bromann (approved):
👍  thanks!

PR #7359, opened 2021-09-27
Fix broken iFrame e2e test

christian-bromann:
This patch https://github.com/webdriverio/webdriverio/commit/b93e7f40fbda87cd77ec525b2a92c86cfb4c58cc disabled an e2e test that seem to start failing for unknown reasons. Running the test locally seems fine. I suggest instead of using a 3rd party page to replace it with an own blank page that allows us to verify that switching to the iframe was successul.


erwinheitzman (approved):
Nice going Christian

PR #7491, opened 2021-10-01
Removing install instructions - not needed with npx - now simpler

andreasmarkussen:
Removing the comments on installing wdio, since we now use npx, and this removes the need to install wdio

## Proposed changes

This PR is to simplify the Getting Started page, and it did confuse us when we stated on it. 
See the title. this is a very simple change

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/website/docs/GettingStarted.md
+++ b/website/docs/GettingStarted.md
@@ -9,22 +9,21 @@ Welcome to the WebdriverIO documentation. It will help you to get started fast. 
 These are the docs for the latest version (__>=7.x__) of WebdriverIO. If you are still using an older version, please visit the [old documentation websites](/versions)!
 :::
 
-## Installation
+## Installation considerations
 
-Use npm or Yarn to install the WebdriverIO test runner in your Node.js project. See [system requirements](#system-requirements).
+There are two ways of using WebdriverIO - you can either add it to the application you plan to test, or you can create a brand new NodeJS project. 
 
-```bash npm2yarn
-npm install @wdio/cli
-```
+If you want to do an experiment, then it makes sense to create your own new project, but if you want to use it to test an existing NodeJS app, then it makes sense to add WebdriverIO to this project. 
 
-This single command downloads the WebdriverIO CLI tool that helps you set up WebdriverIO in your project.
+The installation is the same no matter what. 
 
 ## Set Up
 
-Once you've installed the CLI you can bootstrap a _Hello World_ test suite into your project by running:
+In order to use wdio, you need to specify a folder, either the current folder with `npx wdio .` or a specific folder with a package.json file. 
+If the folder does not exists, wdio will create the folder and make it into a nodejs project. 
 
 ```bash
-npx wdio config
+npx wdio /path/to/project
 ```
 
 ![Set Up](../static/img/setup.gif)


website/docs/GettingStarted.md
```diff
@@ -9,22 +9,21 @@ Welcome to the WebdriverIO documentation. It will help you to get started fast.
 These are the docs for the latest version (__>=7.x__) of WebdriverIO. If you are still using an older version, please visit the [old documentation websites](/versions)!
 :::
 
-## Installation
+## Installation considerations
```
christian-bromann, line 12:
```suggestion
## Installation
```

website/docs/GettingStarted.md
```diff
@@ -9,22 +9,21 @@ Welcome to the WebdriverIO documentation. It will help you to get started fast.
...
 
-Use npm or Yarn to install the WebdriverIO test runner in your Node.js project. See [system requirements](#system-requirements).
+There are two ways of using WebdriverIO - you can either add it to the application you plan to test, or you can create a brand new NodeJS project.
```
christian-bromann, line 14: If we mention that there are two ways to install wdio we should also clearly describe them. I think we can just use the starter toolkit `npx wdio .` and assume someone is currently in an existing project

website/docs/GettingStarted.md
```diff
@@ -9,22 +9,21 @@ Welcome to the WebdriverIO documentation. It will help you to get started fast.
...
-```
+If you want to do an experiment, then it makes sense to create your own new project, but if you want to use it to test an existing NodeJS app, then it makes sense to add WebdriverIO to this project. 
 
-This single command downloads the WebdriverIO CLI tool that helps you set up WebdriverIO in your project.
+The installation is the same no matter what. 
 
 ## Set Up
 
-Once you've installed the CLI you can bootstrap a _Hello World_ test suite into your project by running:
+In order to use wdio, you need to specify a folder, either the current folder with `npx wdio .` or a specific folder with a package.json file.
```
christian-bromann, line 22: Let's assume the user adds WebdriverIO to an existing project. I believe this is the most common use case

website/docs/GettingStarted.md
```diff
@@ -9,22 +9,21 @@ Welcome to the WebdriverIO documentation. It will help you to get started fast.
...
 
 ## Set Up
 
-Once you've installed the CLI you can bootstrap a _Hello World_ test suite into your project by running:
+In order to use wdio, you need to specify a folder, either the current folder with `npx wdio .` or a specific folder with a package.json file. 
+If the folder does not exists, wdio will create the folder and make it into a nodejs project. 
 
 ```bash
-npx wdio config
+npx wdio /path/to/project
```
christian-bromann, line 26:
```suggestion
npx wdio .
```

christian-bromann (commented):
I like the idea of simplifying this. Added some comments.

website/docs/GettingStarted.md
andreasmarkussen: Could be - but for learning the framework having it next to it, is what I just did for some demos and tutorials

website/docs/GettingStarted.md
andreasmarkussen: Well - I just did the Webdriver in a folder for an experiment. 
Perhaps also if you are doing an appium project on an ios and android project, then there is no existing nodejs project?

christian-bromann: You right. I think my concern was just about saying that there are two ways without explicitly documenting both

website/docs/GettingStarted.md
christian-bromann: can we word this so it works for both use cases?

Issue #7609, opened 2021-10-26
[🐛 Bug]: Some failed SauceLabs run result in Unnamed job descriptions


gabriel-kohen-by:
### WebdriverIO Version

7.7.4

### Node.js Version

14

### Mode

WDIO Testrunner

### Which capabilities are you using?

```typescript
import config from './wdio.shared.conf';

const isDebugMode = process.env.DEBUG;

config.user = process.env.SAUCE_USERNAME;
config.key = process.env.SAUCE_ACCESS_KEY;
config.bail = ("SAUCE_BAIL" in process.env)  ? process.env.SAUCE_BAIL : 1;
config.region = 'us';
const BUILD_PREFIX = process.env.BUILD_PREFIX ? `GitHub Actions-` : '';
const defaultBrowserSauceOptions = {
  build: `${BUILD_PREFIX}Luminate Portal Dashboard Mfe build-${new Date().toISOString()}`,
  screenResolution: '1600x1200',
  extendedDebugging: true,
  capturePerformance: true,
  timeZone: 'Chicago'
};
const defaultBrowserDesktopSauceOptions = {
  screenResolution: '1600x1200'
};
config.services?.push([
  'sauce',
  {
    sauceConnect: isDebugMode,
    setJobNameInBeforeSuite: true
  }
]);
config.mochaOpts= {...config.mochaOpts, bail: (config.bail === 1) }

config.waitforTimeout = 20000;

config.capabilities = [
  {
    browserName: 'chrome',
    browserVersion: 'latest',
    platformName: 'Windows 10',
    'sauce:options': {
      ...defaultBrowserSauceOptions,
      ...defaultBrowserDesktopSauceOptions
    }
  },
  {
    browserName: 'chrome',
    browserVersion: 'latest',
    platformName: 'macOS 10.15',
    'sauce:options': {
      ...defaultBrowserSauceOptions,
      ...defaultBrowserDesktopSauceOptions
    }
  },
  {
    browserName: 'MicrosoftEdge',
    browserVersion: 'latest',
    platformName: 'Windows 10',
    'sauce:options': {
      ...defaultBrowserSauceOptions,
      ...defaultBrowserDesktopSauceOptions
    }
  },
  {
    browserName: 'firefox',
    browserVersion: 'latest',
    platformName: 'macOS 10.15',
    'sauce:options': {
      ...defaultBrowserSauceOptions,
      ...defaultBrowserDesktopSauceOptions
    }
  }
];

export { config };
[...]

PR #7672, opened 2021-11-09
Set the job name before a suite is run

gabriel-kohen-by:
## Proposed changes

Per https://github.com/webdriverio/webdriverio/issues/7609 we need to account for suite `before` logic will be long running or failing and we want to give intication early on running the suite's tests of what is running.

## Types of changes

- [ ] Documentation update

## Checklist


christian-bromann (changes requested):
Thanks for taking a stab at this!

Let's add one or multiple test cases [here](https://github.com/gabriel-kohen-by/webdriverio/blob/main/packages/wdio-sauce-service/tests/service.test.ts#L60-L66) and ensure that the if statement is only executed if situation is as expected.

gabriel-kohen-by:
@christian-bromann, thanks for the feedback. I've tweaked and added some assertions to account for the new logic. Let me know if there is anything I need to change.

packages/wdio-sauce-service/tests/service.test.ts
```diff
@@ -485,7 +507,7 @@ test('_uploadLogs should not fail in case of a platform error', async () => {
     const service = new SauceService(
         {},
         {},
-        { outputDir: '/foo/bar' } as any
+            { outputDir: '/foo/bar' } as any
```
christian-bromann, line 510:
```suggestion
        { outputDir: '/foo/bar' } as any
```

christian-bromann (approved):
LGTM 👍  thanks a lot!

gabriel-kohen-by:
@christian-bromann , not sure why the e2e failed on the Mac. Can't think the changes are related to this failure. Is the e2e flaky in nature?

christian-bromann:
@gabriel-kohen-by yeah, the windows e2e tests aren't that stable atm. Will go ahead and merge. Thanks a lot!

gabriel-kohen-by:
Thanks for merging @christian-bromann. How does the release process works as far as making it to GA?

christian-bromann:
I just need to kick off a GitHub action to release a new version. Will do so now.

wswebcreation:
This breaks setting the job name in Jasmine, because the `suiteTitle` is `Jasmine__TopLevel__Suite` and the `fullName` is `null`, so no alternative. Will fix this in another PR

PR #7677, opened 2021-11-10
Explicitly wait for window handle to be picked up in DevTools

### LINKED ISSUE #7619: [🐛 Bug]: Bug in `newWindow` makes hard to work with tabs and use functions such as switchWindow/switchToWindow/getWindowHandles
### WebdriverIO Version

7.16.3

### Node.js Version

14.16

### Mode

WDIO Testrunner

```typescript
browserName: chrome / chromium
browserVersion: 92
platform': linux / macos
automation protocol: devtools / puppeteer
```

### What happened?

We find the `wdio` API dealing with tabs and window management slightly confusing and not working in the way it is supposed to. 

We strongly believe that the problematic function is `newWidow`, which makes it hard to manage tabs and use other `wdio` APIs of `switchWindow`, `switchToWindow`, and `getWindowHandles`.

The first issue that find confusing is having two functions, `switchWindiow` and `switchToWindow`. They have similar names, basically responsible for the same functionality, but each has different arguments. It is probably better to have a single function that deals with switching windows/tabs. 

From the perspective of business logic and automation functionality, we present many scripts with inconsistent/confusing results. 

### Problematic code examples

**Example 1**

```ts
it.only('', async () => {
    await browser.url('https://google.com');
    const _0 = await browser.getWindowHandle();
    const _1 = await browser.newWindow('https://espn.com');
    const _2 = await browser.newWindow('https://netflix.com');

    console.log({
        _0,
        _1,
        _2,
    });
    await browser.pause(5000);
});
``` 

We would expect it to have three different handles, but they are all the same. So we can spread `await browser.wait(2000)`, which will have a different result between `_0, _1` and `_2`, but still incorrect. It all leads to a place where we cant use the `swithToWindow` properly. 

**Example 2**

```ts
it.only('', async () => {
    await browser.url('https://google.com');
    await browser.newWindow('https://www.si.com/', {
        windowName: 'si-window',
    });

    const siTitle = await browser.getTitle();
    console.log({ siTitle });
});
```
This one will result in 
> { siTitle: 'Google' }

The above is true for the `getTitle()` function and other browser and element functions. This can be solved by adding `await browser.pause(3000)`, and it seems that the page title and operations are correct after that. 

**Example 3**

```ts
it.only('', async () => {
    await browser.url('https://google.com');

    await browser.newWindow('https://www.si.com/', {
        windowName: 'si-window',
    });

    await browser.newWindow('https://espn.com', {
        windowName: 'espn-window',
    });

    await browser.switchWindow('si-window');
    console.log(await browser.getTitle());

    await browser.switchWindow('espn-window');
    console.log(await browser.getTitle());
});
```

It also won't work. Most cases would end with an error message like 

> No window found with title, URL or name matching "si-window"

Or messages similar to that one. We've tried to spread some `pause's in different combinations, but it won't help much. 

We have some more combinations of scripts, but most of them are going around the same point. 

### Our workaround and temporary solution

After many tries and retries, we came up with the following workaround that looks consistent. We present two solutions, one is trivial, and another is more elegant.

**Trivial Solution - Using `await browser.pause` and `await browser.getWindowHandle`**

```ts
async function newTab_firstTry(url: string): Promise<string> {
    await browser.newWindow(url);
    await browser.pause(3000);
    return browser.getWindowHandle();
}

it.only('', async () => {
    await browser.url('https://google.com');
    await browser.pause(3000);
    const google = await browser.getWindowHandle();

    const espn = await newTab_firstTry('https://www.espn.com');
    const si = await newTab_firstTry('https://www.si.com');
    const cnn = await newTab_firstTry('https://www.cnn.com');

    await browser.switchToWindow(google);
    console.log(await browser.getTitle());

    await browser.switchToWindow(espn);
    console.log(await browser.getTitle());

    await browser.switchToWindow(si);
    console.log(await browser.getTitle());

    await browser.switchToWindow(cnn);
    console.log(await browser.getTitle());

    console.log({
        google,
        espn,
        si,
        cnn,
    });
});
```

It seems to work fine and to produce the correct result. However, we are afraid of the hard-coded `browser.pause` and its dependency on it. Our experience shows us that using `sleep's` or `pause`s or explicit `wait`s can result in flakiness on ci/cd.

The problem with this solution is that `pause` is manually hard-coded, and our intuition and experiments show us that the update of the `getWindowHandles()` list depends on the page (how heavy is the page we load) and  

**Elegant solution - proposal for `newWindow` update**

```ts
async function waitForNewHandle() {
    const handlesBefore = await browser.getWindowHandles();
    getLogger().info(`Current handles [${handlesBefore.toString()}]`);

    await browser.waitUntil(
        async () => {
            const currentHandles = await browser.getWindowHandles();
            getLogger().info(
                `Current number of handles ${currentHandles.length}`
            );
            return currentHandles.length === handlesBefore.length + 1;
        },
        {
            timeout: 30000,
            timeoutMsg: `Expecting number of tabs to be ${
                handlesBefore.length + 1
            }, but it is ${handlesBefore.length}`,
            interval: 1000,
        }
    );
    getLogger().info('Done');
}

async function newTab_secondTry(url: string): Promise<string> {
    getLogger().info(`Opening new tab for: ${url}`);
    await browser.newWindow(url);
    await waitForNewHandle();
    return browser.getWindowHandle();
}

it.only('', async () => {
    await browser.url('https://google.com');
    await browser.pause(2000);

    const bbc = await newTab_secondTry('https://www.bbc.co.uk');
    const espn = await newTab_secondTry('https://www.espn.com');
    const si = await newTab_secondTry('https://www.si.com');
    const cnn = await newTab_secondTry('https://www.cnn.com');

    await browser.switchToWindow(bbc);
    console.log(await browser.getTitle());

    await browser.switchToWindow(espn);
    console.log(await browser.getTitle());

    await browser.switchToWindow(si);
    console.log(await browser.getTitle());

    await browser.switchToWindow(cnn);
    console.log(await browser.getTitle());

    console.log({
        bbc,
        espn,
        si,
        cnn,
    });
});
```

### Our thoughts on the problem

We believe that the problem is due to synchronization issues between the node script and the browser. Our intuition tells us that the `newWindow` operation is an expensive one, and it takes the browser time to create a new tab, load the page and update the handles. Meanwhile, the `newWindow` operation is not waiting on the browser and now waiting for the update state. Hence, it returns a bad handle, and the function `getWindowHandles()` is also inconsistent. 

We've checked the `webdriverio` code for how the `newWindow` function is implemented and we found out that the problem might occur in 

```ts
await this.execute(newWindowHelper, url, windowName, windowFeatures)

const newTab = tabs.pop()
```

The `this.execute` initiates the asynchronous `window.open`. The command of `this.execute` finishes while the asynchronous process of creating tab and loading page is still going on. Therefore, proceeding to `this.getWindowHandles` results in a place where `wdio` thinks that browser is done, but the browser is not done yet.  

The question is, whether the proposed solution above is a good solution or perhaps the correct solution is to use `devtools` protocol and puppeteer. We were surprised to know that the `newWindow` uses `window.open` instead of using the puppeteer / selenium. 

I wrote everything above. 

### How to reproduce the bug.

I wrote everything above.

### Relevant log output

```typescript
I wrote everything above.
```

### Code of Conduct

christian-bromann:
## Proposed changes

fixes #7619

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

n/a

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -42,9 +42,17 @@ jobs:
       - name: Lint / Dependencies / Unit / Smoke Tests
         run: npm run test
       - name: E2E Protocol Tests
-        run: npm run test:e2e
+        uses: nick-invision/retry@v2
+        with:
+          timeout_minutes: 20
+          max_attempts: 3
+          command: npm run test:e2e
       - name: E2E Edge Launch Tests
-        run: npm run test:e2e:edge
+        uses: nick-invision/retry@v2
+        with:
+          timeout_minutes: 20
+          max_attempts: 3
+          command: npm run test:e2e:edge
   e2e-mac:
     runs-on: macos-latest
     steps:
@@ -57,7 +65,11 @@ jobs:
       - name: Bootstrap Packages
         run: npm run setup-full
       - name: E2E Protocol Tests
-        run: npm run test:e2e
+        uses: nick-invision/retry@v2
+        with:
+          timeout_minutes: 20
+          max_attempts: 3
+          command: npm run test:e2e
       - name: E2E Firefox Launch Tests
         if: ${{ matrix.node-version == '16.x' }}
         run: |
--- a/packages/webdriverio/src/commands/browser/newWindow.ts
+++ b/packages/webdriverio/src/commands/browser/newWindow.ts
@@ -1,5 +1,9 @@
+import { sleep } from '@wdio/utils'
+
 import newWindowHelper from '../../scripts/newWindow'
 import type { NewWindowOptions } from '../../types'
+
+const WAIT_FOR_NEW_HANDLE_TIMEOUT = 3000
 
 /**
  *
@@ -58,10 +62,24 @@ export default async function newWindow (
         throw new Error('newWindow command is not supported on mobile platforms')
     }
 
+    const tabsBefore = await this.getWindowHandles()
     await this.execute(newWindowHelper, url, windowName, windowFeatures)
 
-    const tabs = await this.getWindowHandles()
-    const newTab = tabs.pop()
+    /**
+     * if tests are run in DevTools there might be a delay until
+     * a new window handle got registered, this little procedure
+     * waits for it to exist and avoid race conditions
+     */
+    let tabsAfter = await this.getWindowHandles()
+    const now = Date.now()
+    while ((Date.now() - now) < WAIT_FOR_NEW_HANDLE_TIMEOUT) {
+        tabsAfter = await this.getWindowHandles()
+        if (tabsAfter.length > tabsBefore.length) {
+            break
+        }
+        await sleep(100)
+    }
+    const newTab = tabsAfter.pop()
 
     if (!newTab) {
         throw new Error('No window handle was found to switch to')
--- a/packages/webdriverio/tests/commands/browser/newWindow.test.ts
+++ b/packages/webdriverio/tests/commands/browser/newWindow.test.ts
@@ -5,6 +5,8 @@
 import got from 'got'
 import { remote } from '../../../src'
 
+import type { Capabilities } from '@wdio/types'
+
 describe('newWindow', () => {
     beforeEach(() => {
         global.window.open = jest.fn()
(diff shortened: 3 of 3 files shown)


praveendvd (commented):
LGTM

avikha:
@christian-bromann 
Thank you! It is a super nice solution. 
I've learned so much both from the discussion we had and from the code. Thanks! 
This is a super nice project and I'd like to participate more when I'll have more time.

Issue #7640, opened 2021-11-02
[🐛 Bug]: run is not working when project already has a misconfigured tsconfig.json


Badisi:
### WebdriverIO Version

7.16.3

### Node.js Version

14.17.5

### Mode

Standalone Mode

### What happened?

The starter project generated with `@wdio/cli` does not run if a `tsconfig.json` already exists and is not properly configured.

### What is your expected behavior?

Everything should work without any issues after installing WebdriverIO on an existing project.

### How to reproduce the bug.

The original use case was to install `wdio` on an existing `angular` project.

But for the sake of this issue it can be reproduced as follow:

1) Create a folder `test` then go inside

    ```sh
    mkdir test
    cd test
    ```
2) Create a `tsconfig.json` file

    ```json
    {
      "compilerOptions": {
          "module": "es2020"
      }
    }
    ```
3) Install WebdriverIO with `typescript` as a compiler

    ```sh
    npx wdio .
    ```
4) Run the tests

    ```sh
    wdio run wdio.conf.ts
    ```
5) The following error occurs:
    > SyntaxError: Unexpected token 'export'

### Proposal

Actually, the `@wdio/cli` looks for a `tsconfig.json` file in the current directory and ignores it if it founds one or create one otherwise. But to me the problem is that existing typescript projects will always have a `tsconfig.json` file and those should only serve the main project itself. We should not use the same `tsconfig.json` for both the main project and the tests.

I think something like this would be more suitable:
```
project
├── test
│   ├── pageobjects
│   ├── specs
│   └── tsconfig.json
└ tsconfig.json
```

WebdriverIO would always look up for a `tsconfig.json` in the test folder.
This file would be created during each installation (no need to detect it anymore or to configure it manually afterwards).
And this file would always guarantee that everything is properly configured for wdio to work.

### Relevant log output

```typescript
export const config: WebdriverIO.Config = {
^^^^^^

SyntaxError: Unexpected token 'export'
    at wrapSafe (internal/modules/cjs/loader.js:988:16)
    at Module._compile (internal/modules/cjs/loader.js:1036:27)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1101:10)
    at Module.load (internal/modules/cjs/loader.js:937:32)
    at Function.Module._load (internal/modules/cjs/loader.js:778:12)
    at Module.require (internal/modules/cjs/loader.js:961:19)
    at require (internal/modules/cjs/helpers.js:92:18)
```

[...]

PR #7720, opened 2021-11-18
Fix for bug #7640  run is not working when project already has a misconfigured tsconfig.json

praveendvd:
## Proposed changes

Made changes to move creation of wdio.config and tsconfig to test folder if a tsconfig file already exists under the root project. Fix for [🐛 Bug]: run is not working when project already has a misconfigured tsconfig.json #7640 

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

Have to write tests


packages/wdio-cli/src/commands/config.ts
```diff
@@ -182,7 +189,9 @@ const runConfig = async function (useYarn: boolean, yes: boolean, exit = false)
     }
 
     try {
-        await renderConfigurationFile(parsedAnswers)
+        const configPath = hasFile('test/tsconfig.json') ? 'test' : undefined
+        answers.generateTSConfigFile ? await renderConfigurationFile(parsedAnswers, configPath) :
+            await renderConfigurationFile(parsedAnswers)
```
christian-bromann, line 194:
```suggestion
        await renderConfigurationFile(
            parsedAnswers,
            answers.generateTSConfigFile && hasFile('test/tsconfig.json') ? 'test' : undefined
        )
```

packages/wdio-cli/src/constants.ts
```diff
@@ -21,7 +21,7 @@ WDIO Configuration Helper
 export const CONFIG_HELPER_SUCCESS_MESSAGE = `
 Configuration file was created successfully!
 To run your tests, execute:
-$ npx wdio run wdio.conf.%s
+$ npx wdio run <path>/wdio.conf.%s
```
christian-bromann, line 24:
```suggestion
$ npx wdio run %s/wdio.conf.%s
```

christian-bromann: Then we need to update the `format` function where this const is used to add the path if conditions are met.

packages/wdio-cli/src/commands/config.ts
```diff
@@ -77,11 +77,18 @@ const runConfig = async function (useYarn: boolean, yes: boolean, exit = false)
...
+                    path.join(process.cwd(), 'tsconfig.json'),
+                    JSON.stringify(config, null, 4)
+                )
+            } else {
+                fs.ensureDirSync(path.join(process.cwd(), 'test'))
+                await fs.promises.writeFile(
+                    path.join(process.cwd(), 'test/tsconfig.json'),
+                    JSON.stringify(config, null, 4)
+                )
+            }
```
christian-bromann, line 91: we can simplify this via:

```suggestion
            await fs.promises.writeFile(
                path.join(process.cwd(), (hasFile('tsconfig.json') ? 'test' : ''), 'tsconfig.json'),
                JSON.stringify(config, null, 4)
            )
```

praveendvd:
@christian-bromann  I am not sure why the type failure here, is there anything i need to change in my commit ? it looks like non commit related issue

packages/wdio-cli/src/templates/wdio.conf.tpl.ejs
```diff
@@ -4,6 +4,34 @@
     // Runner Configuration
     // ====================
     //<%
+    if(answers.generateTSConfigFile && hasFile('test/tsconfig.json')){ %>
```
Badisi, line 7: Are we not supposed to generate this part no matter where the `tsconfig.json` is ?

For instance, if it's a new project, the `tsconfig.json` will be at the root of the project but it would still require `autoCompileOpts`, don't you think ?

praveendvd: It picks up that automatically if it's in root folder , only if it's in non root location then only we need to specify the tsconfig location

packages/wdio-cli/src/commands/config.ts
```diff
@@ -78,8 +78,9 @@ const runConfig = async function (useYarn: boolean, yes: boolean, exit = false)
                 }
             }
 
+            fs.ensureDirSync(path.join(process.cwd(), 'test'))
             await fs.promises.writeFile(
-                path.join(process.cwd(), 'tsconfig.json'),
+                path.join(process.cwd(), (hasFile('tsconfig.json') ? 'test' : ''), 'tsconfig.json'),
```
christian-bromann, line 83: I am seeing `hasFile('tsconfig.json')` here and in line 186, let's create a variable so we only need to call this function once.

packages/wdio-cli/src/constants.ts
```diff
@@ -284,6 +284,12 @@ export const QUESTIONNAIRE = [{
         : hasFile('tsconfig.json')
             ? COMPILER_OPTIONS.ts // default to TypeScript
             : COMPILER_OPTIONS.nil // default to no compiler
+}, {
+    type: 'confirm',
+    name: 'generateTSConfigFile',
+    message: 'Do you want WebdriverIO to autogenerate tsconfig file?',
```
christian-bromann, line 290: I wonder if this is the correct question to ask here. What happened if a user says `no`? They would see a bunch of TS errors due to a missing tsconfig, right?

packages/wdio-cli/src/commands/config.ts
```diff
@@ -65,7 +65,7 @@ const runConfig = async function (useYarn: boolean, yes: boolean, exit = false)
         if (!hasPackage('ts-node')) {
             packagesToInstall.push('ts-node', 'typescript')
         }
-        if (!hasFile('tsconfig.json')){
+        if (answers.generateTSConfigFile) {
```
christian-bromann, line 68: So I suggest to remove this if statement and always generate a tsconfig file.

christian-bromann (commented):
To simplify the code, how about we say: always generate a tsconfig to `./test/tsconfig.json`?

packages/wdio-cli/src/constants.ts
praveendvd: @christian-bromann  Yes thats why we added automatic creation of it ( in previous implementation ) . IN this case there is a console message saying what needs to be added to the tsconfig file so if user types  no still we are giving this console message so user have to add those to tsconfig file .

packages/wdio-cli/src/commands/config.ts
praveendvd: @christian-bromann 

what if there already exists a tsconfig file as the main issue mentioned. we need that check to know whether to create the tsconfig in current directory or inside test

packages/wdio-cli/src/constants.ts
christian-bromann: Why do we need this question then?

packages/wdio-cli/src/commands/config.ts
praveendvd: @christian-bromann  will do that 🥳

packages/wdio-cli/src/commands/config.ts
christian-bromann: Why don't we just always create a custom tsconfig file in `./test`? I think it makes the codebase easier and for the user it doesn't really matter, right?

packages/wdio-cli/src/constants.ts
praveendvd: @christian-bromann  now we removed the automatic creation , if user says yes then we creates the file in current or test folder depending on availability of already existing tsconfig file. else we wont do anything . do you want to remove the question ?

packages/wdio-cli/src/commands/config.ts
praveendvd: @christian-bromann  As mentioned before, if we create inside test , then wdio.config should be also inside test folder . So the structure ts project looks different than for all other compilers . 

in our previous comment it was decided that we will create inside test folder only if a tsconfig file exists in the root directory

praveendvd: https://github.com/webdriverio/webdriverio/issues/7640#issuecomment-972694037

christian-bromann: > then wdio.config should be also inside test folder . So the structure ts project looks different than for all other compilers 

We can just set [`autoCompileOpts `](https://webdriver.io/docs/typescript#configuration) to point to the correct `tsconfig.json` as you already implemented it in this PR.

> in our previous comment it was decided that we will create inside test folder only if a tsconfig file exists in the root directory

Correct.

praveendvd: @christian-bromann 

> in our previous comment it was decided that we will create inside test folder only if a tsconfig file exists in the root directory - correct

This is how its implemented now with autocompile option be added if tsconfig file is not in root . So i will keep it the way it is ? or you want tsconfig to be in test folder in all cases ?

packages/wdio-cli/src/constants.ts
christian-bromann: Yes, I think we clarified in https://github.com/webdriverio/webdriverio/issues/7640#issuecomment-972694037 that we always want to create a tsconfig file. Now I also suggest to put the tsconfig file always in `./test` because it simplifies the implementation and has no difference for the user. Then we don't need this question.

packages/wdio-cli/src/commands/config.ts
christian-bromann: > or you want tsconfig to be in test folder in all cases ?

Let's always put it in `/test` because it simplifies the implementation and it makes no difference for the user.

packages/wdio-cli/src/constants.ts
praveendvd: @christian-bromann will do that 🥳

christian-bromann: Awesome, sorry if I wasn't clear enough at the beginning.

packages/wdio-cli/src/commands/config.ts
```diff
@@ -78,8 +101,9 @@ const runConfig = async function (useYarn: boolean, yes: boolean, exit = false)
                 }
             }
 
+            fs.ensureDirSync(path.join(process.cwd(), 'test'))
```
christian-bromann: if we can use `await`, we should use it
```suggestion
            await fs.ensureDir(path.join(process.cwd(), 'test'))
```

packages/wdio-cli/src/commands/config.ts
```diff
@@ -58,14 +58,37 @@ const runConfig = async function (useYarn: boolean, yes: boolean, exit = false)
...
+    const parsedPaths = getPathForFileGeneration(answers)
+
+    const parsedAnswers: ParsedAnswers = {
+        ...answers,
+        runner: runnerPackage.short as 'local',
+        framework: frameworkPackage.short,
+        reporters: reporterPackages.map(({ short }) => short),
+        services: servicePackages.map(({ short }) => short),
+        packagesToInstall,
+        isUsingTypeScript: answers.isUsingCompiler === COMPILER_OPTIONS.ts,
```
christian-bromann, line 74:
```suggestion
        isUsingTypeScript: answers.isUsingCompiler === COMPILER_OPTIONS.ts,
        tsConfigFilePath: path.join(process.cwd(), 'test', 'tsconfig.json'),
```

packages/wdio-cli/src/commands/config.ts
```diff
@@ -78,8 +101,9 @@ const runConfig = async function (useYarn: boolean, yes: boolean, exit = false)
                 }
             }
 
+            fs.ensureDirSync(path.join(process.cwd(), 'test'))
             await fs.promises.writeFile(
-                path.join(process.cwd(), 'tsconfig.json'),
+                path.join(process.cwd(), 'test', 'tsconfig.json'),
```
christian-bromann:
```suggestion
                parsedAnswers.tsConfigFilePath,
```

packages/wdio-cli/src/commands/config.ts
```diff
@@ -158,31 +182,11 @@ const runConfig = async function (useYarn: boolean, yes: boolean, exit = false)
...
-        destPageObjectRootPath: parsedPaths.destPageObjectRootPath,
-        relativePath : parsedPaths.relativePath
-    }
-
     try {
-        await renderConfigurationFile(parsedAnswers)
+        await renderConfigurationFile(
+            parsedAnswers,
+            parsedAnswers.isUsingTypeScript ? 'test' : undefined
+        )
```
christian-bromann, line 189:
```suggestion
        await renderConfigurationFile(parsedAnswers)
```

packages/wdio-cli/src/utils.ts
```diff
@@ -218,11 +218,11 @@ export function convertPackageHashToObject(pkg: string, hash = '$--$'): Supporte
     }
 }
 
-export async function renderConfigurationFile (answers: ParsedAnswers) {
+export async function renderConfigurationFile (answers: ParsedAnswers, configPath = '') {
```
christian-bromann, line 221:
```suggestion
export async function renderConfigurationFile (answers: ParsedAnswers) {
```

packages/wdio-cli/src/utils.ts
```diff
@@ -218,11 +218,11 @@ export function convertPackageHashToObject(pkg: string, hash = '$--$'): Supporte
...
     const tplPath = path.join(__dirname, 'templates/wdio.conf.tpl.ejs')
     const filename = `wdio.conf.${answers.isUsingTypeScript ? 'ts' : 'js'}`
     const renderedTpl = await renderFile(tplPath, { answers })
-    return fs.promises.writeFile(path.join(process.cwd(), filename), renderedTpl)
+    return fs.promises.writeFile(path.join(process.cwd(), configPath, filename), renderedTpl)
```
christian-bromann, line 225:
```suggestion
    return fs.promises.writeFile(parsedAnswers.tsConfigFilePath, renderedTpl)
```

christian-bromann (commented):
More simplifcations

packages/wdio-cli/src/utils.ts
praveendvd: @christian-bromann  parsedAnswers.tsConfigFilePath will have somepath/tsconfig.json , how will we pass this to write File of config file ? i think we need to keep the path.join here

christian-bromann: ah right, nevermind then

packages/wdio-cli/src/commands/config.ts
```diff
@@ -158,31 +183,10 @@ const runConfig = async function (useYarn: boolean, yes: boolean, exit = false)
...
-        destSpecRootPath: parsedPaths.destSpecRootPath,
-        destPageObjectRootPath: parsedPaths.destPageObjectRootPath,
-        relativePath : parsedPaths.relativePath
-    }
-
     try {
-        await renderConfigurationFile(parsedAnswers)
+        await renderConfigurationFile(
+            parsedAnswers
+        )
```
christian-bromann, line 189:
```suggestion
        await renderConfigurationFile(parsedAnswers)
```

christian-bromann (approved):
👍

christian-bromann:
@praveendvd can you check the failing unit tests?

praveendvd:
@christian-bromann  the error is cannot read writeFile of undefined , but the fs.promise.writefile is already mocked in test (haven't changed anything )  . I am not sure why this error is thrown. 

Do you have any suggestions else I will debug tomorrow

praveendvd:
@christian-bromann  had to change the import fs to fs-extra will fix it

packages/wdio-cli/tests/commands/__snapshots__/config.test.ts.snap
```diff
@@ -299,6 +299,7 @@ Object {
     "services": Array [
       undefined,
     ],
+    "tsConfigFilePath": "/mnt/d/MyProjects/webdriverio/all/linux/webdriverio/test/tsconfig.json",
```
praveendvd, line 302: "tsConfigFilePath": "/home/runner/work/webdriverio/webdriverio/test/tsconfig.json",

praveendvd, line 302:
```suggestion
    "tsConfigFilePath": "/home/runner/work/webdriverio/webdriverio/test/tsconfig.json",
```

praveendvd:
@christian-bromann  The snap shot of linux has 

  "tsConfigFilePath": "/mnt/d/MyProjects/webdriverio/all/linux/webdriverio/test/tsconfig.json",


and for 

windows 

    "tsConfigFilePath": "D:\\a\\webdriverio\\webdriverio\\test\\tsconfig.json",


should i create a snapshot resolver ? or any other quick fix for this thats already implemented

christian-bromann:
@praveendvd remove the object from the snapshot

praveendvd:
@christian-bromann  The failure in mac is not this pr related i guess

praveendvd:
@christian-bromann  this error is only for this pull request i am not able to understand what the error means. It just says failed

packages/wdio-cli/src/utils.ts
```diff
@@ -222,7 +222,13 @@ export async function renderConfigurationFile (answers: ParsedAnswers) {
...
     const filename = `wdio.conf.${answers.isUsingTypeScript ? 'ts' : 'js'}`
     const renderedTpl = await renderFile(tplPath, { answers })
-    return fs.promises.writeFile(path.join(process.cwd(), filename), renderedTpl)
+    return fs.promises.writeFile(
+        path.join(
+            process.cwd(),
+            answers.isUsingTypeScript ? 'test' : '', filename
+        ),
+        renderedTpl
+    )
```
christian-bromann, line 231: I think we want to revert this change because we don't want the wdio.conf being generated within `/test/wdio.conf.ts`.

praveendvd: Hi @christian that was the discussion in previous comment to keep both config and tsconfig in test folder

christian-bromann: This must be a misunderstanding then. I remember we decided to have the `tsconfig.json` within the test folder but keep `wdio.conf.ts` in the root directory. With this change the current wdio.conf having this config:

```ts
        tsNodeOpts: {
            transpileOnly: true,
            project: 'test/tsconfig.json' // <-- wdio.conf needs to be in the root directory.
        }
```

wouldn't work. Therefore let's keep the wdio.conf in the root directory and the tsconfig.json in /test

praveendvd: Hi @chritian , the IDE will throw error if tsconfig and wdio config is not in the same directory . That's why we decided keep both in test folder for typescript

praveendvd: @christian-bromann  What do you think ? :(

christian-bromann (approved):
Thanks a lot 👍

Issue #7722, opened 2021-11-19
[🐛 Bug]: Enabling Sauce Connect w/ sauce-service with W3C EMUSIM capabilities presents invalid capability error


abSauce:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

7.5.0

### Node.js Version

15.14.0

### Mode

WDIO Testrunner

### Which capabilities are you using?

```typescript
{
        browserName: 'safari',
        platformName: 'iOS',
        "appium:platformVersion": '14.3',
        "appium:deviceName": 'iPhone XS Simulator',
        "sauce:options": {
            appiumVersion: '1.20.1',
            build: process.env.BUILD_NAME + build_info,
        }        
    }

or 

{
        browserName: 'Chrome',
        platformName: 'Android',
        "appium:platformVersion": '10.0',
        "apppium:deviceName": 'Android GoogleAPI Emulator',
        "sauce:options": {

            build: process.env.BUILD_NAME + build_info,
        }
    },
```

### What happened?

When attempting to use the provided W3C EMUSIM capabilities (as listed [here](https://saucelabs.com/platform/platform-configurator)) with Sauce Connect started using sauce-service (sauceConnect: true), the client returns this error in the logs.

`ERROR @wdio/runner: Error: Invalid or unsupported WebDriver capabilities found ("tunnelIdentifier"). Ensure to only use valid W3C WebDriver capabilities (see https://w3c.github.io/webdriver/#capabilities).
    at Object.startWebDriverSession (/Users/anthonybrown/Desktop/TrainingTests/mega-test/node_modules/webdriver/build/utils.js:38:19)
    at Function.newSession (/Users/anthonybrown/Desktop/TrainingTests/mega-test/node_modules/webdriver/build/index.js:58:59)
    at Object.remote (/Users/anthonybrown/Desktop/TrainingTests/mega-test/node_modules/webdriverio/build/index.js:67:43)
    at processTicksAndRejections (node:internal/process/task_queues:94:5)
    at async Runner._startSession (/Users/anthonybrown/Desktop/TrainingTests/mega-test/node_modules/@wdio/runner/build/index.js:216:56)
[...]

PR #7726, opened 2021-11-21
Fix 7722

wswebcreation:
## Proposed changes
This PR fixes https://github.com/webdriverio/webdriverio/issues/7722, it also:

- removes EMUSIM checks, EMUSIM is now W3C compatible
- removes legacy RDC, that one is decommissioned in Sauce
- updates tests for JWP/W3C

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments


packages/wdio-utils/tests/envDetector.test.ts
```diff
@@ -46,14 +46,17 @@ describe('sessionEnvironmentDetector', () => {
...
         expect(sessionEnvironmentDetector({ capabilities: chromeCaps, requestedCapabilities }).isW3C).toBe(true)
         expect(sessionEnvironmentDetector({ capabilities: geckoCaps, requestedCapabilities }).isW3C).toBe(true)
         expect(sessionEnvironmentDetector({ capabilities: safariCaps, requestedCapabilities }).isW3C).toBe(true)
         expect(sessionEnvironmentDetector({ capabilities: safariDockerNbVCaps, requestedCapabilities }).isW3C).toBe(true)
         expect(sessionEnvironmentDetector({ capabilities: safariDockerNpNCaps, requestedCapabilities }).isW3C).toBe(true)
         expect(sessionEnvironmentDetector({ capabilities: edgeCaps, requestedCapabilities }).isW3C).toBe(true)
+        /**
+         * Legacy / JWP responses
+         */
+        expect(sessionEnvironmentDetector({ capabilities: appiumCaps, requestedCapabilities }).isW3C).toBe(false)
```
wswebcreation, line 58: @christian-bromann 

The W3C check is used validating the caps before the driver starts and for returned caps from the driver (sessionEnvironmentDetector) The first one CAN properly check if the Appium response is W3C, the second one CAN NOT, at least not what I can find.

Should we have a different `sessionEnvironmentDetector` + `isW3C` check?

packages/wdio-utils/src/envDetector.ts
```diff
@@ -20,30 +20,31 @@ export function isW3C (capabilities?: Capabilities.DesiredCapabilities) {
...
+         * Appium can be JWP (Appium 1.x) or W3C (Appium 1.x and 2.x).
+         * JWP for Appium has `platformName` and `platformVersion` as valid
+         * caps which are W3C valid caps for Desktop.
          */
-        (capabilities.platformVersion || Object.prototype.hasOwnProperty.call(capabilities, 'setWindowRect'))
+        isMobile(capabilities) ?
+            /**
+             * The check `isMobile` is not enough to determine if Appium is running with W3C caps
+             */
+            Object.keys(capabilities).find((cap) => cap.startsWith('appium:')) :
```
christian-bromann, line 33: Can we move this into `isMobile` then?

packages/wdio-utils/src/envDetector.ts
```diff
@@ -20,30 +20,31 @@ export function isW3C (capabilities?: Capabilities.DesiredCapabilities) {
...
+                ((capabilities.platformName || capabilities.browserVersion) &&
+            /**
+             * local safari and BrowserStack don't provide platformVersion therefore
+             * check also if setWindowRect is provided
+             */
+            (capabilities.platformVersion || Object.prototype.hasOwnProperty.call(capabilities, 'setWindowRect')))
+            )
     )
-    return Boolean(hasW3CCaps || isAppium)
+    return Boolean(hasW3CCaps)
```
christian-bromann, line 47: Is there an Appium version that doesn't support any W3C? Afaik Appium (at least the last couple of versions) has W3C support. Now this might be different if it is used with a vendor e.g. Sauce Labs but then we explicitly have a check for that.

packages/wdio-utils/tests/envDetector.test.ts
christian-bromann: > The W3C check is used validating the caps before the driver starts

It is indeed a mixture out of received capabilities and requested ones. There is not really a deterministic way to define this, especially when using vendors like Sauce Labs.

What are the advantages of creating two different methods for it? The `sessionEnvironmentDetector` method gets both requested and received capabilities and therefor has everything to determine the environment.

wswebcreation: Well, there are a few things that we need to be aware of:

- Appium `1.x` is JWP/W3C compatible, but not strict on W3C caps. Meaning you can still send W3C commands to a session which is started with JWP caps
- Appium `2.x` is / should be W3C strict, also on the caps
- Appium `1.x` will be used for a long time by users / cloud vendors so we need to support that
- As fas as I know / form what I've seen is that Appium `1.x` uses caps that are JWP, but also W3C compatible (like `platformName`)
- As far as I've seen we can't rely on the `sessionEnvironmentDetector` to determine correctly if it was a W3C Appium session (from what I've seen the `appium:` prefixes are stripped out in the session response

I'm not sure if we should handle that by a single W3C method that is used for both received capabilities and requested ones, but I can be wrong 😉

packages/wdio-utils/src/envDetector.ts
wswebcreation: I don't think so, because `isMobile` does not determine if it's mobile AND W3C because the starting caps can still be JWP, which in case of Sauce should lead to `tunnelIdentifier` instead of `'sauce:options': {tunnelIdentifier:'tunnelname'}`

packages/wdio-utils/src/envDetector.ts
wswebcreation: It has W3C support but as mentioned above there is a diff between how it was started (JWP caps or W3C caps) and version `1.x` and `2.x`

packages/wdio-utils/src/envDetector.ts
christian-bromann: but if someone applies Appium specific capabilities, we can assume that they want to start a mobile Appium session correct? So `isMobile` should return true?

packages/wdio-utils/src/envDetector.ts
christian-bromann: So what you are saying is that if someone uses Appium e.g. using Sauce Labs then WebdriverIO should pick JWP because of lack of support for W3C? In this case as I mentioned above it would be great to have a `if saucelabs && ...` statement to not generalise this.

wswebcreation: Hmm, I'm not talking about the lack of support for W3C, I'm saying that `isMobile` is not per definition W3C, not with how it was started and not with the commands it receives. For Appium `2.x` we can say it's W3C, `1.x` is just a "mess" with being able to support both / a mix

packages/wdio-utils/src/envDetector.ts
wswebcreation: Yes, `isMobile` determines if it's mobile, bot not if it's a W3C session

packages/wdio-utils/src/envDetector.ts
christian-bromann: > For Appium 2.x we can say it's W3C, 1.x is just a "mess" with being able to support both / a mix

What is the intention of this change then. Yes, with Appium 1.x it is kind of ambiguous and currently WebdriverIO assumes W3C support is somewhat given even though there might be issues in some cases.

packages/wdio-utils/src/envDetector.ts
christian-bromann: Correct. I think we should generalise this then and check if there are any capabilities that have the format `%s:%s`

packages/wdio-utils/src/envDetector.ts
wswebcreation: The `isW3C` method is also used in the Sauce Service, but then on the request parameters. By validating it's mobile and W3C it's not Legacy (for Sauce) and thus needs to add the `sauce:options` in the `launcher`, see also https://github.com/webdriverio/webdriverio/blob/f52c8a5851d965255051602ab9d8edf31acf2ba2/packages/wdio-sauce-service/src/utils.ts#L65

packages/wdio-utils/src/envDetector.ts
wswebcreation: 🤔 , with a generic method it would still not say if it would be a _valid_ Appium W3C session. 
If for example `goog:chromeOptions` or `sauce:options` are used it's still not a valid Appium W3C session. A valid one would be **with** `appium:%s`, if they provide other wrong options which should have been in the format `%s:%s` then it's the driver/vendor who should fail

wswebcreation:
@christian-bromann 

What's the best way to move forward on this one?

packages/wdio-sauce-service/tests/service.test.ts
```diff
@@ -504,7 +504,7 @@ test('_uploadLogs should not fail in case of a platform error', async () => {
     const service = new SauceService(
         {},
         {},
-        { outputDir: '/foo/bar' } as any
+            { outputDir: '/foo/bar' } as any
```
christian-bromann, line 507:
```suggestion
        { outputDir: '/foo/bar' } as any
```

packages/wdio-utils/src/envDetector.ts
christian-bromann: The goal of this method (`isW3C`) is to find out whether or not the session is a W3C session or not, no matter if run on mobile or desktop. My suggestion was to check if any of the keys within the session response have the following format: `%s:%s`. If one of them has this format we have to assume W3C compatibility.

christian-bromann: Also can we format this to be:
```js
const hasW3CCaps = <condition>
    ? <value if true>
    : <value if false>
```
makes it easier to read IMO. I was overseeing the `:` at end of that one line.

packages/wdio-utils/src/envDetector.ts
christian-bromann: > The isW3C method is also used in the Sauce Service, but then on the request parameters.

This is a problem. We should modify this part then.

wswebcreation: Will do, thanks

packages/wdio-utils/src/envDetector.ts
wswebcreation: @christian-bromann 

I've checked the session response data of a view mobile sessions and also from desktop and I didn't see any `%s:%s` coming back. Should this be returned in a W3C response?

wswebcreation:
Closing this for now. Sauce will release W3C support on RDC at the beginning of December so we will then also need to have a new fix.
I'll work on figuring out how to solve the Appium part the best way

Issue #7544, opened 2021-10-12
Display detailed stack trace on ETIMEDOUT error


DudaGod:
**Is your feature request related to a problem? Please describe.**

Hello.

In case when request to browser hangs for some reason I get an error like this:
```
2021-10-12T12:43:52.302Z ERROR webdriver: Request failed with status undefined due to TimeoutError: Timeout awaiting 'request' for 5000ms
(node:98332) UnhandledPromiseRejectionWarning: RequestError: Timeout awaiting 'request' for 5000ms
// ...
```

This detailed stack trace does not tell me which command was called when an error occurred and with what arguments it was called. So I suggest to add this information to stack trace something here - https://github.com/webdriverio/webdriverio/blob/main/packages/webdriver/src/request/index.ts#L205-L207.

I can do it myself if you agree that improving the stack trace in this case is really useful.

Example of the test:
```
const wdio = require('webdriverio');

let client;

(async function () {
    client = await wdio.remote({
        protocol: 'http',
        hostname: 'localhost',
        port: 4444,
        path: '/wd/hub',
        automationProtocol: 'webdriver',

        capabilities: {
            browserName: 'chrome',
            browserVersion: '85.0',
        },

        connectionRetryTimeout: 5000,
        connectionRetryCount: 0,
        logLevel: 'error'
    });

    await client.url('http://localhost:3000');
    await client.deleteSession();
})()
    .catch(async (e) => {
        console.error(e.stack)
        await client.deleteSession()
    })
```

My slow server to reproduce the problem:
```
const express = require('express');
const Promise = require('bluebird');
const app = express();
const port = 3000;

app.get('/', async (req, res) => {
    await Promise.delay(60000);
    res.send('Hello World!');
});

app.listen(port, () => {
  console.log(`Example app listening at http://localhost:${port}`)
});
```

If I change `logLevel` to `trace` for example then I can see in the log which command failed. Example:
```
2021-10-12T12:55:35.157Z INFO webdriver: COMMAND navigateTo("http://localhost:3000/")
2021-10-12T12:55:35.157Z INFO webdriver: [POST] http://localhost:4444/wd/hub/session/c3a9593b15c3368f983a39cdaa506d4d/url
2021-10-12T12:55:35.157Z INFO webdriver: DATA { url: 'http://localhost:3000/' }
2021-10-12T12:55:40.167Z ERROR webdriver: Request failed with status undefined due to TimeoutError: Timeout awaiting 'request' for 5000ms
[...]

PR #7753, opened 2021-11-25
feat(webdriver): extend timeout error with info about exec command

DudaGod:
## Proposed changes

Detailed information about the problem is described here - https://github.com/webdriverio/webdriverio/issues/7544.

After these changes, the timeout error of one of the commands will look like this:
```
2021-11-25T09:30:07.853Z ERROR webdriver: Request failed with status undefined due to TimeoutError: Timeout awaiting 'request' for 5000ms when running "url" with method "POST" and args "{"url":"http://localhost:3000/"}"
(node:45219) UnhandledPromiseRejectionWarning: RequestError: Timeout awaiting 'request' for 5000ms when running "url" with method "POST" and args "{"url":"http://localhost:3000/"}"
// ... stack trace
```

Before the changes:
```
2021-11-25T09:30:07.853Z ERROR webdriver: Request failed with status undefined due to TimeoutError: Timeout awaiting 'request' for 5000ms
(node:45219) UnhandledPromiseRejectionWarning: RequestError: Timeout awaiting 'request' for 5000ms
// ... stack trace
```

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/packages/webdriver/src/request/index.ts
+++ b/packages/webdriver/src/request/index.ts
@@ -7,7 +7,7 @@ import logger from '@wdio/logger'
 import { transformCommandLogResult } from '@wdio/utils'
 import type { Options } from '@wdio/types'
 
-import { isSuccessfulResponse, getErrorFromResponseBody } from '../utils'
+import { isSuccessfulResponse, getErrorFromResponseBody, getTimeoutError } from '../utils'
 
 const pkg = require('../../package.json')
 
@@ -203,7 +203,9 @@ export default abstract class WebDriverRequest extends EventEmitter {
              * handle timeouts
              */
             if ((response as RequestLibError).code === 'ETIMEDOUT') {
-                return retry(response, 'Request timed out! Consider increasing the "connectionRetryTimeout" option.')
+                const error = getTimeoutError(response, fullRequestOptions)
+
+                return retry(error, 'Request timed out! Consider increasing the "connectionRetryTimeout" option.')
             }
 
             /**
--- a/packages/webdriver/src/utils.ts
+++ b/packages/webdriver/src/utils.ts
@@ -10,7 +10,8 @@ import { Options, Capabilities } from '@wdio/types'
 import RequestFactory from './request/factory'
 import { WebDriverResponse } from './request'
 import command from './command'
-import { VALID_CAPS } from './constants'
+import { transformCommandLogResult } from '@wdio/utils'
+import { VALID_CAPS, REG_EXPS } from './constants'
 import type { JSONWPCommandError, SessionFlags } from './types'
 
 const log = logger('webdriver')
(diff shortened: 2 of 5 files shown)


DudaGod:
@christian-bromann Hello. Can you look at this PR when you have some free time?

packages/webdriver/tests/request.test.ts
```diff
@@ -399,24 +400,44 @@ describe('webdriver request', () => {
...
+                const spy = jest.spyOn(utils, 'extendTimeoutErrByExecCmd')
+
+                const req = new WebDriverRequest('GET', '/timeout', {}, true)
+                const reqOpts = {
+                    protocol: 'https',
+                    hostname: 'localhost',
+                    port: 4445,
+                    path: '/',
+                } as Options.WebDriver
+                await req.makeRequest(reqOpts, 'foobar').then((res) => res, (e) => e)
```
christian-bromann, line 434: There is nothing we do with the result here, so:
```suggestion
                await req.makeRequest(reqOpts, 'foobar')
```

packages/webdriver/src/request/index.ts
```diff
@@ -203,6 +203,7 @@ export default abstract class WebDriverRequest extends EventEmitter {
              * handle timeouts
              */
             if ((response as RequestLibError).code === 'ETIMEDOUT') {
+                extendTimeoutErrByExecCmd(response, fullRequestOptions)
                 return retry(response, 'Request timed out! Consider increasing the "connectionRetryTimeout" option.')
```
christian-bromann, line 207: I suggest instead of modifying the response error object, to return a new error we can continue to use. Maybe let's rename `extendTimeoutErrByExecCmd` to `getTimeoutError`
```suggestion
                const error = extendTimeoutErrByExecCmd(response, fullRequestOptions)
                return retry(error, 'Request timed out! Consider increasing the "connectionRetryTimeout" option.')
```

packages/webdriver/tests/request.test.ts
DudaGod: I make request to `/timeout` route which is mocked and always return timeout error. I can rewrite to `try catch` but in adjacent tests used solution with `then`.

christian-bromann: Maybe then something like this?
```suggestion
                await req.makeRequest(reqOpts, 'foobar')
                    // ignore error
                    .catch((e) => e)
```

DudaGod:
fixed

packages/webdriver/src/utils.ts
```diff
@@ -323,3 +324,46 @@ export const getSessionError = (err: JSONWPCommandError, params: Partial<Options
...
+ */
+export const getTimeoutError = (error: Error, requestOptions: Options.RequestLibOptions): Error => {
+    const cmdName = getExecCmdName(requestOptions)
+    const cmdArgs = getExecCmdArgs(requestOptions)
+
+    const cmdInfoMsg = `when running "${cmdName}" with method "${requestOptions.method}"`
+    const cmdArgsMsg = cmdArgs ? ` and args ${cmdArgs}` : ''
+
+    const timeoutErr = new Error(`${error.message} ${cmdInfoMsg}${cmdArgsMsg}`)
+    return Object.assign(timeoutErr, error)
```
DudaGod, line 339: used for add special fields like `name`, `code`, `timings`, `event`

christian-bromann (approved):
LGTM 👍 

Thanks for the contribution!

DudaGod:
@christian-bromann can you tell when the version with this change will be released?

christian-bromann:
@DudaGod thanks for pinging, just triggered a new release. It will be part of `v6.10.11`.

Issue #7709, opened 2021-11-17
[💡 Feature]: Allow second parameter to getCSSProperty to target pseudo elements, e.g. '::after'


nikolajl:
### Is your feature request related to a problem?

I need to check a css style that is implemented as pseudo element `::after`.

In the browser this is possible using `window.getComputedStyle(element, '::after')`. However, there is not an obvious way to do this in on the Element in WebdriverIO.

Under the hood, `getCSSProperty()` uses getComputedStyle, but only allows 1 parameter. Allowing a second parameter would be an obvious simple solution.

### Describe the solution you'd like.

Add second parameter to `getCSSProperty()`

Example:
```
const contentAfter = await $('#selector').getCSSProperty('content', '::after')
```

Parameter would be added here.
https://github.com/webdriverio/webdriverio/blob/f8584110532a33ad899ff66eefde427d2d0c7481/packages/webdriverio/src/commands/element/getCSSProperty.ts#L73
```
export default async function getCSSProperty (
    this: WebdriverIO.Element,
    cssProperty: string,
    pseudoElement?: string // or '::before' | '::after'
) {
```

and then passed on to this one:
https://github.com/webdriverio/webdriverio/blob/f8584110532a33ad899ff66eefde427d2d0c7481/packages/devtools/src/scripts/getElementCSSValue.ts#L5
```
export default function getElementCSSValue (_: HTMLElement, elem: HTMLElement, propertyName: string, pseudoElement?: string ): string {
    return (window.getComputedStyle(elem, pseudoElement) as any)[propertyName]
}
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

christian-bromann:
Thanks for proposing the idea @nikolajl! I think the structure you suggest works out, maybe we can be a bit more stricter about the pseudo element strings? So rather than defining it as `string`, we could list `'::after' | '::before' | etc`.

Any contributions that implements this suggestion are highly appreciated. Please take a look into our [contribution guidelines](https://webdriver.io/docs/contribute) and let us know if you have any questions. Cheers!

Andrewnt219:
Hi @christian-bromann, I want to help with this issue.

christian-bromann:
Awesome, let me know if you have any questions.

Andrewnt219:
[...]

PR #7754, opened 2021-11-25
feat(webdriverio): support pseudo element in getCSSProperty

Andrewnt219:
## Proposed changes

Refs: #7709

`getCSSProperty` utilizes browser `getComputedStyle` which allows
passing a second parameter to get the pseudo element instead of
the real element. So I just pass the pseudo element down the call chain.

The type defined by `getComputedStyle` for pseudo elements is
`string | null | undefined` so I do the same. Constrain the type
to literal string can be hard to maintain.

## Types of changes

- [x] Documentation update

## Checklist

## Further comments

I want to add the test for the new feature as well but have trouble figuring out how. I guess it is something from Web Platform Tests, but still don't know how I can create a pseudo-element to test with.

This is the test suit I'm taking references from
https://github.com/webdriverio/webdriverio/blob/1f059837311bb05ecbf7a8442cfb57b04036be29/packages/webdriverio/tests/commands/element/getCSSProperty.test.ts#L8-L14

### DIFF
--- a/packages/wdio-protocols/protocols/webdriver.json
+++ b/packages/wdio-protocols/protocols/webdriver.json
@@ -679,7 +679,14 @@
         "name": "propertyName",
         "description": "name of the CSS property to retrieve"
       }],
-      "parameters": [],
+      "parameters": [
+        {
+          "type": "String|Null",
+          "required": false,
+          "name": "pseudoElement",
+          "description": "the pseudo-element to match. Omitted (or null) for real elements"
+        }
+      ],
       "returns": {
         "type": "string",
         "name": "cssValue",
--- a/packages/webdriverio/src/commands/element/getCSSProperty.ts
+++ b/packages/webdriverio/src/commands/element/getCSSProperty.ts
@@ -19,6 +19,12 @@ import { parseCSS } from '../../utils'
  * <example>
     :example.html
     <label id="myLabel" for="input" style="color: #0088cc; font-family: helvetica, arial, freesans, clean, sans-serif, width: 100px">Some Label</label>
+    <style>
+        #myLabel::after {
+            content: "";
+            height: 1px;
+        }
+    </style>
     :getCSSProperty.js
     it('should demonstrate the getCSSProperty command', async () => {
         const elem = await $('#myLabel')
@@ -62,17 +68,33 @@ import { parseCSS } from '../../utils'
         //         value: 100
         //     }
         // }
+
+        var height = await elem.getCSSProperty('height', '::after');
+        console.log(height);
+        // outputs the following:
+        // {
+        //     property: 'height',
+        //     value: '1px',
+        //     parsed: {
+        //         type: 'number',
+        //         string: '1px',
+        //         unit: 'px',
+        //         value: 1
+        //     }
+        // }
     })
  * </example>
  *
  * @alias element.getCSSProperty
  * @param  {String}      cssProperty css property name
+ * @param  {?String}      [pseudoElement] the pseudo-element to match. Omitted (or null) for real elements
  * @return {CSSProperty}             The specified css of the element
  *
  */
 export default async function getCSSProperty (
     this: WebdriverIO.Element,
-    cssProperty: string
+    cssProperty: string,
+    pseudoElement?: string | null
 ) {
     /**
      * Getting the css value of a shorthand property results in different results
@@ -81,13 +103,13 @@ export default async function getCSSProperty (
      * otherwise expand it and run the command for each longhand property.
      */
     if (!cssShorthandProps.isShorthand(cssProperty)) {
-        const cssValue = await this.getElementCSSValue(this.elementId, cssProperty)
+        const cssValue = await this.getElementCSSValue(this.elementId, cssProperty, pseudoElement)
         return parseCSS(cssValue, cssProperty)
     }
 
     const properties = cssShorthandProps.expand(cssProperty)
     let cssValues = await Promise.all(
-        properties.map((prop) => this.getElementCSSValue(this.elementId, prop))
+        properties.map((prop) => this.getElementCSSValue(this.elementId, prop, pseudoElement))
     )
 
     /**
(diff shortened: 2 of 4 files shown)


packages/wdio-protocols/protocols/webdriver.json
```diff
@@ -679,7 +679,14 @@
...
       }],
-      "parameters": [],
+      "parameters": [
+        {
+          "type": "String|Null",
+          "required": false,
+          "name": "pseudoElement",
+          "description": "the pseudo-element to match. Omitted (or null) for real elements"
+        }
+      ],
```
christian-bromann, line 689: Can you provide a reference where this is defined within the [WebDriver spec](https://w3c.github.io/webdriver/webdriver-spec.html)?

packages/webdriverio/src/commands/element/getCSSProperty.ts
```diff
@@ -62,17 +68,33 @@ import { parseCSS } from '../../utils'
...
  * @param  {String}      cssProperty css property name
+ * @param  {?String}      [pseudoElement] the pseudo-element to match. Omitted (or null) for real elements
  * @return {CSSProperty}             The specified css of the element
  *
  */
 export default async function getCSSProperty (
     this: WebdriverIO.Element,
-    cssProperty: string
+    cssProperty: string,
+    pseudoElement?: string | null
```
christian-bromann, line 97: Let's make the second argument and object as it allows us to easier extend it moving on with other parameters. Also, let's ensure to have a strict typing. There are just a handful of possible pseudo elements defined, so rather than defining it as `string` we should define it it as a list of strings, e.g. `'::after' | ...`

packages/webdriverio/src/commands/element/getCSSProperty.ts
```diff
@@ -81,13 +103,13 @@ export default async function getCSSProperty (
      * otherwise expand it and run the command for each longhand property.
      */
     if (!cssShorthandProps.isShorthand(cssProperty)) {
-        const cssValue = await this.getElementCSSValue(this.elementId, cssProperty)
+        const cssValue = await this.getElementCSSValue(this.elementId, cssProperty, pseudoElement)
```
christian-bromann, line 106: The WebDriver spec afaik doesn't allow to fetch for pseudo elements. A way to work around this would be to check whether this option is given and if so, call `execute` and fetch the css property through running the same script we have in the devtools package.

christian-bromann (changes requested):
Once commends are addressed it would be great to also add some [unit/smoke tests](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#test-changes).

christian-bromann:
> `getCSSProperty` utilizes browser `getComputedStyle`

Only when you run through the `devtools` package, aka. running tests with Puppeteer rather than WebDriver.

> I want to add the test for the new feature as well but have trouble figuring out how. I guess it is something from Web Platform Tests, but still don't know how I can create a pseudo-element to test with.

This feature is currently not supported in WebDriver. My suggestion is to work around it by calling `execute` and fetch the css property via JS script rather than using a WebDriver command. At the end of the day we expect some sort of web environment if we are interested in a CSS value so I don't see any combat issue with this.

packages/webdriverio/src/commands/element/getCSSProperty.ts
Andrewnt219: So something like this?

```ts
const cssValue = this.execute(`window.getComputedStyle(${this.elementId}, ${pseudoElement})[${cssProperty}]`)
return parseCSS(cssValue, cssProperty)

```

packages/webdriverio/src/commands/element/getCSSProperty.ts
Andrewnt219: > Let's make the second argument and object

Do you mean...

```ts
export default async function getCSSProperty (
    this: WebdriverIO.Element,
    { cssProperty, pseudoElement }: {cssProperty: string, pseudoElement?: string | null }
) {
```

... or

```ts
export default async function getCSSProperty (
    this: WebdriverIO.Element,
    cssProperty: string,
    { pseudoElement } : { pseudoElement?: string | null }
) {
```

packages/wdio-protocols/protocols/webdriver.json
Andrewnt219: Do I add it in the description, ref doesn't seem to work?
```diff
- "description": "the pseudo-element to match. Omitted (or null) for real elements" 
+ "description": "the pseudo-element to match. Omitted (or null) for real elements. See: https://drafts.csswg.org/cssom/#dom-window-getcomputedstyle"
```

Andrewnt219:
> This feature is currently not supported in WebDriver. My suggestion is to work around it by calling execute and fetch the css property via JS script rather than using a WebDriver command. At the end of the day we expect some sort of web environment if we are interested in a CSS value so I don't see any combat issue with this.

How do I make a pseudo-element to use for testing? Do I add it by JS as well?

packages/wdio-protocols/protocols/webdriver.json
christian-bromann: That is not what I meant here. You are defining a command for the WebDriver spec. However the spec has to ability to fetch pseudo elements. WebdriverIO generates its command set from these information and if someone would use this parameter it would not work.

packages/webdriverio/src/commands/element/getCSSProperty.ts
christian-bromann: no, this works:

```js
const cssValue = this.execute(
    (elem, pseudoElement, cssProperty) => window.getComputedStyle(elem, pseudoElement)[cssProperty],
    element,
    pseudoElement,
    cssProperty
)
return parseCSS(cssValue, cssProperty)
```

christian-bromann:
> How do I make a pseudo-element to use for testing? Do I add it by JS as well?

not sure what you mean here

Andrewnt219:
Sorry, I probably don't have enough experience to solve the issue, so I'll close this PR. Thank you so much for your guidance @christian-bromann.

PR #7779, opened 2021-12-03
feat: add cucumber step name in Sauce

wswebcreation:
## Proposed changes
Cucumber has a feature, scenario and a step. We now only push the feature and scenario name to Sauce Labs. This PR will also push the test step. This makes it "easier" to debug which command belongs to which step, see also below.

**NOTE:** I've added `--` to have an _indentation_ in Sauce, spaces will be trimmed and `&nbsp;` will not be translated to spaces in Sauce Labs

image

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments
-

### DIFF
--- a/packages/wdio-sauce-service/src/service.ts
+++ b/packages/wdio-sauce-service/src/service.ts
@@ -230,7 +230,7 @@ export default class SauceService implements Services.ServiceInstance {
             return
         }
 
-        return this.setAnnotation(`sauce:context=Feature:${this._suiteTitle}`)
+        return this.setAnnotation(`sauce:context=Feature: ${this._suiteTitle}`)
     }
 
     beforeScenario (world: Frameworks.World) {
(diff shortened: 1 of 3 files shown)


packages/wdio-sauce-service/src/service.ts
```diff
@@ -230,7 +230,7 @@ export default class SauceService implements Services.ServiceInstance {
             return
         }
 
-        return this.setAnnotation(`sauce:context=Feature:${this._suiteTitle}`)
+        return this.setAnnotation(`sauce:context=Feature: ${this._suiteTitle}`)
```
erwinheitzman, line 233: is this added space intentional?

erwinheitzman:
How about we add a single `-` for Scenario's so it becomes a logical tree like so:
```
Feature
-Scenario
--Step
```

wswebcreation:
> How about we add a single `-` for Scenario's so it becomes a logical tree like so:
> 
> ```
> Feature
> -Scenario
> --Step
> ```

Looks fine for me, I'll fix that and push it

erwinheitzman (approved):

christian-bromann (approved):
Very nice 👍

Issue #7580, opened 2021-10-21
[🐛 Bug]: isExisting calls findElements (multiple), a potentially expensive op


nextlevelbeard:
### WebdriverIO Version

latest

### Node.js Version

14

### Mode

WDIO Testrunner

### Which capabilities are you using?

```shell
{
  platformName: "iOS"
}
```

### What happened?

Like the title says, under the hood, webdriverIO calls findElements (multiple) when using `isExisting` and `waitForExist`.
https://github.com/webdriverio/webdriverio/blob/f1313735cd3441dc172d3955479575854e64a306/packages/webdriverio/src/commands/element/isExisting.ts#L58

Only `findElement` (single) calls should be needed to assert whether or not the element / selector exists.

This becomes a problem because `findElements` calls can be quite expensive.
An example would be large scrollable lists on iOS apps, they can take 30+ seconds.

### What is your expected behavior?

No call to `findElements` should be made on isExisting as it is not needed and in fact can needlessly hang the command.

### How to reproduce the bug.

Context:
I'm testing a mobile app with Appium, testing a scrollable list.
Scrollable lists on iOS always possess all the list items, unlike Android which loads the list items into view as you scroll/swipe up and down. Therefore finding multiple elements on iOS can lead to a time consuming operation for large lists. In the same situation however, single findElement call is cheap and fast. Assume a `findElements` call takes 30+ seconds to resolve and we are trying to purposely avoid it

The flow goes something like this:
1. Wait for the (scrollable) list container
2. Then wait for the list items to load within the scrollable list
3. Irrelevant, do whatever else, expectation, whatever

This translates to some code resembling this:

```js
get list() {
  return $('id=list');
}

get items() {
  return $$('id=list-item');
}

waitForItems = (title) => {
  // We previously already waited for the list container
  // We want to wait for the list items to load within the scrollable list
  // Assume they appear instantly

  // Option 1, takes 30+ seconds because of findElements call
  browser.waitUntil(() => this.items.length);

  // Option 2, takes 30+ seconds because of findElements call
  browser.waitUntil(() => $('id=list-item').isExisting());

  // Option 3, takes 30+ seconds because of findElements call
  $('id=list-item').waitForExist();

  // Workaround to save time
  // Option 4, findElements not called, findElement called, 2 seconds
  browser.waitUntil(() => !!$('id=list-item').elementId);
}
```

### Relevant log output

No relevant logs.

### Code of Conduct

[...]

PR #7816, opened 2021-12-13
fix(#7580): update code with the suggested solution

tmjpedro:
## Proposed changes

fixes https://github.com/webdriverio/webdriverio/issues/7580

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/webdriverio/src/commands/element/isExisting.ts
+++ b/packages/webdriverio/src/commands/element/isExisting.ts
@@ -62,5 +62,5 @@ export default async function isExisting (this: WebdriverIO.Element) {
     }
 
     const command = this.isReactElement ? this.parent.react$.bind(this.parent) : this.parent.$.bind(this.parent)
-    return command(this.selector as string).then((res) => Boolean(res.elementId))
+    return command(this.selector as string).then((res) => Boolean(res.elementId), () => false)
 }
--- a/packages/webdriverio/src/commands/element/isExisting.ts
+++ b/packages/webdriverio/src/commands/element/isExisting.ts
@@ -61,6 +61,6 @@ export default async function isExisting (this: WebdriverIO.Element) {
         )
     }
 
-    const command = this.isReactElement ? this.parent.react$$.bind(this.parent) : this.parent.$$.bind(this.parent)
-    return command(this.selector as string).then((res) => res.length > 0)
+    const command = this.isReactElement ? this.parent.react$.bind(this.parent) : this.parent.$.bind(this.parent)
+    return command(this.selector as string).then((res) => Boolean(res.elementId), () => false)
 }
--- a/packages/webdriverio/src/commands/element/isExisting.ts
+++ b/packages/webdriverio/src/commands/element/isExisting.ts
@@ -64,5 +64,5 @@ export default async function isExisting (this: WebdriverIO.Element) {
     const command = this.isReactElement
         ? this.parent.react$.bind(this.parent)
         : this.parent.$.bind(this.parent)
-    return command(this.selector as string).then(async (res: Promise<WebdriverIO.Element>) => Boolean((await res).elementId), () => false)
+    return Boolean((await command(this.selector as string)).elementId)
 }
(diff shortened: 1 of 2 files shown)


packages/webdriverio/src/commands/element/isExisting.ts
```diff
@@ -61,6 +61,6 @@ export default async function isExisting (this: WebdriverIO.Element) {
         )
     }
 
-    const command = this.isReactElement ? this.parent.react$$.bind(this.parent) : this.parent.$$.bind(this.parent)
-    return command(this.selector as string).then((res) => res.length > 0)
+    const command = this.isReactElement ? this.parent.react$.bind(this.parent) : this.parent.$.bind(this.parent)
+    return command(this.selector as string).then((res) => Boolean(res.elementId))
```
christian-bromann, line 65: Otherwise the command would raise an error rather than returning `false`

```suggestion
    return command(this.selector as string).then((res) => Boolean(res.elementId), () => false)
```

christian-bromann (changes requested):

packages/webdriverio/src/commands/element/isExisting.ts
tmjpedro: done, thanks 💪

packages/webdriverio/src/commands/element/isExisting.ts
```diff
@@ -61,6 +61,6 @@ export default async function isExisting (this: WebdriverIO.Element) {
         )
     }
 
-    const command = this.isReactElement ? this.parent.react$$.bind(this.parent) : this.parent.$$.bind(this.parent)
-    return command(this.selector as string).then((res) => res.length > 0)
+    const command = this.isReactElement ? this.parent.react$.bind(this.parent) : this.parent.$.bind(this.parent)
+    return command(this.selector as string).then((res : any) => Boolean(res.elementId), () => false)
```
christian-bromann, line 65: We should avoid using `any`.

```suggestion
    return command(this.selector as string).then((res: WebdriverIO.Element) => Boolean(res.elementId), () => false)
```

christian-bromann (changes requested):

packages/webdriverio/src/commands/element/isExisting.ts
```diff
@@ -61,6 +61,13 @@ export default async function isExisting (this: WebdriverIO.Element) {
         )
     }
 
-    const command = this.isReactElement ? this.parent.react$$.bind(this.parent) : this.parent.$$.bind(this.parent)
-    return command(this.selector as string).then((res) => res.length > 0)
+    const command = this.isReactElement
+        ? this.parent.react$.bind(this.parent)
+        : this.parent.$.bind(this.parent)
+    this.elementId = (await command(this.selector as string)).elementId
```
christian-bromann, line 67: Does this throw an error if `$` or `react$` can't find an element? What would be the return value and why does `command(this.selector).then(...)` doesn't return an element?

tmjpedro: is not throw an error if can't find an element.
I got your point when can't find an element, I'm adding `undefined` to the elementId.

tmjpedro: Updated also added 2 tests to make sure that we return false if the element can\'t be found.

christian-bromann:
@tmjpedro can we pause on this? It seems that using `$` is not a good idea either due to implicit waits. See also https://github.com/webdriverio/webdriverio/issues/7580#issuecomment-995882638. Let's evaluate better options.

tmjpedro:
> @tmjpedro can we pause on this? It seems that using `$` is not a good idea either due to implicit waits. See also [#7580 (comment)](https://github.com/webdriverio/webdriverio/issues/7580#issuecomment-995882638). Let's evaluate better options.

@christian-bromann sure, actually I feel I need to learn more about core elements to implement a robust solution. I will close the PR. thanks!

PR #7839, opened 2021-12-22
Sauce service before step RDC fix

denis-a1qa:
## Proposed changes
Add return in case using RDC to prevent issues like on screenshot:
image

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments


christian-bromann:
@wswebcreation didn't we just removed this `if` statement?

denis-a1qa:
@christian-bromann this issue was introduced here:
https://github.com/webdriverio/webdriverio/pull/7779

wswebcreation:
@christian-bromann 

No, this is a good catch, totally forgot about RDC. PR LGTM

christian-bromann:
@wswebcreation thanks for confirming.

@denis-a1qa can we move the `if` statement into `setAnnotation` so we ensure that we don't attempt this in other hooks? Or is this just not possible in RDC when setting the job title?

wswebcreation:
@christian-bromann 
Not sure what you mean, but the job tittle can also  be set for RDC, the annotation with context doesn’t work

christian-bromann:
Ahh right, so just setting the context via annotation doesn't work. I believe we have some more places where we need to add this if statement then.

wswebcreation:
I think that's already covered @christian-bromann, this PR is only for Cucumber. But maybe it's good to always call the methods and let the method determine if it's AND RDC AND a context call, then don't set it. This will remove the _complexity_ from all other methods

christian-bromann (approved):
@wswebcreation let's do this as part of a separate PR.

Thanks @denis-a1qa 👍

christian-bromann (approved):
@wswebcreation let's do this as part of a separate PR.

Thanks @denis-a1qa 👍

christian-bromann:
fyi: the build is broken due to typing issues that I am currently fixing.

Issue #7427, opened 2021-09-14
direct connect behavior not implemented correctly


jlipps:
The current directConnect behavior appears to not be implemented correctly. It's defined here: https://github.com/webdriverio/webdriverio/blob/main/packages/webdriver/src/index.ts#L44-L52

It merely looks for directConnect information in the params, but it should be looking in the capability response from the new session request. That's the whole point--for a load balancer to be able to respond with specific direct server connection information in its new session response, so that subsequent requests can be faster. The current implementation is basically just a reimplementation of the hostname and port parameters that can already be passed in.

The test for this feature just tests the current behavior, which I think is incorrect. For an example of a correct implementation, check out [web2driver](https://github.com/projectxyzio/web2driver/blob/master/src/session.js#L34-L63)'s wrapping of webdriver.

(I discovered this issue when I went to remove web2driver's handling of this case because I thought webdriver already handled it).

christian-bromann:
Thanks for reporting @jlipps , if these direct connect parameters are suppose to come from a session response then this is indeed wrong.

jlipps:
Yes, that's the case. Here's a reference article: https://appiumpro.com/editions/86-connecting-directly-to-appium-hosts-in-distributed-environments

jlipps:
The funny thing is I have some kind of memory of maybe implementing this myself for wdio? So maybe it got moved around and accidentally put somewhere else? Or maybe I am misremembering and I did not implement this feature in wdio.

PR #5853, opened 2020-09-23
Tracking doc: rewrite `webdriver` package into TypeScript

christian-bromann:
This is a tracking issue for organising the effort of re-writing the code base from vanilla JS into TypeScript. If you are reading this and are interested getting involved, this is a great chance to do so. Here are the rules:

- if you are interested converting source files from `/packages/webdriver` into TypeScript please comment on this thread and mention the files you want to convert to avoid duplicate work
- there is no reason to convert the whole package at once - smaller PRs make it easier to review the code
- if you make a PR, reference this issue so we can keep track of the progress
- if you make a PR, label it with `TypeScript` and the package name `webdriver` so we can easier find all PRs
- if you make a PR, assign it to the `TypeScript Rewrite` project

Happy rewriting 😅

PR #7894, opened 2022-01-10
Fix directConnect

KazuCocoa:
## Proposed changes

Fixes https://github.com/webdriverio/webdriverio/issues/7427

The behavior was changed in https://github.com/webdriverio/webdriverio/pull/5853. This PR tries to bring the feature back to be expected. cc @jlipps 

After some testing, it needed to update `client.options` rather than modify `params` to apply hostnames etc correctly.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

Let me try to run tests on my local... then will tweak documentation.

### DIFF
--- a/packages/wdio-types/src/Capabilities.ts
+++ b/packages/wdio-types/src/Capabilities.ts
@@ -451,6 +451,13 @@ export interface AppiumW3CCapabilities {
     'appium:printPageSourceOnFindFailure'?: boolean;
     'appium:nativeWebTap'?: boolean;
     'appium:options'?: AppiumCapabilities
+
+    // Users as directConnect feature by the server
+    // https://appiumpro.com/editions/86-connecting-directly-to-appium-hosts-in-distributed-environments
+    'appium:directConnectProtocol'?: string;
+    'appium:directConnectHost'?: string;
+    'appium:directConnectPort'?: number;
+    'appium:directConnectPath'?: string;
 }
 
 /**
--- a/packages/wdio-types/src/Capabilities.ts
+++ b/packages/wdio-types/src/Capabilities.ts
@@ -367,6 +367,13 @@ export interface AppiumCapabilities {
     eventTimings?: boolean;
     enablePerformanceLogging?: boolean;
     printPageSourceOnFindFailure?: boolean;
+
+    // Users as directConnect feature by the server
+    // https://appiumpro.com/editions/86-connecting-directly-to-appium-hosts-in-distributed-environments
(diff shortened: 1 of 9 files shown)


packages/wdio-types/src/Capabilities.ts
```diff
@@ -367,6 +367,13 @@ export interface AppiumCapabilities {
     eventTimings?: boolean;
     enablePerformanceLogging?: boolean;
     printPageSourceOnFindFailure?: boolean;
+
+    // Users as directConnect feature by the server
+    // https://appiumpro.com/editions/86-connecting-directly-to-appium-hosts-in-distributed-environments
+    directConnectProtocol?: string;
+    directConnectHost?: string;
+    directConnectPort?: number;
+    directConnectPath?: string;
```
christian-bromann, line 376: Can we also add them to `AppiumW3CCapabilities`?

christian-bromann: In setup direct connect we should then do:

```js
const directConnectProtocol = params.capabilities.directConnectProtocol || params.capabilities['appium:directConnectProtocol']
```

packages/wdio-types/src/Options.ts
```diff
@@ -183,10 +183,7 @@ export interface WebDriver extends Connection {
     /**
      * Appium direct connect options (see: https://appiumpro.com/editions/86-connecting-directly-to-appium-hosts-in-distributed-environments)
      */
-    directConnectProtocol?: string
-    directConnectHost?: string
-    directConnectPort?: number
-    directConnectPath?: string
+    enableDirectConnect?: boolean;
```
christian-bromann, line 186: Can we document this option [here](https://github.com/webdriverio/webdriverio/blob/main/website/docs/Options.md#webdriver-options)?

christian-bromann (commented):
Thanks for fixing this 🙏 

Some suggestions, otherwise LGTM

KazuCocoa:
@christian-bromann Could you review again when you have time?
I've tested that current implementation worked expectedly as https://appiumpro.com/editions/86-connecting-directly-to-appium-hosts-in-distributed-environments

Btw, does the e2e-mac work...?
The scenario worked when I modified the config and expected value as localhost or a host, but it is failing.
(As of current implementation, this directConnect should not affect the test case though..)

christian-bromann (approved):
Awesome 👍  LGTM

The e2e tests are very flaky atm, I haven't had the chance to look into them. We can just move forward with this.

Issue #7873, opened 2022-01-04
[🐛 Bug]: duration property in the afterSuite hook is undefined


jamesmortensen:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

7.16.9

### Node.js Version

v14.17.5

### Mode

WDIO Testrunner

### What happened?

We were attempting to get the duration of tests running in a describe suite, but in the afterSuite hook, we are seeing the duration as undefined:

```{
[0-0]   type: 'afterSuite',
[0-0]   title: 'Describe block A',
[0-0]   parent: '',
[0-0]   fullTitle: 'Describe block A',
[0-0]   pending: false,
[0-0]   file: '/Users/james/Dev/test.spec.js',
[0-0]   duration: undefined
[0-0] }
```

Inside wdio.conf.js:
```
afterSuite: function (suite) {
        // this fires when the describe suite is completed and it contains the file that was run.
        console.info(suite);
}
```

### What is your expected behavior?

It looks like someone intended to implement behavior where the duration of an entire suite is tracked and made available in the afterSuite hook.

### How to reproduce the bug.

To reproduce, write a single spec.js test file with a few simple tests, and then add the afterSuite hook and log the "suite" object to see its properties.

### Relevant log output

```typescript
{
[0-0]   type: 'afterSuite',
[0-0]   title: 'Describe block A',
[0-0]   parent: '',
[0-0]   fullTitle: 'Describe block A',
[0-0]   pending: false,
[0-0]   file: '/Users/james/Dev/test.spec.js',
[0-0]   duration: undefined
[0-0] }
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

christian-bromann:
@jamesmortensen Thanks for reporting!

The duration should definitely be set no matter if it is `0` or any other minimal duration.

Any contributions that resolves the bug are highly appreciated. Please take a look into our [contribution guidelines](https://webdriver.io/docs/contribute) and let us know if you have any questions. Cheers!

praveendvd:
@christian-bromann  I will take this up 

@jamesmortensen  which framework  are you using , is there a  minimum reproducible example

jamesmortensen:
[...]

PR #7989, opened 2022-02-08
#Fix for https://github.com/webdriverio/webdriverio/issues/7873#issuecomment-1032964198 duration missing in mocha hook

praveendvd:
## Proposed changes

Added duration property to mocha event if its missing : Fix for https://github.com/webdriverio/webdriverio/issues/7873

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/wdio-mocha-framework/src/index.ts
+++ b/packages/wdio-mocha-framework/src/index.ts
@@ -43,6 +43,7 @@ class MochaAdapter {
     private _suiteCnt: Map<string, number> = new Map()
     private _hookCnt: Map<string, number> = new Map()
     private _testCnt: Map<string, number> = new Map()
+    private _suiteStartDate!: Date
 
     constructor(
         private _cid: string,
@@ -202,8 +203,11 @@ class MochaAdapter {
 
         switch (hookName) {
         case 'beforeSuite':
+            this._suiteStartDate = new Date()
+            break
         case 'afterSuite':
             params.payload = this._runner?.suite.suites[0]
+            params.payload.duration = params.payload.duration || (new Date() as any) - (this._suiteStartDate as any)
             break
         case 'beforeTest':
         case 'afterTest':


packages/wdio-mocha-framework/src/index.ts
```diff
@@ -43,6 +43,7 @@ class MochaAdapter {
     private _suiteCnt: Map<string, number> = new Map()
     private _hookCnt: Map<string, number> = new Map()
     private _testCnt: Map<string, number> = new Map()
+    private _suiteStartDate!: Date
```
christian-bromann, line 46: Let's initiate it with a value so we don't need to worry about it not being defined.
```suggestion
    private _suiteStartDate: number = Date.now()
```

packages/wdio-mocha-framework/src/index.ts
```diff
@@ -202,8 +203,11 @@ class MochaAdapter {
 
         switch (hookName) {
         case 'beforeSuite':
+            this._suiteStartDate = new Date()
```
christian-bromann, line 206: Let's directly work with the timestamp here.
```suggestion
            this._suiteStartDate = Date.now()
```

packages/wdio-mocha-framework/src/index.ts
```diff
@@ -202,8 +203,11 @@ class MochaAdapter {
...
+            break
         case 'afterSuite':
             params.payload = this._runner?.suite.suites[0]
+            params.payload.duration = params.payload.duration || (new Date() as any) - (this._suiteStartDate as any)
```
christian-bromann, line 210:
```suggestion
            params.payload.duration = params.payload.duration || (Date.now() - this._suiteStartDate)
```

christian-bromann (commented):
After adding tests for this, it is good to go! Great job!

praveendvd:
@christian-bromann  Hi I fixed  the failing test by adding a check whether payload is an Object . This is required only for  this test: https://github.com/webdriverio/webdriverio/blob/0477291da874c7497df3e3975dd11084470636bd/packages/wdio-mocha-framework/tests/adapter.test.ts#L176

I believe the suites will be a array of object and not string . so should i add that check of whether payload is an object ? or just override suites with an object

    adapter['_runner']!.suite.suites[0] = { 'title': 'foo' }
    let result = adapter.prepareMessage('beforeSuite')


Or should we update the mock it self, and make the suites array an array of object instead of string

https://github.com/webdriverio/webdriverio/blob/0477291da874c7497df3e3975dd11084470636bd/packages/wdio-mocha-framework/tests/__mocks__/mocha.ts#L29

packages/wdio-mocha-framework/src/index.ts
```diff
@@ -202,8 +203,13 @@ class MochaAdapter {
 
         switch (hookName) {
         case 'beforeSuite':
+            this._suiteStartDate = Date.now()
+        // eslint-disable-next-line no-fallthrough
```
christian-bromann, line 207: Let's use break here rather than having to ignore eslint
```suggestion
        break;
```

packages/wdio-mocha-framework/src/index.ts
```diff
@@ -202,8 +203,13 @@ class MochaAdapter {
...
         case 'afterSuite':
             params.payload = this._runner?.suite.suites[0]
+            if (params.payload instanceof Object && typeof (params.payload) === 'object') params.payload.duration =
+                params.payload.duration ||
+                (Date.now() - this._suiteStartDate)
```
christian-bromann, line 212: No need to check for instanceof here
```suggestion
            if (typeof params.payload === 'object') {
                params.payload.duration = params.payload.duration ||
                (Date.now() - this._suiteStartDate)
            }
```

christian-bromann (commented):
@praveendvd we can we just have a single unit test for `prepareMessage`.

packages/wdio-mocha-framework/src/index.ts
praveendvd: > Let's use break here rather than having to ignore eslint

@christian-bromann  the suite object is the same for before and after suite , thats why there was no break there . If we add the break the beforeSuite will have only type property. All other information will be removed

packages/wdio-mocha-framework/src/index.ts
praveendvd: @christian-bromann I added that to ensure payload is not null as type of returns object for null

praveendvd: @christian-bromann i removed the check all together

packages/wdio-mocha-framework/tests/adapter.test.ts
```diff
@@ -181,7 +181,14 @@ test('prepareMessage', async () => {
...
     expect(result.file).toBe('/foo/bar.test.js')
-})
+
+    await new Promise((r) => setTimeout(r, 5000))
+    result = adapter.prepareMessage('afterSuite')
+    expect(result.type).toBe('afterSuite')
+    expect(result.title).toBe('first suite')
+    expect(result.duration).toBeDefined()
+    expect(result.duration >= 5000 && result.duration <= 5020).toBeTruthy()
+}, 6000)
```
praveendvd, line 191: @christian-bromann 

Hi the test is infered as Mocha.test but it is jest test so the 6000 will show error in the vscode . Do you want to add an explicit import for jest as 

`import {describe, expect, test} from '@jest/globals'.`

or add a type for test as

`let test:jest.It`

or just ignore the error

image

packages/wdio-mocha-framework/tests/adapter.test.ts
```diff
@@ -181,7 +181,14 @@ test('prepareMessage', async () => {
     expect(result.type).toBe('afterTest')
     expect(result.title).toBe('foobar')
     expect(result.file).toBe('/foo/bar.test.js')
-})
+
+    await new Promise((r) => setTimeout(r, 5000))
```
christian-bromann, line 185: That is not necessary, just mock the `_suiteStartDate` property to be `Date.now() - 3000`

packages/wdio-mocha-framework/src/index.ts
praveendvd: @christian-bromann 

https://github.com/webdriverio/webdriverio/blob/08644c01ba6537cce209d57b4c534cba3f96c288/packages/wdio-mocha-framework/src/index.ts#L208-L209

The params.payload step is commond for both before and aftersuite , thats why we didnt add a break before

praveendvd: so if i add break , the common statement of params.payload wont get executed . SHould i repeat that inside first case ?

christian-bromann: yeah .. my goal here is to get rid of this eslint comment .. at the end of the day we use it to follow its rules 😉

packages/wdio-mocha-framework/tests/adapter.test.ts
```diff
@@ -181,6 +181,13 @@ test('prepareMessage', async () => {
     expect(result.type).toBe('afterTest')
     expect(result.title).toBe('foobar')
     expect(result.file).toBe('/foo/bar.test.js')
+    // @ts-ignore
+    adapter._suiteStartDate = Date.now() - 5000
```
christian-bromann, line 185: to access private fields, do
```suggestion
    adapter['_suiteStartDate'] = Date.now() - 5000
```

christian-bromann (approved):
👍  thanks for confirming, let's move forward with this
LGTM

packages/wdio-mocha-framework/src/index.ts
```diff
@@ -202,8 +203,12 @@ class MochaAdapter {
 
         switch (hookName) {
         case 'beforeSuite':
+            this._suiteStartDate = Date.now()
+            params.payload = this._runner?.suite.suites[0]
+            break
         case 'afterSuite':
             params.payload = this._runner?.suite.suites[0]
+            params.payload.duration = params.payload.duration || (Date.now() - this._suiteStartDate)
```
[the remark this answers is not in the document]
christian-bromann: @Joosakur can you raise a separate issue?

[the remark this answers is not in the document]
Joosakur: Sure!

PR #8002, opened 2022-02-11
Add new eslint rules and directory structure - await-expect, no-debug…

WillBrock:
## Proposed changes

Adding 3 new eslint rules.
- await-expect - Enforce await in front of expect statements
- no-debug - Enforce no browser.debug() statements
- no-pause - Enforce no browser.pause() statements

## Types of changes

- [x] Documentation update

## Checklist

## Further comments

This is still a draft and need to add the appropriate documentation.


packages/eslint-plugin-wdio/lib/rules/await-expect.js
```diff
@@ -0,0 +1,37 @@
...
+            },
+            create : function(context) {
+                return {
+                    CallExpression(node) {
+                        const name = node.callee.name
+
+                        if (name === 'expect') {
+                            const parentType = node.parent && node.parent.parent && node.parent.parent.parent ? node.parent.parent.parent.type : false
+
+                            if (parentType && parentType !== 'AwaitExpression') {
```
christian-bromann, line 25: I wonder if it makes sense to also check the expression that is being expected, if it isn't a WebdriverIO object there might not be a need for `await`, e.g.:
```js
const title = browser.getTitle();
expect(title).toBe('foobar');
```
Given it might be very difficult to determine that I suggest to only explicitly apply this rule for all assertions we've defined in expect-webdriverio, what do you think?

christian-bromann (commented):
Thanks for improving our eslint strength 💪 

Can we maybe use TypeScript for all this? Or is there a specific reason not to?

packages/eslint-plugin-wdio/lib/rules/await-expect.js
christian-bromann: That is impossible to say given that a linter doesn't run your code. If we only do this for expect-webdriverio assertions we can be assured this is the case though.

WillBrock: I rarely use typescript and haven't set up a project with it before in a long time but I'm not sure I really see the benefit to using typescript with eslint. I could use optional chaining but that would be the only benefit I think of vs having the overhead of dealing with builds and typescript.

I can see if it's possible to only enforce the await if it's a Webdriver object.

WillBrock: Yeah, determining if it is a Webdriver object would't be possible.

One issue I see with not awaiting things like .toBe that I've ran across.

```
const title = browser.getTitle();

// Both of the below work
await expect(title).toBe('foobar');
expect(await title).toBe('foobar');
```

Ideally people should be doing the await on the title either before or while inside expect but if they forget the test can pass when it should have failed.

christian-bromann:
> I rarely use typescript and haven't set up a project with it before in a long time but I'm not sure I really see the benefit to using typescript with eslint. I could use optional chaining but that would be the only benefit I think of vs having the overhead of dealing with builds and typescript.

For one you could replace things like:
```js
const parentType = node.parent && node.parent.parent && node.parent.parent.parent ? node.parent.parent.parent.type : false
```

into:

```ts
const parentType = node.parent?.parent?.parent?.type : false
```

Another reason to use TypeScript here is that it simplifies further development. Dealing with AST is difficult if you don't know the types required for the node you are inspecting. I build our codemod without TS because I didn't knew better and would never do this again because it makes much more sense to use TS for its type safety.

Lastly I don't think adding TS for this linter package adds more burden on the build. There are only 3 files to compile. I believe we should stick with TS across the whole code base.

WillBrock:
@christian-bromann I switched this over to TypeScript and only enforced awaiting on expect for expect-webdriverio matchers.

I was getting some vscode errors with some of the typescript things so I left those to `any`. I might need some help with some of that. I feel like I'm missing something simple.

packages/eslint-plugin-wdio/lib/utils/helpers.ts
```diff
@@ -0,0 +1,14 @@
...
+
+    if (
+        callee &&
+        callee?.object?.name === 'browser' &&
+        callee?.property?.name === command
+    ) {
+        return true
+    }
+
+    return false
```
christian-bromann, line 12:
```suggestion
    return (
        callee &&
        callee?.object?.name === 'browser' &&
        callee?.property?.name === command
    )
```

christian-bromann (commented):
One little suggestion. I will check out the code tomorrow and see if we can replace some of the `any`s with the proper types.

packages/eslint-plugin-wdio/lib/utils/helpers.ts
WillBrock: Ah good call, I committed an update.

Thanks for looking at this.

WillBrock:
I fixed a few of the types. There are still 2 things I set to `any` which deal with the AST nodes.

packages/eslint-plugin-wdio/lib/index.ts
```diff
@@ -0,0 +1,28 @@
...
+    recommended: {
+        globals: {
+            $                  : false,
+            $$                 : false,
+            browser            : false,
+            driver             : false,
+            expect             : false,
+            multiremotebrowser : false,
+        },
+        rules : {}
```
christian-bromann, line 21: Shouldn't the rules imported above be recommended?

christian-bromann (approved):
I will go ahead and merge this.

Awesome work @WillBrock 🎉

packages/eslint-plugin-wdio/lib/index.ts
WillBrock: Yeah probably, I'll create a branch that does that along with adding some docs for it.

christian-bromann: I added this in https://github.com/webdriverio/webdriverio/pull/8095

PR #8151, opened 2022-03-23
chore(wdio-types): remove implicit dependency on ts-node

boneskull:
Before this, `@wdio/types` needs `ts-node` installed to be used (which is not present in its `dependencies`, `devDependencies` or otherwise). This is pretty annoying if you don't use `ts-node` otherwise.

I have copied the relevant types from `ts-node` and put them into `src/Compiler.ts`.

I have added `typescript` as a `peerDependency`, because `@wdio/types` now depends on types in the `typescript` package itself. Given that `typescript` is _probably installed anyway_ by consumers, it did not seem appropriate to add it to `dependencies`.  I can understand if we just want to drop this from `peerDependencies` altogether, but it _does_ mean that there's still an implicit dependency, which doesn't really solve the problem.

## Proposed changes

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/wdio-types/package.json
+++ b/packages/wdio-types/package.json
@@ -31,5 +31,8 @@
   "dependencies": {
     "@types/node": "^17.0.4",
     "got": "^11.8.1"
+  },
+  "peerDependencies": {
+    "typescript": "^4.6.2"
   }
 }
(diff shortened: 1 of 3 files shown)


boneskull:
cc @christian-bromann

boneskull:
Another way to approach this would be to move the types which needed `ts-node` originally into a separate package or otherwise split these up.  As a consumer of `@wdio/types`, I don't actually need any of the types that rely on it.

christian-bromann (approved):
Thanks 👍 

Even though we might run into issues when these types change upstream I think this is the best solution for now to move forward.

Issue #8048, opened 2022-02-23
[💡 Feature]: Allow passing more capabilites via REPL


gaunap:
### Is your feature request related to a problem?

Appium breaks up every 60s if no new command is send when using repl for iOS:
`wdio repl ios  --path /wd/hub --platformVersion 15.2 --deviceName 'iPhone 13'`

This makes testing very cumbersome.

The newCommandTimeout of the capabilities cannot be overwritten.

### Describe the solution you'd like.

I like to pass more capbilities to Appium. E.g. using this kind of signature: --"appium:newCommandTimeout" 9999999

### Additional context

My setup is MacOS 12.2.1, wdio 7.16.13, Appium-Server-GUI 1.22 and iphone Simulator 13

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

christian-bromann:
Thanks for reporting!

Any contributions that implements this enhancement are highly appreciated. Please take a look into our [contribution guidelines](https://webdriver.io/docs/contribute) and let us know if you have any questions. Cheers!

praveendvd:
@christian-bromann when i run `wdio repl chrome` it just creates a new wdio project , it doesnt do anything

christian-bromann:
@praveendvd the repl starts for me:

```
❯ mkdir repltest
❯ cd repltest/
❯ npm i @wdio/cli

added 303 packages, and audited 304 packages in 7s

38 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities
❯ npx wdio repl chrome
2022-03-23T12:36:45.496Z INFO devtools:puppeteer: Initiate new session using the DevTools protocol
[...]

PR #8156, opened 2022-03-26
Added ability to pass capability to repl through config file : Fix for https://github.com/webdriverio/webdriverio/issues/8048

praveendvd:
## Proposed changes

Fix for https://github.com/webdriverio/webdriverio/issues/8048

Now user can pass all config using the config file , instead of passing each individual capability

introduce two new args : configPath and capability

configPath : path to config (relative / full)
capability : index or key

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

<a href="https://gitpod.io/#https://github.com/webdriverio/webdriverio/pull/8156"></a>


praveendvd:
@christian-bromann  

The repl session --help shows only few supported args. But webdriver remote support more args and throws the error with supported flags like:

image


but if we do wdio repl --help

image

ony these args are listed. I am not sure where is the remaining args are coming from. WIll try to fix it along with this PR

praveendvd:
@christian-bromann  :

packages/wdio-cli/src/commands/run.ts  its comming fromthis one , will work one it thank you 🤩🤩

praveendvd:
https://github.com/webdriverio/webdriverio/blob/11be174dbad723750c8ab3fc9bc197f2f3534d1c/packages/wdio-cli/src/index.ts#L54

```
    const params = {...argv.argv};
    console.log({...argv.argv},params)
```

@christian-bromann  could you help me with this here `{...argv.argv}` prints the args but params prints it as empty object : 

image

I am not sure why.

Here , I am trying to catch the error and explicitly throw supported yargs error for repl instead of wdio command

praveendvd:
Now the help shows :

image

but if any error is thrown like session not created , port not open etc:  the error will be:

image

it still has the excluded options. I am not able to figure out from where this yargs is comming

packages/wdio-cli/src/commands/repl.ts
```diff
@@ -14,28 +16,42 @@ const IGNORED_ARGS = [
...
+      }
+  } as const
 
 export const builder = (yargs: yargs.Argv) => {
     return yargs
-        .options(pickBy(cmdArgs, (_, key) => !IGNORED_ARGS.includes(key)))
+        .options(pickBy({ ...cmdArgs, ...runCmdArgs }, (_, key) => !IGNORED_ARGS.includes(key)))
         .example('$0 repl firefox --path /', 'Run repl locally')
+        .example('$0 repl firefox --configPath wdio.config.js --capability 0', 'Run repl using first capability from config file')
+        .example('$0 repl firefox --configPath wdio.config.js --capability browserA', 'Run repl using browserA capability from config file')
```
christian-bromann, line 54:
```suggestion
        .example('$0 repl firefox --configPath wdio.config.js --capability browserA', 'Run repl using browserA capability from config file when using Multiremote')
```

packages/wdio-cli/src/commands/repl.ts
```diff
@@ -45,7 +61,10 @@ export const builder = (yargs: yargs.Argv) => {
 }
 
 export const handler = async (argv: ReplCommandArguments) => {
-    const caps = getCapabilities(argv)
+    const config = argv.configPath && new ConfigParser()
+    config && config.addConfigFile(argv.configPath)
+    const requiredCaps:any = config && (config.getCapabilities() as any)[argv.capability]
```
christian-bromann, line 66: Can we type this properly?

website/docs/repl.md
```diff
@@ -5,17 +5,23 @@ title: REPL interface
...
-wdio repl chrome
+npx wdio repl chrome
 ```
 
 This would open a Chrome browser that you can control with the REPL interface. Make sure you have a browser driver running on port `4444` in order to initiate the session. If you have a [Sauce Labs](https://saucelabs.com) (or other cloud vendor) account, you can also directly run the browser on your command line in the cloud via:
 
 ```sh
 wdio repl chrome -u $SAUCE_USERNAME -k $SAUCE_ACCESS_KEY
 ```
+Capabilities can be passed to the repl through the supported commandline arguments as mentioned in the below sections. More capabilites than the supported commandline arguments can be passed using the wdio config file. Wdio support both capabilities list and mutiremote capability, you can specify which capability to use from list or multiremote using --capability flag eg:
```
christian-bromann, line 19:
```suggestion
Capabilities can be passed to the repl through the supported commandline arguments as mentioned in the below sections. More capabilites than the supported commandline arguments can be passed using the wdio config file. Wdio support both capabilities list and mutiremote capabilities, you can specify which capability to use from list or multiremote using `--capability` flag e.g.:
```

christian-bromann:
We can ignore the failing mac tests for now, happy to approve and merge once we can fix the [typing issue](https://github.com/webdriverio/webdriverio/pull/8156#discussion_r843731185).

praveendvd:
> We can ignore the failing mac tests for now, happy to approve and merge once we can fix the [typing issue](https://github.com/webdriverio/webdriverio/pull/8156#discussion_r843731185).

I haven't written the unit test for this , wouldn't that be required for this ?

christian-bromann:
> I haven't written the unit test for this , wouldn't that be required for this ?

A unit test would be great but not required for me. Just the [typing issue](https://github.com/webdriverio/webdriverio/pull/8156#discussion_r843731185) should get fixed.

praveendvd:
@christian-bromann  I have two more issues to solve , one is that 

**FIrst issue:**

https://github.com/webdriverio/webdriverio/pull/8156/files#diff-8f5e23e7352b2714cf08962512e3ae8da39ec1917e1a2ffacaee8133cb2fbe32R18

This line has <options> mandatory  so even if the browser name is there in the config file we have to provide a name here. if i remove the <option> from this string , it makes it non mandatory but then the devtool doesnt spin up automatically  even if i give the command `npx wdio repl chrome`

Could you help me in knowing which file does this export.command is used in and how to make it nn mandatory 

**Second issue**


It seems that repl expect pure w3c capabilities and not desired capabilities. So for example if the capability have something like maxINstance , it will throw error saying its not 23c capability .

do we have any inbuilt method to convert desiredcaps to w3c caps ? or is there any logic that i can use wer we can filter unwanted properties out ?

christian-bromann:
> FIrst issue:

I think the expected behavior (that also would be backwards compatible) is that we start a simple Chrome session like we do right now if someone runs `npx wdio repl chrome` and if the user runs `npx wdio repl -p /wdio.conf.ts -c 0` we check for the config file and run with the first capability. Therefor I think `<options>` should be optional anyway and we only do the new behavior if these params are given.

> Second issue

Which repl command did you run here?

praveendvd:
> > FIrst issue:
> 
> I think the expected behavior (that also would be backwards compatible) is that we start a simple Chrome session like we do right now if someone runs `npx wdio repl chrome` and if the user runs `npx wdio repl -p /wdio.conf.ts -c 0` we check for the config file and run with the first capability. Therefor I think `<options>` should be optional anyway and we only do the new behavior if these params are given.
> 
> > Second issue
> 
> Which repl command did you run here?

@christian-bromann  the first issue is that options are not optional , it's a mandatory argument . I am not sure where it's made mandatory .

praveendvd:
@christian-bromann  it's not from the yargs in repl.ts , it's from another yargs instance from some other starting point file. Will investigate it

christian-bromann:
> Will investigate it

Let me know if I can help!

praveendvd:
@christian-bromann  this is ready to be merged , if tests are not required

praveendvd:
@christian-bromann  could you help me out here , i am a gettign a compilation error from index file which i have not modified. I am not sure where the error coming from , its not throwing the error locally

christian-bromann:
> could you help me out here

I am currently working on this issue in #8303

christian-bromann:
@praveendvd mind rebasing on top of current `main` branch? Just pushed a [committ](https://github.com/webdriverio/webdriverio/pull/8303) that fixes some issues we had with TypeScript and testing.

praveendvd:
@christian-bromann  did it , could u have a look

packages/wdio-cli/src/utils.ts
```diff
@@ -266,6 +269,11 @@ export const validateServiceAnswers = (answers: string[]): Boolean | string => {
 }
 
 export function getCapabilities(arg: ReplCommandArguments) {
+    let finalCapabilities
```
christian-bromann, line 272: If we return early we don't need this.
```suggestion
```

packages/wdio-cli/src/utils.ts
```diff
@@ -287,8 +295,18 @@ export function getCapabilities(arg: ReplCommandArguments) {
...
+    } else if (/(js|ts)$/.test(arg.option)) {
+        const config = new ConfigParser()
+        config.addConfigFile(arg.option)
+        let requiredCaps = config.getCapabilities()
+        if (Object.prototype.hasOwnProperty.call(arg, 'capabilities')) {
+            requiredCaps = (requiredCaps as (DesiredCapabilities | W3CCapabilities)[])[Number(arg.capabilities)] ||
+                (requiredCaps as MultiRemoteCapabilities)[arg.capabilities]
+        }
+        const requiredW3CCaps = pickBy(requiredCaps, (_, key) => !IGNORED_CAPABILITIES.includes(key))
+        finalCapabilities = { capabilities: { ...(requiredW3CCaps as W3CCapabilities) } }
```
christian-bromann, line 307:
```suggestion
        return { capabilities: { ...(requiredW3CCaps as W3CCapabilities) } }
```

packages/wdio-cli/src/utils.ts
```diff
@@ -287,8 +295,18 @@ export function getCapabilities(arg: ReplCommandArguments) {
...
         return { capabilities: { browserName: 'Safari', ...IOS_CONFIG, ...optionalCapabilites } }
+    } else if (/(js|ts)$/.test(arg.option)) {
+        const config = new ConfigParser()
+        config.addConfigFile(arg.option)
+        let requiredCaps = config.getCapabilities()
+        if (Object.prototype.hasOwnProperty.call(arg, 'capabilities')) {
+            requiredCaps = (requiredCaps as (DesiredCapabilities | W3CCapabilities)[])[Number(arg.capabilities)] ||
+                (requiredCaps as MultiRemoteCapabilities)[arg.capabilities]
+        }
+        const requiredW3CCaps = pickBy(requiredCaps, (_, key) => !IGNORED_CAPABILITIES.includes(key))
```
christian-bromann, line 306: For which options are you looking here?

packages/wdio-cli/src/utils.ts
```diff
@@ -287,8 +295,18 @@ export function getCapabilities(arg: ReplCommandArguments) {
...
+        finalCapabilities = { capabilities: { ...(requiredW3CCaps as W3CCapabilities) } }
     }
-    return { capabilities: { browserName: arg.option } }
+    return finalCapabilities || { capabilities: { browserName: arg.option } }
```
christian-bromann, line 309:
```suggestion
    return { capabilities: { browserName: arg.option } }
```

website/docs/repl.md
```diff
@@ -17,6 +17,22 @@ This would open a Chrome browser that you can control with the REPL interface. M
...
+```
+
+Repl could also be ran using the capabilities from the webdriverIO config file. Wdio supports capabilities object; or ; multiremote capability list or object. 
+
+If the config file uses capabilities object then just pass the path to config file, else if its a multiremote capability then, specify which capability to use from list or multiremote using the positional argument . Note: for list we consider zero based index. 
+
+```sh
+wdio repl "./path/to/wdio.config.js" 
+wdio repl "./path/to/wdio.config.js" 0 -p 9515
+wdio repl "./path/to/wdio.config.js" "capabilityPropertyName" -p 9515
```
christian-bromann, line 33: Maybe we add some examples, e.g. "if your config capabilities look like this you can run a repl for this capability by executing this".

christian-bromann (commented):
Added some comments, let me know if you have any questions.

packages/wdio-cli/src/commands/repl.ts
```diff
@@ -34,12 +34,15 @@ export const cmdArgs = {
...
 export const builder = (yargs: yargs.Argv) => {
     return yargs
-        .options(pickBy(cmdArgs, (_, key) => !IGNORED_ARGS.includes(key)))
+        .options(pickBy({ ...cmdArgs, ...runCmdArgs }, (_, key) => !IGNORED_ARGS.includes(key)))
         .example('$0 repl firefox --path /', 'Run repl locally')
         .example('$0 repl chrome -u <SAUCE_USERNAME> -k <SAUCE_ACCESS_KEY>', 'Run repl in Sauce Labs cloud')
         .example('$0 repl android', 'Run repl browser on launched Android device')
         .example('$0 repl "./path/to/your_app.app"', 'Run repl native app on iOS simulator')
         .example('$0 repl ios -v 11.3 -d "iPhone 7" -u 123432abc', 'Run repl browser on iOS device with capabilities')
+        .example('$0 repl "./path/to/wdio.config.js"', 'Run repl using capabilities from wdio.config.js')
```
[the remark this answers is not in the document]
praveendvd: @christian-bromann  if we provide capabilities directly (neither multiremote or multi capabilities] eg capability:{browsername:chrome}

packages/wdio-cli/src/commands/repl.ts
```diff
@@ -34,12 +34,15 @@ export const cmdArgs = {
...
+        .example('$0 repl "./path/to/wdio.config.js" 0 -p 9515', 'Run repl using capabilities from multiremotecapabilty in wdio.config.js')
```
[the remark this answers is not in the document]
praveendvd: yes thats cprrect , we have 3 ways to pass capability , 

https://webdriver.io/docs/options/#capabilities
https://webdriver.io/docs/options/#capabilities-1
https://webdriver.io/docs/multiremote

packages/wdio-cli/src/utils.ts
praveendvd: @christian-bromann  maxInstance , specs , etc . These can be passed as desired capability but are not valid w3c

packages/wdio-cli/src/commands/repl.ts
```diff
@@ -14,7 +14,7 @@ const IGNORED_ARGS = [
 
 export const command = 'repl <option> [capabilities]'
 export const desc = 'Run WebDriver session in command line'
-export const cmdArgs = {
+export const cmdArgs: { [k in keyof ReplCommandArguments]?:object } = {
```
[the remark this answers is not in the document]
praveendvd: @christian-bromann  we are passing it for the yargs so its a object and in replCommand its a string . so its not compatible

packages/wdio-cli/src/commands/repl.ts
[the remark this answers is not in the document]
praveendvd: changed to yargs.Options

packages/wdio-cli/src/commands/repl.ts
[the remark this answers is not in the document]
praveendvd: yargs.Options is used instead , thank you 🎉

packages/wdio-cli/src/commands/repl.ts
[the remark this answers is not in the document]
christian-bromann: A `wdio.conf.js` has no single capability object, this is only allowed in standalone but not using the testrunner.

packages/wdio-cli/src/utils.ts
christian-bromann: Can we use this object: https://github.com/webdriverio/webdriverio/blob/main/packages/webdriver/src/constants.ts#L143-L147 ?

praveendvd: @christian-bromann  We have vendor specific capabilities like "goog:chromeoptions" etc

praveendvd: So we are removing unwanted properties and allowing all other capabilities

christian-bromann: Let's update the `pickBy` filter to allow only the valid WebDriver capabilities + the ones that have a colon (":"), we can remove the rest

christian-bromann: I am suggestion to reuse the `webdriver` constant because creating a new list of capabilities here makes it more difficult to maintain this code because, e.g. if we add a new option we need to update that list and all other lists too, so having a minimal amount of lists would be desireable.

packages/wdio-cli/src/utils.ts
```diff
@@ -287,6 +294,17 @@ export function getCapabilities(arg: ReplCommandArguments) {
         return { capabilities: { browserName: 'Chrome', ...ANDROID_CONFIG, ...optionalCapabilites } }
     } else if (/ios/.test(arg.option)) {
         return { capabilities: { browserName: 'Safari', ...IOS_CONFIG, ...optionalCapabilites } }
+    } else if (/(js|ts)$/.test(arg.option)) {
+        const config = new ConfigParser()
+        config.autoCompile()
+        config.addConfigFile(arg.option)
+        let requiredCaps = config.getCapabilities()
+        if (Object.prototype.hasOwnProperty.call(arg, 'capabilities')) {
```
christian-bromann, line 302: better:
```suggestion
        if (typeof arg.capabilities !== 'undefined') {
```
because it is more readable imo

packages/wdio-cli/src/utils.ts
```diff
@@ -287,6 +294,17 @@ export function getCapabilities(arg: ReplCommandArguments) {
...
+            requiredCaps = (requiredCaps as (DesiredCapabilities | W3CCapabilities)[])[Number(arg.capabilities)] ||
+                (requiredCaps as MultiRemoteCapabilities)[arg.capabilities]
+        }
+        const requiredW3CCaps = pickBy(requiredCaps, (_, key) => !IGNORED_CAPABILITIES.includes(key))
+        return { capabilities: { ...(requiredW3CCaps as W3CCapabilities) } }
```
christian-bromann, line 307: before we return, let's make sure this object exists and if not (e.g. a user picks a capability that doesn't exist) throw an error

packages/wdio-cli/src/utils.ts
praveendvd: @christian-bromann 
image

packages/wdio-cli/src/utils.ts
praveendvd: @christian-bromann  the webdriver session throws the error saying capability is invalid, should we explicitly throw ?

christian-bromann: No, what I meant was, let's say the capabilities are these:

```
// wdio.conf.js
exports.config = {
  // ...
  capabilities: [{ ... }] 
  // ...
}
```

and someone runs:

```sh
npx wdio repl "./wdio.conf.js" 123
```

Given that there is no capability that can be index with 123, we should throw an error.

packages/wdio-cli/src/utils.ts
```diff
@@ -11,6 +11,10 @@ import type { Options, Capabilities, Services } from '@wdio/types'
 
 import { ReplCommandArguments, Questionnair, SupportedPackage, OnCompleteResult, ParsedAnswers } from './types'
 import { EXCLUSIVE_SERVICES, ANDROID_CONFIG, IOS_CONFIG, QUESTIONNAIRE } from './constants'
+import { ConfigParser } from '@wdio/config'
+import { DesiredCapabilities, MultiRemoteCapabilities, W3CCapabilities } from '@wdio/types/build/Capabilities'
+import pickBy from 'lodash.pickby'
+import _ from 'lodash'
```
christian-bromann, line 17: Let's not import all of `lodash` , it is a massive library and many of it isn't used here. Better add a specific method as dependency and import it. In this case I would recommend to just use basic JS to check if `requiredW3CCaps` exists.

packages/wdio-cli/src/utils.ts
```diff
@@ -11,6 +11,10 @@ import type { Options, Capabilities, Services } from '@wdio/types'
 
 import { ReplCommandArguments, Questionnair, SupportedPackage, OnCompleteResult, ParsedAnswers } from './types'
 import { EXCLUSIVE_SERVICES, ANDROID_CONFIG, IOS_CONFIG, QUESTIONNAIRE } from './constants'
+import { ConfigParser } from '@wdio/config'
+import { DesiredCapabilities, MultiRemoteCapabilities, W3CCapabilities } from '@wdio/types/build/Capabilities'
```
christian-bromann, line 15: It is not recommended to import from the `build` folder. What happens when the file changes? Better do use the imports provided by the package, e.g.:

```ts
import { Capabilities } from '@wdio/types'
// and us it via e.g. `Capabilities.MultiRemoteCapabilities`
```

packages/wdio-cli/src/utils.ts
christian-bromann: I think it makes sense to move `VALID_CAPS` into `wdio-protocols` export it there and import it from the CLI and make `@wdio/protocols` a peer dependency.

Note: you can't do relative imports across package boundaries because when these packages are installed through NPM the folder structure might differ.

praveendvd: @christian-bromann  i tried adding it to to the index file of protol package and adding it in depenedency but was showing no export member . COuld you show one example how to do this or which file to add this

praveendvd:
@christian-bromann  did all the changes 😍 i just had to compile it to get the error go

packages/wdio-cli/src/utils.ts
```diff
@@ -287,6 +290,19 @@ export function getCapabilities(arg: ReplCommandArguments) {
...
+        config.addConfigFile(arg.option)
+        let requiredCaps = config.getCapabilities()
+        if (typeof arg.capabilities !== 'undefined') throw ('Error!!!: Please provide index/namedProperty of capability to use from the capabilities array/object in wdio config file')
+        requiredCaps = (requiredCaps as (Capabilities.DesiredCapabilities | Capabilities.W3CCapabilities)[])[Number(arg.capabilities)] ||
+                (requiredCaps as Capabilities.MultiRemoteCapabilities)[arg.capabilities]
+        const requiredW3CCaps = pickBy(requiredCaps, (_, key) => !WEBDRIVERIO_SPECIFIC_CAPABILITIES.includes(key))
+        if (!Object.keys(requiredW3CCaps).length)
+            throw (`Error!!! : No capability found in config file with the provided capability index/namedProperty: ${arg.capabilities}. Please check the capability in your wdio config file.`)
+        return { capabilities: { ...(requiredW3CCaps as Capabilities.W3CCapabilities) }
+        }
```
christian-bromann, line 305:
```suggestion
        return { capabilities: { ...(requiredW3CCaps as Capabilities.W3CCapabilities) } }
```

packages/wdio-cli/src/utils.ts
```diff
@@ -287,6 +290,19 @@ export function getCapabilities(arg: ReplCommandArguments) {
...
+        const config = new ConfigParser()
+        config.autoCompile()
+        config.addConfigFile(arg.option)
+        let requiredCaps = config.getCapabilities()
+        if (typeof arg.capabilities !== 'undefined') throw ('Error!!!: Please provide index/namedProperty of capability to use from the capabilities array/object in wdio config file')
+        requiredCaps = (requiredCaps as (Capabilities.DesiredCapabilities | Capabilities.W3CCapabilities)[])[Number(arg.capabilities)] ||
+                (requiredCaps as Capabilities.MultiRemoteCapabilities)[arg.capabilities]
+        const requiredW3CCaps = pickBy(requiredCaps, (_, key) => !WEBDRIVERIO_SPECIFIC_CAPABILITIES.includes(key))
+        if (!Object.keys(requiredW3CCaps).length)
+            throw (`Error!!! : No capability found in config file with the provided capability index/namedProperty: ${arg.capabilities}. Please check the capability in your wdio config file.`)
```
christian-bromann, line 303:
```suggestion
        if (!Object.keys(requiredW3CCaps).length) {
            throw (`No capability found in given config file with the provided capability indexed/named property: 
       }
       ${arg.capabilities}. Please check the capability in your wdio config file.`)
```

packages/wdio-cli/src/utils.ts
```diff
@@ -287,6 +290,19 @@ export function getCapabilities(arg: ReplCommandArguments) {
         return { capabilities: { browserName: 'Chrome', ...ANDROID_CONFIG, ...optionalCapabilites } }
     } else if (/ios/.test(arg.option)) {
         return { capabilities: { browserName: 'Safari', ...IOS_CONFIG, ...optionalCapabilites } }
+    } else if (/(js|ts)$/.test(arg.option)) {
+        const config = new ConfigParser()
+        config.autoCompile()
+        config.addConfigFile(arg.option)
+        let requiredCaps = config.getCapabilities()
+        if (typeof arg.capabilities !== 'undefined') throw ('Error!!!: Please provide index/namedProperty of capability to use from the capabilities array/object in wdio config file')
```
christian-bromann, line 298:
```suggestion
        if (typeof arg.capabilities !== 'undefined') {
            throw ('Please provide index/named property of capability to use from the capabilities array/object in wdio config file')
        }
```

christian-bromann (commented):
@praveendvd thanks for the update.

I probably wasn't clear enough with my suggestion. Rather then defining webdriverio specific caps I suggested to take [this list](https://github.com/webdriverio/webdriverio/pull/8156#discussion_r870173747) and move it from the `webdriver` package into `@wdio/protocol` and filter only these including ones that contain a colon (`:`). Given that WebDriver capabilities don't change often this makes it less likely that someone forgets about maintaining the list

praveendvd:
@christian-bromann are u suggesting to pic capability form the VALID_CAPS , instead of filtering out invalid caps ?

But the valid_caps is not exhaustive , the vendor specific capabilities like chromeoptions, Firefox option etc are not there in the list .

The main intension of providing this was to allow user to give all supported w3c/json capabilities  to repl

christian-bromann:
Given the WebDriver protocol has been around for a while we can start being more strict about capabilities being W3C compatible. Every browser, vendor and service supports W3C caps these days. So yes, imho we should pick valid caps rather than filtering out invalid ones because the invalid cap list is likely to increase over time and it makes it hard to maintain. Unless you see a reason why we should stay JSONWP compatible here.

praveendvd:
@christian-bromann  done , could you have a look

christian-bromann (commented):
The code changes look good. The last thing to do would be some unit tests [here](https://github.com/webdriverio/webdriverio/blob/main/packages/wdio-cli/tests/utils.test.ts#L387-L401)

praveendvd:
> The code changes look good. The last thing to do would be some unit tests [here](https://github.com/webdriverio/webdriverio/blob/main/packages/wdio-cli/tests/utils.test.ts#L387-L401)

@christian-bromann the getCapability is not entering the actual configparser , i cannot see any where its being mocked also . But the config.getCapabilities returns capabilities no matter wat. Could you help me with this

praveendvd:
@christian-bromann  Also adbport, 

https://appium.io/docs/en/writing-running-appium/caps/

have amny caps which we dont have in the VALID_CAPS array , how are these handled ?

christian-bromann:
> i cannot see any where its being mocked

It is mocked [here](https://github.com/webdriverio/webdriverio/blob/main/packages/wdio-config/tests/__mocks__/%40wdio/config.ts#L22-L27). Every package in this mono repo has its mock in their own package.

> have amny caps which we dont have in the VALID_CAPS array , how are these handled ?

All of them should be written with a vendor prefix, for the adb port e.g. it is `appium:adbPort`.

praveendvd:
@christian-bromann  https://github.com/webdriverio/webdriverio/blob/216622868b810344b12bc8b3ed6c32586d08c114/packages/wdio-types/src/Capabilities.ts#L486

This is supported without vendor prefix  , should we update this file ?

christian-bromann:
> should we update this file ?

Yes, we should provide this type with an additional vendor prefixed version.

praveendvd:
@christian-bromann  could you let me know how to reset taht mock , i am not able to get around it

christian-bromann:
To reset a mock you can call [`mockReset`](https://jestjs.io/docs/mock-function-api#mockfnmockreset)

packages/wdio-cli/tests/utils.test.ts
```diff
@@ -398,6 +398,15 @@ describe('getCapabilities', () => {
     it('should return driver with capabilities for desktop', () => {
         expect(getCapabilities({ option: 'chrome' } as any)).toMatchSnapshot()
     })
+
+    it.only('should return driver with capabilities for wdio.config.js', () => {
+        jest.mock('@wdio/config', () => {
+            return {
+                configMock: jest.resetAllMocks()
```
praveendvd, line 405: @christian-bromann  i tried everything , but still the entire methods from config file is returning mocked values. I am not sure how to get the actual values

praveendvd: @christian-bromann 

https://github.com/webdriverio/webdriverio/blob/f51f23bae870f350aee8a71650fafc1eddfab5f0/packages/webdriver/tests/index.test.ts#L1

Adding unmok to first line of the test works. But it will affect other tests. nay tipson how to handle this

christian-bromann: No need to unmock this. What you want is to set a return value for the `getCapabilities` method. Or what are you trying to achieve?

praveendvd: @christian-bromann  I want to cover below tests , 

1. When config file not exist it throws the error
2. When capability not provided , it throws error
3. When capability not found it throws error

Happy path

1. Parses capability array from config.js
2. Parses capbility object from config.ts

praveendvd: @christian-bromann  sorted , could you have a look 😍

packages/wdio-cli/src/commands/repl.ts
```diff
@@ -14,7 +14,7 @@ const IGNORED_ARGS = [
 
 export const command = 'repl <option> [capabilities]'
 export const desc = 'Run WebDriver session in command line'
-export const cmdArgs = {
+export const cmdArgs: { [k in keyof ReplCommandArguments]?:yargs.Options } = {
```
christian-bromann, line 17:
```suggestion
export const cmdArgs: { [k in keyof ReplCommandArguments]?: yargs.Options } = {
```

packages/wdio-cli/tests/utils.test.ts
```diff
@@ -391,19 +391,54 @@ describe('getCapabilities', () => {
...
+        jest.unmock('@wdio/config')
+        jest.resetModules()
+        const getCapabilities = require('../src/utils').getCapabilities
+        expect(()=>getCapabilities({ option: './packages/wdio-cli/tests/__testData__/wdio.conf.js' } as any)).toThrowErrorMatchingSnapshot()
+    })
+
+    it('should through capability not found', () => {
+        jest.unmock('@wdio/config')
+        jest.resetModules()
+        const getCapabilities = require('../src/utils').getCapabilities
```
christian-bromann, line 419: This function is already imported, see https://github.com/webdriverio/webdriverio/pull/8156/files#diff-9f7a65a7afdec8311f3413b7434378fcb1535c60ce7e8b6609f297a3e9bb7b20R20

packages/wdio-cli/tests/utils.test.ts
```diff
@@ -391,19 +391,54 @@ describe('getCapabilities', () => {
...
+        expect(()=>getCapabilities({ option: 'packages/wdio-cli/tests/__testData__/wdio.conf.js', capabilities: 5 } as any)).toThrowErrorMatchingSnapshot()
+    })
+
+    it('should get capability from wdio.conf.js', () => {
+        jest.unmock('@wdio/config')
```
christian-bromann, line 424: I don't think it is a good idea to actually use the real configparser for these tests. Let's mock them out by importing the mock at the top of the file:
```ts
import { ConfigParser } from '@wdio/config'
```
and then specify functionalities of certain methods, e.g.:
```ts
jest.spyOn(ConfigParser.prototype, 'getCapabilities').mockReturnValue([{ browserName: 'chrome' }]);
```

This will make it easier to test all error cases for your if statement. Let me know if you have any questions to that.

christian-bromann:
Great progress @praveendvd , I added some comments above. I would like to unit test code more in isolation therefor it is better to mock out classes from other wdio packages.

praveendvd:
@christian-bromann  with unmock approach all tests are passing as expected . COuld you let me know  how to mock them. Also why the gitpod stage is coming up in my PR ?

packages/wdio-cli/tests/utils.test.ts
```diff
@@ -385,19 +385,59 @@ test('validateServiceAnswers', () => {
 })
 
 describe('getCapabilities', () => {
+    afterEach(() => {
+        jest.mock('@wdio/config')
```
christian-bromann, line 389: No need to mock, Jest automatically looks into all `__mocks__` directories and it therefor will find the mock here: `/packages/wdio-config/test/__mocks__/@wdio/config.ts`.
```suggestion
```

packages/wdio-cli/tests/utils.test.ts
```diff
@@ -385,19 +385,59 @@ test('validateServiceAnswers', () => {
...
+        jest.resetModules()
```
christian-bromann, line 390: This is not needed
```suggestion
```

packages/wdio-cli/tests/utils.test.ts
christian-bromann: The `addConfigFile` method is already mocked here: https://github.com/webdriverio/webdriverio/blob/main/packages/wdio-config/tests/__mocks__/@wdio/config.ts#L10. There should be no need to mock it. What error are you seeing?

christian-bromann (commented):
Can you rebase on top of current `main` branch? It will allow me to spin up your environment on Gitpod which makes it easier to review the PR. See details about what this is in the projectcommitters channel on Gitter

christian-bromann:
> Also why the gitpod stage is coming up in my PR ?

Ignore this for now, I will have to find out what's the best way to deal with that and what are the issues with it.

praveendvd:
@christian-bromann  Any suggestion on how to write unit test without mocking

christian-bromann:
@praveendvd it should work by spying on the methods, e.g.:

```ts
import { ConfigParser } from '@wdio/config'
const mock = jest.spyOn(ConfigParser, 'addConfigFile')
mock.mockImplementation(() => {
  throw new Error('ups')
})
```

praveendvd:
@christian-bromann  even the suggested mock approach doesnt work :( it takes the old implementation from __mocks__ any suggestions on how to fix this test

packages/wdio-cli/tests/utils.test.ts
```diff
@@ -385,25 +385,56 @@ test('validateServiceAnswers', () => {
...
     })
 
     it('should return driver with capabilities for desktop', () => {
         expect(getCapabilities({ option: 'chrome' } as any)).toMatchSnapshot()
     })
+
+    it('should throw config not found error', () => {
+        jest.mock('@wdio/config', () => {
+            return { ...jest.requireActual('@wdio/config') }
+        })
```
christian-bromann, line 410: Why do you want to use the actual package here and not the mock?

packages/wdio-cli/tests/__testData__/wdio.conf.js
```diff
@@ -0,0 +1,31 @@
+import path from 'path'
```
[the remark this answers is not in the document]
christian-bromann: do we still need the files in `__testData__`?

packages/wdio-protocols/src/index.ts
```diff
@@ -66,3 +66,9 @@ export {
     JSONWPCommandsAsync, MJSONWPCommandsAsync, SauceLabsCommandsAsync,
     SeleniumCommandsAsync, GeckoCommandsAsync
 }
+
+export const VALID_CAPS = [
```
christian-bromann, line 70: Can we rename this to `CAPABILITY_KEYS`

christian-bromann:
Re testing, let me take a look and see if we can manage to mock the config package.

christian-bromann:
Hey @praveendvd,
after a week of vacation I was finally able to look into this. It was a bit trickier than I thought. The problem was that spying [method classes](https://jestjs.io/docs/es6-class-mocks#mocking-a-specific-method-of-a-class) in this case didn't work because it seems like `jest.fn()` is not recognised as function in the prototype of the class. So what I had to do is write a new mock only for this file, e.g.:

```js
jest.mock('@wdio/config', () => ({
    ConfigParser: class ConfigParserMock {
        addConfigFile () {}
        autoCompile () {}
        getCapabilities () {}
    }
}))
```

which would allow me to mock on these function methods we use for the utility function. In order to mock these functions ahead of calling and initiating the `ConfigParser` you had to use `jest.spy` as follows:

```js
import { ConfigParser } from '@wdio/config'
const addConfigFileMock = jest.spyOn(ConfigParser.prototype, 'addConfigFile')
addConfigFileMock.mockImplementationOnce(() => {
    const error: any = new Error('ups')
    error.code = 'MODULE_NOT_FOUND'
    throw error
})
```

Issue #8376, opened 2022-05-26
[🐛 Bug]: CLI option "--spec" does not override the spec option defined in device capabilities, which can render that option useless


hammzj:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

latest

### Node.js Version

14.16.10

### Mode

WDIO Testrunner

### Which capabilities are you using?

```typescript
{
  "platform": "LINUX",
  "webStorageEnabled": false,
  "takesScreenshot": true,
  "javascriptEnabled": true,
  "databaseEnabled": false,
  "networkConnectionEnabled": true,
  "locationContextEnabled": false,
  "warnings": {},
  "desired": {
    "platformName": "Android",
    "goog:chromeOptions": {
      "args": [
        "--no-first-run",
        "--disable-fre",
        "--no-default-browser-check",
        "--webview-enable-modern-cookie-same-site"
      ]
    },
    "app": "/Users/zachhamm/Documents/GitHub/PROJECT/apps/android/APP.apk",
    "locale": "US",
    "language": "en",
    "orientation": "PORTRAIT",
    "automationName": "UiAutomator2",
    "noReset": false,
    "fullReset": true,
    "allowTestPackages": true,
    "newCommandTimeout": 240,
    "printPageSourceOnFindFailure": true,
    "setWebContentsDebuggingEnabled": true,
    "clearDeviceLogsOnStart": true,
    "logcatFormat": "brief",
    "deviceName": "pixel_3a",
    "avd": "Pixel_3a_API_31",
    "platformVersion": "12.0"
  },
  "platformName": "Android",
  "goog:chromeOptions": {
    "args": [
      "--no-first-run",
      "--disable-fre",
      "--no-default-browser-check",
      "--webview-enable-modern-cookie-same-site"
    ]
  },
  "app": "/Users/zachhamm/Documents/GitHub/PROJECT/apps/android/APP.apk",
  "locale": "US",
  "language": "en",
  "orientation": "PORTRAIT",
  "automationName": "UiAutomator2",
  "noReset": false,
  "fullReset": true,
  "allowTestPackages": true,
  "newCommandTimeout": 240,
  "printPageSourceOnFindFailure": true,
  "setWebContentsDebuggingEnabled": true,
  "clearDeviceLogsOnStart": true,
[...]

Issue #7735, opened 2021-11-23
[🐛 Bug]: Running a single scenario using the scenario line number is not working in @wdio/cucumber-framework version 7.11.0 and later


nirmalcrafty:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

latest

### Node.js Version

v14.18.1

### Mode

WDIO Testrunner

### Which capabilities are you using?

```typescript
{
"browserName": "chrome",
}
```

### What happened?

Update "@wdio/cucumber-framework" to any version higher than 7.10.1 and notice if you try to ,Run scenario by line number
eg 

wdio.js run wdio.conf.js --spec           
  ./tests/e2e/a.feature:5

It will run the entire feature file with all the scenarios

### What is your expected behavior?

The scenario which is present in the specified line number should be executed

### How to reproduce the bug.

In package.json upgrade the "@wdio/cucumber-framework": "7.16.6", do an npm install and run any scenario in a feature file by passing the below command

say

wdio.js run wdio.conf.js --spec ./tests/e2e/a.feature:5

Note all the scenarios in the a.feature will be running, instead of the scenario present in line number 5

### Relevant log output

```typescript
The results of all the scenarios executed is displayed
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

christian-bromann:
Thanks for reporting!

This might be a regression due to latest Cucumber update.

Any contributions that resolves the bug are highly appreciated. Please take a look into our [contribution guidelines](https://webdriver.io/docs/contribute) and let us know if you have any questions. Cheers!

nirmalcrafty:
Hi @christian-bromann , only relative paths, are not working, if we send the spec with the absolute path with a colon and line number it works

For eg in Windows

wdio.js run wdio.conf.js --spec c:/tests/e2e/a.feature:5 - This works

but not

wdio.js run wdio.conf.js --spec tests/e2e/a.feature:5

This is because the method getFeatureUriToLinesMapping in pickle_filter.js in @cucumber/cucumber package has now the uris constructed without an absolute path

const uri = path_1.default.normalize(match[1]);

and it used to be 
const uri = path_1.default.resolve(cwd, match[1]);

[...]

PR #8432, opened 2022-06-14
Fix for issue 8376

hammzj:
## Proposed changes
If `--spec` is defined on the command line, it should overwrite and ignore any spec patterns defined in both the configuration file and additionally in the capabilities array.

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

## Further comments

I can't say for sure if this is a breaking change, but if anyone has scripts pre-defined with `--spec` option, but they are anticipating that their capability-defined specs to run, this will affect them. However, we now have the option to define the hierarchy for spec priority as discussed in #8376:

> `--spec` > capability > config 

<a href="https://gitpod.io/#https://github.com/webdriverio/webdriverio/pull/8432"></a>

### DIFF
--- a/website/docs/OrganizingTestSuites.md
+++ b/website/docs/OrganizingTestSuites.md
@@ -1,19 +1,25 @@
 ---
-id: organizingsuites
-title: Organizing Test Suite
+id: organizingsuites title: Organizing Test Suite
 ---
 
-As projects grow, inevitably more and more integration tests are added. This increases build time and slows productivity.
+As projects grow, inevitably more and more integration tests are added. This increases build time and slows
+productivity.
 
-To prevent this, you should run your tests in parallel. WebdriverIO already tests each spec (or _feature file_ in Cucumber) in parallel within a single session. In general, try to test only a single feature per spec file. Try to not have too many or too few tests in one file. (However, there is no golden rule here.)
+To prevent this, you should run your tests in parallel. WebdriverIO already tests each spec (or _feature file_ in
+Cucumber) in parallel within a single session. In general, try to test only a single feature per spec file. Try to not
+have too many or too few tests in one file. (However, there is no golden rule here.)
 
-Once your tests have several spec files, you should start running your tests concurrently. To do so, adjust the `maxInstances` property in your config file. WebdriverIO allows you to run your tests with maximum concurrency—meaning that no matter how many files and tests you have, they can all run in parallel.  (This is still subject to certain limits, like your computer’s CPU, concurrency restrictions, etc.)
+Once your tests have several spec files, you should start running your tests concurrently. To do so, adjust
+the `maxInstances` property in your config file. WebdriverIO allows you to run your tests with maximum
+concurrency—meaning that no matter how many files and tests you have, they can all run in parallel.  (This is still
+subject to certain limits, like your computer’s CPU, concurrency restrictions, etc.)
 
 > Let's say you have 3 different capabilities (Chrome, Firefox, and Safari) and you have set `maxInstances` to `1`. The WDIO test runner will spawn 3 processes. Therefore, if you have 10 spec files and you set `maxInstances` to `10`, _all_ spec files will be tested simultaneously, and 30 processes will be spawned.
 
 You can define the `maxInstances` property globally to set the attribute for all browsers.
 
-If you run your own WebDriver grid, you may (for example) have more capacity for one browser than another. In that case, you can _limit_ the `maxInstances` in your capability object:
+If you run your own WebDriver grid, you may (for example) have more capacity for one browser than another. In that case,
+you can _limit_ the `maxInstances` in your capability object:
 
 ```js
 // wdio.conf.js
@@ -36,11 +42,14 @@ exports.config = {
 
 ## Inherit From Main Config File
 
-If you run your test suite in multiple environments (e.g., dev and integration) it may help to use multiple configuration files to keep things manageable.
+If you run your test suite in multiple environments (e.g., dev and integration) it may help to use multiple
+configuration files to keep things manageable.
 
-Similar to the [page object concept](PageObjects.md), the first thing you’ll need is a main config file. It contains all configurations you share across environments.
+Similar to the [page object concept](PageObjects.md), the first thing you’ll need is a main config file. It contains all
+configurations you share across environments.
 
-Then create another config file for each environment, and supplement the the main config with the environment-specific ones:
+Then create another config file for each environment, and supplement the the main config with the environment-specific
+ones:
 
 ```js
 // wdio.dev.config.js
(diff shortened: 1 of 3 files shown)


christian-bromann:
@hammzj thanks for raising the issue.

> I did not handle for `--exclude`, but only `--spec`. I can always go back and change this to also include `--exclude` to override any pre-defined config or capability patterns for that argument.

Can you explain a bit more what you mean by that. I believe we should be consistent with our features and therefore have the same order of things for the `--exclude` as well as for the `--spec` parameter.

Can you also add documentation around this? I am assuming https://webdriver.io/docs/organizingsuites#run-selected-tests would be a good place, what do you think?

hammzj:
@christian-bromann, I would need to test these possible cases for `--exclude`:

1. **Setup**: CLI-defined exclusion pattern + config-defined exclusion pattern
  a. Result: CLI-defined pattern should be used 
1. **Setup**:CLI-defined exclusion pattern + capability-defined exclusion pattern
  a. Result: CLI-defined pattern should be used 
1. **Setup**: Config-defined exclusion pattern + capability-defined exclusion pattern
  a. Result: Capability-defined pattern should be used 
1. **Setup**: Config-defined exclusion pattern + capability-defined exclusion pattern + CLI-defined exclusion pattern
  a. Result: CLI-defined pattern should be used 

As for the documentation, I'll take a look, but I find the documentation to be very concise as to what I am attempting to do. I can add a note/warning about the which spec pattern will take priority when defined for the driver.

hammzj:
I added new documentation. Please review and let me know if it's adequate!

website/docs/OrganizingTestSuites.md
```diff
@@ -1,19 +1,25 @@
 ---
-id: organizingsuites
-title: Organizing Test Suite
+id: organizingsuites title: Organizing Test Suite
 ---
 
-As projects grow, inevitably more and more integration tests are added. This increases build time and slows productivity.
+As projects grow, inevitably more and more integration tests are added. This increases build time and slows
```
christian-bromann, line 5: Can we revert all these `\n` changes?

website/docs/OrganizingTestSuites.md
```diff
@@ -1,19 +1,25 @@
 ---
-id: organizingsuites
-title: Organizing Test Suite
+id: organizingsuites title: Organizing Test Suite
```
christian-bromann, line 2:
```suggestion
id: organizingsuites
title: Organizing Test Suite
```

website/docs/OrganizingTestSuites.md
```diff
@@ -104,37 +113,43 @@ wdio wdio.conf.js --suite login --suite otherFeature
 
 ## Grouping Test Specs To Run Sequentially
 
-As described above, there are benefits in running the tests concurrently.  However, there are cases where it would be beneficial to group tests together to run sequentially in a single instance.  Examples of this are mainly where there is a large setup cost e.g. transpiling code or provisioning cloud instances, but there are also advanced usage models that benefit from this capability.
+As described above, there are benefits in running the tests concurrently. However, there are cases where it would be
```
christian-bromann, line 116: same here, you can configure your IDE to break lines automatically

website/docs/OrganizingTestSuites.md
```diff
@@ -104,37 +113,43 @@ wdio wdio.conf.js --suite login --suite otherFeature
...
-        ],
-        "./test/specs/test_b*.js",
-    ],
+[
+"./test/specs/test_login.js",
+"./test/specs/test_product_order.js",
+"./test/specs/test_checkout.js"
+],
+"./test/specs/test_b*.js",
+],
```
christian-bromann, line 131: Can we revert this as well?

website/docs/OrganizingTestSuites.md
```diff
@@ -104,37 +113,43 @@ wdio wdio.conf.js --suite login --suite otherFeature
...
-        ],
-        allb: ["./test/specs/test_b*.js"]
+end2end: [
+[
+"./test/specs/test_login.js",
+"./test/specs/test_product_order.js",
+"./test/specs/test_checkout.js"
+]
+],
+allb: ["./test/specs/test_b*.js"]
```
christian-bromann, line 148: same

website/docs/OrganizingTestSuites.md
hammzj: Sure -- I think it might have been from a formatting shortcut that wraps lines.

hammzj:
Reverted commit  to prevent line breaks, then updated the docs with only the new changes

website/docs/OrganizingTestSuites.md
```diff
@@ -166,6 +166,9 @@ Note that each test file is running in a single test runner process. Since we do
 
 This feature will help you to accomplish the same goal.
 
+When the `--spec` option is provided, it will override any patterns defined by the config or capability level's `specs`
+parameter.
```
christian-bromann, line 170:
```suggestion
When the `--spec` option is provided, it will override any patterns defined by the config or capability level's `specs` parameter.
```

website/docs/OrganizingTestSuites.md
```diff
@@ -188,6 +191,9 @@ Or, exclude a spec file when filtering using a suite:
 wdio wdio.conf.js --suite login --exclude ./test/specs/e2e/login.js
 ```
 
+When the `--exclude` option is provided, it will override any patterns defined by the config or capability
+level's `exclude` parameter.
```
christian-bromann, line 195:
```suggestion
When the `--exclude` option is provided, it will override any patterns defined by the config or capability level's `exclude` parameter.
```

website/docs/OrganizingTestSuites.md
```diff
@@ -228,3 +234,80 @@ This is helpful with large test suites when you already know that your build wil
...
+
+## Hierarchy of spec patterns
+
+When declaring what specs to run, there is a certain hierarchy defining what pattern will take precedence when defined.
+Currently, this is how it works, from highest priority to lowest:
+> CLI `--spec` argument > capability `specs` pattern > config `specs` pattern
+
+If only the config parameter is given, it will be used for all capabilities. However, if defining the pattern at the
+capability level, it will be used instead of the config pattern. Finally, any spec pattern defined on the command line
+will override all other patterns given.
```
christian-bromann, line 247:
```suggestion
When declaring what specs to run, there is a certain hierarchy defining what pattern will take precedence when defined. Currently, this is how it works, from highest priority to lowest:

> CLI `--spec` argument > capability `specs` pattern > config `specs` pattern

If only the config parameter is given, it will be used for all capabilities. However, if defining the pattern at the capability level, it will be used instead of the config pattern. Finally, any spec pattern defined on the command line will override all other patterns given.
```

website/docs/OrganizingTestSuites.md
```diff
@@ -228,3 +234,80 @@ This is helpful with large test suites when you already know that your build wil
...
+
+### Using capability-defined spec patterns
+
+When you define a spec pattern at the capability level, it will override any patterns defined at the config level. This
+is useful when needing to separate tests based on differentiating device capabilities. In cases like this, it is more
+useful to use a generic spec pattern at the config level, and more specific patterns at the capability level.
+
+For example, let's say you had two directories, with one for Android tests, and one for iOS tests.
+
+Your config file may define the pattern as such, for non-specific device tests:
```
christian-bromann, line 257:
```suggestion
When you define a spec pattern at the capability level, it will override any patterns defined at the config level. This is useful when needing to separate tests based on differentiating device capabilities. In cases like this, it is more useful to use a generic spec pattern at the config level, and more specific patterns at the capability level.

For example, let's say you had two directories, with one for Android tests, and one for iOS tests.

Your config file may define the pattern as such, for non-specific device tests:
```

website/docs/OrganizingTestSuites.md
```diff
@@ -228,3 +234,80 @@ This is helpful with large test suites when you already know that your build wil
...
+
+```js
+{
+    specs: ['tests/general/**/*.js']
+}
+```
+
+but then, you will have different capabilities for your Android and iOS devices, where the patterns could look like
+such:
```
christian-bromann, line 266:
```suggestion
but then, you will have different capabilities for your Android and iOS devices, where the patterns could look like such:
```

website/docs/OrganizingTestSuites.md
```diff
@@ -228,3 +234,80 @@ This is helpful with large test suites when you already know that your build wil
...
+{
+  "platformName": "iOS",
+  "specs": [
+    "tests/ios/**/*.js"
+  ]
+}
+```
+
+If you require both of these capabilities in your config file, then the Android device will only run the tests under
+the "android" namespace, and the iOS tests will run only tests under the "ios" namespace!
```
christian-bromann, line 287:
```suggestion
If you require both of these capabilities in your config file, then the Android device will only run the tests under the "android" namespace, and the iOS tests will run only tests under the "ios" namespace!
```

christian-bromann (commented):
minor formatting

website/docs/OrganizingTestSuites.md
```diff
@@ -228,3 +232,74 @@ This is helpful with large test suites when you already know that your build wil
 The `bail` option expects a number, which specifies how many test failures can occur before WebDriver stop the entire testing run. The default is `0`, meaning that it always runs all tests specs it can find.
 
 Please see [Options Page](Options.md) for additional information on the bail configuration.
+
+
```
christian-bromann, line 236:
```suggestion

```

mikesalvia:
@hammzj so I figured out kind of what happened. Since "--spec" is provided we skip all config parsing which in turn is preventing filtering. I think we need to check if it spec is provided and clear specs from configuration then still parse the config.

hammzj:
My original thought with this change is that it makes sense to skip all other filtering because a user-defined `--spec` option on the command line _should_ override anything else passed in, including config or capability-defined spec/exclusion patterns. A user is choosing to specify what files to run should take preference over anything else defined in a configuration file, even if the device may not be suited to run that spec.

I looked at #7735 and if this PR is preventing the use of feature file line numbers, then I can understand the need to make a change here; however, it seems that in terms of the order of precedence, then the Cucumber line number would be equal to the level of where the CLI-defined spec is, which will require some new criteria around this concept.

If I'm reading this correctly, then
> **Cucumber line number** OR `--spec` > capability patterns > config patterns

In this case, when both the line number and `--spec` are provided to a script, we would need to perform filtering for both the line number and what would be contained in the spec option, merge them in the array of test files, and then skip over the capability or config patterns as it has been doing since this PR. If only the line number is provided, then it needs to filter and add them to the test files array before we look at the `--spec` option entirely. 

Else, we could just allow providing `**/*.feature:##:##` into the `--spec` option which would be the same as `cucumber **/*.feature:##:##` without needing to handle both options independently.

These solutions seems to make the most sense in retaining both filtering possibilities.

Let me know if this is helpful!

PR #8465, opened 2022-06-23
Allow to skip calling `releaseAction` command for click

### LINKED ISSUE #8440: [🐛 Bug]: Make `releaseAction` call within `click` command optional

### WebdriverIO Version

latest

### Node.js Version

6.14.15

### Mode

WDIO Testrunner

```typescript
{
    browserName: 'firefox',
    browserVersion: '101.0.1',
}
```

### What happened?

[This page ](https://the-internet.herokuapp.com/context_menu) contains a box that will trigger a JavaScript alert when right-clicking in it. After accepting the alert, the actual context menu appears, as seen in the attachment image. However, performing a right-click with WebdriverIO 'click' method causes the alert to disappear really quickly, thus never being able to retrieve the alert text with `browser.getAlertText()`.  Instead, using dispatchEvent call on the element does not lead to the alert disappearing.

image

The right-clicking action should perform the same output as if done manually.

### How to reproduce the bug.

```ts
describe('Context Menu Test', () => {
    it ('should alert on right click', async() => {
        await browser.url('https://the-internet.herokuapp.com/context_menu')
        await browser.$('#hot-spot').click( { button: 2 } )

        // However, this keeps the alert visible until accepted or dismissed.

        /*await browser.$('#hot-spot').$(function() {
            (this as HTMLElement).dispatchEvent(new Event('contextmenu'))
        })*/

        const alertText = await browser.getAlertText();
        await browser.acceptAlert();
        expect(alertText).toBe('You selected a context menu')
    })
})
```

### Relevant log output

```typescript
[0-0] RUNNING in firefox - C:Users\me\webdriverio-getting-started\test\specs\contextmenu.spec.ts
[0-0] 2022-06-20T11:14:40.275Z ERROR webdriver: Request failed with status 404 due to no such alert: unknown error
[0-0] Error in "Context Menu Tests should alert on right click"
no such alert: unknown error
    at async UserContext.<anonymous> (C:\Users\me\webdriverio-getting-started\test\specs\contextmenu.spec.ts:10:27)
[0-0] FAILED in firefox - C:\Users\me\webdriverio-getting-started\test\specs\contextmenu.spec.ts
```

### Code of Conduct

praveendvd:
## Proposed changes

fixes #8440

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

<a href="https://gitpod.io/#https://github.com/webdriverio/webdriverio/pull/8465"></a>

### DIFF
--- a/packages/webdriverio/src/commands/element/click.ts
+++ b/packages/webdriverio/src/commands/element/click.ts
@@ -8,8 +8,13 @@ const log = logger('webdriverio/click')
  *
  * Click on an element.
  *
- * Note: This issues a WebDriver `click` command for the selected element, which generally scrolls to and then clicks the
- * selected element. However, if you have fixed-position elements (such as a fixed header or footer) that cover up the
+ * This issues a WebDriver `click` command for the selected element , which generally scrolls to and then clicks the
+ * selected element when no options are passed. When options object is passed it uses action class instead of webdriver click which
+ * give added capabilities like passing button type, coordinates etc. By default, when using options a release action
+ * command is send after performing the click action, this could cause some alerts to be closed unexpectedly
+ * (issue: https://github.com/webdriverio/webdriverio/issues/6423) To fix this pass `option.skipRelease`.
+ *
+ * Note: If you have fixed-position elements (such as a fixed header or footer) that cover up the
  * selected element after it is scrolled within the viewport, the click will be issued at the given coordinates, but will
  * be received by your fixed (overlaying) element. In these cased the following error is thrown:
  *
@@ -64,6 +69,10 @@ const log = logger('webdriverio/click')
         const myButton = await $('#myButton')
         await myButton.click({ button: 2, x: 30, y: 40 }) // opens the contextmenu 30 horizontal and 40 vertical pixels away from location of the button (from the center of element)
     })
+    it('should skip sending releaseAction command that cause unexpected alert closure', async () => {
+        const myButton = await $('#myButton')
+        await myButton.click({ button: 2, x: 30, y: 40, skipRelease:true }) //skips sending releaseActions
+    })
  * </example>
  *
  * @alias element.click
@@ -75,7 +84,7 @@ const log = logger('webdriverio/click')
  * @param {number=}           options.y      Number (optional)
  * @param {number=}           options.skipRelease         Boolean (optional)
  */
-export default async function click (
+export default async function click(
     this: WebdriverIO.Element,
     options?: ClickOptions
 ) {
--- a/packages/webdriverio/src/commands/element/click.ts
+++ b/packages/webdriverio/src/commands/element/click.ts
@@ -11,8 +11,7 @@ const log = logger('webdriverio/click')
  * This issues a WebDriver `click` command for the selected element , which generally scrolls to and then clicks the
  * selected element when no options are passed. When options object is passed it uses action class instead of webdriver click which
  * give added capabilities like passing button type, coordinates etc. By default, when using options a release action
- * command is send after performing the click action, this could cause some alerts to be closed unexpectedly
- * (issue: https://github.com/webdriverio/webdriverio/issues/6423) To fix this pass `option.skipRelease`.
+ * command is send after performing the click action, pass `option.skipRelease=true` to skip this action.
  *
  * Note: If you have fixed-position elements (such as a fixed header or footer) that cover up the
  * selected element after it is scrolled within the viewport, the click will be issued at the given coordinates, but will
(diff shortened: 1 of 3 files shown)


packages/webdriverio/src/commands/element/click.ts
```diff
@@ -73,6 +73,7 @@ const log = logger('webdriverio/click')
  * @param {string= | number=} options.button can be one of [0, "left", 1, "middle", 2, "right"] (optional)
  * @param {number=}           options.x      Number (optional)
  * @param {number=}           options.y      Number (optional)
+ * @param {number=}           options.skipRelease         Number (optional)
```
christian-bromann, line 76:
```suggestion
 * @param {number=}           options.skipRelease         Boolean (optional)
```

christian-bromann (commented):
Could we add a test for it?

erwinheitzman:
Quick question, what does the release action do exactly?Does it mean it holds down the button until released?
If so, I would like to suggest is to change `skipRelease`  to `holdButton` which can be set to true. This name makes a lot more sense to me as a user as no user knows the WebdriverIO internals and they will have to guess what it does when reading while holdButton is clear right away.

The above only applies if my assumption is correct and can be ignored otherwise but please do share what the release action does :)

christian-bromann:
@erwinheitzman great comment! It is easy to confuse this but release action just tells the driver that no more actions are following. So you can still run an action command that does: mouse down, mouse move, mouse up but still don't call release.  From what I experienced by listening to other folks issues is that every implementation here seems to behave differently based on whether release was called or not and this is just to provide control over this.

My goal with the action class implementation is to get rid of all these uncertainties as we will have battle tested interactions. From there we can learn how to modify them and make them work for even more usecases without breaking the ones we already cover.

praveendvd:
@christian-bromann updated docs and added tests. Going through test i saw some issues : 

https://github.com/webdriverio/webdriverio/blob/a4d1500b69e3716d9576f135ed9d41673da85f9e/packages/webdriverio/tests/commands/element/click.test.ts#L41

here we are not clearing the mock , so the test is actually validating the previous mock call not the new once we should be valudating got.mock.calls[4] or something not [2] . [2] is previous mock call

packages/webdriverio/src/commands/element/click.ts
```diff
@@ -8,8 +8,13 @@ const log = logger('webdriverio/click')
  *
  * Click on an element.
  *
- * Note: This issues a WebDriver `click` command for the selected element, which generally scrolls to and then clicks the
- * selected element. However, if you have fixed-position elements (such as a fixed header or footer) that cover up the
+ * This issues a WebDriver `click` command for the selected element , which generally scrolls to and then clicks the
+ * selected element when no options are passed. When options object is passed it uses action class instead of webdriver click which
+ * give added capabilities like passing button type, coordinates etc. By default, when using options a release action
+ * command is send after performing the click action, this could cause some alerts to be closed unexpectedly
+ * (issue: https://github.com/webdriverio/webdriverio/issues/6423) To fix this pass `option.skipRelease`.
```
christian-bromann, line 15: I am not sure if it is worth mentioning this here because we don't really know what was responsible for the alert to disappear and if this case is true for all similar scenarios.

christian-bromann: I think we can simply say, if this flag is set WebdriverIO won't call the [`releaseAction`](https://w3c.github.io/webdriver/webdriver-spec.html#releasing-actions) command

christian-bromann (commented):
> here we are not clearing the mock before this click , so the test is actually validating the previous mock call not the new once we should be valudating got.mock.calls[4] or something not [2] . [2] is previous mock call

Then we should call `mockClear` in a test hook

packages/webdriverio/src/commands/element/click.ts
```diff
@@ -64,6 +68,10 @@ const log = logger('webdriverio/click')
         const myButton = await $('#myButton')
         await myButton.click({ button: 2, x: 30, y: 40 }) // opens the contextmenu 30 horizontal and 40 vertical pixels away from location of the button (from the center of element)
     })
+    it('should skip sending releaseAction command that cause unexpected alert closure', async () => {
+        const myButton = await $('#myButton')
+        await myButton.click({ button: 2, x: 30, y: 40, skipRelease:true }) //skips sending releaseActions
```
christian-bromann, line 73:
```suggestion
        await myButton.click({ button: 2, x: 30, y: 40, skipRelease:true }) // skips sending releaseActions
```

christian-bromann (approved):
Thanks, LGTM 👍

PR #4065, opened 2022-06-30
Simplify special keys in add/set value commands

christian-bromann:
Currently to set some value and press enter the following is required:

```js
browser.addValue(["foobar", "Enter"])
```

How about simplifying this down to:

```js
browser.addValue("foobar {Enter}")
```

?

<a href="https://gitpod.io/#https://github.com/webdriverio/webdriverio/pull/4065"></a>

### DIFF
--- a/packages/webdriverio/src/commands/element/addValue.ts
+++ b/packages/webdriverio/src/commands/element/addValue.ts
@@ -1,29 +1,13 @@
-import logger from '@wdio/logger'
-import { transformToCharString } from '../../utils/index.js'
-
-const log = logger('addValue')
-
-export type CommandOptions = {
-    translateToUnicode?: boolean
-}
-
-export type Value = string | number
-
-const isNumberOrString = (input: unknown) => typeof input === 'string' || typeof input === 'number'
-
-const isValidType = (value: unknown) => (
-    isNumberOrString(value) ||
-    Array.isArray(value) && value.every((item) => isNumberOrString(item))
-)
-
 /**
  *
- * Add a value to an object found by given selector. You can also use unicode
- * characters like Left arrow or Back space. WebdriverIO will take care of
- * translating them into unicode characters. You’ll find all supported characters
- * [here](https://w3c.github.io/webdriver/webdriver-spec.html#keyboard-actions).
- * To do that, the value has to correspond to a key from the table. It can be disabled
- * by setting `translateToUnicode` optional parameter to false.
+ * Add a value to an input or textarea element found by given selector.
+ *
+ * :::info
+ *
+ * If you like to use special characters, e.g. to copy and paste a value from one input to another, use the
+ * [`keys`](/docs/api/browser/keys) command.
+ *
+ * :::
  *
  * <example>
     :addValue.js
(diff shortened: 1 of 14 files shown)


christian-bromann:
This patch now contains following change:

- Sending an array of input values (the way it was allowed using JSONWireProtocol) to `setValue` and `addValue` is not possible anymore, these commands only take `string` or `number` values
- A new `Key` interface is exported of the `webdriverio` module that has a typed interface for all special and modifier keys
- This interface also contains a `Ctrl` property that sends the control modifier key based on the operating system the browser is running, e.g. `Command` on Mac or `Control` on Windows or Linux

packages/webdriverio/src/constants.ts
```diff
@@ -320,3 +321,70 @@ export const ERROR_REASON = [
...
+    F4: UNICODE_CHARACTERS.F4,
+    F5: UNICODE_CHARACTERS.F5,
+    F6: UNICODE_CHARACTERS.F6,
+    F7: UNICODE_CHARACTERS.F7,
+    F8: UNICODE_CHARACTERS.F8,
+    F9: UNICODE_CHARACTERS.F9,
+    F10: UNICODE_CHARACTERS.F10,
+    F11: UNICODE_CHARACTERS.F11,
+    F12: UNICODE_CHARACTERS.F12,
+    Command: UNICODE_CHARACTERS.Command,
```
erwinheitzman, line 388: In the w3c specs this key is called the `Meta` key. There's also a differentiation between left and right but I'm not sure if that matters, since it's in the spec I think it was worthy to mention both :)

Reference: https://www.w3.org/TR/uievents-code/#key-alphanumeric-functional

packages/webdriverio/src/commands/element/addValue.ts
```diff
@@ -38,23 +22,24 @@ const isValidType = (value: unknown) => (
...
-
-    if (!this.isW3C) {
-        return this.elementSendKeys(this.elementId, transformToCharString(value, translateToUnicode) as any as string)
+    /**
+     * The JSONWireProtocol allowed array values and use of special characters when adding a value to an input.
+     * With the W3C protocol this was not possible anymore. This is a type check to ensure users are aware of
+     * this transition.
+     */
+    if (Array.isArray(value) || !value || typeof value === 'boolean') {
+        throw new Error(
```
erwinheitzman, line 38: How do you feel about using only one type (string)?

I have personally never liked using anything else since it feels odd to throw anything but a string into an input element. It's not a strong opinion but I'm curious if you feel the same.

erwinheitzman (commented):
Nice work Christian, left some small comments for you to read 👍

erwinheitzman:
Would you say that like having the webdriverio.Key property, I could use the same approach for the Actions API that I was working on, so we would also have a webdriverio.Actions property?

packages/webdriverio/src/constants.ts
christian-bromann: Interesting! The [WebDriver spec](https://w3c.github.io/webdriver/webdriver-spec.html#keyboard-actions) defines two different values for `Meta` which I assume is the one for left and right. I will revisit these tables to see if they were misinterpreted or got updated.

packages/webdriverio/src/commands/element/addValue.ts
christian-bromann: No strong opinion either but I tend towards allowing `number` as well because it simplifies usage when an number is being passed in e.g. from parsing a CSV file or whatever.

christian-bromann:
> Would you say that like having the webdriverio.Key property, I could use the same approach for the Actions API that I was working on, so we would also have a webdriverio.Actions property?

Yep, that would be the plan I suggest. I think it is the best package to have these interface exported because it would work for standalone as well as for wdio testrunner while these interfaces are not needed for the protocol modules (`webdriver` or `devtools`).

Issue #6346, opened 2021-01-25
afterSession hook is clearing debug log


RafalSkorka:
**Environment (please complete the following information):**
 - **WebdriverIO version:** 6.12.1
 - **Mode:** WDIO Testrunner
 - **If WDIO Testrunner, running sync/async:** sync
 - **Node.js version:** 12.16.1
 - **NPM version:** 6.14.11
 - **Browser name and version:** Chrome 88
 - **Platform name and version:** Windows 10
 - **Additional wdio packages used (if applicable):** none

**Config of WebdriverIO**
Generated with `npx wdio config -y ` with these changes:
```
    outputDir: 'reports',
    logLevel: 'debug',
    afterSession: function (config, capabilities, specs) {
        console.log(`afterSession hook!`);
    },
```
**Describe the bug**
When `outputDir` and `afterSessions` are defined, setting logLevel to values 'debug' or 'trace' makes the `wdio-0-0.log` almost empty.
It is because this log is cleared after `deleteSession` command is executed, but `executeHooksWithArgs` function logs a message there afterwards clearing its previous contents. 

**To Reproduce**
* Run the `Getting Started` steps.
* Modify the config as above.

**Expected behavior**
The `wdio-0-0.log` is not cleared in the `afterSession` hook.

**Log**
Here is the only entry after running such project:
```
2021-01-25T22:00:24.094Z DEBUG @wdio/sync: Finished to run "afterSession" hook in 0ms
```

**Additional context**
This was introduced with https://github.com/webdriverio/webdriverio/issues/6068

christian-bromann:
I have experienced this too and was wondering how I reproduced it. Thanks for taking the initial research, you want to take a stab at it?

RafalSkorka:
I am sorry, I will not be able to tackle this soon :(

wyhmichael:
I am experiencing exactly the same with the same setup 

```
WebdriverIO version: 6.12.1
Mode: WDIO Testrunner
If WDIO Testrunner, running sync/async: sync
Node.js version: 12.16.1
NPM version: 6.14.11
Browser name and version: Chrome 88
Platform name and version: Windows 10
Additional wdio packages used (if applicable): none
```

christian-bromann:
This will be fixed as part of #6577

RafalSkorka:
@christian-bromann I am still experiencing this issue after updating to 7.2.0 :(

wyhmichael:
[...]

Issue #6068, opened 2020-11-05
Cannot change outputDir for a new Appium session while testing


RedMickey:
Hello,
I am a maintainer of [React Native Tools](https://github.com/microsoft/vscode-react-native) VS Code extension, which is used for debugging React Native applications in VS Code. We use WebdriverIO package for testing of the extension.

**Environment:**
 - **WebdriverIO version:** 6.6.7
 - **Mode:** Standalone mode
 - **Node.js version:** 10.19.0
 - **NPM version:** 6.13.4
 - **Platform names and versions:** Windows 10, macOS 10.15.7, Linux Ubuntu 18.04
 - **Appium**: 1.18.2

**Describe the bug**
We use `Appium` as a server. In our tests we create a WebdriverIO session per each test. We'd like to have logs from each session in separate `wdio.log` files. We specify `outputDir` parameter for each session, but the only first configuration is taken into account. So logs from the second session are written in the `.log` file of the first session and so on.

**To Reproduce**
Try to create 2 WebdriverIO sessions sequentially with different `outputDir` parameters.

1. Start Appium using the command `appium` in the terminal
2. Add capabilities to the configurations below and run the script

```js
const wdio = require("webdriverio");

const opts1 = {
    path: '/wd/hub',
    port: 4723,
    capabilities: { /*any capabilities*/ },
    logLevel: "trace",
    outputDir: `${__dirname}/1`
};

const opts2 = {
  path: '/wd/hub',
  port: 4723,
  capabilities: { /*any capabilities*/ },
  logLevel: "trace",
  outputDir: `${__dirname}/2`
};
  
  async function main () {
    let client = await wdio.remote(opts1);

    console.log("ses1-sleep");
    await sleep(4000);

    await client.deleteSession();

    console.log("ses1 - end");
    await sleep(15000);

    console.log("ses2");
    client = await wdio.remote(opts2);
  
    console.log("ses2-sleep");
    await sleep(6000);

    console.log("ses2 - pre end");
    await client.deleteSession();
  }
  
  main();

async function sleep(time) {
  await new Promise(resolve => {
      const timer = setTimeout(() => {
          clearTimeout(timer);
          resolve();
      }, time);
  });
}
```

After that logs from the both sessions were written to the log file for the first session
![image](https://user-images.githubusercontent.com/33267199/98209318-c6049a80-1f4f-11eb-96a3-5e66b0668a13.png)

**Expected behavior**
[...]

PR #8506, opened 2022-07-08
patch for 6068, adding a flag about clearing logger on session end

wyhmichael:
## Proposed changes

This is patch for #6068, and workaround for issue mentioned in https://github.com/webdriverio/webdriverio/issues/6346#issuecomment-1179299272, by adding a flag to allow turn on/off logger cleanup on session end

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

<a href="https://gitpod.io/#https://github.com/webdriverio/webdriverio/pull/8506"></a>

### DIFF
--- a/packages/webdriver/src/command.ts
+++ b/packages/webdriver/src/command.ts
@@ -101,7 +101,7 @@ export default function (
 
             this.emit('result', { method, endpoint, body, result })
 
-            if (command === 'deleteSession' && process.env.WDIO_WORKER_ID) {
+            if (command === 'deleteSession' && !process.env.WDIO_WORKER_ID) {
                 logger.clearLogger()
             }
             return result.value
--- a/packages/webdriver/src/command.ts
+++ b/packages/webdriver/src/command.ts
@@ -101,7 +101,7 @@ export default function (
 
             this.emit('result', { method, endpoint, body, result })
 
-            if (command === 'deleteSession' && !process.env.KEEP_WDIO_LOGGER) {
+            if (command === 'deleteSession' && process.env.WDIO_WORKER_ID) {
                 logger.clearLogger()
             }
             return result.value
--- a/packages/webdriver/src/command.ts
+++ b/packages/webdriver/src/command.ts
@@ -101,7 +101,7 @@ export default function (
 
             this.emit('result', { method, endpoint, body, result })
 
-            if (command === 'deleteSession' && this.options.clearLoggerOnSessionEnd) {
+            if (command === 'deleteSession' && !process.env.KEEP_WDIO_LOGGER) {
                 logger.clearLogger()
             }
             return result.value
--- a/packages/webdriver/src/command.ts
+++ b/packages/webdriver/src/command.ts
@@ -101,7 +101,7 @@ export default function (
 
             this.emit('result', { method, endpoint, body, result })
 
-            if (command === 'deleteSession') {
+            if (command === 'deleteSession' && this.options.clearLoggerOnSessionEnd) {
                 logger.clearLogger()
             }
             return result.value
--- a/packages/webdriver/tests/command.test.ts
+++ b/packages/webdriver/tests/command.test.ts
@@ -75,31 +75,6 @@ const scope: BaseClient = new FakeClient()
 type mockResponse = (...args: any[]) => any
 
 describe('command wrapper', () => {
-    beforeEach(() => {
-        delete process.env.KEEP_WDIO_LOGGER
-    })
-    
-    function doProperRequest (additionalParams: number | undefined = undefined) {
-        const commandFn = commandWrapper(commandMethod, commandPath, commandEndpoint)
-        const requestMock = require('../src/request/node')
-        const resultFunction = commandFn.call(scope, '123', 'css selector', '#body', additionalParams) as unknown as mockResponse
-        expect(requestMock.mock.calls).toHaveLength(1)
-        expect(resultFunction({ value: 14 })).toBe(14)
-        if (additionalParams) {
-            expect(resultFunction({ value: 'foobarboo' })).toBe('foobarboo')
-        }
-
-        const [method, endpoint, { using, value, customParam }] = requestMock.mock.calls[0]
-        expect(method).toBe('POST')
-        expect(endpoint).toBe('/session/:sessionId/element/123/element')
-        expect(using).toBe('css selector')
-        expect(value).toBe('#body')
-        if (additionalParams) {
-            expect(customParam).toBe(additionalParams)
-        }
-        requestMock.mockClear()
-    }
-    
     it('should fail if wrong arguments are passed in', () => {
         const commandFn = commandWrapper(commandMethod, commandPath, commandEndpoint)
         expect(commandFn)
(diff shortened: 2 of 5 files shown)


christian-bromann:
Adding an option to provide a workaround is not an ideal solution. Can we analyse what the use cases are here and maybe find a solution that doesn't require to add this option?

wyhmichael:
Hi @christian-bromann sorry for the delay, I made a sample repo to show the use cases of wido logger - https://github.com/wyhmichael/wdio-logger-afterSession-issue

There're 3 tests: 

1. loggerOnAfterSessionHook.e2e.ts (runner mode)
2. loggerOnSessionReload.e2e.ts (runner mode)
3. sequentialWDIOSessions.e2e.js (standalone mode) : This test is meant to simulate the issue mentioned in https://github.com/webdriverio/webdriverio/issues/6068

With the introduction of `clearLogger` in https://github.com/webdriverio/webdriverio/issues/6068, the use case 1 and 2 above are behaving like how it was before where all the logs would be written and kept in the log file, now the log file would be cleared whenever `afterSession` and `reloadSession` called (which trigger `deleteSession` command to trigger `clearLogger`)

christian-bromann:
How about we disable the `clearLogger` function when the testrunner is running, e.g. `WDIO_LOG_PATH` is set as environment variable. This would still allow the use case for what it got implemented and would fix the bug you are experiencing.

christian-bromann:
It would be:
```js
if (!process.env.WDIO_LOG_PATH) {
  logger.clearLogger();
}
```

but yes!

wyhmichael:
Thanks @christian-bromann! 

I tried to update [this condition](https://github.com/webdriverio/webdriverio/pull/8506/files#diff-b08c37e616f7015c3c66a75fde5d6a5f9e58955dcc8be341320d65536e492037R104) to be 

```
if (command === 'deleteSession' && !process.env.WDIO_LOG_PATH) {
```

but looks like [`process.env.WDIO_LOG_PATH`](https://github.com/webdriverio/webdriverio/search?q=WDIO_LOG_PATH) always got set in both standalone and runner mode

And also, I tried more with [sequentialWDIOSessions.e2e.js](https://github.com/wyhmichael/wdio-logger-afterSession-issue/blob/main/test/specs/sequentialWDIOSessions.e2e.js#L46), looks like both `process.env.WDIO_LOG_PATH` and `clearLogger` needed in order to switch log path at run time.

So looks like the following would not work, but I am not 100% sure if I use it correctly, any further suggestions?
```
if (!process.env.WDIO_LOG_PATH) {
  logger.clearLogger();
}
```

wyhmichael:
Thanks @christian-bromann ! updated the PR, please take another look 👍

christian-bromann:
> I didn't find any docs to update

I think I will cover this with v8.

Can we extend the unit tests to ensure we cover this code section?

packages/webdriver/src/command.ts
```diff
@@ -101,7 +101,7 @@ export default function (
 
             this.emit('result', { method, endpoint, body, result })
 
-            if (command === 'deleteSession') {
+            if (command === 'deleteSession' && !process.env.KEEP_WDIO_LOGGER) {
```
christian-bromann, line 104: As I mentioned in my post, let's define a `WDIO_WORKER_CID` environment variable within `@wdio/local-runner` and check against its existence here.

packages/webdriver/tests/command.test.ts
```diff
@@ -75,6 +75,31 @@ const scope: BaseClient = new FakeClient()
...
+        expect(method).toBe('POST')
+        expect(endpoint).toBe('/session/:sessionId/element/123/element')
+        expect(using).toBe('css selector')
+        expect(value).toBe('#body')
+        if (additionalParams) {
+            expect(customParam).toBe(additionalParams)
+        }
+        requestMock.mockClear()
+    }
+
```
christian-bromann, line 102: I don't think these tests are relevant to the change you made above, let's revert them.

packages/webdriver/tests/command.test.ts
```diff
@@ -258,18 +282,43 @@ describe('command wrapper result log', () => {
...
+        clearLoggerSpy.mockClear()
+    })
+
+    it('should NOT call clearLogger on deleteSession cmd if env var - KEEP_WDIO_LOGGER set', () => {
+        const resultFunction = getRequestCallback(deleteSessionCmd.method, deleteSessionCmd.path, deleteSessionCmd.endpoint) as unknown as mockResponse
+        process.env.KEEP_WDIO_LOGGER = 'true'
+        resultFunction({})
+        expect((log.info as jest.Mock).mock.calls).toHaveLength(0)
+        expect(clearLoggerSpy).not.toHaveBeenCalled()
+        clearLoggerSpy.mockClear()
```
christian-bromann, line 322: This test si not needed and can be incorporated into `should be no result in log if there is value in response`

packages/webdriver/tests/command.test.ts
wyhmichael: reverted

packages/webdriver/src/command.ts
wyhmichael: kk, I am not sure what else the usage of WDIO_WORKER_ID would be, but using it anyways

wyhmichael:
@christian-bromann thanks for the feedback, please take another look at the updated changes

packages/webdriver/src/command.ts
```diff
@@ -101,7 +101,7 @@ export default function (
 
             this.emit('result', { method, endpoint, body, result })
 
-            if (command === 'deleteSession') {
+            if (command === 'deleteSession' && process.env.WDIO_WORKER_ID) {
```
christian-bromann, line 104:
```suggestion
            if (command === 'deleteSession' && !process.env.WDIO_WORKER_ID) {
```

christian-bromann (commented):
I think we want this the other way around. Only clear logger when we don't run the wdio testrunner,

packages/webdriver/src/command.ts
wyhmichael: Updated as suggested, and also updated unit tests

christian-bromann (approved):
Thanks for the fix 👍

christian-bromann:
It seems that one of the unit tests is not happy:

```
FAIL packages/webdriver/tests/command.test.ts
  ● Test suite failed to run
    Cannot spy the clearLogger property because it is not a function; undefined given instead
      160 |     const log = logger('webdriver')
      161 |     jest.spyOn(log, 'info').mockImplementation((string) => string)
    > 162 |     const clearLoggerSpy = jest.spyOn(logger, 'clearLogger')
          |                                 ^
      163 |
      164 |     function getRequestCallback (method: string, path: string, endpoint: Protocols.CommandEndpoint) {
      165 |         const commandFn = commandWrapper(method, path, endpoint)
      at ModuleMocker.spyOn (node_modules/jest-mock/build/index.js:783:15)
      at packages/webdriver/tests/command.test.ts:162:33
      at Object.<anonymous> (packages/webdriver/tests/command.test.ts:159:1)
```

wyhmichael:
hi @christian-bromann , I wonder if you can help with unit test failure, I have been struggle with it. What I experienced so far: 

if I do: 
1. go under `webdriverio\packages\webdriver`
2. run `tsc .\tests\command.test.ts`
3. run `npx jest -i .\tests\command.test.js` ===> this would work and pass

but I do: 
1. go under `webdriverio`
2. run `npm run test` ===> the unit test would fail for me, I have been trying but haven't got any clue on fixing this yet :(

christian-bromann (approved):
👍  thanks a lot

PR #8542, opened 2022-07-20
Make WebDriverIO compatible with pnpm without hoisting

dgp1130:
This makes `devtools` compatible with pnpm without relying on hoisting the `debug` package.

## Proposed changes

Currently you cannot import `webdriverio` in pnpm with [`hoist=false`](https://pnpm.io/npmrc#hoist) or it fails with the error:

```
Error: Cannot find module 'debug'
Require stack:
- /home/doug/Source/webdriverio-pnpm/node_modules/.pnpm/devtools@7.20.7/node_modules/devtools/build/utils.js
- /home/doug/Source/webdriverio-pnpm/node_modules/.pnpm/devtools@7.20.7/node_modules/devtools/build/devtoolsdriver.js
- /home/doug/Source/webdriverio-pnpm/node_modules/.pnpm/devtools@7.20.7/node_modules/devtools/build/index.js
- /home/doug/Source/webdriverio-pnpm/node_modules/.pnpm/webdriverio@7.20.7/node_modules/webdriverio/build/utils/index.js
- /home/doug/Source/webdriverio-pnpm/node_modules/.pnpm/webdriverio@7.20.7/node_modules/webdriverio/build/middlewares.js
- /home/doug/Source/webdriverio-pnpm/node_modules/.pnpm/webdriverio@7.20.7/node_modules/webdriverio/build/multiremote.js
- /home/doug/Source/webdriverio-pnpm/node_modules/.pnpm/webdriverio@7.20.7/node_modules/webdriverio/build/index.js
- /home/doug/Source/webdriverio-pnpm/index.js
```

The reason for this is because pnpm (without hoisting) limits package visibility so Node resolution [cannot import packages which aren't listed as hard dependencies](https://pnpm.io/symlinked-node-modules-structure). This is an issue for `devtools` since it requires _but does not depend on_ the `debug` package used by `puppeteer-core` in order to apply a monkey patch. This was previously done by resolving `puppeteer-core/node_modules/debug`, however pnpm does not follow this structure. See https://github.com/dgp1130/rules_prerender/issues/48#issuecomment-1188624568 for more context on the problem.

The solution proposed here is to have `devtools` require `puppeteer-core` normally (which works because it's explicitly listed as a dependency) and then require `debug` _from_ `puppeteer-core` using the [`paths` option for `require.resolve()`](https://nodejs.org/api/modules.html#requireresolverequest-options). In my local testing, this fixed the issue however I'm not sure of a great regression test to add without actually installing WebDriverIO in a pnpm workspace.

## Types of changes

## Checklist

    * I confirmed that this works in my local testing, but not sure of the best way to write a regression test.

<a href="https://gitpod.io/#https://github.com/webdriverio/webdriverio/pull/8542"></a>

### DIFF
--- a/packages/devtools/src/utils.ts
+++ b/packages/devtools/src/utils.ts
@@ -1,5 +1,3 @@
-import fs from 'fs'
-import path from 'path'
 import { execFileSync } from 'child_process'
 import logger from '@wdio/logger'
 import { commandCallStructure, isValidParameter, getArgumentType, canAccess } from '@wdio/utils'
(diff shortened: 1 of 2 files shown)


christian-bromann:
@dgp1130 thanks for the PR. I updated the target branch to `v7` as we have moved the `v8` development into the `main` branch. Can you check the failing unit test?

dgp1130:
@christian-bromann, thanks didn't realize there were tests for this. I had to basically rewrite them entirely as they were mocking `path.resolve()` and `fs.existsSync()` which didn't fit well with this model. I updated it to mock `require.resolve()` and `require()` directly (via dependency injection as Jest can't really mock `require()` AFAICT). I also included tests for pnpm and the associated failure cases.

It's hard to be 100% sure since this isn't _actually_ running the Node resolution algorithm, rather returning what I believe the Node resolution algorithm would return in such a case. Hopefully this is good enough for now, though a real end-to-end test which runs in a pnpm environment would probably give the most reliable answer.

I'll try to check tomorrow to verify that this modification still works when run in my reproduction, which can hopefully give an informal answer to that question.

christian-bromann (approved):
Thanks 👍 

I am not entirely sure if patching `debug` is still working this way since there also have been changes in Puppeteer in regards to that. I revisited and adapted the fix for the upcoming `v8` so it is not very crucial.

pnpm support is definitely desired here. I will see if I can find a tool that ensures we test for this.

PR #7214, opened 2021-07-31
webdriverio: implemented Actions class

erwinheitzman:
## Proposed changes

Implement Actions class. For more information see #7213

## Types of changes

[//]: # (What types of changes does your code introduce to WebdriverIO?)
[//]: # (_Put an `x` in the boxes that apply_)

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update

## Checklist

[//]: # (_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._)

- [x] I have read the [CONTRIBUTING](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md) doc
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)
- [x] I have added proper type definitions for new commands (if appropriate)

## Further comments

I would love an open discussion about this idea.

The class can currently be used like so:
```js
    await Actions
      .moveToElement(await $("#password"))
      .pause(1000)
      .click()
      .sendKeys("abracadabra") // sets password
      .sendKeys(await $("#username"), "invalid") // sets username
      .pause(2000)
      .contentClick() // right click at cursor position (which is still at center of the password element)
      .pause(1000)
      .perform(); // perform all actions in one go
```

Issue #7213, opened 2021-07-31
Implement something similar to action class in selenium for easier use of perform action


praveendvd:
**Is your feature request related to a problem? Please describe.**

Currently perform action requires knowledge on w3c action pragmatic implementation

**Describe the solution you'd like**

It would be great to wrap these actions in something like builder pattern like in selenium

`await browser.chainActions().moveTo(2,3).pause(1).moveTo(2,3).doubleClick().build()`

**Describe alternatives you've considered**

creating my own wrap like

```
moveTo: ()=> { 
actionArry.push(the action object
return this
)

build: async () => await browser.performActions(actionArr)
```

christian-bromann:
What API would you suggest?

praveendvd:
> What API would you suggest?

Like a chain action class that exposes all available actions as method , and finally a build step that triggers perform actions as shown in my example .

praveendvd:
> What API would you suggest?

In gitter we have lots questions on this , thought having something like this can make it easier

erwinheitzman:
I have taken a small example and modified it a little to match what it could look like for us.
This reminds me of how Protractor implemented the Actions as well and it felt very easy to use. It would also allow us to assign proper typings so that people don't make mistakes which makes me excited 😃 
```js
const actions = new Actions();
actions.moveToElement(element).click().perform();
```

erwinheitzman:
This would also mean that our `dragAndDrop` and `keys` command and probably more would become part of this instead as you would have a single entity for performing actions like `clickAndHold` `contextClick` `doubleClick` `dragAndDrop` `keyDown` `keyUp` etc

praveendvd:
> This would also mean that our `dragAndDrop` and `keys` command and probably more would become part of this instead as you would have a single entity for performing actions like `clickAndHold` `contextClick` `doubleClick` `dragAndDrop` `keyDown` `keyUp` etc

do we need to create action object? instead we can do like broser method , else it will be an entirely different implementation than other things . i think its better to extend the browser to have chainaction method

erwinheitzman:
I'm currently trying a POC as we speak and even though I felt this would be handy I now feel that we could have the following:
```js
browser
  .click() // would perform click at cursor location
  .contextClick() // would perform context-click at cursor location
```
with
```js
$('some selector')
  .click() // would perform click at the center of the element
  .contextClick() // would perform context-click at the center of the element
```

What do you think about that @praveendvd?

[...]

PR #8595, opened 2022-07-26
WebDriver Action Interface

christian-bromann:
## Left to Do

- [x] documentation
- [x] some e2e tests
- [x] allow chainable webdriverio elements as origin in pointer events
- [x] ~implement swipe and pinch commands~ __Won't Do:__ as support for actions is still pretty low

## Proposed changes

implements #7213

This patch implements an interface to call WebDriver actions. The interface looks as follows:

- an `action` command that allows to start a `key`, `pointer` or `wheel` action
```ts
browser.action('key')
browser.action('wheel', { id: 'someaction1' }) // second parameters allows custom id or parameters
browser.action('pointer', { arguments: { pointerType: 'pen' } }) // second parameters allows custom id or parameters
```
these commands return action classes that allow to chain certain actions. Every of these action classes have a `perform` property that allows to execute the action, e.g.:

```ts
// excerpt from the tests, none of the parameters are mandatory
await browser.action('key', { id: 'foobar' })
    .down('foo')
    .pause(100)
    .up('bar')
    .perform()

await browser.action('pointer')
    .down({
        button: 1,
        width: 2,
        height: 3,
        pressure: 4,
        tangentialPressure: 5,
        tiltX: 6,
        tiltY: 7,
        twist: 8,
        altitudeAngle: 9,
        azimuthAngle: 10,
    })
    .pause(100)
    .move({
        button: 0,
        width: 1,
        height: 2,
        pressure: 3,
        tangentialPressure: 4,
        tiltX: 5,
        tiltY: 6,
        twist: 7,
        altitudeAngle: 8,
        azimuthAngle: 9,
        x: 10,
        y: 11,
        duration: 12,
        origin: 'viewport'
    })
    .up({ button: 2 })
    .cancel()
    .perform()

await browser.action('wheel', { id: 'foobar' })
    .pause(100)
    .scroll({
        x: 0,
        y: 1,
        deltaX: 2,
        deltaY: 3,
        duration: 4
    })
    .perform()
```

Running `.perform(true)` will skip the `releaseActions` call.

- parallel action moves can be executed by `actions` command. It takes an array of different action interfaces as shown above, e.g.:

```ts
await browser.actions([
    browser.action('key').down().up()
    browser.action('pointer').down().up()
    browser.action('wheel').scroll()
])
```

## Types of changes

- [ ] ~Documentation update~ __Won't do:__ due to lack of support in remote devices

## Checklist

## Further comments

supersedes #7214

<a href="https://gitpod.io/#https://github.com/webdriverio/webdriverio/pull/8595"></a>


erwinheitzman:
Great work Christian though I would have appreciated a heads up before taking over the work that I started. 

Personally I find the API to be a bit overwhelming. I was hoping to bring our users an API that does not force them to think about things like `pointer`, `touch`, `wheel` but rather thinking of `keyboard` and `mouse` and we handle the input modes under the hood (think of setting `touch` automatically when mobile).

```js
    .move({
        button: 0,
        width: 1,
        height: 2,
        pressure: 3,
        tangentialPressure: 4,
        tiltX: 5,
        tiltY: 6,
        twist: 7,
        altitudeAngle: 8,
        azimuthAngle: 9,
        x: 10,
        y: 11,
        duration: 12,
        origin: 'viewport'
    })
  ```
  
Especially this example points out what I would like to try and avoid. If they want such finegrained control (which is very rare in my opinion), we already provide the performActions API for that.

I do like the where you are going with this setup, it makes me think of Playwright's API:
- keyboard: https://playwright.dev/docs/api/class-keyboard
- mouse: https://playwright.dev/docs/api/class-mouse

It's also a bit similar to Cypress's setup.

Regarding the parallel actions, do you have an example of a usecase for this?

**Comment by christian-bromann** [comment]
@erwinheitzman thanks for the feedback!

> Personally I find the API to be a bit overwhelming.

In what sense to you feel this API overwhelming? The example I put above might be a bit confusing here. All of them are optional and can be omitted, e.g. drag and drop can be written like this:

```ts
    return this.action('pointer')
        .move({ origin: await $('#drag')  })
        .down()
        .pause(10)
        .move({ origin: await $('#drop') })
        .up()
        .perform()
```

Which seems to me very minimal but very flexible at the same time. Note that these commands are more interfaces to the protocol and we still have commands like `await $('#drag').dragAndDrop($('#drop'))` to wrap certain common interactions into one single command. These I haven't yet implemented bug could think of commands like

- swipe
- pinch
- zoom
- what else?

> Regarding the parallel actions, do you have an example of a usecase for this?

An example would be a pinch zoom which has the following payload:

```json
 {
   "actions": [
     {
       "type": "pointer",
       "id": "finger1",
       "parameters": {"pointerType": "touch"},
       "actions": [
         {"type": "pointerMove", "duration": 0, "x": 100, "y": 100},
         {"type": "pointerDown", "button": 0},
         {"type": "pause", "duration": 500},
         {"type": "pointerMove", "duration": 1000, "origin": "pointer", "x": -50, "y": 0},
         {"type": "pointerUp", "button": 0}
       ]
     }, {
       "type": "pointer",
       "id": "finger2",
       "parameters": {"pointerType": "touch"},
       "actions": [
         {"type": "pointerMove", "duration": 0, "x": 100, "y": 100},
         {"type": "pointerDown", "button": 0},
         {"type": "pause", "duration": 500},
         {"type": "pointerMove", "duration": 1000, "origin": "pointer", "x": 50, "y": 0},
         {"type": "pointerUp", "button": 0}
       ]
     }
   ]
 }
```
and could be executed as follows:
```js
await browser.actions([
    browser.action('pointer', { arguments: { pointerType: 'touch' }})
        .move({ x: 100, y: 100 })
        .down()
        .pause(500)
        .move({ duration: 1000, x: -50, y: 0 })
        .up()
    ,
    browser.action('pointer', { arguments: { pointerType: 'touch' }})
        .move({ x: 100, y: 100 })
        .down()
        .pause(500)
        .move({ duration: 1000, x: 50, y: 0 })
        .up()
])
```
Again, I can see implementing a nice command for this like `pinch` that has less arguments and runs that action.

pjcalvo:
Hi,

I was thinking that making the entire object the first parameter would make it easier to reuse steps. As an additional way to implement this:

something like:



`browser.action({type:’pointer’, params: ‘…’, id: ‘’someid}`


What do you think?

christian-bromann:
> What do you think?

I don't quite understand what you described, do you have some "imaginary" code examples?

erwinheitzman:
@christian-bromann that's a nice example, let me show an example of what I would like to use myself as a user:

```js
await browser.mouse.
        .move(await $('#drag')) // don't await here as the element needs to be fetched when the command gets performed?
        .down()
        .pause(10)
        .move(0, 100) move by x and y positions
        .up()
        .perform()
        
await mouse.
        .move(await $('#elem-to-click'))
        .click({ button: 2 })
        .perform()
```

The complexity that I was referring to are the options coming from performActions.
While it's nice to have support for some really advanced options, I feel like we should either give our users an example of when such options could be used in our documentation.
One of our users suggested a `Recipies` page on our website for things like this which might be perfect for this.

For example, I have never seen many of the  options until I just searched for them specifically and found them here: https://w3c.github.io/webdriver/#pointer-actions.
I don't think many people know what all of these options do (I know I don't) and feel that if we decide to allow all of this, we should definitely provide good examples for their usecases.

Let me know what you think of the two points :)

christian-bromann:
> The complexity that I was referring to are the options coming from performActions.

Gotcha, I assume it has to do with:

```ts
await browser.mouse.
```

vs.

```ts
browser.action('pointer', { arguments: { pointerType: 'touch' }})
```

Please note that my example should showed what is possible. In reality 99% of the time user go with default values, so you could write it like this too:

```ts
browser.action('pointer')
```

I am hesitant towards the idea to just have it as `browser.mouse` given that you can also refer to a tablet pen or touch events when e.g. used for mobile applications. Appium has deprecated their `touchAction` command in favor of the actions interface so staying a bit general with `pointer` seems like a better wording to me.

> I don't think many people know what all of these options do (I know I don't) and feel that if we decide to allow all of this, we should definitely provide good examples for their usecases.

Sounds good to me. I am working on some documentation around this. Maybe have an example that draws a WebdriverIO logo in our [guinea pig](http://guinea-pig.webdriver.io/draw.html#) page.

christian-bromann:
I have to skip on pinch as it is poorly supported by browser, e.g. Firefox doesn't have `touch` as `pointerType` yet, creating a funky back and forth on https://hammerjs.github.io/:
play

Chrome is not much better:
play

christian-bromann:
@erwinheitzman the last commit added support for easier usage of pointer methods, e.g. `.down('left')` vs. `.down({ button: 0 })` (same for `up`) as well as `.move({ x: 1, y: 2 })` vs. `.move(1, 2)`. If there is anything else you see can be simplified, let me know.

Before merging I will add a documentation section for actions (similar what we have for the `mock` object) to provide more information on individual properties of every action type.

christian-bromann:
> Appium has specific gesture helpers per platform, ran by `execute`: [iOS](https://github.com/appium/appium-xcuitest-driver#mobile-pinch) and [Android](https://github.com/appium/appium-uiautomator2-driver#mobile-gesture-commands)
> Maybe that helps.

@nextlevelbeard thanks for pointing to these docs. Would be very nice to have a common interface for mobile and browser but given that browsers are not there yet it won't gonna implement these commands for them yet.

christian-bromann:
@erwinheitzman I added some docs, the PR is ready to be shipped, mind reviewing this?

erwinheitzman:
Regarding the proxy error, I think you are running into the same error that I ran into when I started implementing this. Somewhere in WebdriverIO we are manually throwing a "not yet implemented" error, is this the error you are referring to Christian?

packages/devtools/src/commands/performActions.ts
```diff
@@ -180,6 +183,27 @@ export default async function performActions(
...
+                    const location = await getElementRect.call(this, { elementId: singleAction.origin[ELEMENT_KEY] })
+                    await page.mouse.wheel({
+                        deltaX: location.x,
+                        deltaY: location.y - windowSize.height
+                    })
+                } else if (singleAction.x || singleAction.y) {
+                    await page.mouse.move(singleAction.x || 0, singleAction.y || 0)
+                }
+
+                await page.mouse.wheel({
```
erwinheitzman, line 199: This always get's executed, it feels like we should only scroll once when an origin is assigned. Is this correct?

christian-bromann:
> is this the error you are referring to Christian?

Nope, this was Sauce / Appium specific. I am not sure what error you experienced but it could be related to running actions using the devtools package and Puppeteer. I did extend the commands to support the new wheel action type though.

packages/webdriverio/src/commands/browser/action.ts
```diff
@@ -0,0 +1,146 @@
...
+ *
+ * ```ts
+ * browser.action('key')
+ * ```
+ *
+ * It returns a `KeyAction` object that supports the following actions:
+ *
+ * - `down(value: string)`: generates a key down action
+ * - `up(value: string)`: generates a key up action
+ * - `pause(ms: number)`: indicate that an input source does nothing during a particular tick
```
erwinheitzman, line 29: do we want to add a `press` and/or `type` action?

- press would do both a key down and key up action in one
- type would do the same as press but for multiple keys in one

packages/webdriverio/src/commands/browser/action.ts
```diff
@@ -0,0 +1,146 @@
...
+ *
+ * ### Pointer input source
+ *
+ * A pointer input source is an input source that is associated with a pointer-type input device. The type can be
+ * specified when invoking the `action` command, e.g.:
+ *
+ * ```ts
+ * browser.action('key'. {
+ *     parameters: { pointerType: 'mouse' } // "mouse" is default value, also possible: "pen" or "touch"
```
erwinheitzman, line 38: Did you mean to add a mouse action here? It feels like we are mixing keyboard and mouse and actions with performActions

packages/webdriverio/src/commands/browser/action.ts
```diff
@@ -0,0 +1,146 @@
...
+ * })
+ * ```
+ *
+ * It returns a `PointerAction` object that supports the following actions:
+ *
+ * - `down (button: 'left' | 'middle' | 'right')`: creates an action to press a single key
```
erwinheitzman, line 44: can the button be left out with `left` being the default?

packages/devtools/src/commands/performActions.ts
christian-bromann: Good catch, I updated the implementation as it had an error anyway. I hope the updated version is more clear.

erwinheitzman:
> > is this the error you are referring to Christian?
> 
> Nope, this was Sauce / Appium specific. I am not sure what error you experienced but it could be related to running actions using the devtools package and Puppeteer. I did extend the commands to support the new wheel action type though.

I think you are right about this, good call!

packages/webdriverio/src/commands/browser/action.ts
christian-bromann: Yes, the default values are here: https://github.com/webdriverio/webdriverio/blob/83472c73148134924cbbbc844b91b136faf26433/packages/webdriverio/src/utils/actions/pointer.ts#L20-L41

erwinheitzman:
> @erwinheitzman the last commit added support for easier usage of pointer methods, e.g. `.down('left')` vs. `.down({ button: 0 })` (same for `up`) as well as `.move({ x: 1, y: 2 })` vs. `.move(1, 2)`. If there is anything else you see can be simplified, let me know.
> 
> Before merging I will add a documentation section for actions (similar what we have for the `mock` object) to provide more information on individual properties of every action type.

if wheel and mobile support is lacking we should clearly mention this (maybe somewhere on top) so that are users are well aware of this limitation. can you add this (I haven't seen it if you have already)?

packages/webdriverio/src/commands/browser/action.ts
christian-bromann: `press` doesn't really suggest that a key up action is emitted. I can see adding `type` for that particular action though (e.g. emitting a key down and up event with the same button/params).

> do the same as press but for multiple keys in one

Can you share a specific use case here?

christian-bromann:
> if wheel and mobile support is lacking we should clearly mention this (maybe somewhere on top) so that are users are well aware of this limitation. can you add this (I haven't seen it if you have already)?

I was thinking about this but then decided not to add this explicitly because so far the error messages from the driver (either e.g. Safaridriver or Appium) did provide a good suggestion that the action isn't supported. Furthermore if we would explicitly throw an error message saying this action is not supported, we would need to check new Appium and driver versions if support was shipped.

However I think it is a good idea to add an informational banner saying that some browser or mobile devices might not support this.

christian-bromann:
Added:

> Support for this command and specific actions can differ based on the environment. Progress on the development
can be followed on [wpt.fyi](https://wpt.fyi/results/webdriver/tests/perform_actions?label=experimental&label=master&aligned). For mobile you might want to use Appium specific gesture commands on [iOS](https://github.com/appium/appium-xcuitest-driver#mobile-pinch)
and [Android](https://github.com/appium/appium-uiautomator2-driver#mobile-gesture-commands).

PR #8597, opened 2022-07-28
Enforce file extensions in imports

christian-bromann:
## Proposed changes

Let's enforce this as we can't import directories anymore using ESM. This patch fixes all imports (mostly within tests).

<a href="https://gitpod.io/#https://github.com/webdriverio/webdriverio/pull/8597"></a>


erwinheitzman:
Maybe a silly question but can't we now update to native import/export instead?

christian-bromann:
> Maybe a silly question but can't we now update to native import/export instead?

Yes, since we moved to ESM. ESM imports require however a file extension everywhere which is why I added this eslint rule. For all vitest tests it appears that this wasn't needed, either because TypeScript or Vitest has some smart module resolution and understands to look into the right place.

erwinheitzman:
If you can resolve the conflicts then I'll review again :)

christian-bromann:
@erwinheitzman rebased. Not sure if this PR needs a big review, given it is basically just adding file extensions to the imports.

christian-bromann:
merged into `main`

Issue #7354, opened 2021-08-28
Import globals instead of relying on globals


fregante:
**Is your feature request related to a problem? Please describe.**

Globals bring a variety of issues that need to be addressed separately and could be avoided by just not using globals now that we have CJS and ESM.

Namely this is ESLint and TypeScript issues (I'm not looking for a solution to each)

**Describe the solution you'd like**

Like Jest, offer a [`injectGlobals: false` option](https://jestjs.io/docs/configuration#injectglobals-boolean) and document a file to import:
```js
import {expect, jest, test} from '@jest/globals';
```

I tried using `const {default: browser} = require("webdriverio/build/commands/browser")` but the functions refer to `this` so there must be something that I'm missing.

**Additional context**

No setup, type work out of the box thanks to my `require` (if only my import was right)

<img width="703" alt="Screen Shot" src="https://user-images.githubusercontent.com/1402241/131212360-d52be330-b0fe-4a0e-a048-d2cc58bebe82.png">

christian-bromann:
> Globals bring a variety of issues

Can you elaborate what kind of issues you are referring to?

I definitely believe moving away from globals is a cleaner way of handling the browser variable. It would require some refactoring across the project to get this working though. So far I haven't had anyone with severe issues using the global variable.

fregante:
> > Globals bring a variety of issues
> 
> Can you elaborate what kind of issues you are referring to?

ESLint `no-undef` reports, I don’t even lack of completions, TypeScripts compile errors. Each of these issues have its own solution/config, but an `import` would work out of the box

christian-bromann:
> ESLint `no-undef` reports, I don’t even lack of completions

For that we offer an [Eslint plugin](https://www.npmjs.com/package/eslint-plugin-wdio).

I guess we can optionally offer to import these. Any other opinions?

erwinheitzman:
eslint allows you to define globals for this very reason, add `browser` as a global and you're ready to go

fregante:
I think you have not read my post.

1. I'm not looking for a solution, I know _more configuration_ fixes each of the 3 issues
2. I don't think it's reasonable for a project to require additional configuration just to get the basics working
3. `no-undef` only fixes one of the 3 issues

erwinheitzman:
I read it but a bit too quick as I am doing this during my coffee break ^^;

[...]

PR #8631, opened 2022-08-02
Introduce `@wdio/globals` to allow implicit imports of WebdriverIO global variables

christian-bromann:
## Proposed changes

see #7354

This patch adds a new option called `injectGlobals` (by default `true`) that allows user to tell WebdriverIO __not__ inject global variables (`browser`, `driver`, `$` and `$$`) into the global scope. A new package called `@wdio/globals` handles setting these global variables now and can be used to import them if `injectGlobals` is set to `false`.

#### Breaking Change

This patch has an impact on TypeScript users who now have to setup global support by setting `@wdio/globals` in their `tsconfig.json`, e.g.:

```patch
{
    "compilerOptions": {
-        "types": ["node", "webdriverio/async"]
+        "types": ["node", "@wdio/globals/types"]
    }
}
```

The types `webdriverio/async` or `webdriverio/sync` were removed as we don't differentiate anymore between sync and async. Documentation was added to the API section to reference all global variables:

Screenshot 2022-08-02 at 12 27 07

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

n/a

<a href="https://gitpod.io/#https://github.com/webdriverio/webdriverio/pull/8631"></a>


erwinheitzman:
Can we also fix the framework globals? Currently if a team is using a unit test framework using describe/it/expect (Jest, Jasmine, Mocha), it will conflict thus one is not able to place the files in the same project.

christian-bromann:
> it will conflict thus one is not able to place the files in the same project.

As mentioned in the doc changes, WebdriverIO has no influence on frameworks like Jasmine or Mocha how they set globals. It is impossible for `@wdio/globals` to detect which framework is being used and export the correct types. I've experienced the issue you've mentioned and know it can be resolved setting e.g. an additional tsconfig for WebdriverIO tests. So I would suggest adding more docs around this topic rather than trying to solve this within `@wdio/globals`.

erwinheitzman:
Gotcha, I am looking at some workarounds and it seems like it is indeed possible, Thanks!

packages/wdio-globals/.npmignore
```diff
@@ -0,0 +1,4 @@
+src
```
erwinheitzman, line 1: Instead of maintaining the .npmignore files we can instead define only what we want to include in our publication. This can be done by populating the `files` array in the package.json which is easier to maintain.

Reference: https://docs.npmjs.com/cli/v8/using-npm/developers#keeping-files-out-of-your-package

packages/wdio-globals/src/index.ts
```diff
@@ -0,0 +1,82 @@
+type SupportedGlobals = 'browser' | 'driver' | 'multiremotebrowser' | '$' | '$$'
```
erwinheitzman, line 1: I saw `define` and `it` mentioned in earlier examples, do they need to be added here?

packages/wdio-globals/README.md
```diff
@@ -0,0 +1,22 @@
+WDIO Globals
+============
+
+> A helper utility for importing global variables directly
+
+This package can be used to import global variables explicitly, e.g.
+
+```ts
+import { browser, describe, it } from '@wdio/globals'
```
erwinheitzman, line 9: If we can import the `describe` and `it`, can we also import `expect`, `beforeEach` etc?

packages/wdio-globals/src/index.ts
```diff
@@ -0,0 +1,82 @@
...
+declare namespace WebdriverIOAsync {
+    interface Browser {}
+    interface Element {}
+    interface MultiRemoteBrowser {}
+}
+
+declare namespace WebdriverIO {
+    interface Browser extends BrowserSync, WebdriverIOAsync.Browser { }
+    interface Element extends ElementSync, WebdriverIOAsync.Element { }
+    // @ts-expect-error
```
erwinheitzman, line 20: why is this expect-error needed here?

packages/wdio-globals/src/index.ts
```diff
@@ -0,0 +1,82 @@
...
+                throw new Error(GLOBALS_ERROR_MESSAGE)
+            }
+
+            return globals.get(key)[prop]
+        }
+    }
+}
+
+export const browser = new Proxy(
+    class Browser {} as any as WebdriverIO.Browser,
```
erwinheitzman, line 46: I think we can do the following instead
```suggestion
    class Browser extends WebdriverIO.Browser {},
```
This also goes for the similar instances below.

packages/wdio-globals/src/index.ts
```diff
@@ -0,0 +1,82 @@
...
+    proxyHandler('browser')
+)
+export const driver = new Proxy(
+    class Browser {} as any as WebdriverIO.Browser,
```
erwinheitzman, line 50: I think we can do the following instead
```suggestion
    class Browser extends WebdriverIO.MultiRemoteBrowser {},
```

tests/mocha/test.ts
```diff
@@ -10,8 +10,15 @@ describe('Mocha smoke test', () => {
         }
     })
 
-    it('should return sync value', () => {
-        expect(browser).toHaveTitle('Mock Page Title')
+    it('has globals set up', () => {
+        expect(typeof global.browser).not.toBe('undefined')
```
erwinheitzman, line 14:
```suggestion
        expect(typeof global.browser).not.toBeDefined()
```

tests/mocha/test.ts
```diff
@@ -10,8 +10,15 @@ describe('Mocha smoke test', () => {
...
+        expect(typeof global.driver).not.toBe('undefined')
```
erwinheitzman, line 15:
```suggestion
        expect(typeof global.driver).not.toBeDefined()
```

packages/wdio-globals/.npmignore
christian-bromann: I have no strong opinions how we want to handle this, let's create an issue for it though and have a separate PR for it, if we like to do it for all packages.

packages/wdio-globals/src/index.ts
christian-bromann: I assume you mean `describe` and `it`? Those are added by Mocha and Jasmine and we don't handle them in WebdriverIO (see our issue conversation).

packages/wdio-globals/README.md
christian-bromann: Good catch, `describe` and `it` don't work. It was initially an idea to include them but found out later that it is very difficult to implement this.

packages/wdio-globals/src/index.ts
christian-bromann: I don't think this works given that `WebdriverIO.Browser` is a type and not a class.

nextlevelbeard:
Hey @christian-bromann , not sure if relevant but Cucumber users also need to import a bunch of stuff to use the framework.

We made `@wdio/cucumber-framework` export these some time ago: https://github.com/webdriverio/webdriverio/blob/9a2680078b629b8d2857b6cbd275736c3847fce9/packages/wdio-cucumber-framework/src/index.ts#L395-L413

So currently, user can either import their own version, if their project has a `@cucumber/cucumber` dependency:
```ts
import { Given, When, Then } from '@cucumber/cucumber'
```
or they can use what's exported by `@wdio/cucumber-framework` which might not be up to date but also useful and makes `@cucumber/cucumber` dependecy not required:
```ts
import { Given, When, Then } from '@wdio/cucumber/framework'
```

So, could `injectGlobals` being true also inject everything from `@wdio/cucumber-framework`?
Would this be a good practice?

christian-bromann:
> Would this be a good practice?

I would not touch the behavior of how test frameworks like Cucumber deal with globals. If Cucumber wants to import `Given`, `When`, `Then` WebdriverIO should not change that and people should do this no matter of the WebdriverIO configuration. Generally I started to prefer importing globals implicitly after working with `vitest`. I think there is no best practice, more a preference question.

packages/wdio-globals/src/index.ts
christian-bromann: It is because I wasn't able to figure out how to properly type the Mutliremote instance to allow:
```ts
browser.click() // click for both instances at the same time
```
and
```ts
browser.browserName.click() // click for single browser instance
```

christian-bromann:
To get this finally working I had to do a little change. Rather than putting `@wdio/globals` directly into `types` of `tsconfig.json` it has to be `@wdio/globals/types` because to make it work

- I have to have an ambient file(`@wdio/globals/types`) that defines global WebdriverIO types
- and a module file `/build/index.d.ts` that allows to import these as objects

There was no way to have them within a single file. It could also be `webdriverio/types` but I felt it fits better within `@wdio/globals`.

Feedback appreciated.

erwinheitzman:
@christian-bromann would it make sense to instead use `@wdio/types/globals`? that way we can put all our types under the same structure like so:

`@wdio/types/globals`
`@wdio/types/foo`
`@wdio/types/bar`

christian-bromann:
> that way we can put all our types under the same structure like so:

What do you mean by that? What other structures to we have? I think the problem with this idea is that `webdriverio` has `@wdio/types` as dependency which means we can't have `@wdio/types` also depend on `webdriverio` (which is needed because the ambient file imports the webdriverio types).

erwinheitzman:
What I means it, can't we make this part of `@wdio/types` so that we can import from `@wdio/types/globals`?

christian-bromann:
> or maybe from `@wdio/types` directly?

That would not work because we have to export a module to be able to import the types in other packages. Afaik in order to make a `WebdriverIO` global namespace available it needs to be defined within an ambient typescript file. Those don't have imports and exports.

I don't think it makes a bit difference to use `@wdio/types/globals` or `@wdio/globals/types`, the problem with the first is that it would introduce a cyclic import issue which causes side effects and might not even work in the first place.

erwinheitzman:
I don't have a strong opinion regarding `webdriverio/types` vs `@wdio/globals` so I think we can merge this unless you want to wait for more feedback :)

PR #8657, opened 2022-08-08
reloadSession with DevTools breaks switchWindow command

olexme:
## Proposed changes

This is a fix for reloadSession method, which was braking possibility to switchWindow for new windows opened after session was reloaded.
The problem was debugged/diagnosed and it turned out that Puppeteer browser in devToolsDriver stops listen on
targetcreated and targetdestroyed events after the session is reloaded. Proposed fix initialises the same listeners in devToolsDriver constructor as well as after session is reloaded and new Browser object is created. It unloads also listeners old handlers to prevent memory leaks and undesired behaviour.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

<a href="https://gitpod.io/#https://github.com/webdriverio/webdriverio/pull/8657"></a>

### DIFF
--- a/packages/devtools/src/devtoolsdriver.ts
+++ b/packages/devtools/src/devtoolsdriver.ts
@@ -100,7 +100,7 @@ export default class DevToolsDriver {
         this.activeListeners.push({ emitter, eventName, boundHandler })
     }
 
-    private offActiverListeners() {
+    private offAllListeners() {
         this.activeListeners.forEach(({ emitter, eventName, boundHandler }) => {
             emitter.off(eventName, boundHandler)
         })
@@ -109,12 +109,12 @@ export default class DevToolsDriver {
 
     /**
      * Inits browser listeners and sets initial handlers for given pages.
-     * Function is intended to be used while reloading DevTools session.
-     * @param browser pupeteer Browser
-     * @param pages pupeteer page array
+     * Function is also intended to be used while reloading DevTools session.
+     * @param browser Puppeteer Browser
+     * @param pages Puppeteer page array
      */
     initBrowser(browser: Browser, pages: Page[]) {
-        this.offActiverListeners()
+        this.offAllListeners()
         this.elementStore = new ElementStore()
         this.windows = new Map()
         this.activeDialog = undefined
--- a/packages/devtools/src/devtoolsdriver.ts
+++ b/packages/devtools/src/devtoolsdriver.ts
@@ -15,6 +15,7 @@ import * as commands from './commands/index.js'
 import ElementStore from './elementstore.js'
 import { validate, sanitizeError } from './utils.js'
 import { DEFAULT_IMPLICIT_TIMEOUT, DEFAULT_PAGELOAD_TIMEOUT, DEFAULT_SCRIPT_TIMEOUT } from './constants.js'
+import { EventEmitter } from 'puppeteer-core/lib/cjs/puppeteer/common/EventEmitter.js'
 
 const __dirname = path.dirname(url.fileURLToPath(import.meta.url))
 const log = logger('devtools')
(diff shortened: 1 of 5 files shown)


packages/devtools/src/devtoolsdriver.ts
```diff
@@ -108,6 +94,51 @@ export default class DevToolsDriver {
         log.trace(`Switching to window handle with id ${pageIds[0]}`)
     }
 
+    private addListener(emitter: EventEmitter, eventName: string, handler: any) {
+        const boundHandler = handler.bind(this)
+        emitter.on(eventName, boundHandler)
+        this.activeListeners.push({ emitter, eventName, boundHandler })
+    }
+
+    private offAllListeners() {
```
erwinheitzman, line 103: I'd prefer something more descriptive of what the purpose of this method is.
```suggestion
    private cleanupListeners() {
```

packages/devtools/src/devtoolsdriver.ts
```diff
@@ -108,6 +94,51 @@ export default class DevToolsDriver {
...
+        this.activeListeners.forEach(({ emitter, eventName, boundHandler }) => {
+            emitter.off(eventName, boundHandler)
```
erwinheitzman, line 105: do we need to run this when the listeners are removed anyway?

erwinheitzman (commented):
Thanks for picking this up!

I left some feedback for you to look at though I don't know enough about this so say anything about the functioning of the code without testing this locally (which I currently can't).

So if anyone can take a look at that, I'd appreciate that :)

packages/devtools/src/devtoolsdriver.ts
olexme: done

packages/devtools/src/devtoolsdriver.ts
olexme: We should unplug listeners from emitters in case a listener instance isn't destroyed together with emitter. Here pages and browsers emitters can be created and destroyed (new windows, session reload) but DevToolsDriver with its listeners remains all session. Otherwise the emitters won't be garbage-collected and we could have also unpredicted behavior (e.g. any event from hanging window), however the latter is less probable here.

erwinheitzman: I wonder if we could automatically do this on an event so that it always happens and not just when explicitly calling `cleanupListeners` as it would be more robust and better for maintenance. Do you happen to know if this is possible?

packages/devtools/src/devtoolsdriver.ts
```diff
@@ -30,11 +31,9 @@ export default class DevToolsDriver {
     currentFrame?: Page
     currentWindowHandle?: string
     currentFrameUrl?: string
-
+    activeListeners: {emitter: EventEmitter, eventName: string, boundHandler: any}[] = []
```
christian-bromann, line 34: Can we define this as `ActiveListener` type in `types.ts` for better readability?

christian-bromann (commented):
One comment, otherwise LGTM 👍 

Also, mind raising the same PR for the `v7` branch so we can release this before v8?

packages/devtools/src/devtoolsdriver.ts
olexme: done

packages/devtools/src/devtoolsdriver.ts
olexme: It could be probably possible to do it attaching some logic to `targetdestroyed` event handler, but it would require some more investigation/tests - probably out of scope of this PR.

olexme:
> One comment, otherwise LGTM 👍
> 
> Also, mind raising the same PR for the `v7` branch so we can release this before v8?

PR for `v7` done [here](https://github.com/webdriverio/webdriverio/pull/8687). Would be great to have a patch sooner than later - thanks.

Issue #8628, opened 2022-08-01
[🐛 Bug]: element.click inconsistent behaviour


DirkoOdendaal:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

v7.20.8

### Node.js Version

v16.13.0

### Mode

Standalone Mode

### Which capabilities are you using?

```typescript
{
      maxInstances: 5,
      browserName: 'chrome',
      acceptInsecureCerts: true
}
```

### What happened?

When you have a page with a button that is not in the viewport, calling element.click() does not scroll the element into view.
But, if you add an empty options object, then this does work, something like element.click({}).

### What is your expected behavior?

From the documentation I understand that with no options passed it should scroll into view.
If that's not the case, then I'm wondering why in some cases the code does scroll the button into view and in other cases not?

https://webdriver.io/docs/api/element/click
This issues a WebDriver click command for the selected element , which generally scrolls to and then clicks the selected element when no options are passed.

### How to reproduce the bug.

I created a repo where this is reproduced:
https://github.com/DirkoOdendaal/ionic-5-angular-webdriverio

All you need to do is run npm i to install dependencies.
Start the app with npm start and then run e2e tests with npm run wdio.

There is a spec file where 1 test passes with element.click({}) (e2e/wdio/src/specs/tab1-success.spec.ts)
and one that fails with element.click() (e2e/wdio/src/specs/tab1-failure.spec.ts).

Exact same test, just with the click command updated.

### Relevant log output

```typescript
Passing:

2022-08-01T14:15:51.092Z INFO @wdio/local-runner: Run worker command: run
2022-08-01T14:15:51.628Z INFO webdriver: Initiate new session using the WebDriver protocol
2022-08-01T14:15:51.702Z INFO webdriver: [POST] http://localhost:9515/session
2022-08-01T14:15:51.702Z INFO webdriver: DATA {
  capabilities: {
    alwaysMatch: { browserName: 'chrome', acceptInsecureCerts: true },
    firstMatch: [ {} ]
  },
  desiredCapabilities: { browserName: 'chrome', acceptInsecureCerts: true }
}
2022-08-01T14:15:53.167Z INFO webdriver: COMMAND navigateTo("http://localhost:4200/tabs/tab1")
[...]

PR #8658, opened 2022-08-08
Assign empty object if options are undefined

DirkoOdendaal:
## Proposed changes

If the click command is called with no options, the click action on the element is executed, but when passing empty options or just a button left option, then the code first scrolls the button into view and then clicking.
By just setting the options to an empty object if it is not passed in, the code follows the same pattern for all options.
Resolves bug: [8628](https://github.com/webdriverio/webdriverio/issues/8628)

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

By setting options to an empty object, it can follow the current code flow.
I also looked into setting a scroll into view just before the elementClick command, but that introduces another place where you scroll and click, adding more complexity to the code. So I rather chose a simple fix that follows the existing code paths.

<a href="https://gitpod.io/#https://github.com/webdriverio/webdriverio/pull/8658"></a>

### DIFF
--- a/packages/webdriverio/src/commands/element/click.ts
+++ b/packages/webdriverio/src/commands/element/click.ts
@@ -147,7 +147,7 @@ export default async function click(
             () => null,
             (err) => log.warn(`Failed to call "releaseAction" command due to: ${err.message}, ignoring!`))
 
-        return
+        return null
     }
 
     const { width, height } = await this.getElementSize(this.elementId)
--- a/packages/webdriverio/src/commands/element/click.ts
+++ b/packages/webdriverio/src/commands/element/click.ts
@@ -88,7 +88,7 @@ export default async function click(
     options?: ClickOptions
 ) {
     if (typeof options === 'undefined') {
-        return this.elementClick(this.elementId)
+        options = {}
     }
 
     if (typeof options !== 'object' || Array.isArray(options)) {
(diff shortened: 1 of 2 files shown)


packages/webdriverio/src/commands/element/click.ts
```diff
@@ -147,7 +147,7 @@ export default async function click(
             () => null,
             (err) => log.warn(`Failed to call "releaseAction" command due to: ${err.message}, ignoring!`))
 
-        return
+        return null
```
erwinheitzman, line 150: Unnessecary

packages/webdriverio/src/commands/element/click.ts
```diff
@@ -88,7 +88,7 @@ export default async function click(
     options?: ClickOptions
 ) {
     if (typeof options === 'undefined') {
-        return this.elementClick(this.elementId)
+        options = {}
```
erwinheitzman, line 91: This isn't the way to assign a default object, instead it should be:
```js
export default async function click(
    this: WebdriverIO.Element,
    options?: ClickOptions = {}
) {
```
After which this part can be removed.

However, doing this completely changes the behavior of the click command and I know that @christian-bromann wanted to have the elementClick in here specifically when no options are passed. So Christian, do we still want this?

PR #8672, opened 2022-08-08
Align click command consistency

DirkoOdendaal:
## Proposed changes

If the click command is called with no options, the click action on the element is executed, but when passing empty options or just a button left option, then the code first scrolls the button into view and then clicking.
By just setting the options to an empty object if it is not passed in, the code follows the same pattern for all options.
Resolves bug: [8628](https://github.com/webdriverio/webdriverio/issues/8628)

## Types of changes

- [ x] Bugfix (non-breaking change which fixes an issue)
- [ ] Documentation update

## Checklist

- [ x] I have read the [CONTRIBUTING](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md) doc

## Further comments

By setting options to an empty object, it can follow the current code flow.
I also looked into setting a scroll into view just before the elementClick command, but that introduces another place where you scroll and click, adding more complexity to the code. So I rather chose a simple fix that follows the existing code paths.

<a href="https://gitpod.io/#https://github.com/webdriverio/webdriverio/pull/8672"></a>

### DIFF
--- a/packages/webdriverio/src/commands/element/click.ts
+++ b/packages/webdriverio/src/commands/element/click.ts
@@ -86,7 +86,7 @@ export default async function click(
     options?: ClickOptions
 ) {
     if (typeof options === 'undefined') {
-        return this.elementClick(this.elementId)
+        options = {}
     }
 
     if (typeof options !== 'object' || Array.isArray(options)) {
--- a/packages/webdriverio/src/commands/element/click.ts
+++ b/packages/webdriverio/src/commands/element/click.ts
@@ -86,7 +86,8 @@ export default async function click(
     options?: ClickOptions
 ) {
     if (typeof options === 'undefined') {
-        options = {}
+        await this.scrollIntoView()
+        return this.elementClick(this.elementId)
     }
 
     if (typeof options !== 'object' || Array.isArray(options)) {
--- a/packages/webdriverio/tests/commands/element/click.test.ts
+++ b/packages/webdriverio/tests/commands/element/click.test.ts
@@ -21,7 +21,7 @@ describe('click test', () => {
         await elem.click()
 
         expect(got.mock.calls[2][0].pathname)
-            .toBe('/session/foobar-123/element/some-elem-123/click')
+            .toBe('/session/foobar-123/actions')
     })
 
     it('should allow to left click on an element by passing a button type', async () => {
--- a/packages/webdriverio/tests/commands/element/click.test.ts
+++ b/packages/webdriverio/tests/commands/element/click.test.ts
@@ -21,7 +21,7 @@ describe('click test', () => {
         await elem.click()
 
         expect(got.mock.calls[2][0].pathname)
-            .toBe('/session/foobar-123/actions')
+            .toBe('/session/foobar-123/element/some-elem-123/click')
     })
 
     it('should allow to left click on an element by passing a button type', async () => {
--- a/packages/webdriverio/tests/middleware.test.ts
+++ b/packages/webdriverio/tests/middleware.test.ts
@@ -65,8 +65,8 @@ describe('middleware', () => {
         const subSubElem = await subElem.$('#subsubfoo')
         subSubElem.elementId = undefined as any
 
-        //Success returns undefined
-        expect(await subSubElem.click()).toEqual(undefined)
+        //Success returns a null
+        expect(await subSubElem.click()).toEqual(null)
         expect(vi.mocked(warn).mock.calls).toHaveLength(1)
         expect(vi.mocked(warn).mock.calls).toEqual([['Request encountered a stale element - terminating request']])
         // @ts-ignore mock feature
--- a/packages/webdriverio/tests/middleware.test.ts
+++ b/packages/webdriverio/tests/middleware.test.ts
@@ -65,8 +65,8 @@ describe('middleware', () => {
         const subSubElem = await subElem.$('#subsubfoo')
         subSubElem.elementId = undefined as any
 
-        //Success returns a null
-        expect(await subSubElem.click()).toEqual(null)
+        //Success returns undefined
+        expect(await subSubElem.click()).toEqual(undefined)
         expect(vi.mocked(warn).mock.calls).toHaveLength(1)
         expect(vi.mocked(warn).mock.calls).toEqual([['Request encountered a stale element - terminating request']])
         // @ts-ignore mock feature


packages/webdriverio/src/commands/element/click.ts
```diff
@@ -86,7 +86,7 @@ export default async function click(
     options?: ClickOptions
 ) {
     if (typeof options === 'undefined') {
-        return this.elementClick(this.elementId)
+        options = {}
```
christian-bromann, line 89: I don't think we can consider this as a solution. The original idea here is to use the standard WebDriver click if no special options are passed in. Errors such as `element not interactable` are legit issues and should be investigated. The reason why you don't get this when using the actions API is because these commands are not connected to an element and therefor those "intractability" checks aren't made.

christian-bromann (changes requested):

packages/webdriverio/src/commands/element/click.ts
DirkoOdendaal: Okay cool.
Let me work on another solution then that would cover all the scenarios.

christian-bromann: The one (`click({})`) might just not throw an error, can you verify if the click at least works? I wonder why the element is not intractable. As last option I could imagine to retry the click using the action command instead.

DirkoOdendaal: Okay, so what I did now was just to add the scrollIntoView before the click.
Reason is that the rest of the code does the scroll to.
If that's not the preferred way of doing it, then it's fine, but I wonder why the other command actually do it then.

christian-bromann: The WebDriver protocol already defines that for every click the browser should scroll the element into viewport. But it only scrolls it to the edge of the viewport and it might help to scroll it into the center of it.

Did you test the new behavior with your use case?

DirkoOdendaal: Yeah, works perfectly.

Can also raise it to v7 branch if you're happy with it.

christian-bromann: If you add a unit test that would be great. And yes, the same pr to `v7` helps to release it to the current `stable` version.

christian-bromann:
@erwinheitzman what do you think about this? My only concern is that a click now is 2 WebDriver commands rather than 1.

packages/webdriverio/src/commands/element/click.ts
BorisOsipov: >I did now was just to add the scrollIntoView before the click.

To be honest I am not happy about this. You add unnecessary scroll for hundreds of thousands users who don't need it.
For me `$().click()` scrolls s webdriver spec defines. Seems it will be better if you cover your corner case by custom click command in your own project or raise an issue in chrome\geckodriver repo.

BorisOsipov:
>I did now was just to add the scrollIntoView before the click.

To be honest I am not happy about this. You add unnecessary scroll for hundreds of thousands users who don't need it.
For me `$().click()` scrolls as webdriver spec defines. Seems it will be better if you cover your corner case by custom click command in your own project or raise an issue in chrome\geckodriver repo.

christian-bromann:
Thanks for the feedback @BorisOsipov .. I share your opinion.

erwinheitzman:
Rather looking at it from a big project where we ran into this issue as well (the same Ionic upgrade), we had to take this into account where I expected WebdriverIO to be able to handle this.

Why scroll to something only sometimes but sometimes not? I would say that consistency is more important then the performance cost as it's so little that I doubt anyone will notice it.

I was honestly surprised that it won't scroll because it's completely accessible and possible, it's just that because of some edgecase, webdriver won't.

erwinheitzman:
It's easy to create a wrapper so it's not like it's hard to work around it but I do feel like it would be nice of webdriver handled this by default :)

christian-bromann:
@erwinheitzman @DirkoOdendaal can you share a little bit more context? When you say you both have Ionic apps that means we are talking about mobile native apps, right? In this case I would suggest to only do the scrolling for mobile because WebDriver indeed scrolls the element into viewport.

erwinheitzman:
> @erwinheitzman @DirkoOdendaal can you share a little bit more context? When you say you both have Ionic apps that means we are talking about mobile native apps, right? In this case I would suggest to only do the scrolling for mobile because WebDriver indeed scrolls the element into viewport.

Hybrid app actually, Ionic started to use (open) shadow dom after which commands like `click` didn't scroll to the elements anymore. What I also noticed is that the `getText` command returned `''` instead of the actual textContent of the element. Since it was an app of another team I only assisted in explaining what I thought was the reason as to why the click no longer acted the same after the change.

Strange thing is that they could still access the elements like they weren't inside shadow dom. In other words they still used $ and $$ which work fine when you first scroll to the element and then execute the `click` or `getText` command.

I might be able to retrieve some additional information at some point but I do think this is all the info I have to share right now.

DirkoOdendaal:
> > @erwinheitzman @DirkoOdendaal can you share a little bit more context? When you say you both have Ionic apps that means we are talking about mobile native apps, right? In this case I would suggest to only do the scrolling for mobile because WebDriver indeed scrolls the element into viewport.
> 
> Hybrid app actually, Ionic started to use (open) shadow dom and it's my assumption that this is the reason why commands like `click` didn't scroll to the elements anymore. What I also noticed is that the `getText` command returned `''` instead of the actual textContent of the element. Since it was an app of another team I only assisted in explaining what I thought was the reason as to why the click no longer acted the same after the change. The update we did was from Ionic v5 to v6.
> 
> Strange thing is that they could still access the elements like they weren't inside shadow dom. In other words they still used $ and $$ which work fine when you first scroll to the element and then execute the `click` or `getText` command.
> 
> I might be able to retrieve some additional information at some point but I do think this is all the info I have to share right now.
> 
> Examples:
> 
> ```js
> elem.click() // doesn't click
> 
> // became
> elem.scrollIntoView()
> elem.click() // does click
> ```
> 
> ```js
> elem.getText() // returns ''
> 
> // became
> elem.scrollIntoView()
> elem.getText() // returns 'foo'
> ```
> 
> Since this is my assuption, another suspicion I have is that it might be a virtual dom that might be the reason.

So what Erwin explains here is exactly how we started.
I thought it might just be Ionic 6, but in ht eexample project I supplied, it used Ionic 5. Which is where I got confused.

If the correct thing to do is to try and raise a bug with webdriver then that's 100% fine.

I'll run that project on the other browsers too and see if it also happens.

erwinheitzman:
@DirkoOdendaal I think it only appeared in v6 for us as we aren't using many ionic components, it might be that the components that we use where affected in v6 as two components where updated in that version. I can't say for sure though as it happened outside my team inside a large organization so I was contacted to help out the teams to provide a solution and provide an answer of why their tests where suddenly failing but I have little knowledge about what components we use and such.

christian-bromann:
@DirkoOdendaal I'ld suggest we only scroll into view if we get an `element not intractable` error as part of a retry. I also raised https://github.com/w3c/webdriver/pull/1680 to change the current behavior of browser scrolling the element just to the end of the viewport rather than to the center.

What do you think?

wswebcreation:
@christian-bromann 

I've talked with @DirkoOdendaal about this last week, but based on the W3C specs this would be an external bug in ChromeDriver because, according to that spec, it should do [this](https://www.w3.org/TR/webdriver/#element-click)

> The [Element Click](https://www.w3.org/TR/webdriver/#dfn-element-click) [command](https://www.w3.org/TR/webdriver/#dfn-commands) [scrolls into view](https://www.w3.org/TR/webdriver/#dfn-scrolls-into-view) the [element](https://www.w3.org/TR/webdriver/#dfn-elements) if it is not already [pointer-interactable](https://www.w3.org/TR/webdriver/#dfn-pointer-interactable), and clicks its [in-view center point](https://www.w3.org/TR/webdriver/#dfn-center-point).

Solving it in WDIO would be good, but then we also need to update our docs on this. Reason for mentioning this is that WDIO also has a few commands from the W3C specs that have additional actions in it which are not clearly described in the docs.

What do you think?

christian-bromann:
@wswebcreation @DirkoOdendaal 

> Solving it in WDIO would be good

How?

> has a few commands from the W3C specs that have additional actions in it which are not clearly described in the docs.

Which ones? WebdriverIO should clearly communicate protocol commands vs. framework commands which come with additional features and I am all in for making sure these are clearly documented.

christian-bromann:
I will go ahead and close the PR given there doesn't seem to be any movement anymore. I am all ears for suggestions and improvements on this topic.

PR #8732, opened 2022-08-17
remove unnecessary await from example

monkpit:
A minor fix for the docs - browser.addLocatorStrategy has a return type of `void` so it doesn't need to be awaited.


christian-bromann (approved):
Thanks 👍

PR #8762, opened 2022-08-25
Allow fetching elements by a11y name

christian-bromann:
## Proposed changes

Aria labels are a common way to fetch elements. WebdriverIO already offers to fetch an element by its role but not by its name. For that this patch suggest to allow fetching an element by its label using e.g. `aria/Foobar`. WebdriverIO will translate this into an xPath that selects the element in the following order:

1. elements that have an aria-label property set
2. elements that have an aria-labelledby property and get their name by another element
3. elements that have a certain text containing the selector
4. elements that have a title containing the selector
5. `img` tags with a certain `alt` property
6. `input` elements with a `label`

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

I would like to write a little blog post about this to announce the feature.


nextlevelbeard:
Cool feature @christian-bromann 👍
Will we make any changes to the mobile accessibility operator `~` i.e. `$('~Foobar')`
Should `~` be deprecated in favor of `aria/` ?

packages/webdriverio/src/utils/findStrategy.ts
```diff
@@ -152,6 +156,27 @@ export const findStrategy = function (selector: SelectorStrategy, isW3C?: boolea
...
+    // .//*[@aria-labelledby = string(.//*[normalize-space() = "target 2"]/@id)]
+    case 'aria': {
+        const label = stringSelector.slice(ARIA_SELECTOR.length)
+        const conditions = [
+            // element has direct aria label
+            `.//*[@aria-label = "${label}"]`,
+            // aria label is recevied by other element with aria-labelledBy
+            `.//*[@aria-labelledby=(//*[normalize-space() = "${label}"]/@id)]`,
+            // aria label is received from element content
+            `.//*[normalize-space() = "${label}"]`,
```
erwinheitzman, line 168: This would retrieve any element containing this text in any of it's child elements.
I believe that instead we need `//*[text() = "${label}"]`

website/docs/Selectors.md
```diff
@@ -168,6 +168,72 @@ const parent = await paragraph.$('..')
 console.log(await parent.getTagName()) // outputs: "body"
 ```
 
+## Accessibility Name Selector
+
+For querying elements based on their accessiblity name. The accessible name is part of what is announced by a screen reader when that element receives focus. The value of the accessible name can be both visual content or hidden text alternatives.
```
erwinheitzman, line 173:
```suggestion
Query elements by their accessible name. The accessible name is what is announced by a screen reader when that element receives focus. The value of the accessible name can be both visual content or hidden text alternatives.
```

erwinheitzman (commented):
Nice going Christian

erwinheitzman:
Where does the code decide in what order to check if the conditions are truthy because the order in which it checks matters right?

erwinheitzman:
@nextlevelbeard thanks for the review, we have some doubts about what naming to take for this and your opinion is appreciated.

The actual name for it is Accessible name, we are thinking of `$('aria/Foobar')` or `$('a11y/Foobar')`, what do you think?

christian-bromann:
@nextlevelbeard @erwinheitzman thanks for the review!

> Will we make any changes to the mobile accessibility operator `~` i.e. `$('~Foobar')`

Mhm .. that would be an option even though the mobile accessibility operator is quite different from the web one. However I can see them being represented in the same way.

packages/webdriverio/src/utils/findStrategy.ts
christian-bromann: I can not really see a difference between `.//` and `//` given that we scan the whole document anyway. See http://xpather.com/1iZkrkxL

packages/webdriverio/src/utils/findStrategy.ts
```diff
@@ -152,6 +156,27 @@ export const findStrategy = function (selector: SelectorStrategy, isW3C?: boolea
...
+            // aria label is received by its title attribute
+            `.//*[@titel="${label}"]`,
+            // images with an alt tag
```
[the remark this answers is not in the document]
christian-bromann: The accessibility name of an image with an alt tag is the alt tag, that is also what the [spec](https://www.w3.org/TR/accname-1.1/#step2D) says:

> For example, in HTML, the img element's alt attribute defines a text alternative string, and the label element provides text for the referenced form element. In SVG2, the desc and title elements provide a description of their parent element.

Maybe I am understanding this incorrect but devtools seem to approve my assumption here:
Screenshot 2022-08-31 at 08 56 49
Screenshot 2022-08-31 at 08 57 07

christian-bromann:
> Where does the code decide in what order to check if the conditions are truthy because the order in which it checks matters right?

We have a set of [conditions](https://github.com/webdriverio/webdriverio/blob/6a43879fbee13d7e107f62c23daf9f49418e0610/packages/webdriverio/src/utils/findStrategy.ts#L161-L183) we check for and we will pick whatever element comes first. So if we find an element using the first xpath we will pick that one. I tried to be according to the spec.

christian-bromann:
Ok, regarding the selector strategy format, we have to decide between:

- `aria/Foobar`
- `a11y/Foobar`
- `~Foobar` (merge with the mobile accessibility operator)

Which should we pick? Thinking about it now I think it makes sense to merge them with the mobile version. WDYAT?

christian-bromann:
I added a small blog post on the new selector: https://github.com/webdriverio/webdriverio/blob/cb-aria-label/website/blog/2022-09-01-accessibility-selector.md

erwinheitzman:
I'm fine with `~Foobar` :)

erwinheitzman:
> > Where does the code decide in what order to check if the conditions are truthy because the order in which it checks matters right?
> 
> We have a set of [conditions](https://github.com/webdriverio/webdriverio/blob/6a43879fbee13d7e107f62c23daf9f49418e0610/packages/webdriverio/src/utils/findStrategy.ts#L161-L183) we check for and we will pick whatever element comes first. So if we find an element using the first xpath we will pick that one. I tried to be according to the spec.

Wouldn't the following be closer to the spec (moved title to be last)?
```
        const conditions = [
            // aria label is recevied by other element with aria-labelledBy
            // https://www.w3.org/TR/accname-1.1/#step2B
            `.//*[@aria-labelledby=(//*[normalize-space() = "${label}"]/@id)]`,
            // aria label is recevied by other element with aria-labelledBy
            // https://www.w3.org/TR/accname-1.1/#step2B
            `.//*[@aria-describedby=(//*[normalize-space() = "${label}"]/@id)]`,
            // element has direct aria label
            // https://www.w3.org/TR/accname-1.1/#step2C
            `.//*[@aria-label = "${label}"]`,
            // aria label is received by its title attribute
            // inputs with a label
            // https://www.w3.org/TR/accname-1.1/#step2D
            `.//input[@id = (//label[normalize-space() = "${label}"]/@for)]`,
            // images with an alt tag
            // https://www.w3.org/TR/accname-1.1/#step2D
            `.//img[@alt="${label}"]`,
            // aria label is received from element content
            // https://www.w3.org/TR/accname-1.1/#step2G
            `.//*[normalize-space() = "${label}"]`
            // https://www.w3.org/TR/accname-1.1/#step2D
            `.//*[@title="${label}"]`,
        ]
```

christian-bromann:
> Wouldn't the following be closer to the spec (moved title to be last)?

According to https://www.w3.org/TR/accname-1.2/#terminology I interpret the provided list as follows:

A) if an element has no indication nothing, return an empty string
B) return a accessibility name from `aria-labelledby` or `aria-describedby`
C) compute it from `aria-label`
D) compute it from `title` or `label` properties and if you open the comment section here you can see:

> For example, in HTML, the img element's alt attribute defines a text alternative string, and the label element provides text for the referenced form element. In SVG2, the desc and title elements provide a description of their parent element.

E) value of a textbox, alt text of a button etc.
F) return accessibility name from CSS pseudo classes
G) if it is a text node, return the text
H/I) not sure but probably not too relevant

We surely can't be 100% correct to the spec because xPath doesn't allow us to e.g. fetch an element by its pseudo css properties nor do we have access to the actual a11y tree.

christian-bromann:
> I'm fine with `~Foobar` :)

I just realised that if we merge them together and pick `accessibility id` when `isMobile` is true and this new one if it is false then everyone running web tests using Appium are screwed. So I gonna stay with `aria/*`.

christian-bromann:
> Should ~ be deprecated in favor of aria/ ?

@nextlevelbeard no, because the current `~` is for native mobile testing and results in a selector strategy called `accessibility id` which is used Appium to fetch an element by its mobile accessible id. As mentioned above, merging them is difficult because we can't tell whether a user is currently running web tests or native tests with Appium. Therefore we stick with `aria/*` for the web accessibility name.

nextlevelbeard:
> > I'm fine with `~Foobar` :)
> 
> I just realised that if we merge them together and pick `accessibility id` when `isMobile` is true and this new one if it is false then everyone running web tests using Appium are screwed. So I gonna stay with `aria/*`.

@christian-bromann  Couldn't this easily be checked by grabbing the current context?
i.e. if `isMobile` plus context equals `native` use mobile `accessibility id` strategy, otherwise use this feature.

PR #8770, opened 2022-08-29
docs: change function based on explanation in previous paragraph

basarat:
Previous paragraph says:

> Therefore we need to replace `forEach` with `map` which returns that promise


christian-bromann:
It says:

> The above example looks transformed like this:

And the example further above was:

```js
const elems = $$('div')
elems.forEach((elem) => {
    elem.click()
})
```

So I don't think it makes sense to switch to `map`. Thanks for raising a PR though.

PR #8863, opened 2022-09-16
Throw error if select element/option is disabled 

harsha509:
## Proposed changes

* Currently WDIo select, selects the option even if the Select tag/ option is disabled for selection. We need to throw an error is Select tag/option is disabled for selection

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/webdriverio/src/commands/element/selectByAttribute.ts
+++ b/packages/webdriverio/src/commands/element/selectByAttribute.ts
@@ -40,6 +40,14 @@ export default async function selectByAttribute (
     attribute: string,
     value: string | number
 ) {
+
+    /**
+     * Throw error if Select element is disabled
+     */
+    if (!(await this.isElementEnabled(this.elementId))) {
+        throw new Error(`Select element is disabled and may not be used.`)
+    }
+
     /**
      * convert value into string
      */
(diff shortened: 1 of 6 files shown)


christian-bromann:
@harsha509 thanks for raising the PR!

For some reason I was not able to reproduce this problem. Here is an example script I used:

```js
await browser.url('http://guinea-pig.webdriver.io/')

const select = await $('#selectTest')
await browser.execute(
    (elem) => elem.setAttribute('disabled', true),
    await select.$$('option')[2])
await browser.pause(2000)
await select.selectByVisibleText('tres')
await expect(select).toHaveValue('someValue3')
```

As you can see I disable the 3rd option and then try to use one of the select commands to change the value. However the test fails and I am not able to modify the select input.

Can you provide a reproducible example?

harsha509:
Hi @christian-bromann ,

https://github.com/harsha509/test_selector_wdio

christian-bromann:
> The test fails only on assertion, but it allows the user to select disabled option according to logic.

Ah, so you would like to have the attempt of changing a disabled input fail. That could be an interesting approach, the only problem I see with this is that it requires an additional command which could slow down the tests if there are a lot of inputs.

@webdriverio/project-committers any feedback, suggestions?

christian-bromann:
@harsha509 another option would be to extend our [`CustomRequestError`](https://github.com/webdriverio/webdriverio/blob/53eefeb7ad55d3d9ac0f83d5c6d5f02d0ea5da62/packages/webdriver/src/utils.ts#L238-L265) class and modify the error message there. What do you think?

PR #8905, opened 2022-09-25
GitHub Workflows security hardening

sashashura:
This PR adds explicit [permissions section](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to workflows. This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -20,6 +20,7 @@ on:
 env:
   NPM_TOKEN: ${{secrets.NPM_TOKEN}}
 
+permissions: {}
 jobs:
   authorize:
     runs-on: ubuntu-latest
@@ -32,6 +33,8 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
   release:
+    permissions:
+      contents: write
     needs: [authorize]
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,6 +9,9 @@ on:
       - v7
   pull_request:
 
+permissions:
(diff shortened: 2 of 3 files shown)


.github/workflows/publish.yml
```diff
@@ -32,6 +33,8 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
   release:
+    permissions:
+      contents: write
```
christian-bromann, line 37: Shouldn't this also say `read` as we also have `actions/checkout` in place? Or is it implied that when you can write you automatically can read as well.

christian-bromann (commented):
Thanks for the PR, one question.

christian-bromann:
Also, mind making the same PR for the `v7` branch? Thank you!

.github/workflows/publish.yml
sashashura: It assumes read too, i.e. read-write.

christian-bromann (approved):
👍

PR #8920, opened 2022-09-26
GitHub Workflows security hardening

sashashura:
as asked in https://github.com/webdriverio/webdriverio/pull/8905#issuecomment-1257419380


christian-bromann (approved):
Thanks 👍

Issue #8730, opened 2022-08-16
[🐛 Bug]: The `exclude` capability parameter no longer ignores desired spec files when the --spec command is used to run a test


dragosMC91:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

7.20.9

### Node.js Version

12.22.7

### Mode

WDIO Testrunner

### Which capabilities are you using?

```typescript
[
        {
            browserName: 'chrome',
            'goog:platformName': 'desktop',
            exclude: [
                './test/specs/*mobile*.js'
            ],
            'goog:loggingPrefs': { browser: 'WARNING' },
            'goog:chromeOptions': {
                args: [
                    '--no-sandbox',
                    '--disable-dev-shm-usage',
                    '--window-position=1050,210',
                    '--window-size=1366,768',
                ],
            },
        },
        {
            browserName: 'chrome',
            'goog:platformName': 'mobile',
            exclude: [
                './test/specs/*desktop*.js'
            ],
            'goog:loggingPrefs': { browser: 'WARNING' },
            'goog:chromeOptions': {
                mobileEmulation: {
                    deviceName: 'iPhone 8',
                },
                args: [
                    '--window-position=1050,210',
                    'use-mobile-user-agent',
                ],
                prefs: {
                    protocol_handler: {
                        excluded_schemes: {
                            sms: false,
                            tel: false,
                        },
                    },
                },
            },
        },
    ]
```

### What happened?

After updating to the latest wdio cli version I noticed that the `exclude` capability was no longer filtering out certain tests from execution based on some naming rules.

A quick look made me thing that this issue was coming from https://github.com/webdriverio/webdriverio/pull/8456/files

### What is your expected behavior?

Given an array of capabilities, when I run a single spec file using the `--spec` command I expect the that test to run only with the capabilities which have an `exclude` attribute that does not match the name of the current test.

### Steps to reproduce:
1. pull https://github.com/dragosMC91/wdio-exclude-pattern-bug-highlight
2. yarn install in the root (not that i'm using yarn **1.22.10** and node v **12.22.7**)
3. run the dummy test via `yarn test --spec test/specs/example.desktop.test.js`

### Expected result
[...]

PR #9013, opened 2022-10-16
Fix exclude pattern when --spec is passed via cli

dragosMC91:
## Proposed changes

This fixes the issue mentioned in this ticket: https://github.com/webdriverio/webdriverio/issues/8730.

The current problem is that the `exclude` pattern is only taken into account when the `--spec` CLI option is not provided. This is a powerful option which adds the possibility to filter tests in a fine-grained manner and should always be applied when spec/suite options were passed or not.

The proposed changes achieve 2 things:
1. make the `exclude` option independent of the `spec`/ `suite` option.
2. simplify the logic within `getSpecs()` to make it more obvious what happens there and when

## Types of changes

- [x] Documentation update

## Checklist

## Further comments

While I was not able to setup wdio locally and run tests, I created a dummy project and did the following validations:
- [x] Run `wdio wdio.conf.js` with config specs and capability specs to validate capability specs take precedence
- [x] Run `wdio wdio.conf.js --spec someSpec` with config specs and capability specs to validate CLI spec takes precedence
- [x] Run `wdio wdio.conf.js --suite someSuite` and validate only tests from that suite are run 
- [x] Run `wdio wdio.conf.js --suite someSuite --spec someSpec` and validate tests from the suite and the single spec are run

Also ran all of the above while having defined a capability `exclude` to check the desired patterns are filtered out.

## Potential problems
This generates a small consistency issue when it comes to precedence of spec / exclude options as follows: 
> CLI `--spec` argument > capability `specs` pattern > config `specs` pattern
> CLI `--exclude` argument > config `exclude` pattern > capability `exclude` pattern

This is because we don't know if the `exclude` option is coming from the config or CLI like we do for the `spec` option. This can be addressed but is out of scope for this PR and if it's decided to be pursued, should be addressed in a separate PR


hammzj:
Hey @dragosMC91 , jumping on here over from #8730.

I originally made the change for the `--spec` pattern but didn't add any changes for `--exclude`, so it's great to see some fixes added here!

There's one thing I'd like to point out, though, and it's that there are good number of expectations that need to be met when using the exclude parameter, and I made note of what tests are needed [over in my original PR's comment here](https://github.com/webdriverio/webdriverio/pull/8432#issuecomment-1162020655). The combination of both `--spec` and `--exclude` can get tricky, so I'd recommended to please take a look at the outcomes mentioned in that comment to ensure that the expectations in each of those cases still pass.

dragosMC91:
> Hey @dragosMC91 , jumping on here over from #8730.
> 
> I originally made the change for the `--spec` pattern but didn't add any changes for `--exclude`, so it's great to see some fixes added here!
> 
> There's one thing I'd like to point out, though, and it's that there are good number of expectations that need to be met when using the exclude parameter, and I made note of what tests are needed [over in my original PR's comment here](https://github.com/webdriverio/webdriverio/pull/8432#issuecomment-1162020655). The combination of both `--spec` and `--exclude` can get tricky, so I'd recommended to please take a look at the outcomes mentioned in that comment to ensure that the expectations in each of those cases still pass.

hello @hammzj, thanks for taking the time to reply to this PR

I used this dummy repo (PR) to validate all my changes: https://github.com/dragosMC91/wdio-exclude-pattern-bug-highlight/tree/debug-config-parser-changes

Results for the mentioned test cases: 
1. Pass
2. Pass
3. Fail - I already mentioned this in the PR description in the `Potential problems` section
4. Pass
5. Result: CLI-defined pattern should be used - I think this makes most sense, it's also consistent with the behavior of the `spec` option

I also ran combinations of the above with `spec` and `suite` cli options but I doubt I covered all combinations, the most obvious ones are working fine it seems

hammzj:
Awesome! Very glad to see these tests addressed and are consistent. I think for test 3, yes, it would be out of scope as the capability exclusion pattern is removed prior to the config exclusion pattern being defined, if I recall correctly.

These tests will still need to be added to the WebDriverIO suite with an assertion of which spec pattern should be expected based on the `--spec`/`--exclude` pattern combinations. I'm not too familiar with how to set up those tests in this suite, but it looks like you have a grasp on which pattern is expected to be run. 

Nice work here!

dragosMC91:
> Awesome! Very glad to see these tests addressed and are consistent. I think for test 3, yes, it would be out of scope as the capability exclusion pattern is removed prior to the config exclusion pattern being defined, if I recall correctly.
> 
> These tests will still need to be added to the WebDriverIO suite with an assertion of which spec pattern should be expected based on the `--spec`/`--exclude` pattern combinations. I'm not too familiar with how to set up those tests in this suite, but it looks like you have a grasp on which pattern is expected to be run.
> 
> Nice work here!
> 
> E: Also, @christian-bromann will need to merge/check this PR as I'm just a contributor on this issue -- I've just have been messing with this one thing for a bit 😃

I was not able to build the project locally on my mac machine and run/ extend unit tests...followed the steps here https://webdriver.io/docs/contribute#set-up-project , used nvm with node 14 lts but `npm run build` fails, is there any step from the setup guide missing ?

I will also try on my windows/ linux machine once im done with work see if it works there and I can add some new tests and fix the failing ones.

Any pointers on how to fix `npm run build` are appreciated :D

christian-bromann:
@dragosMC91 update your branch, currently `main` is passing.

dragosMC91:
> @dragosMC91 please update your branch, currently `main` is passing.

@christian-bromann i followed the steps from https://webdriver.io/docs/contribute/#make-a-pull-request but i can't setup the project, `npm run build` fails. Tried on both my macbook and my linux (WSL) machine. I want to run all tests locally and extend existing tests if i figure out how they work.
Im using as suggested in the docs node 14 lts. The output for the build command on my linux WLS (ubuntu 20) machine is:
```
➜  webdriverio git:(fix-exclude-pattern-when-cli-spec-passed) npm run build

> build
> run-s clean:build generate:typings compile copy


> clean:build
> rimraf ./packages/*/build


> generate:typings
> node ./scripts/type-generation/generate-all-typings.js

file:///home/dragoscampean/Documents/webdriverio/scripts/constants.js:1
import webdriver from '../packages/wdio-protocols/protocols/webdriver.json' assert { type: 'json' }
                                                                            ^^^^^^

SyntaxError: Unexpected identifier
    at Loader.moduleStrategy (internal/modules/esm/translators.js:145:18)
ERROR: "generate:typings" exited with 1.
```

dragosMC91:
> You are on a wrong Node.js version. Call `nvm install`. I will update the contributing docs to make sure this is mentioned.

ok this fixed it, i can build now and see the failing tests and will update them accordingly asap, i hope tomorrow after work

packages/wdio-config/tests/configparser.test.ts
```diff
@@ -728,9 +728,16 @@ describe('ConfigParser', () => {
...
+
+            // first validate that the conf fixture 'spec' was merged successfully
+            expect(configParser.getSpecs()).toHaveLength(1)
+
+            configParser.merge({ exclude: [path.join(__dirname, '**', '*conf*')] })
+
+            // then after merging an exclude containing a glob pattern, validate that the exclude
+            // attribute contains multiple items and the filtering on the spec attribute works
+            expect(configParser.getConfig().exclude.length).toBeGreaterThan(0)
+            expect(configParser.getSpecs()).toHaveLength(0)
```
dragosMC91, line 740: this test was added via this PR: https://github.com/webdriverio/webdriverio/pull/4544/files but from my understanding, in its previous form, it was not actually validating what it stated... the test was passing even without the changes from https://github.com/webdriverio/webdriverio/pull/4544/files#diff-e0b205fb022f00db1b2dafbf906129bd7627eeea8db65d3be9285f6aa215c691R217-R221

The only reason it started failing in my PR is because i eliminated duplicated from `getSpecs()` so i tried to make it more robust

packages/wdio-config/tests/configparser.test.ts
```diff
@@ -728,39 +728,55 @@ describe('ConfigParser', () => {
...
+
             configParser.merge({ exclude: [INDEX_PATH] })
-            const specs = configParser.getSpecs()
-            expect(specs).toHaveLength(4)
+            // after overriding the config exclude with cli exclude check that the initial config exclude is discarded
+            expect(configParser.getSpecs()).toHaveLength(4)
         })
 
-        it('should overwrite exclude if piped into cli command with params', async () => {
+        it('should overwrite config and capabilities exclude if piped into cli command', async () => {
```
dragosMC91, line 754: it looks to me as if `getFilePaths()` has a hidden bug:
```
ConfigParser.getFilePaths(['wdio-config/src/index.ts', 'wdio-config/src/constants.ts',
    'wdio-config/src/types.ts', 'wdio-config/src/utils.ts'], undefined, this._pathService)
```
only returns the `index.ts` file which is problematic to all tests using this because they can allow bugs to pass uncaught by the unit tests (i tested this on the latest `main` branch).

I only refactored the tests which were related to take the above mentioned into account.

packages/wdio-config/tests/configparser.test.ts
```diff
@@ -728,39 +728,55 @@ describe('ConfigParser', () => {
...
-            configParser.merge({})
+            expect(configParser.getSpecs()).toHaveLength(3)
 
-            const utilsPath = path.join(__dirname, '..', 'src', 'utils.ts')
-            const indexPath = path.join(__dirname, '..', 'src', 'index.ts')
-            const specs = configParser.getSpecs([indexPath, utilsPath], [utilsPath])
-            expect(specs).toEqual([indexPath])
+            configParser.merge({ exclude: [FIXTURES_CONF] })
+            const specs = configParser.getSpecs([FIXTURES_CONF, FIXTURES_CONF_RDC], [FIXTURES_CONF_RDC])
+            expect(specs).toEqual([FIXTURES_CONF_RDC])
```
dragosMC91, line 761: there were multiple problems with this test:
1. no CLI exclude param was provided
2. because of the [issue mentioned before](https://github.com/webdriverio/webdriverio/pull/9013/files#r1002671129), when taking into account the validation:
```
            const specs = configParser.getSpecs([indexPath, utilsPath], [utilsPath])
            expect(specs).toEqual([indexPath])
```
even if a bug were introduced which made it such that the `utilsPath` capability exclude is no longer taken into account when it should, the above statement would still pass, missing the bug

dragosMC91:
@christian-bromann with a small delay i made the promised updates to the tests, lots of work last week and i didn't get around to do it then.

The tests were a bit harder to navigate at first and I am probably not very apt at reading them but it looked to me like some of them, at least the ones i went through, were not really validating what they were supposed to, and would have passed even a bug was introduced in the area they were safeguarding.

I added some comments in the code directly to get a better view of the intent of some of these tests and in the PR for future contributors.

Do you think these comments make sense/ are useful and should i invest some time in the future to review more of the unit tests and address the ones which are problematic ?

packages/wdio-config/tests/configparser.test.ts
hammzj: I kept having this issue when writing unit tests as well. It made it extremely difficult to figure out what was going wrong.

dragosMC91: maybe some veteran could help us figure out whats going on here, if not I will just create a ticket to be investigated separately

dragosMC91:
@christian-bromann sorry to be pushy but kind reminder for the above whenever you have time to review

packages/wdio-config/tests/configparser.test.ts
christian-bromann: The configparser tests use a mocked file system which makes writing tests around it a bit difficult. Ideally I would love to rewrite this but this is how things are atm, see https://github.com/webdriverio/webdriverio/blob/6a4dd4402e82b6bf87ed802adb0883b08b16dffd/packages/wdio-config/tests/configparser.test.ts#L138-L150

christian-bromann:
@dragosMC91 thanks for all the work on this. The code looks good to me.  The config parser module is the only one that is a bit tricky to test due to its abstracted custom file system API it uses. Is there anything left to do or is this good to go?

Also would you mind to make the same changes in the `v7` branch so we can release it for these versions too?

christian-bromann:
The errors are unrelated to your PR, please update your branch and they should be gone.

dragosMC91:
> @dragosMC91 thanks for all the work on this. The code looks good to me. The config parser module is the only one that is a bit tricky to test due to its abstracted custom file system API it uses. Is there anything left to do or is this good to go?
> 
> Also would you mind to make the same changes in the `v7` branch so we can release it for these versions too?

Good morning, thank you for looking over the changes!

This PR is good to go, reviewing/fixing the existing config parser tests should be done in a separate PR if you think that's a good idea.

I made a branch from v7 and cherry picked the commits from this PR and removed the config parser async await stuff because i saw that's not yet implemented there: https://github.com/webdriverio/webdriverio/pull/9103/files.
Also made a small commit to fix a failing test on windows in both PRs https://github.com/webdriverio/webdriverio/pull/9103/commits/805a598f00ec972b62fbc0a96f1c741355998be5.

If anything else needs doing i will do it tomorrow, will be away for the weekend.

dragosMC91:
@christian-bromann I just updated the branch again to contain the latest changes in `main`, did the same with https://github.com/webdriverio/webdriverio/pull/9103 and rebased with latest `webdriverio/v7`. Can you please trigger the ci?

christian-bromann:
Thanks, I will review once I get the time.

christian-bromann (approved):
👍 Thanks a lot!

packages/wdio-config/src/lib/ConfigParser.ts
```diff
@@ -190,17 +190,28 @@ export default class ConfigParser {
...
+        // when CLI --spec is explicitly specified, this._config.specs contains the filtered
+        // specs matching the passed pattern else the specs defined inside the config are returned
+        let specs = ConfigParser.getFilePaths(this._config.specs, undefined, this._pathService)
         let exclude = ConfigParser.getFilePaths(this._config.exclude!, undefined, this._pathService)
         let suites = Array.isArray(this._config.suite) ? this._config.suite : []
 
-        /**
-         * check if user has specified a specific suites to run
-         */
+        // only use capability excludes if (CLI) --exclude or config exclude are not defined
```
damencho, line 204: @dragosMC91 @christian-bromann Why is that? 
I have a case where I want to exclude tests via `config: { exclude` and via `capabilities: {
                ....,
                'wdio:exclude': [`
                 
Can we change that to not be the case and use both?
I changed the code to work for my case:
```
        if (Array.isArray(capExclude)) {
            exclude = [ ...exclude, ...ConfigParser.getFilePaths(capExclude, this._config.rootDir, this._pathService)]
        }
```
Is that acceptable?

damencho: Yeah, I see 3 tests failing, seems this is a no go .. Anyway I'm curious why cannot both excludes work together and be merged?

christian-bromann: I'ld suggest to create a new issue and treat this as a feature request. Thanks!

PR #9091, opened 2022-10-25
fix(junit-reporter): strip ansi escape codes from error stack

alk-sdavid:
## Proposed changes

If the error message contains some ANSI escape codes such as colors, the stacktrace is not cleaned [as the error message](https://github.com/webdriverio/webdriverio/blob/7659670b5a8d12662f4b942e2ddd86e69ed8b2e0/packages/wdio-junit-reporter/src/index.ts#L192), for example:

```shell
Expected: [32m"Local [7ma[27mdministrator"[39m
Received: [31m"Local [7mA[27mdministrator"[39m
    at Context.<anonymous> ...
```

So if an error stack exists, I striped ANSI escape codes from it as well.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

None


christian-bromann (approved):
👍  Thanks! Mind making the same PR for the `v7` branch so we can release it also for `v7,x` versions.

alk-sdavid:
@christian-bromann done, I backported this fix on v7: https://github.com/webdriverio/webdriverio/pull/9094

On this other hand, there are some smoke tests failing on windows, but it seems unrelated to my changes, are they just some flaky tests?

packages/wdio-junit-reporter/src/index.ts
```diff
@@ -201,7 +201,7 @@ class JunitReporter extends WDIOReporter {
                         // default
                         testCase.error(test.error.message)
                     }
-                    testCase.standardError(`\n${test.error.stack}\n`)
+                    testCase.standardError(`\n${test.error.stack?.replace(ansiRegex, '')}\n`)
```
erwinheitzman, line 204: I don't know where these ASCII characters are coming from but I think we might have to apply this to the output on line 212 as well, what do you think?

erwinheitzman (commented):
Thanks for this improvement, the code looks good to me.
I left a question for you before I approve it.

packages/wdio-junit-reporter/src/index.ts
alk-sdavid: indeed, I'll do that

alk-sdavid: @erwinheitzman finally it's not so trivial since this output is processed from objects and stringified by `JSON.stringify` which escaped back slashes and so the regex does not match escaped ansi codes. In the end I think it's not necessary to strip ansi codes from this output, because it seems to be mostly wdio internal logs (such as commands), and if a command send some ansi codes, I guess it's fine to log them.

erwinheitzman (approved):
LGTM

alk-sdavid:
Hello, can you run the CI workflows?

Issue #6777, opened 2021-04-28
@wdio/browserstack-service: Name is changed after every test even if the name capability is set


facusantillo:
[//]: # NOTE: This repository only maintains packages that are listed in the [Readme](https://github.com/webdriverio/webdriverio/blob/main/README.md#packages). Please make sure that your issue is directly caused by one of these packages and if not file an issue in the correct 3rd party package repository.

**Environment (please complete the following information):**
 - **WebdriverIO version:** [7.5.2]
 - **Mode:** [WDIO Testrunner]
 - **If WDIO Testrunner, running sync/async:** [async]
 - **Node.js version:** [12.19]
 - **NPM version:** [6.14.8]
 - **Browser name and version:** [appium]
 - **Platform name and version:** [macOS]
 - **Additional wdio packages used (if applicable):** [@wdio/browserstack-service]

**Config of WebdriverIO**
config.services = [
    [
        'browserstack',
        {
            args: {
                browserstackLocal: true
            },
            command: 'browserstack'
        }
    ]
],
config.capabilities = [
    {
        platformName: env.PLATFORM_NAME === 'Android' ? 'Android' : 'iOS',
        deviceName: env.DEVICE_NAME,
        platformVersion: env.PLATFORM_VERSION,
        orientation: env.DEVICE_ORIENTATION,
        automationName: env.PLATFORM_NAME === 'Android' ? 'UiAutomator2' : 'XCUITest',
        app: 'bs://' + env.BS_APP_ID,
        build: env.BS_BUILD,
        project: env.BS_PROJECT,
        name: env.BS_NAME,
        noReset: true,
        newCommandTimeout: 2400,
        recreateChromeDriverSessions: true,
        chromedriver_autodownload: true,
        relaxedSecurity: true,
        'browserstack.debug': true,
        autoGrantPermissions: true
    }
];
config.excludeDriverLogs = ['bugreport', 'server'];
config.user = env.BS_USER;
config.key = env.BS_KEY;
exports.config = config;

**Describe the bug**
When the suite starts executing on browserstack, every time a test is finished is overriding the name capability on the config

**To Reproduce**
Run a suite with wdio ./your_config_file --suite 'your suite with multiple tests in a describe'

**Expected behavior**
Running multiple tests from a same spec file shouldn't update the name capability if this is set.

**Log**
If applicable, add logs to help explain your problem.
[...]

PR #9107, opened 2022-10-31
@wdio/browserstack-service: Add session naming options

### LINKED ISSUE #7876: [🐛 Bug]: For mocha tests, BrowserStack session name is set as '<the last test title's describe> - <test title>'

### WebdriverIO Version

7.5.7

### Node.js Version

12

### Mode

Standalone Mode

```typescript
{
  os: 'OS X',
  osVersion: 'Catalina',
  browserName: 'Chrome',
  browserVersion: 'latest',
}

Wdio and other packages in use:
@wdio/browserstack-service
@wdio/cli
@wdio/local-runner
@wdio/mocha-framework
@wdio/selenium-standalone-service
@wdio/spec-reporter
webdriverio
selenium-standalone (^6.15)
```

### What happened?

homepage.spec.ts
```
describe('Homepage', () => {
  describe('header', () => {
    it('is rendered', () => {});
  }
  describe('navigation', () => {
    it('is present on the page', () => {});
  }
  describe('footer', () => {
    it('is present on the page', () => {});
  }
})
```
category.spec.ts
```
describe('Category Page', () => {
  describe('header', () => {
    it('is rendered', () => {});
  }
  describe('navigation', () => {
    it('is present on the page', () => {});
  }
  describe('footer', () => {
    it('is present on the page', () => {});
  }
})
```
**Current Behavior:**
With Mocha tests, the session name for the above spec files is same which is "footer - is present on the page".  (last test's parent describe - last test's title) as per https://github.com/webdriverio/webdriverio/blob/main/packages/wdio-browserstack-service/src/service.ts#L95

I have tried to rename the session name using BrowserStack REST API but it is overriding the session name by BrowserStack-service after the hook code.

**Additional Info:**
Fix done for Jasmine tests #6777 
https://github.com/webdriverio/webdriverio/pull/7198 

In the above case, the expected session name for homepage.spec.ts should be - `'Homepage'`
and for category.spec.ts, should be - `'Category Page'`. 
Expect the main/global describe title of the spec file as a session name so we can differentiate between different spec files. 

### How to reproduce the bug.

To reproduce the issue, run the mocha tests shown above(nested describe) in BrowserStack and observe the BrowserStack session names.

### Relevant log output

```typescript
The session name for both the spec files is the same - "footer - is present on the page"
```

### Code of Conduct

SCG82:
## Proposed changes

- Set the session name in `beforeSuite` and/or `beforeTest` instead of `afterTest`
  - allow the user the option to update the name at any given point in time.
  - account for cases where there is a long running `before` function, or one that can fail.
- Add options `setSessionName` and `setSessionStatus` (both default to `true`)
- Add option `sessionNamePrependTopLevelSuiteTitle `
  - For nested suites, set session name to: `<outer suite title> - <inner suite title>`
- Add option `sessionNameOmitTestTitle `
  - Set session name to: `<suite title>`
- Add option `sessionNameFormat` to allow customizing the session name using configuration, capabilities, suite title, test title (Mocha only)

Setting `sessionNamePrependTopLevelSuiteTitle =true` and `sessionNameOmitTestTitle=true` will resolve #7876 using the consensus from the discussion in #6777.

Examples, `sessionNamePrependTopLevelSuiteTitle` and `sessionNameOmitTestTitle` set to `true`:
```javascript
describe('webdriver.io', () => {
    describe('header bar', () => {
        it('should have logo', () => {})
    })
    describe('body', () => {
        xit('font should be...', () => {})
    })
})

// Session name before the fix: "header bar - should have logo"
// Session name after the fix: "webdriver.io - header bar"
```

Deeply nested suites:
```javascript
describe('webdriver.io', () => {
    describe('v6', () => {
        describe('header bar', () => {
            it('should have inline search', () => {})
        })
    })
})

// Session name before the fix: "header bar - should have inline search"
// Session name after the fix: "webdriver.io - header bar"
```

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/wdio-browserstack-service/README.md
+++ b/packages/wdio-browserstack-service/README.md
@@ -1,7 +1,7 @@
-WebdriverIO Browserstack Service
+WebdriverIO BrowserStack Service
 ==========
 
-> A WebdriverIO service that manages local tunnel and job metadata for Browserstack users.
+> A WebdriverIO service that manages local tunnel and job metadata for BrowserStack users.
 
 ## Installation
 
@@ -17,7 +17,7 @@ Instructions on how to install `WebdriverIO` can be found [here.](https://webdri
 
 ## Configuration
 
-WebdriverIO has Browserstack support out of the box. You should set `user` and `key` in your `wdio.conf.js` file. This service plugin provides support for [Browserstack](https://www.browserstack.com/automate/node#setting-local-tunnel) Tunnel](https://www.browserstack.com/automate/node#setting-local-tunnel). Set `browserstackLocal: true` also to activate this feature.
+WebdriverIO has BrowserStack support out of the box. You should set `user` and `key` in your `wdio.conf.js` file. This service plugin provides support for [BrowserStack](https://www.browserstack.com/automate/node#setting-local-tunnel) Tunnel](https://www.browserstack.com/automate/node#setting-local-tunnel). Set `browserstackLocal: true` also to activate this feature.
 Reporting of session status on BrowserStack will respect `strict` setting of Cucumber options.
 
 ```js
@@ -40,19 +40,13 @@ export const config = {
 In order to authorize to the BrowserStack service your config needs to contain a [`user`](https://webdriver.io/docs/options#user) and [`key`](https://webdriver.io/docs/options#key) option.
 
 ### browserstackLocal
-Set this to true to enable routing connections from Browserstack cloud through your computer.
-
-Type: `Boolean`<br />
-Default: `false`
-
-### preferScenarioName
-Cucumber only. Set this to true to enable updating the session name to the Scenario name if only a single Scenario was run. Useful when running in parallel with [wdio-cucumber-parallel-execution](https://github.com/SimitTomar/wdio-cucumber-parallel-execution).
+Set this to true to enable routing connections from BrowserStack cloud through your computer.
 
 Type: `Boolean`<br />
 Default: `false`
 
 ### forcedStop
-Set this to true to kill the Browserstack process on complete, without waiting for the browserstack stop callback to be called. This is experimental and should not be used by all. Mostly necessary as a workaraound for [this issue](https://github.com/browserstack/browserstack-local-nodejs/issues/41).
+Set this to true to kill the BrowserStack Local process on complete, without waiting for the BrowserStack Local stop callback to be called. This is experimental and should not be used by all. Mostly necessary as a workaraound for [this issue](https://github.com/browserstack/browserstack-local-nodejs/issues/41).
 
 Type: `Boolean`<br />
 Default: `false`
@@ -141,8 +135,53 @@ services: [
 ]
 ```
 
+### preferScenarioName
+
+Cucumber only. Set the BrowserStack Automate session name to the Scenario name if only a single Scenario ran.
+Useful when running in parallel with [wdio-cucumber-parallel-execution](https://github.com/SimitTomar/wdio-cucumber-parallel-execution).
+
+Type: `Boolean`<br />
+Default: `false`
+
+### sessionNameFormat
+
+Customize the BrowserStack Automate session name format.
+
+Type: `Function`<br />
+Default (Cucumber/Jasmine): `(config, capabilities, suiteTitle) => suiteTitle`<br />
+Default (Mocha): `(config, capabilities, suiteTitle, testTitle) => suiteTitle + ' - ' + testTitle`
+
+### sessionNameOmitTestTitle
+
+Mocha only. Do not append the test title to the BrowserStack Automate session name.
+
+Type: `Boolean`<br />
+Default: `false`
+
+### sessionNamePrependTopLevelSuiteTitle
+
+Mocha only. Prepend the top level suite title to the BrowserStack Automate session name.
+
+Type: `Boolean`<br />
+Default: `false`
+
+### setSessionName
+
+Automatically set the BrowserStack Automate session name.
+
+Type: `Boolean`<br />
+Default: `true`
+
+### setSessionStatus
+
+Automatically set the BrowserStack Automate session status (passed/failed).
+
+Type: `Boolean`<br />
+Default: `true`
+
 ### opts
-Specified optional will be passed down to BrowserstackLocal.
+
+BrowserStack Local options.
 
 Type: `Object`<br />
 Default: `{}`
@@ -185,7 +224,7 @@ opts = { f: "/my/awesome/folder" };
 
 #### Force Start
 
-To kill other running Browserstack Local instances -
+To kill other running BrowserStack Local instances -
 
 ```js
 opts = { force: "true" };
--- a/packages/wdio-browserstack-service/src/service.ts
+++ b/packages/wdio-browserstack-service/src/service.ts
@@ -5,6 +5,7 @@ import type { Browser, MultiRemoteBrowser } from 'webdriverio'
 
 import { getBrowserDescription, getBrowserCapabilities, isBrowserstackCapability, getParentSuiteName } from './util.js'
 import type { BrowserstackConfig, MultiRemoteAction, SessionResponse } from './types'
+import { DEFAULT_OPTIONS } from './constants.js'
 
 const log = logger('@wdio/browserstack-service')
 
@@ -14,15 +15,18 @@ export default class BrowserstackService implements Services.ServiceInstance {
     private _scenariosThatRan: string[] = []
     private _failureStatuses: string[] = ['failed', 'ambiguous', 'undefined', 'unknown']
     private _browser?: Browser<'async'> | MultiRemoteBrowser<'async'>
+    private _suiteTitle?: string
     private _fullTitle?: string
+    private _options: BrowserstackConfig & Options.Testrunner
 
     constructor (
-        private _options: BrowserstackConfig & Options.Testrunner,
+        options: BrowserstackConfig & Options.Testrunner,
         private _caps: Capabilities.RemoteCapability,
         private _config: Options.Testrunner
     ) {
+        this._options = { ...DEFAULT_OPTIONS, ...options }
         // added to maintain backward compatibility with webdriverIO v5
-        this._config || (this._config = _options)
+        this._config || (this._config = this._options)
         // Cucumber specific
         const strict = Boolean(this._config.cucumberOpts && this._config.cucumberOpts.strict)
         // See https://github.com/cucumber/cucumber-js/blob/master/src/runtime/index.ts#L136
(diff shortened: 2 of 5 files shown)


christian-bromann:
@SCG82 it seems like unit tests are failing 🤔

SCG82:
> @SCG82 it seems like unit tests are failing 🤔

Ready for review

packages/wdio-browserstack-service/src/service.ts
```diff
@@ -95,7 +97,12 @@ export default class BrowserstackService implements Services.ServiceInstance {
             }
         } else {
             // Mocha
-            this._fullTitle = `${test.parent} - ${test.title}`
+            // eslint-disable-next-line no-lonely-if
```
erwinheitzman, line 100: This rule is quite valid and I don't agree disabling it: https://eslint.org/docs/latest/rules/no-lonely-if

packages/wdio-browserstack-service/src/service.ts
```diff
@@ -95,7 +97,12 @@ export default class BrowserstackService implements Services.ServiceInstance {
...
+            if (test.parent && test.parent !== this._suiteTitle) {
+                this._fullTitle = `${this._suiteTitle} - ${test.parent}`
```
erwinheitzman, line 102: Feels like this needs to happen recursively when there is more than one parent suite, in other words, I do not think this handles cases like this:
```js
describe('my component', () => {
  describe('scenario A', () => {
    describe('and some other condition is truthy', () => {
      it('example', () => {
        // this now has three parents, does this work?
      });
    });
  });
});
```

erwinheitzman (changes requested):
Thank you for this PR, I have left a question and a request for change

packages/wdio-browserstack-service/src/service.ts
SCG82: For this example, it will set the session name to: `my component - and some other condition is truthy`. For comparison, the sauce labs service will set the job name to `my component` and add context of `and some other condition is truthy - example`. 

I can't figure out how to obtain `scenario A` from any of the hooks. Any ideas?

SCG82:
Made a few changes, and added some tests (based on the sauce service tests). Also fixed the type errors in the service test.

packages/wdio-browserstack-service/src/service.ts
SCG82: Another option is to use one of the following as the session name:
- `this._suiteTitle` (which would be "my component" in the above example)
- `test.parent` ("and some other condition is truthy")

or keep the current title (test.parent - test.title) except set it before the test (so users may change it without being overridden).

christian-bromann:
@Ankit098 what do you think?

christian-bromann:
@SCG82 mind updating this branch, it should resolve the build issues.

christian-bromann:
@Ankit098 what do you think?

Ankit098 (commented):
1. in the following scenario:
```
describe("describe title", () => {
  it("it title", () => {
  })
})
```
the session name after new changes will be `describe title` (from `describe title - it title` earlier).. and by extension the `it block`'s name gets dropped even in nested describe.. I think we should include the it block's title.
2. for nested desribes, like either:
```
describe("describe1", () => {
  describe("describe2", () => {
    describe("describe3", () => {
      it("it1", () => {
      })
      it("it2", () => {
      })
    })
  })
})
```
or
```
describe("describe1", () => {
  describe("describe2", () => {
    it("it1", () => {
    })
  })
  describe("describe3", () => {
    it("it2", () => {
    })
  })
})
```
the session name picked will be `describe1 - describe3` here we can introduce some [annotations](https://www.browserstack.com/docs/automate/selenium/annotations) to differentiate the different blocks' execution in the logs on browserstack dashboard. +@alliv8

SCG82:
I can make this change, although it would be preferable if BrowserStack introduced a "subtitle" / "context" in which to display the current `it` step (similar to SauceLabs' 'context') so that the job title can be set by the user (in `before`) and not overridden.

SCG82:
@christian-bromann @Ankit098 ready for review

SCG82:
@christian-bromann @Ankit098 ready for another review

alliv8:
I think adding more verbosity would be a personal choice and whatever changes we make, we should not modify existing behaviour for users (leads to users saying why did you change this; I didn't ask for it), and rather extend functionality in cases where the outcome is undesired or add options for more advanced use-cases.

In terms of BrowserStack adding further classification, we've attempted to solve this using console annotators, which can be seen and filtered within the session log context to highlight different `describe` and `it` blocks. The service could automatically use the beforeHook for each `describe` and  `it` context to be passed to the BrowserStack session.

I'm wondering if we could create options to support the different preferences that have been brought up, and keep current behaviour as default?

```
sessionNamingOpts: {
  includeLastStep: true / false (default true) -- 
  maxDescribeLevels: 1  
}
```

or we instead allow for users to decide, and disable the service naming and test marking logic

```
setSessionName: true / false
```

SCG82:
> we should not modify existing behaviour for users

Existing behavior is unchanged by this PR, except that the job name is set before the suite/test instead of only after the test has completed.

SCG82:
@alliv8 @Ankit098 @christian-bromann Added `setSessionName`, `setSessionStatus` options + tests and updated README.

SCG82:
@erwinheitzman can I get another review?

packages/wdio-browserstack-service/README.md
```diff
@@ -141,8 +135,53 @@ services: [
...
+Default: `false`
+
+### setSessionName
+
+Automatically set the BrowserStack Automate session name.
+
+Type: `Boolean`<br />
+Default: `true`
+
+### setSessionStatus
```
christian-bromann, line 175: Just curious, I would love to understand when people are interested to turn this off. Is it because these additional request slow down the test run?

christian-bromann (approved):
Overall fantastic change set, great work 👍 

Please see my last question, which I would like to see answered before merging. Having this options seems great but if not really needed we could avoid adding it as it means also more stuff to maintain 😉

packages/wdio-browserstack-service/README.md
SCG82: For my use case I'd like to set the session status/reason in an `after` hook without it being overridden.

SCG82:
changes since last review:
* apply `sessionNameFormat` in all locations where the session name is set.

packages/wdio-browserstack-service/README.md
christian-bromann: Why are you setting the name in the after hook? Just curious.

christian-bromann:
@Ankit098 this change looks good to me. Can you do a second review and give me your 👍 ?

packages/wdio-browserstack-service/README.md
ablok: It's what I would do as well. I would like to write the path of the module there as session name.
Unless browserstack starts supporting it's within a session.

alliv8: For session name:

We've seen instances of incompatibility of the cucumber wdio parallel plugin and the service's naming logic at a scenario level.

In this case, users have opted to not use the service entirely.

Another one was where describe blocks were written in Korean, and Browserstack currently doesn't support those characters in the name.

While the above are either bugs or fixable limitations, at least the flexibility helps users to fix things themselves, and not have to troubleshoot a behaviour where they set context manually and then it gets overridden on BrowserStack via the service.

For session status:

I've seen patterns where the session reason is set to a grouped message based on all the assertions that have failed in a session.

Something like
"1 of 3 tests failed"

Currently we set the reason based on a single failure

christian-bromann: Makes sense to me, thanks for clarifying @alliv8

SCG82: > Why are you setting the name in the after hook? Just curious.

Not the name but the status (passed/failed) and a descriptive `reason` that conveys additional information if `passed`.

bhanuagarwal73 (approved):

bhanuagarwal73 (approved):

Ankit098 (approved):
LGTM! @SCG82 @christian-bromann

Ankit098:
A minor improvement that can be done in a future PR is to add a js executor in the block where the session name is now being updated to add annotations in the logs to better distinguish the different describes/its being executed in the same session - as right now the session name gets updated with each it block and the final name is the last it block - having the annotations will put the different describes/its as visually different blocks in the logs, making logs more meaningful and debugging easier.

SCG82:
> A minor improvement that can be done in a future PR is to add a js executor in the block where the session name is now being updated to add annotations in the logs to better distinguish the different describes/its being executed in the same session - as right now the session name gets updated with each it block and the final name is the last it block - having the annotations will put the different describes/its as visually different blocks in the logs, making logs more meaningful and debugging easier.

Does https://github.com/webdriverio/webdriverio/pull/9124 handle this?

bhanuagarwal73:
Hey @Ankit098 @christian-bromann  I am not seeing sessionNameOmitTestTitle in bstack:options.
Generally using as below rather than at service level

"bstack:options": runtimeArgs.service === 'browserstack' ? {
            "projectName": "xxxxxxx",
            "buildName": `${runtimeArgs.env} - ${runtimeArgs.brand} - ${runtimeArgs.suite} - ${runtimeArgs.browserName}`,
            "os": runtimeArgs.os,
            "osVersion": runtimeArgs.osVersion,
            "deviceName": runtimeArgs.device,
            "networkLogs": true,
            "consoleLogs": "errors",
            "browserVersion": "latest",
            "local": true
        } : {}

I don't want to define at service level:
services: [runtimeArgs.service == 'browserstack' ? [runtimeArgs.service, {
        "sessionNameOmitTestTitle": true
    }] : runtimeArgs.service],

christian-bromann:
@bhanuagarwal73 you have to add `@wdio/browserstack-service` to your `"types"` property in `tsconfig.json`.

bhanuagarwal73:
@christian-bromann  added but still not seeing it.
image

{
    "compilerOptions": {
        "esModuleInterop": true,
        "types": [
            "node",
            "webdriverio/async",
            "@wdio/mocha-framework",
            "@wdio/browserstack-service"
        ],
        "target": "ES5",
        "downlevelIteration": true
    }
}

christian-bromann:
It works on my machine™️
Screenshot 2022-11-25 at 10 21 40

christian-bromann:
This option has to be applied to the service not to your capabilities.

bhanuagarwal73:
That's what asked in initial comments. Thanks for the response as there is no way over bstack:Options, will go with  service level :)

christian-bromann:
I am not sure what you mean by that.

bhanuagarwal73:
Nothing buddy. I am saying ok, will add in service

ssomise5:
Can some one assist , how to use the above fix . Do we need to have specific wdio version

ssomise5:
can someone help me with this bug fix

Issue #9279, opened 2022-12-01
[🐛 Bug]: v8 @wdio/utils - safeImport finds and imports the CommonJS version of a package


RossVertizan:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

next - 8.0.0-alpha.631

### Node.js Version

v18.12.0

### Mode

WDIO Testrunner

### Which capabilities are you using?

```typescript
Not related to capabilities
```

### What happened?

In order to make the vitaqai-service work, I have a slightly modified version of mocha - ```vitaqai-mocha```.  In order to load this I have a minimally modified version of the ```@wdio/mocha-framework``` called ```wdio-vitaqai-mocha-framework```, which simply imports Mocha from ```vitaqai-mocha``` as opposed to ```mocha```.

In order to support the CommonJS (require) structure of WebdriverIO V7 (and below) and the ESM (import) structure of WebdriverIO V8 (and above), I have prepared a version of the  ```wdio-vitaqai-mocha-framework`` package which has both CJS and ESM variants and the necessary directives in the ```package.json``` file to load the correct version.

This approach works fine in V7.  However in V8, the CJS variant is loaded and this subsequently causes errors, as packages are then loaded using require and I get this message:
```Error [ERR_REQUIRE_ESM]: require() of ES Module Y:\vitaqai_itc_testing\vitaq-example-v8\node_modules\@wdio\logger\build\index.js from Y:\vitaqai_itc_testing\vitaq-example-v8\node_modules\wdio-vitaqai-mocha-framework\build\cjs\index.js not supported.```

### What is your expected behavior?

I expected that the ESM variant of the package would be loaded in preference to the CJS variant

### How to reproduce the bug.

I am not going to provide a reproducible test case at this point as I am happy to work through this and submit a pull request.  All I need from the team is guidance on what is expected behaviour in V8.  It may be that my expectations are incorrect.  If necessary I will provide a simple example later.

I will add what I have done so far in a first reply to this issue.

### Relevant log output

```typescript
[0-0] Starting VitaqAI client for connection to Vitaq API ...
[0-0] ==> Connected to Vitaq
[0-0] ==> Sending authentication request
[0-0] ==> Authenticated
[0-0] ==> Registered
[0-0] ==> Python script launched successfully!
[...]

PR #9309, opened 2022-12-05
Implemented import-meta-resolve to import ESM modules in safeImport

RossVertizan:
## Proposed changes

The modifications are to the safeImport method in the file utils.ts in the package wdio-utils, to address the issue raised in #9279.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

Just some quick notes:
1. I have used the import-meta-resolve package as has been used elsewhere and discussed in #9279.
2. Since resolve throws an error if the package is not found I have used a nested try-catch, which I don't really like, but was better (IMO) than an empty catch block.  I have nesting so that I can first look on the ```import.meta.url``` path and then on the   ```localNodeModules``` path, to best replicate the existing solution.
3. I preferentially import the ESM module and if that is undefined try to load the module found by ```require.resolve```
4. I also has to add ```.js`` to the types imports in the wdio-protocols package as they were failing eslint checks.


packages/wdio-utils/src/utils.ts
```diff
@@ -176,6 +178,21 @@ export async function safeImport (name: string): Promise<Services.ServicePlugin
...
+         */
+        try {
+            esmImportPath = await resolve(name, import.meta.url)
+        } catch (err: any) {
+            try {
+                esmImportPath = await resolve(name, pathToFileURL(localNodeModules).toString())
+            } catch (err: any) {
+                console.debug(`Could not determine esmImportPath using import.meta.url or localNodeModules for ${name}`)
+            }
+        }
```
christian-bromann, line 194: We don't want to do `await resolve(name, import.meta.url)` as we don't want users to load WebdriverIO dependencies, when they should load deps from their project.

Also we can remove `require.resolve` calls as `import-meta-resolve` also supports CJS modules. So I'ld suggest to replace `require.resolve` with `await resolve(...)`.

RossVertizan: Agreed

RossVertizan:
@christian-bromann  - OK, I have updated the file in accordance with what we have discussed.  I have left your original comment.  Do we want to keep the remark about being able to link packages?

christian-bromann (approved):
LGTM 👍

christian-bromann:
@RossVertizan I resolved the conflict through the UI

christian-bromann:
I will merge and release once the tests pass.

PR #9354, opened 2022-12-08
feat: ability to send headers when connect to browser using puppeteer

DudaGod:
## Proposed changes

I have browsers pool in some cloud. I want that only users with access will be able to connect to them. So they must provide token through headers. But wdio does not send headers when connect to browser using puppeteer, because puppeteer didn't support it before. But from puppeteer-core@19.4.0 it became possible (PR in puppeteer - https://github.com/puppeteer/puppeteer/pull/9314).

Moreover I want to support it for webdriverio@7, but branch v7 - https://github.com/webdriverio/webdriverio/tree/v7 use puppeteer-core@13.1.3 - https://github.com/webdriverio/webdriverio/blob/v7/packages/webdriverio/package.json#L79. And I'm not sure that I can bump it to 19. What should I do with it?

### DIFF
--- a/packages/devtools/package.json
+++ b/packages/devtools/package.json
@@ -43,7 +43,7 @@
     "chrome-launcher": "^0.15.0",
     "edge-paths": "^2.1.0",
     "import-meta-resolve": "^2.1.0",
-    "puppeteer-core": "19.3.0",
+    "puppeteer-core": "19.4.0",
     "query-selector-shadow-dom": "^1.0.0",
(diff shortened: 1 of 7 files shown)


DudaGod:
> Can we update the docs to reflect that [headers](https://webdriver.io/docs/configuration#headers) are also used when connecting to Puppeteer using the devtools automation protocol?

Fixed. And can you answer for my question about webdriverio@7? Thank you.

christian-bromann (approved):
👍  Thanks!

PR #9395, opened 2022-12-13
@wdio/cli: kill child processes on SIGINT

### LINKED ISSUE #9393: [🐛 Bug]: interrupting cli test execution via ctrl + c does not appear to kill the session with V8

### WebdriverIO Version

8.0.11

### Node.js Version

18.12.0

### Mode

WDIO Testrunner

```typescript
{
  browserName: "chrome",
  browserVersion: 'latest',
  platformName: "Windows 10"
}
```

### What happened?

Tested on: macOS Monterey

When I run a spec via cli I want to stop execution during the test run with ctrl + c. However, with v8, test execution persists even after my attempt to kill the process.

I have run into this when running tests on my local server as well as on BrowserStack. Upon starting test execution I ran `ps -axl | grep wdio` and verified the processes began and upon my attempt to kill the test session I ran it again and saw the processes were not removed.

Test session should be removed and any running wdio process should be killed

### How to reproduce the bug.

Kill the process before the test can complete via ctrl + c (then compare to v7)

```js
await browser.pause(10000)
```

### Relevant log output

```typescript
no relevant logs
```

### Code of Conduct

SCG82:
## Proposed changes

Fixes #9393

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/wdio-cli/src/commands/run.ts
+++ b/packages/wdio-cli/src/commands/run.ts
@@ -200,7 +200,7 @@ export async function handler(argv: RunCommandArguments) {
             removeSignalListeners()
             if (tsProcess.pid) {
                 // gracefully kill child process
-                kill(tsProcess.pid, 'SIGINT')
+                process.kill(tsProcess.pid, 'SIGINT')
             }
         }
         const removeSignalListeners = () => {
--- a/packages/wdio-cli/src/commands/run.ts
+++ b/packages/wdio-cli/src/commands/run.ts
@@ -214,7 +214,6 @@ export async function handler(argv: RunCommandArguments) {
          * ensure process is killed according to result of new process
          */
         tsProcess.on('close', (code) => process.exit(code || 0))
-
         tsProcess.on('exit', removeSignalListeners)
         tsProcess.on('error', executeKill)
 
--- a/packages/wdio-cli/src/commands/run.ts
+++ b/packages/wdio-cli/src/commands/run.ts
@@ -1,6 +1,7 @@
 import path from 'node:path'
 import fs from 'node:fs/promises'
 import cp from 'node:child_process'
+import { kill } from 'node:process'
 import type { Argv } from 'yargs'
 
 import Launcher from '../launcher.js'
@@ -183,6 +184,7 @@ export async function handler(argv: RunCommandArguments) {
             path.join(path.dirname(wdioConf), 'tsconfig.json')
         )
         const hasLocalTSConfig = await fs.access(localTSConfigPath).then(() => true, () => false)
+
         const tsProcess = cp.spawn(nodePath, process.argv.slice(1), {
             cwd: process.cwd(),
             detached: true,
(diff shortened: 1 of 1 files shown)


packages/wdio-cli/src/commands/run.ts
```diff
@@ -194,10 +196,33 @@ export async function handler(argv: RunCommandArguments) {
             }
         })
 
+        const executeKill = () => {
+            removeSignalListeners()
+            if (tsProcess.pid) {
+                // gracefully kill child process
+                kill(tsProcess.pid, 'SIGINT')
```
christian-bromann, line 203: No need to import process implicitly if we don't do it for others either.
```suggestion
                process.kill(tsProcess.pid, 'SIGINT')
```

packages/wdio-cli/src/commands/run.ts
```diff
@@ -194,10 +196,33 @@ export async function handler(argv: RunCommandArguments) {
...
+            process.removeListener('SIGTERM', executeKill)
+            process.removeListener('exit', executeKill)
+            process.removeListener('beforeExit', executeKill)
+        }
+
         /**
          * ensure process is killed according to result of new process
          */
         tsProcess.on('close', (code) => process.exit(code || 0))
+
```
christian-bromann, line 217:
```suggestion
```

packages/wdio-cli/src/commands/run.ts
```diff
@@ -194,10 +196,33 @@ export async function handler(argv: RunCommandArguments) {
...
+        const executeKill = () => {
+            removeSignalListeners()
+            if (tsProcess.pid) {
+                // gracefully kill child process
+                kill(tsProcess.pid, 'SIGINT')
+            }
+        }
+        const removeSignalListeners = () => {
+            process.removeListener('SIGINT', executeKill)
+            process.removeListener('SIGTERM', executeKill)
```
christian-bromann, line 208: This would kill the TS process with `SIGINT` rather `SIGTERM` .. I am not sure if this matters a lot

christian-bromann (commented):
Some comments, but overall LGTM 👍

packages/wdio-cli/src/commands/run.ts
SCG82: I'm considering using async-exit-hook, or another exit-hook, but haven't found one that passes the exit code to the hook. Do you know of any?

christian-bromann: Maybe it is worth considering a different way to do the whole initial config parsing, maybe through a web-worker that passes down the config object. However this seems problematic as we can't serialize e.g. `onPrepare` hooks through that.

Compared to `v7` we are now running into this because compiling TS with `ts-node` in ESM environment requires passing in these `NODE_OPTIONS`.

SCG82: I ended up forking `exit-hook` and modified it to pass the exit code in the callback handler: https://www.npmjs.com/package/@scg82/exit-hook. Would you like me to use this here?

christian-bromann: What is the added value using your `exit-hook` fork? If the current solution solves the issue I would prefer that as it doesn't require us to maintain another dependency.

SCG82:
I agree with minimizing the use of dependencies. I can open a PR to add passing the exit code to the hook for the exit-hook-async package that’s already being used in wdio-cli.  The benefits to using exit-hook (or exit-hook-async) are minor in this case (unhandled promise rejection, pm2 cluster termination, and a couple additional signals). So if those are not needed/wanted here, then I can remove everything except the SIGINT listener, which will solve the Ctrl + c issue.On Dec 13, 2022, at 10:27 PM, Christian Bromann ***@***.***> wrote:﻿
@christian-bromann commented on this pull request.



In packages/wdio-cli/src/commands/run.ts:
> @@ -194,10 +196,33 @@ export async function handler(argv: RunCommandArguments) {
             }
         })
 
+        const executeKill = () => {
+            removeSignalListeners()
+            if (tsProcess.pid) {
+                // gracefully kill child process
+                kill(tsProcess.pid, 'SIGINT')
+            }
+        }
+        const removeSignalListeners = () => {
+            process.removeListener('SIGINT', executeKill)
+            process.removeListener('SIGTERM', executeKill)

What is the added value using your exit-hook fork? If the current solution solves the issue I would prefer that as it doesn't require us to maintain another dependency.

christian-bromann:
Actually let's work with https://www.npmjs.com/package/execa here, it seems to be what vitest is doing too.

christian-bromann:
What do you think of https://github.com/webdriverio/webdriverio/pull/9399

christian-bromann:
Let's re-open and decide which solution makes more sense.

christian-bromann:
Ok, I did some more testing and not only fixes `execa` the original issue but it also makes the issues we have with Windows go away. So I will go with that for now. Thanks anyway for looking into this!

SCG82:
> What do you think of https://github.com/webdriverio/webdriverio/pull/9399?

LGTM! I might do this for my projects as well.

PR #9496, opened 2022-12-31
fix workaround for Safari 12.0.3

kyryloonufriiev:
## Proposed changes

The issue is that when the element's `getRect` command is executed there is a Safari workaround to extract a rect from the browser executing js snippet. The workaround applied when one of the properties (x, y, width, or height) is a falsy (`!`) value, previously (in wdio 7) it was applied only if one of the properties were equal to null (`== null`). The problem happens when on appium you extract a region that has one of those properties `0`, it is a valid value, but js snippet execution fails everything on appium.

## Types of changes

I am replacing `!` with `=== undefined`.

- [ ] Documentation update

## Checklist

## Further comments

I didn't find any tests in this area which is why I don't know how you guys usually test things like that. Feel free to add a test or instruct me on what I should add.

### DIFF
--- a/packages/webdriverio/src/utils/index.ts
+++ b/packages/webdriverio/src/utils/index.ts
@@ -381,7 +381,7 @@ export async function getElementRect(scope: WebdriverIO.Element) {
      * getElementRect workaround for Safari 12.0.3
      * if one of [x, y, height, width] is undefined get rect with javascript
      */
-    if (Object.keys(defaults).some((key: keyof typeof defaults) => !rect[key])) {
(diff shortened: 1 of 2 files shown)


christian-bromann (commented):

christian-bromann:
ping @kyryloonufriiev

kyryloonufriiev:
@christian-bromann sorry, I was away last week. I did add a unit test, how does it look now?

christian-bromann (approved):
LGTM 👍

PR #9616, opened 2023-01-16
Update ConfigurationFile.md

TiKevin83:
## Proposed changes

Updated to reflect change from v8 announcement and the behavior we see locally in v8. The 'specs' option now refers to a relative path from the configuration file running instead of the previous behavior where it was relative to where 'wdio' was ran out of.

## Types of changes

- [x] Documentation update

## Checklist


christian-bromann:
Thanks for the PR @TiKevin83 , there were some more places with the same wording. I pushed a commit with all additional changes.

Issue #7932, opened 2022-01-17
[📖 Docs]: Limitations for nth selector using TypeScript does not work


a2f0:
### Pre-check

- [X] I know I can edit the docs but prefer to file this issue

### Describe the improvement

Unclear documentation

### Description of the improvement / report

It appears that the [documentation for limitations](https://webdriver.io/docs/autowait/#limitation) to select the nth element of an array using `waitUnitl` does not work when using TypeScript, or at least fails TypeScript validation because the `success` case returns an object of type `WebdriverIO.Element`  where waitUntil only accepts a function that returns ` boolean | Promise<boolean>`.  The error with the example verbatim is in the attached screenshot.

![Screen Shot 2022-01-17 at 8 49 56 AM](https://user-images.githubusercontent.com/1569853/149780792-49671b65-43ea-40ef-b7a6-a54904fc0880.png)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

christian-bromann:
Thanks for reporting!

I believe `waitUntil` returns the truthy value returned by the callback. The typings and docs should reflect that.

Any contributions that resolves the bug are highly appreciated. Please take a look into our [contribution guidelines](https://webdriver.io/docs/contribute) and let us know if you have any questions. Cheers!

harsha509:
HI @christian-bromann ,

Looks like the typing's are fine, as `waitUntil` should return a Boolean value.
```
waitUntil(this: WebdriverIO.Browser | WebdriverIO.Element, condition: () => boolean|Promise< boolean >, { timeout, interval, timeoutMsg }?: Partial<WaitUntilOptions>): Promise<true | void>;
```

Looks like code sample in docs  should be updated , like below
```
    await browser.waitUntil(
      async () => {
        const elems = await $$('div');
        if (elems.length !== 2) {
          return false
        }
        return elems[2].getText() === "Some text"
      }, {
        timeoutMsg: 'Never found enough div elements',
      })
```

Please  suggest, will give the PR for this in docs!

christian-bromann:
@harsha509 thanks for investigating. I think the types might be wrong here. I believe `waitUntil` is implemented so it returns the `truthy` value, e.g.:
```js
const value = await browser.waitUntil(() => 'foobar');
console.log(value); // outputs: "foobar"
```

Can you confirm? If so I suggest to not expect a boolean as return value but a generic.

nextlevelbeard:
[...]

PR #9767, opened 2023-02-14
Have the return type of waitUntil be the return type of the condition

### LINKED ISSUE #9679: [📖 Docs]: Use waitUntil with $$ element array fetch command, in the context of typescript, but failed to pass the typescript check
### Pre-check

- [X] I know I can edit the docs but prefer to file this issue

### Describe the improvement

Unclear documentation

### Description of the improvement / report

I found the below code snippet on the page: https://webdriver.io/docs/autowait#limitations

```
const div = await browser.waitUntil(async () => {
    const elems = await $$('div')
    if (elems.length !== 2) {
        return false
    }

    return elems[2]
}, {
    timeoutMsg: 'Never found enough div elements'
})
await div.click()
```

I thought that it means that the condition of waitUntil can return an element/elements array directly, so I implemented something like this in my code(android device automation): 

```
let elements = await driver.waitUntil(async () => {
	let elems = await $$(`//*[contains(@resource-id, "${id}")]`);
	if (elems.length < 1) {
		return false;
	}
	return elems; // here I return an element array
});
return elements;
```

but the typescript check said this: 

Argument of type '() => Promise<false | ElementArray>' is not assignable to parameter of type '() => boolean | Promise<boolean>'.
  Type 'Promise<false | ElementArray>' is not assignable to type 'boolean | Promise<boolean>'.
    Type 'Promise<false | ElementArray>' is not assignable to type 'Promise<boolean>'.
      Type 'false | ElementArray' is not assignable to type 'boolean'.
        Type 'ElementArray' is not assignable to type 'boolean'.ts(2345)

In the meanwhile, the functionality works without error. It seems that the condition callback should only return boolean values. Is the code snippet I found in the docs legitimate? What should I do to fulfill the checks?  

### Code of Conduct

christian-bromann:
## Proposed changes

fixes #9679 #7932

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

n/a

### DIFF
--- a/packages/webdriverio/src/commands/browser/waitUntil.ts
+++ b/packages/webdriverio/src/commands/browser/waitUntil.ts
@@ -2,9 +2,8 @@ import Timer from '../../utils/Timer.js'
 import type { WaitUntilOptions } from '../../types.js'
 
 /**
- *
  * This wait command is your universal weapon if you want to wait on something. It expects a condition
- * and waits until that condition is fulfilled with a truthy value.
+ * and waits until that condition is fulfilled with a truthy value to be returned.
  *
  * A common example is to wait until a certain element contains a certain text (see example).
  *
@@ -31,7 +30,7 @@ import type { WaitUntilOptions } from '../../types.js'
  *
  *
  * @alias browser.waitUntil
- * @param {Function#Boolean}  condition  condition to wait on
+ * @param {Function}          condition  condition to wait on until returning a truthy value
  * @param {WaitUntilOptions=} options    command options
  * @param {Number=}           options.timeout     timeout in ms (default: 5000)
  * @param {String=}           options.timeoutMsg  error message to throw when waitUntil times out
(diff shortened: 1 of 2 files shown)


nextlevelbeard:
@christian-bromann I was thinking it probably makes sense not only to exclude `false`, but all falsy values:
`null`, `undefined`, `""`, `0`, `0n`, `NaN`, `-0`, and finally `false`

Here is an example that is not excluding 0 when it should:

```ts
const waitUntilElems = await browser.waitUntil(async () => {
        const elems = await $$('elems')
        // Here we only ever return an ElementArray
        return elems.length && elems
    })
    // This should be expected
    expectType<WebdriverIO.ElementArray>(waitUntilElems)
    // What is actually expected: Promise<0 | ElementArray>
```

christian-bromann:
> Here is an example that is not excluding 0 when it should:

Doesn't this work already?

> Fix could be to make this change:

Mind raising a PR?

PR #9771, opened 2023-02-17
#9539 Support various configurations of CJS/ESM with(out) TS

### LINKED ISSUE #9539: [🐛 Bug]: ts page class is imported as .js and failing test when updating the import

### WebdriverIO Version

8.0.13

### Node.js Version

18

### Mode

WDIO Testrunner

```typescript
{
    browserName: 'chrome'
    browserVersion: 'latest',
    platform: 'macOS'
}
```

### What happened?

i am using typescript with webdriverio and page objects files are also in .ts
In actual test import for page are written as .js which should not be the case .

When I remove the file extension (same approach as v7) and run then i get error 

page import statement should not have file extension or based on actual file and test should run successfully.

### How to reproduce the bug.

git clone https://github.com/testSadab/WebdriverIOMochaTS.git
git checkout import_issue
npm install
npm run wdio

### Relevant log output

```typescript
git link: https://github.com/testSadab/WebdriverIOMochaTS/tree/import_issue
logs: https://github.com/testSadab/WebdriverIOMochaTS/blob/import_issue/log.txt
```

### Code of Conduct

Jmcosel:
## Proposed changes

This fixes #9539 - the configuration wizard for WebdriverIO has many different paths you can go, and with the switch to defaulting to ESM for the module system over CJS, there were use cases in which the generated test files and configuration were not actually working. This PR aims to solve all of those.

## Types of changes

- [ ] Documentation update

## Checklist
- [] I have added proper type definitions for new commands (if appropriate)

## Further comments

There is a weird bug I don't know how to solve where if a pure ESM project is generated using the `wdio` command (and it detecting there's no configuration so it goes to run that first before returning to the run mode), it will fail with an error like so: `ERROR @wdio/config:ConfigParser: Failed loading configuration file: file:///path/to/wdio.conf.js: Unexpected token 'export'`
The strange part is subsequent runs of the same command, skipping the config wizard this time, work just fine. Additionally, using TypeScript as a compiler also works fully, even if you haven't set up the project yet.

I'm also aware that trying to run after the configuration wizard will have problems, but only if you have both a `package.json` and `tsconfig.json` present, and they are mismatched with the module systems. I feel like at that point it's the user's responsibility to fix that since I don't really want to touch files that were already made.


Jmcosel:
@christian-bromann I did my best to try and find all the use cases, can you double check my work? This was much harder than it looks ;D

packages/wdio-cli/src/commands/run.ts
```diff
@@ -146,23 +146,22 @@ export async function launch(wdioConfPath: string, params: Partial<RunCommandArg
...
-    const canAccessConfigPath = await fs.access(configPath).then(
+    const canAccessConfigPath = await fs.access(`${configPathNoExtension}.js`).then(
         () => true,
-        () => false
+        () => fs.access(`${configPathNoExtension}.ts`).then(
+            () => true, () => false
+        )
     )
+    const wdioConf = configPath.includes(process.cwd())
+        ? configPath : path.join(process.cwd(), configPath)
```
christian-bromann, line 159:
```suggestion
        ? configPath
        : path.join(process.cwd(), configPath)
```

packages/wdio-cli/src/commands/run.ts
```diff
@@ -146,23 +146,22 @@ export async function launch(wdioConfPath: string, params: Partial<RunCommandArg
...
 
-    const canAccessConfigPath = await fs.access(configPath).then(
+    const canAccessConfigPath = await fs.access(`${configPathNoExtension}.js`).then(
         () => true,
-        () => false
+        () => fs.access(`${configPathNoExtension}.ts`).then(
+            () => true, () => false
+        )
     )
+    const wdioConf = configPath.includes(process.cwd())
```
christian-bromann, line 158: This should be moved to the top, if `canAccessConfigPath` fails for e.g. a `wdio.conf.js` but passes for a `wdio.conf.ts` then we continue here with `wdio.conf.js`

packages/wdio-cli/src/utils.ts
```diff
@@ -611,21 +601,36 @@ export function runProgram (command: string, args: string[], options: SpawnOptio
  * create package.json if not already existing
  */
 export async function createPackageJSON (parsedAnswers: ParsedAnswers) {
-    if (!parsedAnswers.createPackageJSON) {
-        return
+    const packageJsonExists = await fs.access(path.resolve(process.cwd(), 'package.json')).then(() => true, () => false)
+
+    // Use the exisitng package.json if it already exists.
+    if (packageJsonExists) { return }
```
christian-bromann, line 607:
```suggestion
    if (packageJsonExists) {
        return
    }
```

packages/wdio-cli/src/utils.ts
```diff
@@ -611,21 +601,36 @@ export function runProgram (command: string, args: string[], options: SpawnOptio
...
+    // If a user said no to creating a package.json, but it doesn't exist, abort.
+    if (parsedAnswers.createPackageJSON === false) {
+        if (!packageJsonExists) {
+            /* istanbul ignore next */
+            console.log(`No WebdriverIO configuration found in "${parsedAnswers.wdioConfigPath}"`)
+
+            /* istanbul ignore next */
+            return !process.env.VITEST_WORKER_ID && process.exit(0)
+        } return
+    }
```
christian-bromann, line 618:
```suggestion
    // If a user said no to creating a package.json, but it doesn't exist, abort.
    if (parsedAnswers.createPackageJSON === false) {
        /* istanbul ignore if */
        if (!packageJsonExists) {
            console.log(`No WebdriverIO configuration found in "${parsedAnswers.wdioConfigPath}"`)
            return !process.env.VITEST_WORKER_ID && process.exit(0)
        }
        
        return
    }
```

christian-bromann (commented):
Thanks for raising the PR, added some comments

packages/wdio-cli/src/commands/config.ts
```diff
@@ -80,10 +80,10 @@ export const parseAnswers = async function (yes: boolean): Promise<ParsedAnswers
          */
         : answers.specs
             /**
-             * if we have autogenerated spec files, put the tsconfig one above the spec file dir
+             * if we have autogenerated spec files, put the tsconfig two levels above the spec file dir
```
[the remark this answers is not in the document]
Jmcosel: I thought about this and you're right, I should have just put it at the project root directory.

packages/wdio-cli/src/commands/run.ts
Jmcosel: This wasn't formatted for me automatically and ESLint doesn't have a problem with it.
Are the EditorConfig rules supposed to do that?

christian-bromann: There is EsLint rule for that, I can work on adding one to make this more obvious.

packages/wdio-cli/src/commands/config.ts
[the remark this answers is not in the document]
Jmcosel: Honestly I don't really understand the use case here, I don't know why we would need to make a second tsconfig file if spec files are made within the project. I'm going to guess the intention was not to overwrite a tsconfig file for a larger project if they want their e2e-tests to be a subfolder/subproject within it... should I just remove these conditions?

packages/wdio-cli/src/utils.ts
Jmcosel: This wasn't formatted for me automatically and ESLint doesn't have a problem with it.
Are the EditorConfig rules supposed to do that?

packages/wdio-cli/src/commands/config.ts
```diff
@@ -167,7 +168,7 @@ export async function handler(argv: ConfigCommandArguments, runConfigCmd = runCo
  */
 export async function missingConfigurationPrompt(command: string, message: string, useYarn = false, runConfigCmd = runConfigCommand) {
     const configMessage = command === 'run'
-        ? `Error: Could not execute "run" due to missing configuration, file "${message}" not found! Would you like to create one?`
+        ? `Error: Could not execute "run" due to missing configuration, file "${path.parse(message).name}[.js/.ts]" not found! Would you like to create one?`
```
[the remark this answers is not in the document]
Jmcosel: Good catch, thank you.

Jmcosel:
@christian-bromann Rebased and resolved _almost_ all your requests.
I have that outstanding question on whether that conditional code is useful or if I should just remove it.

packages/wdio-cli/src/commands/config.ts
```diff
@@ -158,22 +159,53 @@ export async function handler(argv: ConfigCommandArguments, runConfigCmd = runCo
...
  * @param {Function} runConfigCmd   runConfig method to be replaceable for unit testing
  */
-export async function missingConfigurationPrompt(command: string, message: string, useYarn = false, runConfigCmd = runConfigCommand) {
-    const configMessage = command === 'run'
-        ? `Error: Could not execute "run" due to missing configuration, file "${message}" not found! Would you like to create one?`
-        : `Error: Could not execute "${command}" due to missing configuration. Would you like to create one?`
+export async function missingConfigurationPrompt(command: string, configPath: string, useYarn = false, runConfigCmd = runConfigCommand) {
+
+    const message = (
+        `Could not execute "${command}" due to missing configuration, file` +
```
Jmcosel, line 200: Oops I missed a space here, will correct after we address all other outstanding issues.

packages/wdio-cli/src/commands/config.ts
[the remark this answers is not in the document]
christian-bromann: Yes, I think having a separated tsconfig file makes sense as the type setup might be different when writing tests.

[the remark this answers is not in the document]
Jmcosel: Ok I put it back the way it was beofre.

Jmcosel:
@christian-bromann Made final changes. Builds above are immediately failing, not sure why?

christian-bromann:
@Jmcosel seems like an issue with `package-lock.json` due to dependency updates. Can you please update your branch?

Jmcosel:
@christian-bromann Great, ty, should be good to go now.

christian-bromann (approved):
Great job, LGTM 👍

Issue #9805, opened 2023-02-21
[🐛 Bug]: Scroll Into View issues with scrolling to an element


kevin-longmate:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

8.3.9

### Node.js Version

18.12.1

### Mode

WDIO Testrunner

### Which capabilities are you using?

```typescript
This is happening on all capabilities I've tried - chrome, firefox and safari latest and older versions through browserstack so don't believe it matters
```

### What happened?

We have recently updated from version 7.26 to 8.3.9 and noticed a number of issues where our tests are scrolling elements in to view

When scrolling in to view an element on a page, when the element is near the top of the page, or above the currently displayed part of the website, webdriver appears to scroll down taking the element out of view instead.

See attached video and code to reproduce below. There are two scrolls in each test, the element used in all cases is 
`//h3[text()="Get Started With WebdriverIO within Seconds"]`

The first test, I scroll to it normally and it works (although the scroll puts it behind the navbar at the top of the page). Then I scroll with the options `{block: "center"}`, which should put the element in the middle of the page (scrolling up from where we are) however it instead scrolls down so that the "Watch talks about webdriver io" is displayed at the top of the page, which is below the element I want. The element I want is then further out of view above the visible part of the page.

The page reloads for the second test and I use scrollIntoView with `{block: "center"}` first, which works ok and the element I want is in the middle of the page. I then scroll again to that element without using `{block: "center"}` and once again it sends is up above the visible area of the page so that I'm looking at the "Watch talks about webdriver io" bit, with the element I want to be in view well above the top of the page.

Video taken from browserstack
https://user-images.githubusercontent.com/46672685/220308880-15a032c5-26f7-48a4-867c-f0c697e5d2ba.mp4

### What is your expected behavior?

When calling scroll in to view with any options, the element should be scroll in to view correctly. Using this without any options should have it show at the top of the page, using it with `{block: "center"}` should show it in the middle of the page

### How to reproduce the bug.

describe("scrolling tests", () => {
it("scroll testing 1", async () => {
    await browser.url("https://webdriver.io/");
    await browser.maximizeWindow();
    await browser.pause(5000);
    const element = await $(
      "//h3[text()="Get Started With WebdriverIO within Seconds"]"
    );
[...]

PR #9820, opened 2023-02-24
update for scroll

kevin-longmate:
## Proposed changes

Fix for https://github.com/webdriverio/webdriverio/issues/9805
Scrolling was ok as long as the element location hadn't changed before (that is, a previous scroll had taken place as part of the same test). The second (or subsequent) scrolls were getting correct distance to scroll but were starting from the location of the element that had already moved, so it was scrolling down more than necessary.

Changing the `origin` of the scroll to be `browser` rather than `this` corrects this issue.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/webdriverio/src/commands/element/scrollIntoView.ts
+++ b/packages/webdriverio/src/commands/element/scrollIntoView.ts
@@ -79,7 +79,7 @@ export async function scrollIntoView (
 
     try {
         return await browser.action('wheel')
-            .scroll({ origin: this, duration: 200, deltaY, deltaX })
+            .scroll({ origin: browser, duration: 200, deltaY, deltaX })
             .perform()
     } catch (err: any) {
         log.warn(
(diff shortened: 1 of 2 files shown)


christian-bromann (changes requested):
I don't think this is a solution to the problem. `origin` takes an element and `browser` is not an element. I assume this just creates the effect of not applying the origin at all.

kevin-longmate:
I've not been able to find any alternative that works for this unfortunately 😞 using `browser` doesn't stop the origin being applied, if origin is removed completely then the scroll into view function scrolls differently (but still incorrectly) than it does when `browser` is in there, so `origin: browser` is doing something.

I can't work out what `origin: element` (or `origin: browser` honestly) actually does for this scroll so I think this might need someone more familiar with the code than I am to work on it.
I can say that the `deltaX` and `deltaY` seem to be getting the correct numbers to scroll the element in to view and the `x: number` and `y: number` have always been `0` when I try, which I think seems right as when scrolling from the top of the page the scroll is correct.

For some reason, though, the scroll doesn't seem to be happening from a starting point of 0,0 after already having scrolled something else in to view, which is where it's breaking.

Hope that helps someone look at it

PR #9874, opened 2023-03-01
fix(commands): successive scrollIntoView

OBe95:
## Proposed changes

Fixes [#9805](https://github.com/webdriverio/webdriverio/issues/9805)
Subsequent `scrollIntoView` on a given element, is wrongly scrolling the page
It seems that we have 2 issues:
- the `origin` used when scrolling is different than the one used when computing the deltas
- the `deltaX` seems to be wrongly calculated (using `height` instead of `width`)

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments


packages/webdriverio/src/commands/element/scrollIntoView.ts
```diff
@@ -56,19 +56,20 @@ export async function scrollIntoView (
...
      */
     if (typeof options === 'boolean' || typeof options.block === 'string' || typeof options.inline === 'string') {
-        const htmlElem = await browser.$('html')
-        const viewport = await htmlElem.getSize()
+        origin = await browser.$('html')
+        const viewport = await origin.getSize()
         const elemSize = await this.getSize()
         if (options === true || (options as ScrollIntoViewOptions).block === 'start') {
-            deltaY += viewport.height - elemSize.height
+            deltaY += viewport.width - elemSize.width
```
christian-bromann, line 70: Can you explain how to determine the vertical delta when looking at the `wdith`of the element?

christian-bromann (commented):
I wonder if we can have any e2e tests for this, please see the `./e2e` folder, so we can make sure this doesn't cause any regression.

packages/webdriverio/src/commands/element/scrollIntoView.ts
OBe95: Unless I'm missing something, but from my understanding the vertical delta is computed in the `deltaX` variable in the following lines

Here it was computing the horizontal delta but based on the `height` which doesn't make sense

OBe95:
Not sure if it's an automatic message regarding e2e, but I did cover it by scrolling once and getting the `window` scroll state, then doing successive scrolling and checking that the `window` scroll state matches the first one (assuming here that one scroll works just fine)

packages/webdriverio/src/commands/element/scrollIntoView.ts
OBe95: My bad, I meant to update the `deltaX`
They were both based the `height`, I will push an update soon

christian-bromann:
I would suggest the following method of testing this:

- run `elem.scrollIntoView()`
- get scroll position
- run web API: `browser.execute((elem) => elem.scrollIntoView(), [elem])`
- compare if both are the same

I think this better tells us whether our command behaves the same way as the web API.

OBe95:
@christian-bromann I pushed an update to handle all options, add e2es and update the snapshots for existing unit tests

christian-bromann (approved):
This is awesome, thanks so much 👍

christian-bromann:
> ** Test / e2e-windows (pull_request) ** Failing after 25m
> ** Test / e2e-mac (pull_request) ** Failing after 9m

Seems both fail by `5px` different, I will go ahead and merge this and fix the test failure in `main`

niklasschaeffer:
This fix has been leading to an weird debug message

```
ERROR webdriver: Request failed with status 400 due to invalid argument: invalid argument
[0-0] from invalid argument: 'delta x' must be an int
[0-0]   (Session info: chrome=111.0.5563.64)
```

km9134:
Has anyone noticed any issues when using the block parameter in this function? When i call `scrollIntoView({ behavior: 'smooth',  block: 'top' })` I am presented with an error `TypeError: Cannot read properties of undefined (reading 'y')`

Notably if i use `scrollIntoView({ behavior: 'smooth',  block: 'center' })` i am not presented with any issues

christian-bromann:
@niklasschaeffer thanks for reporting, mind raising a PR with changes that would round the `deltaX` and `deltaY` values?

OBe95:
Let me know @niklasschaeffer if you want me to take care of it 😄

niklasschaeffer:
@OBe95 @christian-bromann there should be a pr now. dont know if it actually fixes the problem :D

PR #9917, opened 2023-03-09
check for mocha in hook

nurdtechie98:
For certain user defined hooks like beforeAll and afterAll in cucumber, the test argument being passed to beforeHook and afterHook is undefined. Leading to exception

```
[0-0] 2023-03-06T17:13:57.110Z ERROR @wdio/utils:shim: TypeError: Cannot read properties of undefined (reading 'parent')
[0-0]     at InsightsHandler.beforeHook (/Users/sean.darley/projects/dig-navigation-ui-search/test/node_modules/@wdio/browserstack-service/build/insights-handler.js:39:35)
```

## Proposed changes
As a part of this PR we are checking for framework mocha explicitly for which the above logic works fine, and skip sending data for other frameworks. We'll tackle other frameworks on a case by case basis and unblock users from running cucumber test for now.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/wdio-browserstack-service/src/insights-handler.ts
+++ b/packages/wdio-browserstack-service/src/insights-handler.ts
@@ -52,28 +52,28 @@ export default class InsightsHandler {
     }
 
     async beforeHook (test: Frameworks.Test, context: any) {
-        const fullTitle = `${test.parent} - ${test.title}`
-        const hookId = uuidv4()
-        this._tests[fullTitle] = {
-            uuid: hookId,
-            startedAt: (new Date()).toISOString()
-        }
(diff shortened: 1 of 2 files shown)


christian-bromann:
> We'll tackle other frameworks on a case by case basis and unblock users from running cucumber test for now.

The goal should it be to ensure that the structure of these hook arguments is the same for all frameworks (Mocha, Jasmine and Cucumber) so that building reporters and services is simple.

nurdtechie98:
> The goal should it be to ensure that the structure of these hook arguments is the same for all frameworks
Definitely! that would be the most ideal scenario

Issue #9886, opened 2023-03-05
[🐛 Bug]: browser.getLogs is not a function


ShootingKing-AM:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

latest

### Node.js Version

16.13.2

### Mode

Standalone Mode

### Which capabilities are you using?

```typescript
capabilities: [{
    
        // maxInstances can get overwritten per capability. So if you have an in-house Selenium
        // grid with only 5 firefox instances available you can make sure that not more than
        // 5 instances get started at a time.
        maxInstances: 5,
        //
        browserName: 'chrome',
        acceptInsecureCerts: true,
        // 'goog:chromeOptions': {
        //     args: [
        //         '--no-sandbox',
        //         '--disable-infobars',
        //         '--headless',
        //         '--disable-gpu',
        //         '--window-size=1440,735'
        //     ],
        // }
        'goog:chromeOptions': {
            // binary: '/squashfs-root/obsidian', // Path to your Electron binary
            args: [
                '--no-sandbox',
                '--disable-dev-shm-usage',
                '--disable-gpu',
                '--disable-software-rasterizer',
                '--window-size=1440,735'
            ],
            debuggerAddress: '127.0.0.1:8888'
            // args: [/* cli arguments */] // Optional, perhaps 'app=' + /path/to/your/app/
        }
        // If outputDir is provided WebdriverIO can capture driver session logs
        // it is possible to configure which logTypes to include/exclude.
        // excludeDriverLogs: ['*'], // pass '*' to exclude all driver session logs
        // excludeDriverLogs: ['bugreport', 'server'],
    }],
```

### What happened?

```sh
$ npm run wdio

> obsidian-sample-plugin@0.9.7 wdio
> wdio run ./wdio.conf.ts


Execution of 1 workers started at 2023-03-05T13:25:44.841Z

2023-03-05T13:25:44.903Z INFO chromedriver: Initiate Chromedriver Launcher (v8.1.1)
2023-03-05T13:25:44.904Z INFO @wdio/cli:launcher: Run onPrepare hook
[...]

PR #9929, opened 2023-03-10
[Docs] Remove ambiguity and simplify docs

ShootingKing-AM:
## Proposed changes

Remove ambiguity and simplify docs

## Types of changes

- [x] Documentation update

## Checklist

## Further comments

- Simplified docs, remove "browser automation session" -> "chrome session"
- WebdriverIO has to be referred as WDIO instead, since Webdriver has many meanins, driver, communication protocol, automation protocol and only last two letters modify the meaning.
- Removed implicit pre knowledge requirement as to what "Appium" means, which is not required for the purposes of these docs, simple comparison between chrome and Appium is used as example
- used Chromium protocol example instead of ambiguous Webdriver example (can refer to again, W3C standard, this package, driver, protocol etc.. etc..)

Linked: #9886

### DIFF
--- a/website/docs/api/Browser.md
+++ b/website/docs/api/Browser.md
@@ -9,9 +9,9 @@ The browser object is the session instance you use to control the browser or mob
 
 The session is initialized by the test runner. The same goes for ending the session. This is also done by the test runner process.
 
-The browser object provides all methods from the [communication protocol](Protocols.md) section, e.g. [Chromium](/docs/api/chromium) communication protocol (if you are using Chromedriver) as well as commands listed within this section. Available protocol commands depend on the type of session. If you run an automated browser session, none of the Appium [commands](/docs/api/appium) will be available on the browser object and vice versa.
+The browser object provides all methods from the [communication protocol](Protocols.md) section, e.g. [Chromium](/docs/api/chromium) communication protocol (if you are using Chromedriver) as well as methods listed within this section. Available methods depend on the type of session. If you run an Chrome test session, none of the Appium [methods](/docs/api/appium) will be available on the browser object and vice versa.
 
-Besides all commands from the [API](API.md), the `browser` object provides some more information you might be interested in during your test run:
+Besides all methods from the [API](API.md), the `browser` object provides some more information you might be interested in during your test run:
 
 ## Properties
 
--- a/website/docs/api/Browser.md
+++ b/website/docs/api/Browser.md
@@ -9,6 +9,8 @@ The browser object is the session instance you use to control the browser or mob
 
 The session is initialized by the test runner. The same goes for ending the session. This is also done by the test runner process.
 
+The browser object provides all methods from the [communication protocol](Protocols.md) section, e.g. [Chromium](/docs/api/chromium) communication protocol (if you are using Chromedriver) as well as commands listed within this section. Available protocol commands depend on the type of session. If you run an automated browser session, none of the Appium [commands](/docs/api/appium) will be available on the browser object and vice versa.
+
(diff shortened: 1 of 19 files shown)


christian-bromann:
Thanks for putting up the PR, mind signing the CLA?

ShootingKing-AM:
Omg! 
Done.

website/docs/api/Browser.md
```diff
@@ -9,7 +9,9 @@ The browser object is the session instance you use to control the browser or mob
 
 The session is initialized by the test runner. The same goes for ending the session. This is also done by the test runner process.
 
-Besides all commands from the [API](API.md), the `browser` object provides some more information you might be interested in during your test run:
+The browser object provides all methods from the [communication protocol](Protocols.md) section, e.g. [Chromium](/docs/api/chromium) communication protocol (if you are using Chromedriver) as well as methods listed within this section. Available methods depend on the type of session. If you run an Chrome test session, none of the Appium [methods](/docs/api/appium) will be available on the browser object and vice versa.
```
christian-bromann, line 12: I don't think highlighting Chromium makes sense here. Can we paraphrase this more general, saying that "the browser object provides [protocol methods](/docs/api/protocols) based on the detected WebDriver backend".

website/docs/api/Protocols.md
```diff
@@ -1,24 +1,24 @@
 ---
 id: protocols
-title: Protocol Commands
+title: Communication Protocols
 ---
 
-WebdriverIO is an automation framework that relies on various of automation protocols to control a remote agent, e.g. a browser, mobile device or television. Based on the remote device different protocols come into play. These commands are assigned to the [Browser](/docs/api/browser) or [Element](/docs/api/element) Object depending on the session information by the remote server (e.g. browser driver).
+WebdriverIO (WDIO) is an automation framework that relies on various internal protocols to control a remote agent, for e.g. a browser, mobile device or television. Based on the remote device different protocols come into play. These commands are assigned to the [Browser](/docs/api/browser) or [Element](/docs/api/element) Object depending on the session information by the remote agent's server (e.g. browser driver) automatically.
```
christian-bromann, line 6: > various internal protocols

I don't think these protocols are internal. The WebDriver protocol is a web standard so not internal at all.

website/docs/api/Browser.md
ShootingKing-AM: I don't have particular interest with Chromium either. But the package name is WebdriverIO, the example Webdriver, which are confusing, hence since its just an example, we can take anything other than "webdriver" to make it simpler and easier for quick documentation reading. Please note people will not read all letters of the word to comprehend the word.

Chromium is common knowledge, and webdriver backend etc., again will require some pre-requisite knowledge, that its a std, protocol, etc.etc. Chromium is understood as browser, for understanding webdriver users have to reasearch what is webdriver, etc. etc. which is not essential to use WDIO package.

The majority users are here to just install the WDIO and run tests, they hardly (and rightly) do not care about the technologies that WDIO is using to accomplish its tasks.

I understand you are your intention of being accurate in documenting the code, but its not required for majority audiences to know what they don't need to know for just running a test. (WDIO should automatically take care of the rest, and it does). Majority audience are here to just use the package as testing facilitator.

These are my inputs, please do feel free to force changes or reject the PR.

website/docs/api/Protocols.md
ShootingKing-AM: Internal in the sense under the hood. Testing developers the main audience of the documentation would not need to have any idea of Webdriver or other protocols that WDIO uses. WDIO is handling it from them. They just need to install WDIO, config it, use WDIO methods to get the tests done imo

Issue #10010, opened 2023-03-20
[💡 Feature]: shared-store atomic operations


pedrorfernandes:
### Is your feature request related to a problem?

I have a test scenario where I need to access specific url's, e.g.: `['url01.com', 'url02.com', 'url03.com']`

The wdio workers should share these URL's and never clash. Once a worker is finished, it "frees" up that url for another worker.

Possibly a better way to handle this is through network/load balancing, however I wanted to try this with `shared-data` service, setting up the url sharing between workers:

```javascript
// worker A
await getValue("*"); // { 'url01.com': true, 'url02.com': true } booleans represent isUrlAvailable

// worker B
await getValue("*"); // { 'url01.com': true, 'url02.com': true }

// worker A chooses `url01.com`
await setValue("url01.com", false)
// worker B chooses `url01.com`
await setValue("url01.com", false)

// both A + B workers now have the same URL
// clash !
```

This is basically a shared queue management problem between node worker threads

### Describe the solution you'd like.

I could implement in `shared-store` another server operation `'/compareAndSet'` to avoid collisions when writing data

`compareAndSet(key, expectedValue, newValue)` only sets `newValue` if `expectedValue` is the current value in the store

```javascript
// worker A
await getValue("*"); // { 'url01.com': true, 'url02.com': true }

// worker B
await getValue("*"); // { 'url01.com': true, 'url02.com': true }

// worker A chooses `url01.com`
await compareAndSet("url01.com", true, false)
// worker B chooses `url01.com`
await compareAndSet("url01.com", true, false) // throws error, 'url01.com' is not `=== true` anymore
// worker B must retry another value
```

### Describe alternatives you've considered.

I thought about a different service where we just configure a queue of resources to be shared `[1, 2, 3, 4]` and the operations would just be `/take` a resource and `/return` a resource, this is easier on the `beforeSession` for each worker since the server would automatically dispense the value the worker needs and it doesn't need to retry until a value is free

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

christian-bromann:
Hey @pedrorfernandes, thanks for filing the issue.

This seems indeed like useful idea. We would appreciate any help on this so please help us out! Have a look at our [contributing guidelines](https://github.com/webdriverio/webdriverio/blob/master/CONTRIBUTING.md) and reach out if you have any questions. Cheers!

Ajneesh21:
Hey looks intresting and I would like to try this out if I am assigned this issue.

[...]

PR #10029, opened 2023-03-22
shared store resource pool #10010

pedrorfernandes:
Solves #10010

## Proposed changes

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/wdio-shared-store-service/README.md
+++ b/packages/wdio-shared-store-service/README.md
@@ -72,7 +72,7 @@ If the worker threads are competing for resources that must be assigned for each
 
 ```js
 // wdio.conf.js
-import { setResourcePool, getValueFromPool, addValueToPool} from '@wdio/shared-store-service'
+import { setResourcePool, getValueFromPool, addValueToPool } from '@wdio/shared-store-service'
 
 export const config = {
     maxInstances: 2,
--- a/packages/wdio-shared-store-service/README.md
+++ b/packages/wdio-shared-store-service/README.md
@@ -66,6 +66,33 @@ IMPORTANT! Every spec file should be atomic and isolated from others' specs.
 The idea of the service is to deal with very specific environment setup issues.
 Please avoid sharing test execution data!
 
+### Resource Pools
+
+If the worker threads are competing for resources that must be assigned for each worker, you can use Resource Pool API:
+
+```js
+// wdio.conf.js
+import { setResourcePool, getValueFromPool, addValueToPool} from '@wdio/shared-store-service'
+
+export const config = {
+    maxInstances: 2,
+    // ...
+    onPrepare: async function (config, capabilities) {
+        await setResourcePool('availableUrls', ['url01.com', 'url02.com'])
+    },
+    // ...
+    beforeSession: async (conf) => {
+        conf.baseUrl = await getValueFromPool('availableUrls');
+    },
+    // ...
+    afterSession: async (conf) => {
+        // worker returns the used resource for next workers to use
+        await addValueToPool('availableUrls', conf.baseUrl);
+    }
+```
+
+This example ensures that both workers never use the same `baseUrl`. A unique url is only assigned to one worker until it's released by it.
+
 ## Configuration
 
 Add `shared-store` to the services list and the `sharedStore` object will be accessible to you on the [`browser` scope](https://webdriver.io/docs/api/browser) in your test.
--- a/packages/wdio-shared-store-service/src/client.ts
+++ b/packages/wdio-shared-store-service/src/client.ts
@@ -3,6 +3,7 @@ import got from 'got'
 import logger from '@wdio/logger'
 
 import type { JsonCompatible, JsonPrimitive, JsonObject, JsonArray } from '@wdio/types'
+import type { GetValueOptions } from 'src'
 
 const log = logger('@wdio/shared-store-service')
 
@@ -64,9 +65,9 @@ export const setResourcePool = async (key: string, value: JsonArray) => {
  * @param {string}  key
  * @param {*}       value
  */
-export const getValueFromPool = async (key: string) => {
+export const getValueFromPool = async (key: string, options: GetValueOptions) => {
     const baseUrl = await baseUrlPromise
-    const res = await got.get(`${baseUrl}/getValueFromPool`, { json: { key }, responseType: 'json' }).catch(errHandler)
+    const res = await got.get(`${baseUrl}/getValueFromPool/${key}${options?.timeout ? `?timeout=${options.timeout}` : '' }`, { responseType: 'json' }).catch(errHandler)
     return res?.body ? (res.body as JsonObject).value : undefined
 }
 
--- a/packages/wdio-shared-store-service/src/client.ts
+++ b/packages/wdio-shared-store-service/src/client.ts
@@ -64,9 +64,9 @@ export const setResourcePool = async (key: string, value: JsonArray) => {
  * @param {string}  key
  * @param {*}       value
  */
-export const takeValueFromPool = async (key: string) => {
+export const getValueFromPool = async (key: string) => {
     const baseUrl = await baseUrlPromise
-    const res = await got.post(`${baseUrl}/takeValueFromPool`, { json: { key }, responseType: 'json' }).catch(errHandler)
+    const res = await got.get(`${baseUrl}/getValueFromPool`, { json: { key }, responseType: 'json' }).catch(errHandler)
     return res?.body ? (res.body as JsonObject).value : undefined
 }
 
--- a/packages/wdio-shared-store-service/src/client.ts
+++ b/packages/wdio-shared-store-service/src/client.ts
@@ -1,11 +1,8 @@
-import type { Response } from 'got'
+import type { RequestError } from 'got'
 import got from 'got'
-import logger from '@wdio/logger'
 
 import type { JsonCompatible, JsonPrimitive, JsonObject, JsonArray } from '@wdio/types'
 import type { GetValueOptions } from 'src'
-
-const log = logger('@wdio/shared-store-service')
 
 let baseUrlResolve: Parameters<ConstructorParameters<typeof Promise>[0]>[0]
 const baseUrlPromise = new Promise<string>((resolve) => {
@@ -82,6 +79,6 @@ export const addValueToPool = async (key: string, value: JsonPrimitive | JsonCom
     return res?.body ? (res.body as JsonObject).value : undefined
 }
 
-const errHandler = (err: Response<Error>) => {
-    log.warn(err.statusCode, err.statusMessage, err.url, err.body)
+const errHandler = (err: RequestError) => {
+    throw new Error(`${err.response?.body || 'Shared store server threw an error'}`)
 }
--- a/packages/wdio-shared-store-service/src/client.ts
+++ b/packages/wdio-shared-store-service/src/client.ts
@@ -6,12 +6,21 @@ import type { JsonCompatible, JsonPrimitive, JsonObject, JsonArray } from '@wdio
 
 const log = logger('@wdio/shared-store-service')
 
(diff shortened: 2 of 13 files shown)


packages/wdio-shared-store-service/src/client.ts
```diff
@@ -29,38 +39,46 @@ export const getValue = async (key: string): Promise<string | number | boolean |
...
 
-    log.info('Shared store server not yet started, collecting value')
-    pendingValues.set(key, value)
+/**
+ *
+ * @param {string}  key
+ * @param {*}       value
+ */
+export const setResourcePool = async (key: string, value: JsonArray) => {
+    const setPromise = baseUrlPromise.then((baseUrl) => {
```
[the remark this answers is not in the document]
pedrorfernandes: If we `await baseUrlPromise` and `setResourcePool()` is inside an `onPrepare` hook, then it would be stuck forever because this service only sets an url after the users hooks are finished, correct? If this is not true then awaiting in all functions here is way simpler

packages/wdio-shared-store-service/src/server.ts
```diff
@@ -2,9 +2,12 @@ import type { AddressInfo } from 'node:net'
 import polka from 'polka'
 import { json } from '@polka/parse'
 
-import type { JsonCompatible, JsonPrimitive, JsonObject } from '@wdio/types'
+import type { JsonCompatible, JsonPrimitive, JsonObject, JsonArray } from '@wdio/types'
+
+export type ResourcePoolStore = { [x: string]: JsonArray }
 
 const store: JsonObject = {}
+const resourcePoolStore: ResourcePoolStore = {}
```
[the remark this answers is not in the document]
pedrorfernandes: updated

packages/wdio-shared-store-service/src/server.ts
```diff
@@ -46,6 +49,44 @@ export const startServer = () => new Promise<{ port: number, app: PolkaInstance
...
+            const value = req.body.value as JsonCompatible | JsonPrimitive
+
+            if (!Array.isArray(value)) {
+                throw new Error('Resource pool must be an array of values')
+            }
+
+            resourcePoolStore[key] = value
+            return res.end()
+        })
+        .post('/takeValueFromPool', (req, res) => {
```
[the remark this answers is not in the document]
pedrorfernandes: updated, had to add get to types

packages/wdio-shared-store-service/src/server.ts
```diff
@@ -46,6 +49,44 @@ export const startServer = () => new Promise<{ port: number, app: PolkaInstance
...
+
+            if (!pool) {
+                throw new Error(`'${key}' resource pool is does not exist. Set it first`)
+            }
+
+            if (pool.length === 0) {
+                throw new Error(`'${key}' resource pool is empty. Return values to it first`)
+            }
+
+            const value = pool.shift()
```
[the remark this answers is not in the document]
pedrorfernandes: I've tried this:

```javascript
.get('/getValueFromPool', (req, res) => {
            const key = req.body.key as string
            let value = getValueFromPool(key);

            if (value) {
                res.end(JSON.stringify({ value }))
                return
            }

            setTimeout(function secondAttempt() {
                value = getValueFromPool(key);
                if (!value) {
                    throw new Error(`'${key}' resource pool is empty. Set values to it first using 'setResourcePool'`)
                }
                res.end(JSON.stringify({ value }))
            }, 1000)
        })
```

It may be best for the user to specify the desired timeout amount, however this doesn't allow for further retries or faster polling. Should we let that down to the user?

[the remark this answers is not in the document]
pedrorfernandes: Implemented timeout, please check now

packages/wdio-shared-store-service/src/service.ts
```diff
@@ -1,7 +1,9 @@
 import type { JsonCompatible, JsonPrimitive, Services } from '@wdio/types'
 
-import { getValue, setValue, setPort } from './client.js'
+import { getValue, setValue, setPort, setResourcePool, getValueFromPool, addValueToPool } from './client.js'
 import type { SharedStoreServiceCapabilities } from './types.js'
+import type { JsonArray } from '@wdio/types'
```
christian-bromann, line 5: We already import types in line 1, let's merge these.

christian-bromann (commented):
One comment but the implementation looks good to me. Let's add some tests for it.

packages/wdio-shared-store-service/src/service.ts
pedrorfernandes: oops, added now
will work on tests tomorrow

packages/wdio-shared-store-service/src/server.ts
```diff
@@ -46,6 +51,57 @@ export const startServer = () => new Promise<{ port: number, app: PolkaInstance
...
+
+            if (!Array.isArray(value)) {
+                next(new Error('Resource pool must be an array of values'))
+                return
+            }
+
+            resourcePoolStore.set(key, value)
+            return res.end()
+        })
+        .post('/getValueFromPool/:key', async (req, res, next) => {
```
pedrorfernandes, line 66: @christian-bromann setting this to `.get` caused me issues, unit tests that throw errors would hang for 2s instead of an immediate reply. `.post()` throws the error immediately

packages/wdio-shared-store-service/src/server.ts
```diff
@@ -46,6 +51,57 @@ export const startServer = () => new Promise<{ port: number, app: PolkaInstance
...
+            const key = req.params.key as string
+
+            if (!resourcePoolStore.has(key)) {
+                next(new Error(`'${key}' resource pool does not exist. Set it first using 'setResourcePool'`))
```
pedrorfernandes, line 70: @christian-bromann I had to change throwing to `next(new Error`, this is the correct way for polka to throw error messages

packages/wdio-shared-store-service/README.md
```diff
@@ -66,6 +66,33 @@ IMPORTANT! Every spec file should be atomic and isolated from others' specs.
 The idea of the service is to deal with very specific environment setup issues.
 Please avoid sharing test execution data!
 
+### Resource Pools
+
+If the worker threads are competing for resources that must be assigned for each worker, you can use Resource Pool API:
+
+```js
+// wdio.conf.js
+import { setResourcePool, getValueFromPool, addValueToPool} from '@wdio/shared-store-service'
```
christian-bromann, line 75:
```suggestion
import { setResourcePool, getValueFromPool, addValueToPool } from '@wdio/shared-store-service'
```

packages/wdio-shared-store-service/src/server.ts
christian-bromann: Let's fix the unit test then, this should definitely be a get method. Given it is not really a user facing API I am happy to ignore the url path.

packages/wdio-shared-store-service/src/server.ts
christian-bromann: sounds good, let's do:
```suggestion
                return next(new Error(`'${key}' resource pool does not exist. Set it first using 'setResourcePool'`))
```

packages/wdio-shared-store-service/tests/client.test.ts
```diff
@@ -30,12 +30,11 @@ const baseUrl = `http://localhost:${port}`
...
     describe('when used in launcher process', () => {
         it('should not post before server has started', async () => {
-            await setValue('foo', 'bar')
-            await setValue('bar', 'foo')
+            setValue('foo', 'bar')
+            setValue('bar', 'foo')
             expect(got.post).toBeCalledTimes(0)
-            await new Promise((resolve) => setTimeout(resolve, 300))
             setPort(port)
-            await new Promise((resolve) => setTimeout(resolve, 300))
```
christian-bromann, line 38: Why changing this test? It doesn't seem to be connected with your feature.

packages/wdio-shared-store-service/tests/server.test.ts
```diff
@@ -15,12 +15,19 @@ describe('WdioSharedStoreService exports', () => {
...
 
     beforeAll(async () => {
+
         result = await startServer()
         const baseUrl = `http://localhost:${result.port}`
         setUrl = `${baseUrl}/set`
         getUrl = `${baseUrl}/get`
+        setResourcePoolUrl = `${baseUrl}/setResourcePool`
+        getValueFromPoolUrl = `${baseUrl}/getValueFromPool`
+        addValueToPoolUrl = `${baseUrl}/addValueToPool`
```
christian-bromann, line 30: Actually can we change the urls for the internal API to:
```suggestion
        setResourcePoolUrl = `${baseUrl}/pool/set`
        getValueFromPoolUrl = `${baseUrl}/pool/get`
        addValueToPoolUrl = `${baseUrl}/pool/add`
```

christian-bromann (changes requested):
Some comments, great work so far 👍

packages/wdio-shared-store-service/src/server.ts
pedrorfernandes: figured it out, `got` has a retry mechanism on error, I've set all `throwHttpErrors: false, retry: { limit: 0 }`

packages/wdio-shared-store-service/src/server.ts
pedrorfernandes: changed to 
```javascript
return next(`'${key}' resource pool does not exist. Set it first using 'setResourcePool'`)`
```
because new Error serializes to `{}`, serializing all errors by hand is too cumbersome
Polka recommends to just `next(string)` because this throws the error

I've added a catch + throw error on the client to read this message

packages/wdio-shared-store-service/tests/server.test.ts
pedrorfernandes: done, i've only updated the internal api, don't know if you want to also change function names of the JS api

packages/wdio-shared-store-service/src/client.ts
```diff
@@ -29,40 +37,48 @@ export const getValue = async (key: string): Promise<string | number | boolean |
...
+export const addValueToPool = async (key: string, value: JsonPrimitive | JsonCompatible) => {
+    const baseUrl = await baseUrlPromise
+    const res = await got.post(`${baseUrl}/pool/add`, { json: { key, value }, responseType: 'json' }).catch(errHandler)
+    return res?.body ? (res.body as JsonObject).value : undefined
 }
 
-const errHandler = (err: Response<Error>) => {
-    log.warn(err.statusCode, err.statusMessage, err.url, err.body)
+const errHandler = (err: RequestError) => {
+    throw new Error(`${err.response?.body || 'Shared store server threw an error'}`)
```
pedrorfernandes, line 83: @christian-bromann please validate this

before: the polka server would crash on error throw and all wdio process crashed suddenly
now: the polka server returns an error, this client throws it to wdio, the message is logged, the worker crashes, but wdio process continues on to the next test

i think this was the desired behaviour, not to immediately crash all running suites

packages/wdio-shared-store-service/README.md
pedrorfernandes: done

packages/wdio-shared-store-service/src/server.ts
pedrorfernandes: also, to note: `GET` operations should not modify the state of data, that's why got was retrying, assuming that it wouldn't make a difference in the server. A `POST` operation is more suited to our api call because it's actually modifying internal data

packages/wdio-shared-store-service/src/client.ts
christian-bromann: Sounds good to me 👍

christian-bromann (approved):
Thank you for all this work. I want to include this into the next release so I will go ahead, merge this and address the missing pieces separately.

Issue #10040, opened 2023-03-23
[🐛 Bug]: Parent Elements cannot be selected by the Text


RahulARanger:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

8.6.7

### Node.js Version

18.14.2

### Mode

Standalone Mode

### Which capabilities are you using?

```typescript
{
browserName: "chrome",
maxInstances: 1
}
```

### What happened?

For our project, we have preferred [selecting an element by text](https://webdriver.io/docs/selectors/#element-with-certain-text). This helped us select very minor elements easily and quickly but recently we observed it was not able to select the required elements.

After deep dive we have observed, the following change

For instance, for the below snipped, in order to select the button
```html
<button>
      <span>Button Text</span>
</button>
```

we wrote the following script,

```js
$("button=Button Text").click()
```

but now we are now required to write 
```js
$("span=Button Text").click()
```

In the first case the main advantage was;

1. Allow Users to select the Parent elements based on the text regardless of its child nodes
```js
expect($("button#button-id")).toHaveText("Button Text")
```
2.  From the User perspective, both tags have the same Text.

And after we updated our packages we faced this issue, after some exploring, I really accept the motive behind #9992 which is technically correct since span is the only element having that text but from the user perspective, it is also valid to say that the button tag has the same text.

### What is your expected behavior?

Both methods must be allowed to select any element with the requested text.

```js
$("button=Button Text").click()
```
```js
$("span=Button Text").click()
```

### How to reproduce the bug.

To reproduce this bug, please execute this following standalone code:

### Aim: To select the Following Button for Sign In

<img width="960" alt="image" src="https://user-images.githubusercontent.com/57987728/227316687-c59c846c-574f-4b5e-822f-d83a0f088419.png">

### Target Element's HTML

```html
<a class="gb_ja gb_ka gb_ge gb_gd">
<span class="gb_ie">Sign in</span>
</a>
```

#### Automation Script

```js
import { remote } from 'webdriverio'

const browser = await remote({
    capabilities: {
        browserName: 'chrome',
        'goog:chromeOptions': {
            args: process.env.CI ? ['headless', 'disable-gpu'] : []
        }
    }
})

await browser.url('https://www.google.com/')


[...]

PR #10046, opened 2023-03-24
Allow selectors to further search for elements by text and tag name

RahulARanger:
## Proposed changes

Aims to solve Issue: #10040.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

Currently, for selecting an element by Tag Name and Content, we are following the XPath of the form: ``.//{'*' if not tag else tag}[normalize-space(text()) = "{text}"]``

Example: `.//span[normalize-space(text()) = "Sign in"]`

It fetches the text specific to that node and normalizes 
it.  but we would not be able to find the parent element as it might have some child nodes. 

```html
<button><span>Click Me!</span></button>
```

In order to select the element *Button*, we cannot have the selector: `$("button=Click Me!")` to select the button. but we can select *span* with a similar syntax: `$("span=Click Me!")`.

#### Suggestion: 

If we would have used the following XPath expression of the form: ``.//{'*' if not tag else tag}[normalize-space() = "{text}"]``

Example: `.//button[normalize-space() = "Click Me!"]`

we can select both button and span based on our request. as it is converting the text (inside of its child nodes if exists) and normalizes it.

But that had an edge case:

```html
<div class="not-me">
 <div class="sub">
  <div class="main">
   Deep
  </div>
 </div>
</div>
```

Selector: `$(div=Deep)` would select all the *div s* with "not-me' as the first found. Expected result is the *div* with class "main". But the Current Implementation would select the right selector.

So the Fix is to cater to both the type of scenarios. keeping the current implementation as stricter part and the suggestion as the looser part.

Suggested Selector is form: `.//tag[normalize-space(text()) = "text"] | .//tag[not(.//tag[normalize-space(text()) = "text"]) and normalize-space() = "text"]`

Explanation

1. Look for the expression: `.//tag[normalize-space(text()) = "text"]` and if it exists,
   * `not(.//tag[normalize-space(text()) = "text"]` becomes false so right side of union operator is null
   * we pick only the left side which is `.//tag[normalize-space(text()) = "text"]` _same as existing implemetation_
2. If `.//tag[normalize-space(text()) = "text"]` doesn't exist,
   * then right side of the union operator is `.//tag[true and normalize-space(text()) = "text"]`

### DIFF
--- a/e2e/browser-runner/lit.test.js
+++ b/e2e/browser-runner/lit.test.js
@@ -122,6 +122,27 @@ describe('Lit Component testing', () => {
             expect(await $('div*=me').getHTML(false)).toBe('Find me')
         })
 
+        it('fetches the parent element by content correctly', async () => {
+            render(
+                html`<button><span>Click Me!</span></button>`,
+                document.body
+            )
+            expect(await $('span=Click Me!').getHTML(false)).toBe('Click Me!')
+            expect(await $('button=Click Me!').getHTML(false)).toBe('<span>Click Me!</span>')
+            const elem = $('button=Click Me!')
+            await expect(elem).toHaveText('Click Me!')
+        })
+
+        it('fetches the element with multiple text nodes by the content', async () => {
+            render(
+                html`<p>
+                Find
+                me</p>`, document.body
+            )
+            const elem = await $('p=Find me')
+            await expect(elem).toHaveText('Find me')
+        })
+
         it('fetches element by JS function', async () => {
             expect((await $(() => document.body).getTagName()).toLowerCase()).toBe('body')
         })
--- a/e2e/browser-runner/lit.test.js
+++ b/e2e/browser-runner/lit.test.js
@@ -134,11 +134,10 @@ describe('Lit Component testing', () => {
         })
 
         it('fetches the element with multiple text nodes by the content', async () => {
-            render(
-                html`<p>
-                Find
-                me</p>`, document.body
-            )
+            const container = document.createElement('p')
+            container.appendChild(document.createTextNode('Find'))
+            container.appendChild(document.createTextNode(' me'))
+            render(container, document.body)
             const elem = await $('p=Find me')
             await expect(elem).toHaveText('Find me')
         })
--- a/packages/webdriverio/src/utils/findStrategy.ts
+++ b/packages/webdriverio/src/utils/findStrategy.ts
@@ -246,7 +246,7 @@ export const findStrategy = function (selector: SelectorStrategy, isW3C?: boolea
             throw new Error(`InvalidSelectorMatch: Strategy 'xpath extended' has failed to match '${stringSelector}'`)
         }
         const PREFIX_NAME: Record<string, string> = { '.': 'class', '#': 'id' }
-        const conditions = []
+        const conditions: Array<string> = []
         const [
             tag,
             prefix, name,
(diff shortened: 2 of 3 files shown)


e2e/browser-runner/lit.test.js
```diff
@@ -122,6 +122,27 @@ describe('Lit Component testing', () => {
...
+            const elem = $('button=Click Me!')
+            await expect(elem).toHaveText('Click Me!')
+        })
+
+        it('fetches the element with multiple text nodes by the content', async () => {
+            render(
+                html`<p>
+                Find
+                me</p>`, document.body
+            )
```
christian-bromann, line 141: I am not sure of Lit renders this as two text nodes, so let's do this:
```js
const container = document.createElement('p')
container.appendChild(document.createTextNode('Find'))
container.appendChild(document.createTextNode('me'))
document.body.appendChild(container)
```

christian-bromann (commented):
Awesome work, one suggestion.

e2e/browser-runner/lit.test.js
RahulARanger: sure

christian-bromann (approved):
👍  Thanks a lot!

And congrats on your first WebdriverIO contribution 🎉

lacell75:
I have the same issue in my side. I cannot find a tr by text now: `$(tr*=good)`
Example, I need to get the line with the text 'good' to clik on the action 2 : $(tr*=good).$('tbutton').click() -> Not working now
`<table>
 <tbody>
  <tr>
   <td>
     my information
   </td>
   <td>
    other information
   </td>
   <td>
     Action
    <button>
    </button>
   </td>
  </tr>
  <tr>
   <td>
   the good information
   </td>
   <td>
    any information
   </td>
   <td>
    Action
    <button>
    </button>
   </td>
  </tr>
 </tbody>
</table>`

christian-bromann:
You might want to try ` $(td*=good).parentElement().$('button').click()`

lacell75:
> You might want to try ` $(td*=good).parentElement().$('button').click()`

Thanks Christian, that works.

RahulARanger:
> But I reproduced another issue. If an element is present inside the element and the text is after this element, the search doesn't work. Example :
> 
> ```
>   await browser.url('https://webdriver.io/blog');
>   await $('[itemprop="blogPost"]').waitForDisplayed();
>   console.log(await $('[itemprop="articleBody"]').$('p*=play around with').isExisting()); // text is before the "a" element -> return true
>   console.log(await $('[itemprop="articleBody"]').$('p*=a thousand words').isExisting()); // text is after the "a" element -> return false
> ```
> 
> ```
> <p>The WebdriverIO framework is a versatile tool that offers a lot of features for you to play around with. The goal of our 
>      <a href="https://webdriver.io" target="_blank" rel="noopener noreferrer">project documentation</a> is to communicate these features well and give you an understanding, on how they could be applied in your project. A central contributor to this are code examples. Many times they can convey the principle idea of a feature like a picture that is worth a thousand words.
> </p>
> ```

The example has 3 elements, 2 TextNodes, and one "A" Tag. `//p[contains(text(), "a thousand words")]` (which is equivalent to `await $("p*=a thousand words")`) searches the text by the first element only.

Hence `$('p*=play around with')` this is passing as it is the first text node.  You can try the XPath expression if needed
`$('[itemprop="articleBody"]').$("//p[text()[contains(.,'is worth a thousand')]]")`

lacell75:
For me, the example has only 2 elements. An element with <P> tag and one with "A" inside the "P" element

Issue #9993, opened 2023-03-17
[🐛 Bug]: Allure Reporter / Retries not listed anymore.


muhserks:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

8.5.6

### Node.js Version

18

### Mode

WDIO Testrunner

### Which capabilities are you using?

```typescript
"@wdio/allure-reporter": "^8.5.0"
"@wdio/mocha-framework": "^8.5.6",
```

### What happened?

It seems that the allure reporter creates a new step instead of a new entry in the "Retries" Tab (if a test fails and retries are configure in wdio).
I believe this also leads to unstable builds in jenkins.
In previuos version (wdio7 and corresponding allure reporter version) retries were listed in the "Retries" Tab and did not made jenkins unstable.

I attached some screenshots for comparison.
![Allure-8 5 0](https://user-images.githubusercontent.com/91063753/225873424-4ac12055-a4c4-4676-b4bd-8eee4dc36333.png)
![Allure-7 26 0_Screenshot_1](https://user-images.githubusercontent.com/91063753/225873458-875f571b-744b-44e3-996b-a80393f02511.png)
![Allure-7 26 0_Screenshot_2](https://user-images.githubusercontent.com/91063753/225873488-faba1f48-0c63-40e6-883b-7fedd0501a63.png)

### What is your expected behavior?

Retries are correctly listed in the retry tab and hopefully the publisher will not consider it as unstable anymore.

### How to reproduce the bug.

Let a test fail and look into the generated allure report.

### Relevant log output

```typescript
n.a.
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

christian-bromann:
Hey,
thanks for filing the issue. This seems indeed like a regression from the v1 Allure reporter integration. We would appreciate any help on this so please help us out! Have a look at our [contributing guidelines](https://github.com/webdriverio/webdriverio/blob/master/CONTRIBUTING.md) and reach out if you have any questions. Cheers!

BorisOsipov:
cc @epszaw

praveendvd:
@christian-bromann  i will take this up

praveendvd:
[...]

PR #10053, opened 2023-03-26
Added histroy and testcaseid

praveendvd:
## Proposed changes

Allure checks if a test case is unique by using history and testcaseid , so exiplicitly added cid as these ids 

fix for https://github.com/webdriverio/webdriverio/issues/9993

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments


christian-bromann (commented):
@epszaw is there a way to set these information without setting a private property?

epszaw:
> @epszaw is there a way to set these information without setting a private property?

Yes. If you talk about unit tests, it's possible to use `getResults` (`packages/wdio-allure-reporter/tests/helpers/wdio-allure-helper.ts`) utility to read test results as plain json. I guess, the parameters should be present.

packages/wdio-allure-reporter/src/reporter.ts
```diff
@@ -240,6 +240,8 @@ export default class AllureReporter extends WDIOReporter {
 
         if (cid) {
             this._state.currentTest.addLabel(LabelName.THREAD, cid)
+            this._state.currentTest.historyId = cid
```
BorisOsipov, line 243: You cannot use `cid` as the `historyId` because the `historyId` must remain the same for each test run. Actually, it is a unique identifier for the specific test between all runs of specific test.
`cid` is a random value.

I'd suggest to use `md5(suiteName+testCaseName)` as  `historyId`.

BorisOsipov (changes requested):
See comments.

praveendvd:
@BorisOsipov made the changes thank you so much could you have a look

packages/wdio-allure-reporter/src/reporter.ts
```diff
@@ -26,6 +26,7 @@ import type {
 import {
     TYPE as DescriptionType
 } from './types.js'
+import { md5 } from 'allure-js-commons'
```
christian-bromann, line 29: We already import `allure-js-commons` at the top, so let's add `md5` to it.

BorisOsipov:
@praveendvd Looks good. It seems that it would be nice to add unit tests for these changes to avoid any potential regressions.

christian-bromann (approved):
I would love to include this into the next release so I will go ahead and merge it.
Thanks @praveendvd for the fix 👍

PR #10129, opened 2023-04-06
docs(mock): 'matches' -> 'calls' property

KuznetsovRoman:
## Proposed changes

Change `Mock` interface in [docs](https://webdriver.io/docs/api/mock#properties)  (use `calls` property instead of `matches`)

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/website/docs/api/Mock.md
+++ b/website/docs/api/Mock.md
@@ -25,7 +25,7 @@ A mock object contains the following properties:
 | `url` | `String` | The url passed into the mock command |
 | `filterOptions` | `Object` | The resource filter options passed into the mock command |
 | `browser` | `Object` | The [Browser Object](/docs/api/browser) used to get the mock object. |
-| `matches` | `Object[]` | Information about matching browser requests, containing properties such as `url`, `method`, `headers`, `initialPriority`, `referrerPolic`, `statusCode`, `responseHeaders` and `body` |
+| `calls` | `Object[]` | Information about matching browser requests, containing properties such as `url`, `method`, `headers`, `initialPriority`, `referrerPolic`, `statusCode`, `responseHeaders` and `body` |
 
 ## Methods
 


christian-bromann:
Thanks for raising the PR.

I think this is an issue with the docs. The user is suppose to use `calls` to receive these information. `matches` is more an internal object for this.

KuznetsovRoman:
> Thanks for raising the PR.
> 
> I think this is an issue with the docs. The user is suppose to use `calls` to receive these information. `matches` is more an internal object for this.

Ok, i will update the docs in this PR. Also, about mocks: i need an ability to check if mock (spy) has some pending requests. Will you approve a feature, which emits something like `onRequest`, `onResponse` on request and response accordingly?

christian-bromann:
> Will you approve a feature, which emits something like `onRequest`, `onResponse` on request and response accordingly?

Seems like a reasonable feature request. Yeah totally!

> (i see how it can be done with `devtools` mock, but not with `webdriver` one)

For WebDriver we will have to use the new WebDriver Bidi features which have not landed yet. While you can already get request information, it is not yet possible to mock/replace requests and responses.

KuznetsovRoman:
Also, do you mind if i add documentation about `WDIO_LOG_LEVEL` environment variable? It looks like a part of public api (because [here](https://github.com/webdriverio/webdriverio/blob/62e24d4dd9d58f3b4168866189d60e7e8da09a9b/packages/wdio-logger/src/node.ts#L162-L164) it assumes, it can be set from the outside) but is not being documented. I want to use this env variable in my code (i already asked about this in gitter, but didn't got the answer to my question), but i'm scared of using undocumented variables

christian-bromann:
> Also, do you mind if i add documentation about `WDIO_LOG_LEVEL` environment variable?

Good call, let's add it to https://webdriver.io/docs/api/environment

christian-bromann (approved):
LGTM 👍

PR #10167, opened 2023-04-11
fix(logs): WDIO_LOG_LEVEL

KuznetsovRoman:
## Proposed changes

1) fix log output with set `WDIO_LOG_LEVEL` env variable. 

Currently if you set WDIO_LOG_LEVEL env variable, you will still see `info` level logs from `devtools` and `webdriver` because `devtools.newSession` and `webdriver.newSession` defaulting `logLevel` to `info` ([devtools](https://github.com/webdriverio/webdriverio/blob/main/packages/devtools/src/index.ts#L40) and [webdriver](https://github.com/webdriverio/webdriverio/blob/main/packages/webdriver/src/index.ts#L23)). Though its not needed because  `setLogLevelsConfig` already sets right logLevels on `wdio.remote` [here](https://github.com/webdriverio/webdriverio/blob/main/packages/webdriverio/src/index.ts#L44), and `getLogger` [already has this fallback with default value](https://github.com/webdriverio/webdriverio/blob/main/packages/wdio-logger/src/node.ts#L120).

2) add documentation about `WDIO_LOG_LEVEL` env variable.

## Types of changes

- [x] Documentation update

## Checklist


packages/webdriver/src/index.ts
```diff
@@ -22,8 +22,8 @@ export default class WebDriver {
     ): Promise<Client> {
         const params = validateConfig(DEFAULTS, options)
 
-        if (!options.logLevels || !options.logLevels.webdriver) {
-            logger.setLevel('webdriver', params.logLevel!)
+        if (params.logLevel && (!options.logLevels || !options.logLevels.webdriver)) {
```
KuznetsovRoman, line 25: as in [devtools](https://github.com/webdriverio/webdriverio/blob/main/packages/devtools/src/index.ts#L42)

KuznetsovRoman:
one e2e test ("should allow to use then/catch/finally") randomly fell on windows due to `test timeout`.

packages/devtools/src/constants.ts
```diff
@@ -77,7 +77,6 @@ export const DEFAULTS: Options.Definition<Options.WebDriver> = {
      */
     logLevel: {
         type: 'string',
-        default: 'info',
```
christian-bromann, line 80: Why removing the default values?

packages/webdriver/src/constants.ts
```diff
@@ -67,7 +67,6 @@ export const DEFAULTS: Options.Definition<Required<Options.WebDriver>> = {
      */
     logLevel: {
         type: 'string',
-        default: 'info',
```
christian-bromann, line 70: same

packages/devtools/src/constants.ts
KuznetsovRoman: I mentioned above, in "proposed changes"

> 
Currently if you set WDIO_LOG_LEVEL env variable, you will still see `info` level logs from `devtools` and `webdriver` because `devtools.newSession` and `webdriver.newSession` defaulting `logLevel` to `info` ([devtools](https://github.com/webdriverio/webdriverio/blob/main/packages/devtools/src/index.ts#L40) and [webdriver](https://github.com/webdriverio/webdriverio/blob/main/packages/webdriver/src/index.ts#L23)). Though its not needed because  `setLogLevelsConfig` already sets right logLevels on `wdio.remote` [here](https://github.com/webdriverio/webdriverio/blob/main/packages/webdriverio/src/index.ts#L44), and `getLogger` [already has this fallback with default value](https://github.com/webdriverio/webdriverio/blob/main/packages/wdio-logger/src/node.ts#L120).

christian-bromann: Let's keep this anyway as it better describes the default behavior. The default log level for `@wdio/logger` can maybe change in the future and we would like to keep the current default in `devtools` and `webdriver`.

KuznetsovRoman: Whats about

```ts
    options.logLevel = process.env.WDIO_LOG_LEVEL ?? options.logLevel
    const params = validateConfig(DEFAULTS, options)
```
[in devtools/src/index.ts](https://github.com/webdriverio/webdriverio/blob/main/packages/devtools/src/index.ts#L40)

or

```ts
    const params = validateConfig(DEFAULTS, options)
    params.logLevel = process.env.WDIO_LOG_LEVEL?? options.logLevel ?? params.logLevel
```

to set priority as:
1. `process.env.WDIO_LOG_LEVEL`
2. `options.logLevel`
3. `DEFAULT`

`DEFAULT` priority being higher than `process.env.WDIO_LOG_LEVEL` makes no sense

christian-bromann: That works for me 👍

KuznetsovRoman: Fixed:

```ts
const envLogLevel = process.env.WDIO_LOG_LEVEL as Options.WebDriverLogTypes | undefined
options.logLevel = envLogLevel ?? options.logLevel // before validateConfig, so we could validate process.env.WDIO_LOG_LEVEL
const params = validateConfig(DEFAULTS, options)
```

christian-bromann (approved):
LGTM 👍 

Thank you!

PR #10188, opened 2023-04-13
feat: export cjs interface for wdio-cli

### LINKED ISSUE #9869: [🐛 Bug]: CJS require of @wdio/cli and @wdio/reporter not implemented

### WebdriverIO Version

8.4

### Node.js Version

16.13.2

### Mode

WDIO Testrunner

_No response_

### What happened?

Tests are written in a cjs project with the testrunner started programmatically. When upgrading to v8 some wdio modules are no longer possible to require. 
```
const Launcher = require('@wdio/cli').default
const WDIOReporter = require('@wdio/reporter').default;
```
fails since they only support esm import.
A temporary workaround is to use ```const { Launcher } = await import('@wdio/cli');```
but in the case of @wdio/reporter that is a bit trickier.

To be able to use 
```
const Launcher = require('@wdio/cli').default
const WDIOReporter = require('@wdio/reporter').default;
```

### How to reproduce the bug.

Use 
```
require('@wdio/cli').default
```
or
```
require('@wdio/reporter').default
```
in v8

### Relevant log output

```typescript
Launcher failed to start the test Error [ERR_REQUIRE_ESM]: require() of ES Module
```

### Code of Conduct

DirkoOdendaal:
Fixes #9869 

## Types of changes

- [ x ] Bugfix (non-breaking change which fixes an issue)
- [ ] Documentation update

## Checklist

## Further comments

Export interface for CJS

### DIFF
--- a/packages/wdio-cli/src/cjs/index.ts
+++ b/packages/wdio-cli/src/cjs/index.ts
@@ -1,63 +1,53 @@
 import type { RunCommandArguments } from '../types.js'
 import type { Options, Capabilities } from '@wdio/types'
-import type { EndMessage } from '../launcher.js'
+import logger from '@wdio/logger'
+
+const log = logger('@wdio/cli:launcher')
 
 export class Launcher {
 
-    esmLauncher: any
+    #esmLauncher: any
 
     constructor(
         private configFilePath: string,
         private args: Partial<RunCommandArguments> = {},
         private isWatchMode = false
     ) {
-        import('../index.js').then(esmModule => this.esmLauncher = new esmModule.Launcher(this.configFilePath, this.args, this.isWatchMode))
+        import('../launcher.js').then(launcher => {
+            log.info('Initializing main Launcher')
+            this.#esmLauncher = new launcher.default(this.configFilePath, this.args, this.isWatchMode)
+            return this.#esmLauncher
+        })
     }
 
     /**
      * run sequence
-     * @return  {Promise}  that only gets resolves with either an exitCode or an error
+     * @return  {Promise}  that only gets resolved with either an exitCode or an error
      */
-    async run() {
-        return this.esmLauncher.run()
+    async run(): Promise<undefined | number> {
+        return this.#esmLauncher.run()
     }
 
     /**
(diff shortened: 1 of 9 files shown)


packages/wdio-cli/src/cjs/index.ts
```diff
@@ -0,0 +1,115 @@
+import type { RunCommandArguments } from '../types.js'
+import type { Options, Capabilities } from '@wdio/types'
+import type { EndMessage } from '../launcher.js'
+
+export class Launcher {
+
+    esmLauncher: any
```
christian-bromann, line 7:
```suggestion
    #esmLauncher: any
```

packages/wdio-cli/src/cjs/index.ts
```diff
@@ -0,0 +1,115 @@
...
+    ) {
+        this.esmLauncher.startInstance(specs, caps, cid, rid, retries)
+    }
+
+    /**
+     * generates a runner id
+     * @param  {Number} cid capability id (unique identifier for a capability)
+     * @return {String}     runner id (combination of cid and test id e.g. 0a, 0b, 1a, 1b ...)
+     */
+    getRunnerId (cid: number) {
```
christian-bromann, line 85: can we make a bunch of these methods private so we don't have to expose them here? The launcher has been build before private class functions in JavaScript were a thing. Maybe we can optimize the code here so we only have to expose the `run` method which is the only one we've documented.

DirkoOdendaal: Some of the methods I refactored to private methods. Not sure how many people actually use this, so kept a few methods public for now.
Also happy making most of them private.

christian-bromann: > Also happy making most of them private.

Let's make all of them private except `run`. We don't have anything else documented and therefor people should not use any other method. If folks start to report this, we can think about making them public again.

DirkoOdendaal: Yeah you are right. Just did that.

christian-bromann (approved):
Nicely done 👌

PR #10297, opened 2023-05-01
Check for hostname before setting isMac to true

therealbrad:
## Proposed changes

[//]: # Check if hostname is set when determining if instance is running locally. Without this check, a local mac os instance can send Command key to a remote Window instance.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments


christian-bromann:
@therealbrad thanks for taking a stab at this. Let's make sure we add proper unit tests so we don't run into any regressions moving on. There are already [existing unit tests](https://github.com/webdriverio/webdriverio/blob/main/packages/webdriverio/tests/commands/browser/action.test.ts#L58-L74) for this case. Also we need you to have the CLA signed. Thank you!

christian-bromann (commented):
LGTM 👍 

Once the CLA is signed I can go ahead and merge this.

christian-bromann (approved):
Thank you 👍

christian-bromann:
@therealbrad there seems to be an failure in the test?

therealbrad:
@christian-bromann My bad. I meant to remove that test because I didn't see a way to run unit tests and set browser.options.hostname to null. Is there a way to do that?

christian-bromann:
I thought `browser.options.hostname = 'local'` would do the trick because it is the same value as `this.instance` if not we should track down where `this.instance` is defined and find a way to modify that value for our test purposes. You are always free to write an individual unit test for the action class itself if it makes writing the test easier.

therealbrad:
@christian-bromann I changed it to explicitly check if running locally via hostname instead of assuming it. Seems to work but I don't know if this breaks anything else.

packages/webdriverio/src/utils/actions/key.ts
```diff
@@ -20,7 +20,7 @@ export default class KeyAction extends BaseAction {
             // check capabilities first
             platformName && platformName.match(/mac(\s)*os/i) ||
             // if not set, expect we run locally
-            os.type().match(/darwin/i)
+            (this.instance.options.hostname?.match(/local/i) && os.type().match(/darwin/i))
```
christian-bromann, line 23: `hostname` could also be `127.0.0.1`. I think it is impossible to really know whether the session runs on ones machine or not. So let's just check for `localhost` and `127.0.0.1`

christian-bromann:
Thank you @therealbrad

Issue #10394, opened 2023-05-16
[🐛 Bug]: Cucumber doc strings are not displayed on reports generated by spec-reporter


ductoan0295:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

8.10.2

### Node.js Version

18.14.2

### Mode

WDIO Testrunner

### Which capabilities are you using?

```typescript
{
  browserName: 'chrome',
  browserVersion: 'latest',
  platformName: 'Windows 10'
}
```

### What happened?

Doc strings of a cucumber step do not display on the report generated by wdio spec-reporter.

### What is your expected behavior?

Doc strings of a cucumber step display on the report generated by wdio spec-reporter.

The report content should be:
```
 "spec" Reporter:
------------------------------------------------------------------
[chrome 113.0.5672.93 Windows 10 #0-0] Running: chrome (v113.0.5672.93) on Windows 10
[chrome 113.0.5672.93 Windows 10 #0-0] Session ID: 0a438b3d6040c9f9144fec6c8d87c05c
[chrome 113.0.5672.93 Windows 10 #0-0]
[chrome 113.0.5672.93 Windows 10 #0-0] » \test\features\test.feature
[chrome 113.0.5672.93 Windows 10 #0-0] Test
[chrome 113.0.5672.93 Windows 10 #0-0] Docstring
[chrome 113.0.5672.93 Windows 10 #0-0]    ✓ Given I have the following docstring
[chrome 113.0.5672.93 Windows 10 #0-0]        First line
[chrome 113.0.5672.93 Windows 10 #0-0]        Second line
[chrome 113.0.5672.93 Windows 10 #0-0]        Third line
[chrome 113.0.5672.93 Windows 10 #0-0]    ✓ When I run test
[chrome 113.0.5672.93 Windows 10 #0-0]    ✓ Then I find it in the reporter
[chrome 113.0.5672.93 Windows 10 #0-0]
[chrome 113.0.5672.93 Windows 10 #0-0] 3 passing (996ms)


Spec Files:      1 passed, 1 total (100% completed) in 00:00:20
```

### How to reproduce the bug.

Create a webdriverio project using cucumber framework, spec reporter for chrome testing and add the following feature and step definition files

**test.feature**
```
Feature: Test
    Scenario: Docstring
[...]

PR #10416, opened 2023-05-19
Spec repoter - add docstring to the final report

ductoan0295:
## Proposed changes

Add docstring to the final report for cucumber tests generated by spec-reporter. This PR is created to solve this bug #10394.

The `argument` property of each `eventsToReports` has `string | Argument` type, but current logic only handles the Argument case. So the datatable of executed cucumber tests which has `Argument` type will display correctly on the report but the `docstring` will be missed.

This fix adds the logic to handle when the `argument` property is a string.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/packages/wdio-spec-reporter/tests/index.test.ts
+++ b/packages/wdio-spec-reporter/tests/index.test.ts
@@ -9,7 +9,8 @@ import {
     SUITES_NO_TESTS,
     SUITES_WITH_DATA_TABLE,
     SUITES_NO_TESTS_WITH_HOOK_ERROR,
-    SUITES_MULTIPLE_ERRORS
+    SUITES_MULTIPLE_ERRORS,
+    SUITES_WITH_DOC_STRING
 } from './__fixtures__/testdata.js'
 
 vi.mock('chalk')
@@ -395,10 +396,9 @@ describe('SpecReporter', () => {
             expect(result).toMatchSnapshot()
         })
 
-        it('should not print if data table format is not given', () => {
+        it('should print doc string', () => {
             tmpReporter.getOrderedSuites = vi.fn(() => {
-                const suites = Object.values(JSON.parse(JSON.stringify(SUITES_WITH_DATA_TABLE))) as any[]
-                suites[0].hooksAndTests[0].argument = 'some different format'
+                const suites = Object.values(JSON.parse(JSON.stringify(SUITES_WITH_DOC_STRING))) as any[]
                 return suites
             })
             const result = tmpReporter.getResultDisplay()
--- a/packages/wdio-spec-reporter/src/index.ts
+++ b/packages/wdio-spec-reporter/src/index.ts
@@ -352,13 +352,24 @@ export default class SpecReporter extends WDIOReporter {
                 // Output for a single test
                 output.push(`${testIndent}${chalk[this.getColor(state)](this.getSymbol(state))} ${testTitle}`)
 
-                // print cucumber data table cells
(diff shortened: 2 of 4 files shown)


packages/wdio-spec-reporter/tests/index.test.ts
```diff
@@ -395,10 +396,9 @@ describe('SpecReporter', () => {
             expect(result).toMatchSnapshot()
         })
 
-        it('should not print if data table format is not given', () => {
+        it('should print doc string', () => {
```
christian-bromann, line 399: Why do we replace the test? Can we just add one?

ductoan0295: Currently, this case tests if the `argument` is a string, there will nothing to show. The new behavior is when the `argument` is a string, the doc string will be printed in the report for the related test. Behavior changes required this case to be replaced IMO.

christian-bromann: We also want to ensure nothing bad happens if the argument is empty, so let's keep the existing test and just add a new one.

packages/wdio-spec-reporter/tests/index.test.ts
```diff
@@ -395,7 +396,7 @@ describe('SpecReporter', () => {
             expect(result).toMatchSnapshot()
         })
 
-        it('should not print if data table format is not given', () => {
+        it('should not print if argument is a single line doc string', () => {
```
ductoan0295, line 399: I've changed the name and the expected result of the `should not print if data table format is not given` test. The existing name and expectation are not appropriate when supporting doc strings which is a string. It will be covered by the new test case `should not print if the argument is undefined`.

packages/wdio-spec-reporter/tests/index.test.ts
```diff
@@ -405,6 +406,36 @@ describe('SpecReporter', () => {
...
+                const suites = Object.values(JSON.parse(JSON.stringify(SUITES_WITH_DOC_STRING))) as any[]
+                suites[0].hooksAndTests[0].argument = ''
+                suites[0].hooksAndTests[1].argument = ''
+                return suites
+            })
+            const result = tmpReporter.getResultDisplay()
+            expect(result).toMatchSnapshot()
+        })
+
+        it('should not print if the argument is undefined', () => {
```
ductoan0295, line 429: This case will cover `should not print if data table format is not given`.

packages/wdio-spec-reporter/tests/index.test.ts
```diff
@@ -395,7 +396,7 @@ describe('SpecReporter', () => {
             expect(result).toMatchSnapshot()
         })
 
-        it('should not print if data table format is not given', () => {
+        it('should not print if argument is a single line doc string', () => {
             tmpReporter.getOrderedSuites = vi.fn(() => {
                 const suites = Object.values(JSON.parse(JSON.stringify(SUITES_WITH_DATA_TABLE))) as any[]
                 suites[0].hooksAndTests[0].argument = 'some different format'
```
ductoan0295, line 402: The current code support only data table (type of Argument) and string can be used to test for wrong data table format input. The changing for doc string supports will not allow this as the argument is a string for doc string.

packages/wdio-spec-reporter/tests/index.test.ts
ductoan0295: Please check my solution for this concerning test case.

christian-bromann (approved):
Awesome, LGTM 👍

PR #10452, opened 2023-05-26
Allure reporter add environment variables in config

### LINKED ISSUE #10321: [🐛 Bug]: Allure reporter adding environment value overwrites the last value

### WebdriverIO Version

latest

### Node.js Version

v16.13.2

### Mode

WDIO Testrunner

_No response_

### What happened?

When using allureReporter.addEnvironment() the new value overwrites the last value.

The problem comes from https://github.com/webdriverio/webdriverio/blob/f0c607340b4f31a4ae48509115335283f488e38f/packages/wdio-allure-reporter/src/reporter.ts#L740.

I guess the reason is that in allure reporter you should now set all environment variables at once as it is a `Record<string,string>`, which makes IMHO also sense.

I have two approaches to solve the problem, and I'm happy to implement whichever one the community prefers.

1. Setting all variables in wdio.conf in the reporter configuration is the first approach. Normally, you would not/should not change environment variables during the tests. So IMHO something like this makes sense:
```
reporters: [['allure', {
    outputDir: 'allure-results',
    environment: [
        a:'a',
        b:'b'
    ]
}]],
```
2. Changing `addEnvironment(name: string, value: string)` to `setEnvironment(Record<string,string>)` is the second approach.

The second one is more flexible, but in my opinion, the first one is the right approach because the second one opens the door to mistakes and makes it possible to overwrite the environment variable information.

### How to reproduce the bug.

Just write this anywhere in your test:
```
allureReporter.addEnvironment('d', 'c');
allureReporter.addEnvironment('a', 'b');
```
### Relevant log output

```typescript
a b
```

### Code of Conduct

m4hdyar:
## Proposed changes

Closes #10321 
With this pull request we can use  `reportedEnvironmentVars` in the config instead of using `addEnvironment` in tests

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] Documentation update

## Checklist

## Further comments


christian-bromann:
Thanks for raising the issue.

I am a bit concerned removing `addEnvironment` entirely because it could potentially break other peoples test. I would recommend better deprecate it and log a warning that the method will be removed soon. Does `addEnvironment` still has a function in the reporter?

m4hdyar:
> Thanks for raising the pull request.
> 
> I am a bit concerned removing `addEnvironment` entirely because it could potentially break other peoples test. I would recommend better deprecate it and log a warning that the method will be removed soon. Does `addEnvironment` still has a function in the reporter?

`addEnvironment` still has a function, but when you use it a second time, it will overwrite the first value. This means you cannot add more than one environment variable. So, if you only need to add one environment variable, it works correctly, although that is not usually the case. 
If we keep it, it will also override environment variables that you write in the config.

I understand the importance of not disrupting people's tests. But the error is relatively easy to fix. However, in my opinion, deprecating addEnvironment without correcting its functionality would result in incorrect functionality without displaying any error, leading to unpredictable behavior for users.

On the other hand, if we choose to deprecate and also correct the behavior of `addEnvironment`, it would involve introducing a significant amount of unnecessary complexity. 

What do you think?

christian-bromann:
> However, in my opinion, deprecating addEnvironment without correcting its functionality would result in incorrect functionality without displaying any error, leading to unpredictable behavior for users.

Why we can't deprecated and notify people at the same time about the error, even give them some instructions how to use the new interface. I am not sure why keeping `addEnvironment` imposes more complexity but I would rather want to keep it with a deprecation warning.

BorisOsipov:
@m4hdyar @christian-bromann

The best solution for me would be to deprecate addEnvironment and display a warning message with a link to the documentation. Additionally, we should modify addEnvironment to have no functionality at all. This approach will alert users to switch to the new method and provide an easy way to migrate. Of course, we may encounter some issues where users complain that addEnvironment is not working, but there is no better solution available.

christian-bromann:
Sounds reasonable @BorisOsipov  👍

packages/wdio-allure-reporter/src/common/api.ts
```diff
@@ -144,11 +144,11 @@ export function addTag(tag: string) {
...
  * Add environment value
  * @name addEnvironment
- * @param {string} name - environment name
- * @param {string} value - environment value
+ * @deprecated addEnvironment is deprecated. Use reportedEnvironmentVars in config instead.
  */
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
 export function addEnvironment (name: string, value: string) {
-    tellReporter(events.addEnvironment, { name, value })
+    console.warn('addEnvironment is deprecated and has no longer any functionality. Use reportedEnvironmentVars in wdio config instead. Read more in https://webdriver.io/docs/allure-reporter. ')
```
christian-bromann, line 151:
```suggestion
    console.warn('⚠️ addEnvironment is deprecated and has no longer any functionality. Use reportedEnvironmentVars in wdio config instead. Read more in https://webdriver.io/docs/allure-reporter.')
```

christian-bromann (approved):
LGTM 👍 

@BorisOsipov any comments?

BorisOsipov (approved):

BorisOsipov:
LGTM

christian-bromann:
@m4hdyar mind signing the CLA?

christian-bromann:
That should work, thanks!

Issue #10358, opened 2023-05-09
[🐛 Bug]: @wdio/allure-reporter cucumber step reporter adds blank steps before and after each scenario 


JordiSAGE:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

latest

### Node.js Version

v18.16.0

### Mode

Standalone Mode

### Which capabilities are you using?

```typescript
package.json:

  "dependencies": {
    "@cucumber/cucumber": "^9.1.0",
    "@types/chai": "4.3.4",
    "@types/cucumber": "^7.0.0",
    "@types/node": "^18.16.0",
    "@types/source-map-support": "^0.5.6",
    "@wdio/allure-reporter": "^8.10.1",
    "@wdio/cli": "^8.10.1",
    "@wdio/config": "^8.10.1",
    "@wdio/cucumber-framework": "^8.10.1",
    "@wdio/devtools-service": "^8.10.1",
    "@wdio/dot-reporter": "^8.10.1",
    "@wdio/local-runner": "^8.10.1",
    "@wdio/shared-store-service": "^8.10.1",
    "@wdio/spec-reporter": "^8.10.1",
    "@wdio/types": "^8.10.1",
    "allure-commandline": "2.21.0",
    "bytenode": "1.4.0",
    "chai": "~4.3.7",
    "cucumber-tag-expressions": "2.0.3",
    "devtools": "8.8.4",
    "dotenv": "^16.0.3",
    "fs-copy-file-sync": "1.1.1",
    "image-size": "^1.0.2",
    "npm": "^9.6.4",
    "npm-check-updates": "^16.10.8",
    "pdfkit": "^0.13.0",
    "puppeteer": "^19.9.1",
    "rimraf": "5.0.0",
    "source-map-support": "0.5.21",
    "wdio-cucumber-reporter": "^0.0.2",
    "wdio-typescript-service": "0.0.3",
    "wdio-wait-for": "^3.0.3",
    "webdriverio": "^8.10.1"
  }

Allure report config:

    reporters: [
        'dot',
        'spec',
        [
            'allure',
            {
                outputDir: path.resolve(cfgReportDir + '/allure-results'),
                disableWebdriverStepsReporting: true,
                disableWebdriverScreenshotsReporting: false,
                useCucumberStepReporter: true,
[...]

Issue #9928, opened 2023-03-10
[🐛 Bug]: If use before/suiteSetup hook, then the first test would be duplicated in allure report


7artur7:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

8.5.8

### Node.js Version

16.14.2 (OS X 12.5)

### Mode

Standalone Mode

### Which capabilities are you using?

```typescript
capabilities: [{
        maxInstances: 5,
        browserName: 'chrome',
        acceptInsecureCerts: true
    }],
```

### What happened?

If use `before` hook (or `suiteSetup` if you are using `TDD` interface) using `mocha` test framework, then the first test would be duplicated inside of `allure` report. If change hook to `beforeEach` or `setup` (for `TDD`) then test wouldn't be duplicated inside of report.

https://imgur.com/SLre3Oc.png - result with `before` hook
https://imgur.com/K8owHqc.png - result with `beforeEach`

If add steps inside of test, then one test will contain the steps, but another one would be missing them. 

### What is your expected behavior?

Tests aren't duplicated inside of report

### How to reproduce the bug.

Setup a new project 

```
npm init wdio ./path/to/new/project -- --yes
```

Install `allure` reporter

```
npm install @wdio/allure-reporter --save-dev
```

Install `allure` cli

```
npm install -g allure-commandline --save-dev
```

Add `allure` to `wdio.conf.js` that was automatically generated in the root folder:

```
    reporters: [
        'spec',
        ['allure', {
            outputDir: 'report/allure-results',
            disableWebdriverStepsReporting: true,
            disableWebdriverScreenshotsReporting: true,
        }]],
```

Open test that was automatically generated in `./test/specs/example.e2e.js` and update it to use `before` hook. You can place `await LoginPage.open()` into the hook: 

```
import LoginPage from '../pageobjects/login.page.js'
import SecurePage from '../pageobjects/secure.page.js'

describe('My Login application', () => {
    before(async () => {
        await LoginPage.open()
    })
    it('should login with valid credentials', async () => {

        await LoginPage.login('tomsmith', 'SuperSecretPassword!')
        await expect(SecurePage.flashAlert).toBeExisting()
        await expect(SecurePage.flashAlert).toHaveTextContaining(
            'You logged into a secure area!')
    })
})
```

[...]

PR #10475, opened 2023-05-30
@wdio/allure-reporter cucumber step reporter adds blank steps before and after each scenario bug fixing

JordiSAGE:
## Proposed changes

PR for fixing the bug: https://github.com/webdriverio/webdriverio/issues/10358

## Types of changes

- [ ] Documentation update

## Checklist


JordiSAGE:
> @JordiSAGE All Cucumber hooks have empty title, thats why you have empty steps in report. If you exclude all empty steps from report users get lost all information attached to such steps(images, params, link, etc). It seems unacteaptable

This is just limited to the hook steps with empty title, but in that case, what do you suggest in order to fix the issue? Maybe to limit the lines 544 and 578 when (!disableMochaHooks)?
Thanks for the suggestion!

JordiSAGE:
> > @JordiSAGE All Cucumber hooks have empty title, thats why you have empty steps in report. If you exclude all empty steps from report users get lost all information attached to such steps(images, params, link, etc). It seems unacteaptable
> 
> @BorisOsipov This is just limited to the hook steps with empty title on the onHookStart and onHookEnd events, but in that case, what do you suggest in order to fix the issue? Maybe to limit the lines 544 and 578 when (!disableMochaHooks)? Thanks for the suggestion!

BorisOsipov:
@JordiSAGE 

Reporting hooks in Allure can be challenging. I have spent a few days trying to make it work effectively within the scope of https://github.com/webdriverio/webdriverio/issues/9928, but without success.

I would suggest keeping the hooks reported. However, if the hook.title is empty and it is a cucumber framework with useCucumberStepReporter set to true, I recommend forcing the hook title to be set as "Unnamed cucumber hook."

I understand that it may not be the ideal solution and that it can result in "empty" hooks for some users. However, this approach allows for the inclusion of custom steps within hooks and attachments, providing more flexibility and functionality.

dstoutnutrien:
@BorisOsipov Could there be a `disableCucumberHooks` option which defaults to false, so users who don't want those empty steps can remove them explicitly?

BorisOsipov:
@dstoutnutrien 

Yes, it is a great option. However, imagine a scenario where you don't care about failed hooks because you don't use them. But suddenly, you end up using them and have the disableCucumberHooks option enabled. This could result in false positive report results and introduce bugs into production.

What is the best solution: keeping useless hooks in the report forever or missing out on bugs in the system under test due to an ugly report?

Another solution is for contributors to thoroughly check all possible cases when a before/after hook fails with different artifacts. It is important to ensure that none of these cases are missed, especially when a failed hook has no effect on reporting a scenario as failed.

BorisOsipov:
I think we should always have an excellent report for scenarios in which tests fail, for the sake of instances when they are successful. Even if they have some unsightly empty steps.

rsivasubramaniam:
@JordiSAGE @BorisOsipov 

Hello Fellas,

Having used + customized the Java version of Allure -- My 2 cents below as the **Java + Allure + Cuke** version is much more mature than the **WDIO + Allure + Cuke** version

1. Yes the

BorisOsipov:
@dstoutnutrien

>In what case might reporting that blank step/hook be helpful?

In all cases :) It may contain test related data 
image

I've checked with current fix tests aren't run at all with error smth about missing current step\test. So it is definitially breaking change.

Also in case of failed After hook test marked as `Passed` - it is ridiculous.
image

JordiSAGE:
As @dstoutnutrien mentioned, I can't see the case in which these blank lines would be helpful, even the onHookEnd already includes the test fail creation in case of error, and honestly, the lines 544 and 578 seem to be added by mistake.

JordiSAGE:
@BorisOsipov Following your suggestions, I've reverted the previous changes, but alternatively I've added the disableMochaHooks checking sentence, that in case that this is false, it will not show the hook as passed, but still keeping it in case of error, I hope it will match your expectations.
Thanks for your review!

image

image

christian-bromann:
@BorisOsipov good to go?

BorisOsipov:
@christian-bromann unfortunately, I am not sure. I will flook into this tomorrow.

BorisOsipov:
@JordiSAGE 

> alternatively I've added the disableMochaHooks checking sentence, that in case that this is false, it will not show the hook as passed (when disableMochaHooks is set to true), but still keeping it in case of error, I hope it will match your expectations.


I am not happy with such an approach. First of all, disableMochaHooks  should not affect any cases unrelated to Mocha integration. As a Cucumber user, such an approach is non-obvious.

Secondly, I still insist on reporting all hooks in Cucumber to avoid cases where custom steps in hooks are not being reported. To achieve this, the best approach is to report all hooks. Due to a limitation in Cucumber, we cannot obtain the hook names. Therefore, if the `hook.title` is empty and it is a cucumber framework with `useCucumberStepReporte`r set to true, I recommend forcing the hook title to be set as "Unnamed cucumber hook" or simply "Unnamed hook". It will allow reporting all hooks, even if their "body" is empty and seemingly useless (although they may contain some user steps, attachments, parameters, etc.).

I understand that such an approach may make the report a bit ugly and verbose for you, but for the common case, it is appropriate because we need to cover all user cases.

I want to doubly emphasize that I am aware the report may look a bit ugly, but we have to maintain backward compatibility.

BorisOsipov (changes requested):
See comment in PR

JordiSAGE:
Hi again @BorisOsipov and @christian-bromann 

It can be a good solution, putting the message "Unnamed cucumber hook" or "Unnamed hook", in case to keep coherence in the code and to avoid cases where custom steps in hooks are not being reported, as you mentioned, but this is not clear functionally speaking, since the end users will see a blank space or the messages you proposed, those would be completely irrelevant as well as confusing for them, hope you understand, other than that, again from my point of view, this should have been discussed as part of the bug itself, or when it has been reported, even rejecting it as a bug, thus avoiding for having to provide the different solutions, again I hope that makes sense, so at this point I should insist on removing the blank passed steps (with or without messages), as you mentioned when `useCucumberStepReporter` is set to true, the `hook.title` is empty, and additionally by checking that the hook doesn't contain any steps, attachments, parameters, etc.
For doing so, I've removed the previous changes, and made the required change to only show the onTestStart step when the title is not empty or it contains any attachment and exclusively for when the `useCucumberStepReporter` variable is set to true, otherwiser keeping the steps as before.

Thank you again for your support and patience!

BorisOsipov:
>t I should insist on removing the blank passed steps (with or without messages), as you mentioned when useCucumberStepReporter is set to true, the hook.title is empty, and additionally by checking that the hook doesn't contain any steps, attachments, parameters, etc.

Nice. Just do it :)

>For doing so, I've removed the previous changes, and made the required change to only show the onTestStart step when the title is not empty or it contains any attachment and exclusively for when the useCucumberStepReporter variable is set to true, otherwiser keeping the steps as before.

These changes aren't effective. You simply moved the check for hook.title from one function to another. Additionally, the datatable for hooks is always empty. But the worst part is that this code still leads to corrupted reports. 
You should never allow cases where you don't call startStep/Test(testTitle) but call endStep/Test() 
and vice versa. It leads to a broken test-step-substep hierarchy and a broken report.

It seems that I haven't explained the problem very clearly. Let me do it.
There is a repository with an example available at https://github.com/BorisOsipov/wdio-allure-10475

In this case, we have before and after hooks where I used the Allure API to enrich the report with additional data.
https://github.com/BorisOsipov/wdio-allure-10475/blob/master/steps.ts#L21-L40
https://github.com/BorisOsipov/wdio-allure-10475/blob/master/steps.ts#L43-L49

Additionally, I have some hooks where I don't use the Allure API.
https://github.com/BorisOsipov/wdio-allure-10475/blob/master/steps.ts#L43-L49
https://github.com/BorisOsipov/wdio-allure-10475/blob/master/wdio.conf.ts#L56-L58

Currently, all tests have passed, and I have a correct but visually unappealing report (it contains steps with empty names).
But all my data, steps, attachments are correct placed as I wanted.
See screen: https://imgur.com/a/ni2akyA

If I make any changes to the reporter where we remove "empty" hooks from the report, I will have a partially correct report where all empty steps are gone. However, all my custom Allure steps, attachments, and other elements will be mixed into the test body, creating a chaotic and confusing presentation.

See screen: https://imgur.com/a/ZTbwtnq

So the goal of this bug fix should be to remove all empty steps from the report, but not to remove steps where I have my own created attachments, sub-steps, etc. Or even better, give the "empty" named steps some names like "Unnamed cucumber hook" or "Unnamed hook" to help users expand the step and see the sub-steps and data.

I hope it is now clear why we can't simply remove hooks where the hook.title is empty.

JordiSAGE:
Hi @BorisOsipov 

First, apologies for the mess in the code, I can see your point, and I'm doing my best to understand all the different cases for providing the best code I could as requested in the bug, but as shown I still don't have your experience to be able to do it.
I'm going to base my coming changes on the resources you gave me and provide the best possible solution.
Stay tuned!

Thank you again for your suggestions and resources!

BorisOsipov:
Hi @JordiSAGE 

Yes that's good👍
Could you add a unit test `wdio-allure-reporter/tests/cucumber.suite.test.ts` to ensure we don't run into any regression?

JordiSAGE:
Hi again @BorisOsipov 

I've added the corresponding mocha tests on cucumber test file to both test empty hooks and hooks with dependant steps attached.
Hope this coverage will be enough, fell free to let me know!

Thank you in advance.

BorisOsipov:
Hi @JordiSAGE Yes, these tests are sufficient.

Please, have a look into failed unit tests. The current implementation seems to be flawed
https://github.com/webdriverio/webdriverio/actions/runs/5380296095/jobs/9765373726?pr=10475#step:7:527

JordiSAGE:
Hello @BorisOsipov

I've fixed the issues when running the cucumber steps by modifying the code provided on onHookEnd method, apart from improve the mocha test cases, but there is still an issue on https://github.com/webdriverio/webdriverio/actions/runs/5380296095/jobs/9765373726?pr=10475#step:7:576
I can't manage to get rid of the Unhandled exception, more than to put a .catch() method on the promise, what makes the step as passed, as follows:
`it('should add failed custom step', async () => {
    await AllureReporter.step('custom step', () => {
        return Promise.reject(new Error('an error here')).catch(function () {
            console.log("Promise Rejected");
       })
    })

    const [, call] = vi.mocked(process.emit).mock.calls[0] as any
    expect(process.emit).toHaveBeenCalledTimes(1)
    expect(call.steps).toHaveLength(1)
    expect(call.steps[0]).toMatchObject({
        name: 'custom step',
        status: Status.PASSED,
        stage: Stage.FINISHED,
    })
})`

This part has been written by @christian-bromann some time ago, so maybe he can help on it, otherwise I don't know what to do to keep it as exception without throwing the unhandled error.

Thank you again!

christian-bromann (approved):
It seems like you were able to resolve that issue. The code looks good to me 👍 

@BorisOsipov should we merge?

packages/wdio-allure-reporter/tests/cucumber.suite.test.ts
```diff
@@ -189,9 +189,9 @@ describe('reporter option "useCucumberStepReporter" set to true', () => {
         })
 
         it('should have the console log add', () => {
-            expect(allureResult.steps).toHaveLength(3)
+            expect(allureResult.steps).toHaveLength(2)
             expect(allureResult.steps[1].attachments).toHaveLength(1)
-            expect(allureResult.steps[1].attachments[0].name).toEqual('Console Logs')
+            expect(allureResult.steps[1].attachments[0].name).toEqual('Page source')
```
BorisOsipov, line 194:
```suggestion
            expect(allureResult.steps[0].attachments[0].name).toEqual('Console Logs')
```

packages/wdio-allure-reporter/tests/cucumber.suite.test.ts
```diff
@@ -189,9 +189,9 @@ describe('reporter option "useCucumberStepReporter" set to true', () => {
         })
 
         it('should have the console log add', () => {
-            expect(allureResult.steps).toHaveLength(3)
+            expect(allureResult.steps).toHaveLength(2)
             expect(allureResult.steps[1].attachments).toHaveLength(1)
```
BorisOsipov, line 193:
```suggestion
            expect(allureResult.steps[0].attachments).toHaveLength(1)
```

BorisOsipov (approved):
LGTM

BorisOsipov:
>@BorisOsipov should we merge?

Yes, I fixed test a bit and now CI jobs fail on bidi generation. IDK how to fix it.

christian-bromann:
> IDK how to fix it.

I am on it and will merge this PR as soon as we have a green build again.

christian-bromann:
Great job @JordiSAGE 👏 thanks a lot!

JordiSAGE:
Thank you @BorisOsipov and @christian-bromann for your help, appreciate it!

Issue #10495, opened 2023-06-05
[🐛 Bug]: browser.waitUntil() is not throwing custom error message


nils-hoyer:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

8.8.3

### Node.js Version

v18.12.0

### Mode

WDIO Testrunner

### Which capabilities are you using?

```typescript
automationProtocol: 'devtools', 
browserName: 'chrome',
'goog:chromeOptions': {
   args: ['--incognito', '--window-size=1920,1080'],
 },
```

### What happened?

the function browser.waitUntil() is not throwing custom error message

### What is your expected behavior?

throwing the custom error message over the internal error message.

### How to reproduce the bug.

```typescript
describe('browser.waitUntil() tests', async () => {
    it('should throw custom error message', async () => {
        await browser.waitUntil(() => false, { timeoutMsg: 'timeoutMsg will be ignored' });
    });
});
```

### Relevant log output

```typescript
[chrome 113.0.0.0 win32 #0-0] browser test
[chrome 113.0.0.0 win32 #0-0]    ✖ should throw custom error message
[chrome 113.0.0.0 win32 #0-0]
[chrome 113.0.0.0 win32 #0-0] 1 failing (10.1s)
[chrome 113.0.0.0 win32 #0-0]
[chrome 113.0.0.0 win32 #0-0] 1) browser test should throw custom error message
[chrome 113.0.0.0 win32 #0-0] waitUntil condition failed with the following reason: return value was never truthy
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

christian-bromann:
Thanks for reporting!



We greatly appreciate any contributions that help resolve the bug. While we understand that active contributors have their own priorities, we kindly request your assistance if you rely on this bug being fixed. We encourage you to take a look at our [contribution guidelines](https://webdriver.io/docs/contribute) or join our friendly [Discord](http://discord.webdriver.io/) development server, where you can ask any questions you may have. Thank you for your support, and cheers!

rashiq231:
@christian-bromann, @nils-hoyer, as per the waitUntil implementation, condition uses a promise based timer, which stops when the function is resolved.
[...]

PR #10535, opened 2023-06-11
waitUntil, if given should return custom error message when condition returns always false

rashiq231:
## Proposed changes
waitUntil should return a custom error message if given, else should return a default error message when the condition always returns false ( not resolving to falsy)
This change is to resolve an issue reported in the #10495 , where waitUntil doesn't throw a custom error message if given.
With this change, we check for a condition if the given function always returns falsy and has a custom error message.
## Types of changes

- [ ] Documentation update

## Checklist

## Further comments


christian-bromann (changes requested):
I think the best solution is to do the following change in line 80 of `packages/webdriverio/src/utils/Timer.ts#L80`:

```patch
-                 return this._checkCondition(new Error('return value was never truthy'))
+                return this._checkCondition(new Error(TIMEOUT_ERROR))
```

Then the right timeout message should be thrown as defined in line 50

rashiq231:
updated as per your comment. Could you please verify now?

packages/webdriverio/tests/commands/browser/waitUntil.test.ts
```diff
@@ -191,6 +191,65 @@ describe('waitUntil', () => {
...
+        let val
+        // @ts-ignore uses expect-webdriverio
+        expect.assertions(2)
+        try {
+            val = await browser.waitUntil(()=>n, {
+                timeout: 500,
+                timeoutMsg: 'Custom error message',
+                interval: 200
+            }
+            )
```
christian-bromann, line 205:
```suggestion
            val = await browser.waitUntil(() => n, {
                timeout: 500,
                timeoutMsg: 'Custom error message',
                interval: 200
            })
```

packages/webdriverio/tests/commands/browser/waitUntil.test.ts
```diff
@@ -191,6 +191,65 @@ describe('waitUntil', () => {
...
+        let error
+        let val
+        // @ts-ignore uses expect-webdriverio
+        expect.assertions(2)
+        try {
+            val = await browser.waitUntil(()=>n, {
+                timeout: 500,
+                interval: 200
+            }
+            )
```
christian-bromann, line 224:
```suggestion
            val = await browser.waitUntil(() => n, {
                timeout: 500,
                interval: 200
            })
```

packages/webdriverio/tests/commands/browser/waitUntil.test.ts
```diff
@@ -191,6 +191,65 @@ describe('waitUntil', () => {
...
+        } catch (err: any) {
+            error = err
+        } finally {
+            expect(error).toBeUndefined()
+            expect(val).toBe(n)
+        }
+    })
+
```
christian-bromann, line 252: We have this test already in line 168
```suggestion
```

christian-bromann (approved):
LGTM 👍

christian-bromann:
@rashiq231 I think we need to update one more unit test:

```
 FAIL  packages/webdriverio/tests/utils/Timer.test.ts > timer > promise > should be rejected without promise
AssertionError: expected Error: timeout to match object Error: return value was never truthy
  - Expected  - 1
  + Received  + 1

    Error {
  -   message: 'return value was never truthy',
  +   message: 'timeout',
    }
```

rashiq231:
okay... I will update the test case for Timer.
But, a little clarification, shouldn't we need to maintain the existing error message - 'return value was never truthy', when ever the function returns false instead of resolving to false?
Shouldn't the timeout error which we updated in the timer should occur when the given time is exhausted?

christian-bromann:
The docs implicitly say:

> It expects a condition and waits until that condition is fulfilled with a truthy value to be returned.

So there is no reason to have an error message stating the obvious and use the timeout error because that is what happens at the end: the timer times out because no truthy value was returned.

christian-bromann (approved):
LGTM 👍

christian-bromann:
Congratulations on your first WebdriverIO contribution! This project can't live without the participation of the community. We would love to see more from you, so let us know if we can help to find interesting areas for you to contribute to. We appreciate you 🙏 ❤️

rashiq231:
@christian-bromann , Thank you for your help and support :)

Issue #10391, opened 2023-05-15
[🐛 Bug]: @wdio-junit-reporter working on mac os x but not windows


OluJS:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

8

### Node.js Version

18.16.0

### Mode

WDIO Testrunner

### What happened?

On my mac os x device, I'm able to get generate a populated junit report using the @wdio/junit-reporter reporter. However, cloning the same repo on a windows machine results in an 'empty' junit report.

### What is your expected behavior?

For the junit reporter to return test results when running on a windows machine

### How to reproduce the bug.

To reproduce this bug, you can clone the following repository, install the node_modules by running `npm i` and then run the tests using `npm run wdio`.

The repository -> https://github.com/OluJS/WebdriverIO-JUnit

Or you can just setup a new wdio project and choose junit as one of the reporter options

### Relevant log output

```typescript
mac os x junit report output:
 "junit" Reporter:
<?xml version="1.0" encoding="UTF-8"?>
<testsuites>
  <testsuite name="The Internet Guinea Pig Website" timestamp="2023-05-15T22:22:27" time="2.991" tests="2" failures="0" errors="0" skipped="0">
    <properties>
      <property name="specId" value="0"/>
      <property name="featureName" value="The Internet Guinea Pig Website"/>
      <property name="capabilities" value="chrome.113_0_5672_92.macosx"/>
      <property name="featureFile" value="file://./features/login.feature"/>
    </properties>
    <testcase classname="CucumberJUnitReport-chrome.113_0_5672_92.macosx.The Internet Guinea Pig Website" name="As a user I can log into the secure area" time="2.135">
      <system-out><![CDATA[
✅ Given I am on the login page✅ When I login with tomsmith and SuperSecretPassword!✅ Then I should see a flash message saying You logged into a secure area!
]]></system-out>
    </testcase>
    <testcase classname="CucumberJUnitReport-chrome.113_0_5672_92.macosx.The Internet Guinea Pig Website" name="As a user I can log into the secure area" time="0.838">
      <system-out><![CDATA[
✅ Given I am on the login page✅ When I login with foobar and barfoo✅ Then I should see a flash message saying Your username is invalid!
]]></system-out>
    </testcase>
  </testsuite>
</testsuites>


windows output:
[...]

PR #10603, opened 2023-06-20
@wdio/junit-reporter: Fixes issue with junit reporter not working on windows

lukefitz1:
## Proposed changes

This is a bug fix for [this bug](https://github.com/webdriverio/webdriverio/issues/10391). 

The reporter checks that the filepath for the spec file matches the expected filepath, but on Windows, due to forward slashes and 1 leading slash, the comparison will always fail. This change removes the leading slash and replaces forward slashes with backslashes when tests are running on windows. Tested on macOS (Ventura), Windows 11, and Ubuntu just to verify that the change didn't affect mac/linux users. 

## Types of changes

- [ ] Documentation update

## Checklist


christian-bromann (commented):
Thank you for looking into this.

I think instead of using string `replace` methods here we should use Node.js APIs to handle paths and transform them the way we need it, e.g. use `url.pathToFileUrl` etc. Would you mind checking this option?

lukefitz1:
@christian-bromann Thanks for the suggestion, it did end up simplifying the code, no more ternary. Please give it another look when you have a chance. Thanks!

christian-bromann:
@lukefitz1 looks good, it seems that tests are failing though, mind taking a look?

christian-bromann:
ping @lukefitz1

PR #10731, opened 2023-07-14
(webdriver): re-export command

christian-bromann:
## Proposed changes

Make what is documented [here](https://webdriver.io/docs/customcommands/#add-more-webdriver-commands) work when using CJS.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

Thanks @jlipps for reporting.

### DIFF
--- a/packages/webdriver/src/cjs/index.ts
+++ b/packages/webdriver/src/cjs/index.ts
@@ -1,6 +1,13 @@
 import type { Options } from '@wdio/types'
 import type { Client, AttachOptions } from '../types.js'
 
+function command (method: string, encodeUri: string, commandInfo: any, doubleEncodeVariables = false) {
+    return async function protocolCommand(this: unknown, ...args: unknown[]) {
+        const commandESM = await import('../command.js')
+        return commandESM.default(method, encodeUri, commandInfo, doubleEncodeVariables).apply(this, args)
+    }
+}
+
 class WebDriver {
     static async newSession(
         options: Options.WebDriver,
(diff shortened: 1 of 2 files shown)


packages/webdriver/src/cjs/index.ts
```diff
@@ -32,20 +39,19 @@ class WebDriver {
...
-    return async function protocolCommand(this: unknown, ...args: unknown[]) {
-        const commandESM = await import('../command.js')
-        return commandESM.default(method, encodeUri, commandInfo, doubleEncodeVariables).apply(this, args)
+    static get command () {
+        return command
     }
 }
+
+module.exports = WebDriver
+exports.command = command
```
jlipps, line 57:
```suggestion
module.exports.command = command
```

christian-bromann: `module.exports = WebDriver` which has `command` as static property, so it should be the same, right?

christian-bromann: the test seem to pass:

```js
const WebDriver = require('../../packages/webdriver')
const command = require('../../packages/webdriver').command

assert.equal(typeof WebDriver.command, 'function')
assert.equal(typeof command, 'function')
```

jlipps: oh I didn't see the refactor above, this was a suggestion based on the initial commit. let me try.

jlipps: yep this should work now

christian-bromann: 👍

PR #10767, opened 2023-07-19
(webdriverio): manage browser drivers

christian-bromann:
## Proposed changes

This patch will change the current behavior that would default the execution to use `devtools` as automation engine and instead will setup and run a browser driver. Ultimately we want to deprecate `devtools` as WebDriver will provide everything needed. The maintenance of the package is not manageable and given all the new features in Bidi there doesn't seem much value to have WebdriverIO users run tests through Puppeteer.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

Missing items:

- [x] write unit tests for new code
- [x] add e2e tests to ensure all browser start correctly in CI
- [x] add docs on Capabilities (especially document new driver caps)
- [x] Remove driver services from cli
- [x] verify that people that still use driver services can continue to use it
- [x] write blog post


packages/webdriverio/src/driver/index.ts
```diff
@@ -0,0 +1,174 @@
...
+    } else if (caps.browserName.toLowerCase() === 'firefox') {
+        driver = 'GeckoDriver'
+        driverProcess = await startGeckodriver({
+            ...(caps['wdio:geckodriverOptions'] || {}),
+            port
+        })
+    } else if (['microsoftedge', 'edge'].includes(caps.browserName.toLowerCase())) {
+        driver = 'EdgeDriver'
+        driverProcess = await startEdgedriver({
+            ...(caps['wdio:geckodriverOptions'] || {}),
```
nextlevelbeard, line 161:
```suggestion
            ...(caps['wdio:edgedriverOptions'] || {}),
```

packages/webdriverio/package.json
```diff
@@ -78,19 +80,22 @@
     "aria-query": "^5.0.0",
     "css-shorthand-properties": "^1.1.1",
     "css-value": "^0.0.1",
-    "devtools": "8.12.1",
-    "devtools-protocol": "^0.0.1170846",
+    "decamelize": "^6.0.0",
```
erwinheitzman, line 83: aren't we missing the chromedriver package?

packages/webdriverio/src/driver/index.ts
```diff
@@ -0,0 +1,174 @@
...
+
+import type { Capabilities } from '@wdio/types'
+
+import detectBackend from './detectBackend.js'
+import { parseParams } from './utils.js'
+import { SUPPORTED_BROWSERNAMES } from '../constants.js'
+import type { RemoteOptions } from '../types'
+
+const log = logger('webdriver')
+const __dirname = url.fileURLToPath(new URL('.', import.meta.url))
```
erwinheitzman, line 26: since NodeJs already uses a global with the same name, I would change this to:
`const dirname`

erwinheitzman (commented):
Left some feedback, love seeing this 👍

packages/webdriverio/package.json
christian-bromann: The `chromedriver` NPM package installs Chromedriver on install rather then runtime which would be less ideal for this scenario (all other drivers are able to dynamically detect the browser and install the right driver). The Chrome team has worked on an NPM package (`@puppeteer/browser`) that allows dynamic installing Chromedriver and Chrome.

nextlevelbeard:
@christian-bromann Since you are working on this.
Having a hard time trying to set time zones.

Using selenium-standalone, `browser.setTimeZone` does not seem to be working for Chrome at least.

From what I gather there is no API to do this. Maybe on CDP.
One way to do it is to set the `TZ` variable when launching the browser.
Having the variable present when running the test-runner is not enough.
Can you provide a way to inject environment variables when launching the browsers?

AFAIK there is no way to do that at the moment.
For component testing there is (through Vite) but not for E2E tests.

christian-bromann:
> Can you provide a way to inject environment variables when launching the browsers?

Can you explain what you mean by that. Please create a new issue for this because it is not necessarily within the scope of this PR.

christian-bromann:
🎉

owens-ben:
hi, when does this get released?

christian-bromann:
It is, see release notes.

owens-ben:
ah thanks, sorry i was only looking at the 18.4.2 notes

PierreFritsch:
> # [v8.14.0 (2023-08-01)](https://github.com/webdriverio/webdriverio/blob/main/CHANGELOG.md#v8140-2023-08-01)
>
> 🚀 New Feature
>
> * wdio-browser-runner, wdio-cli, wdio-devtools-service, wdio-runner, wdio-selenium-standalone-service, wdio-types, wdio-utils, wdio-webdriver-mock-service, webdriver, webdriverio
>     * https://github.com/webdriverio/webdriverio/pull/10767 (webdriverio): manage browser drivers ([@christian-bromann](https://github.com/christian-bromann))

PR #10804, opened 2023-07-26
(@wdio/globals): fix method calls on global objects so that they're correctly bound

jan-molak:
## Proposed changes

At the moment, `proxyHandler` from `@wdio/globals` returns fields from global objects without binding them to their receivers:
https://github.com/webdriverio/webdriverio/blob/bd86d6975458ad8bba3fd87f2af505e1537204db/packages/wdio-globals/src/index.ts#L20-L30

This means that when the field happens to be a method referencing `this` (like the `getPuppeteer` method), it loses the context of its receiver object and doesn't work:
https://github.com/webdriverio/webdriverio/blob/caa2302eb1123d5595a01f147374d5430ea451e7/packages/webdriverio/src/commands/browser/getPuppeteer.ts#L45-L68

For example, the following code sample DOESN'T WORK as `getPuppeteer()` is not bound to `browser` and returns an `undefined`.

```typescript
import { browser } from '@wdio/globals';

const puppeteer = await browser.getPuppeteer();
// returns undefined
```

## Types of changes

This pull request addresses the issue by binding method calls to their respective receiver objects.

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/wdio-globals/src/index.ts
+++ b/packages/wdio-globals/src/index.ts
@@ -24,7 +24,12 @@ function proxyHandler (key: SupportedGlobals) {
                 throw new Error(GLOBALS_ERROR_MESSAGE)
             }
 
-            return globals.get(key)[prop]
+            const receiver = globals.get(key)
+            const field = receiver[prop]
+
+            return typeof field === 'function'
+                ? field.bind(receiver)
+                : field
         }
     }
 }


christian-bromann (commented):
@jan-molak thanks for raising the PR.

I never stumbled upon this problem. Can we add a unit test to ensure this doesn't fail us again?

jan-molak:
Hey @christian-bromann, sure thing, I was looking for a good place to do it but struggled to find it. Are there any existing specs in this are I could extend, or should I make something new?

christian-bromann:
I think `packages/webdriverio/tests/commands/browser/getPuppeteer.test.ts` is the best place to test the change. I wonder if this bug was the reason why I called `browser.getPuppeteer.call(...)` instead of just `browser.getPuppeteer(...)`

jan-molak:
Looks like `packages/webdriverio/tests/commands/browser/getPuppeteer.test.ts` uses `remote` to instantiate a browser rather than the global `browser` object, so doesn't trigger the bug 🤔.

For example, this test works with and without the patch:
```typescript
    it('should work', async () => {
        browser = await remote({
            baseUrl: 'http://foobar.com',
            capabilities: {
                browserName: 'chrome',
            },
        })
        const pptr = await browser.getPuppeteer()
        expect(typeof pptr).toBe('object')
    })
```

Are there any tests specifically around `@wdio/globals` that I could extend?

christian-bromann:
I recommend adding a smoke test then. It runs a "fake" session but allows you to import the browser via `@wdio/globals`. You can just add a test case to the mocha one: `tests/mocha/test.ts`

jan-molak:
Hey @christian-bromann, so it looks like adding tests in this area might be a bit challenging.

I started by adding a simple test to `tests/mocha/test.ts`:

```typescript
    it('should provide access to Puppeteer', async () => {
        const puppeteer = await global.browser.getPuppeteer()
        expect(puppeteer).toBeDefined()
    })
```

This failed with:
```
[0-0] TypeError in "Mocha smoke test.should provide access to Puppeteer"
TypeError: caps.browserVersion.split is not a function
    at Browser.getPuppeteer (file:///Users/jan/Projects/jan-molak/webdriverio/packages/webdriverio/build/commands/browser/getPuppeteer.js:95:59)
    at Browser.wrapCommandFn (file:///Users/jan/Projects/jan-molak/webdriverio/packages/wdio-utils/build/shim.js:72:38)
    at async Context.<anonymous> (file:///Users/jan/Projects/jan-molak/webdriverio/tests/mocha/test.ts:117:27)
```

## 🐛 #2 Mock config issue

I tracked it down to this line, which defines the browser version as `number` instead of 
`string`
https://github.com/webdriverio/webdriverio/blob/caa2302eb1123d5595a01f147374d5430ea451e7/packages/wdio-webdriver-mock-service/src/mocks/newSession.ts#L7

I changed that as follows:
```diff
- browserVersion: 64.0,
+ browserVersion: '79.0',
```

This is to avoid triggering this clause that prevents using Puppeteer with anything older than v79.0:

https://github.com/webdriverio/webdriverio/blob/caa2302eb1123d5595a01f147374d5430ea451e7/packages/webdriverio/src/commands/browser/getPuppeteer.ts#L98-L104

This allowed me to get further, but resulted in the following error message:

```
[0-0] Error in "Mocha smoke test.should provide access to Puppeteer"
Error: Could't find a websocket url within returned capabilities to connect to! Make sure you have "moz:debuggerAddress" set to `true` in your Firefox capabilities
    at Browser.getPuppeteer (file:///Users/jan/Projects/jan-molak/webdriverio/packages/webdriverio/build/commands/browser/getPuppeteer.js:113:23)
    at Browser.wrapCommandFn (file:///Users/jan/Projects/jan-molak/webdriverio/packages/wdio-utils/build/shim.js:72:38)
    at async Context.<anonymous> (file:///Users/jan/Projects/jan-molak/webdriverio/tests/mocha/test.ts:117:27)
```

## 🐛 #3 Error message issue

Setting `moz:debuggerAddress` to `true`, as per the error message causes URL parsing error, so the error message should be adjusted at https://github.com/webdriverio/webdriverio/blob/caa2302eb1123d5595a01f147374d5430ea451e7/packages/webdriverio/src/commands/browser/getPuppeteer.ts#L125

## 🐛 #4 Can't use Puppeteer with mock WebdriverIO

Setting `moz:debuggerAddress` to an actual debugger URL, like `http://localhost:1234` allows me to pass the configuration stage, but fails since Puppeteer connection is not mocked out by `wdio-service-mock`:

```
[0-0] system in "Mocha smoke test.should provide access to Puppeteer"
FetchError: Failed to fetch browser webSocket URL from http://localhost:1234/json/version: request to http://localhost:1234/json/version failed, reason: connect ECONNREFUSED 127.0.0.1:1234
```

christian-bromann:
Dang .. I guess testing that `this.puppeteer` is accessible is the only way to test it. Then I suggest to add a full e2e test. I suggest the best place would be `https://github.com/webdriverio/webdriverio/blob/main/e2e/wdio/headless/test.e2e.ts` then.

jan-molak:
@christian-bromann - I've added a new e2e test file as modifying `test.e2e.ts` makes Husky pre-commit checks complain about unrecognised Mocha global functions, `window`, and them like. I'm guessing that the pre-commit check was introduced after the e2e test, so they're a bit out of sync.

Anyway, hope the test is OK.

e2e/wdio/headless/puppeteer.e2e.ts
```diff
@@ -0,0 +1,12 @@
+import { describe, it } from 'mocha'
+import * as puppeteer from 'puppeteer-core'
```
christian-bromann, line 2:
```suggestion
import { Browser } from 'puppeteer-core'
```

e2e/wdio/headless/puppeteer.e2e.ts
```diff
@@ -0,0 +1,12 @@
...
+
+import { browser } from '../../../packages/wdio-globals/build/index.js'
+
+describe('WebdriverIO', () => {
+    it('should provide access to Puppeteer', async () => {
+        const puppeteerInstance = await browser.getPuppeteer()
+
+        expect(puppeteerInstance).toBeInstanceOf(puppeteer.Browser)
```
christian-bromann, line 10:
```suggestion
        expect(puppeteerInstance).toBeInstanceOf(Browser)
```

christian-bromann (commented):
Let's import `Browser` directly if possible

christian-bromann (approved):
Awesome, lgtm 👍

PR #10816, opened 2023-07-29
(@wdio/cli): corrected TypeScript loading

jan-molak:
## Proposed changes

Change https://github.com/webdriverio/webdriverio/commit/429273377b76628dff1a691061982041b13f5dd3 introduced a regression where TypeScript files get loaded using both `-r ts-node/register` AND `--loader ts-node/esm/transpile-only`, preventing source maps from working correctly:
https://github.com/webdriverio/webdriverio/blob/54cbada9d273982612181f2bfae21394e2cd8d6e/packages/wdio-cli/src/commands/run.ts#L180

Even though [one part of ts-node documentation](https://github.com/TypeStrong/ts-node#mocha-7-and-newer) seems to indicate that specifying both flags _should be_ supported, another part talks about specifying [only one of the flags at the time](https://github.com/TypeStrong/ts-node#other).  

Looking at the source code of [`ts-node/esm/transpile-only`](https://github.com/TypeStrong/ts-node/blob/47d4f45f35e824a2515e17383a563be7dba7d8ff/esm/transpile-only.mjs#L7), we can see that the method [`registerAndCreateEsmHooks`](
https://github.com/TypeStrong/ts-node/blob/47d4f45f35e824a2515e17383a563be7dba7d8ff/src/esm.ts#L102-L107) already calls `register(opts)`, which means specifying the flag `-r ts-node/register` is redundant.

From what I've observed, using both flags results in TS Node incorrectly mapping the resulting JavaScript to the original TypeScript, which means that stack traces point to incorrect lines.

I've modified `@wdio/cli` running locally on a CJS project to remove the `ts-node/register` flag and the stack traces started to work correctly again.  

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/e2e/wdio/headless/ts-node.e2e.ts
+++ b/e2e/wdio/headless/ts-node.e2e.ts
@@ -1,7 +1,7 @@
-import { describe, it } from 'mocha'
-
+// eslint-disable-next-line no-undef
 describe('WebdriverIO', () => {
 
+    // eslint-disable-next-line no-undef
     it('should support source maps when running TypeScript files', async () => {
         const location = getInvocationLocation()
 
--- a/e2e/wdio/headless/ts-node.e2e.ts
+++ b/e2e/wdio/headless/ts-node.e2e.ts
@@ -1,0 +1,16 @@
+import { describe, it } from 'mocha'
+
+describe('WebdriverIO', () => {
+
+    it('should support source maps when running TypeScript files', async () => {
+        const location = getInvocationLocation()
+
+        expect(location).toMatch(/ts-node.e2e.ts:6:26/)
+    })
+})
+
+function getInvocationLocation(): string {
+    const error = new Error('example error')
+    const lines = error.stack.split('\n')
+    return lines[2]
+}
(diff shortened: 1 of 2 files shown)


nextlevelbeard:
Great work on this @jan-molak! 
Can confirm I'm also seeing incorrect stack-traces.
But not sure the tests that were added here are the best approach, each time someone changes the file it might need updating which is not ideal.

jan-molak:
Thanks! 

> But not sure the tests that were added here are the best approach, each time someone changes the file it might need updating which is not ideal.

I added a separate test file to limit issues like that, but if you have another suggestion I'm happy to review the proposed approach!

christian-bromann (approved):
I have been adding this because it was the only way to use ESM import statement with TypeScript in an CJS project. Anyway, @erwinheitzman has mentioned issues with this too so happy to revert it.

Thanks 👍

jan-molak:
> it was the only way to use ESM import statement with TypeScript in an CJS project. 

Yeah, it looks like mixing CJS and ESM is quite a challenge. I'm starting to understand @sindresorhus with his ESM-only approach. Maybe that's what we should consider for WebdriverIO 9 and Serenity/JS 4 ;-)

Issue #9495, opened 2022-12-30
[🐛 Bug]: Since v8 --spec only works with full path to spec file


synedra-mpe:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

latest

### Node.js Version

16

### Mode

WDIO Testrunner

### What happened?

Before v8 it was possible to use the `--spec` option with only mentioning the spec file. wdio then figured out, which spec file to load, depending on den given specs glob pattern config. This was very convenient to quickly limit test execution to just one spec file and not the whole suite.

### What is your expected behavior?

Entering, for configured specs glob pattern `test/itests/js/specs/mainProduct/**/*Test.js` 

`wdio run --spec LoginTest.js` should be the same as eg.

`wdio run --spec test/itests/js/specs/mainProduct/LoginTest.js`

### How to reproduce the bug.

Make use of the `--spec` cli option in a wdio project without giving the full path to the spec js file.

### Relevant log output

```
2022-12-30T16:00:31.536Z WARN @wdio/config:ConfigParser: pattern test/itests/js/specs/mainProduct/**/*Test.js did not match any file
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

christian-bromann:
Thanks for reporting!



Any contributions that resolves the bug are highly appreciated. Please take a look into our [contribution guidelines](https://webdriver.io/docs/contribute) and let us know if you have any questions. Cheers!

ifero:
I have the same issue. Wanted to bump wdio version to v8+
but after the upgrade it doesn't find any spec anymore with the same error as @synedra-mpe
In v7 I just had to run the test-suite without specifying any `--spec` details
In v8 it fails

```
WARN @wdio/config:ConfigParser: pattern __tests__/e2e/features/**/*.features did not match any file
```

ifero:
I believe it has something to do with the changes in [THIS LINE](https://github.com/webdriverio/webdriverio/blob/e053d0cf8e0da651237a45f40ce19fe61dbc031e/packages/wdio-config/src/lib/ConfigParser.ts#L399)
which now includes `rootDir` too as parameter

ebonom-dazn:
[...]

PR #10882, opened 2023-08-07
Allow to run single spec file without full path (--spec)

jemishgopani:
## Proposed changes
These changes should address this issue: https://github.com/webdriverio/webdriverio/issues/9495

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments


jemishgopani:
Hi @christian-bromann 

Can you please take a look on this?

packages/wdio-config/src/lib/ConfigParser.ts
```diff
@@ -203,11 +205,13 @@ export default class ConfigParser {
...
-        if (exclude.length > 0) {
-            this._config.exclude = this.setFilePathToFilterOptions(exclude, this._config.exclude!)
+        if (!ignoreSpecs) {
+            if (spec.length > 0) {
+                this._config.specs = this.setFilePathToFilterOptions(spec, this._config.specs!)
+            }
+            if (exclude.length > 0) {
+                this._config.exclude = this.setFilePathToFilterOptions(exclude, this._config.exclude!)
+            }
         }
```
christian-bromann, line 215:
```suggestion
        if (!ignoreSpecs && spec.length > 0) {
            this._config.specs = this.setFilePathToFilterOptions(spec, this._config.specs!)
        }
        if (!ignoreSpecs && exclude.length > 0) {
            this._config.exclude = this.setFilePathToFilterOptions(exclude, this._config.exclude!)
        }
```

packages/wdio-config/src/lib/ConfigParser.ts
```diff
@@ -37,6 +37,7 @@ interface MergeConfig extends Omit<Partial<TestrunnerOptionsWithParameters>, 'sp
 }
 
 export default class ConfigParser {
+    #ignoreSpec = false
```
christian-bromann, line 40: Can we add a comment mentioning for what this is being used? I can see that in some month this will be difficult to understand

packages/wdio-config/src/utils.ts
```diff
@@ -182,7 +181,7 @@ export function makeRelativeToCWD (files: (string | string[])[] = []): (string |
 
         returnFiles.push(file.startsWith('file:///')
             ? url.fileURLToPath(file)
-            : path.resolve(process.cwd(), file))
+            :file)
```
christian-bromann, line 184:
```suggestion
            : file)
```

christian-bromann (commented):
Some comments but looks good to me

packages/wdio-config/src/lib/ConfigParser.ts
jemishgopani: Thanks for the suggestion!!

packages/wdio-config/src/lib/ConfigParser.ts
jemishgopani: Added

packages/wdio-config/src/utils.ts
jemishgopani: Updated!

jemishgopani:
@christian-bromann Addressed your comments.

jemishgopani:
Hi @christian-bromann 
I have fixed test can you please run workflow?

tests/smoke.runner.js
```diff
@@ -89,7 +89,7 @@ const mochaAsyncTestrunner = async () => {
         'mochaAsyncTestrunner',
         path.resolve(__dirname, 'helpers', 'command.hook.config.js'),
         {
-            spec: ['./tests/mocha/test-async.ts']
+            spec: ['../mocha/test-async.ts']
```
christian-bromann, line 92: This really doesn't make sense, does it?

packages/wdio-config/src/utils.ts
```diff
@@ -182,7 +181,7 @@ export function makeRelativeToCWD (files: (string | string[])[] = []): (string |
 
         returnFiles.push(file.startsWith('file:///')
             ? url.fileURLToPath(file)
-            : path.resolve(process.cwd(), file))
+            : file)
```
christian-bromann, line 184: Maybe this change should be reverted?

jemishgopani: I have tried to make minimal changes can you please review again?

packages/wdio-config/src/utils.ts
```diff
@@ -182,7 +182,9 @@ export function makeRelativeToCWD (files: (string | string[])[] = []): (string |
 
         returnFiles.push(file.startsWith('file:///')
             ? url.fileURLToPath(file)
-            : path.resolve(process.cwd(), file))
+            : (file.includes('/'))
```
christian-bromann, line 185: Can you explain this change?

jemishgopani: Here is what happens if we use only the spec file name `test.spec.js` in cli this method will return the spec file path with CWD but we need only the file name to run the test. and the second case if we give the path of spec `./src/spec/test.spec.js` in cli then we need the spec path with CWD path to find the relevant file.

So I have added a check like if we see any `/` in --spec cli then return `path.resolve(process.cwd(), file))` otherwise return only spec file name.

jemishgopani: And I have checked this is working as expected if you have any other suggestions then please give me.

jemishgopani: @christian-bromann  I have checked on Windows and it is working perfectly. And also I have added a comment as per your suggestion please take a look again.

christian-bromann:
Thank you! I will review and merge once I get the build green again.

jemishgopani:
> Thank you! I will review and merge once I get the build green again.

Thank You!!

packages/wdio-config/src/utils.ts
erwinheitzman: This change does not work for me when testing on my Windows machine and it also breaks passing relative paths for me. Tested on Windows 11, Node 18, with both the Windows Terminal and CMD.

erwinheitzman:
I found that instead of updating this, the makeRelativeToCWD function can be removed as the glob package already handles this for you when correctly configured. I feel that this is where we should fix it instead

packages/wdio-config/src/utils.ts
jemishgopani: Hi @erwinheitzman 
Thanks for checking this I have added fix can you please check?

erwinheitzman:
Just did some research and using the matchBase property of the glob package, passing `my-file.js` will find the test file from the root (cwd) which already resolves this :)

christian-bromann:
Thanks for the tip @erwinheitzman , I think that makes the solution a bit more simpler. I raised a new PR with changes suggested by @jemishgopani and you, see https://github.com/webdriverio/webdriverio/pull/10921

Issue #10860, opened 2023-08-03
[🐛 Bug]: Worker starts with downloading the chromedriver from incorrect path what leads to a failure


HannaTarasevich:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

8.14.4

### Node.js Version

18.17.0

### Mode

WDIO Testrunner

### Which capabilities are you using?

```typescript
"@wdio/cli": "^8.14.4",
    "@wdio/cucumber-framework": "^8.14.0",
    "@wdio/local-runner": "^8.14.3",
    "@wdio/spec-reporter": "^8.14.0"
```

### What happened?

I setup the project from the scratch today (3 August 2023) with JS, Cucumber, local-runner and spec-reporter. After the successful message in the console, I run "npm run wdio" and this job failed due to an error (please find repository here https://github.com/HannaTarasevich/wdio-cucumber-bug):
![image](https://github.com/webdriverio/webdriverio/assets/79576380/e43e6674-0493-4518-acae-3ea0d19bc55f)

According to my investigation, there is incorrect link to Chromedriver download. I'm using Chrome locally, current version is 115.0.5790.171, but WDIO tries downloading the same version of Chromedriver, but actually there is no such version (Chrome for Testing also does not have this version, only desktop has):
![image](https://github.com/webdriverio/webdriverio/assets/79576380/d6e8fa4b-b829-4175-819d-5af9da320a7f)

### What is your expected behavior?

WDIO should take correct version from https://googlechromelabs.github.io/chrome-for-testing/ and https://github.com/GoogleChromeLabs/chrome-for-testing#json-api-endpoints, not using Chrome desktop versions

### How to reproduce the bug.

1) Please clone repo https://github.com/HannaTarasevich/wdio-cucumber-bug
2) Run npm/yarn install
3) Run "npm run wdio"

### Relevant log output

```typescript
"C:\Program Files\nodejs\npm.cmd" run wdio

> wdio
> wdio run ./wdio.conf.js


Execution of 1 workers started at 2023-08-03T19:29:27.307Z

2023-08-03T19:29:27.335Z INFO @wdio/cli:launcher: Run onPrepare hook
2023-08-03T19:29:27.340Z INFO @wdio/cli:launcher: Run onWorkerStart hook
[...]

PR #10899, opened 2023-08-08
[Bugfix] Fix Chromedriver download in case of lack of specified version

HannaTarasevich:
## Proposed changes

This pull request is aimed to fix the bug https://github.com/webdriverio/webdriverio/issues/10860 available version of Chrome for desktop is 115.0.5790.171, but the stable Chromedriver version is 115.0.5790.170. Previously WDIO checked if Chromedriver is installed based on the Chrome version installed. But the current Chrome logic is that a different build version for stable Chrome and Chromedriver is possible, and this leads to a WDIO error in attempting to download Chromedriver 115.0.5790.171.

Current changes include the following Chromedriver flow:

1. If there is a downloaded specified Chromedriver version (build number based on installed Chrome)
2. If not downloaded - check if it is possible to download this specified Chromedriver version
3. If not possible - check if the stable Chromedriver version has already been downloaded
4. If not downloaded - it downloads the stable Chromedriver version

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/webdriver/src/driver/utils.ts
+++ b/packages/webdriver/src/driver/utils.ts
@@ -8,12 +8,13 @@ import logger from '@wdio/logger'
 import { getChromePath } from 'chrome-launcher'
 import {
     install, canDownload, resolveBuildId, detectBrowserPlatform, Browser, ChromeReleaseChannel,
-    computeExecutablePath, type InstallOptions
+    computeExecutablePath, type InstallOptions, type BrowserPlatform
 } from '@puppeteer/browsers'
 import type { EdgedriverParameters } from 'edgedriver'
 import type { Capabilities, Options } from '@wdio/types'
 
 import { DEFAULTS } from '../constants.js'
+import fsp from 'node:fs/promises'
 
 const log = logger('webdriver')
 const EXCLUDED_PARAMS = ['version', 'help']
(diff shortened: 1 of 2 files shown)


packages/webdriver/src/driver/utils.ts
```diff
@@ -114,6 +115,49 @@ export async function setupChrome(caps: Capabilities.Capabilities, cacheDir: str
...
+    const installOptions: InstallOptions & { unpack?: true } = {
+        ...chromedriverOptions,
+        cacheDir,
+        platform,
+        buildId,
+        browser: Browser.CHROMEDRIVER,
+        unpack: true,
+        downloadProgressCallback: (downloadedBytes: number, totalBytes: number) => downloadProgressCallback('Chromedriver', downloadedBytes, totalBytes)
+    }
+    if (!(await hasChromedriverInstalled(chromedriverBinaryPath))) {
```
christian-bromann, line 137: No need for `hasChromedriverInstalled` to be a function.
```suggestion
    const hasChromedriverInstalled = await fsp.access(chromedriverBinaryPath)
        .then(() => true, () => false)
    // Check if specified Chromedriver version is installed
    const installOptions: InstallOptions & { unpack?: true } = {
        ...chromedriverOptions,
        cacheDir,
        platform,
        buildId,
        browser: Browser.CHROMEDRIVER,
        unpack: true,
        downloadProgressCallback: (downloadedBytes: number, totalBytes: number) => downloadProgressCallback('Chromedriver', downloadedBytes, totalBytes)
    }
    if (!hasChromedriverInstalled) {
```

packages/webdriver/src/driver/utils.ts
```diff
@@ -114,6 +115,49 @@ export async function setupChrome(caps: Capabilities.Capabilities, cacheDir: str
...
+        // Check if specified version of Chromedriver is available for download
+        const isCombinationAvailable = await canDownload(installOptions)
+        // If specified chromedriver version is not installed and there is no specified version to download, work with the latest stable version
+        if (!isCombinationAvailable){
+            const stableChromedriverBuildId = await resolveBuildId(Browser.CHROMEDRIVER, platform, ChromeReleaseChannel.STABLE)
```
christian-bromann, line 142: This assumes that the user wants to test on a stable release. What happens this error appears for an older version?

christian-bromann (commented):
Thanks for taking a stab to solve this bug. My only concern is that we should make sure to respect the Chrome version we have set up or identified prior setting up Chromedriver. For example what happens the user sets a Chrome version which has no Chromedriver correspondent?

HannaTarasevich:
> Thanks for taking a stab to solve this bug. My only concern is that we should make sure to respect the Chrome version we have set up or identified prior setting up Chromedriver. For example what happens the user sets a Chrome version which has no Chromedriver correspondent?

Great point, thanks! I'll continue work on this point

christian-bromann:
@HannaTarasevich due to the amount of bug reports that came in I tried to come up with a new approach, see #10902

PR #10902, opened 2023-08-09
(webdriver): find good known version when Chromedriver can't be installed with found `buildId`

christian-bromann:
## Proposed changes

We have observed in CI that a Chrome browser is found on the machine, e.g. `v115.0.5790.171` but a driver with the same `buildId` doesn't exist. In this case let's make the driver manager a bit more versatile and have it detect the next known good version.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

n/a


christian-bromann:
@harsha509 yes, this is the same case. When WebdriverIO detects a Chrome browser on the machine it doesn't seem to be guaranteed that there is a Chromedriver with the same `buildId`. For this case this patch will find a last good known version for it to execute.

packages/webdriver/src/driver/index.ts
```diff
@@ -85,17 +86,33 @@ export async function startWebDriver (options: Options.WebDriver) {
...
+                 */
+                log.warn(`Couldn't download Chromedriver v${buildId}: ${err.message}, trying to find known good version...`)
+                const majorVersion = buildId.split('.')[0]
+                const knownGoodVersions: any = await got('https://googlechromelabs.github.io/chrome-for-testing/known-good-versions.json').json()
+                const knownGoodVersion = knownGoodVersions.versions.filter(({ version }: { version: string }) => version.startsWith(majorVersion)).pop()
+                if (!knownGoodVersion) {
+                    throw new Error(`Couldn't find known good version for Chromedriver v${majorVersion}`)
+                }
+                loggedBuildId = knownGoodVersion.version
+                return install({ ...chromedriverInstallOpts, buildId: loggedBuildId })
```
HannaTarasevich, line 115: Can we have a case when known good version has already been installed and can be used from cache?

HannaTarasevich (commented):
Based on this PR, I'm closing this one https://github.com/webdriverio/webdriverio/pull/10899

packages/webdriver/src/driver/index.ts
HannaTarasevich: So this way, console won't contain message about the version of Chromedriver in use, right?
Like we have for else flow in row 118

PR #10905, opened 2023-08-10
(@wdio/selenium-standalone-service): good bye 👋

christian-bromann:
## Proposed changes

With v8.14.0 and the new [driver management](https://webdriver.io/blog/2023/07/31/driver-management) this service is no longer needed and can be removed from the code repository.

## Types of changes

- [x] Cleanup (non-breaking change which fixes an issue)
- [ ] Documentation update

## Checklist

## Further comments

n/a


christian-bromann:
@erwinheitzman yes, the package will stay around, I will just deprecate it.

answer-equals-42:
@christian-bromann @erwinheitzman I am kinda new to the testing domain. Please excuse me, if this is a stupid question, but how do I properly migrate? We were using some older 8- and up versions.

The thing is we want to make sure that both the local setup and the setup on GitLab CI are running selenium and not for example chromium. I had some issues with the setup once I removed the "@wdio/selenium-standalong-service" and the config.services where it used to say "config.services = ['selenium-standalone']".

It used to run on a selenium server on localhost. Now it is telling me, that it runs on the DevProtocol using Pupeteer and chromedriver. I've also asked a question on StackOverflow: https://stackoverflow.com/questions/77949565/how-to-properly-replace-wdio-selenium-standalone-service-service.

Do we need to configure something additionally? I've read that there is Google Chrome for Testing now. Is the new setup now using this browser when I use it locally? I don't quite get it.

Thank you in advance for the help.

erwinheitzman:
Hi @answer-equals-42 no need to apologise, could you please share the logs of a run?

erwinheitzman:
Apologies, I was in a meeting when I typed my previous message and I just noticed that you added a part of the log over on Stackoverflow.

If it is running with the devtools protocol it means you either did not upgrade to a version that has the built-in driver manager yet or perhaps you are using the devtools service.
Could you share your package.json and config?

erwinheitzman:
@answer-equals-42 I think it is a local thing on your end as I cannot reproduce it with that repo. For me it is running just fine cloning that and installing the packages

answer-equals-42:
@erwinheitzman It is running now, because I have modified config.services from undefined to an empty array. However, I wanted to know if this still runs on selenium because it says chromedriver in the logs and refers to my local chrome installation as you can see in the logs. Is it now using chromedriver or selenium?

image

erwinheitzman:
I think that Selenium isn't what you think it is. WebdriverIO either uses Webdriver or Devtools (protocols), the selenium server was used in the past only to download and start the drivers, which is now built into webdriverio itself.

You are running WebdriverIO with the same protocol as before (webdriver) instead of devtools.

Regarding the local browser binary, you can add the browserVersion property to your capabilities. I recommend to set it to "stable" and if you run your tests with this, it will go ahead and download the latest stable Chrome for Testing along with the matching driver for it.

Good luck 😎

PR #10972, opened 2023-08-19
 deprecating array of args support in appium service

tamil777selvan:
## Proposed changes

deprecating array of args support in appium service 

## Types of changes

- [x] Documentation update

## Checklist

## Further comments


packages/wdio-appium-service/src/utils.ts
```diff
@@ -23,6 +23,36 @@ export function getFilePath (filePath: string, defaultFilename: string): string
...
+                } else {
+                    acc[key.replace(/^-+/g, '')] = nextItem
+                }
+            }
+            return acc
+        }, {})
+    }
+
+    return args
+}
```
christian-bromann, line 54: Let's return early
```suggestion
export function defragCliArgs(args?: AppiumServerArguments | Array<string>) {
    if (!Array.isArray(args)) {
        return args
    }

    return args.reduce((acc: KeyValueArgs, currentItem, index, array) => {
        const nextItem = array[index + 1]

        if (
            !currentItem?.toString().startsWith('--') &&
            !currentItem?.toString().startsWith('-')
        ) {
            return acc
        }

        const [key, value] = (currentItem ?? '').toString().split('=')

        if (value !== undefined) {
            acc[key.replace(/^-+/g, '')] = value
        } else if (
            nextItem &&
            (nextItem?.toString().startsWith('--') ||
                nextItem?.toString().startsWith('-'))
        ) {
            acc[key.replace(/^-+/g, '')] = true
        } else {
            acc[key.replace(/^-+/g, '')] = nextItem
        }
        return acc
    }, {})
}
```

christian-bromann (commented):
I think it would be more clear if the `defragCliArgs` method could be incorporated into `formatCliArgs`. Also can we have more tests for the new function?

tamil777selvan:
> I think it would be more clear if the `defragCliArgs` method could be incorporated into `formatCliArgs`. Also can we have more tests for the new function?

At first, I considered changing the `formatCliArgs` function instead of creating a new one. However, because `_args` is used in `_setCapabilities`, which also needs the `port` value, I ended up needing this new method.

Also as suggested, I have updated the new method and added more tests to check it's functionality. Please share your thoughts on this.

christian-bromann:
@tamil777selvan I am a bit confused. It seems to be that with `defragCliArgs` we now support to pass in Appium args as object __and__ array. If this is the case I suggest to only go with one option since it can get confusing otherwise. E.g. update the docs and make sure check the type of the `_args` object in the service.

tamil777selvan:
> @tamil777selvan I am a bit confused. It seems to be that with `defragCliArgs` we now support to pass in Appium args as object **and** array. If this is the case I suggest to only go with one option since it can get confusing otherwise. E.g. update the docs and make sure check the type of the `_args` object in the service.

Hi @christian-bromann, The [documentation](https://webdriver.io/docs/appium-service#args) explains its capability to accept objects/arrays as inputs for 'args'. Could we consider discontinuing the array support in this scenario? Please share your opinion.

christian-bromann:
> Could we consider discontinuing the array support in this scenario? Please share your opinion.

Yeah, I would suggest that. Given array support never worked, we can safely discontinue it. It is less confusing to only support 1 type here. Would you mind making that change?

tamil777selvan:
> > Could we consider discontinuing the array support in this scenario? Please share your opinion.
> 
> Yeah, I would suggest that. Given array support never worked, we can safely discontinue it. It is less confusing to only support 1 type here. Would you mind making that change?

Yeah sure, on it.

tamil777selvan:
Hi @christian-bromann, I've eliminated the array support for arguments and enforced the type to be an object. Please share your opinion.

packages/wdio-appium-service/README.md
```diff
@@ -86,9 +86,9 @@ export const config = {
 Map of arguments for the Appium server, passed directly to `appium`.
 
 See [the documentation](https://appium.github.io/appium.io/docs/en/writing-running-appium/server-args/index.html) for possible arguments.
-The arguments should be supplied in lower camel case, so `--pre-launch true` becomes `preLaunch: true` or passed as an array.
+The arguments are supplied in lower camel case. For instance, `debugLogSpacing: true` transforms into `--debug-log-spacing true`, or they can be supplied as outlined in the Appium documentation.
```
christian-bromann, line 89: Boolean flags should not have a value.
```suggestion
The arguments are supplied in lower camel case. For instance, `debugLogSpacing: true` transforms into `--debug-log-spacing`, or they can be supplied as outlined in the Appium documentation.
```

Can you link _Appium documentation_ to an Appium resource that could be of help here?

packages/wdio-appium-service/README.md
```diff
@@ -102,10 +102,10 @@ export const config = {
                 // ...
                 debugLogSpacing: true,
                 platformName: 'iOS',
+                '-p': 1234,
+                '--relaxed-security': true
```
christian-bromann, line 106: I don't think we should document such usage and encourage users to write out the full property name.

packages/wdio-appium-service/tests/launcher.test.ts
```diff
@@ -388,6 +389,16 @@ describe('Appium launcher', () => {
...
         })
+
+        test('should result in an error when args is not of object type', async () => {
+            const options = {
+                args: []
+            }
+            const launcher = new AppiumLauncher(options, [], {} as any)
+            const error = await launcher.onPrepare().catch((err) => err)
+            const expectedError = new Error('Args should be an object')
+            expect(error).toEqual(expectedError)
```
christian-bromann, line 400:
```suggestion
            const expectedError = new Error('Args should be an object')
            await expect(launcher.onPrepare()).rejects.toEqual(expectedError)
```

packages/wdio-appium-service/README.md
tamil777selvan: Understood, but would this apply equally to aliases as well?

packages/wdio-appium-service/README.md
tamil777selvan: Furthermore, the documentation link is connected in line 88; however, it points to the outdated Appium version. Would it be advisable to directly link to the documentation [here](https://github.com/appium/appium/blob/master/packages/appium/docs/en/cli/args.md) on GitHub?

christian-bromann: yes, that makes sense

packages/wdio-appium-service/README.md
christian-bromann: I am not sure if we should support these, I think if someone e.g. sets `{ p: 1234 }` we can create `['-p', 1234]` out of it. What do you think?

tamil777selvan: I've observed that this approach isn't applicable to all aliases. 

For example: if I have `{ p: 1234 }`, following the formatting, it would result in `['--p', 1234]`, which Appium would accept.

However, consider a scenario where I provide an address in the alias, such as `{ a: 'foo' }`. The formatting would lead to `['--a', 'foo']`, triggering an ambiguous error in Appium.

To address this situation, it becomes necessary to either permit the use of full property names in the args or to consider discontinuing the support by updating the documentation. Please share your opinion.

christian-bromann: From the user experience point of view I feel like this makes the configuration more clear anyway. What do you think?

tamil777selvan: Certainly, that sounds like a solid plan. I'll integrate those adjustments and make the necessary updates to the documentation.

tamil777selvan: Hi @christian-bromann, I have implemented all the required modifications. Please review them at your convenience.

packages/wdio-appium-service/src/launcher.ts
```diff
@@ -118,6 +118,13 @@ export default class AppiumLauncher implements Services.ServiceInstance {
     }
 
     async onPrepare() {
+        /**
+         * Throws an error if `this._options.args` is defined and is an array.
+         * @throws {Error} If `this._options.args` is an array.
+         */
+        if (this._options.args && Array.isArray(this._options.args)) {
```
christian-bromann, line 125: this will be sufficient
```suggestion
        if (Array.isArray(this._options.args)) {
```

christian-bromann (commented):
Since we are already at this, could we extend the `AppiumServerArguments` and actually fill it up with given server properties rather than just having: `[capability: string]: any`. The rest looks good to me!

tamil777selvan:
> Since we are already at this, could we extend the `AppiumServerArguments` and actually fill it up with given server properties rather than just having: `[capability: string]: any`. The rest looks good to me!

I have included all the parameters and corresponding types utilised by the Appium server CLI. Please review them at your convenience.

christian-bromann (approved):
This is awesome, thank you so much 👍

PR #10976, opened 2023-08-20
Allow to run single spec file without full path (--spec)

jemishgopani:
## Proposed changes
- I have added this type of change again @christian-bromann because the earlier fix is not working as expected if we don't have `wdio.conf.js` in the project root dir.
- previous fixes only work when we have `wdio.conf.js` in the project root dir. So I have added the fix to search specific spec and file path patterns in project root dir.
Previous [PR#10921](https://github.com/webdriverio/webdriverio/pull/10921)
Issue [ISSUE#9495](https://github.com/webdriverio/webdriverio/issues/9495)
## Types of changes

- [ ] Documentation update

## Checklist

## Further comments


christian-bromann (commented):
I don't think this is the right solution to this. This makes the `cwd` always the root dir which is not always the case.

jemishgopani:
> I don't think this is the right solution to this. This makes the `cwd` always the root dir which is not always the case.

Can you please give me an example in which case this will not work?

christian-bromann:
Currently WebdriverIO is able to pick up a spec file by name e.g. `--spec foobar.test.ts`. This changeset (without having thoroughly tested it) does not seem like the right approach because you overwrite `this._config.rootDir` with `this.#projectRootDir` which always is `process.cwd()`. Currently the `rootDir` property [is defined](https://github.com/webdriverio/webdriverio/blob/20abfa8958c76740e1c5d3865eada03cc3c6a0c9/packages/wdio-config/src/lib/ConfigParser.ts#L57) to be the directory where the config file is located. All files will be resolved based on that.

With your change I am afraid that files won't be resolved properly anymore if the config is e.g. inside a `test` folder. For example:

```
<root>
  - /tests
    - wdio.conf.js
    - /specs
      - foobar.e2e.js
```

if the `wdio.conf.js` defines its specs via `./specs/**/*.e2e.js` and you run the wdio command from root, it won't find any specs because it tries to resolve the relative path from root.

jemishgopani:
> Currently WebdriverIO is able to pick up a spec file by name e.g. `--spec foobar.test.ts`. This changeset (without having thoroughly tested it) does not seem like the right approach because you overwrite `this._config.rootDir` with `this.#projectRootDir` which always is `process.cwd()`. Currently the `rootDir` property [is defined](https://github.com/webdriverio/webdriverio/blob/20abfa8958c76740e1c5d3865eada03cc3c6a0c9/packages/wdio-config/src/lib/ConfigParser.ts#L57) to be the directory where the config file is located. All files will be resolved based on that.
> 
> With your change I am afraid that files won't be resolved properly anymore if the config is e.g. inside a `test` folder. For example:
> 
> ```
> <root>
>   - /tests
>     - wdio.conf.js
>     - /specs
>       - foobar.e2e.js
> ```
> 
> if the `wdio.conf.js` defines its specs via `./specs/**/*.e2e.js` and you run the wdio command from root, it won't find any specs because it tries to resolve the relative path from root.

I'll go through this today later and let's see what can I do.

jemishgopani:
@christian-bromann I've tried to fix can you please take a look again?

packages/wdio-config/src/lib/ConfigParser.ts
```diff
@@ -37,6 +37,7 @@ interface MergeConfig extends Omit<Partial<TestrunnerOptionsWithParameters>, 'sp
 }
 
 export default class ConfigParser {
+    #projectRootDir = process.cwd()
```
christian-bromann, line 40: this seems to be a static value that never changes which means we should not attach it to the class

packages/wdio-config/src/lib/ConfigParser.ts
```diff
@@ -258,18 +259,18 @@ export default class ConfigParser {
         const multiRun = this._config.multiRun
         // when CLI --spec is explicitly specified, this._config.specs contains the filtered
         // specs matching the passed pattern else the specs defined inside the config are returned
-        let specs = ConfigParser.getFilePaths(this._config.specs!, this._config.rootDir, this._pathService)
-        let exclude = ConfigParser.getFilePaths(this._config.exclude!, this._config.rootDir, this._pathService)
+        let specs = this.findSpecs(this._config.specs!)
```
christian-bromann, line 262: Can we just amend the `ConfigParser.getFilePaths` instead of introducing a new function?

packages/wdio-config/src/lib/ConfigParser.ts
```diff
@@ -320,21 +321,17 @@ export default class ConfigParser {
...
-                this._config.rootDir,
-                this._pathService
-            )
+            const globMatchedFiles = <string[]> this.findSpecs(this._pathService.glob(filteredFile, this.#projectRootDir))
             if (this._pathService.isFile(filteredFile)) {
                 filesToFilter.add(
                     this._pathService.ensureAbsolutePath(
                         filteredFile,
-                        path.dirname(this.#configFilePath)
+                        this.#projectRootDir
```
christian-bromann, line 334: files should always resolve based on the location of the config file not the root path

christian-bromann:
@jemishgopani can you mind update your branch?

jemishgopani:
> 

Yes, butneed some time. I can make it draft for now.

erwinheitzman:
This has now been resolved and can be closed iirc right @christian-bromann ?

christian-bromann:
> * because the earlier fix is not working as expected if we don't have `wdio.conf.js` in the project root dir.

@erwinheitzman I think for @jemishgopani the fix doesn't help due to this

jemishgopani:
Hi @christian-bromann 
This PR is ready for review

Thanks!!

jemishgopani:
@christian-bromann can I get review here?

packages/wdio-config/src/lib/ConfigParser.ts
```diff
@@ -63,7 +63,11 @@ export default class ConfigParser {
          * rather than relative to the config file
          */
         if (_initialConfig.spec) {
-            _initialConfig.spec = makeRelativeToCWD(_initialConfig.spec) as string[]
+            // Find dir path for single spec using glob pattern
+            const specsDirPath = _initialConfig.spec[0].includes(path.sep) || _initialConfig.spec[0].includes('/')
+                ? process.cwd()
+                : path.dirname(path.join(process.cwd(), _pathService.glob(_initialConfig.spec[0], process.cwd())[0]))
+            _initialConfig.spec = makeRelativeToCWD(_initialConfig.spec, specsDirPath) as string[]
```
christian-bromann, line 70: I don't really understand this, e.g. the following section:

```
_initialConfig.spec[0].includes(path.sep) || _initialConfig.spec[0].includes('/')
```

Why do we need to check for both here? In which cases is `path.sep` not sufficient?

Also we are only looking at `_initialConfig.spec[0]` but what if there are multiple specs in this array from different directories?

erwinheitzman: Agreed, and correct me if I am wrong but I am pretty sure the specs should be relative to the config and not use process.cwd()

erwinheitzman:
@jemishgopani we both took a look

erwinheitzman:
@jemishgopani anything we can do to help you get this merged?

jemishgopani:
@erwinheitzman 
Sorry, I took long because I was occupied with some other tasks, please give me some time I'm going to draft this pull request for now and  I'll take a look this weekend

erwinheitzman:
@jemishgopani friendly reminder 😃

PR #10995, opened 2023-08-22
(@wdio/utils): support setup of Firefox browser through @puppeteer/browser

christian-bromann:
## Proposed changes

This patch enables downloading and setting up Firefox Nightly if `browserVersion` is set to `latest`.

## Types of changes

- [x] Polish (non-breaking change which fixes an issue)
- [ ] Documentation update

## Checklist

## Further comments

n/a

### DIFF
--- a/packages/wdio-utils/src/driver/index.ts
+++ b/packages/wdio-utils/src/driver/index.ts
@@ -21,7 +21,7 @@ import {
 } from './utils.js'
 import { SUPPORTED_BROWSERNAMES } from '../constants.js'
 
-export type ChromedriverParameters = InstallOptions & Omit<EdgedriverParameters, 'port' | 'edgeDriverVersion' | 'customEdgeDriverPath'>
+export type ChromedriverParameters = Partial<InstallOptions> & Omit<EdgedriverParameters, 'port' | 'edgeDriverVersion' | 'customEdgeDriverPath'>
 declare global {
     namespace WebdriverIO {
         interface ChromedriverOptions extends ChromedriverParameters {}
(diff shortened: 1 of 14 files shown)


packages/wdio-utils/src/driver/index.ts
```diff
@@ -107,6 +107,12 @@ export async function startWebDriver (options: Options.WebDriver) {
         /**
          * Firefox
          */
+        const { executablePath } = await setupPuppeteerBrowser(cacheDir, caps)
+        caps['moz:firefoxOptions'] = deepmerge(
+            { binary: executablePath },
+            caps['moz:firefoxOptions'] || {}
+        )
+        delete caps.browserVersion
```
nextlevelbeard, line 115: I understand `geckodriver` is very specific with the versions passsed to session creation requests, but won't we loose valuable information for all kinds of reporters and services by deleting the `browserVersion`?

FYI I worked around this by getting semver info and modifying the version like so:
```suggestion
    // Set proper version in capabilities
    // Geckodriver does not like 0-based patch versions
    caps.browserVersion = `${major}.${minor}${patch ? `.${patch}` : ''}`
```

christian-bromann: Good point, yeah would prefer to keep the version. I will look get this added.

christian-bromann: I actually looked into where we pass information through and it seems all reporter get the `browserVersion` from the capabilities object returned by the driver. Deleting them doesn't harm that.

erwinheitzman: Apologies, seems like I forgot to re-build after switching. It was indeed related and have fixed the cause. I have also made it so that the edge capabilities are only added when on Windows.

PR #11010, opened 2023-08-25
Adopt latest Cucumber API interface

tamil777selvan:
## Proposed changes

These modifications enable the utilisation of the latest Cucumber API interface for execution.

## Types of changes

- [x] Documentation update

## Checklist

## Further comments


nextlevelbeard:
Great to see work on this 👍 personal favorite feature I wanted
PS: fixes #11010

packages/wdio-cucumber-framework/README.md
```diff
@@ -141,6 +104,26 @@ Only retries the features or scenarios with tags matching the expression (repeat
...
+
+Type: `Boolean`<br />
+Default: `false`
+
+### failAmbiguousDefinitions
+**Please note that this is a @wdio/cucumber-framework specific option and not recognized by cucumber-js itself**
+Treat ambiguous definitions as errors.
+
+Type: `Boolean`<br />
+Default: `false`
```
christian-bromann, line 125: Is there a reason why the order was changed? Did anything change here?

packages/wdio-cucumber-framework/README.md
```diff
@@ -44,70 +44,39 @@ Type: `String[]`<br />
...
+Example: `[path.join(__dirname, 'step-definitions', 'my-steps.js')]`
 
 ### strict
 Fail if there are any undefined or pending steps
 
 Type: `Boolean`<br />
 Default: `false`
 
-### tagExpression
+### tags
```
christian-bromann, line 79: Can we make this backwards compatible? Possibly add a deprecation warning?

christian-bromann (commented):
Fantastic work 👏 do you see anything that could potentially make user tests break after the update?

tamil777selvan:
> Fantastic work 👏 do you see anything that could potentially make user tests break after the update?

With these changes, the main area that will be affected is the capability to [skip tests](https://webdriver.io/docs/frameworks/#skipping-tests-in-cucumber) based on specific conditions. I believe this feature is widely used by many users. I'm currently working on finding a solution to address this impact and come up with an alternative approach.

packages/wdio-cucumber-framework/README.md
tamil777selvan: Not really, as these caps are specific to WebdriverIO, I've adjusted them accordingly.

packages/wdio-cucumber-framework/README.md
tamil777selvan: Sure thing, I'll make sure to add a deprecation warning to give users a heads-up about the changes and ensure backwards compatibility.

tamil777selvan:
Hi @christian-bromann, I've managed to include all the current features in this by adopting the latest version of Cucumber. Kindly take a moment to review and share your thoughts on these updates.

packages/wdio-cucumber-framework/src/constants.ts
```diff
@@ -1,50 +1,40 @@
...
-export const NOOP = function () {}
+export const FILE_PROTOCOL = 'file://'
 
-export const CUCUMBER_HOOK_DEFINITION_TYPES = [
-    'beforeTestRunHookDefinitionConfigs',
-    'beforeTestCaseHookDefinitionConfigs',
-    'afterTestCaseHookDefinitionConfigs',
-    'afterTestRunHookDefinitionConfigs',
-] as const
+const cucumberFormatter = path.resolve(url.fileURLToPath(import.meta.url), '../cucumberFormatter.js')
```
christian-bromann, line 10: Can we just use a file url string for windows and linux based OS?
```suggestion
const cucumberFormatter = url.pathToFileURL(path.resolve(url.fileURLToPath(import.meta.url), '../cucumberFormatter.js')).href
```

packages/wdio-cucumber-framework/src/index.ts
```diff
@@ -48,123 +59,106 @@ const {
...
-        const reporterOptions = {
-            capabilities: this._capabilities,
-            ignoreUndefinedDefinitions: Boolean(this._cucumberOpts.ignoreUndefinedDefinitions),
-            failAmbiguousDefinitions: Boolean(this._cucumberOpts.failAmbiguousDefinitions),
-            tagsInTitle: Boolean(this._cucumberOpts.tagsInTitle),
-            scenarioLevelReporter: Boolean(this._cucumberOpts.scenarioLevelReporter)
-        }
-        this._cucumberReporter = new CucumberReporter(this._eventBroadcaster, this._pickleFilter, reporterOptions, this._cid, this._specs, this._reporter)
+        // backwards compatibility for tagExpression usage
+        this._cucumberOpts.tags = this._cucumberOpts.tags ? this._cucumberOpts.tags : this._cucumberOpts.tagExpression
```
christian-bromann, line 134:
```suggestion
        this._cucumberOpts.tags = this._cucumberOpts.tags || this._cucumberOpts.tagExpression
```

christian-bromann: Also, should we add a deprecation warning here?

packages/wdio-cucumber-framework/src/index.ts
```diff
@@ -174,75 +168,68 @@ class CucumberAdapter {
...
+                const pickles = Gherkin.compile(doc, '', uuidFn)
+                const tags = pickles.map((pickle) => pickle.tags.map((tag) => tag.name))
+                const generatedTag = generateSkipTagsFromCapabilities(this._capabilities, tags)
+                return generatedTag.length > 0 ? generatedTag.join(' and '): []
+            }).flat(1)
+    }
+
     async init() {
+        if (this._generateSkipTags) {
+            this._cucumberOpts.tags = this._cucumberOpts.tags ? this.generateDynamicSkipTags().concat(this._cucumberOpts.tags).join(' and ') : this.generateDynamicSkipTags().toString()
```
christian-bromann, line 216:
```suggestion
            this._cucumberOpts.tags = this.generateDynamicSkipTags().concat(this._cucumberOpts.tags || []).join(' and ')
```

packages/wdio-cucumber-framework/src/index.ts
```diff
@@ -174,75 +168,68 @@ class CucumberAdapter {
...
-            this._specs = this.filterSpecsByTagExpression([this._specs]).flat(1)
-        }
-
-        if (this._specs.length === 0) {
-            this._hasTests = false
-            return this
+        if (this._cucumberOpts.tags) {
+            this._specs = this.filterSpecsByTagExpression([this._specs]).flat(
+                1
+            )
```
christian-bromann, line 225: minor formatting
```suggestion
            this._specs = this.filterSpecsByTagExpression([this._specs]).flat(1)
```

packages/wdio-cucumber-framework/src/index.ts
```diff
@@ -253,58 +240,68 @@ class CucumberAdapter {
...
-                eventBroadcaster: this._eventBroadcaster,
-                options: this._cucumberOpts as any as IRuntimeOptions,
-                supportCodeLibrary,
-                eventDataCollector: this._eventDataCollector,
-                pickleIds: this._cucumberReporter!.eventListener.getPickleIds(this._capabilities)
+            const outStream = new Writable({
+                write(chunk, encoding, callback) {
+                    callback()
+                },
+            })
```
christian-bromann, line 261: Are we using this stream anywhere else? Does it stream some important information?

tests/cucumber/test.feature
```diff
@@ -17,7 +17,7 @@ Feature: Example feature
     Scenario: Retry Check
         Then  I should fail once but pass on the second run
 
-    @skip(browserName="chrome")
+    @skip{browserName="chrome"}
```
christian-bromann, line 20: Can you provide information about the reasoning of this change?

tests/typings/cucumber/config.ts
```diff
@@ -3,9 +3,7 @@ import { expectType } from 'tsd'
 const config: WebdriverIO.Config = {
     cucumberOpts: {
         timeout: 123,
-        require: ['123'],
-        // @ts-expect-error
-        scenarioLevelReporter: 'wrong param'
```
christian-bromann, line 8: Why removing this test?

christian-bromann (commented):
Some comments. Overall very happy to see this change 🤩

packages/wdio-cucumber-framework/src/constants.ts
tamil777selvan: At first, I took the same approach. When we use `url.pathToFileURL`, the file URL comes with the `file:///` prefix with three slashes (`///`), which works well for Linux-based systems. However, Cucumber internally modifies the provided file URL and raises an error if there are three slashes (`///`) present on windows. 

Upon further investigation, I observed that Cucumber employs the split [operation](https://github.com/cucumber/cucumber-js/blob/85b1e5ed13d73d037d39641399f3fdff66faa023/src/configuration/option_splitter.ts#L5), which creates issues on Windows. As a result, I've explicitly added a double slash prefix to the file path for Windows.

packages/wdio-cucumber-framework/src/index.ts
tamil777selvan: The purpose of this stream is to actually suppress the standard output returned from cucumber-js. In real-world scenarios, this behaviour will be taken over by the `cucumberFormatter`. However, during unit testing, we are suppressing the output intentionally.

packages/wdio-cucumber-framework/src/index.ts
tamil777selvan: It's marked as deprecated in both the types and documentation. However, if necessary, we can generate a warning message as below. I would appreciate your thoughts on this approach.

```
if (this._cucumberOpts.tagExpression) {
    log.warn("'tagExpression' is deprecated. Use 'tags' instead.");
}
```

tests/typings/cucumber/config.ts
tamil777selvan: Oops, my mistake. While piecing things together to restore the old features, I accidentally missed to add it back.

tests/cucumber/test.feature
tamil777selvan: In the past, we used `parseGherkinMessageStream` beforehand, granting us the ability to filter tests based on tags or custom criteria. This allowed us to manupulate the list of `pickleId` returned to the `cucumber Runtime` for execution.

However, with this new approach, we've transitioned to use the `cucumber API` for test execution, which prevents the prior use of `parseGherkinMessageStream`. Consequently, to skip test cases based on capabilities, we now need a different method by generating a dynamic cucumber tags with a `not` condition to address this challenge.

Moreover, when we encounter tags like `@skip(browser="chrome")`, cucumber's tag expression encounters an issue due to reserved characters `(` and `)`. To overcome this problem within the tag expression, this modification is necessary. As a solution, we generate dynamic tags like `not @skip{browser="chrome"}`, which is compatible with the tag expression.

packages/wdio-cucumber-framework/src/index.ts
christian-bromann: Your suggestion is a good approach. I believe there are a lot of non TS users out there that won't see the type deprecation. While this log message probably will be hard to spot as well, it is the best we can do for now until we have a better deprecation strategy in general.

packages/wdio-cucumber-framework/src/index.ts
christian-bromann: Gotcha, let's make sure we close the stream at the end so we don't have any pending listeners hanging around.

tests/cucumber/test.feature
christian-bromann: Gotcha. That makes sense. I just thinking how much this will impact users Cucumber tests and whether merging this PR requires a minor or major version bump.

tamil777selvan: It would probably be best to categorise this as a major change, mainly due to the shift in how things are executed underneath.

christian-bromann: I agree, but this would mean we can't merge it until we work and plan to release WebdriverIO v9. I am not worried about the changes as we have good test coverage and always can fix bugs when reported by users. If we can find a way to keep the old behavior supported I can imagine releasing this as `minor` update. Can you think of a workaround we can keep in place until we release v9?

tamil777selvan: I believe I made an error when escaping special characters. However, with the revised regex, I managed to restore the previous behaviour without causing any impact. Let me update the changes.

tamil777selvan: Hi @christian-bromann, I've made adjustments to the tags, ensuring they are returned with the escape character, and this modification has been accepted by the tag-expression. Thus, from a user standpoint, there shouldn't be any significant changes. Kindly take a look at these revisions and let me know your thoughts.

christian-bromann (approved):
This is an excellent change set, fantastic job 👍

christian-bromann:
I will look at the Windows test in `main`, these failures aren't related to this PR.

christian-bromann:
@tamil777selvan just letting you know that you have picked up two issues with an `💸 Expensible` tag which means you need 8 more of those to be able to claim $1000 from our [fund](https://opencollective.com/webdriverio).

packages/wdio-cucumber-framework/src/constants.ts
davidjgoss: Sorry this is such a pain point with Cucumber!

I just released 9.5.1 of cucumber-js which includes support for expressing each format as a 1- or 2-item array instead of a string, so it's explicit what is the formatter name/path and (optionally) what is the target and there is no split operation. So with the configuration below it could be something like:

```json
format: [
  [cucumberFormatter]
]
```

And that value should be fine as a regular file URL regardless of platform - if you're able to give this a try, let me know how it goes.

Also more generally, feel free to start a thread in https://github.com/cucumber/cucumber-js/discussions/2041 with any feedback you have on the new API. I'm conscious that the docs are pretty thin still, so I'm happy to fill any gaps.

tamil777selvan: Thanks @davidjgoss for handling this. I'll make the necessary changes and will get in touch if I encounter any issues.

PR #11057, opened 2023-08-30
fix(@wdio-utils): secure driver connections when using the built-in driv…

erwinheitzman:
There's two issues with the current built-in driver manager which this PR resolves.

1. the allow-ips and allow-domains where set to a wildcard by default (empty string functions as a wild card too) which caused the chromedriver to allow remote connections
2. instead of using localhost like we do in other places, 0.0.0.0 was specifically set instead. Although our tests where working, for some reason this caused a huge slowdown (talking about 20+ seconds)

## Proposed changes

I have changed the behavior to be in line with how the selenium-standalone and chromedriver packages work.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/e2e/wdio/wdio.local.conf.ts
+++ b/e2e/wdio/wdio.local.conf.ts
@@ -40,6 +40,11 @@ export const config: Options.Testrunner = {
         'moz:firefoxOptions': {
             args: ['-headless']
         }
+    }, {
+        browserName: 'edge',
+        'ms:edgeOptions': {
+            args: ['headless', 'disable-gpu']
+        }
     }],
 
     /**
@@ -59,14 +64,5 @@ export const config: Options.Testrunner = {
 if (os.platform() === 'darwin') {
     (config.capabilities as Capabilities.Capabilities[]).push({
         browserName: 'safari'
-    })
-}
-
-if (os.platform() === 'win32') {
-    (config.capabilities as Capabilities.Capabilities[]).push({
-        browserName: 'edge',
-        'ms:edgeOptions': {
-            args: ['headless', 'disable-gpu']
-        }
     })
 }
--- a/e2e/wdio/wdio.local.conf.ts
+++ b/e2e/wdio/wdio.local.conf.ts
@@ -40,11 +40,6 @@ export const config: Options.Testrunner = {
         'moz:firefoxOptions': {
             args: ['-headless']
         }
-    }, {
-        browserName: 'edge',
-        'ms:edgeOptions': {
-            args: ['headless', 'disable-gpu']
-        }
     }],
 
     /**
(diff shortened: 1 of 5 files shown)


erwinheitzman:
For some reason Firefox 119a is downloaded which seems to have some issues starting, why are we using nightly builds by default? I suggest we download the latest stable version instead

christian-bromann:
> why are we using nightly builds by default? I suggest we download the latest stable version instead.

We don't, we have tests though that verify we are able to setup Firefox Nightly.

> It also sees the edge version as undefined when locally running on a machine that does not have Edge installed.

I will have to look, it build seems sometimes unstable which is a reason to look into.

christian-bromann:
@erwinheitzman it seems like the current `main` build is green so the changes might be related to the failing build. Note that we run some e2e tests where we test whether we can setup browser in CI.

erwinheitzman:
> @erwinheitzman it seems like the current `main` build is green so the changes might be related to the failing build. Note that we run some e2e tests where we test whether we can setup browser in CI.

Apologies, seems like I forgot to re-build after switching branches I guess even though I feel like I did.
It was indeed related and I have pushed a fix.
I have also made sure the Edge capabilities are now only added when running Windows.

erwinheitzman:
One note, I believe we can also use 127.0.0.1 instead of the localhost reference if preferred but I would have to check if that works just as well as setting localhost

e2e/wdio/wdio.local.conf.ts
```diff
@@ -66,3 +61,12 @@ if (os.platform() === 'darwin') {
         browserName: 'safari'
     })
 }
+
+if (os.platform() === 'win32') {
```
christian-bromann, line 65: We have Edge support in all OS'es so we should test the ability to set this up in all as well.

erwinheitzman: Oh right since it is Chromium based, totally forgot that. I'll take a look at it tomorrow, thanks :)

erwinheitzman:
I am really confused as to why the tests that are failing are suddenly failing though I have a feeling that it does not like that we run the driver on localhost while the rest of the codebase seems to use 0.0.0.0 instead. Do you know why this might be the case Christian?

I'm also not sure about why we use 0.0.0.0 instead of 127.0.0.1 or localhost everywhere, as localhost is what is used by most programs for knowing if you are running locally or not and programs that do support using 0.0.0.0 will internally have to map 0.0.0.0 to localhost for it to work (like all browsers internally do for example). Can you give some feedback on why we use this?

christian-bromann:
> Do you know why this might be the case Christian?

I don't which is the reason why I had to put `0.0.0.0` in the first place.

> with such a vulnerability

While it is not advised to start a server on `0.0.0.0` if it is not clearly intended to be publicly available, this is definitely __not__ a vulnerability in almost all of the cases. Your local machine is not exposing this port to the internet unless you make it so. None of your colleagues will be able to access your driver started on your local machine.

Again I am all for it to restrict the access but I haven't seen a way to successfully initiate a session otherwise.

erwinheitzman:
I'll revert some of the changes and instead of trying to change 0.0.0.0 to 127.0.0.1/localhost and will whitelist only 0.0.0.0 as a valid IP to connect so that at least the connection is secure.

This simplifies the change greatly and allows me to continue with the upgrade to v8 on my end while on our (the webdriverio side) end we can investigate how to make the transition to run on 127.0.0.1/localhost in the future as currently I cannot get all the tests green with the time that I have.

I think this is quite fair 👍

christian-bromann (approved):
LGTM 👍

Issue #5505, opened 2020-06-16
Can't use devTools with multiremote


vojtech-cerveny:
**Environment (please complete the following information):**
 - **WebdriverIO version:** 6.1.17
 - **Mode:** WDIO Testrunner
 - **If WDIO Testrunner, running sync/async:** sync/async
 - **Node.js version:** v12.13.1
 - **Browser name and version:** Chrome latest
 - **Platform name and version:** Ubuntu 18.04
 - **Additional wdio packages used (if applicable):** 
```
    "@wdio/cli": "^6.1.17",
    "@wdio/devtools-service": "^6.1.14",
    "@wdio/local-runner": "^6.1.17",
    "@wdio/mocha-framework": "^6.1.17",
    "@wdio/spec-reporter": "^6.1.14",
    "@wdio/sync": "^6.1.14",
```
**Config of WebdriverIO**
```js
  runner: 'local',
  services: ['devtools'],  
  capabilities: {
    creator: {
      capabilities: {
        browserName: 'chrome',
      },
    },
    spectator: {
      capabilities: {
        browserName: 'chrome',
      },
    },
  },
```

**Describe the bug**
It is not possible to use devTools using multiremote. If I add `cdp` fuction, it returns `creator.cdp is not a function` 

```
ERROR @wdio/devtools-service: Couldn't connect to chrome: TypeError: Cannot read property 'goog:chromeOptions' of undefined
    at findCDPInterface (/home/vojtech-cerveny/Documents/dev/bee/bee-plugin-frontend/node_modules/@wdio/devtools-service/build/utils.js:33:52)
    at DevToolsService.before (/home/vojtech-cerveny/Documents/dev/bee/bee-plugin-frontend/node_modules/@wdio/devtools-service/build/index.js:64:61)
   ....
```

**To Reproduce**
```js
/* global creator, spectator */

describe('Network monitoring', () => {
  it('should have the right title', () => {

    creator.cdp('Network', 'enable')
    creator.on('Network.responseReceived', params => {
      console.log(`Loaded ${params.response.url}`)
    })
    creator.url('http://url/')

    // I need to wait for something from Network, and parse it and use it in second remote. 
    // spectator.url('https://webdriver.io/docs/multiremote.html')
    // expect(creator).toHaveTitle('WebdriverIO · Next-gen browser and mobile automation test framework for Node.js')

    // spectator.url('https://webdriver.io/docs/multiremote.html')
[...]

PR #11059, opened 2023-08-31
devtools commands for multiremote

ccharnkij:
## Proposed changes

Continuation from [#5505](https://github.com/webdriverio/webdriverio/issues/5505), this feature allows devtools commands to be used for multiremote. The commands can be used with `browser` object itself, which produces array result for all instances, or with an individual instance.

```js
// example cap
cap = 
{
  chrome1: {
    capabilities: {
      browserName: 'chrome'
    }
  },
 chrome2: {
    capabilities: {
      browserName: 'chrome'
    }
  },
}

await browser.cdp('Network', 'getCookies') // returns an array containing result for both instance
await chrome1.cdp('Network', 'getCookies')  // returns only result for chrome1
```

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

This change is pretty much done by moving the code from devtools/index.ts to devtools/commands.ts and having a separate command object for each browser instance. All commands can be used by an individual instance. However, because I try to do this while wanting to make as little changes as possible, not all commands can be used by `browser` and those command are: `getNodeId` , `getNodeIds`, `startTracing`, `endTracing`, `getTraceLogs`, and `getPageWeight` (the commands that originally exist in commands.ts file). If needed, I can make more change to allow those commands to be used by `browser`.

Also, performance testing with lighthouse is a little bit tricky to use. Because of the way `beforeCommand` work, I don't have a way to determine which instance the command is coming from, and this can cause the code to hang because `onLoadEventFired` and `onFrameNavigated` are not triggered.

```js
// This works
await chrome1.enablePerformanceAudits()
await chrome1.url('https://webdriver.io')
await chrome1.disablePerformanceAudits()

// This will cause timeout
await chrome1.enablePerformanceAudits()
await chrome1.url('https://webdriver.io')
await chrome2.url('https://webdriver.io') // <---this line will causes chrome1 to hang because beforeCommand triggers 
                                       //  for both instances and chrome1 will get stuck waiting for onLoadEventFired to finish

```

So for performance testing, a user will have to do one instance at a time while having `enablePerformanceAudits` and `disablePerformanceAudits` before and after it.

Nothing change if a user is using normal mode.

### DIFF
--- a/e2e/wdio/headless/multiremoteTest.e2e.ts
+++ b/e2e/wdio/headless/multiremoteTest.e2e.ts
@@ -18,4 +18,22 @@ describe('main suite 1', () => {
         await browserB.keys(Key.Enter)
         await expect(browserB.$('.inputMessage')).toHaveAttribute('placeHolder', 'Type here...')
     })
+
+    it('should return devtools result', async () => {
+        await browserA.url('https://webdriver.io')
+        await browserB.url('https://google.com')
+
+        const cookiesA = await browserA.cdp('Network', 'getCookies')
+        const cookiesB = await browserB.cdp('Network', 'getCookies')
+        const cookies = await browser.cdp('Network', 'getCookies')
+
+        expect(Object.keys(cookiesA).length).toBe(1)
+        expect(Object.keys(cookiesA)).toEqual(['cookies'])
+        expect(Object.keys(cookiesB).length).toBe(1)
+        expect(Object.keys(cookiesB)).toEqual(['cookies'])
+        expect(Object.keys(cookies).length).toBe(2)
+
+        const score = await browser.checkPWA()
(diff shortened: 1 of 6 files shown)


ccharnkij:
Hi @christian-bromann, how has this been going?

christian-bromann:
@ccharnkij it seems like the coverage has dropped given all the changes, you think we can add some more unit tests to stay above treshold?

ccharnkij:
@christian-bromann this one failed due to some ECONNRESET error. Is there any way I can rerun it w/o additional commit?

ccharnkij:
@christian-bromann nvm, looks like it goes well now.

christian-bromann (approved):
Great change set , nice work 👌

PR #11071, opened 2023-09-02
(docs): add best practices guide

erwinheitzman:
## Proposed changes

I've written an extensive best practices guide based on my knowledge and experience. I do not think it is done and am planning to extend it at some point but I believe this to be a great start that can help our users to write better tests.

## Types of changes

- [x] Documentation update

## Checklist


website/docs/BestPractices.md
```diff
@@ -0,0 +1,186 @@
...
+
+When you have some asynchronous code that you want to repeat, it is important to know that not all loops can do this.
+
+The following will not click the elements in order.
+
+```js
+// 👎
+elements.forEach(async (element) => {
+    await element.click();
+});
```
christian-bromann, line 145: Actually our interface allows this and I am using it regularly e.g. to fetch a set of elements and get their text I would do:

```ts
console.log(await $$('h1').map((h1) => h1.getText()))
```

Something we should point out here is that your example above would run all clicks in parallel while you actually want to run them sequentially. We should point that out. Using array methods on elements should __not__ be discouraged.

christian-bromann (commented):
This is a great start to begin collecting best practices. Add my two cents.

website/docs/BestPractices.md
erwinheitzman: > Actually our interface allows this and I am using it regularly e.g. to fetch a set of elements and get their text I would do:
> 
> ```ts
> console.log(await $$('h1').map((h1) => h1.getText()))
> ```
> 
> Something we should point out here is that your example above would run all clicks in parallel while you actually want to run them sequentially. We should point that out. Using array methods on elements should **not** be discouraged.

You are correct with that example, however let me show you what I wanted to make our users aware of.

Given we have a table with table rows that have some text in order from 0 to 19 as show here:
Screenshot 2023-09-02 at 22 24 50

When I use forEach like shown in the screenshot, you'll see that the text is returned in a random order:
Screenshot 2023-09-02 at 22 25 15

And when using the for...of loop it returns the text as expected:
Screenshot 2023-09-02 at 22 26 56

I think we are both right in this case, what do you think?

erwinheitzman: I have updated the example that might make it more clear but I'm unsure at this point if it is actually any clearer or not

website/docs/BestPractices.md
```diff
@@ -0,0 +1,203 @@
...
+
+This guide aims to share our best practices that help you write performant and resilient tests. 
+
+## Use resilient selectors
+
+__Note:__ To find out all the possible selectors WebdriverIO supports, checkout our [Selectors](./Selectors.md) page.
+
+Use selectors that are resilient to changes in the DOM.
+
+```js reference useResilientSelectors
```
christian-bromann, line 16:
```suggestion
```js
```

website/docs/BestPractices.md
```diff
@@ -0,0 +1,203 @@
...
+Using a single strategy
+
+```js
+// 👎
+await $('table').$('tr').$('td');
+```
+
+``` js
+// 👍
+await $('table').custom$('myCustomStrategy', 'tr').$('td');
```
christian-bromann, line 37: Let's use a different selector strategy and not `custom$`.

website/docs/BestPractices.md
```diff
@@ -0,0 +1,203 @@
...
+```
+
+Another reason why you do not want to chain elements and would want to use a single selector is because it locates the elements separately and since DOM interactions are expensive, this should be avoided.
```
christian-bromann, line 40: > DOM interactions are expensive

I think this is not correct. We don't interact with the DOM, we just fetch elements. The expensive part is making the request.

website/docs/BestPractices.md
```diff
@@ -0,0 +1,203 @@
...
+```
+
+The following will click the elements in order.
+
+```js
+// 👍
+for (const element of elements) {
+    await executeSomeLogic();
+}
+```
```
christian-bromann, line 172: I don't understand this example. There is no click happening. Also the behavior is the same, it is just a different loop.

website/docs/BestPractices.md
erwinheitzman: You have a strategy in mind? Since this is the only strategy I found that I believe requires chaining?

website/docs/BestPractices.md
erwinheitzman: As shown in my example earlier the execution is not the same.
You can replicate this using browser.keys, it will run differently with the for...of loop compared to other loops.

website/docs/BestPractices.md
christian-bromann: let's say e.g. `$('header').$('>>>.component').$('aria/button')`

website/docs/BestPractices.md
christian-bromann: Yes, this I do understand because you have a for..of loop and a forEach one but the PR suggests two for loops that have the same effect.

harsha509: forEach doesn't print chars in expected order as I believe `forEach` expects sync code, so async/await has no effect when forEach is used! i don't recommend use of foeEach with async/awaits!

website/docs/BestPractices.md
erwinheitzman: That's a good one though did you know you can actually do `$('>>> header .component aria/button')`? i have personally used this a lot to make life super easy when dealing with shadow DOM 👍

erwinheitzman:
I think I resolved all the feedback 👍

website/docs/BestPractices.md
```diff
@@ -0,0 +1,223 @@
...
+await expect(button).toHaveText('Submit');
+
+// 👎
+await expect(button).toBeDisplayed();
+await expect(button).click();
+```
+
+```js
+// 👍
+await expect(button).click();
```
christian-bromann, line 139: This doesn't work, we don't have a `click` matcher

erwinheitzman:
Thanks for all the great feedback Christian!

christian-bromann (approved):
LGTM 👍 great job!

PR #11078, opened 2023-09-04
@wdio/cucumber-framework: Add original coordinates when resetting Cucumber Support Library

nextlevelbeard:
## Proposed changes

Adds [original coordinates](https://github.com/cucumber/cucumber-js/blob/ad7f17d86d8fcf99d323240946cd83095e0b3c59/src/api/types.ts#L86)
```ts
export interface ISupportCodeCoordinates {
    requireModules: string[];
    requirePaths: string[];
    importPaths: string[];
}
```
when resetting the Cucumber Support Library.
In some cases, when a user defines a custom World like so:

```ts

import type { ITestCaseHookParameter } from '@cucumber/cucumber';
import { Before, World, setWorldConstructor } from '@wdio/cucumber-framework';

export class Scenario extends World {
  /** My custom World prop */
  myProp!: { myKey: string };
}

setWorldConstructor(Scenario);

Before(async function doStuff(this: Scenario, scenario: ITestCaseHookParameter) {
  this.myProp = { myKey: 'Test!' }
});
```

Cucumber follows different logic for dealing with the support library, as [seen here](https://github.com/cucumber/cucumber-js/blob/ad7f17d86d8fcf99d323240946cd83095e0b3c59/src/api/run_cucumber.ts#L34-L37):
```ts
  const supportCoordinates =
    'World' in configuration.support
      ? configuration.support.originalCoordinates
      : configuration.support
```

**This only appears to be a problem when the `parallel` option is set to a non-zero value.**

This fix includes the third argument for the original coordinates in the `reset()` call of the `SuportCodeLibraryBuilder`.

Prevents errors such as 
```ts
TypeError: Cannot read properties of undefined (reading 'getInvocationParameters')
    at Object.run (/Users/user/repos/repo/node_modules/@cucumber/cucumber/lib/runtime/step_runner.js:39:47)
    at TestCaseRunner.invokeStep (/Users/user/repos/repo/node_modules/@cucumber/cucumber/lib/runtime/test_case_runner.js:95:44)
    at TestCaseRunner.runHook (/Users/user/repos/repo/node_modules/@cucumber/cucumber/lib/runtime/test_case_runner.js:200:27)
    at /Users/user/repos/repo/node_modules/@cucumber/cucumber/lib/runtime/test_case_runner.js:171:39
    at TestCaseRunner.aroundTestStep (/Users/user/repos/repo/node_modules/@cucumber/cucumber/lib/runtime/test_case_runner.js:120:38)
    at TestCaseRunner.runAttempt (/Users/user/repos/repo/node_modules/@cucumber/cucumber/lib/runtime/test_case_runner.js:158:24)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async TestCaseRunner.run (/Users/user/repos/repo/node_modules/@cucumber/cucumber/lib/runtime/test_case_runner.js:136:35)
    at async Worker.runTestCase (/Users/user/repos/repo/node_modules/@cucumber/cucumber/lib/runtime/parallel/worker.js:82:9)
    at async Worker.receiveMessage (/Users/user/repos/repo/node_modules/@cucumber/cucumber/lib/runtime/parallel/worker.js:64:13)
```

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/wdio-cucumber-framework/src/index.ts
+++ b/packages/wdio-cucumber-framework/src/index.ts
@@ -249,7 +249,11 @@ class CucumberAdapter {
 
         try {
             await this.registerRequiredModules()
-            supportCodeLibraryBuilder.reset(this._cwd, this._newId)
+            supportCodeLibraryBuilder.reset(this._cwd, this._newId, {
+                requireModules: this._cucumberOpts.requireModule,
+                requirePaths: this._cucumberOpts.require,
+                importPaths: this._cucumberOpts.import,
+            })
 
             this.addWdioHooksAndWrapSteps(this._config, supportCodeLibraryBuilder)
 


nextlevelbeard:
@tamil777selvan Mind having a look at this?

tamil777selvan:
> @tamil777selvan Mind having a look at this?

hi @nextlevelbeard, Given that this is an options parameter, it shouldn't cause an error. Also, in your `cucumberOpts`, have you used `import` instead of `require` to import step definitions and hooks, especially if you're working with ESM?

I've tested it with the provided custom world, and it worked as expected for me. Could you provide any reproducible examples or scenarios where this issue occurs?

nextlevelbeard:
> > @tamil777selvan Mind having a look at this?
> 
> hi @nextlevelbeard, Given that this is an options parameter, it shouldn't cause an error. Also, in your `cucumberOpts`, have you used `import` instead of `require` to import step definitions and hooks, especially if you're working with ESM?
> 
> I've tested it with the provided custom world, and it worked as expected for me. Could you provide any reproducible examples or scenarios where this issue occurs?

Yep, put `parallel` as a non-zero integer.
Seems to occur if Cucumber is running in parallel, a worker [Coordinator](https://github.com/cucumber/cucumber-js/blob/ad7f17d86d8fcf99d323240946cd83095e0b3c59/src/api/runtime.ts#L10) is setup instead of the Runtime, each worker then appears to require/import the support code.

I know we are not supposed to use that feature with WDIO but that's how I found the error.
Would be nice to refactor parallelism to be compatible with Cucumber works. 

Another problem I found with the support code library in parallel, when importing the globals from `@wdio/globals` package and referencing `browser` in Cucumber hooks (Before, After), it triggers an Error:
```shell

Error: No browser instance registered. Don't import @wdio/globals outside of the WDIO testrunner context. Or you have two two different "@wdio/globals" packages installed.
Error: No browser instance registered. Don't import @wdio/globals outside of the WDIO testrunner context. Or you have two two different "@wdio/globals" packages installed.
    at Object.get (file:///Users/user/repos/repo/node_modules/@wdio/globals/build/index.js:15:23)
    at Scenario.doStuff (file:///Users/user/repos/repo/e2e/steps/hooks.steps.ts:49:36)
```

tamil777selvan:
> I know we are not supposed to use that feature with WDIO but that's how I found the error.

In this situation, we shouldn't permit the use of parallel counting to be specified in the `cucumberOpts`, and we should keep it as an internal feature.

> Would be nice to refactor parallelism to be compatible with Cucumber works.

I'm uncertain if we can immediately implement this change because the current functionality is aligned with the worker nodes that were generated by the 'runner' process and is functioning as intended.

tamil777selvan:
> Seems like each WDIO worker should correspond to one Cucumber worker when `maxInstances` is higher than 1.
> 
> At the moment, `runCucumber` is called for each worker, potentially producing separate sets of report/results, publishing a `cucumber.io` report for each worker, etc.

Then it would be a better approach not to relay on the `publish plugin` from `Cucumber` and instead we should work on extending the `wdio-reporter` interface to create a `cucumber message` report that can be shared at the end of the process.

christian-bromann:
It seems to me that WebdriverIO currently doesn't support Cucumber's parallel feature. I raised a PR (https://github.com/webdriverio/webdriverio/pull/11099) to ensure we fail if the `parallel` option is set. 

I suggest to close this PR and raise a new issue to discuss support for Cucumber parallel feature. What do you all think?

nextlevelbeard:
Created https://github.com/webdriverio/webdriverio/issues/11101

christian-bromann (approved):
LGTM 👍

PR #11120, opened 2023-09-08
(webdriverio): fix scrollIntoView

erwinheitzman:
fix scrollIntoView by aligning it with it's native counterpart for both horizontal and vertical scrolling as well as scrolling multiple times. Changed scroll and scrollIntoView duration to 0

## Types of changes

- [ ] Documentation update

## Checklist


christian-bromann:
This is amazing 🎉  let's adjust the coverage treshold if it keeps failing for Windows.

christian-bromann (approved):
LGTM 🙌

packages/webdriverio/src/commands/element/scrollIntoView.ts
```diff
@@ -61,55 +61,50 @@ export async function scrollIntoView (
...
-        windowY = Math.round(windowY)
-    }
+    // take into account the current scroll position
+    deltaX = Math.round(deltaX - scrollX)
+    deltaY = Math.round(deltaY - scrollY)
 
     try {
         return await browser.action('wheel')
-            .scroll({ duration: 200, x: windowX, y: windowY, deltaX, deltaY, origin })
+            .scroll({ duration: 0, x: deltaX, deltaY, origin: this })
```
blackforest-t, line 107: Hello, @christian-bromann, @erwinheitzman 
Could you please tell me what's the reason for using `x: deltaX` here?
Why can't we just use `deltaX`, `deltaY` instead?

erwinheitzman: I think I made when changing this, it should be y: deltaY

blackforest-t: i see. So it will be changed in future releases?
I just didn't want to create a bug for this minor thing. Scroll is working for me, but with warning and fallback to web api

erwinheitzman: Yeah I can correct this, thing is it won't show up in the release notes because we use the issues to generate the release notes.

Regarding the warning, it falls back to the native web scrollIntoView when the webdriver approach throws an error so it would be interesting to find out why it's failing at that attempt and with what error.

blackforest-t: got it. Thanks

PR #11130, opened 2023-09-11
(@wdio/utils): don't setup a driver nor browser if using devtools protocol

### LINKED ISSUE #11129: [💡 Feature]: disable automatic driver setup when using devtools automation protocol
### Is your feature request related to a problem?

Version `8.14` introduced [automatic web driver configuration](https://webdriver.io/blog/2023/07/31/driver-management). While this is great, it can be an annoyance when using the `devtools` automation protocol. My tests run in a network environment with limited access to the Internet and they fail because of this. While this can be disabled (#11128), it's counterintuitive to specify a `port` or a driver `binary` to disable the download despite it will never be used.

### Describe the solution you'd like.

At first glance, it should be as simple as wrapping this:

https://github.com/webdriverio/webdriverio/blob/aa35c291fb496b80350766d674385ea7b4d1516a/packages/wdio-cli/src/launcher.ts#L138-L144

with an `if` statement as follows:

```ts
if (config.automationProtocol !== 'webdriver') {
    await Promise.all([
        setupDriver(config, caps),
        setupBrowser(config, caps)
    ])
}
```

However, this wouldn't work with configurations where multiple automation protocols are used, for example:

```ts
export const config: WebdriverIO.Config = {
  beforeSession(config, capabilities) {
    if (capabilities.browserName === 'chrome') {
      config.automationProtocol = 'devtools'
    } else {
      config.automationProtocol = 'webdriver'
    }
  }
  // ...
}
```

Thus, in order for this to work, the `setup*()` method invocations must be moved after the `onWorkerStart` and before the actual runner is started (e.g. line `437` below):

https://github.com/webdriverio/webdriverio/blob/aa35c291fb496b80350766d674385ea7b4d1516a/packages/wdio-cli/src/launcher.ts#L431-L439

### Describe alternatives you've considered.

_No response_

### Additional context

_No response_

### Code of Conduct

christian-bromann:
## Proposed changes

When a user defines `devtools` as automation protocol we should not set up any drivers.

fixes #11129

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

n/a

### DIFF
--- a/packages/wdio-utils/src/driver/manager.ts
+++ b/packages/wdio-utils/src/driver/manager.ts
@@ -30,7 +30,12 @@ function mapCapabilities (
                 const isMultiremote = Boolean(multiremoteCaps[Object.keys(cap)[0]].capabilities)
 
                 if (isMultiremote) {
-                    return Object.values(multiremoteCaps).map((c) => c.capabilities)as Capabilities.Capabilities[]
+                    return Object.values(multiremoteCaps).map((c) => {
+                        if (c.automationProtocol === 'devtools') {
+                            return
+                        }
+                        return c.capabilities
+                    }) as Capabilities.Capabilities[]
                 } else if (w3cCaps.alwaysMatch) {
                     return w3cCaps.alwaysMatch
                 }
@@ -38,6 +43,9 @@ function mapCapabilities (
             }).flat()
             : Object.values(caps as Capabilities.MultiRemoteCapabilities).map((mrOpts) => {
                 const w3cCaps = mrOpts.capabilities as Capabilities.W3CCapabilities
+                if (mrOpts.automationProtocol === 'devtools') {
+                    return
+                }
                 if (w3cCaps.alwaysMatch) {
                     return w3cCaps.alwaysMatch
                 }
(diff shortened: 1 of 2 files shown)


packages/wdio-utils/src/driver/manager.ts
```diff
@@ -47,15 +55,19 @@ function mapCapabilities (
...
         // - we are not about to run a cloud session
         !definesRemoteDriver(options) &&
         // - we are not running Safari (driver already installed on macOS)
         !isSafari(cap.browserName) &&
         // - driver options don't define a binary path
-        !getDriverOptions(cap).binary
-    ))
+        !getDriverOptions(cap).binary &&
+        // - user is not defining "devtools" as automation protocol
+        (options as Options.WebdriverIO).automationProtocol !== 'devtools'
```
marcogrcr, line 69: Should the `setup*()` methods and this one receive an `options: Options.WebdriverIO` so you don't have to do this casting?

christian-bromann: in hindsight, yes .. good catch, will push a fix.

PR #11206, opened 2023-09-19
(webdriverio): fix findStrategy for querying elements by classname and partial text

### LINKED ISSUE #11154: [🐛 Bug]: The class locator combined with Partial Link Text is returning error for nested elements in wdio v8

### WebdriverIO Version

8.16.3

### Node.js Version

16.19.0

### Mode

WDIO Testrunner

```typescript
{
browserName: chrome,
browserVersion: 116
}
```

### What happened?

I have selectors based on the partial text, it should find me the element in which the text is.
For example, I have the following HTML:

```HTM
  <div role="tab" class="mdc-tab mat-mdc-tab">
    <span class="mdc-tab__ripple"></span>
    <div class="mat-ripple mat-mdc-tab-ripple"></div>
    <span class="mdc-tab__content">
      <span class="mdc-tab__text-label">TextToFind sample</span>
    </span>
    <span class="mdc-tab-indicator">
      <span class="mdc-tab-indicator__content" style=""></span>
    </span>
  </div>
```
  
And I have a selector to select this whole element: `.mat-mdc-tab*=TextToFind`
  
By using browser.$(<selector>) in the wdio v7, that selector gave me exactly the `<div role="tab" class="mdc-tab mat-mdc-tab">...</div>` element, and now (in v8) it gives me error (no element found). 
  
I've noticed such behavior only with class selectors combined with Link/Partial link.
But if I change, for example, the class selector to the attribute selector, like this `[role="tab"]*=TextToFind`
Then it works fine and gives exactly what I need (and what I had in v7).

To sum it up:
* selector `.mat-mdc-tab*=TextToFind` returns error, no element found.
* selector `[role="tab"]*=TextToFind` returns the whole element with nested elements where text has been found.

Which is not expected behavior for me. I'm expecting that the logic would be the same as it was in v7. I'm expecting the locator  `.mat-mdc-tab*=TextToFind` to return me the whole element with nested elements where text has been found.

### How to reproduce the bug.

Create an HTML file:
```HTM
  <div role="tab" class="mdc-tab mat-mdc-tab">
    <span class="mdc-tab__ripple"></span>
    <div class="mat-ripple mat-mdc-tab-ripple"></div>
    <span class="mdc-tab__content">
      <span class="mdc-tab__text-label">TextToFind sample</span>
    </span>
    <span class="mdc-tab-indicator">
      <span class="mdc-tab-indicator__content" style=""></span>
    </span>
  </div>
```

Try to find element by using the REPL interface:
`browser.$('.mat-mdc-tab*=TextToFind').getHTML()` 

Those actions should return HTML like: `<div role="tab" class="mdc-tab mat-mdc-tab">...</div>` but there's error instead

### Relevant log output

```typescript
N/A
```

### Code of Conduct

erwinheitzman:
fixes #11154 (a detailed explanation of the fix and the issue can be found in there)

## Proposed changes

Updated the strategy to match classnames exactly instead of partially.
Example, `$('.bar*=Find')` matches:
```
<div class="bar">
<div class="foo bar">
```
but not:
```
<div class="foo-bar">
<div class="foobar">
```

I have also installed the cross-env package as a dev dependency as it is required by the project in order to run some tests.

## Types of changes

- [ ] Documentation update

## Checklist


package.json
```diff
@@ -97,6 +97,7 @@
     "chalk": "^5.3.0",
     "cheerio": "^1.0.0-rc.12",
     "copyfiles": "^2.4.1",
+    "cross-env": "^7.0.3",
```
christian-bromann, line 100: Any reason this dependency was added?

packages/webdriverio/tests/findStrategy.test.ts
```diff
@@ -113,25 +113,25 @@ describe('selector strategies helper', () => {
     it('should find an element by tag name + class + content', () => {
         const element = findStrategy('div.some-class=some random text with "§$%&/()div=or others')
         expect(element.using).toBe('xpath')
-        expect(element.value).toBe('.//div[contains(@class, "some-class") and normalize-space(text()) = "some random text with "§$%&/()div=or others"] | .//div[contains(@class, "some-class") and not(.//div[contains(@class, "some-class") and normalize-space(text()) = "some random text with "§$%&/()div=or others"]) and normalize-space() = "some random text with "§$%&/()div=or others"]')
+        expect(element.value).toBe('.//div[contains(concat(" ",@class," "), " some-class ") and normalize-space(text()) = "some random text with "§$%&/()div=or others"] | .//div[contains(concat(" ",@class," "), " some-class ") and not(.//div[contains(concat(" ",@class," "), " some-class ") and normalize-space(text()) = "some random text with "§$%&/()div=or others"]) and normalize-space() = "some random text with "§$%&/()div=or others"]')
```
christian-bromann, line 116: Do you know why:

```
concat(" ",@class," "), " some-class ")
```
differs from
```
contains(@class, "some-class")
```
?

package.json
erwinheitzman: Yeah it is used in our project to be able to for example run some tests (the Lit component tests in my case)

packages/webdriverio/tests/findStrategy.test.ts
erwinheitzman: Yeah, contains "bar" matches "foo-bar" whereas instead we want to look for "bar".

Now you can also use @class="bar" but this has the downside that when you have an element with a class "foo bar baz" it Will not match.

The solution I applied matches a class in a classlist without matching too much (classes that only match partially)

erwinheitzman: I think I updated the PR description after your feedback was placed. I think the description shows a good example as well

christian-bromann: That makes sense, thanks! Can you confirm that `concat(" ",@class," "), " foo ")` matches all of the following elements:

- `<div class="foo bar baz"></div>`
- `<div class="bar foo baz"></div>`
- `<div class="bar baz foo"></div>`

I am asking because the selector suggests that there has to be a space left and right from the class name.

package.json
christian-bromann: I think it is only used in the `e2e` directory which is why we have it defined [here](https://github.com/webdriverio/webdriverio/blob/3670a36edb2f873d6797781ddc63fe4e5c5fca96/e2e/package.json#L48)

packages/webdriverio/tests/findStrategy.test.ts
erwinheitzman: Better yet, I'll add a testcase for it ;)

christian-bromann (approved):
Nicely done 👌

PR #11222, opened 2023-09-21
(@wdio/config): memoize ConfigParser.getSpecs

### LINKED ISSUE #4880: Cache list of spec files

erwinheitzman:
## Proposed changes

Memoized the getSpecs method of the ConfigParser to cache the results for all workers.

Fixes #4880 

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,6 +9,9 @@
       "workspaces": [
         "packages/*"
       ],
+      "dependencies": {
+        "lodash.memoize": "^4.1.2"
+      },
       "devDependencies": {
         "@octokit/rest": "^19.0.13",
         "@preact/preset-vite": "^2.5.0",
@@ -17562,6 +17565,11 @@
       "resolved": "https://registry.npmjs.org/lodash.ismatch/-/lodash.ismatch-4.4.0.tgz",
       "integrity": "sha512-fPMfXjGQEV9Xsq/8MTSgUf255gawYRbjwMyDbcvDhXgV7enSZA0hynz6vMPnpAb5iONEzBHBPsT+0zes5Z301g==",
       "dev": true
+    },
+    "node_modules/lodash.memoize": {
+      "version": "4.1.2",
+      "resolved": "https://registry.npmjs.org/lodash.memoize/-/lodash.memoize-4.1.2.tgz",
+      "integrity": "sha512-t7j+NzmgnQzTAYXcsHYLgimltOV1MXHtlOWf6GjL9Kj8GK5FInw5JotxvbOs+IvV1/Dzo04/fCGfLVs7aXb4Ag=="
     },
     "node_modules/lodash.merge": {
       "version": "4.6.2",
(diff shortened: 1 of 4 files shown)


christian-bromann:
The solution looks good to me, mind verifying if updating the `spec` parameter in the `onPrepare` or `onWorkerStart` hook still works. I am actually not sure if this currently works but I could see someone trying to modify this as part of service functionality.

erwinheitzman:
Currently it does not play nicely with the merge. We merge A LOT apparently and it looks like this is because it is making changes to the config indirectly instead of returning a new object which we could also cache.

erwinheitzman:
From doing some testing I think we would have to memoize the initialise method so that we don't execute a merge and run all of the underlying methods if nothing has changed. Will have to do more investigation though, to be continued =]

erwinheitzman:
Going to close this for now as I have to focus on other things first

PR #11244, opened 2023-09-25
(@wdio/allure-reporter): partially revert change to fix missing attachments

### LINKED ISSUE #11215: Screenshot on Failure is not attached to Allure report
Screenshot on Failure is not attached to Allure report.
I have added a reproducible example here - https://github.com/sushantsoni5392/wdio-mocha-boilerplate

I investigated a bit, it seems like the code doesn't go past line [ #493](https://github.com/webdriverio/webdriverio/blame/main/packages/wdio-allure-reporter/src/reporter.ts#L493).

`this._state.currentStep` is undefined

image

If I revert the change made in [this MR](https://github.com/webdriverio/webdriverio/commit/3bde885cabc0a39f19c58ffbd27f63d6a2c6a274), the screenshot is attached. But I do not know why this change was done in the first place.

_Originally posted by @sushantsoni5392 in https://github.com/webdriverio/webdriverio/issues/11164#issuecomment-1725603766_

erwinheitzman:
Partially reverted a previous change (https://github.com/webdriverio/webdriverio/commit/27b98700cd092b68df57849056783d5ae6b6980e) to fix #11215 

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/packages/wdio-allure-reporter/src/reporter.ts
+++ b/packages/wdio-allure-reporter/src/reporter.ts
@@ -407,13 +407,14 @@ export default class AllureReporter extends WDIOReporter {
     onTestStart(test: TestStats | HookStats) {
         const { useCucumberStepReporter } = this._options
         this._consoleOutput = ''
+        const testTitle = test.currentTest ? test.currentTest : test.title
 
         if (useCucumberStepReporter) {
             const testObj = test as TestStats
             const argument = testObj?.argument as Argument
             const dataTable = argument?.rows?.map((a: { cells: string[] }) => a?.cells)
 
-            this._startStep(test.title)
+            this._startStep(testTitle)
 
             if (dataTable) {
                 this.attachFile('Data Table', stringify(dataTable), ContentType.CSV)
@@ -421,7 +422,7 @@ export default class AllureReporter extends WDIOReporter {
             return
         }
 
-        this._startTest(test.title, test.cid)
+        this._startTest(testTitle, test.cid)
     }
 
     onTestPass() {
--- a/packages/wdio-allure-reporter/tests/allureSteps.test.ts
+++ b/packages/wdio-allure-reporter/tests/allureSteps.test.ts
@@ -28,8 +28,7 @@ describe('reporter allure steps API', () => {
         })
     })
 
-    // TODO: fails on main
-    it.skip('should add failed custom step', async () => {
+    it('should add failed custom step', async () => {
         await AllureReporter.step('custom step', () => (
             Promise.reject(new Error('an error here'))))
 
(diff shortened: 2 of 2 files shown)


christian-bromann:
> The disabled test failed both on main and my branch locally.

I don't see any failing unit tests in `main` 🤔

erwinheitzman:
I just double checked again by checking out main, doing `npm i` > `npm run setup` > `npx vitest ./packages/wdio-allure-reporter/tests`

erwinheitzman:
Screenshot 2023-09-25 at 16 42 26

erwinheitzman:
Updating to node v18.18.0 did not make a change

erwinheitzman:
Enabled the test, my assumption is that it will pass on CI but fails locally. I have more failing tests, like an element property test that returns a position of 16.5 instead of 16 for example but I never bothered looking into those.

BorisOsipov:
@erwinheitzman It is bad. By doing this, you allow the worst bugs to appear. For example, it may produce reports where a **failed** test is marked as **passed**. I belive it is better to see a correct report than to be able to attach a screenshot in some cases.

erwinheitzman:
The change was made not too long ago so I'm assuming that it has been like that for a while, if that is correct, I don't think it is that bad while currently people really seem to be running into issues with the attachments.

When a fix introduces new bugs that are quite impactful I would much rather revert the code and rethink the fix we did than to leave our users crippled in the meantime.

I want to stress that I simple reverted a change, I don't do anything new.

erwinheitzman:
@BorisOsipov this is the code I partially reverted: https://github.com/webdriverio/webdriverio/commit/27b98700cd092b68df57849056783d5ae6b6980e

erwinheitzman:
I can add back the following part too if that helps:
```js
        if (this._state.currentTest?.wrappedItem.name === testTitle) {
            // Test already in progress, most likely started by a before each hook
            return
        }
```

BorisOsipov:
>When a fix introduces new bugs that are quite impactful I would much rather revert the code and rethink the fix we did than to leave our users crippled in the meantime.

I think if a fix solves the critical level problem and introduces a medium problem, that is no reason to put users back on the critical problem. 
The worst thing is that the user can easily discover missing attachments and at the same time it is hard to discover that you have problems in the SUT when there are failed tests but the reporter shows almost all green :(

>this is the code I partially reverted: https://github.com/webdriverio/webdriverio/commit/27b98700cd092b68df57849056783d5ae6b6980e

Yes, and I don't understand why.
In the[ original issue comment](https://github.com/webdriverio/webdriverio/issues/11215#issue-1905350762) user says that revert https://github.com/webdriverio/webdriverio/commit/3bde885cabc0a39f19c58ffbd27f63d6a2c6a274
helps. And I have an understanding of why it is of help :)

erwinheitzman:
I tried that locally and that did not resolve the issue for me but hey, I know nothing about the Allure reporter, I'm just doing my best.

Regarding the change I made, I made an assumption like I said, and only if that would be true I could see this not being that bad of a change. Seems I was wrong and I'll let you handle this.

Issue #11208, opened 2023-09-19
[🐛 Bug]: in the onPrepare hook, using getValue from shared store service crashes.


nbowman-lulus:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

latest

### Node.js Version

LTS (18.18)

### Mode

WDIO Testrunner

### What happened?

In the onPrepare hook, using getValue crashes. 

### What is your expected behavior?

using getValue and setValue in the onPrepare hook as written up in the documentation https://www.npmjs.com/package/@wdio/shared-store-service should work.

### How to reproduce the bug.

Minimal repro case:
https://github.com/nbowman-lulus/wdio_test

npx wdio run .\wdio.conf.js --spec .\test\specs\test.e2e.js

Windows output:
```
npx wdio run .\wdio.conf.js --spec .\test\specs\test.e2e.js
Execution of 1 workers started at 2023-09-19T21:21:40.922Z
beginning of onPrepare

Ended WebDriver sessions gracefully after a SIGINT signal was received!
```

comment out ln 156 of wdio.conf and re-run same:

```
npx wdio run .\wdio.conf.js --spec .\test\specs\test.e2e.js

Execution of 1 workers started at 2023-09-19T21:23:29.709Z

beginning of onPrepare
[0-0] RUNNING in chrome - file:///<SNIP>/wdio_test/test/specs/test.e2e.js
[13892:20064:0919/142335.724:ERROR:policy_logger.cc(154)] :components\enterprise\browser\controller\chrome_browser_cloud_management_controller.cc(163) Cloud management controller initialization aborted as CBCM is not enabled.

DevTools listening on ws://127.0.0.1:55842/devtools/browser/82b7a1f1-bdb3-4598-bc12-0fca0d491626
[0-0] beginning test
[0-0] shared store values:
[0-0] {test: value}
```

### Relevant log output

```typescript
2023-09-19T21:21:40.966Z DEBUG @wdio/utils:initialiseServices: initialise service "shared-store" as NPM package
2023-09-19T21:21:41.036Z INFO @wdio/cli:launcher: Run onPrepare hook
2023-09-19T21:21:41.038Z INFO @wdio/local-runner: Shutting down spawned worker
[...]

PR #11254, opened 2023-09-27
Throw err if getter's is used before server init

tamil777selvan:
## Proposed changes

Proposal for #11208 

Throw an error when attempting to call `getValue`, `getValueFromPool`, `addValueToPool` before the server is initialised.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/wdio-shared-store-service/tests/client.test.ts
+++ b/packages/wdio-shared-store-service/tests/client.test.ts
@@ -41,6 +41,12 @@ describe('client', () => {
     })
 
     describe('when used in launcher process', () => {
+        it('should not return a value if the server has not been initialized', () => {
+            expect(getValue('*')).rejects.toThrowError('Attempting to use `getValue` before the server has been initialized.')
+            expect(getValueFromPool('*')).rejects.toThrowError('Attempting to use `getValueFromPool` before the server has been initialized.')
+            expect(addValueToPool('*', '')).rejects.toThrowError('Attempting to use `addValueToPool` before the server has been initialized.')
+        })
+
         it('should not post before server has started', async () => {
             const result1 = await setValue('foo', 'bar')
             const result2 = await setValue('bar', 'foo')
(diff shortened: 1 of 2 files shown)


christian-bromann (approved):
LGTM 👍

packages/wdio-shared-store-service/tests/client.test.ts
```diff
@@ -41,6 +41,12 @@ describe('client', () => {
     })
 
     describe('when used in launcher process', () => {
+        it('Error should be thrown when attempting to retrieve a value before server initialization', () => {
+            expect(getValue('*')).rejects.toThrowError('Attempting to use `getValue` before the server has been initialized.')
+            expect(getValueFromPool('*')).rejects.toThrowError('Attempting to use `getValueFromPool` before the server has been initialized.')
+            expect(addValueToPool('*', '')).rejects.toThrowError('Attempting to use `addValueToPool` before the server has been initialized.')
```
erwinheitzman, line 47: This is not retrieving, put it in a separate testcase?

erwinheitzman (approved):
Approved but I do have a suggestion for the testcase

packages/wdio-shared-store-service/tests/client.test.ts
tamil777selvan: Considering that these are negative tests and should run before invoking `setPort(port)`, which is responsible for resolving the value of `isBaseUrlReady`, I've made this changes. Please share your feedback on this. If necessary, I'm open to move this to a new test case.

christian-bromann, line 47: I agree, this should be:
```suggestion
            expect(getValueFromPool('*', '')).rejects.toThrowError('Attempting to use `addValueToPool` before the server has been initialized.')
```

christian-bromann (approved):
👍

packages/wdio-shared-store-service/tests/client.test.ts
christian-bromann: I think it makes sense this way.

Issue #11213, opened 2023-09-20
[💡 Feature]: Limited console logging for chrome driver download in WebdriverIO v8


Zarmeena:
### Is your feature request related to a problem?

Chrome browser on my machine updated and while executing the scripts, it's taking too long to download the latest chrome driver. Also, logging of each step is done which is flooding the command prompt window leading to difficulty in checking other console logs

### Describe the solution you'd like.

Can we not limit the console logging lines and update the download status differently?

Attached the current logging in Windows command prompt

![wdioV8_Logger](https://github.com/webdriverio/webdriverio/assets/46337025/196b622c-1ba8-43cd-8c90-b8e9b4e6f7dd)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

erwinheitzman:
I think we can overwrite the same line instead of posting a new line every time, would that help?

Zarmeena:
> I think we can overwrite the same line instead of posting a new line every time, would that help?

Sure @erwinheitzman . That would be better!

christian-bromann:
Thanks for raising this issue.

We greatly appreciate any contributions that help implement this feature request. While we understand that active contributors have their own priorities, we kindly request your assistance if you rely on this feature being implemented. We encourage you to take a look at our [contribution guidelines](https://webdriver.io/docs/contribute) or join our friendly [Discord](http://discord.webdriver.io/) development server, where you can ask any questions you may have. Thank you for your support, and cheers!

erwinheitzman:
Resolved by https://github.com/webdriverio/webdriverio/pull/11255

PR #11255, opened 2023-09-27
Updating driver download logs

tamil777selvan:
## Proposed changes

Solves #11213

Limit the displayed logs to indicate the download progress of drivers.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments


erwinheitzman:
Can you explain how using stdout instead of the logger resolves the issue?

tamil777selvan:
> Can you explain how using stdout instead of the logger resolves the issue?

When we utilize `process.stdout.write`, we can update the same logged line by appending `\r` at the end of the message. However, the `logger` internally relies on `console.log`, which doesn't support this functionality.

After implementing these changes, the output appears as follows. Please share your opinion on this.

https://github.com/webdriverio/webdriverio/assets/49207189/4840fa0d-3a93-42f3-bea4-26213511abaf

christian-bromann:
@tamil777selvan this is awesome. How about extending the logger with another method, e.g. `progress` which does exactly this (adding the `\r`) and includes the logging prefix?

tamil777selvan:
> @tamil777selvan this is awesome. How about extending the logger with another method, e.g. `progress` which does exactly this (adding the `\r`) and includes the logging prefix?

Sure @christian-bromann, I'll give the suggested change a try.

erwinheitzman:
I was thinking the same thing as Christian when I read your first reply. Glad to see we are all aligned and I am looking forward to the new option. Learning something new as well (the \r was new to me)

packages/wdio-utils/src/driver/utils.ts
```diff
@@ -84,7 +84,8 @@ export const downloadProgressCallback = (artifact: string, downloadedBytes: numb
         return
     }
     const percentage = ((downloadedBytes / totalBytes) * 100).toFixed(2)
-    log.info(`Downloading ${artifact} ${percentage}%`)
+    // @ts-expect-error Invoking custom loglevel
```
christian-bromann, line 87: mhm .. I wonder if we can make the logger export with the `progress` method so we don't have to do this. Given we might want to use this one elsewhere in the future it might make sense to extend the types.

packages/wdio-logger/src/node.ts
```diff
@@ -111,6 +112,14 @@ const wdioLoggerMethodFactory = function (this: log.Logger, methodName: log.LogL
     }
 }
 
+const progress = function (this: any, data: string) {
+    const level = 'progress'
+    const timestampFormatter = chalk.gray(new Date().toISOString())
+    const levelFormatter = chalk[COLORS[level]](level.toUpperCase())
+    const nameFormatter = chalk.whiteBright(this.name)
+    process.stdout.write(`${timestampFormatter} ${levelFormatter} ${nameFormatter}: ${data}\r`)
```
christian-bromann, line 120: What happens if WebdriverIO writes to a log file, e.g. if `outputDir` is set?

tamil777selvan: When the `outputDir` is configured, the progress will continue to be displayed in the terminal but won't be recorded in the log files.

packages/wdio-utils/src/driver/utils.ts
tamil777selvan: Agreed, I'll start figuring a way on extending the types to include 'progress'.

packages/wdio-logger/src/node.ts
christian-bromann: We should treat this custom log level similar to `INFO` which is why I would suggest to check the loglevel in here and don't print anything if loglevel is e.g. `ERROR`

packages/wdio-utils/src/driver/utils.ts
```diff
@@ -84,10 +84,15 @@ export const downloadProgressCallback = (artifact: string, downloadedBytes: numb
...
     }
     const percentage = ((downloadedBytes / totalBytes) * 100).toFixed(2)
-    log.info(`Downloading ${artifact} ${percentage}%`)
+    log.progress(`Downloading ${artifact} ${percentage}%`)
     lastTimeCalled = Date.now()
 }
 
+const _install = async (args: InstallOptions & { unpack?: true | undefined }) => {
+    await install(args)
+    log.progress('')
```
christian-bromann, line 93: Can you comment why we do this?

christian-bromann (commented):
One question before merging. Nice take on the docs!

packages/wdio-utils/src/driver/utils.ts
tamil777selvan: Initially, `log.progress` is used within the `downloadProgressCallback` to show the download status. To ensure that subsequent log entries maintain their formatting, it's must to clear the `log.progress` output at the end.

Unfortunately, the `install` method doesn't include a built-in callback mechanism for this purpose. Consequently, we must manually invoke `log.progress` with an empty string to clear it after the `install` method is executed.

To avoid duplicating these lines of code, I've introduced this method.

christian-bromann: Can you comment this in the code just to document this?

tamil777selvan: sure on it.

christian-bromann (approved):
LGTM 👍

PR #11274, opened 2023-09-30
add publishCucumberReport functionality

### LINKED ISSUE #7300: Allow Cucumber adapter to publish reports to reports.cucumber.io
Allow Cucumber adapter to publish reports to https://reports.cucumber.io/.
According to the documentation it's a matter of passing a simple `publish` flag to the the `cucumber-js` binary.
And yet currently, passing `publish` as `true` in webdriverIO's config ( `cucumberOpts` ) does not seem to work.

tamil777selvan:
## Proposed changes

Incorporate the `publishCucumberReport` functionality to consolidate and upload results from all workers.

Fixes #7300 

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/wdio-cucumber-framework/README.md
+++ b/packages/wdio-cucumber-framework/README.md
@@ -141,13 +141,46 @@ Default: ``
 
 ***Please note that this option would be deprecated in future. Use [`tags`](#tags) config property instead***
 
-#### profile
+### profile
 Specify the profile to use.
 
 Type: `string[]`<br />
 Default: `[]`
 
 ***Kindly take note that only specific values (worldParameters, name, retryTagFilter) are supported within profiles, as `cucumberOpts` takes precedence. Additionally, when using a profile, make sure that the mentioned values are not declared within `cucumberOpts`.***
+
+## Publishing Report
+
+Cucumber provides a feature to publish your test run reports to `https://reports.cucumber.io/`, which can be controlled either by setting the `publish` flag in `cucumberOpts` or by configuring the `CUCUMBER_PUBLISH_TOKEN` environment variable. However, when you use `WebdriverIO` for test execution, there's a limitation with this approach. It updates the reports separately for each feature file, making it difficult to view a consolidated report.
+
+To overcome this limitation, we've introduced a promise-based method called `publishCucumberReport` within `@wdio/cucumber-framework`. This method should be called in the `onComplete` hook, which is the optimal place to invoke it. `publishCucumberReport` requires the input of the report directory where cucumber message reports are stored.
+
+You can generate `cucumber message` reports by configuring the `format` option in your `cucumberOpts`. It's highly recommended to provide a dynamic file name within the `cucumber message` format option to prevent overwriting reports and ensure that each test run is accurately recorded.
+
+Before using this function, make sure to set the following environment variables:
+- CUCUMBER_PUBLISH_REPORT_URL: The URL where you want to publish the Cucumber report. If not provided, the default URL 'https://messages.cucumber.io/api/reports' will be used.
+- CUCUMBER_PUBLISH_REPORT_TOKEN: The authorization token required to publish the report. If this token is not set, the function will exit without publishing the report.
+
+Here's an example of the necessary configurations and code samples for implementation:
+
+```javascript
+import { v4 as uuidv4 } from 'uuid'
+import { publishCucumberReport } from '@wdio/cucumber-framework';
+
+export const config = {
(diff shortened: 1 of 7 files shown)


packages/wdio-cucumber-framework/README.md
```diff
@@ -141,13 +141,46 @@ Default: ``
...
+
+```javascript
+import { v4 as uuidv4 } from 'uuid'
+import { publishCucumberReport } from '@wdio/cucumber-framework';
+
+export const config = {
+    // ... Other Configuration Options
+    cucumberOpts: {
+        // ... Cucumber Options Configuration
+        format: [['message', `./reports/${uuidv4()}.ndjson`]]
```
erwinheitzman, line 174: Going through https://github.com/cucumber/messages/blob/main/messages.md#envelope and https://github.com/cucumber/messages it sounds like you can add more than just `message`, is this correct? If so, I feel like we should add another example in the format array so this is a bit more clear for users that are new to this concept.
What do you think?

erwinheitzman: Also, is the `.ndjson` a typo?

packages/wdio-cucumber-framework/src/index.ts
```diff
@@ -481,6 +483,39 @@ class CucumberAdapter {
...
+ * @async
+ * @param {string} cucumberMessageDir - The directory path that holds Cucumber NDJSON files.
+ * @returns {Promise<void>} - A Promise that resolves when the report is successfully published.
+ * @throws {Error} - Throws an error if there are issues with file reading or the publishing process.
+ */
+const publishCucumberReport = async (cucumberMessageDir: string): Promise<void> => {
+    const url = process.env.CUCUMBER_PUBLISH_REPORT_URL || 'https://messages.cucumber.io/api/reports'
+    const token = process.env.CUCUMBER_PUBLISH_REPORT_TOKEN
+    if (!token) {
+        return
```
erwinheitzman, line 497: Would be nice to let the user know that the token is missing instead of a silent ignore :)

packages/wdio-cucumber-framework/src/index.ts
```diff
@@ -481,6 +483,39 @@ class CucumberAdapter {
...
+
+    const { headers } = await got(url, {
+        headers: {
+            Authorization: `Bearer ${token}`,
+        }
+    })
+
+    const { location } = headers
+
+    const files = (await readdir(path.resolve(cucumberMessageDir))).filter((file) => path.extname(file) === '.ndjson')
```
erwinheitzman, line 508: resolve should only be needed if there's more than one argument.

```
The right-most parameter is considered {to}. Other parameters are considered an array of {from}.

Starting from leftmost {from} parameter, resolves {to} to an absolute path.

If {to} isn't already absolute, {from} arguments are prepended in right to left order, until an absolute path is found. If after using all {from} paths still no absolute path is found, the current working directory is used as well. The resulting path is normalized, and trailing slashes are removed unless the path gets resolved to the root directory.
```

erwinheitzman: Is `.ndjson` a typo?

packages/wdio-cucumber-framework/src/index.ts
```diff
@@ -481,6 +483,39 @@ class CucumberAdapter {
...
+
+    const cucumberMessage = (await Promise.all(files.map((file) => readFile(path.resolve(`${cucumberMessageDir}/${file}`), 'utf8')))).join('')
```
erwinheitzman, line 510: I think `normalize` is better here as well as changing the forward slash to (better for OS compatibility):
```
path.join(cucumberMessageDir, file)
```

erwinheitzman (commented):
Looks good overall but I have some questions and suggestions 👍 

Nice work!

packages/wdio-cucumber-framework/src/index.ts
tamil777selvan: Using the suggested "normalize" approach should work well. I'll implement this change now.

packages/wdio-cucumber-framework/src/index.ts
tamil777selvan: Roger, I'll implement this change now.

christian-bromann (approved):
Awesome stuff 👍 

@erwinheitzman please verify if your comments got addressed and merge

erwinheitzman (approved):
Awesome works 👏

PR #11368, opened 2023-10-09
WebdriverIO project generator for Serenity/JS

jan-molak:
## Proposed changes

Per my discussion with @christian-bromann, I'd like to extend the WebdriverIO project wizard to generate Serenity/JS projects, too.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments


christian-bromann (approved):
This PR seems to only contains improvements for unit tests and some wording. No signs of Serenity in here. If this is intentional, let's go ahead and merge this as it a good patch itself.

LGTM 👍

jan-molak:
@christian-bromann - I'm adding Serenity now :)

jan-molak:
@christian-bromann - I added project generators for Serenity/JS with Cucumber, Jasmine, and Mocha. Please let me know if you're happy with the overall structure and integration points.

I still need to think about additional tests I could write to ensure that EJS template rendering works as expected.

packages/wdio-cli/src/templates/wdio.conf.tpl.ejs
```diff
@@ -232,7 +251,13 @@ if (answers.isUsingTypeScript) {
     // specFileRetriesDeferred: false,
     //
     <%- include('snippets/reporters', { reporters: answers.reporters }) %>
-    <% if(answers.framework === 'mocha') { %>
+    <% if(answers.serenityAdapter === 'mocha') { %>//
```
christian-bromann, line 254: Can we move all the framework opts into their own template snippets?

christian-bromann (commented):
One comment, otherwise LGTM

What do you think @erwinheitzman @tamil777selvan ?

erwinheitzman (approved):
LGTM 👍

packages/wdio-cli/src/commands/config.ts
```diff
@@ -66,7 +67,8 @@ export const parseAnswers = async function (yes: boolean): Promise<ParsedAnswers
         presetPackage.package,
         ...reporterPackages.map(reporter => reporter.package),
         ...pluginPackages.map(plugin => plugin.package),
-        ...servicePackages.map(service => service.package)
+        ...servicePackages.map(service => service.package),
+        ...serenityPackages,
```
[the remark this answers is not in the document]
jan-molak: I think so because `servicePackages` are determined based on what the developer selects in the configuration wizard, `serenityPackages` are separate from that concept. Is that what you meant?

packages/wdio-cli/src/templates/wdio.conf.tpl.ejs
jan-molak: Yeah, good call; let me do this now.

jan-molak: Done in https://github.com/webdriverio/webdriverio/pull/11368/commits/68a437c2e7894a647e9ff03e277fc8ad046dcf0e#diff-cc58b213dd871b45469e751b91589bd4836c606b4dbb8af89fc9cc38782fec66

packages/wdio-cli/src/templates/wdio.conf.tpl.ejs
```diff
@@ -231,58 +235,14 @@ if (answers.isUsingTypeScript) {
...
-    },
-    <% } %>
+    <%- include('snippets/reporters', { reporters: answers.reporters })
+    %><%- answers.serenityAdapter   === 'mocha'    ? include('snippets/serenity-js/mocha', { answers }) : ''
+    %><%- answers.framework         === 'mocha'    ? include('snippets/mocha', { answers }) : ''
+    %><%- answers.serenityAdapter   === 'jasmine'  ? include('snippets/serenity-js/jasmine', { answers }) : ''
+    %><%- answers.framework         === 'jasmine'  ? include('snippets/jasmine', { answers }) : ''
+    %><%- answers.serenityAdapter   === 'cucumber' ? include('snippets/serenity-js/cucumber', { answers }) : ''
+    %><%- answers.framework         === 'cucumber' ? include('snippets/cucumber', { answers }) : ''
+    %>
```
christian-bromann, line 245: Why splitting between Serenity and standard framework opts? Are they differ by a lot?

jan-molak: They differ quite a bit for Cucumber since S/JS supports all the major versions of Cucumber, which means supporting legacy parameters, tag filters from before tag expressions, and so on.

For Mocha and Jasmine, most configuration options are the same, but there are some slight differences, like support for `strict` flag in S/JS Mocha, but not in WDIO, or support for `cleanStack` in WDIO but not in S/JS
- https://serenity-js.org/api/jasmine-adapter/interface/JasmineConfig/ vs https://github.com/webdriverio/webdriverio/blob/main/packages/wdio-jasmine-framework/src/types.ts
- https://serenity-js.org/api/mocha-adapter/interface/MochaConfig/ vs https://github.com/webdriverio/webdriverio/blob/main/packages/wdio-mocha-framework/src/types.ts

Thinking about it, we could try and standardise the adapters in the future 🤔

packages/wdio-cli/src/commands/config.ts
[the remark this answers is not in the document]
erwinheitzman: I see it now, I was on my phone and it was a bit hard to read and I read serenity twice but it's service and serenity.

tamil777selvan (commented):
I have a minor suggestion. Could we have a single template instead of both `js.ejs` and `ts.ejs` as like we currently have?

packages/wdio-cli/src/templates/wdio.conf.tpl.ejs
jan-molak: Sure, I can do that.

jan-molak:
> I have a minor suggestion. Could we have a single template instead of both `js.ejs` and `ts.ejs` as like we currently have?

That's a good idea; I'll need to introduce some helpers to avoid having to repeat `if` statements.

jan-molak:
@christian-bromann and @tamil777selvan - thanks for your feedback. I merged WebdriverIO and Serenity/JS configuration templates. I've also removed separate templates for CJS/ESM and instead generate TS, CJS and ESM from TypeScript templates.

Serenity/JS code sample templates are reasonably involved as they contain classes, functions, interfaces, parameters, return types, and so on. To avoid polluting WDIO EJS templates with a bazillion `if` statements, I introduced an `EjsHelpers` class that helps to generate different kinds of files from the same TS file.

The best example of using the helpers can be seen at https://github.com/webdriverio/webdriverio/pull/11368/files#diff-d2b3a22b9c707586c3011b43acc439b12e3c952a0bf5f70f58cf257558406908

You'll also find unit tests demonstrating the behaviour of `EjsHelpers` at https://github.com/webdriverio/webdriverio/pull/11368/files#diff-99bb565c721ebef00c7ccea7cab374547e3c4a25a78d03a0e321c888aaefec5c

I think I've arrived at quite an elegant solution that could be used to simplify other WDIO templates too if needed.

Looking forward to hearing your thoughts!

christian-bromann (approved):
LGTM 👍

Great work!

tamil777selvan (approved):
LGTM 👍

PR #11397, opened 2023-10-10
changed order for preventing some errors

udarrr:
## Proposed changes

I just would like to suggest to check existence of chromedriver version before downloading and prevent the error when version of browser don't match with driver version each time when we're starting test execution.

2023-10-10T18:14:01.637Z WARN webdriver: Couldn't download Chromedriver v117.0.5938.150: Download failed: server returned code 404. URL: https://edgedl.me.gvt1.com/edgedl/chrome/chrome-for-testing/117.0.5938.150/win64/chromedriver-win64.zip, trying to find known good version...
2023-10-10T18:14:02.270Z INFO webdriver: Download of Chromedriver v117.0.5938.149 was successful

### DIFF
--- a/packages/wdio-utils/src/driver/utils.ts
+++ b/packages/wdio-utils/src/driver/utils.ts
@@ -267,8 +267,8 @@ export async function setupChromedriver (cacheDir: string, driverVersion?: strin
             try{
                 await _install({ ...chromedriverInstallOpts, buildId })
                 log.info(`Download of Chromedriver v${buildId} was successful`)
-            } catch {
-                throw new Error(`Couldn't have founded known good version for Chromedriver v${buildId}`);
+            } catch (err: any) {
+                throw new Error(`Couldn't download Chromedriver v${buildId}: ${err.message}`);
             }
         }         
         executablePath = computeExecutablePath({
--- a/packages/wdio-utils/src/driver/utils.ts
+++ b/packages/wdio-utils/src/driver/utils.ts
@@ -4,7 +4,6 @@ import fsp from 'node:fs/promises'
 import path from 'node:path'
 import cp from 'node:child_process'
 
-import got from 'got'
 import decamelize from 'decamelize'
 import logger from '@wdio/logger'
 import {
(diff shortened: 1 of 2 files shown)


packages/wdio-utils/src/driver/utils.ts
```diff
@@ -201,12 +201,38 @@ export function getCacheDir (options: Pick<Options.WebDriver, 'cacheDir'>, caps:
...
+    let prefix
+    if (fullVersion) {
+        prefix = fullVersion.match(/^[+-]?([0-9]+)/)
+    }
+    return prefix && prefix.length > 0 ? prefix[0] : ''
+}
+
+export async function checkKnownBuild (build: string) {
+    try {
+        const knownGoodVersions: any = await got('https://googlechromelabs.github.io/chrome-for-testing/known-good-versions.json').json()
```
christian-bromann, line 214: We should use `@puppeteer/browsers` primitives here, the package has interfaces that validates whether a given version is available or not.

erwinheitzman:
@christian-bromann you know the driver a bit better than me, do you know if this also needs to be applied to the other drivers?

christian-bromann:
> do you know if this also needs to be applied to the other drivers?

No, only Chrome has the restriction where we have to have the same driver version as the Chrome version.

packages/wdio-utils/src/driver/utils.ts
udarrr: yes that's it, i've added 'canDownload' instead of the part of code

packages/wdio-utils/src/driver/utils.ts
```diff
@@ -201,12 +201,30 @@ export function getCacheDir (options: Pick<Options.WebDriver, 'cacheDir'>, caps:
...
+    let prefix
+    if (fullVersion) {
+        prefix = fullVersion.match(/^[+-]?([0-9]+)/)
+    }
+    return prefix && prefix.length > 0 ? prefix[0] : ''
+}
+
+export async function getKnownBuild (build: string) {
+    log.warn(`Chromedriver v${build} don't exist, trying to find known good version...`)
+    const knownGoodVersions: any = await got('https://googlechromelabs.github.io/chrome-for-testing/known-good-versions.json').json()
```
christian-bromann, line 214: This still uses a download url instead of the Puppeteer package.

udarrr: Oh yes! The func needn't at all 'resolveBuildId' can parse major version and return good known. I've added changes

PR #11419, opened 2023-10-13
Expose `findStrategy`

### LINKED ISSUE #11216: [🐛 Bug]: `findStrategy` cannot be imported in v8

### WebdriverIO Version

8.16.11

### Node.js Version

v18.17.1

### Mode

Standalone Mode

```typescript
N/A
```

### What happened?

I am migrating from v7 to v8. I first consulted the blog post here:
https://webdriver.io/blog/2022/12/01/webdriverio-v8-released

then I set in. most of my early issues are covered by the blog post. but I'm running into (what I believe to be) an effect of the cjs -> esm change: certain subfolders under the `webdriverio` package are now truly "private" and will not resolve for me.

```
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './build/utils/findStrategy' is not defined by "exports" in /home/michaele/git/caliper/server/node_modules/webdriverio/package.json
    at new NodeError (node:internal/errors:405:5)
    at exportsNotFound (node:internal/modules/esm/resolve:359:10)
    at packageExportsResolve (node:internal/modules/esm/resolve:695:9)
    at resolveExports (node:internal/modules/cjs/loader:567:36)
    at Module._findPath (node:internal/modules/cjs/loader:636:31)
    at Module._resolveFilename (node:internal/modules/cjs/loader:1063:27)
    at Module._load (node:internal/modules/cjs/loader:922:27)
    at Module.require (node:internal/modules/cjs/loader:1143:19)
    at require (node:internal/modules/cjs/helpers:121:18)
    at Object.<anonymous> (/home/michaele/git/caliper/server/node_modules/@blend/wdio-shared/dist/src/methods/browser.js:9:24) {
  code: 'ERR_PACKAGE_PATH_NOT_EXPORTED'
}
```

the package I own (`wdio-shared`, listed in the stacktrace above) imports and builds against `findStrategy`:
```ts
// ...
import { findStrategy } from 'webdriverio/build/utils/findStrategy';
// ...
```
the `webdriverio` exports do not allow the import, which is breaking for me:
```json
  "exports": {
    ".": [
      {
        "types": "./build/index.d.ts",
        "import": "./build/index.js",
        "require": "./build/cjs/index.js"
      },
      "./build/cjs/index.js"
    ],
    "./package.json": "./package.json"
  },
```

I haven't found a way to effectively use this function anymore.
```
michaele@remote-dev-00:~/git/caliper/server/node_modules/webdriverio$ grep 'findStrategy' -r .
./build/utils/findStrategy.js:export const findStrategy = function (selector, isW3C, isMobile) {
./build/utils/findStrategy.d.ts.map:{"version":3,"file":"findStrategy.d.ts","sourceRoot":"","sources":["../../src/utils/findStrategy.ts"],"names":[],"mappings":"AA4BA,KAAK,gBAAgB,GAAG,MAAM,GAAG;IAAE,IAAI,EAAE,MAAM,CAAC;IAAC,IAAI,EAAE,MAAM,CAAA;CAAE,CAAA;AA4F/D,eAAO,MAAM,YAAY,aAAuB,gBAAgB,UAAU,OAAO,aAAa,OAAO;;;CAmLpG,CAAA"}
./build/utils/findStrategy.d.ts:export declare const findStrategy: (selector: SelectorStrategy, isW3C?: boolean, isMobile?: boolean) => {
./build/utils/findStrategy.d.ts://# sourceMappingURL=findStrategy.d.ts.map
./build/utils/index.js:import { findStrategy } from './findStrategy.js';
./build/utils/index.js:        const { using, value } = findStrategy(selector, this.isW3C, this.isMobile);
./build/utils/index.js:        const { using, value } = findStrategy(selector, this.isW3C, this.isMobile);
./build/commands/element/shadow$$.js:import { findStrategy } from '../../utils/findStrategy.js';
./build/commands/element/shadow$$.js:        const { using, value } = findStrategy(selector, this.isW3C, this.isMobile);
./build/commands/element/shadow$.js:import { findStrategy } from '../../utils/findStrategy.js';
./build/commands/element/shadow$.js:        const { using, value } = findStrategy(selector, this.isW3C, this.isMobile);
```

I checked `@wdio/utils` as well, which does not seem to be a public analog to `webdriverio/utils`, and neither exposes nor includes the `findStrategy` implementation.
https://github.com/webdriverio/webdriverio/tree/main/packages/wdio-utils

I would like to know what my available options are, as a replacement. or to request that this method be exposed in some fashion, if there is no readily available replacement.

_No response_

### How to reproduce the bug.

```bash
nvm i 18.17.1
node -e 'require("webdriverio/build/utils/findStrategy")'
```

### Relevant log output

```typescript
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './build/utils/findStrategy' is not defined by "exports" in /home/michaele/git/caliper/server/node_modules/webdriverio/package.json
    at new NodeError (node:internal/errors:405:5)
    at exportsNotFound (node:internal/modules/esm/resolve:359:10)
    at packageExportsResolve (node:internal/modules/esm/resolve:695:9)
    at resolveExports (node:internal/modules/cjs/loader:567:36)
    at Module._findPath (node:internal/modules/cjs/loader:636:31)
    at Module._resolveFilename (node:internal/modules/cjs/loader:1063:27)
    at Module._load (node:internal/modules/cjs/loader:922:27)
    at Module.require (node:internal/modules/cjs/loader:1143:19)
    at require (node:internal/modules/cjs/helpers:121:18)
    at Object.<anonymous> (/home/michaele/git/caliper/server/node_modules/@blend/wdio-shared/dist/src/methods/browser.js:9:24) {
  code: 'ERR_PACKAGE_PATH_NOT_EXPORTED'
}
```
```


### Code of Conduct

### PR DESCRIPTION
**Pull request by michaele-blend**
## Proposed changes

fixes #11216

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/webdriverio/package.json
+++ b/packages/webdriverio/package.json
@@ -56,6 +56,13 @@
       },
       "./build/cjs/index.js"
     ],
+    "./public": [
+      {
+        "types": "./build/public.d.ts",
+        "import": "./build/public.js"
+      },
+      "./build/public.js"
+    ],
     "./package.json": "./package.json"
   },
   "types": "./build/index.d.ts",
--- a/packages/webdriverio/package.json
+++ b/packages/webdriverio/package.json
@@ -56,13 +56,6 @@
       },
       "./build/cjs/index.js"
     ],
-    "./public": [
-      {
-        "types": "./build/public.d.ts",
-        "import": "./build/public.js"
-      },
-      "./build/public.js"
-    ],
     "./package.json": "./package.json"
   },
   "types": "./build/index.d.ts",
--- a/packages/webdriverio/src/index.ts
+++ b/packages/webdriverio/src/index.ts
@@ -17,6 +17,7 @@ import type * as elementCommands from './commands/element.js'
 
 export * from './types.js'
 export * from './utils/interception/types.js'
+export { findStrategy } from './utils/findStrategy.js'
 
 export const Key = KeyConstant
 export const SevereServiceError = SevereServiceErrorImport
@@ -34,7 +35,7 @@ export const SevereServiceError = SevereServiceErrorImport
  * @return browser object with sessionId
  * @see <a href="https://webdriver.io/docs/typescript">Typescript setup</a>
  */
-export const remote = async function(
+export const remote = async function (
     params: RemoteOptions,
     remoteModifier?: (client: WebDriverTypes.Client, options: Options.WebdriverIO) => WebDriverTypes.Client
 ): Promise<WebdriverIO.Browser> {
--- a/packages/webdriverio/src/index.ts
+++ b/packages/webdriverio/src/index.ts
@@ -35,7 +35,7 @@ export const SevereServiceError = SevereServiceErrorImport
  * @return browser object with sessionId
  * @see <a href="https://webdriver.io/docs/typescript">Typescript setup</a>
  */
-export const remote = async function (
+export const remote = async function(
     params: RemoteOptions,
     remoteModifier?: (client: WebDriverTypes.Client, options: Options.WebdriverIO) => WebDriverTypes.Client
 ): Promise<WebdriverIO.Browser> {
--- a/packages/webdriverio/src/public.ts
+++ b/packages/webdriverio/src/public.ts
@@ -1,0 +1,1 @@
+export { findStrategy } from './utils/findStrategy.js'
--- a/packages/webdriverio/src/public.ts
+++ b/packages/webdriverio/src/public.ts
@@ -1,1 +1,0 @@
-export { findStrategy } from './utils/findStrategy.js'

### CODE REVIEW

**Comment by erwinheitzman** [comment]
Thank you for the PR, I don't think there's a need for a public interface, simply add it to the current types

**Comment by christian-bromann** [comment]
`findStrategy` is an internal utility, can you elaborate on why you need this in your tests?

**Comment by michaele-blend** [comment]
certainly,
my company has an internal wrapper library for webdriverio, where we primarily extend the browser and elements interfaces.
as detailed in #11216, I am in the process of upgrading from v7 to v8.
some of our functions accept selector strings of unknown type (css or xpath), and perform browser-side selector dispatch (`document.evaluate` / `document.querySelectorAll`).
rather than implement our own selector string disambiguation, we rely on `findStrategy`.

a viable alternative to exposing `findStrategy` for us would be to implement our own, or to vendor a copy of your implementation.
relying on your `findStrategy` is preferable for us, because it is thorough and receives [regular updates](https://github.com/webdriverio/webdriverio/commits/main/packages/webdriverio/src/utils/findStrategy.ts).

**Comment by erwinheitzman** [comment]
We already allow users do defined custom selector strategies and also allow users to find elements using a function of which I feel both would achieve the goal that you want to achieve. Could you confirm if this is indeed correct or not?

**Comment by erwinheitzman** [comment]
@michaele-blend What do you think?

**Comment by michaele-blend** [comment]
@erwinheitzman apologies for the delayed response,

> We already allow users to define custom selector strategies and also allow users to find elements using a function of which I feel both would achieve the goal that you want to achieve. Could you confirm if this is indeed correct or not?

perhaps I'm misunderstanding something, but I don't believe this will work for me.

my functions do not look like this:
```ts
const strategy = wdio.findStrategy(selector);
const el = await browser.executeAsync((strategy, selector, done) => { /* lookup */ }, strategy, selector);
await el.doStuff();
```
my functions instead look more like this:
```ts
const strategy = wdio.findStrategy(selector);
await browser.executeAsync((strategy, selector, done) => {
  const get = () => {
    switch (strategy) {
      case 'xpath': return /* document.evaluate */;
      case 'css selector': return /* document.querySelectorAll */;
      default: /* throw */
    }
  }
  const el = get();
  // do stuff
}, strategy, selector);
```
we have two forms, one is a synchronous script which asserts selected elements are not zero-size.
another is an asynchronous animation wait helper, which ensures bounding box and opacity stability across multiple animation frames, before resolving. (run this as a script rather than issuing multiple webdriver commands to control for latency differences on developer and CI machines)

is there some approach using custom lookup strategies that I'm overlooking?
if so, I may not need to call into `findStrategy` at all.

erwinheitzman:
@michaele-blend the examples really give more context here.

If we are talking about only two forms, I really wonder why you'd go so far instead of creating a page object that reflects the two forms (one object for each form).

Perhaps I am still misunderstanding but you want to use xpath and document.evaluate for one and css and querySelectorAll (equal to $$) for the other so I feel like two simple objects already solve this problem but let me know if I misunderstood your challenge.
```
export class FormA {
  getSomeElement() {
    return $$('.foo'); // querySelectorAll equivalent
  }
}

export class FormB {
  getSomeOtherElement() {
    return $(() => document.evaluate('//button[@type="submit'']'));
  }
}
```

erwinheitzman:
@michaele-blend ping

christian-bromann:
Closing due to inactivity.

michaele-blend:
@erwinheitzman I do not mean literal HTML `form` tags. I meant "we have two [implementation] forms". put another way, we have two use cases for `findStrategy`. our two use cases are generic to what is being selected for.
I should have chosen my words better, apologies for the confusion.

I've consulted the docs once more, I see
https://webdriver.io/docs/api/browser/execute
> Arguments may be any JSON-primitive, array, or JSON object. JSON objects that define a **WebElement reference** will be converted to the corresponding DOM element. Likewise, any WebElements in the script result will be returned to the client as WebElement JSON objects.

is your recommendation that we instead pass element references (return of `$`) as arguments to the exec?
this would probably work for the non-zero size assertion we have. but I worry that the element references might grow stale over the course of the function call for the "bounding box and opacity stability across multiple animation frames" use case, previously mentioned. I would prefer to use a selector and perform the lookup repeatedly, immediately before using the element. but thinking on this a bit more, we can probably just check the [`isConnected`](https://developer.mozilla.org/en-US/docs/Web/API/Node/isConnected) property to know if the element reference has gone stale, and reject in this case.

I'll try this out and report back.

Issue #11243, opened 2023-09-25
[💡 Feature]: allow sharding tests for better scalability in large CI systems


muhqu:
### Is your feature request related to a problem?

We have over 100 tests and like to run them across multiple test executors in our CI environment. A single machine can only run a few browsers in parallel due to memory consumption. Being able to run the tests across multiple machines allows for better scaling.

### Describe the solution you'd like.

The [playwright](https://playwright.dev/docs/test-parallel#shard-tests-between-multiple-machines) and [vitest](https://vitest.dev/guide/cli.html#shard) test runners support a `--shard` parameter, that evenly splits the number of tests to run.
e.g. if you have 100 tests `--shard=1/4` would execute test 1-25, `--shard=2/4` would execute test 26-50, etc…

### Describe alternatives you've considered.

To work around the issue I can manually [group tests in test suites](https://webdriver.io/docs/organizingsuites/#grouping-test-specs-in-suites) and setup specific jobs to trigger each test suite to have them run across multiple machines... 

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

erwinheitzman:
Hi @muhqu thank you for sharing this idea.

I think it's quite interesting and am interesting in how you currently run your tests on multiple devices. Do you use a selenium grid for example?

muhqu:
We are currently using playwright, but looking to evaluate alternatives like WebdriverIO as browser interaction layer. Mostly because playwright is too opinionated and does not work well if you require persistent browser context (non-incognito), or if you need to test with Safari preview or Firefox nightly releases…

We’re not using Selenium Grid, but we have used BrowserStack and SauceLabs in the past. However, as the added latency between test code and remote browser had a huge impact on per-test performance, we are now running playwright test runners on multiple machines in parallel using their sharding mechanism. Each test runner also runs multiple workers, but we’re a bit limited by the memory footprint of our application (+4GB per browser).

That’s why I’m looking for a similar feature in WebdriverIO.

christian-bromann:
@muhqu sounds like a valuable feature to have. Let me have a look into this. I am happy to make that contribution if it helps you run your tests more effectively.

nextlevelbeard:
+1

christian-bromann:
This has landed and was released. Closing.

PR #11441, opened 2023-10-16
(@wdio/cli): Support for sharding

christian-bromann:
## Proposed changes

implements #11243

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

I will extend the docs once the feature is released with an example how to use sharding with Github Action.

### DIFF
--- a/packages/wdio-cli/src/launcher.ts
+++ b/packages/wdio-cli/src/launcher.ts
@@ -3,13 +3,14 @@ import path from 'node:path'
 import exitHook from 'async-exit-hook'
 
 import logger from '@wdio/logger'
-import { ConfigParser } from '@wdio/config'
+import { ConfigParser, validateConfig } from '@wdio/config'
 import { initialisePlugin, initialiseLauncherService, sleep, setupDriver, setupBrowser } from '@wdio/utils'
 import type { Options, Capabilities, Services } from '@wdio/types'
 
 import CLInterface from './interface.js'
-import type { HookError } from './utils.js'
 import { runLauncherHook, runOnCompleteHook, runServiceHook } from './utils.js'
+import { TESTRUNNER_DEFAULTS } from './constants.js'
+import type { HookError } from './utils.js'
 import type { RunCommandArguments } from './types.js'
 
 const log = logger('@wdio/cli:launcher')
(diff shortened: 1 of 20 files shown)


packages/wdio-cli/src/launcher.ts
```diff
@@ -244,6 +246,12 @@ class Launcher {
              * fail if no specs were found or specified
              */
             if (Object.values(this._schedule).reduce((specCnt, schedule) => specCnt + schedule.specs.length, 0) === 0) {
+                const { total, current } = config.shard
+                if (total > 1) {
+                    log.info(`No specs to execute in shard ${current}/${total}, exiting!`)
+                    return resolve(0)
```
erwinheitzman, line 252: If we exit with code 1 when there are no tests, I would expect the same here

erwinheitzman (commented):
Omg I really love this, clean code, well done on this Christian. Just one comment on the exit code 👍

packages/wdio-cli/src/launcher.ts
christian-bromann: For sharding, this would be not desireable. For example: let's say a user has a CI setup that by defaults uses 10 workers and calls wdio always with `--shard 1/10` ... `--shard 10/10`. Now let's say we have 100 test files and provide a dynamic parameter like `--spec="SOME_PATTERN"` which can resolves in 9 test files. In this case just because the dynamic spec parameter causes let's files to be selected than for what sharding is set up for should not cause the worker to break.

erwinheitzman (approved):
There's one merge conflict that needs to be resolved but other than that this LGTM 👍

kamal-kaur04:
@christian-bromann, was debugging an issue inside `browserstack-wdio-service` package wherein we used to send test status extending WDIO Reporter. We had some checks as per framework value inside the [config](https://github.com/webdriverio/webdriverio/blob/712767b49161e483a51eb12cf44074acd29e018d/packages/wdio-browserstack-service/src/reporter.ts#L42). Now, it seems that the value isn't in the config object anymore. Seems like we have removed framework  and other few parameters from `WDIO_DEFAULTS` in the [commit](https://github.com/webdriverio/webdriverio/pull/11441/commits/9b805887b984c7af5fb2dd20a4d6a8e66db90f1d) of this PR that is creating some test reporting failures. So, anyway to get this info inside `onRunnerStart` hook back?

christian-bromann:
> Now, it seems that the value isn't in the config object anymore.

Which value?

> So, anyway to get this info inside `onRunnerStart` hook back?

What information?

kamal-kaur04:
> We had some checks as per framework value inside the [config](https://github.com/webdriverio/webdriverio/blob/712767b49161e483a51eb12cf44074acd29e018d/packages/wdio-browserstack-service/src/reporter.ts#L42)

^ [this.config.framework](https://github.com/webdriverio/webdriverio/blob/712767b49161e483a51eb12cf44074acd29e018d/packages/wdio-browserstack-service/src/reporter.ts#L128)

christian-bromann:
These information should be still accessible. Can you please point out which service or reporter hook used to have this information?

christian-bromann:
Please create a separate issue for this. Any contributions are welcome!

kamal-kaur04:
@christian-bromann Sure, what was the reason of removing these keys from WDIO_DEFAULTS?

christian-bromann:
These are testrunner specific defaults and should not be a concern for the `webdriverio` package. They were [moved](https://github.com/webdriverio/webdriverio/pull/11441/commits/9b805887b984c7af5fb2dd20a4d6a8e66db90f1d#diff-77007d1fd1f17a3a42b172c92c6000f6ac64d45f7091c24822551c25d0a9d45fR675) to the CLI package.

PR #11469, opened 2023-10-18
(@wdio/utils): separate between common and node utils

christian-bromann:
## Proposed changes

Our browser runner currently has to resolve a lot of Node.js primitives through Vite which can fail at all times due to transient dependencies. This PR attempts to separate Node.js related imports and ones that can run on all platforms.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments


erwinheitzman:
Is it not possible to keep the types instead of setting it to any?

christian-bromann:
> Is it not possible to keep the types instead of setting it to any?

It is but it would require a more sophisticated build setup which doesn't justify the value of keeping the types, because we don't ship them to users. They were there to keep us in check but ultimately we don't change a lot in these CJS files which is why we can also live without them.

PR #11470, opened 2023-10-19
fix(types): Element and WebdriverIO.Element definitions do not match

Badisi:
## Proposed changes

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

Was migrating my project from wdio `8.14` to `8.19` and got this blocking issue:
```ts
The 'this' context of type 'import("/node_modules/webdriverio/build/types").Element' is not assignable to method's 'this' of type 'WebdriverIO.Element'.
  Type 'Element' is missing the following properties from type 'Element': sessionStatus, sessionNew, sessionEnd, sessionSubscribe, and 25 more.ts(2684)
```

The problem is that type definition between [Element](https://github.com/webdriverio/webdriverio/blob/main/packages/webdriverio/src/types.ts#L283) and [WebdriverIO.Element](https://github.com/webdriverio/webdriverio/blob/main/packages/webdriverio/src/types.ts#L493) are no more in sync.

### DIFF
--- a/packages/webdriverio/src/types.ts
+++ b/packages/webdriverio/src/types.ts
@@ -280,7 +280,7 @@ export interface ElementBase extends InstanceBase, ElementReference, CustomInsta
      */
     error?: Error
 }
-export interface Element extends ElementBase, ProtocolCommands, Omit<BrowserCommandsType, keyof ElementCommandsType>, ElementCommandsType {}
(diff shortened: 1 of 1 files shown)


christian-bromann (approved):
Nice catch. I think initially I thought that bidi commands should be only called from the `browser` object but I did a quick check and:

```ts
const elem = await browser.$('body')
console.log(typeof elem.scriptAddPreloadScriptCommand)
```

actually prints out `function` for this one. So make sense to merge this even though it is currently not our desired behavior.

Thanks for the contribution!

PR #11529, opened 2023-10-25
changed moveTo to like it's been implemented in click

udarrr:
## Proposed changes

change moveTo to click similarity, because of flaky moveTo

## Types of changes

I'd like to suggest to change it to a way like it's been implemented in click method with 'origin: this' instead of calculation full offset. Even more i don't understand why we should calculate it instead of using 'origin: this' as click method perform particularly the same, with stable behavior in each cases, just with some additional actions like 'down' and 'up'.

### DIFF
--- a/e2e/wdio/headless/test.e2e.ts
+++ b/e2e/wdio/headless/test.e2e.ts
@@ -4,7 +4,7 @@ describe('main suite 1', () => {
     it('foobar test', async () => {
         const browserName = (browser.capabilities as WebdriverIO.Capabilities).browserName
         await browser.url('http://guinea-pig.webdriver.io/')
-        await expect((await $('#useragent').getText()).toLowerCase()).toContain(browserName)
+        await expect((await $('#useragent').getText()).toLowerCase()).toContain(browserName ? browserName.replace(' ', '') : browserName)
     })
 
     it('should allow to check for PWA', async () => {
--- a/e2e/wdio/headless/test.e2e.ts
+++ b/e2e/wdio/headless/test.e2e.ts
@@ -99,6 +99,13 @@ describe('main suite 1', () => {
     })
 
     describe('moveTo tests', () => {
+        const inputs = [
+            undefined,
+            { xOffset: 10 },
+            { yOffset: 10 },
+            { xOffset: 25, yOffset: 25 },
+        ]
+
         before(async () => {
             await browser.url('http://guinea-pig.webdriver.io/')
             await (await browser.$('a[href="pointer.html"]')).click()
@@ -117,18 +124,26 @@ describe('main suite 1', () => {
             expect(value.endsWith('center\n')).toBe(true)
         })
 
-        it('moveTo without iframe with xOffset and yOffset regarding center', async () => {
-            const elmRectChild = await browser.getElementRect(await browser.$('#child').elementId)
-            const elmRectParent = await browser.getElementRect(await browser.$('#parent').elementId)
-            await (await browser.$('#parent')).moveTo({ xOffset: elmRectParent.width/2 - elmRectChild.width/2 + 1, yOffset: elmRectParent.height - elmRectChild.height + 1 })
-            const value = await (await browser.$('#text')).getValue()
-            expect(value.endsWith('out\n')).toBe(true)
+        inputs.forEach((input) => {
+            it(`moves to position x,y outside of iframe when passing the arguments ${JSON.stringify(input)}`, async () => {
+                await browser.execute(() => {
+                    const mouse = { x:0, y:0 }
+                    document.onmousemove = function(e){ mouse.x = e.clientX, mouse.y = e.clientY }
+                    //@ts-ignore
+                    document.mouseMoveTo = mouse
+                })
+                await (await browser.$('#parent')).moveTo()
+                const rectBefore = await browser.execute('return document.mouseMoveTo') as {x: number, y: number}
+                await (await browser.$('#parent')).moveTo(input)
+                const rectAfter = await browser.execute('return document.mouseMoveTo') as {x: number, y: number}
+                expect(rectBefore.x + (input && input?.xOffset ? input?.xOffset : 0)).toEqual(rectAfter.x)
+                expect(rectBefore.y + (input && input?.yOffset ? input?.yOffset : 0)).toEqual(rectAfter.y)
+            })
         })
 
         it('moveTo in iframe', async () => {
             const iframe = await browser.$('iframe.code-tabs__result')
             await browser.switchToFrame(iframe)
-            await (await browser.$('#parent')).scrollIntoView()
             await (await browser.$('#parent')).moveTo()
             const value = await (await browser.$('#text')).getValue()
             expect(value.endsWith('center\n')).toBe(true)
@@ -140,12 +155,21 @@ describe('main suite 1', () => {
             expect(value.endsWith('center\n')).toBe(true)
         })
 
-        it('moveTo in iframe with xOffset and yOffset regarding center', async () => {
-            const elmRectChild = await browser.getElementRect(await browser.$('#child').elementId)
-            const elmRectParent = await browser.getElementRect(await browser.$('#parent').elementId)
-            await (await browser.$('#parent')).moveTo({ xOffset: elmRectParent.width/2 - elmRectChild.width/2 + 1, yOffset: elmRectParent.height - elmRectChild.height + 1 })
-            const value = await (await browser.$('#text')).getValue()
-            expect(value.endsWith('out\n')).toBe(true)
+        inputs.forEach((input) => {
+            it(`moves to position x,y inside of iframe when passing the arguments ${JSON.stringify(input)}`, async () => {
+                await browser.execute(() => {
+                    const mouse = { x:0, y:0 }
+                    document.onmousemove = function(e){ mouse.x = e.clientX, mouse.y = e.clientY }
+                    //@ts-ignore
+                    document.mouseMoveTo = mouse
+                })
+                await (await browser.$('#parent')).moveTo()
+                const rectBefore = await browser.execute('return document.mouseMoveTo') as {x: number, y: number}
+                await (await browser.$('#parent')).moveTo(input)
+                const rectAfter = await browser.execute('return document.mouseMoveTo') as {x: number, y: number}
+                expect(rectBefore.x + (input && input?.xOffset ? input?.xOffset : 0)).toEqual(rectAfter.x)
+                expect(rectBefore.y + (input && input?.yOffset ? input?.yOffset : 0)).toEqual(rectAfter.y)
+            })
         })
     })
 
--- a/e2e/wdio/headless/test.e2e.ts
+++ b/e2e/wdio/headless/test.e2e.ts
@@ -118,8 +118,9 @@ describe('main suite 1', () => {
         })
 
         it('moveTo without iframe with xOffset and yOffset regarding center', async () => {
-            const elmRect = await browser.getElementRect(await browser.$('#parent').elementId)
-            await (await browser.$('#parent')).moveTo({ xOffset: elmRect.width -  1, yOffset: elmRect.height - 1 })
+            const elmRectChild = await browser.getElementRect(await browser.$('#child').elementId)
+            const elmRectParent = await browser.getElementRect(await browser.$('#parent').elementId)
+            await (await browser.$('#parent')).moveTo({ xOffset: elmRectParent.width/2 - elmRectChild.width/2 + 1, yOffset: elmRectParent.height - elmRectChild.height + 1 })
             const value = await (await browser.$('#text')).getValue()
             expect(value.endsWith('out\n')).toBe(true)
         })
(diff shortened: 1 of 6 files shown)


udarrr:
> Hi @udarrr
> 
> Could you provide examples of actual cases in which the command is not doing what you would expect it to do? This way we can verify that this indeed fixes the problem(s)

Hi @erwinheitzman  unfortunately i can't because i know reproduceable a example with no public resources, but from my perspective here something wrong. Let me explain

we have implementation clicking 

something like 

if (this.isW3C) {
        await browser.action('pointer', {
            parameters: { pointerType: 'mouse' }
        })
            .move({
                origin: this,
                x: xoffset,
                y: yoffset
            })
            .down({ button })
            .up({ button })
            .perform(skipRelease)
        return
    }

it's moving pointer to element with scrolling and then down and up pointer, if we'd like to be moved to element without pressing pointer i believe we should just use 'move' without down and up. It's the same without useless calculations. And it works everywhere even with my apps.

erwinheitzman:
> > Hi @udarrr
> 
> > 
> 
> > Could you provide examples of actual cases in which the command is not doing what you would expect it to do? This way we can verify that this indeed fixes the problem(s)
> 
> 
> 
> Hi @erwinheitzman  unfortunately i can't because known reproduceable an example with no public resources, but from my perspective here something wrong. Let me explain
> 
> 
> 
> we have implementation of clicking, click https://github.com/webdriverio/webdriverio/blob/main/packages/webdriverio/src/commands/element/click.ts already have an implementation of moveTo but with some extra actions 
> 
> 
> 
> something like 
> 
> 
> 
> if (this.isW3C) {
> 
>         const browser = getBrowserObject(this)
> 
>         await browser.action('pointer', {
> 
>             parameters: { pointerType: 'mouse' }
> 
>         })
> 
>             .move({
> 
>                 origin: this,
> 
>                 x: xoffset,
> 
>                 y: yoffset
> 
>             })
> 
>             **.down({ button })
> 
>             .up({ button })**
> 
>             .perform(skipRelease)
> 
>         return
> 
>     }
> 
> 
> 
> it's moving pointer to element with scrolling and then down and up pointer, if we'd like to be moved to element without pressing pointer i believe we should just use 'move' without down and up. It's the same without useless calculations. And it works everywhere even with my apps.
> 
> 

I think I understand what you mean and I feel like we should create a utility function to retrieve the coordinates of an element with the passed offsets as right now we would have to maintain it in two places. Also the click implementation is still different from the changes you suggest so that will also be resolved by doing this.
Do you have a clear picture of what it is I am suggesting? 🙂

erwinheitzman:
Additionally there might be more commands that could benefit from such a utility function, like the scrollIntoView command might be able to use it (it does scroll of course)

udarrr:
> Do you have a clear picture of what it is I am suggesting?

To be honest no :) it's not clear for me. Could you clarify that ?

this is click implementation 

await browser.action('pointer', {
parameters: { pointerType: 'mouse' }
})
.move({
origin: this,
x: xoffset,
y: yoffset
})
.down({ button })
.up({ button })
.perform(skipRelease)

this is my implementation of moveTo

await browser.action('pointer', {
parameters: { pointerType: 'mouse' }
})
.move({
origin: this,
x: xoffset,
y: yoffset
})
.perform()

udarrr:
> Additionally there might be more commands that could benefit from such a utility function, like the scrollIntoView command might be able to use it (it does scroll of course)

I am still confused, 'scrollIntoView ' using 'wheel' at the moment click and moveTo using 'pointer'

udarrr:
Even dnd use origin in the project instead of calculations

image

udarrr:
> Hi @udarrr
> 
> Could you provide examples of actual cases in which the command is not doing what you would expect it to do? This way we can verify that this indeed fixes the problem(s)

Generally i was able to reproduce it. Here is description https://github.com/webdriverio/webdriverio/issues/11534 There are two issue.

1. It can't be performed properly inside of iframe with current version of moveTo but can be resolved as i've done it above.
2. My assumption that 'move' method can scroll it to element was wrong that's why if element not into view 'moveTo' method throw an Error 'move target out of bounds' Maybe it have sense to rewrite it to the version

    const browser = getBrowserObject(this);
    await this.scrollIntoView(scrollOptions)
    await browser.pause(scrollAnimationTimeout)
    return browser.action('pointer', { parameters: { pointerType: 'mouse' } })
        .move({origin: this, x: 0, y: 0 })
        .perform();

but i am not convinced

erwinheitzman:
The command adheres to the webdriver specs by throwing the error and has always acted like this and there are other options to do this (action command).

What I meant is that the whole part of the coordinates can be abstracted away into a function that can be shared between moveTo and click

udarrr:
> What I meant is that the whole part of the coordinates can be abstracted away into a function that can be shared between moveTo and click

I got it, but i'd like to suggest don't use it at all. Currently click are using the xoffset,yoffset just like arguments without calculations, the same could be for moveTo because by default it already moves to center of element

xoffset - {number} X offset to move to, relative to the top-left corner of the element. If not specified, the mouse will move to the middle of the element.
yoffset - {number} Y offset to move to, relative to the top-left corner of the element. If not specified, the mouse will move to the middle of the element.

> The command adheres to the webdriver specs by throwing the error and has always acted like this and there are other options to do this (action command) so I would say that it is up to the user to scroll for now.

Ok i'll create for that different discussion not like a part of the pr. Maybe it could be a good feature to add it.

packages/webdriverio/src/commands/element/moveTo.ts
```diff
@@ -1,4 +1,4 @@
-import { getElementRect, getScrollPosition, getBrowserObject } from '../../utils/index.js'
```
christian-bromann, line 1: We should get rid of `getScrollPosition` as it is nowhere used no more.

christian-bromann (commented):
I think overall this makes sense. I wonder if there is an easy way to create some e2e tests for this similar to what we have [`/e2e/wdio/headless/test.e2e.ts#L123`](https://github.com/webdriverio/webdriverio/blob/abb33f0043cdf9415d7611e559b5fccbd79b7003/e2e/wdio/headless/test.e2e.ts#L123).

@erwinheitzman wdyt?

packages/webdriverio/src/commands/element/moveTo.ts
udarrr: Ok i'll create some test and get rid of getScrollPosition

udarrr: Could you look into it https://github.com/webdriverio/guinea-pig/pull/2. There is an extra page for moveTo checks
Is it correct project for resource http://guinea-pig.webdriver.io ?

christian-bromann: @udarrr I merged the PR and the guinea pig page was deployed

erwinheitzman:
> I think overall this makes sense. I wonder if there is an easy way to create some e2e tests for this similar to what we have [`/e2e/wdio/headless/test.e2e.ts#L123`](https://github.com/webdriverio/webdriverio/blob/abb33f0043cdf9415d7611e559b5fccbd79b7003/e2e/wdio/headless/test.e2e.ts#L123).
> 
> @erwinheitzman wdyt?

Yeah that's totally possible.

We can moveTo the same element with different offsets and also things like not passing and offsets, something along the lines of:
```js
const inputs = [
  [undefined, { x: 9999, y: 9999 }],
  [{ xOffset: 10 }, { x: 9999, y: 9999 }],
  [{ yOffset: 10 }, { x: 9999, y: 9999 }],
  [{ xOffset: 25, yOffset: 25 }, { x: 9999, y: 9999 }],
  [{ xOffset: Number.MAX_SAFE_INTEGER, yOffset: Number.MAX_SAFE_INTEGER }, { x: 9999, y: 9999 }],
]

inputs.forEach(([input, output]) => {
  it(`moves to position x ${output.x} y ${output.y} when passing the arguments ${JSON.stringify(input)}`, () => {
    await browser.execute(() => {
      let mouse = { x:0, y:0 }
      document.onmousemove = function(e){ mouse.x = e.clientX, mouse.y = e.clientY }
    })
    
    await elem.moveTo(input);
    
    const { x, y } = await browser.execute(() => mouse)

    expect(x).toEqual(output.x)
    expect(y).toEqual(output.y)
  })
})
```

packages/webdriverio/src/commands/element/moveTo.ts
udarrr: @christian-bromann  I was able to add some test, please check it.

udarrr:
> Yeah that's totally possible.
> 
> We can moveTo the same element with different offsets and also things like not passing and offsets, something along the lines of:
> 
> ```js
> const inputs = [
>   [undefined, { x: 9999, y: 9999 }],
>   [{ xOffset: 10 }, { x: 9999, y: 9999 }],
>   [{ yOffset: 10 }, { x: 9999, y: 9999 }],
>   [{ xOffset: 25, yOffset: 25 }, { x: 9999, y: 9999 }],
>   [{ xOffset: Number.MAX_SAFE_INTEGER, yOffset: Number.MAX_SAFE_INTEGER }, { x: 9999, y: 9999 }],
> ]
> 
> inputs.forEach(([input, output]) => {
>   it(`moves to position x ${output.x} y ${output.y} when passing the arguments ${JSON.stringify(input)}`, () => {
>     await browser.execute(() => {
>       let mouse = { x:0, y:0 }
>       document.onmousemove = function(e){ mouse.x = e.clientX, mouse.y = e.clientY }
>     })
>     
>     await elem.moveTo(input);
>     
>     const { x, y } = await browser.execute(() => mouse)
> 
>     expect(x).toEqual(output.x)
>     expect(y).toEqual(output.y)
>   })
> })
> ```
> 
> Note that this code is an example and I can by no means guarantee it works

That's great an example, where are you been early... generally i've added test cases already not so refined but still)

udarrr:
Tests are failed don't related to last changes, because of 

[headless chrome 120.0.6092.0 linux #1-0] Expected substring: "**headless chrome**"
[headless chrome 120.0.6092.0 linux #1-0] Received string:    "mozilla/5.0 (x11; linux x86_64) applewebkit/537.36 (khtml, like gecko) **headlesschrome**/120.0.6092.0 safari/537.36"

Any thoughts ?

udarrr:
I've fixed 2 test cases regarding message above and finally it can be merged)

e2e/wdio/headless/test.e2e.ts
```diff
@@ -98,6 +98,81 @@ describe('main suite 1', () => {
...
+        ]
+
+        before(async () => {
+            await browser.url('http://guinea-pig.webdriver.io/')
+            await (await browser.$('a[href="pointer.html"]')).click()
+            await (await browser.$('#parent')).waitForExist()
+        })
+
+        it('moveTo without iframe', async () => {
+            await (await browser.$('#parent')).moveTo()
```
christian-bromann, line 116:
```suggestion
            await browser.$('#parent').moveTo()
```

e2e/wdio/headless/test.e2e.ts
```diff
@@ -98,6 +98,81 @@ describe('main suite 1', () => {
...
+            const value = await (await browser.$('#text')).getValue()
```
christian-bromann, line 117:
```suggestion
            const value = await browser.$('#text').getValue()
```

e2e/wdio/headless/test.e2e.ts
```diff
@@ -98,6 +98,81 @@ describe('main suite 1', () => {
...
+            expect(value.endsWith('center\n')).toBe(true)
+        })
+
+        it('moveTo without iframe with 0 offsets', async () => {
+            await (await browser.$('#parent')).moveTo({ xOffset: 0, yOffset: 0 })
```
christian-bromann, line 122: same with the other occurences

udarrr: Done

erwinheitzman (approved):
This is looking really good, thank you very much for your hard work 👏

vitaly-kiselev-qa:
@udarrr 
Hello.

The comment states that `xOffset` and `yOffset` are relative to the **top-left corner** of the element:
https://github.com/webdriverio/webdriverio/blob/main/packages/webdriverio/src/commands/element/moveTo.ts#L16

Looks like it is no longer true since they are passed unchanged to `moveTo`:
https://github.com/webdriverio/webdriverio/blob/main/packages/webdriverio/src/commands/element/moveTo.ts#L44

And `move (params: PointerActionMoveParams)` expects `x` and `y` relative to the **center** of a specific element:
https://webdriver.io/docs/api/browser/action/#pointer-input-source

christian-bromann:
@vitaly-kiselev-qa feel free to raise a PR if you have a concrete suggestions in mind.

vitaly-kiselev-qa:
@christian-bromann thank you for your answer.

My comment was more a question not a suggestion.

It looks like, the behaviour has been changed and the documention has not. 
The question is — was the change intentional or not?

For my code this was a breaking change and I had to rewrite it (calculate coordinates in a new way).

christian-bromann:
@vitaly-kiselev-qa mind raising an issue with a description of this, I think this should be fixed.

Issue #11531, opened 2023-10-25
[🐛 Bug]: Allure Report missing testCaseId (TestOps cannot assign id to a test result)


AlexKorTutu:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

8.20.0

### Node.js Version

v18.13.0

### Mode

WDIO Testrunner

### What happened?

I updated from a pretty old setup
"@wdio/allure-reporter": "^7.16.10",
"webdriverio": "^7.20.5"
which used aoutdated xml format for allure results

up to 
"webdriverio": "^8.20.0",
"@wdio/allure-reporter": "^8.20.0",
which produces json results

with the update test ops lost ability to assign test results to test-cases. 
I found out that allure testOps unlike allure report uses different unique ID to assign cases
testCaseId - test ops,  historyId - report
https://allurereport.org/docs/how-it-works/test-result-file/#historyid-string - both can be found in docs here

adding testCaseId to the result.json file manually before upload to testOps solves the problem 

### What is your expected behavior?

allure report should have both historyId and testCaseId which can be generated at the same time, since they use pretty much the same structure.

testCaseId = md5(fullName, sort(names(parameters)) // see https://help.qameta.io/support/solutions/articles/101000480600-cooking-the-allureid

which is exactly how historyId is generated now 
https://github.com/webdriverio/webdriverio/blob/main/packages/wdio-allure-reporter/src/utils.ts#L189

### How to reproduce the bug.

upload these files to testOps
[noTestCaseId-result.json](https://github.com/webdriverio/webdriverio/files/13164160/noTestCaseId-result.json)
[withTestCaseId-result.json](https://github.com/webdriverio/webdriverio/files/13164161/withTestCaseId-result.json)
second file has historyId added as testCase Id 
"testCaseId":"f6edc25502e8c60dcadb37f34b9efc73", "historyId":"f6edc25502e8c60dcadb37f34b9efc73"

![image](https://github.com/webdriverio/webdriverio/assets/40691993/057e34e0-955b-48a6-84e1-965bc7e13c09)

### Relevant log output

```typescript
nothing to share here
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

BorisOsipov:
[...]

PR #11532, opened 2023-10-25
add testCaseId for allure testOps

AlexKorTutu:
## Proposed changes

[//]: # Support both allure report and allureTestOps by adding TestCaseId

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/wdio-allure-reporter/src/utils.ts
+++ b/packages/wdio-allure-reporter/src/utils.ts
@@ -177,7 +177,7 @@ export const getSuiteLabels = ({ tags }: SuiteStats): Label[] => {
     }, [])
 }
 
-export const setHistoryId = (test: AllureTest | undefined, suite: AllureGroup | undefined) => {
+export const setAllureIds = (test: AllureTest | undefined, suite: AllureGroup | undefined) => {
     if (!test) {
         return
     }
@@ -186,5 +186,7 @@ export const setHistoryId = (test: AllureTest | undefined, suite: AllureGroup | 
         .sort((a, b) => a.name?.localeCompare(b.name))
         .map(it => it.name + it.value)
         .join('')
-    test.historyId = md5(`${suite?.name}${test.wrappedItem.name}${paramsPart}`)
+    const hash = md5(`${suite?.name}${test.wrappedItem.name}${paramsPart}`)
+    test.historyId = hash
+    test.testCaseId = hash
 }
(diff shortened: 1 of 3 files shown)


packages/wdio-allure-reporter/src/utils.ts
```diff
@@ -186,5 +186,7 @@ export const setHistoryId = (test: AllureTest | undefined, suite: AllureGroup |
         .sort((a, b) => a.name?.localeCompare(b.name))
         .map(it => it.name + it.value)
         .join('')
-    test.historyId = md5(`${suite?.name}${test.wrappedItem.name}${paramsPart}`)
+    const hash = md5(`${suite?.name}${test.wrappedItem.name}${paramsPart}`)
+    test.historyId = hash
+    test.testCaseId = hash
```
BorisOsipov, line 191: You have to check that testCaseId is empty before setting calculated one because user can set id via `allure.addAllureId(customId)` API and we should prefer this id over calculated one.

BorisOsipov (changes requested):
Changes look good. Please add corresponding unit tests.

packages/wdio-allure-reporter/src/utils.ts
AlexKorTutu: turns out they are two different things, allure.addAllureId(customId) actually sets a label:
https://github.com/webdriverio/webdriverio/blob/main/packages/wdio-allure-reporter/src/reporter.ts#L619

test.testCaseId is a set property so i can't even access it without some sort of breaking into private properties of AllureTest  object.

AlexKorTutu: tried to upload a file that has both testCaseId and explicitly set label via allure.addAllureId
allure.addAllureId actually overrides the generated id.
So... it kind of works, but i'm not sure if it's a sustainable thing.

i'll try to write a test in suite.test.ts where i can assert the final result of a report.

BorisOsipov: >turns out they are two different things, allure.addAllureId(customId) actually sets a label:

It is the same thing.

BorisOsipov: >So... it kind of works, but i'm not sure if it's a sustainable thing.

I prefer not to set generated id all if user provides his own. No matter how TestOps itself handles two id's (as field and as 
label) at the same time. If user set id, we need to send only user's id.

So please add guard that will set generated id only if no allureid label exists.

AlexKorTutu: got it

BorisOsipov (approved):

BorisOsipov:
@AlexKorTutu Congratulations on your first WebdriverIO contribution 🎉

BorisOsipov:
Fixes #11531

baev:
There are a few issues with the PR:

1. `testCaseId` should not include test parameters
2. It should be set for all tests (even for the ones with ALLURE_ID)
3. `testCaseId` should not change between the runs 
4. `testCaseId` should not change for reruns

PR #11537, opened 2023-10-26
Fix name selector to allow other characters

aristotelos:
Fix the `name` selector strategy to allow other characters, such as parentheses.
For example, using `$('[name="test (Test)"]')` will now correctly result in using the `name` selector strategy instead of using the `css selector` strategy.

Fixes #11553

## Proposed changes

Fix the `name` selector strategy to allow other characters, such as parentheses.
For example, using `$('[name="test (Test)"]')` will now correctly result in using the `name` selector strategy instead of using the `css selector` strategy.

Fixes #11553

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/packages/webdriverio/src/utils/findStrategy.ts
+++ b/packages/webdriverio/src/utils/findStrategy.ts
@@ -104,7 +104,7 @@ const defineStrategy = function (selector: SelectorStrategy) {
     // Use name strategy if selector queries elements with name attributes for JSONWP
     // or if isMobile is used even when w3c is used
     // e.g. "[name='myName']" or '[name="myName"]'
-    if (stringSelector.search(/^\[name=("|')([a-zA-z0-9\-_.@=[\] ']+)("|')]$/) >= 0) {
+    if (stringSelector.search(/^\[name=(?:"(.[^"]*)"|'(.[^']*)')]$/) >= 0) {
         return 'name'
     }
     // Allow to move up to the parent or select current element
@@ -232,12 +232,12 @@ export const findStrategy = function (selector: SelectorStrategy, isW3C?: boolea
     }
     case 'name': {
         if (isMobile || !isW3C) {
-            const match = stringSelector.match(/^\[name=("|')([a-zA-z0-9\-_.@=[\] ']+)("|')]$/)
+            const match = stringSelector.match(/^\[name=(?:"(.[^"]*)"|'(.[^']*)')]$/)
             if (!match) {
                 throw new Error(`InvalidSelectorMatch. Strategy 'name' has failed to match '${stringSelector}'`)
             }
             using = 'name'
-            value = match[2]
+            value = match[1] || match[2]
         }
         break
     }
--- a/packages/webdriverio/tests/findStrategy.test.ts
+++ b/packages/webdriverio/tests/findStrategy.test.ts
@@ -29,17 +29,36 @@ describe('selector strategies helper', () => {
         expect(element.value).toBe('[name="searchinput"]')
     })
 
-    it('should find an element using "name" method with a . in the name', () => {
-        const element = findStrategy('[name="search.input"]')
-        expect(element.using).toBe('name')
-        expect(element.value).toBe('search.input')
-    })
-
-    it('should find an element using "name" method with an attribute', () => {
-        const element = findStrategy('[name="searchinput[@isDisplayed=\'true\']"]')
-        expect(element.using).toBe('name')
-        expect(element.value).toBe('searchinput[@isDisplayed=\'true\']')
-    })
+    it.each([
+        {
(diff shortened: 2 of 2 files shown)


erwinheitzman (changes requested):
Thank you for the contribution, looks good to me with the exception of a small request for change as the testcase is confusing.

packages/webdriverio/tests/findStrategy.test.ts
```diff
@@ -29,17 +29,35 @@ describe('selector strategies helper', () => {
...
+        {
+            selector: "[name='search with \"double quotes\"']",
+            expectedValue: 'search with "double quotes"',
+        },
+        {
+            selector: "[name=\"search with 'single quotes'\"]",
+            expectedValue: "search with 'single quotes'",
+        },
+        {
+            selector: "[name=\"searchinput[@isDisplayed='true']\"]",
```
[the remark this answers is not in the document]
aristotelos: This test case was already there, it was added in commit c41a299c33 for issue #3964. I have no idea what Youi.tv selectors are so left it intact...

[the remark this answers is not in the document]
aristotelos: Added a comment pointing to Youi.tv.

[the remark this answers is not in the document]
erwinheitzman: Thank you, this indeed adds a lot of context 👍

packages/webdriverio/src/utils/findStrategy.ts
```diff
@@ -232,12 +232,12 @@ export const findStrategy = function (selector: SelectorStrategy, isW3C?: boolea
...
     case 'name': {
         if (isMobile || !isW3C) {
-            const match = stringSelector.match(/^\[name=("|')([a-zA-z0-9\-_.@=[\] ']+)("|')]$/)
+            const match = stringSelector.match(/^\[name=(("([^"]*)")|('([^']*)'))]$/)
             if (!match) {
                 throw new Error(`InvalidSelectorMatch. Strategy 'name' has failed to match '${stringSelector}'`)
             }
             using = 'name'
-            value = match[2]
+            value = match[3] || match[5]
```
erwinheitzman, line 240: Just to be sure, I can't really tell how the match could have changed. Perhaps I am overlooking something but could you please explain this so we can be sure this is indeed correct?

The reason I ask is because it is still a single match right?

erwinheitzman (commented):
I just have a question, I assume it is correct because the tests are passing but from looking at the code it feels off somehow and I just want to make sure we are doing the right thing 🙂

packages/webdriverio/src/utils/findStrategy.ts
aristotelos: There are two possibilities: matching with double quotes or matching with single quotes. Both have different capture groups (group 3 and group 5) in the regular expression. It would be possible to make it a single capture with a single regular expression as before, but then the regular expression would also match e.g. `[name="test']` which is an invalid expression. The other option is to make 2 different regular expressions, one for double quotes and one for single quotes. I thought this would be easiest: get the capture from the double quotes if it exists (capture 3), otherwise get the capture from the single quotes (capture 5).

christian-bromann (approved):
LGTM 👍 

Wdyt @erwinheitzman ?

packages/webdriverio/src/utils/findStrategy.ts
erwinheitzman: I see now, I feel like we should optimize the regex a bit.
```js
\[name=(?:"(.[^"]*)"|'(.[^']*)')]
```
It would be more performant and returns a match on the full regex and 1 group.

What do you think about this?

erwinheitzman:
> LGTM 👍
> 
> Wdyt @erwinheitzman ?

I would like to optimize the regex first if you don't mind 🙂

packages/webdriverio/src/utils/findStrategy.ts
aristotelos: Thanks, pushed an update

christian-bromann:
@erwinheitzman wdyt?

christian-bromann:
Congratulations @aristotelos on your remarkable first contribution to WebdriverIO! This project thrives on the invaluable involvement of our community, and we are truly grateful for your contribution. We eagerly anticipate witnessing more of your exceptional work, so please don't hesitate to inform us if we can assist you in identifying intriguing areas where you can make further contributions. Join our lively [Discord](https://discord.webdriver.io/) channel and reach out to us; we would be delighted to connect with you. Your efforts are deeply appreciated, and we extend our heartfelt gratitude to you. 🙏 ❤️

PR #11542, opened 2023-10-26
(internal): change host from 0.0.0.0 to localhost

erwinheitzman:
## Proposed changes

[//]: # (Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue.)

## Types of changes

[//]: # (What types of changes does your code introduce to WebdriverIO?)
[//]: # (_Put an `x` in the boxes that apply_)

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## Checklist

[//]: # (_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._)

- [ ] I have read the [CONTRIBUTING](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md) doc
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] I have added proper type definitions for new commands (if appropriate)

## Further comments

[//]: # (If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...)

PR #11548, opened 2023-10-30
(webdriverio): add throttleCPU and throttleNetwork commands

erwinheitzman:
## Proposed changes

- Added a throttleNetwork and throttleCPU command, after which the throttle command has become an alias for throttleNetwork.
- Updated the debugging docs to explain the different causes of flakiness and how they can be debugged (using the throttleNetwork and throttleCPU command)

This allows our users to easily throttle both and debug flaky tests.
We can remove the throttle command itself in v9.

## Types of changes

- [x] Documentation update

## Checklist


packages/webdriverio/src/commands/browser/throttle.ts
```diff
@@ -1,157 +1,13 @@
...
-    await client.send(
-        'Network.emulateNetworkConditions',
-        typeof params === 'string'
-            ? NETWORK_PRESETS[params]
-            : params
-    )
-
-    return null
-}
+    const browser = getBrowserObject(this)
```
christian-bromann, line 11: Let's add a deprecation warning here

packages/webdriverio/src/commands/browser/throttleCPU.ts
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    // Connect to Chrome DevTools
+    await this.getPuppeteer()
+    if (!this.puppeteer) {
+        throw new Error('No Puppeteer connection could be established which is required to use this command')
+    }
+
+    const pages = await this.puppeteer.pages()
+    const client = await pages[0].target().createCDPSession()
```
christian-bromann, line 40: Let's double check that `pages` has at least one item. This could potentially fail if users work with multiple tabs.

christian-bromann (commented):
Some minor comments but overall looks good.

erwinheitzman:
Thanks @christian-bromann I have resolved your feedback 👍

packages/webdriverio/src/commands/browser/throttle.ts
```diff
@@ -1,157 +1,14 @@
...
-    await client.send(
-        'Network.emulateNetworkConditions',
-        typeof params === 'string'
-            ? NETWORK_PRESETS[params]
-            : params
-    )
-
-    return null
-}
+    console.log('Deprecated: use throttleNetwork instead')
```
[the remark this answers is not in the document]
christian-bromann: Also we maybe should change the wording to something like:

```
Command "throttle" is deprecated and will be removed with the next major version release! Use `throttleNetwork` instead.
```

erwinheitzman:
Is there a way to improve the failing test?

I don't see why we check the amount of browser commands

christian-bromann (approved):
I would love to see a better deprecation system because this message gets lost in the logs and I wonder if we can better amplify it in the stdout. However this should not block this PR to get merged.

LGTM 👍

erwinheitzman:
Can this be merged with the failing tests?

christian-bromann:
I think we should fix the one that is failing. I can take a look when I have time.

erwinheitzman:
> I think we should fix the one that is failing. I can take a look when I have time.

These are the same tests that are failing for #11542 and I honestly have no idea why it's failing in either of these PR's 😮 
Any pointers or help with this is greatly appreciated ❤️

christian-bromann:
> These are the same tests that are failing for #11542

No quite, in this PR this test is failing:

```
 FAIL  packages/webdriverio/tests/commands/browser.test.ts > browser commands list and strategies
AssertionError: expected [ 'puppeteer', '$$', '$', …(36) ] to deeply equal [ 'puppeteer', '$$', '$', …(37) ]
```

in #11542 the following tests are failing:

```
FAIL  packages/wdio-utils/tests/node/index.test.ts > startWebDriver > should start firefox driver
AssertionError: expected "spy" to be called with arguments: [ { port: 1234, foo: 'bar', …(1) } ]

 FAIL  packages/wdio-utils/tests/node/index.test.ts > startWebDriver > should start no driver or download geckodriver if binaries are defined
AssertionError: expected "spy" to be called with arguments: [ { cacheDir: Any{ …(3) }, …(2) } ]
```

For latter you have to adjust the assertions since you now set different arguments when starting a Chromedriver session. For this PR it seems like the throttle command is not attached to the scope properly.

packages/webdriverio/src/commands/browser.ts
```diff
@@ -29,7 +29,8 @@ export * from './browser/setCookies.js'
 export * from './browser/setTimeout.js'
 export * from './browser/setWindowSize.js'
 export * from './browser/switchWindow.js'
-export * from './browser/throttle.js'
```
christian-bromann, line 32: By removing this export, the command is not attached to the browser scope anymore which would cause a breaking change

packages/webdriverio/src/commands/browser.ts
```diff
@@ -29,7 +29,8 @@ export * from './browser/setCookies.js'
...
+export * from './browser/throttleNetwork.js'
```
christian-bromann, line 32:
```suggestion
export * from './browser/throttle.js'
export * from './browser/throttleNetwork.js'
```

packages/webdriverio/src/commands/browser/throttle.ts
```diff
@@ -1,157 +1,17 @@
...
-        downloadThroughput: -1,
-        uploadThroughput: -1
-    }
-} as const
-
-const NETWORK_PRESET_TYPES = Object.keys(NETWORK_PRESETS)
+const log = logger('webdriverio:throttle')
 
+/**
+ * Alias for throttleNetwork. The browser.throttle command will be removed in v9.
```
christian-bromann, line 8:
```suggestion
 * @deprecated use `browser.throttleNetwork` instead
```

website/docs/Debugging.md
```diff
@@ -127,3 +127,37 @@ If you are an [Atom](https://atom.io/) hacker you can try [`wdio-repl`](https://
...
+### Rendering speed
+To debug device speed related flakiness use the [throttleCPU](https://webdriver.io/docs/api/browser/throttleCPU) command.
+This will cause your pages to render slower which can be caused by many things like running multiple processes in your CI which could be slowing down your tests.
+```js
+await browser.throttleCPU(4)
+```
+
+### Test execution speed
+If your tests do not seem to be affected it is possible that WebdriverIO is faster than the update from the frontend framework / browser.
+This happens when using synchronous assertions, some examples of code that can break because of this:
```
christian-bromann, line 150:
```suggestion

If your tests do not seem to be affected it is possible that WebdriverIO is faster than the update from the frontend framework / browser. This happens when using synchronous assertions since WebdriverIO has no chance to retry these assertions anymore. Some examples of code that can break because of this:
```

christian-bromann (commented):
Resolve build issues.

christian-bromann (approved):
I will go ahead and merge this and gonna fix remaining test issues in `main`.

Thanks a lot, this is awesome 👍

Issue #11544, opened 2023-10-26
[💡 Feature]: Add scroll behavior to w3c moveTo, click with offset arguments, and dnd methods


udarrr:
### Is your feature request related to a problem?

Currently scrolling should be implemented in test level. 
For example if i'd like to move pointer to element with scrolling I should do not so obvious action like scrolling because element can be shifted for some reason.

1. moveTo
```
await (await browser.$("#root div.theme-switcher-menu > button > span")).scrollIntoView()
await browser.pause(animationTimeout)
await (await browser.$("#root div.theme-switcher-menu > button > span")).moveTo()
```

2. w3c click with x,y
```
await (await browser.$("#root div.theme-switcher-menu > button > span")).scrollIntoView()
await browser.pause(animationTimeout)
await (await browser.$("#root div.theme-switcher-menu > button > span")).click({x: 1, y: 1})
```

3. for dragAndDrop it can't be performed because of current implementation 

```
    const browser = getBrowserObject(this);
    return browser.action('pointer')
        .move({ duration: 0, origin, x: 0, y: 0 })
        .down({ button: ACTION_BUTTON })
        .pause(10)
        .move({ duration, origin: targetOrigin, x: targetX, y: targetY })
        .up({ button: ACTION_BUTTON })
        .perform();
```

### Describe the solution you'd like.

I'd like to suggest add scrollIntoView by default to the methods and some arguments for parametrizing the behavior.

1. moveTo

just with arguments by default

`await (await browser.$("#root div.theme-switcher-menu > button > span")).moveTo()`

and with preset arguments

`await (await browser.$("#root div.theme-switcher-menu > button > span")).moveTo({x: 10, y: 10, scrollOptions: options})`

2. the same for w3c clicking with x,y
3. and maybe for dnd if it possible to iterate with the same 'pointer' after using 'wheel'

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

christian-bromann:
@udarrr thanks for raising the issue.

Do you have some examples where this additional `moveTo` step benefits the user?

Note that an element that is not in viewport might not be able to be scrolled into view. Relying on the WebDriver protocol to do its thing and move the element into the viewport is usually a good step. Also be aware that adding these commands can increase test execution time when WebDriver endpoints are hosted in the cloud.

udarrr:
> Do you have some examples where this additional `moveTo` step benefits the user?

[...]

PR #11565, opened 2023-10-31
autoscroll for moveto and w3c click

udarrr:
related to https://github.com/webdriverio/webdriverio/issues/11544


packages/webdriverio/src/commands/element/moveTo.ts
```diff
@@ -30,6 +32,7 @@ export async function moveTo (
      * W3C way of handle the mouse move actions
      */
     const browser = getBrowserObject(this)
+    await moveInViewport(this, scrollTimeout ? scrollTimeout : 2000)
```
erwinheitzman, line 35: Let's get rid of moveInViewport, it's doing the same thing as the scrollIntoView command, so we can use that.

e2e/wdio/headless/secondTest.e2e.ts
```diff
@@ -5,8 +5,8 @@ describe('main suite 1', () => {
         const caps = browser.capabilities as WebdriverIO.Capabilities
         const assertionValue = caps.browserName === 'msedge'
             ? 'headlessedg'
-            : caps.browserName!.toLowerCase()
+            : caps.browserName && caps.browserName.includes('chrome') ? 'chrome/' : caps.browserName!.toLowerCase()
```
erwinheitzman, line 8: Why did this have to be changed?

e2e/wdio/headless/test.e2e.ts
```diff
@@ -107,8 +107,7 @@ describe('main suite 1', () => {
         ]
 
         before(async () => {
-            await browser.url('http://guinea-pig.webdriver.io/')
-            await browser.$('a[href="pointer.html"]').click()
```
erwinheitzman, line 111: Why was the click removed? This potentionally affects other tests

packages/webdriverio/src/commands/element/click.ts
```diff
@@ -123,6 +124,7 @@ export async function click(
 
     if (this.isW3C) {
         const browser = getBrowserObject(this)
+        await moveInViewport(this, 2000)
```
erwinheitzman, line 127: Let's just scroll to it using scrollIntoView. Checking if the element is in the viewport is just as expensive of an operation so in the best case you only do the check but in the worst you do the check, then scroll. Having to do both for multiple elements is expensive. Especially when running in a cloud environment where the roundtrip takes a very long time

erwinheitzman (changes requested):

e2e/wdio/headless/test.e2e.ts
udarrr: The suite of test related to only moveTo tests and it's been changed directly to http://guinea-pig.webdriver.io/pointer.html' not necessary click it here, the rest cases call browser.url with their own links and open it each time when it's been called

packages/webdriverio/src/commands/element/moveTo.ts
udarrr: No exactly the version is 

`export async function isInViewport (elm: WebdriverIO.Element) {
    const browser = getBrowserObject(elm)
    return browser.execute(isElementInViewportScript, {
        [ELEMENT_KEY]: elm.elementId, // w3c compatible
        ELEMENT: elm.elementId // jsonwp compatible
    } as any as HTMLElement)
}
export async function moveInViewport (elm: WebdriverIO.Element, scrollTimeout  = 2000) {
    const browser = getBrowserObject(elm)
    try {
        if (await isInViewport(elm)) {
            await  elm.scrollIntoView({ block : 'nearest', inline: 'nearest', behavior: 'instant' })
            await browser.waitUntil(async ()=> await isInViewport(elm), { timeout: scrollTimeout })
        }
    } catch {
        await  elm.scrollIntoView({ block : 'center', inline: 'center', behavior: 'instant' })
    }
}`

it's checking isInViewport before scrolling, wait for animation in browser.waitUntil and then use scrollIntoView again if it's  still not into view

packages/webdriverio/src/commands/element/click.ts
udarrr: I agree action like waiting for animation could be omitted
```
await browser.waitUntil(async ()=> await isInViewport(elm), { timeout: scrollTimeout })
```
but rest of them still usefull because why we should move it if it's in intoview ?

by default web version of scrollIntoView with the option like 'nearest' have the check, into view or not, under hood, but wdio implementation using action if it's not failed. That's why it's going to be called each time,

udarrr: Using {block: "start"}, the element is aligned at the top of its ancestor.
Using {block: "center"}, the element is aligned at the middle of its ancestor.
Using {block: "end"}, the element is aligned at the bottom of its ancestor.

Using {block: "nearest"}, the element:
is aligned at the top of its ancestor if you're currently below it.
is aligned at the bottom of its ancestor if you're currently above it.
stays put, if it's already in view.

packages/webdriverio/src/utils/inViewport.ts
```diff
@@ -0,0 +1,22 @@
...
+    return browser.execute(isElementInViewportScript, {
+        [ELEMENT_KEY]: elm.elementId, // w3c compatible
+        ELEMENT: elm.elementId // jsonwp compatible
+    } as any as HTMLElement)
+}
+export async function moveInViewport (elm: WebdriverIO.Element, scrollTimeout  = 2000) {
+    const browser = getBrowserObject(elm)
+    try {
+        if (await isInViewport(elm)) {
+            await  elm.scrollIntoView({ block : 'nearest', inline: 'nearest', behavior: 'instant' })
```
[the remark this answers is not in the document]
udarrr: No, in first case it's { block : 'nearest', inline: 'nearest', behavior: 'instant' }
in second { block : 'center', inline: 'center', behavior: 'instant' }

packages/webdriverio/src/commands/element/moveTo.ts
erwinheitzman: But why do this when scrollIntoView already has that fallback?

packages/webdriverio/src/commands/element/click.ts
erwinheitzman: I know, I have refactored the scrollIntoView command not too long ago to verify all this behaviour.

If we are going to send a script to the server it doesn't matter if you send the check or the scroll itself. Scrolling or checking if the element is in the viewport both take about the same amount of time so you gain nothing by first doing the check in my opinion.
I understand why you introduced the check but I think it is better to leave it out and just try to scroll each element into the viewport as it's instant with the new scrollIntoView implementation anyway.

packages/webdriverio/src/utils/inViewport.ts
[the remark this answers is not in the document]
christian-bromann: ah gotcha, thanks for the clarification

packages/webdriverio/src/commands/element/moveTo.ts
udarrr: For the same reason like it's been implemented in isElementClickable here

https://github.com/webdriverio/webdriverio/blob/main/packages/webdriverio/src/scripts/isElementClickable.ts

// Should definitely work even if element was covered with sticky header/footer

packages/webdriverio/src/commands/element/click.ts
erwinheitzman: Having that said, if Christian is fine with the way you implemented it. I am fine with resolving this, I just want to share what I feel would be the best approach 🙂

packages/webdriverio/src/commands/element/moveTo.ts
udarrr: > But why do this when scrollIntoView already has that fallback?

I use it twice

await  elm.scrollIntoView({ block : 'nearest', inline: 'nearest', behavior: 'instant' })
await  elm.scrollIntoView({ block : 'center', inline: 'center', behavior: 'instant' })

unfortunately i can't figure out where is callback in scrollIntoView with the options.

udarrr: let's get imagine, i'm calling scrollIntoView like
await elm.scrollIntoView({ block : 'nearest', inline: 'nearest', behavior: 'instant' })

after some calculation it's failed here and throw error for example

await browser.action('wheel')
            .scroll({ duration: 0, x: deltaX, deltaY, origin: this })
            .perform()

the it's moved to catch and tried to call scrollIntoView web version but with the same options like { block : 'nearest', inline: 'nearest', behavior: 'instant' }, element still isn't available

packages/webdriverio/src/commands/element/click.ts
udarrr: I got it... ok maybe it's worth to delete first check but still have the second ?

udarrr: async function moveInViewport (elm: WebdriverIO.Element) {
        await  elm.scrollIntoView({ block : 'nearest', inline: 'nearest', behavior: 'instant' })

        if (await isInViewport(elm)) {
            await  elm.scrollIntoView({ block : 'center', inline: 'center', behavior: 'instant' })
        }
}

udarrr: Moreover I'd suggest use it only for moveTo

udarrr:
@christian-bromann  @erwinheitzman  Could review it again ?

packages/webdriverio/src/commands/element/moveTo.ts
```diff
@@ -26,10 +28,20 @@ export async function moveTo (
...
+        return await browser.execute(isElementInViewportScript, {
+            [ELEMENT_KEY]: this.elementId, // w3c compatible
+            ELEMENT: this.elementId // jsonwp compatible
+        } as any as HTMLElement)
+    }
     /**
      * W3C way of handle the mouse move actions
      */
     const browser = getBrowserObject(this)
+    await  this.scrollIntoView({ block : 'nearest', inline: 'nearest', behavior: 'instant' })
```
christian-bromann, line 41:
```suggestion
    await this.scrollIntoView({ block : 'nearest', inline: 'nearest', behavior: 'instant' })
```

packages/webdriverio/src/commands/element/moveTo.ts
```diff
@@ -26,10 +28,20 @@ export async function moveTo (
...
+    if (!(await isIntoView())) {
+        await  this.scrollIntoView({ block : 'center', inline: 'center', behavior: 'instant' })
```
christian-bromann, line 43:
```suggestion
        await this.scrollIntoView({ block : 'center', inline: 'center', behavior: 'instant' })
```

packages/webdriverio/tests/commands/element/moveTo.test.ts
```diff
@@ -9,43 +9,6 @@ vi.mock('got')
 vi.mock('@wdio/logger', () => import(path.join(process.cwd(), '__mocks__', '@wdio/logger')))
 
 describe('moveTo', () => {
-    it('should do a moveTo without params', async () => {
```
christian-bromann, line 12: Why do we remove these tests?

christian-bromann (commented):
One question, otherwise LGTM

packages/webdriverio/tests/commands/element/moveTo.test.ts
udarrr: Generally it's my bad, I've thought scroll can work only with e2e here, but the assumption was wrong, it can be run with mock like elem.elementId = { scrollIntoView: vi.fn() }. I've restored the test cases

packages/webdriverio/src/commands/element/moveTo.ts
udarrr: deleted in https://github.com/webdriverio/webdriverio/pull/11565/commits/44f648b9b9a4a145063331541ccf96285badcd2f

packages/webdriverio/src/commands/element/moveTo.ts
udarrr: deleted in https://github.com/webdriverio/webdriverio/pull/11565/commits/44f648b9b9a4a145063331541ccf96285badcd2f

christian-bromann (approved):
LGTM 👍

packages/webdriverio/src/commands/element/click.ts
erwinheitzman: there's already a retry built into the scrollIntoView so this is not needed.

christian-bromann:
> could you trigger new release

Yes, I will package a new release by EOD

PR #11570, opened 2023-11-01
Adding ability to get pseudo-elements css value via getCSSProperty #7709

Pawel1894:
## Proposed changes
Added possibility to select pseudo-element css value. It is not defined in Webdriver spec so I did it via browser execute and did not modify protocols.

I didn't found a way to get browser element other than passing it as argument. I also did split `getCSSProperty` to smaller functions because it was hard to understand with all the comments and pseudoElement ifs.

Example usage:
```js
element.getCSSProperty('width', {
   target: '::before',
   browser
})
```

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

I don't know how to mock window object with `got.ts` to be able to add tests for it in `getCSSProperty.test.ts`. I tried to use vitest spyOn to mock window object but it does not work.

### DIFF
--- a/packages/webdriverio/src/commands/element/getCSSProperty.ts
+++ b/packages/webdriverio/src/commands/element/getCSSProperty.ts
@@ -77,15 +77,12 @@ export async function getCSSProperty (
     cssProperty: string,
     pseudoElement?: PseudoElement,
 ) {
-    const browser = getBrowserObject(this)
-
     if (cssShorthandProps.isShorthand(cssProperty)) {
         const cssValue = await getShorthandPropertyCSSValue.call(
             this,
             {
                 cssProperty,
                 pseudoElement,
-                browser
             }
         )
 
@@ -97,7 +94,6 @@ export async function getCSSProperty (
         {
             cssProperty,
             pseudoElement,
-            browser
         }
     )
 
@@ -106,7 +102,6 @@ export async function getCSSProperty (
 
 type Options = {
     cssProperty: string;
-    browser: WebdriverIO.Browser | WebdriverIO.MultiRemoteBrowser;
     pseudoElement?: PseudoElement;
 }
 
@@ -114,17 +109,16 @@ async function getShorthandPropertyCSSValue(
     this: WebdriverIO.Element,
     options: Options
 ) {
-    const { pseudoElement, browser, cssProperty } = options
+    const { pseudoElement, cssProperty } = options
     const properties = getShorthandProperties(cssProperty)
 
     if (pseudoElement) {
         const cssValues = await Promise.all(
-            properties.map((cssProperty) => getPseudoElementCSSValue(
+            properties.map((prop) => getPseudoElementCSSValue(
                 this,
                 {
-                    browser,
                     pseudoElement,
-                    cssProperty,
+                    cssProperty: prop,
                 }
             ))
         )
@@ -148,7 +142,10 @@ async function getPropertyCSSValue(
     if (pseudoElement) {
         return await getPseudoElementCSSValue(
             this,
-            options
+            {
+                pseudoElement,
+                cssProperty
+            }
         )
     }
     return await this.getElementCSSValue(this.elementId, cssProperty)
@@ -190,12 +187,13 @@ function mergeEqualSymmetricalValue(cssValues: string[]) {
 
 async function getPseudoElementCSSValue (
     elem: WebdriverIO.Element,
-    options: Options
+    options: Required<Options>
 ): Promise<string> {
-    const { browser, cssProperty, pseudoElement }  = options
+    const browser = getBrowserObject(elem)
+    const { cssProperty, pseudoElement }  = options
     const cssValue = await browser.execute(
         (elem: Element, pseudoElement: string, cssProperty: string) => (window.getComputedStyle(elem, pseudoElement) as any)[cssProperty],
-        elem,
+        elem as unknown as Element,
         pseudoElement,
         cssProperty
     )
--- a/packages/webdriverio/src/commands/element/getCSSProperty.ts
+++ b/packages/webdriverio/src/commands/element/getCSSProperty.ts
@@ -1,10 +1,7 @@
 import cssShorthandProps from 'css-shorthand-properties'
-import { parseCSS } from '../../utils/index.js'
+import { getBrowserObject, parseCSS } from '../../utils/index.js'
 
-type PseudoElement = {
-    target: '::before'|'::after',
-    browser: WebdriverIO.Browser | WebdriverIO.MultiRemoteBrowser
- }
+type PseudoElement = '::before'|'::after'
 
 /**
  *
@@ -54,10 +51,7 @@ type PseudoElement = {
         //      }
         // }
 
-        var width = await elem.getCSSProperty('width', {
-            target: '::before',
-            browser
-        })
+        var width = await elem.getCSSProperty('width', '::before')
         console.log(width)
         // outputs the following:
         // {
@@ -83,11 +77,16 @@ export async function getCSSProperty (
     cssProperty: string,
     pseudoElement?: PseudoElement,
 ) {
+    const browser = getBrowserObject(this)
+
     if (cssShorthandProps.isShorthand(cssProperty)) {
         const cssValue = await getShorthandPropertyCSSValue.call(
             this,
-            cssProperty,
-            pseudoElement,
+            {
+                cssProperty,
+                pseudoElement,
+                browser
+            }
         )
 
         return parseCSS(cssValue, cssProperty)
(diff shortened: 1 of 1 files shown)


packages/webdriverio/src/commands/element/getCSSProperty.ts
```diff
@@ -1,6 +1,11 @@
 import cssShorthandProps from 'css-shorthand-properties'
 import { parseCSS } from '../../utils/index.js'
 
+type PseudoElement = {
+    target: '::before'|'::after',
+    browser: WebdriverIO.Browser | WebdriverIO.MultiRemoteBrowser
```
christian-bromann, line 6: Why should users pass in the `browser` instance? You can just use `this` or use the `getBrowserObject` to get the browser instance from an element

Pawel1894: I updated the code, browser is no longer in params, I get it via `getBrowserObject ` inside `getCSSProperty`. As for using `execute` on element it is not working.

Pawel1894:
> It would be nice to get unit tests for this.

As I wrote in PR comment. I have tried to mock window object and make it work with got, but nothing worked, no idea how to do it.

packages/webdriverio/src/commands/element/getCSSProperty.ts
```diff
@@ -83,66 +77,78 @@ export async function getCSSProperty (
...
+
     if (cssShorthandProps.isShorthand(cssProperty)) {
         const cssValue = await getShorthandPropertyCSSValue.call(
             this,
-            cssProperty,
-            pseudoElement,
+            {
+                cssProperty,
+                pseudoElement,
+                browser
```
christian-bromann, line 88: No need to pass this along since you can call `const browser = getBrowserObject(this)` in `getShorthandPropertyCSSValue`

packages/webdriverio/src/commands/element/getCSSProperty.ts
```diff
@@ -83,66 +77,78 @@ export async function getCSSProperty (
...
     }
 
     const cssValue = await getPropertyCSSValue.call(
         this,
-        cssProperty,
-        pseudoElement,
+        {
+            cssProperty,
+            pseudoElement,
+            browser
```
christian-bromann, line 100: same

packages/webdriverio/src/commands/element/getCSSProperty.ts
Pawel1894: uptated - silly mistake by me sorry 😞

packages/webdriverio/src/commands/element/getCSSProperty.ts
```diff
@@ -64,50 +66,137 @@ import { parseCSS } from '../../utils/index.js'
...
+
+async function getPseudoElementCSSValue (
+    elem: WebdriverIO.Element,
+    options: Required<Options>
+): Promise<string> {
+    const browser = getBrowserObject(elem)
+    const { cssProperty, pseudoElement }  = options
+    const cssValue = await browser.execute(
+        (elem: Element, pseudoElement: string, cssProperty: string) => (window.getComputedStyle(elem, pseudoElement) as any)[cssProperty],
+        elem as unknown as Element,
```
Pawel1894, line 196: I don't know why it was ok for typescript earlier when I was passing browser via params, but now i got error that `WebdriverIO.Element` can't be used as `Element`. 
The only solution i found is this ugly casting

packages/webdriverio/src/commands/element/getCSSProperty.ts
Pawel1894: uptated

packages/webdriverio/src/commands/element/getCSSProperty.ts
christian-bromann: Yeah, this is because the browser driver transforms the `WebdriverIO.Element` reference into a web `Element` reference and we haven't found a nice way to teach TypeScript that 😉 I think this is fine for now

packages/webdriverio/src/commands/element/getCSSProperty.ts
```diff
@@ -64,50 +66,137 @@ import { parseCSS } from '../../utils/index.js'
...
+
+    const cssValue = await getPropertyCSSValue.call(
+        this,
+        {
+            cssProperty,
+            pseudoElement,
+        }
+    )
+
+    return parseCSS(cssValue, cssProperty)
```
christian-bromann, line 100: We can simplify this to:
```suggestion
    const getCSSProperty = cssShorthandProps.isShorthand(cssProperty)
        ? getShorthandPropertyCSSValue
        : getPropertyCSSValue

    const cssValue = await getCSSProperty.call(
        this,
        {
            cssProperty,
            pseudoElement,
        }
    )

    return parseCSS(cssValue, cssProperty)
```

christian-bromann:
> I have tried to mock window object and make it work with got, but nothing worked, no idea how to do it.

There is no need to mock the window object. We already have [some unit tests](https://github.com/webdriverio/webdriverio/blob/aa891330bbb517d106024bc95d0c344388423373/packages/webdriverio/tests/commands/element/getCSSProperty.test.ts) for the command which mocks the `got` package to fake requests to browser drivers (so the function within `execute` is never really being run). You can find the mock specification in [`./__mocks__/got.ts`](https://github.com/webdriverio/webdriverio/blob/8c2d65c49ac37a348ea04d9bff33cbc5d20d06ab/__mocks__/got.ts#L187-L198).

Pawel1894:
> > I have tried to mock window object and make it work with got, but nothing worked, no idea how to do it.
> 
> There is no need to mock the window object. We already have [some unit tests](https://github.com/webdriverio/webdriverio/blob/aa891330bbb517d106024bc95d0c344388423373/packages/webdriverio/tests/commands/element/getCSSProperty.test.ts) for the command which mocks the `got` package to fake requests to browser drivers (so the function within `execute` is never really being run). You can find the mock specification in [`./__mocks__/got.ts`](https://github.com/webdriverio/webdriverio/blob/8c2d65c49ac37a348ea04d9bff33cbc5d20d06ab/__mocks__/got.ts#L187-L198).

Nice, I'll look into it on saturday 😄

Pawel1894:
> > > I have tried to mock window object and make it work with got, but nothing worked, no idea how to do it.
> > 
> > 
> > There is no need to mock the window object. We already have [some unit tests](https://github.com/webdriverio/webdriverio/blob/aa891330bbb517d106024bc95d0c344388423373/packages/webdriverio/tests/commands/element/getCSSProperty.test.ts) for the command which mocks the `got` package to fake requests to browser drivers (so the function within `execute` is never really being run). You can find the mock specification in [`./__mocks__/got.ts`](https://github.com/webdriverio/webdriverio/blob/8c2d65c49ac37a348ea04d9bff33cbc5d20d06ab/__mocks__/got.ts#L187-L198).
> 
> Nice, I'll look into it on saturday 😄

Ok, so I tried to make test work , but without success.

When I use `window.getComputedStyle` inside `execute` I'm getting this error:

**ReferenceError: window is not defined**
```js
eval __mocks__/got.ts:207:24
    205|     case `/session/${sessionId}/execute/sync`: {
    206|         // mock execute
    207|         const script = Function(params.json.script)
       |                        ^
    208|         const args = params.json.args.map((arg: any) => (arg && (
```
I tried to look for examples, I used some other window properties like window.scrollY inside my function and it worked.. but this one does not. Do I need to also add this specific function somewhere in mocks to make it work?

christian-bromann:
@Pawel1894 for this test you can try to attach a mocked window object to the global scope, e.g. `globalThis.window = { getComputedStyle: vi.fn() }`

Pawel1894:
> @Pawel1894 for this test you can try to attach a mocked window object to the global scope, e.g. `globalThis.window = { getComputedStyle: vi.fn() }`

So i tried what you recommended, also this but with `mockReturnValue` and `mockImplementation` and I always end up getting with empty object returned from `execute` function.

So I end up with this
```ts
TypeError: Cannot read properties of undefined (reading 'padding-top')
 ❯ eval __mocks__/got.ts:206:24
    204|     case `/session/${sessionId}/execute`:
    205|     case `/session/${sessionId}/execute/sync`: {
    206|         const script = Function(params.json.script)
```

`vi.spyOn(browser, 'execute').mockReturnValue('12px')` <- this worked for me but it won't allow me to write test for shorthand property, so I'm not happy with this solution either.

Pawel1894:
> > also this but with `mockReturnValue` and `mockImplementation`
> 
> I am not sure if you need these. Instead make sure the mock (`./__mocks__/got.ts`) returns the expected value for your needs. Also feel free to stub the browser objects completely if you don't want to use the mock.

@christian-bromann Tbh it is really hard for me to understand how code in `mocks got.ts` works. Do you mean by  _feel free to stub the browser objects_ to mock the `getPseudoElementCSSValue`?

christian-bromann (approved):
I will go ahead and merge this and will add necessary unit tests as an additional PR.

Thanks for the contribution 👍

christian-bromann:
See https://github.com/webdriverio/webdriverio/commit/d68bc2e9f8f60e3e252b005f268b7396245c862d

Pawel1894:
@christian-bromann Thx 👍

PR #11590, opened 2023-11-06
(webdriverio): remove type exports

### LINKED ISSUE #11055: [🐛 Bug]: Browser type primitive mismatches WebdriverIO.Browser global type

### WebdriverIO Version

Tested in 8.15.10 (latest as of writing), introduced in 8.11.0

### Node.js Version

Verified in 18.17.1, 20.5.1

### Mode

Standalone Mode

```typescript
n/a
```

### What happened?

As of PR https://github.com/webdriverio/webdriverio/pull/9640, WebdriverIO's typings include two copies of the `Browser` type, one a global type within the `WebdriverIO` namespace and one a type primitive exported from the module.

PR https://github.com/webdriverio/webdriverio/pull/10447 made an update to to the primitive export, but not the global namespace version.

In practice, this means that code which tries to use `import type { Browser } from 'webdriverio'` is incompatible with webdriverio functions that return `WebdriverIO.Browser`, such as `remote`.

The two copies of the type should match, ideally with one defined in terms of the other to avoid this issue in the future

### How to reproduce the bug.

```ts
// repro.ts
import { type Browser, remote } from 'webdriverio';

async function main() {
    let browser: Browser; // works if using WebdriverIO.Browser instead

    browser = await remote({
        capabilities: { browserName: 'irrelevant' }
    });
}

```

```sh
$ npm init -y
$ npm install webdriverio@8.15.10 typescript@5.2.2
$ npx tsc --noEmit --skipLibCheck ./repro.ts
```

### Relevant log output

```typescript
repro.ts:7:5 - error TS2740: Type 'Browser' is missing the following properties from type 'Browser': sessionStatus, sessionNew, sessionEnd, sessionSubscribe, and 25 more.

7     browser = await remote({
      ~~~~~~~


Found 1 error in repro.ts:
```

### Code of Conduct

christian-bromann:
## Proposed changes

We currently export all WebdriverIO types from the module, including the ones for `Browser` and `Element`. These are not meant to be used for public consumption, instead folks should use the `WebdriverIO` namespace, e.g. `WebdriverIO.Browser`.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

fixes #11055

### DIFF
--- a/packages/webdriverio/src/index.ts
+++ b/packages/webdriverio/src/index.ts
@@ -15,9 +15,7 @@ import { getPrototype, addLocatorStrategyHandler, isStub } from './utils/index.j
 import type { AttachOptions, RemoteOptions } from './types.js'
 import type * as elementCommands from './commands/element.js'
 
-export * from './types.js'
-export * from './utils/interception/types.js'
-
+export { Selector, CustomInstanceCommands } from './types.js'
 export const Key = KeyConstant
 export const SevereServiceError = SevereServiceErrorImport
 


dbjorge:
@christian-bromann This is a breaking change for folks that used the previously-exported types. I definitely understand the desire to avoid multiple definitions and think it's reasonable to collapse down to 1 exposed set of typings, but it should be done in a breaking change release, not a bugfix.

christian-bromann:
I disagree this being a breaking change since we never documented nor advertised the use of these types.

dbjorge:
Of course, it's your project and it's your prerogative to decide what should constitute a breaking change for it, but please understand that many reasonable folks will consider "a type that is exported from the package's index" to be an advertised part of the package's API, regardless of explicit documentation about it. Realistically, more folks discover types through IDE auto-completion than through documentation.

In actual practice, you can find 1000+ cases that this would break between the following two GitHub code search queries (which are not exhaustive):

* [`/import type .* from 'webdriverio'/`](https://github.com/search?q=%2Fimport+type+.*+from+%27webdriverio%27%2F&type=code)
* [`/import \{.*(type|Browser|Element).*\} from 'webdriverio'/`](https://github.com/search?q=%2Fimport+%5C%7B.*%28type%7CBrowser%7CElement%29.*%5C%7D+from+%27webdriverio%27%2F&type=code)

Is it truly so important to get this change in *now*, rather than waiting for the next major version, that it's worth breaking 1000+ known usages?

christian-bromann:
@dbjorge you are making a good point! I am surprised it is used that much. I reverted the change and added a deprecation warning.

PR #11625, opened 2023-11-07
fix jasmine afterTest hook

### LINKED ISSUE #11495: [🐛 Bug]: afterTest hook contains wrong test object but correct error in a specific scenario

### WebdriverIO Version

8.20.2

### Node.js Version

18.6.1

### Mode

Standalone Mode

_No response_

### What happened?

The afterTest hook is called containing the wrong test object but correct error in a specific scenario

Using the example spec from the documentation I have added two tests to demonstrate the issue. The first one fails due to a time out because the last line does not contain the text it is expecting ('d' removed from 'logged'). The second test case fails because of an exception (k not defined).

- When running with Mocha test framework -  the afterTest hook is called correctly.
- When running with the Jasmine test framework - the afterTest hook is called incorrectly. The test object in both cases is the second test case, but the error is different (and correct) in each invocation.

```
import { expect } from '@wdio/globals'
import LoginPage from '../pageobjects/login.page.js'
import SecurePage from '../pageobjects/secure.page.js'

describe('My Login application', () => {
    it('timeout failure', async () => {
        await LoginPage.open()
        await LoginPage.login('tomsmith', 'SuperSecretPassword!')
        await expect(true).toBe(true);
        await expect(SecurePage.flashAlert).toBeExisting()
        await expect(SecurePage.flashAlert).toHaveTextContaining(
            'You logge into a secure area!')
    })
    it('exception fail rather than assert', async () => {
        await LoginPage.open()
        await LoginPage.login('tomsmith', 'SuperSecretPassword!')
        k++; // reference error, k not defined
        await expect(true).toBe(true);
        await expect(SecurePage.flashAlert).toBeExisting()
        await expect(SecurePage.flashAlert).toHaveTextContaining(
            'You logged into a secure area!')
    })
})

```

If you change the order of the test cases (in this case reverse them) so that the timed out test is last then the afterTest hook is called correctly for both Mocha and Jasmin:

```
import { expect } from '@wdio/globals'
import LoginPage from '../pageobjects/login.page.js'
import SecurePage from '../pageobjects/secure.page.js'

describe('My Login application', () => {
    it('exception fail rather than assert', async () => {
        await LoginPage.open()
        await LoginPage.login('tomsmith', 'SuperSecretPassword!')
        k++; // reference error, k not defined
        await expect(true).toBe(true);
        await expect(SecurePage.flashAlert).toBeExisting()
        await expect(SecurePage.flashAlert).toHaveTextContaining(
            'You logged into a secure area!')
    })
   it('timeout failure', async () => {
        await LoginPage.open()
        await LoginPage.login('tomsmith', 'SuperSecretPassword!')
        await expect(true).toBe(true);
        await expect(SecurePage.flashAlert).toBeExisting()
        await expect(SecurePage.flashAlert).toHaveTextContaining(
            'You logge into a secure area!')
    })
})

```

The afterTest hook should contain the correct test object.

### How to reproduce the bug.

1. Create a new wdio project
2. Use the example spec from the documentation but modified in the way shown above in the 'What happened' section.
3. Set test framework to Jasmine (the issue does not occur with Mocha)
4. Add a console.log of the test object return in the afterTest hook (or use debugger/breakdpoint).
5. You will see the same test object in the afterTest hook but the error will be different (and correct) in the two invocations.

### Relevant log output

```typescript
Follow steps in How to reproduce the bug. This can be reproduced with a new project quickly.
```

### Code of Conduct

tamil777selvan:
## Proposed changes

Fixes #11495

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/wdio-utils/src/shim.ts
+++ b/packages/wdio-utils/src/shim.ts
@@ -289,8 +289,12 @@ export async function executeAsync(this: any, fn: Function, retries: Frameworks.
     this.wdioRetries = retries.attempts
 
     try {
+        /**
+         * Decrease the overall timeout by 3 milliseconds to ensure that when a test fails due to a timeout exception in Jasmine,
+         * the test objects are returned instead of being overridden by the values of the next test.
+         */
         // @ts-expect-error
-        const _timeout = this?._runnable?._timeout || globalThis.jasmine?.DEFAULT_TIMEOUT_INTERVAL || timeout
+        const _timeout = (this?._runnable?._timeout || globalThis.jasmine?.DEFAULT_TIMEOUT_INTERVAL || timeout) - 3
         /**
          * Executes the function with specified timeout and returns the result, or throws an error if the timeout is exceeded.
          */
--- a/packages/wdio-utils/src/shim.ts
+++ b/packages/wdio-utils/src/shim.ts
@@ -18,6 +18,8 @@ const ELEMENT_PROPS = [
 const ACTION_COMMANDS = ['action', 'actions']
 const PROMISE_METHODS = ['then', 'catch', 'finally']
 
+const TIME_BUFFER = 3
+
 export async function executeHooksWithArgs<T> (this: any, hookName: string, hooks: Function | Function[] = [], args: any[] = []): Promise<(T | Error)[]> {
     /**
      * make sure hooks are an array of functions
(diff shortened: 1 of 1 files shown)


packages/wdio-utils/src/shim.ts
```diff
@@ -289,8 +289,12 @@ export async function executeAsync(this: any, fn: Function, retries: Frameworks.
     this.wdioRetries = retries.attempts
 
     try {
+        /**
+         * Decrease the overall timeout by 3 milliseconds to ensure that when a test fails due to a timeout exception in Jasmine,
+         * the test objects are returned instead of being overridden by the values of the next test.
+         */
         // @ts-expect-error
-        const _timeout = this?._runnable?._timeout || globalThis.jasmine?.DEFAULT_TIMEOUT_INTERVAL || timeout
+        const _timeout = (this?._runnable?._timeout || globalThis.jasmine?.DEFAULT_TIMEOUT_INTERVAL || timeout) - 3
```
christian-bromann, line 297: Do we know that 3ms are enough for this to not happen? It seems a bit arbitrary.

tamil777selvan: It does not need to be precisely 3ms. It just needs to be a shorter duration than the overall timeout. So, I've used 3ms, making sure it doesn't disrupt the timeouts set & execution timeline.

christian-bromann: Can we move the `3` to a const at the top of the file with a descriptive name and a comment as to why the value is chosen? I think it would greatly help anyone stumbling over it in the future.

tamil777selvan: sure, on it.

christian-bromann (approved):
Awesome, LGTM 👍

PR #11655, opened 2023-11-13
Fix key module export

erwinheitzman:
## Proposed changes

Some keys do not exist due to having two different truths, this has now been corrected.

Example:
Screenshot 2023-11-13 at 14 30 15

## Types of changes

- [ ] Documentation update

## Checklist


christian-bromann:
@erwinheitzman can you describe how you experienced this error? These objects (`Key` vs. `UNICODE_CHARACTERS`) are different on purpose. one reflects the protocol definition and the other is a consumable key interface. `UNICODE_CHARACTERS` defines all possible ways a user could define a key, e.g. `PageUp`, `Pageup`, `Page_Up` with the same value.

In theory you should not have an issue with these objects being different. Can you explain how you got to the error?

erwinheitzman:
Yeah, using Key.Meta throws an error as it does not exist while it is a valid key

erwinheitzman:
@christian-bromann imo any of the keys listed [here](https://w3c.github.io/webdriver/#keyboard-actions) should be exposed by the Key module as these are all valid keys to be passed into the keys command.

christian-bromann:
> Yeah, using Key.Meta throws an error as it does not exist while it is a valid key

The value of `UNICODE_CHARACTERS.Meta` is exposed as `Key.Command`

erwinheitzman:
It would also simplify the code a lot as we only have one list instead of multiple.

Lastly, I don't understand the need for the Ctrl / WDIO_CONTROL key, we solve this by doing:
```js
import { platform } from 'os'

platform() === 'darwin'
    ? 'Meta'
    : 'Control'
```

christian-bromann:
> Lastly, I don't understand the need for the Ctrl / WDIO_CONTROL key, we solve this by doing:

Control keys are different based on the OS. On Windows `Key.Command` does nothing while on MacOS `Key.Control` has no meaning either. Therefor I introduced a key that works based on the OS you are on.

I suggest to use `Key.Ctrl` for that matter and given this confusing I suggest we have better docs on this. Wdty?

erwinheitzman:
> > Lastly, I don't understand the need for the Ctrl / WDIO_CONTROL key, we solve this by doing:
> 
> Control keys are different based on the OS. On Windows `Key.Command` does nothing while on MacOS `Key.Control` has no meaning either. Therefor I introduced a key that works based on the OS you are on.
> 
> I suggest to use `Key.Ctrl` for that matter and given this confusing I suggest we have better docs on this. Wdyt?

Well that's the thing, according to the docs the Meta keys is what is the Windows/Command key. Both the Meta key and Control key do serve a purpose on MacOS as we currently use them in my project and I have successfully used it in the past as well. Instead I think we can get rid of the Ctrl and Command key as they not valid keys.

erwinheitzman:
But to get back to the focus of this PR, we can keep the additional keys. I have merged the two lists into one which is accessible now.

I noticed another list like this in the devtools package with which we can do the same (simply use the one list).

erwinheitzman:
I do have another suggestion and that is to keep the separate lists but to extend from them. So something like this:
```js
export const Key = {
   ...UNICODE_CHARACTERS,
   Ctrl: 'WDIO_COMMAND'
}
```
I think this achieves what you are looking for while also achieving my goal + the benefit of better maintainability.
What do you think @christian-bromann ?

christian-bromann:
> Well that's the thing, according to the docs the Meta key is what is the Windows/Command key.

That is wrong then. The `Meta` key doesn't work on Windows. For example things like `browser.keys(Key.Meta, 'a')` does not trigger a select all in an input element.

> Both the Meta key and Control key do serve a purpose on MacOS

The same or a different purpose?

> I have merged the two lists into one which is accessible now.

Let me provide some context: before we switched to TypeScript we only had the `UNICODE_CHARACTERS` as a mean to define various of ways someone could enter a function key. Given that there are many Selenium examples out there and folks would confuse tools all the time they would see `ArrowRight` being supported in one tool and `Arrow_Right` in another. The reason why `UNICODE_CHARACTERS` defines multiple function keys in different ways is to accommodate that, so that someone who calls `browser.keys(['Arrow_Right'])` has the same working automation as someone calling `browser.keys(['ArrowRight'])`.

Now with TypeScript being introduced we introduced the `Key` interface to have better type support for keys as well as IDE auto-completion. That `Key` interface is meant to provide a user friendly interface of the keyboard.

Therefor merging these lists is not a good idea in my opinion because we would clutter that interface with duplicate entries (e.g. `PageUp`, `Page Up` and `Page_up` doing the same thing). From the maintenance perspective, having two lists is more code but both lists don't change or need change very often.

I would like to understand how your application uses the `Meta` key? Can you describe it a little bit?

erwinheitzman:
Thank you for all the information Christian, that really helps to understand all of this. I'll gladly try to explain everything as best as I can.

For Windows and Linux the Control (Ctrl on the keyboard) key is used when you want to select all text like so:
```js
await browser.keys(Key.Control + 'a')
```

For MacOS the Meta key (Command on the keyboard) is used to select all text like so:
```js
await browser.keys(Key.Meta + 'a')
```

Now it would be nice to handle both in the same way which can be achieved like so:
```js
import { platform } from 'node:os'

await browser.keys(platform() === 'darwin' ? 'Meta' : 'Control')
```

I think the above shows how the Control key and Meta key work on different operating systems but to quote Apple's docs:
```
On keyboards made for Windows PCs, use the Alt key instead of Option, and the Ctrl key or Windows logo key instead of Command.
```
This might also help to understand it.
___________

Regarding having the two lists, I do see why things are the way they are now and will refactor the code back to how it was. However what is a bit unclear is if we would like to keep supporting values like `Arrow_down` as it's not a valid html value nor a valid w3c value while we also offer the Key module offering one API that everyone can use.
Ideally I would work towards having only one list in the future and get rid of the additional keys but if we want to keep them, that's fine, I just don't see the value after all that we offer nowadays :)

christian-bromann:
> Now it would be nice to handle both in the same way so you can run the tests on any OS, which can be achieved like so:

This is exactly why we have `Key.Ctrl` which handles both, see https://github.com/webdriverio/webdriverio/blob/a429320ede209a9d43a10c602e2226988ceacc16/packages/webdriverio/src/constants.ts#L121-L125

> Ideally I would work towards having only one list in the future and get rid of the additional keys but if we want to keep them, that's fine, I just don't see the value after all that we offer nowadays :)

We could add this to our `v9` list and change the behavior there. Wdyt?

erwinheitzman:
That sounds like a solid plan 👍

erwinheitzman:
I have updated the PR to only add the Meta key to the module 😄

christian-bromann:
For what do you particularly use the `Meta` key? Is the intention of adding it related to the fact that there is no `Command` key in Windows?

erwinheitzman:
The code we have is this:
```js
import { platform } from 'os'

const key = platform() === 'darwin'
    ? 'Meta'
    : 'Control';
    
await browser.keys(key);
```

I have written a migration script to instead use the Key module and after running the script the code looks like this:
```js
import { Key } from 'webdriverio';
import { platform } from 'os'

const key = platform() === 'darwin'
    ? Key.Meta
    : Key.Control;
    
await browser.keys(key);
```

christian-bromann:
> The code we have is this:

This code still remains to work.

> I have written a migration script to instead use the Key module and after running the script the code looks like this:

Why don't you want to use `Key.Ctrl` as described above which does exactly this?

Note that `Command` and `Meta have the same value:

```
    'Command': '\uE03D',
    'Meta': '\uE03D',
```

And `Key.Command` is already exposed through the `Key` interface which brings me back to the question of: Is the intention of adding it related to the fact that there is no Command key in Windows and the usage of `Meta` as name seems more familiar? Again, my intention here is to only expose one property per key value to avoid cluttering the `Key` interface.

erwinheitzman:
As a fellow dev on the project I understand this but as a user I just find it very confusing and we will have document this very well because people will be looking for keys like the Meta key and they will be confused when it is not available.

PR #11636, opened 2023-11-09
(webdriverio): remove type support for browser commands on element in…

christian-bromann:
## Proposed changes

fixes #11582

## Types of changes

[//]: # (What types of changes does your code introduce to WebdriverIO?)
[//]: # (_Put an `x` in the boxes that apply_)

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## Checklist

[//]: # (_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._)

- [x] I have read the [CONTRIBUTING](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md) doc
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] I have added proper type definitions for new commands (if appropriate)

## Further comments

n/a

PR #11679, opened 2023-11-17
deleted BidiHandler and browser commands  from Element type

udarrr:
as a part of the pr https://github.com/webdriverio/webdriverio/pull/11636


udarrr:
Oh it seems there are a lot of references... Are we sure we'd like to exclude browser commands from element? @christian-bromann

udarrr:
> Note: if a core package fails to compile you usually get a ton of other TS errors as well since it won't be able to access types of that core package.

Yes there was just one place where it should be changed. Thank you for explanation!

christian-bromann (approved):
LGTM 👍 

Thanks a lot!

PR #11771, opened 2023-12-01
Fix filtering test specs by part of their name using CLI --spec arg

tech-dm-klymenko:
## Proposed changes

Issue: https://github.com/webdriverio/webdriverio/issues/9495
A bugfix for this functionality has been released:

> If the --spec value does not point to a particular spec file, it is instead used to filter the spec filenames defined in your configuration.
> 
> To run all specs with the word “dialog” in the spec file names, you could use:
> 
> `wdio wdio.conf.js --spec dialog`

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments


packages/wdio-config/src/node/ConfigParser.ts
```diff
@@ -199,6 +199,13 @@ export default class ConfigParser {
             this._config.cucumberFeaturesWithLineNumbers = Array.isArray(this._config.spec) ? [...new Set(this._config.spec)] : [this._config.spec]
         }
 
+        /**
+         * Make glob-pattern from CLI --spec arg if it is only a keyword
+         */
+        if (spec.length > 0 && !this.cliSpecArgContainsPathstoSpecs(spec) && !this.isGlobPattern(spec[0])){
+            spec[0] = this.addGlobPatternToSpecCliArg(spec[0])
```
erwinheitzman, line 206: Is your intention to replace the first character of the string or to add something in front of the string? (assuming it's a string and not an array)

erwinheitzman (commented):
Thank you for the contribution, awesome work!

Could you add some tests to verify the behavior? 👍

packages/wdio-config/src/node/ConfigParser.ts
```diff
@@ -305,6 +312,18 @@ export default class ConfigParser {
             throw new Error('The --multi-run flag requires that either the --spec or --suite flag is also set')
         }
 
+        // remove glob-pattern from CLI spec arg before filtering
+        if (this._config.spec !== undefined && this._config.spec.length > 0 && this.isGlobPattern(this._config.spec[0])){
+            this._config.spec[0] = this.removeGlobPatternFromSpecCliArg(this._config.spec[0])
+        }
```
christian-bromann, line 318: Why do we need to remove the glob pattern? Can we support it instead, e.g. allow `foo*` to match files like `foobar` or `foobarloo`?

packages/wdio-config/src/node/ConfigParser.ts
```diff
@@ -305,6 +312,18 @@ export default class ConfigParser {
...
+
+        // filter specs by keyword
+        if (this._config.spec !== undefined){
+            if (!this.cliSpecArgContainsPathstoSpecs(this._config.spec!)){
+                specs = this.filterSpecsFromConfigFileByKeyword(this._config.spec[0], specs)
+            }
+        }
```
christian-bromann, line 325: Never do nested `if`s , especially when there are not further `else`s. It seems to me that we use `cliSpecArgContainsPathstoSpecs` only once so no need to put this into an extra method, instead:
```suggestion
        // filter specs by keyword
        if (this._config.spec){
            specs = this.filterSpecsFromConfigFileByKeyword(this._config.spec[0], specs)
        }
```

`filterSpecsFromConfigFileByKeyword` should handle cases where `this._config.spec[0]` contains a file path

packages/wdio-config/src/node/ConfigParser.ts
christian-bromann, line 206: Is this really needed? Can we just do this later in `filterSpecsFromConfigFileByKeyword`?

packages/wdio-config/src/node/ConfigParser.ts
```diff
@@ -478,4 +497,35 @@ export default class ConfigParser {
...
     }
-}
+
+    filterSpecsFromConfigFileByKeyword(keyword: string, listOfSpecs: Spec[]
+    ) {
+        return listOfSpecs.filter((specPath) => {
+            if (!Array.isArray(specPath)) {
+                const specName = path.basename(specPath)
+                // check that spec name contains --spec arg from CLI
+                return specName.includes(keyword)
```
christian-bromann, line 507: Can we just glob here too? See above comment

christian-bromann (changes requested):
Thanks for taking a stab at it. I made some comments to hopefully simplify the code. Let me know if you have any questions.

tech-dm-klymenko:
@christian-bromann @erwinheitzman 

Thank you for your comments!

Upon further consideration, I revisited the current approach and made adjustments. From my analysis, the main issues stemmed from two points:

1. The `setFilePathToFilterOptions` method consistently used a glob pattern search. While effective when specifying the spec path to the `--spec`, it failed when we pass a keyword as `--spec` argument.

2. During the initialization stage of the `ConfigParser` object, the `merge` method was invoked within its constructor. This, in turn, attempted to obtain paths to the specs through the `setFilePathToFilterOptions` method. However, at this stage, we lacked a list of specs from the config file, leading to an error.

I fixed issues above and now the --spec option works correctly. I have also included smoke tests to verify this functionality.

Please check my PR one more time

packages/wdio-config/src/node/ConfigParser.ts
```diff
@@ -162,7 +162,7 @@ export default class ConfigParser {
      * merge external object with config object
      * @param  {Object} object  desired object to merge into the config object
      */
-    private merge(object: MergeConfig = {}) {
+    private merge(object: MergeConfig = {}, shouldSetPathToSpecs = true) {
```
erwinheitzman, line 165: I think it might make more sense to rename this to `addPathToSpecs`, what do you think?

packages/wdio-config/src/node/ConfigParser.ts
```diff
@@ -478,4 +478,10 @@ export default class ConfigParser {
         const end = current === total ? undefined : specsPerShard * current
         return specs.slice(current * specsPerShard - specsPerShard, end)
     }
-}
+
+    cliSpecArgContainsPathstoSpecs(specArr: string[]) {
+    // check that CLI --spec arg does not contain path to file
+    // If arg contains '/', '\' or '.' => in this case it is filepath
+        return /[/\\.]/.test(specArr[0])
```
erwinheitzman, line 485: I think we can simplify this, @christian-bromann what do you think about this one?
```suggestion
        return specArr[0].includes(path.sep)
```

erwinheitzman (commented):
Added some feedback

packages/wdio-config/src/node/ConfigParser.ts
tech-dm-klymenko: It seemed to me that in this case the modal verb in the name more clearly conveys that this is a boolean flag. However, I don't see any problem changing the name to the one you suggested

packages/wdio-config/src/node/ConfigParser.ts
christian-bromann: Yes, and I think there is not need for making this a function since we don't use this anywhere else

tests/smoke.runner.js
```diff
@@ -679,6 +679,58 @@ const mochaHooksTestrunner = async () => {
...
+const runSpecsWithFlagAllPassed = async () => {
+    const { skippedSpecs } = await launch(
+        'runSpecsWithFlagAllPassed',
+        path.resolve(__dirname, 'tests-cli-spec-arg', 'wdio-with-all-passed.conf.js'),
+        {
+            autoCompileOpts: { autoCompile: false },
+            spec: ['test']
+        }
+    )
+    assert.strictEqual(skippedSpecs, 0)
```
christian-bromann, line 694: I am not sure if this actually verifies that all files were run

packages/wdio-config/src/node/ConfigParser.ts
tech-dm-klymenko: Method has been updated

tests/smoke.runner.js
tech-dm-klymenko: Yes, I agree. Here it would be good to somehow check the number of specs that were launched to make sure that all files were executed. But so far I have not found such an opportunity

christian-bromann: afaik it is currently only possible to inspect the spec report as done in some other smoke tests

packages/wdio-config/src/node/ConfigParser.ts
```diff
@@ -162,7 +162,7 @@ export default class ConfigParser {
      * merge external object with config object
      * @param  {Object} object  desired object to merge into the config object
      */
-    private merge(object: MergeConfig = {}) {
+    private merge(object: MergeConfig = {}, addPathToSpecs = true) {
```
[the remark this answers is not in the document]
tech-dm-klymenko: Jsdoc has been updated

packages/wdio-config/src/node/ConfigParser.ts
```diff
@@ -338,7 +338,7 @@ export default class ConfigParser {
                         path.dirname(this.#configFilePath)
                     )
                 )
-            } else if (globMatchedFiles.length) {
+            } else if (this.cliSpecArgContainsPathstoSpecs(cliArgFileList) && globMatchedFiles.length) {
```
christian-bromann, line 341: No need to put this into a separated function.
```suggestion
            } else if (cliArgFileList[0].includes(path.sep) && globMatchedFiles.length) {
```

christian-bromann: Also can you elaborate why this is needed? Maybe add a comment?

tech-dm-klymenko: The codebase is new to me. And I explored numerous options in search of a robust solution. It is plausible that, with the recent modifications, the necessity of this condition may have diminished. I'll double check this again. However, its presence definitely does not break anything))

tests/smoke.runner.js
tech-dm-klymenko: Tests have been updated

tech-dm-klymenko:
@christian-bromann @erwinheitzman 
I made changes according to your comments

tech-dm-klymenko:
@christian-bromann 
Could you, please, tell me, do you have any other comments regarding the fix without taking into account the tests? Probably I’ll just add the tests as the separate PR later. And current implementation of tests will be reverted to avoid issues in CI

tests/tests-cli-spec-arg/wdio-with-all-passed.conf.js
```diff
@@ -0,0 +1,28 @@
...
+                './mocha.test02.js',
+                './mocha.test03.js'
+            ],
+            [
+                './mocha.test04.js'
+            ]
+        ],
+        outputDir: outputDirPath
+    }
+)
```
christian-bromann, line 28: Don't use `Object.assign` here because it modifies the `baseConfig` for upcoming test runs
```suggestion
export const config = {
    ...baseConfig,
    /**
        * specify test files
        */
    specs:
    [
        './mocha.test01.js',
        [
            './mocha.test02.js',
            './mocha.test03.js'
        ],
        [
            './mocha.test04.js'
        ]
    ],
    outputDir: outputDirPath
}
```

tests/tests-cli-spec-arg/wdio-with-failed.conf.js
```diff
@@ -0,0 +1,21 @@
+import { config as baseConfig } from './base.wdio.conf.js'
+
+export const config = Object.assign(
```
christian-bromann, line 3: same

christian-bromann (commented):
I think it makes sense to fix the test, it seems like we are very close

tests/tests-cli-spec-arg/wdio-with-all-passed.conf.js
```diff
@@ -0,0 +1,34 @@
...
+        [
+            './mocha.test04.js',
+            [
+                './mocha.test02.js',
+                './mocha.test03.js'
+            ],
+            [
+                './mocha.test01.js'
+            ]
+        ],
```
christian-bromann, line 32: It seems like there is an issue resolving files properly when using the `Launcher` interface instead of the CLI, to workaround this problem, make the following change for paths:

```suggestion
        [
            path.resolve(__dirname, 'mocha.test04.js'),
            [
                path.resolve(__dirname, 'mocha.test02.js'),
                path.resolve(__dirname, 'mocha.test03.js')
            ],
            [
                path.resolve(__dirname, 'mocha.test01.js')
            ]
        ],
```

christian-bromann (commented):
I think there is another issue at hand here which is outside of the scope of your PR, I made a suggestion how to work around it.

christian-bromann (approved):
Awesome, thanks a lot! 👍

christian-bromann:
Fantastic first contribution 🎉

tech-dm-klymenko:
Thank you very much for your appreciation! :blush:

PR #11772, opened 2023-12-01
(webdriverio): add isStable, waitForStable commands

### LINKED ISSUE #3106: click should wait for animation to finish
Modern interfaces use animations heavily to, for example, animate dialogs, slide sidebars in/out etc.

In order to handle this, right now, the suggestion is to write a custom command as per the issue submitted here: https://github.com/webdriverio/webdriverio/issues/429

While animated UI was rare before, now that it is common, it would be a benefit to handle this in webdriverio. This would be similar to the idea that `click` waits for the element to exist and scrolls to it before clicking. The `click` command is a smart function and is not a nieve implementation of just clicking.

When calling `click`, the library could run a test as simple as getting the coordinates of the element, waiting until the next `requestAnimationFrame` and if the new coordinates are the same, we know that the element is not animating. If it's not animating, execute the click command. If it is animating, then we do another `requestAnimationFrame` test. There can be a timeout in the unusual case that the element animates forever.

This works as described in the Cypress test runner, an alternative to WebdriverIO. It works really well.

**Additional context**
n/a

erwinheitzman:
## Proposed changes

Resolves #3106

## Types of changes

- [ ] Documentation update

## Checklist


packages/webdriverio/src/commands/element/isStable.ts
```diff
@@ -0,0 +1,69 @@
...
+ * @return {Boolean} true if element is stable, false if unstable
+ * @type state
+ *
+ */
+export async function isStable (this: WebdriverIO.Element) {
+    const browser = getBrowserObject(this)
+    return await browser.executeAsync((elem, done) => {
+        try {
+            const previousPosition = elem.getBoundingClientRect()
+            requestAnimationFrame(() => {
```
BorisOsipov, line 54: Browsers will never call `window.requestAnimationFrame()` callback for background tab(performance optimisation) and drivers will throw meaningless `script timeout error` after `script timeout` time. So if user calls `isStable()` for element in inactive tab, it will be nice to inform him about it somehow in a friendly way.

You can easily reproduce it through test cases such as

```
browser.url("someurl")
$("some <a target="_blank"></a> element").click()
$("any element").isStable() // script timeout error here
```

You can check that the current window is backgroud by the following condition `document.visibilityState === 'hidden'`

P.S. It would also be nice to have an e2e test for such a case or comment on the code why we check the background tab or not.

BorisOsipov (commented):
See comment.

erwinheitzman:
@BorisOsipov although you are right, I feel like we can already ship it as it is as it seems like a bit of an edgecase to use this on a inactive tab. What do you think?

BorisOsipov:
@erwinheitzman Sure we can. But! It seems easy to add code like this 
```
if(document.visibilityState === 'hidden') {
	throw Error("Hey man, you are using isStable for inactive tab, don't do it dude!")
}
```
and forget the case forever. Instead of waiting to see when someone will come to Discord and ask about it in the next few years.

erwinheitzman:
I thought so but it's actually not that easy as the error doesn't seem to bubble up correctly.

Trying some workarounds for it.

erwinheitzman:
> @erwinheitzman I have also played with this in another tool I am a committer of. My version was
> 
> `browser.execute(script, arguments)`
> 
> ```
> function fn (element) {
>   if (document.visibilityState === 'hidden') {
>     throw Error("Blablabla")
>   }
> 
>   return new Promise((resolve) => {
>     window.requestAnimationFrame(() => {
>       const rect1 = element.getBoundingClientRect();
>       window.requestAnimationFrame(() => {
>         const rect2 = element.getBoundingClientRect();
>         if (rect1.x === rect2.x &&
>           rect1.y === rect2.y &&
>           rect1.width === rect2.width &&
>           rect1.height === rect2.height) {
>           resolve(false);
>         } else {
>           resolve(true);
>         }
>       });
>     });
>   });
> }
> ```
> 
> It also doesn't use modern JS features like `for (const prop in previousPosition)`, and probably work in old browsers either. You may find it useful.

Thanks, I know I know but the problem is that it works but the the waitForStable command does not fail when the isStable command throws so I am looking into why it's not catching the error as I expect it would

BorisOsipov:
Ahh get it.

packages/webdriverio/src/commands/element/waitForStable.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    :waitForStable.js
+    it('should detect when element is stable', async () => {
+        const elem = await $('#elem')
+        await elem.waitForStable({ timeout: 3000 });
+    });
+    it('should detect when element is no longer stable', async () => {
+        const elem = await $('#elem')
+        await elem.waitForStable({ reverse: true });
+    });
```
christian-bromann, line 46:
```suggestion
    it('should wait until element movements have settled', async () => {
        const elem = await $('#elem')

        // this command will throw after 3s as the
        // given example runs a continuous animation
        await elem.waitForStable({ timeout: 3000 })
    })

    it('should wait until the element starts moving', async () => {
        const elem = await $('#elem')
        await elem.waitForStable({ reverse: true })
    })
```

packages/webdriverio/src/commands/element/isStable.ts
```diff
@@ -0,0 +1,72 @@
...
+ * </example>
+ *
+ * @alias element.isStable
+ * @return {Boolean} true if element is stable, false if unstable
+ * @type state
+ *
+ */
+export async function isStable (this: WebdriverIO.Element) {
+    const browser = getBrowserObject(this)
+    return await browser.executeAsync((elem, done) => {
```
christian-bromann, line 51: Let's move the function into an external script in `packages/webdriverio/src/scripts` and give it a name so we don't print the whole function content within the wdio logs

christian-bromann: having these scripts in that directory will help us later on compile them to a different target if needed, e.g. to ensure that we support older browser.

christian-bromann (commented):
Overall looks good to me. Some suggestions on examples and code organisation. I kind of start thinking it would be nice to just have a `waitForElementState` rather than having so many waitFor commands.

packages/webdriverio/src/commands/element/isStable.ts
```diff
@@ -0,0 +1,72 @@
...
+                requestAnimationFrame(() => {
+                    const currentPosition = elem.getBoundingClientRect()
+                    for (const prop in previousPosition) {
+                        if (previousPosition[(prop as keyof DOMRect)] !== currentPosition[(prop as keyof DOMRect)]) {
+                            done(false)
+                        }
+                    }
+                    done(true)
+                })
+            })
```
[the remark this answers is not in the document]
christian-bromann: Gotcha, would you mind adding a comment then with the reason why using two frames? Just for prosperity so we know later one why this was implemented this way.

packages/webdriverio/src/commands/element/isStable.ts
```diff
@@ -0,0 +1,58 @@
...
+ * <example>
+    :index.html
+    <head>
+        <style>
+            #has-animation {
+                width: 200px;
+                height: 200px;
+                background-color: red;
+                animation: 3s 0s alternate slidein;
+            }
```
christian-bromann, line 18:
```suggestion
            div {
                width: 200px;
                height: 200px;
                background-color: red;
            }
            #has-animation {
                animation: 3s 0s alternate slidein;
            }
```

packages/webdriverio/src/commands/element/waitForStable.ts
```diff
@@ -0,0 +1,94 @@
...
+ * <example>
+    :index.html
+    <head>
+        <style>
+            #has-animation {
+                width: 200px;
+                height: 200px;
+                background-color: red;
+                animation: 3s 0s alternate slidein;
+            }
```
christian-bromann, line 20:
```suggestion
            div {
                width: 200px;
                height: 200px;
                background-color: red;
            }
            #has-animation {
                animation: 3s 0s alternate slidein;
            }
```

packages/webdriverio/src/scripts/isElementStable.ts
```diff
@@ -0,0 +1,27 @@
...
+ * @return {void}
+ */
+export default function isElementStable(elem: HTMLElement, done: (returnValue: boolean) => void) {
+    if (document.visibilityState === 'hidden') {
+        throw Error('You are are checking for animations on an inactive tab, animations do not run for inactive tabs')
+    }
+    try {
+        const previousPosition = elem.getBoundingClientRect()
+        requestAnimationFrame(() => {
+            requestAnimationFrame(() => {
```
christian-bromann, line 14: Let's add a comment as to why we wait two frames here.

erwinheitzman:
@christian-bromann there seems to be a difference in CI and my machine in how the quotes are handled but the snapshot is identical to that of similar commands, do you see anything I missed/did wrong?

erwinheitzman:
Finally a green build lol

packages/webdriverio/tests/commands/element/isStable.test.ts
```diff
@@ -0,0 +1,71 @@
...
+            done(false);
+          }
+        }
+        done(true);
+      });
+    });
+  } catch (error) {
+    done(false);
+  }
+}).apply(null, arguments)`)
```
christian-bromann, line 55: No need to assert the whole function
```suggestion
            .toBe(expect.stringContaining(`return (function isElementStable(elem, done)`))
```

christian-bromann (approved):
Great effort 👏

viktor-silakov:
Do I understand correctly, if an element is stable for 5 frames in a row and then starts to change, this method (wait for stable) will not work?

christian-bromann:
@viktor-silakov detecting the "stability" of an element is almost impossible to get right all the times. This command is a first approach but there are plenty of use case where this might not work. If you have an example where this command fails, please provide a reproducible example and we are happy to take a look.

Issue #11834, opened 2023-12-10
[🐛 Bug]: Jasmine - Skipping a test in beforeTest global hook is not preventing the test runner from executing the entire test code


HananArgov:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

8.26.0

### Node.js Version

18.12.1

### Mode

WDIO Testrunner

### Which capabilities are you using?

```typescript
{
      browserName: 'chrome',
      browserVersion: 'stable', // or 'stable', 'beta', 'dev' or 'canary'
      acceptInsecureCerts: true,
    },
```

### What happened?

when adding 
`pending('skip message')`
to the `beforeTest` hook in the config, it will not skip the test, and will continue the test

### What is your expected behavior?

to skip the test.
there was a fix for mocha, i've tried to do the same for Jasmine, but the reject will fail the test.
https://github.com/webdriverio/webdriverio/issues/6645

### How to reproduce the bug.

install wdio with the testrunner, and Jasmine.

in the config file add the following:

```
  beforeTest(test, context) {
    pending('test 123');
  },
```

### Relevant log output

```typescript
getting this from log:

[0-0] 2023-12-10T10:01:44.372Z ERROR @wdio/utils:shim: undefined
```
```


### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

**Comment by HananArgov** [comment]
somehow, i've found the solution (I think)
i'm using the same solution that was used for Mocha, and I add a reason to the reject:

```
            if (/^=> marked Pending/.test(e)) {
                return reject(e);
            }
```

the result is

```
 "spec" Reporter:
------------------------------------------------------------------
[firefox 64 mac #0-0] Running: firefox (v64) on mac
[firefox 64 mac #0-0] Session ID: 4b936dd8-3369-49e8-b002-c6464e5797e7
[firefox 64 mac #0-0]
[firefox 64 mac #0-0] » /e2e-specs/test/specs/example.api.ts
[firefox 64 mac #0-0] example
[firefox 64 mac #0-0]    - skipped test
[firefox 64 mac #0-0]
[firefox 64 mac #0-0]       .........Pending Reasons.........
[firefox 64 mac #0-0]          test 123
[firefox 64 mac #0-0]
[...]

PR #11158, opened 2023-09-13
mocha: skip test in custom hooks

tamil777selvan:
## Proposed changes

Fix - #6645

Preventing the execution of hooks when an exception is thrown to skip a test case in Mocha.

## Types of changes

[//]: # (What types of changes does your code introduce to WebdriverIO?)
[//]: # (_Put an `x` in the boxes that apply_)

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## Checklist

[//]: # (_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._)

- [x] I have read the [CONTRIBUTING](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md) doc
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] I have added proper type definitions for new commands (if appropriate)

## Further comments

[//]: # (If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...)

PR #11835, opened 2023-12-10
fix for issue #11834 - fixing skip test for jasmine on beforeTest hook

HananArgov:
## Proposed changes

This is a fix for issue #11834. 
This will enable skipping tests from `beforeTest` hook in config file. just like https://github.com/webdriverio/webdriverio/pull/11158 did for Mocha

the result is that the test is skipped:
image

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

this is an implementation based on a previous one for Mocha: (https://github.com/webdriverio/webdriverio/pull/11158)

### DIFF
--- a/packages/wdio-utils/src/shim.ts
+++ b/packages/wdio-utils/src/shim.ts
@@ -54,7 +54,7 @@ export async function executeHooksWithArgs<T> (this: any, hookName: string, hook
             * in case of jasmine, when rejecting, we need to pass the message of rejection as well
             */
             if (/^=> marked Pending/.test(e)) {
-                return reject(e);
+                return reject(e)
             }
             log.error(e.stack)
             return resolve(e)
--- a/tests/helpers/jasmine-hooks.conf.js
+++ b/tests/helpers/jasmine-hooks.conf.js
@@ -3,6 +3,6 @@ import { config as baseConfig } from './config.js'
 export const config = Object.assign({}, baseConfig, {
     beforeTest() {
         // this is a global function defined by Jasmine
-        pending()
+        pending('skip test')
     }
 })
--- a/tests/helpers/jasmine-hooks.conf.js
+++ b/tests/helpers/jasmine-hooks.conf.js
@@ -2,6 +2,7 @@ import { config as baseConfig } from './config.js'
 
 export const config = Object.assign({}, baseConfig, {
     beforeTest() {
+        // this is a global function defined by Jasmine
         pending()
     }
 })
--- a/tests/helpers/jasmine-hooks.conf.js
+++ b/tests/helpers/jasmine-hooks.conf.js
@@ -1,0 +1,7 @@
+import { config as baseConfig } from './config.js'
+
+export const config = Object.assign({}, baseConfig, {
+    beforeTest() {
+        pending()
+    }
+})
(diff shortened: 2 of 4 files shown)


packages/wdio-utils/src/shim.ts
```diff
@@ -50,6 +50,12 @@ export async function executeHooksWithArgs<T> (this: any, hookName: string, hook
             if (/^(sync|async) skip; aborting execution$/.test(e.message)) {
                 return reject()
             }
+            /**
+            * in case of jasmine, when rejecting, we need to pass the message of rejection as well
+            */
+            if (/^=> marked Pending/.test(e)) {
+                return reject(e)
```
erwinheitzman, line 57: I think it is good that you add the error here but I think we should do the same for Mocha to keep them in line

erwinheitzman (changes requested):

packages/wdio-utils/src/shim.ts
HananArgov: i'm not sure if this is suppose to be in mocha, i'm working on Jasmine, so i'm not that familiar with how Mocha

HananArgov:
@erwinheitzman 
i've added the tests

tests/helpers/jasmine-hooks.conf.js
```diff
@@ -0,0 +1,7 @@
+import { config as baseConfig } from './config.js'
+
+export const config = Object.assign({}, baseConfig, {
+    beforeTest() {
+        pending()
```
christian-bromann, line 5: Can we add a comment that this is a global function defined by Jasmine?

tests/jasmine/test-skipped-hooks.ts
```diff
@@ -0,0 +1,7 @@
+import assert from 'node:assert'
+
+describe('Jasmine smoke test to be skipped via wdio hooks', () => {
+    it('should be skipped', async () => {
+        assert.strictEqual(1, 1)
```
christian-bromann, line 5: How about we let this assertion fail as we expect this test never to be executed.
```suggestion
        assert.strictEqual(1, 2)
```

tests/helpers/jasmine-hooks.conf.js
HananArgov:
```suggestion
        // this is a global function defined by Jasmine
        pending()
```

tests/jasmine/test-skipped-hooks.ts
HananArgov: done

tests/helpers/jasmine-hooks.conf.js
HananArgov: done

christian-bromann:
It seems like the added smoke test fails now 🤔

HananArgov:
> It seems like the added smoke test fails now 🤔

so what do we do? should I return the value back?

christian-bromann:
@HananArgov it seems like in the smoke test we execute a file (`/tests/jasmine/test-skipped-hooks.ts`) that suppose to be skipped right?

HananArgov:
@christian-bromann i've re-written the test, so it will be suitable to a Jasmine test (used `jasmineFailSpecWithNoExpectations` as a template)

PR #11900, opened 2023-12-18
(@wdio/cli): improve TypeScript setup process

erwinheitzman:
## Proposed changes

Improve the CLI process when setting up TypeScript

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/packages/wdio-cli/src/commands/config.ts
+++ b/packages/wdio-cli/src/commands/config.ts
@@ -97,7 +97,7 @@ export const parseAnswers = async function (yes: boolean): Promise<ParsedAnswers
     const parsedPaths = getPathForFileGeneration(answers, projectRootDir)
     const isUsingTypeScript = answers.isUsingCompiler === CompilerOptions.TS
     const wdioConfigFilename = `wdio.conf.${isUsingTypeScript ? 'ts' : 'js'}`
-    const wdioConfigPath = path.resolve(projectRootDir, wdioConfigFilename)
+    const wdioConfigPath = path.resolve(answers.specs ? path.dirname(answers.specs.split(path.sep).filter((s) => !s.includes('*')).join(path.sep)) : projectRootDir, wdioConfigFilename)
 
     return {
         projectName: projectProps?.packageJson.name || 'Test Suite',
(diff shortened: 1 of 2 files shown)


packages/wdio-cli/src/commands/config.ts
```diff
@@ -97,7 +97,7 @@ export const parseAnswers = async function (yes: boolean): Promise<ParsedAnswers
     const parsedPaths = getPathForFileGeneration(answers, projectRootDir)
     const isUsingTypeScript = answers.isUsingCompiler === CompilerOptions.TS
     const wdioConfigFilename = `wdio.conf.${isUsingTypeScript ? 'ts' : 'js'}`
-    const wdioConfigPath = path.resolve(projectRootDir, wdioConfigFilename)
+    const wdioConfigPath = path.resolve(answers.specs ? path.dirname(answers.specs.split(path.sep).filter((s) => !s.includes('*')).join(path.sep)) : projectRootDir, wdioConfigFilename)
```
christian-bromann, line 100:
```suggestion
    const wdioConfigPath = path.resolve(
        answers.specs
            ? path.dirname(answers.specs.split(path.sep).filter((s) => !s.includes('*')).join(path.sep))
            : projectRootDir,
        wdioConfigFilename
    )
```

christian-bromann (changes requested):
I am not sure if using `path.dirname(answers.specs.split(path.sep).filter((s) => !s.includes('*')).join(path.sep))` is a reliable way to determine the path. What happens if someone puts: `./tests/**/e2e/*.js` which is a viable glob pattern. With the current implementation it would be fomatted to `./tests/e2e` which I think is not what we want.

erwinheitzman:
Currently there is no other directory to determine where they want their files right? The root is the root of the project which we don't want to use, so does this mean we need to place a directory in between the root and test directories?

christian-bromann:
So if I understand correctly , we want to create a separate `tsconfig.json` file for WebdriverIO test files. I think it is not straight forward to find a good place to put that file given peoples setups are so different. When someone configures a new WebdriverIO setup we can assume that either:

- it is a complete new project and no files have been created, e.g. `npm init wdio@latest ./idontexist`
- it is an existing project and a `tsconfig.json` file likely exists

When I added TypeScript support to the configuration wizard I covered the first scenario given it was easy to just create a `tsconfig.json` that enables someone to use TypeScript and call it a day.

The second scenario is much more complex since we don't know nor can make any assumptions about the users setup. I don't think it is a good idea to make assumption and e.g. "just" put a `tsconfig.json` in `/test` when we know that there is a chance someone can already have a `tsconfig.json` file there. I think the safest bet is to create a `./.wdio` directory and store wdio and tsconfig files. However this seems like a larger effort.

christian-bromann:
I honestly think the best approach is to just say: if there is a `tsconfig.json` , don't create one. WebdriverIO should not depend on any specific TypeScript settings so it shouldn't be difficult to adjust existing settings. So my personal opinion is: keep it up to the user, maybe extend the existing docs to guide user better.

erwinheitzman:
Well we can also create a tsconfig.wdio.json instead and update the script to run the tests with that, what do you think?

christian-bromann:
> what do you think?

Sounds like a good compromise.

erwinheitzman:
We can still inform the user that they should update their setup to include the types for their IDE but other than that, this would mean we don't have to change much, the logic stays pretty much the same and there is no conflicts (if the user already has a file called tsconfig.wdio.json we do nothing and inform the user as well)

christian-bromann:
@erwinheitzman any updates?

christian-bromann:
I will go ahead and close this as there doesn't seem to be interest in moving this forward. Feel free to re-open if I am mistaken.

PR #11910, opened 2023-12-20
fix(webdriver): respect strictSSL setting for bidi connections

jlipps:
WebDriver BiDi connections were not made with sensitivity to the `strictSSL` WebDriverIO option. This PR attempts to fix that. Tested locally on a self-signed cert. Not really sure how to add a test in this repo, nor do I really want to figure that out 😉


christian-bromann (approved):
LGTM 👍

PR #11915, opened 2023-12-23
(docs): simplify setup using npx

erwinheitzman:
## Proposed changes

I dived into some other create-xxx packages and noticed they use npx, after reading the npm docs I found out that npx always uses the latest version and simplifies things for us as we can remove the `@latest` from our examples and it also explicitly states what package we use which I like.

I also simplified the following cases:
`npx create-wdio .` and `npx create-wdio ./` have become `npx create-wdio`
`npx create-wdio -- --yes` has become `npx create-wdio --yes`
`npx create-wdio ./path/to/project -- --yes` has become `npx create-wdio ./path/to/project --yes`

EDIT: I have also update the yarn, pnpm and bun command examples. For pnpm there's a similar dlx command which acts like npx https://pnpm.io/cli/dlx showing an example of a create package.

## Types of changes

- [x] Documentation update

## Checklist


wswebcreation (commented):
I like the simplification, but......., can we explain a bit more what the default behaviour is, especially for the paths?

For example

```sh
npm init wdio@latest .
```

The above makes it "more clear" to me that it's being installed in the current path

```sh
npx create-wdio
```

is not telling me anything about a path.

Reason for also mentioning this is that you, me, and maybe other people, didn't know that this was the "default" behaviour. We need to keep thinking that (most of) the WDIO users know the complete Node.js ecosystem.

Love to hear your thoughts

erwinheitzman:
> I like the simplification, but......., can we explain a bit more what the default behaviour is, especially for the paths?
> 
> For example
> 
> ```shell
> npm init wdio@latest .
> ```
> 
> The above makes it "more clear" to me that it's being installed in the current path
> 
> ```shell
> npx create-wdio
> ```
> 
> is not telling me anything about a path.
> 
> Reason for also mentioning this is that you, me, and maybe other people, didn't know that this was the "default" behaviour. We need to keep thinking that (most of) the WDIO users know the complete Node.js ecosystem.
> 
> Love to hear your thoughts

We can certainly mention that it will install it in the current directory or the directory that is given. I don't think this is more confusing than `npx create-wdio .` as people that are new to the terminal won't know what the dot stands for anyways, therefore I think clear documentation and an intuitive user interface/experience is more important than being explicit here.
It also clears confusion about questions like: Is `npx create-wdio .` the same as `npx create-wdio ./`?

website/docs/ComponentTesting.md
```diff
@@ -21,9 +21,9 @@ WebdriverIO has built-in support for the following frameworks:
 To set-up WebdriverIO for unit or component testing in the browser, initiate a new WebdriverIO project via:
 
 ```bash
-npm init wdio@latest ./
```
christian-bromann, line 24: Is `npm init wdio@latest ./` still working though? We have now a lot of material out there where we advocate for this command and we should not break these.

christian-bromann: Furthermore, we own the `wdio` NPM package and can use it to shorten this to `npx wdio`, wdyt?

website/i18n/it/docusaurus-plugin-content-docs/current/GettingStarted.md
```diff
@@ -34,20 +34,20 @@ If you're in the root directory of an existing project, run:
...
 ```sh
-npm init wdio ./path/to/new/project
+npx create-wdio ./path/to/new/project
 ```
 
 </TabItem>
 <TabItem value="yarn">
 
 ```sh
-yarn create wdio .
```
christian-bromann, line 50: Why do we remove the path parameter here?

website/i18n/ru/docusaurus-plugin-content-docs/current/desktop-testing/Electron.md
```diff
@@ -19,7 +19,7 @@ You just need a few simple steps to get started.
 To initiate a new WebdriverIO project, run:
 
 ```sh
-npm create wdio@latest ./
+npm create create-wdio
```
christian-bromann, line 22: seems like duplicate `create`

christian-bromann (commented):
When it comes to

```sh
npm init wdio@latest ./
```
vs.
```sh
npx create-wdio
```

I would honestly prefer keeping it as `npm init wdio ./` or `npm create wdio ./`. Looking at many other projects this seems to be the preferred way to go.

Issue #11711, opened 2023-11-22
[🐛 Bug]: Arrow functions and `.contains()` crash IE11 via `isElementDisplayed()`


colinrotherham:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

8.23.4

### Node.js Version

20.9.0

### Mode

Standalone Mode

### Which capabilities are you using?

```typescript
{
  browserName: 'internet explorer',
  browserVersion: '11.285',
  platformName: 'Windows 10'
}
```

### What happened?

Following on from https://github.com/webdriverio/webdriverio/issues/11705 I've got a few more IE11 issues

TypeScript uses `"target": "es2022"` for **./packages/webdriverio/src/scripts** so these aren't downleveled

## Arrow functions

Internet Explorer 11 has no [arrow function support](https://caniuse.com/arrow-functions)

https://github.com/webdriverio/webdriverio/blob/59d349ca847950354d02b9e548f60cc50e7871f0/packages/webdriverio/src/scripts/isElementDisplayed.ts#L134-L136

## Method `.contains()`

Internet Explorer 11 has [`.contains()` support](https://caniuse.com/mdn-api_node_contains) only for `HTMLElement` (not `Document` or other node types)

https://github.com/webdriverio/webdriverio/blob/59d349ca847950354d02b9e548f60cc50e7871f0/packages/webdriverio/src/scripts/isElementDisplayed.ts#L205-L208

### What is your expected behavior?

WebdriverIO v8 should still support IE11

### How to reproduce the bug.

See https://github.com/webdriverio/webdriverio/issues/11705 for the same steps to reproduce

### Relevant log output

```typescript
See https://github.com/webdriverio/webdriverio/issues/11705 for log output
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

PR #11990, opened 2024-01-10
Webdriverio: Fix regression in 11711

WillBrock:
#11711 was to fix issues with older browsers but also added `body` to a check. It looks like this was added by mistake.

Our tests started failing with this change. I don't have a specific reproducible example because it fails on a custom dropdown that we've built in our codebase.  

- [ ] Documentation update

## Checklist

### DIFF
--- a/packages/webdriverio/src/scripts/isElementDisplayed.ts
+++ b/packages/webdriverio/src/scripts/isElementDisplayed.ts
@@ -210,7 +210,7 @@ export default function isElementDisplayed (element: Element): boolean {
     // This is a partial reimplementation of Selenium's "element is displayed" algorithm.
     // When the W3C specification's algorithm stabilizes, we should implement that.
     // If this command is misdirected to the wrong document (and is NOT inside a shadow root), treat it as not shown.
-    if (!isElementInsideShadowRoot(element) && !document.body.contains(element)) {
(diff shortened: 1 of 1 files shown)


christian-bromann (approved):
LGTM 👍

christian-bromann:
> Our tests started failing with this change.

What was the error? Technically I believe that the difference between `document.contains` and `document.body.contains` is that with the first you can also check if a tag is existent in e.g. in `<head>` section of the page. For testing in the context of WebdriverIO I don't think it makes a difference between using one over the other. Still, it would be good to better understand the problem here.

WillBrock:
It's not seeing this Filter input box as being displayed here. 

The error:
```Error: element (".swift-box-option-filter-input") still not displayed after 20000ms```

I didn't dig too deep into the difference between the `document.body` and just `document` since I didn't see any mention of it in the original PR and doesn't seem like that part of the code would be affected with an old browser.

Screenshot 2024-01-10 at 6 38 37 PM

christian-bromann:
Thanks for the context. It seems like `document.contains` is a non standard interface according to this [MDN](https://developer.mozilla.org/pt-BR/docs/Web/API/Document) translation artifact. Interestingly there are no docs in the english version about this 🤷

PR #11992, opened 2024-01-11
feat(wdio-prefix): introduce `wdio:{maxInstances,specs,exclude}` to WebdriverIO.Capabilities

vobu:
### feature
allow the capability config to contain a W3C-compliant `wdio` prefix for `maxInstances`,`specs`, and `exclude`:

```ts
const _config: Testrunner = {
    ...baseConfig,
    capabilities: [
        {
            browserName: 'safari',
            'wdio:maxInstances': 2,
            'wdio:specs': ['prefix-test-01.ts'],
            'wdio:exclude': ['prefix-test-02.ts']
        }
}
```

### type fix
When extending the `WebdriverIO.Capabilites` interface, the property `maxInstances` was missing. This resulted in an error when specifying `maxInstances` by using the extended type.

```typescript
// in some.types.ts
export interface wdi5Config extends WebdriverIO.Config {
      capabilities: wdi5Capabilities[] | wdi5MultiRemoteCapability
}
export interface wdi5Capabilities extends WebdriverIO.Capabilities {
    //...
}

// in wdio.conf.ts
export const config: wdi5Config = {
    // ...
    capabilities: [
            {
                maxInstances: 2, // <-- Error: Object literal may only specify known properties, and 'maxInstances' does not exist in type 'wdi5Capabilities'.ts(2353)
                browserName: "chrome",
                // ...
            }
    ]
}
```
## Types of changes

- [ ] Documentation update

## Checklist


packages/wdio-types/src/Capabilities.ts
```diff
@@ -1130,10 +1134,10 @@ export interface BrowserStackCapabilities {
...
     debug?: boolean
     networkLogs?: boolean
-     /**
-     * https://www.browserstack.com/docs/app-automate/appium/debug-failed-tests/network-logs
-     * Enable viewing the response data in the Network Logs tab on your session
-     */
+    /**
+    * https://www.browserstack.com/docs/app-automate/appium/debug-failed-tests/network-logs
+    * Enable viewing the response data in the Network Logs tab on your session
+    */
```
christian-bromann, line 1140:
```suggestion
     * https://www.browserstack.com/docs/app-automate/appium/debug-failed-tests/network-logs
     * Enable viewing the response data in the Network Logs tab on your session
     */
```

packages/wdio-types/src/Capabilities.ts
```diff
@@ -77,15 +77,19 @@ declare global {
...
             unhandledPromptBehavior?: string;
             /**
              * WebDriver clients opt in to a bidirectional connection by requesting a capability with the name "webSocketUrl" and value true.
              */
-            webSocketUrl?: boolean
+            webSocketUrl?: boolean;
+            /**
+             * Maximum number of total parallel running workers.
+             */
+            maxInstances?: number;
```
christian-bromann, line 92: Can we create a custom interface for WebdriverIO options in capabilities, have this interface extend from it and keep it as is since we only have true w3c capabilities here.

christian-bromann (commented):
Thanks for the proposal, this makes me think of whether we should treat these "special" webdriverio capabilities more W3C conform. Maybe we should prefix them with `wdio:XXX`.

@vobu can you do that in the new interface? Afaik the following should be recognised: `maxInstances`, `specs`, `exclude`.

packages/wdio-types/src/Capabilities.ts
vobu: right on, I'll give it a shot.

packages/wdio-types/src/Capabilities.ts
```diff
@@ -39,9 +39,16 @@ export interface ProxyObject {
     noProxy?: string[];
 }
 
+export interface WdioOptions {
+    /**
+     * Maximum number of total parallel running workers (per capability)
+     */
+    maxInstances?: number;
+}
```
christian-bromann, line 47: Instead of redefining, better do this:
```suggestion
interface WDIOCapabilityOptions extends Pick<TestrunnerOptions, 'specs' | 'exclude' | 'maxInstances'> {}
```

packages/wdio-types/src/Capabilities.ts
```diff
@@ -77,15 +84,15 @@ declare global {
             /**
              * Defines the current session’s strict file interactability.
              */
-            strictFileInteractability?: boolean,
+            strictFileInteractability?: boolean;
```
christian-bromann, line 87: Let's not use `;` they are not needed (not only here but also in other parts of the file.

vobu:
you're too fast @christian-bromann :)
let me work on some more details, then I'll ping when done and ready for re-review

packages/wdio-types/src/Capabilities.ts
vobu: done

packages/wdio-types/src/Capabilities.ts
vobu: `,` → `;`?
this was done by the eslint-prettier setup and autoformat, not by me intentionally...

vobu:
ok, now there should both be the `pick`ed type definitions:
- `wdio`-prefixed options 
- plain options marked `deprecated`

plus a test extending options per capability.

→ ping @christian-bromann :)

packages/wdio-types/src/Capabilities.ts
christian-bromann: can we revert this?

packages/wdio-types/src/Capabilities.ts
```diff
@@ -39,9 +40,45 @@ export interface ProxyObject {
     noProxy?: string[];
 }
 
+export interface CapabilityOptions extends Pick<TestrunnerOptions, 'specs' | 'exclude' | 'maxInstances'> {
```
christian-bromann, line 43: We don't need this if we define them in the interface with new docs.
```suggestion
export interface CapabilityOptions {
```

christian-bromann (commented):
I am 💯 for using prefixed wdio caps, however we currently don't recognise them and need to make enhancements to the code.

packages/wdio-types/src/Capabilities.ts
christian-bromann: yes, we should have a linter for this but that can be done in a different PR

vobu: done: all `;` removed

packages/wdio-types/src/Capabilities.ts
vobu: done

christian-bromann:
> however we currently don't recognise them and need to make enhancements to the code.

The code in question is:
- [596057](https://github.com/webdriverio/webdriverio/blob/d7e4b20dbaf354831eecd7a2298e79af94993207/packages/wdio-config/src/node/ConfigParser.ts#L171-L179)
- https://github.com/webdriverio/webdriverio/blob/c6db464b4b87b7e80b66eda4e3260b7be43b61a6/packages/wdio-cli/src/launcher.ts#L232

christian-bromann:
> Or where do you see the `wdio:specs` and `wdio:exclude` being used?

No, this is exactly how I imagined. Let me know if you have further questions.

vobu:
> > Or where do you see the `wdio:specs` and `wdio:exclude` being used?
> 
> No, this is exactly how I imagined. Let me know if you have further questions.

right on, then I'm on the right track 😄 
Further q's: 
- should the `wdio:specs` and `wdio:exclude` mix in with `config.specs` and `config.exclude` or overwrite?
- if so, which one takes precedence (if `wdio:exclude` excludes `$test-x`, but `config.specs` lists it)? 
  I'd naïvely assume "specific first", so `config.capabilty.wdio:{specs,exclude,maxInstances}` > `config.{specs,exclude,maxInstances}`

christian-bromann:
> * should the `wdio:specs` and `wdio:exclude` mix in with `config.specs` and `config.exclude` or overwrite?

`wdio:specs` and `wdio:exclude` will take precedence and overwrite global configurations.

> * how should the case be handled when multiple `capabilities` list different files? Say:

In this case the first capability only runs `path.resolve('test-03.ts')` and the second only runs `path.resolve('test-05.ts')`

> I'd assume that this should result in "2 instances of Chrome run test-03.ts, but not test-04.ts. And neither test-05.ts nor test-06.ts".

No, this should initiate 1 Chrome and 1 Safari session. A `wdio:spec` defined in a capability does not effect other capabilities. Note that all this functionality is already there. You already define `spec` and `exclude` in the capabilities and it should work. What is needed is that we check for `spec` as well as `wdio:spec` capabilities.

vobu:
hmmm, can't really do sth about the  ❌ e2e-mac ... might be worth a re-trigger (which I can't do due to repo perms)

vobu:
and in terms of docs: wasn't sure whether to put any (yet) - if I should, maybe to here: https://webdriver.io/docs/capabilities#webdriverio-capabilities-to-manage-browser-driver-options ?

packages/wdio-config/src/node/ConfigParser.ts
```diff
@@ -31,7 +31,9 @@ interface TestrunnerOptionsWithParameters extends Omit<Options.Testrunner, 'capa
 
 interface MergeConfig extends Omit<Partial<TestrunnerOptionsWithParameters>, 'specs' | 'exclude'> {
     specs?: Spec[]
+    'wdio:specs'?: Spec[]
     exclude?: string[]
+    'wdio:exclude'?: string[]
```
christian-bromann, line 36: Can we revert changes in this file? These are options and not capabilities. We don't need to support prefixed options as we don't have to adhere to a standardized interface

packages/wdio-types/src/Capabilities.ts
```diff
@@ -5,83 +5,119 @@ import type {
...
+    sslProxyPort?: number
+    socksProxy?: string
+    socksProxyPort?: number
+    socksVersion?: string
+    socksUsername?: string
+    socksPassword?: string
+    noProxy?: string[]
+}
+
+export interface CapabilityOptions {
```
christian-bromann, line 42: Can we remove this interface and merge everything into [`WebdriverIOCapabilities`](https://github.com/webdriverio/webdriverio/blob/0652d3c0212256108d3f505f98b95d97a5b5f591/packages/wdio-types/src/Capabilities.ts#L213)?

christian-bromann (commented):
This looks great, I added some comments. Let's also document these caps in [`/website/docs/Capabilities.md`](https://github.com/webdriverio/webdriverio/blob/63ef7b125b33657f99558a7177c4d504f4cdb032/website/docs/Capabilities.md#L37)

christian-bromann:
> if I should, maybe to here:

Yes, that looks like a good place!

packages/wdio-config/src/node/ConfigParser.ts
vobu: kk, done

packages/wdio-types/src/Capabilities.ts
vobu: kk, done

vobu:
added the [documentation](https://github.com/webdriverio/webdriverio/pull/11992/files#diff-f4f99dfb0ab78eb0d2ced19921eab3dd9d5337e4f98de51c2c6b83db41edf5baR64). 
yet CI fails due to a build error not specific for this commit (i hope 🤣 ).

packages/wdio-cli/src/launcher.ts
```diff
@@ -271,7 +272,9 @@ class Launcher {
      * Format the specs into an array of objects with files and retries
      */
     private _formatSpecs(capabilities: (Capabilities.DesiredCapabilities | Capabilities.W3CCapabilities | Capabilities.RemoteCapabilities), specFileRetries: number) {
-        const files = this.configParser.getSpecs((capabilities as Capabilities.DesiredCapabilities).specs, (capabilities as Capabilities.DesiredCapabilities).exclude)
+        const specs = (capabilities as Capabilities.DesiredCapabilities).specs || (capabilities as CapabilityOptions)['wdio:specs']
+        const excludes = (capabilities as Capabilities.DesiredCapabilities).exclude || (capabilities as CapabilityOptions)['wdio:exclude']
```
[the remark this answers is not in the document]
christian-bromann: I will create a ticket to simplify this in v9 but until then we have to deal with these options

packages/wdio-config/src/node/ConfigParser.ts
```diff
@@ -169,12 +169,17 @@ export default class ConfigParser {
         this._config = deepmerge(this._config, object) as TestrunnerOptionsWithParameters
 
         /**
-         * overwrite config specs that got piped into the wdio command
+         * overwrite config specs that got piped into the wdio command,
+         * also adhering to the wdio-prefixes from a capability
          */
-        if (object.specs && object.specs.length > 0) {
+        if (object['wdio:specs'] && object['wdio:specs'].length > 0) {
```
christian-bromann, line 175: This is why the build is failing, `object` are not capabilities here. We should revert these changes.

website/docs/Capabilities.md
```diff
@@ -61,6 +61,24 @@ If the driver `binary` is set, WebdriverIO won't attempt to download a driver bu
 
 :::
 
+#### `wdio:maxInstances`
```
christian-bromann, line 64: I think this is a bit misleading as these caps are not related to driver options, maybe the section could look like this:

```md
### WebdriverIO Capabilities to manage browser driver options

WebdriverIO defines a set of own capabilities to manage its automated session:

#### `wdio:chromedriverOptions`

Specific options passed into Chromedriver when starting it.

#### `wdio:geckodriverOptions`

Specific options passed into Geckodriver when starting it.

#### `wdio:edgedriverOptions`

Specific options passed into Edgedriver when starting it.

#### `wdio:safaridriverOptions`

Specific options passed into Safari when starting it.

#### `wdio:maxInstances`

Maximum number of total parallel running workers for the specific browser/capability. Takes precedence over [maxInstances](#configuration#maxInstances) and [maxInstancesPerCapability](configuration/#maxinstancespercapability).

Type: `number`

#### `wdio:specs`

Define specs for test execution for that browser/capability. Same as the [regular `specs` configuration option](configuration#specs), but specific to the browser/capability. Takes precedence over `specs`.

Type: `(String | String[])[]`

#### `wdio:exclude`

Exclude specs from test execution for that browser/capability. Same as the [regular `exclude` configuration option](configuration#exclude), but specific to the browser/capability. Takes precedence over `exclude`.

Type: `String[]`

WebdriverIO manages installing and running browser driver for you. WebdriverIO uses a custom capability that allows you to pass in parameters to the driver.

#### Common Driver Options
...
```

packages/wdio-cli/src/launcher.ts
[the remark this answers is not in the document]
vobu: again thanks for the explanation.
made the change you suggested, plus a small type cleanup in the `.map`

website/docs/Capabilities.md
vobu: kk, moved and expanded the section as you suggested.

packages/wdio-config/src/node/ConfigParser.ts
vobu: hmmm, but if the check for `wdio:{specs,exclude}` is removed, we don't get the merged config and the tests (e.g.) `https://github.com/webdriverio/webdriverio/pull/11992/files#diff-e81dd1eb85b53f80e2f3860c68a8712a3b1bc2fa846f47909facc4637baf17d1R616` fail.

Above is of course directly connected to the `MergeConfig` type def:
```ts
interface MergeConfig extends Omit<Partial<TestrunnerOptionsWithParameters>, 'specs' | 'exclude'> {
    specs?: Spec[]
    'wdio:specs'?: Spec[]
    exclude?: string[]
    'wdio:exclude'?: string[]
}
```
What to do? 😄

website/docs/Capabilities.md
```diff
@@ -38,6 +38,43 @@ While the amount of fixed defined capabilities is very low, everyone can provide
...
+
+#### `wdio:exclude`
+
+Exclude specs from test execution for that browser/capability. Same as the [regular `exclude` configuration option](configuration#exclude), but specific to the browser/capability. Takes precedence over `exclude`.
+
+Type: `String[]`
+
+WebdriverIO manages installing and running browser driver for you. WebdriverIO uses a custom capability that allows you to pass in parameters to the driver.
+
+
```
christian-bromann, line 77:
```suggestion
```

packages/wdio-config/src/node/ConfigParser.ts
christian-bromann: Oh ok, thanks for the call out. Let's leave it it as is then.

christian-bromann (approved):
One minor thing that I will manually merge. The rest LGTM 👍 

Thanks a lot!

PR #12042, opened 2024-01-18
(@wdio/browser-runner): retry on dynamic import error

christian-bromann:
tbd


christian-bromann:
Seems to not be solvable through retrying multiple times. I skipped the test in CI and raised https://github.com/webdriverio/webdriverio/issues/12056

Issue #11958, opened 2024-01-02
[🐛 Bug]: <w3c wheel actions sometimes can't perform deltaX,deltaY movement>


udarrr:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

latest

### Node.js Version

20

### Mode

WDIO Testrunner

### What happened?

I've tried to add correct calculation to scrollIntoView method (it's been done successfully) but I've encountered some of unexpected behavior it seems wheel actions sometimes can't perform movement with deltaX and deltaY as with random deltaX,deltaY value as with correct. It's been scrolling as it would be deltaX,deltaY = 0

here is reproducible an example https://github.com/udarrr/actionsDeltaBug 

npm run wdio_bug

### What is your expected behavior?

I can suggest the fix 

https://github.com/udarrr/actionsDeltaBug 

npm run wdio_fix

There are 4 actions

1. Executed script for calculations deltaX,deltaY (script is rewritten version of [compute-scroll-into-view](https://github.com/scroll-into-view/compute-scroll-into-view) calculations)
2. Executed wheel actions with dexltaX,deltaY
3. Executed script for calculation again 
4. Checked was it scrolled into view or not, 
      - if it's been scrolled, fallback to web scrollinToView
      - if it hasn't been scrolled, skip executions web scrollToView

### How to reproduce the bug.

https://github.com/udarrr/actionsDeltaBug

### Relevant log output

```typescript
-
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

christian-bromann:
Thanks for filing the issue @udarrr 

Do you experience the bug in all browser or just on particular. Chances are that the command is executed incorrectly by the browser driver and an issue needs to be reported in the WebDriver protocol.

udarrr:
To be honest i've encountered only with chrome the latest. But let me check the rest browsers

udarrr:
The same for 

chrome
edge

and in firefox it's not working at all with origin, i mean actions can't perform scroll automatically as with delta as without it just with origin error is "Move target (542, 3647) is out of bounds of viewport dimensions (1280, 744)"

christian-bromann:
[...]

Issue #11866, opened 2023-12-13
[🐛 Bug]: element.scrollIntoView() throws "Request failed with status 500 due to move target out of bounds: move target out of bounds"


spencerlavallesonos:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

latest

### Node.js Version

18.17.1

### Mode

WDIO Testrunner

### Which capabilities are you using?

```typescript
N/A, not configured from initial wdio wizard.
```

### What happened?

When using `element.scrollIntoView()` the error/ warning "Request failed with status 500 due to move target out of bounds: move target out of bounds" is thrown even if the `scrollIntoView()` is executed successfully

### What is your expected behavior?

If a move target is in boundary (is currently on the screen and visible), scrolling to it shouldn't throw an error that the target is out of bounds. 
In this example Sonos.com's main "Shop" button on the top navigation-bar. 

### How to reproduce the bug.

Clone [this repo](https://github.com/spencerlavallesonos/wdio-bug-scrollIntoView), follow the readme, see the output. 

### Relevant log output

```typescript
npm run wdio

> wdio
> wdio run ./wdio.conf.js


Execution of 1 workers started at 2023-12-13T23:48:54.651Z

[0-0] RUNNING in chrome - file:///test/specs/wdiobug.test.js
[0-0] Will trigger "Request failed with status 500"
[0-0] 2023-12-13T23:49:09.006Z ERROR webdriver: Request failed with status 500 due to move target out of bounds: move target out of bounds
[0-0]   (Session info: chrome=120.0.6099.71)
[0-0] PASSED in chrome - file:///test/specs/wdiobug.test.js

 "spec" Reporter:
------------------------------------------------------------------
[chrome 120.0.6099.71 mac #0-0] Running: chrome (v120.0.6099.71) on mac
[chrome 120.0.6099.71 mac #0-0] Session ID: 6a5c6510f6bd7785d1caf445b04e4656
[chrome 120.0.6099.71 mac #0-0]
[chrome 120.0.6099.71 mac #0-0] » file:///test/specs/wdiobug.test.js
[chrome 120.0.6099.71 mac #0-0] WDIO Bug
[chrome 120.0.6099.71 mac #0-0]    ✓ element.scrollIntoView() still sending status 500
[...]

PR #12057, opened 2024-01-20
scroll calculations feature

udarrr:
related to https://github.com/webdriverio/webdriverio/issues/11958, https://github.com/webdriverio/webdriverio/issues/11866


e2e/wdio/headless/test.e2e.ts
```diff
@@ -250,11 +250,16 @@ describe('main suite 1', () => {
...
 
                 await browser.execute((elem, _params) => elem.scrollIntoView(_params), searchInput, input as any)
-                const nativeX = await browser.execute(() => window.scrollX)
+                let nativeX = await browser.execute(() => window.scrollX)
 
+                if (Math.floor(wdioX) > Math.floor(nativeX)){
+                    nativeX = nativeX + 1
+                } else if (Math.floor(wdioX) < Math.floor(nativeX)) {
+                    wdioX = wdioX + 1
+                }
```
christian-bromann, line 262: Can you explain this?

packages/webdriverio/src/commands/element/scrollIntoView.ts
```diff
@@ -43,74 +44,59 @@ function scrollIntoViewWeb (
  */
 export async function scrollIntoView (
     this: WebdriverIO.Element,
-    options: ScrollIntoViewOptions | boolean = { block: 'start', inline: 'nearest' }
+    options: ScrollIntoViewOptions & {scrollMode?: 'if-needed' | 'always'} | boolean = { block: 'start', inline: 'nearest', scrollMode: 'if-needed' }
```
christian-bromann, line 47: Let's not mingle standard options with custom modifications. If we introduce a scrollMode we should add a 2nd parameter

packages/webdriverio/src/commands/element/scrollIntoView.ts
```diff
@@ -43,74 +44,59 @@ function scrollIntoViewWeb (
...
+            top: number;
+            left: number;
+            isVisible?: boolean
         }
-
-        let [deltaX, deltaY] = [deltaByOption.start.x, deltaByOption.start.y]
-        if (options === true) {
-            options = { block: 'start', inline: 'nearest' }
+        // eslint-disable-next-line unicorn/prefer-ternary
+        if (deltaPosition && 'isVisible' in deltaPosition && deltaPosition.isVisible && parsedOptions.scrollMode !== 'if-needed') {
```
christian-bromann, line 82: Can we add a comment why we do a execute scroll into view here?

christian-bromann (changes requested):
Can you describe this PR a little bit? What are you trying to resolve? What is the approach?

e2e/wdio/headless/test.e2e.ts
udarrr: deltaX or delataY can be only int not float and we're round it to one side or opposite, whereas web scrollIntoView can work with float that's why there is a difference in a pixel that shoud be calculated for the test

packages/webdriverio/src/commands/element/scrollIntoView.ts
udarrr: Ok but scrollIntoView in wdio isn't webScrollIntoView in general, because it's having 'actions' behind of scenes, and scrollMode is optional here, i believe it won't break anything. please let me know if you're still insist to move it to second argument

udarrr:
> Can you describe this PR a little bit? What are you trying to resolve? What is the approach?

Yes sure. Generally it's correct calculation for deltaX and deltaY, previously when an error 
 'Request failed with status 500 due to move target out of bounds: move target out of bounds' appeared that meant 'action' method has moved scrolling to 'end' even with wrong calculation, then throwed the error, and then called web scrollIntoView. It doesn't have a big sense to call 'actions' if for delta is responsible just scrollIntoView in current implementation, the issue has been described here 
https://github.com/webdriverio/webdriverio/issues/11866, you can reproduce it here https://github.com/udarrr/moveToExample.

e2e/wdio/headless/test.e2e.ts
christian-bromann: I don't understand, why are we adding `+1` here? Can you outline in a concrete example?

packages/webdriverio/src/commands/element/scrollIntoView.ts
christian-bromann: for reference, there seems to be a proposal out there for this: https://github.com/w3c/csswg-drafts/pull/5677

packages/webdriverio/src/scripts/scrollIntoViewDeltaCalcScript.ts
```diff
@@ -0,0 +1,287 @@
+/**
+ * adopted to w3c delta
+ * original source: https://github.com/scroll-into-view/compute-scroll-into-view
+ * Copyright by https://github.com/stipsan
+ */
```
christian-bromann, line 5: I think if we end up using this, we should add `compute-scroll-into-view` as dependency as we should avoid to have maintain this ourselves

packages/webdriverio/src/commands/element/scrollIntoView.ts
```diff
@@ -43,74 +44,59 @@ function scrollIntoViewWeb (
...
+            await scrollIntoViewWeb.call(this, options as any)
```
christian-bromann, line 83: Is there a chance we can do this using the action command?

packages/webdriverio/src/commands/element/scrollIntoView.ts
christian-bromann: I am not a big fan introducing this option for two reasons:

1. it is not (yet) part of the CSS spec
2. I don't see the value for users to have the ability to choose between `if-needed` and `always`

WebdriverIO users using this command mostly want to make sure that an element is within the viewport so they can properly interact with that. We should choose our defaults wisely to ensure this is what the command does at the minimal basis.

packages/webdriverio/src/scripts/scrollIntoViewDeltaCalcScript.ts
udarrr: It's not possible, the dependency has been rewritten totally and adopted for obtaining delta, i even couldn't have mentioned original. Original library provide coordinates for full scroll path, but in case of actions are necessary only deltaX,Y because actions is moving a scroll to end by default.

packages/webdriverio/src/commands/element/scrollIntoView.ts
udarrr: The main benefit 'if-needed' is we don't trigger element to position without necessity of it. For some reason the method can be a part of general actions or checks, and when we're always triggering it can lead to various artifacts because Element can be had animation, locked or something like that. But ok it's new and we should be wary i'll make it like in ScrollIntoViewOptions by default.

packages/webdriverio/src/commands/element/scrollIntoView.ts
udarrr: Oh... these are big calculations... is there a chance to preserve it like this ?)

packages/webdriverio/src/commands/element/scrollIntoView.ts
udarrr: done

packages/webdriverio/src/commands/element/scrollIntoView.ts
christian-bromann: I don't understand 🤔 `scrollIntoViewWeb` runs an `execute` command to run `scrollIntoView` within the browser. What I am asking is if we could use the action command instead.

udarrr: let me clarify
all deltas are moving scrolling regarding 'end'

- when element isn't into view 'actions' behave as 'end' and scrolling it to that, then we can use delta regarding 'end' that to scroll it to 'center' or 'start'
- when element is into view 'actions' behave as 'nearest' and don't trigger scrolling, in the case calculations are different we should determine current position of element regarding possible 'end' then calculate delta

udarrr:
Possible it makes sense to left it just like

```
await browser.action('wheel').scroll({ duration: 0, deltaX: 0, deltaY: 0, origin: this, }).perform()
```
without calculation at all, by default it's handling 'end' and 'nearest'

And for 'center' and 'start' we can use 'action' for scrolling to 'end' and 'scrollIntoViewWeb' for shifting
```
await browser.action('wheel').scroll({ duration: 0, deltaX: 0, deltaY: 0, origin: this, }).perform() 
await scrollIntoViewWeb.call(this, options as any)
```
thay's it like it's working now just without annoyed an error

What's your take on this @christian-bromann  ?

erwinheitzman:
I'm not sure what others think but looking at the code I feel like we are adding a lot of complexity, are we sure this is really needed?

christian-bromann:
@udarrr 

I am also not 100% convinced about this one. Our goal should it be to provide the best result using WebDriver primitives for scrolling. It is fine to me that we do element calculations through execute but the actual scroll should happen through WebDriver. Happy to document the escape hatch to do this via `await browser.execute((elem) => elem.scrollIntoView(), elem)` if an error occurs, or have this escape hatch within a catch block.

udarrr:
thank you for pointing it out, as we're going to rely on webdriver action native only, there should be fixed an issue with delta. I've raised an issue here https://github.com/w3c/webdriver/issues/1798 after clarification i'll be able to proceed with the PR.

christian-bromann:
Sounds good to me. Thanks for clarifying and raising the issue. I will go ahead and close this. Feel free to re-raise any time.

Issue #12094, opened 2024-01-26
[🐛 Bug]: Appium service is not starting against local emulator using latest versions.


lararojasmr:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

latest

### Node.js Version

latest

### Mode

WDIO Testrunner

### Which capabilities are you using?

```typescript
{
  "platformName": "Android",
  "port": 4723,
  "appium:automationName": "UiAutomator2",
  "appium:deviceName": "Android",
  "appium:appPackage": "com.cimobileapp",
  "appium:appActivity": ".MainActivity"
}
```

### What happened?

The problem is that the test runner start but the appium service is stuck in some place that i'm not able to log. I was seeing the @christian-bromann changes in the latest versions and I saw that there was a changes in this part related to the port, etc. Also, this spawn process that not have a timeout so the wdio could be hanging out for hours, becuase the failure occur in the OnPrepare in the appium service, so the wdio timeout is not working in that level.

I have my app running in local, using android studio, i tried removing the port, the hostname, adding it to the capabilities, configuration. 

Some log: 

`2024-01-25T23:59:39.345Z INFO @wdio/appium-service: Will spawn Appium process: node /Users/mlara/intive/sony/qa-automation/node_modules/appium/index.js --base-path / --allow-insecure get_server_logs
2024-01-26T00:01:59.928Z INFO @wdio/local-runner: Shutting down spawned worker
2024-01-26T00:01:59.948Z ERROR @wdio/cli:utils: A service failed in the 'onPrepare' hook
Error: Appium exited before timeout (exit code: 0)
dbug
    at ChildProcess.<anonymous> (file:///Users/mlara/intive/sony/qa-automation/packages/e2e-ci-automation/node_modules/@wdio/appium-service/build/launcher.js:151:24)
    at Object.onceWrapper (node:events:634:26)
    at ChildProcess.emit (node:events:519:28)
    at ChildProcess.emit (node:domain:488:12)
    at Process.ChildProcess._handle.onexit (node:internal/child_process:294:12)

Continue...
2024-01-26T00:01:59.948Z DEBUG @wdio/cli:utils: Finished to run "onPrepare" hook in 140609ms
2024-01-26T00:01:59.951Z INFO @wdio/cli:launcher: Run onComplete hook
[...]

PR #12095, opened 2024-01-26
#12094 Bugfix/appium start function

lararojasmr:
## Proposed changes
I've added a timeout for the function and fix the issue that happened when the promise emit data but is not include the information that is in the conditional, that could be a error.

I added a rejectonce method to ensure that the reject is called once.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments


lararojasmr:
@christian-bromann please check this changes.

christian-bromann (commented):
Thanks for the contribution! The code looks good. Could we add a unit test for this to prevent regression?

lararojasmr:
@christian-bromann think that the current test cases should be good in this case. Do you have any other ideas on this?. I tell you this because i only add a new considerations but the logic is the same.

packages/wdio-appium-service/src/launcher.ts
```diff
@@ -166,25 +170,54 @@ export default class AppiumLauncher implements Services.ServiceInstance {
                     log.info(`Appium started with ID: ${process.pid}`)
                     resolve(process)
                 }
+                // if the data does not include the above string, it is probably an error or debugger message.
+                log.warn(data.toString())
+                resolve('The appium service is starting... Check logs above for possible issues.')
```
christian-bromann, line 175: We resolve the promise with a string here which would cause an error further up where a child process is expected.

christian-bromann (commented):
One additional comment.

christian-bromann:
> Do you have any other ideas on this?

There is the timeout aspect that is new and should be tested. I think it is important to capture the essence of this improvement in a test otherwise someone will change it in the future and might break this behavior.

packages/wdio-appium-service/src/launcher.ts
lararojasmr: Interesting, so we can resolve with the same promise on this case so the appium process finished well but without expect message. 

A quick change if you are agree.

packages/wdio-appium-service/src/launcher.ts
```diff
@@ -154,37 +154,70 @@ export default class AppiumLauncher implements Services.ServiceInstance {
...
     }
-
-    private _startAppium(command: string, args: Array<string>) {
+    private _startAppium(command: string, args: Array<string>, timeout:number = 60000) {
         log.info(`Will spawn Appium process: ${command} ${args.join(' ')}`)
         const process: ChildProcessByStdio<null, Readable, Readable> = spawn(command, args, { stdio: ['ignore', 'pipe', 'pipe'] })
+        // just for validate the first error
+        let errorCaptured = false
+        // to set a timeout for the promise
+        let timeoutId:number
```
christian-bromann, line 163:
```suggestion
        let timeoutId: NodeJS.Timeout
```

packages/wdio-appium-service/src/launcher.ts
```diff
@@ -154,37 +154,70 @@ export default class AppiumLauncher implements Services.ServiceInstance {
...
 
             /**
              * only capture first error to print it in case Appium failed to start.
              */
-            process.stderr.once('data', (err) => { error = err })
+            process.stderr.once('data', (err) => {
+                error = (err || new Error('Appium exited without unknown error message')) as Error
+                log.error(error.toString());
+                rejectOnce(error)
+            })
```
christian-bromann, line 185:
```suggestion
            process.stderr.once('data', (err: Error) => {
                error = err || new Error('Appium exited without unknown error message')
                log.error(error.toString());
                rejectOnce(error)
            })
```

packages/wdio-appium-service/src/launcher.ts
```diff
@@ -154,37 +154,70 @@ export default class AppiumLauncher implements Services.ServiceInstance {
             this._process.kill()
         }
     }
-
-    private _startAppium(command: string, args: Array<string>) {
+    private _startAppium(command: string, args: Array<string>, timeout:number = 60000) {
```
christian-bromann, line 157:
```suggestion
    private _startAppium(command: string, args: Array<string>, timeout = 60000) {
```

christian-bromann: Can we move this into a `const` variable to the top of the file, e.g. `const APPIUM_START_TIMEOUT = 10 * 1000`

Should 10s be sufficient here?

christian-bromann (commented):
Some last comments/suggestions.

packages/wdio-appium-service/src/launcher.ts
lararojasmr: > Can we move this into a `const` variable to the top of the file, e.g. `const APPIUM_START_TIMEOUT = 10 * 1000`
> 
> Should 10s be sufficient here?

i suggest to add 30 sec. It's not a option that we can change now, so we need to be flexible. I agree with you with the const idea.

christian-bromann (approved):
LGTM, thanks 👍

lararojasmr:
> LGTM, thanks 👍

@christian-bromann I saw some TC failing. Anything that I can do? - I've just update the branch.

lararojasmr:
@christian-bromann I sent some fixes :) all test cases are passing now. Sorry hehe

christian-bromann (approved):
LGTM 👍  thanks!

picakia:
Hey, I recently updated from 8.29.1 to 8.29.3 and it looks like this change introduced a bug with logging that generates countless "INFO @wdio/appium-service: Appium started with ID: 496446" messages as can be seen in this screenshot:
image

Reverting back to 8.29.1 resolves the issue

lararojasmr:
Hey, let me check this now. Thanks for reporting it.On 31 Jan 2024, at 09:50, Filip ***@***.***> wrote:﻿
Hey, I recently updated from 8.29.1 to 8.29.3 and it looks like this change introduced a bug with logging that generates countless "INFO @wdio/appium-service: Appium started with ID: 496446" messages as can be seen in this screenshot:
image.png (view on web)
Reverting back to 8.29.1 resolves the issue

christian-bromann:
@lararojasmr feel free to just raise a PR, no need to create an extra issue for it if you have a solution already. Thanks for looking into it.

lararojasmr:
@christian-bromann sorry to late I created a new Issue haha. But I've added the fixes in this PR ☝🏼

PR #12130, opened 2024-01-30
comma can't be typed by keys sequence without extra pause in Electron apps

udarrr:
Bug is browser.keys('something with c,o,m,m,a,s') type it to input as 'something with commas' without commas but have to 'something with c,o,m,m,a,s' in Electron apps

### DIFF
--- a/packages/webdriverio/src/commands/browser/keys.ts
+++ b/packages/webdriverio/src/commands/browser/keys.ts
@@ -57,7 +57,7 @@ export async function keys (
     keySequence.forEach((value) => {
         if (!this.isIOS && value === ','){
             /**
-             * comma can't be printed within sequence without extra pause here more then 200ms
+             * comma can't be typed within sequence without extra pause here more than 200ms
              */
             keyAction.pause(225)
         }
--- a/packages/webdriverio/src/commands/browser/keys.ts
+++ b/packages/webdriverio/src/commands/browser/keys.ts
@@ -54,7 +54,15 @@ export async function keys (
      * W3C way of handle it key actions
      */
     const keyAction = this.action('key')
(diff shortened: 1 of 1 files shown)


packages/webdriverio/src/commands/browser/keys.ts
```diff
@@ -54,7 +54,15 @@ export async function keys (
      * W3C way of handle it key actions
      */
     const keyAction = this.action('key')
-    keySequence.forEach((value) => keyAction.down(value))
+    keySequence.forEach((value) => {
+        if (!this.isIOS && value === ','){
```
christian-bromann, line 58: How is this related to iOS? Should we check for Safari in general?

christian-bromann (commented):
Let's add an e2e test for this, maybe somewhere here: `e2e/wdio/headless/test.e2e.ts`

packages/webdriverio/src/commands/browser/keys.ts
```diff
@@ -54,7 +54,15 @@ export async function keys (
...
+            /**
+             * comma can't be typed within sequence without extra pause here more than 200ms
+             */
+            keyAction.pause(225)
```
christian-bromann, line 62: Can we make this a `const` and move the variable declaration to the top of the file?

christian-bromann (commented):
@wswebcreation what do you think?

udarrr:
> Let's add an e2e test for this, maybe somewhere here: `e2e/wdio/headless/test.e2e.ts`

@christian-bromann it seems the bug relevant only for electron apps, i wasn't able reproduce it in browsers. Any ideas how should it be determined only for electron ?

PR #12206, opened 2024-02-09
(feat): enable scroll as optional in isClickable script

lacell75:
## Proposed changes

Currently, a scroll is perform in isClickable API if the element is not displayed. This is very disruptive in the case where you want to test that the element is not clickable. (a scroll is done wihout asked)
This PR allows to disable the scroll if necessary.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments
N/A

### DIFF
--- a/packages/webdriverio/src/commands/element/isClickable.ts
+++ b/packages/webdriverio/src/commands/element/isClickable.ts
@@ -3,15 +3,18 @@ import { ELEMENT_KEY } from 'webdriver'
 import { getBrowserObject } from '../../utils/index.js'
 import isElementClickableScript from '../../scripts/isElementClickable.js'
 
+import type { ClickableOptions } from '../../types.js'
 /**
  *
  * Return true if the selected DOM-element:
  *
  * - exists
  * - is visible
- * - is within viewport (if not try scroll to it)
+ * - is within viewport
  * - its center is not overlapped with another element
  * - is not disabled
+ *
+ *By default try to scroll it if not within viewport (can set to false in option)
  *
  * otherwise return false.
  *
@@ -30,18 +33,23 @@ import isElementClickableScript from '../../scripts/isElementClickable.js'
         let clickable = await el.isClickable();
         console.log(clickable); // outputs: true or false
 
+        // is clickable without perform the scroll
+        let clickable = await el.isClickable({ scroll: false });
+
         // wait for element to be clickable
-        await browser.waitUntil(() => el.isClickable())
+        await browser.waitUntil(() => el.isClickable());
     });
  * </example>
  *
  * @alias element.isClickable
  * @return {Boolean}            true if element is clickable
  * @uses protocol/selectorExecute, protocol/timeoutsAsyncScript
+ * @param {ClickableOptions=} options           isClickable command options
+ * @param {scroll=}             options.scroll    try scrolling to element if it's not clickable (default: true)
  * @type state
  *
  */
-export async function isClickable (this: WebdriverIO.Element) {
+export async function isClickable (this: WebdriverIO.Element, { scroll = true }: ClickableOptions = {}) {
     if (!await this.isDisplayed()) {
         return false
     }
@@ -58,5 +66,5 @@ export async function isClickable (this: WebdriverIO.Element) {
     return browser.execute(isElementClickableScript, {
         [ELEMENT_KEY]: this.elementId, // w3c compatible
         ELEMENT: this.elementId // jsonwp compatible
-    } as any as HTMLElement)
+    } as any as HTMLElement, scroll)
 }
--- a/packages/webdriverio/src/commands/element/waitForClickable.ts
+++ b/packages/webdriverio/src/commands/element/waitForClickable.ts
@@ -1,5 +1,7 @@
 import type { WaitForOptions } from '../../types.js'
 
+import type { ClickableOptions } from '../../types.js'
+
 /**
  * Wait for an element for the provided amount of milliseconds to be clickable or not clickable.
  *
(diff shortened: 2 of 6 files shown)


erwinheitzman:
Hi @lacell75 Thank you for your contribution!

I like the idea but would like to suggest the idea of adding a `inViewport` property which checks if the element is clickable within the viewport or outside of the viewport depending on it's value.
I think it is more descriptive than `scroll`, and it aligns with our plans for doing something similar with the isDisplayed/waitForDisplayed command.

What do you think?

christian-bromann:
Could you also provide a reproducible example where scrolling would not be desired?

lacell75:
> Could you also provide a reproducible example where scrolling would not be desired?

```
  it('Navigation in bar menu', async () => {
    await browser.url('https://www.google.com/search?q=webdriver+selenium');
    await browser.setWindowSize(750, 750);
    await $('[aria-modal="true"] button:nth-of-type(2)').click();
    await expect($('[role="navigation"] [data-id="trc"]')).toBeDisplayed();
    await expect($('span=Error')).not.toBeDisplayedInViewport(); // test pass
    await expect($('span=Error')).toBeClickable();
    await expect($('span=Error')).not.toBeDisplayedInViewport(); // Test failed and Result is different from first test (isClickable changes the behavior)
    await $('[data-id="trc"] button:nth-of-type(2)').click(); // Click on right button to display "Error"
    await expect($('span=Error')).toBeDisplayedInViewport(); // test pass 

        // The goal of my MR is to deactivate automatic scrolling because it distorts the following checks. With(next PR on expect-webdriverIo) :  await expect($('span=Error')).toBeClickable( {scroll : false });
  });
```

lacell75:
> Hi @lacell75 Thank you for your contribution!
> 
> I like the idea but would like to suggest the idea of adding a `inViewport` property which checks if the element is clickable within the viewport or outside of the viewport depending on it's value. I think it is more descriptive than `scroll`, and it aligns with our plans for doing something similar with the isDisplayed/waitForDisplayed command.
> 
> What do you think?

Yes, I can do it but, how I can meets my need which is to deactivate the automatically scroll action in isClickable script? 
I think we need to have 2 properties in this case, WDYT? Currently, we scroll to the element if it's not clickable in all cases

christian-bromann:
> it('Navigation in bar menu', async () => {

@lacell75 I am failing to reproduce due to this error: 

> Can't call click on element with selector "[aria-modal="true"] button:nth-of-type(2)" because element wasn't found

lacell75:
> > it('Navigation in bar menu', async () => {
> 
> @lacell75 I am failing to reproduce due to this error:
> 
> > Can't call click on element with selector "[aria-modal="true"] button:nth-of-type(2)" because element wasn't found
You can remove this click. This modal was displayed because, I launched the browser in private mode.

christian-bromann:
@lacell75 the test now fails in this line `await expect($('span=Error')).toBeClickable();` because the element is not existing.

lacell75:
Sorry, the datas displayed is moving on google. Please see the new example below, I hope it will be good
```
  await browser.url('https://www.youtube.com');
  await browser.setWindowSize(750, 750);
  try {
    await $('ytd-consent-bump-v2-lightbox #dialog').waitForDisplayed({ timeout: 10000 });
    await $('ytd-consent-bump-v2-lightbox ytd-button-renderer:nth-of-type(2) yt-button-shape').click(); // accept cookies
  } catch (e) {
    // No COOKIE dialog
  }
  await expect($('#scroll-container')).toBeDisplayed();
  await expect($('yt-chip-cloud-chip-renderer:nth-of-type(10)')).not.toBeDisplayedInViewport(); // test pass
  await expect($('yt-chip-cloud-chip-renderer:nth-of-type(10)')).toBeClickable();
  await expect($('yt-chip-cloud-chip-renderer:nth-of-type(10)')).not.toBeDisplayedInViewport(); // Test failed and Result is different from first test (isClickable changes the behavior)
  await $('#right-arrow-button').click(); // Click on right button to display chip expected
  await expect($('yt-chip-cloud-chip-renderer:nth-of-type(10)')).toBeDisplayedInViewport(); // test pass

```

christian-bromann:
I see, this is makes it more clear. I feel like it is a very specific use case as it only clashes when combined with an `isWithinViewport` check. I feel like the better approach to solve this problem is to revert the scroll operation by scrolling back to the original position. What do you think?

lacell75:
Yes, I think it's a good idea.
How I can do it? I remove the scroll option added by me then I add a
scrolling back to original position in this
script packages/webdriverio/src/scripts/isElementClickable.ts after the
scroll for this specific case?
Is that right?

christian-bromann:
> I remove the scroll option added by me then I add a scrolling back to original position in this

Sounds reasonable, yes! I think we should "just" work by negating the numbers, right?

christian-bromann:
> what do you mean by negating the numbers?

Call the same scroll command with the parameters negated.

lacell75:
> revert the scroll operation by scrolling back to the original position

Currently, this is a scrollIntoView that is performed . https://github.com/webdriverio/webdriverio/blob/bff063e5c285c8806257e31abfa18638838cfe4e/packages/webdriverio/src/scripts/isElementClickable.ts#L133
I can't call this scrollIntoView command with the parameters negated to go back to original position , isn't ?

christian-bromann:
> I can't call this scrollIntoView command with the parameters negated to go back to original position , isn't ?

No, you are right. I guess we have to determine the scroll position then and after to reverse the operation. Or have `scrollintoview` return the previous scroll position since it is detemining that in the command.

lacell75:
I didn't find a solution to perform the scroll back. I think I'm going to give up.

erwinheitzman:
@lacell75 I did some investigation and wanted to simplify the setup a bit and reduce the checks done alongside the change so I hope you don't mind that I created this followup #12431 https://github.com/webdriverio/webdriverio/pull/12431.

I can merge this to v8 and edit the release notes to mention you instead of me as you did some of the work already, would that be okay with you?

lacell75:
> @lacell75 I did some investigation and wanted to simplify the setup a bit and reduce the checks done alongside these changes so I hope you don't mind that I created this followup #12431. This PR includes scrolling back to it's original position.
> 
Well done!
> I can merge this to v8 and edit the release notes to mention you instead of me as you did some of the work already, would that be okay with you?

Sure, I'm ok. Thanks you.

christian-bromann:
Let's close in favor of #12431

PR #12210, opened 2024-02-12
fix: WDIO Cucumber Test Tags

sauravdas1997:
## Proposed changes

Currently Browserstack accessibility is not running when tags are passed in accessibilityOptions in case of cucumber, so we are handling the use case for test tags in cucumber.

## Types of changes

- [ ] Documentation update

## Checklist

## Further comments

### DIFF
--- a/packages/wdio-browserstack-service/src/util.ts
+++ b/packages/wdio-browserstack-service/src/util.ts
@@ -343,10 +343,19 @@ export const validateCapsWithA11y = (deviceName?: any, platformMeta?: { [key: st
     return false
 }
 
-export const shouldScanTestForAccessibility = (suiteTitle: string | undefined, testTitle: string, accessibilityOptions?: { [key: string]: any; }) => {
+export const shouldScanTestForAccessibility = (suiteTitle: string | undefined, testTitle: string, accessibilityOptions?: { [key: string]: any; }, world?: { [key: string]: any; }, isCucumber?: boolean ) => {
     try {
         const includeTags = Array.isArray(accessibilityOptions?.includeTagsInTestingScope) ? accessibilityOptions?.includeTagsInTestingScope : []
         const excludeTags = Array.isArray(accessibilityOptions?.excludeTagsInTestingScope) ? accessibilityOptions?.excludeTagsInTestingScope : []
+
+        if (isCucumber) {
+            const tagsList: string[] = []
+            world?.pickle?.tags.map((tag: { [key: string]: any; }) => tagsList.push(tag.name))
+            const excluded = excludeTags?.some((exclude) => tagsList.includes(exclude))
+            const included = includeTags?.length === 0 || includeTags?.some((include) => tagsList.includes(include))
+
+            return !excluded && included
+        }
 
         const fullTestName = suiteTitle + ' ' + testTitle
         const excluded = excludeTags?.some((exclude) => fullTestName.includes(exclude))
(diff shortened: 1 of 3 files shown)


erwinheitzman:
Hi @sauravdas1997 thank you for your contribution!

I have some requests:
1. file an issue that can be coupled to this PR so that we can first verify and discuss if this is an issue
2. describe the changes made
3. add some tests to verify the changed behaviour

Thank you

erwinheitzman:
Hi @sauravdas1997 thank you for your contribution!

This does indeed need to be `name`, however the value also shouldn't be an array, could you make this change and update/add tests were needed? Thank you.

packages/wdio-browserstack-service/tests/util.test.ts
```diff
@@ -1026,13 +1026,30 @@ describe('validateCapsWithA11y', () => {
...
+                }
+            ]
+        }
+    }
     it('returns true if full test name contains includeTags', async () => {
         expect(shouldScanTestForAccessibility('suite title', 'test title', { includeTagsInTestingScope: 'title' })).toEqual(true)
     })
 
-    it('returns false if full test name contains excludeTags', async () => {
+    it.only('returns false if full test name contains excludeTags', async () => {
```
nextlevelbeard, line 1042: Don't forget to remove this `.only`

sauravdas1997: done

sauravdas1997:
> file an issue that can be coupled to this PR so that we can first verify and discuss if this is an issue
This is a fix for a browserstack specific offering that we came across during our internal testing.
> describe the changes made
Added in the PR description
> add some tests to verify the changed behaviour
Added in PR

packages/wdio-browserstack-service/src/util.ts
```diff
@@ -343,11 +343,20 @@ export const validateCapsWithA11y = (deviceName?: any, platformMeta?: { [key: st
...
+export const shouldScanTestForAccessibility = (suiteTitle: string | undefined, testTitle: string, accessibilityOptions?: { [key: string]: any; }, world?: { [key: string]: any; }, isCucumber?: boolean ) => {
     try {
         const includeTags = Array.isArray(accessibilityOptions?.includeTagsInTestingScope) ? accessibilityOptions?.includeTagsInTestingScope : []
         const excludeTags = Array.isArray(accessibilityOptions?.excludeTagsInTestingScope) ? accessibilityOptions?.excludeTagsInTestingScope : []
 
+        if (isCucumber) {
+            const tagsList: string[] = []
+            world?.pickle?.tags.map((tag: { [key: string]: any; }) => tagsList.push(tag.name))
+            const excluded = excludeTags?.some((exclude) => tagsList.includes(exclude))
+            const included = includeTags?.length === 0 || includeTags?.some((include) => tagsList.includes(include))
```
erwinheitzman, line 355: Could you explain why we do an explicit check here `includeTags?.length === 0`? I feel like the include and exclude arrays should be treated the same way so it's a surprise to see this.

erwinheitzman (commented):
Got one question as a comment, looks got overall!

packages/wdio-browserstack-service/src/util.ts
sauravdas1997: Use case: If the user has not specified any includeTags, we should return the shouldScanTestForAccessibility results as true. However, if the user has mentioned a value in includeTags, we should evaluate against it. If the user does not provide excludeTags, no tags should be excluded by default. We will only honour it if the user explicitly specifies tags in excludeTags.

christian-bromann (approved):
LGTM 👍 

@erwinheitzman any objections to merge?

christian-bromann:
I will go ahead and merge. Happy to address remaining issues in a new PR.

Issue #12173, opened 2024-02-05
[🐛 Bug]: getSize() typing expects at least 1 argument


therealbrad:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

latest

### Node.js Version

latest

### Mode

WDIO Testrunner

### What happened?

When using getSize() in TypeScript I get a compile error:
`Expected 1 arguments, but got 0.`

### What is your expected behavior?

[getSize()](https://webdriver.io/docs/api/element/getSize) argument is optional.

### How to reproduce the bug.

https://github.com/therealbrad/wdio-getsize shows the bug
Note the test passes, but TypeScript complains:
```Expected 1 arguments, but got 0.
Property 'height' does not exist on type 'number'.
Property 'width' does not exist on type 'number'.```

### Relevant log output

```typescript
Expected 1 arguments, but got 0.
Property 'height' does not exist on type 'number'.
Property 'width' does not exist on type 'number'.
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

christian-bromann:
Hey @therealbrad , thanks for raising the issue.

It seems like we are loosing the optional information when trying to map the types to the prototype: https://github.com/webdriverio/webdriverio/blob/0cba687a8df2464ca970b26c2bab9cd78e5bdccc/packages/webdriverio/src/types.ts#L20-L24

```ts
type ChainablePrototype = {
    [K in ElementQueryCommands]: (...args: Parameters<$ElementCommands[K]>) => ChainablePromiseElement<ThenArg<ReturnType<$ElementCommands[K]>>>
} & {
    [K in ElementsQueryCommands]: (...args: Parameters<$ElementCommands[K]>) => ChainablePromiseArray<ThenArg<ReturnType<$ElementCommands[K]>>>
}
```

justalemming:
same issue with [getLocation()](https://webdriver.io/docs/api/element/getLocation/). Here the docs also don't specify the `prop` as optional.

PR #12307, opened 2024-02-18
[V8] Updated props to optional for getText and getLocation commands

jemishgopani:
## Proposed changes
Fixes [#12173](https://github.com/webdriverio/webdriverio/issues/12173)
* Make props as optional parameter for getText and getLocation commands

## Types of changes
- [ ] Documentation update

## Checklist


christian-bromann:
Thanks for the contribution! Can you raise the same PR for `main` branch as well?

jemishgopani:
> Thanks for the contribution! Can you raise the same PR for `main` branch as well?

Sure

packages/webdriverio/src/commands/element/getLocation.ts
```diff
@@ -36,7 +36,7 @@ export function getLocation (this: WebdriverIO.Element, prop: keyof Location): P
 export async function getLocation (
     this: WebdriverIO.Element,
     prop?: keyof Location
-): Promise<Location | number> {
+): Promise<any> {
```
christian-bromann, line 39: I don't see the need to make this `any`, what is the reasoning for this change?

```suggestion
): Promise<Location | number> {
```

packages/webdriverio/src/commands/element/getSize.ts
```diff
@@ -37,7 +37,7 @@ export function getSize (this: WebdriverIO.Element, prop: keyof RectReturn): Pro
 export async function getSize (
     this: WebdriverIO.Element,
     prop?: keyof RectReturn
-): Promise<Size | number> {
+): Promise<any> {
```
christian-bromann, line 40: same

packages/webdriverio/src/commands/element/getLocation.ts
jemishgopani: if here we use return type as `Location | number` then the declaration refrence is not working as expected for `getLocation()` it will ask for an 1 argument in the warning because reference for IDE always point to `export function getSize (this: WebdriverIO.Element, prop: keyof RectReturn): Promise<number>` declaration.

packages/webdriverio/src/commands/element/getSize.ts
jemishgopani: Same answer as above

packages/webdriverio/src/commands/element/getSize.ts
```diff
@@ -4,8 +4,8 @@ import { getElementRect } from '../../utils/index.js'
 
 export type Size = Pick<RectReturn, 'width' | 'height'>;
 
-export function getSize (this: WebdriverIO.Element): Promise<Size>;
 export function getSize (this: WebdriverIO.Element, prop: keyof RectReturn): Promise<number>;
+export function getSize (this: WebdriverIO.Element): Promise<Size>;
```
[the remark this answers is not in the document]
jemishgopani: yes I did this accidentally I'll revert this.

packages/webdriverio/src/commands/element/getLocation.ts
christian-bromann: I am not sure if this is the right solution then. We shouldn't return `any` as we loose type information about the return value. I might be wrong here though, do you have any reference documentation that recommends this approach?

jemishgopani: Yup you are correct that is not right, I have updated the code with a different solution

christian-bromann (approved):
Awesome, LGTM 👍

Issue #9926, opened 2023-03-10
[💡 Feature]: Proposal: Allow running the same sped file multiple times in parallel 


WillBrock:
### Is your feature request related to a problem?

I'd like to be able to run the same test multiple times in parallel. The reason for this is to prevent flaky tests from getting introduced into the code. The way that I would use it is that when a new branch gets submitted it will run through all the regular tests like normal but will then detect if it's a new test being added and run that test again x number of times in parallel.  

### Describe the solution you'd like.

Two ways I could see doing this when running tests.

1) Add a --allow-duplicates argument which will then not filter out duplicate tests

```
npx wdio run ./wdio.conf.js --spec example.e2e.js example.e2e.js example.e2e.js --allow-duplicates
```

2) Add a --multi-run argument which will allow you to specify how many times you'd want each spec to run (not sure if multi-run is the best name) 

```
npx wdio run ./wdio.conf.js --spec example.e2e.js --multi-run=10
```

### Additional context

I will implement the change if this is something we can do.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

christian-bromann:
I tend towards solution two with the following constraint: the `--multi-run` parameter can only be applied when also one single `--spec` parameter is applied. The testrunner needs to know which spec file to run multiple times. Or we find a different way to define it.

WillBrock:
I was kind of thinking it would run any test that is specified or matched multiple times. If someone passes a suite that contains 500 tests and the multi-run option is set it would run those, although I don't think someone would want to do that.

I could see in my case to pass multiple spec's. The reason for that is if we've disabled some tests and we are re-introducing them I would want to run each test that is being enable in the branch.

```
npx wdio run ./wdio.conf.js --spec foo.e2e.js bar.e2e.js baz.e2e.js --multi-run=10
```
Probably not something someone would do often but if they wanted they could pass the suite and multi-run. I would actually probably use the suite part in some instances. We run tests each night on fresh databaess to create data for our manual testers and right now I loop over tests to do that but I could use the mulit-run option to do that instead.
```
npx wdio run ./wdio.conf.js --suite FooBar --multi-run=3
```

ArminazK:
hello, imho `--multi-run` could be confused with `multiremote` mode.
To me this sounds like project specific feature. So anyone can loop easily their tests in different ways on current wdio. 
Instead of creating this complex logic, better to focus removing flakiness in tests/framework.
[...]

PR #12372, opened 2024-02-26
(wdio-config): expand and simplify the --multi-run feature

erwinheitzman:
## Proposed changes

At my current project we manually filter our spec files and pass them to webdriverio.
In order to use the --multi-run with our setup I have tested if it is possible to run the following:

- --spec (already worked)
- --suite (already worked)
- --spec with --suite (already worked)
- without any flags (did not work but works now)

## Types of changes

## Checklist

### DIFF
--- a/packages/wdio-config/tests/node/configparser.test.ts
+++ b/packages/wdio-config/tests/node/configparser.test.ts
@@ -853,7 +853,7 @@ describe('ConfigParser', () => {
             expect(specs).toHaveLength(1)
         })
 
-        it('should include specs from suite 3 times with mulit-run', async () => {
+        it('should include specs from suite 3 times with multi-run', async () => {
             const configParser = await ConfigParserForTestWithAllFiles(FIXTURES_CONF)
             await configParser.initialize({ suite: ['functional'], multiRun: 3 })
 
@@ -861,11 +861,12 @@ describe('ConfigParser', () => {
             expect(specs).toHaveLength(3)
         })
 
-        it('should throw an error if multi-run is set but no spec or suite is specified', async () => {
+        it('should allow multi-run to run all tests', async () => {
             const configParser = await ConfigParserForTestWithAllFiles(FIXTURES_CONF)
             await configParser.initialize({ multiRun: 3 })
 
-            expect(() => configParser.getSpecs()).toThrow('The --multi-run flag requires that either the --spec or --suite flag is also set')
+            const specs = configParser.getSpecs()
+            expect(specs).toHaveLength(9)
         })
 
         it('should include spec when specifying a suite unless excluded', async () => {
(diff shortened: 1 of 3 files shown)


wswebcreation (approved):
@erwinheitzman 

This looks good, but do we also know why we initially set the multiRun in combination with the `--spec` or `--suite` flag?

erwinheitzman:
It was a suggestion made by @christian-bromann in the original PR #9926 because he thought that people might accidentally run their whole suite multiple times by accident. If you use the `--multi-run` flag you know you are rerunning your tests and if you don't specify anything, it will run everything. This is always the case so I don't see how it would suddenly be an issue when using this flag.

erwinheitzman:
I'm also planning to rename `--multi-run 5` to `--repeat 5` as I think it is a bit clearer (and because it is similar to the linux repeat cli).

erwinheitzman:
Please do not merge yet, I need to make some additional changes

packages/wdio-config/tests/node/configparser.test.ts
```diff
@@ -1,5 +1,5 @@
 import url from 'node:url'
-import path from 'node:path'
+import path, { resolve } from 'node:path'
```
christian-bromann, line 2: I suggest to either import the default or named functions but we should not mix
```suggestion
import path from 'node:path'
```

packages/wdio-config/tests/node/configparser.test.ts
```diff
@@ -853,14 +853,28 @@ describe('ConfigParser', () => {
...
             const configParser = await ConfigParserForTestWithAllFiles(FIXTURES_CONF)
             await configParser.initialize({ suite: ['functional'], multiRun: 3 })
 
             const specs = configParser.getSpecs()
             expect(specs).toHaveLength(3)
         })
 
+        it('should repeat specs in specific order to fail early', async () => {
+            const spec1 = resolve(__dirname, '../utils.test.ts')
+            const spec2 = resolve(__dirname, 'configparser.test.ts')
```
christian-bromann, line 866:
```suggestion
            const spec1 = path.resolve(__dirname, '../utils.test.ts')
            const spec2 = path.resolve(__dirname, 'configparser.test.ts')
```

christian-bromann (approved):
One last minor thing, after that LGTM 👍

christian-bromann:
> [erwinheitzman](https://github.com/erwinheitzman) added [PR: Breaking Change 💥](https://github.com/webdriverio/webdriverio/labels/PR%3A%20Breaking%20Change%20%3Aboom%3A) and removed [PR: Polish 💅](https://github.com/webdriverio/webdriverio/labels/PR%3A%20Polish%20%3Anail_care%3A) labels [8 hours ago](https://github.com/webdriverio/webdriverio/pull/12372#event-11934625702)

I wonder if we should just mark this as polish as we never documented the order of re-execution. Wdyt?

erwinheitzman:
I think the chance of any impact is slim and also if there is an impact it's likely small as well. Of course it's guessing work but I think polish should be fine indeed 👍

PR #12373, opened 2024-02-26
(docs): update our best practices

erwinheitzman:
## Proposed changes

Updated our best practices with some common examples of things I have seen our users do that they are not aware of that it can be done in a better way.

## Types of changes

## Checklist

### DIFF
--- a/website/docs/BestPractices.md
+++ b/website/docs/BestPractices.md
@@ -92,9 +92,8 @@ await expect(button).toBeDisplayed()
 
 ## Lazy loading and promise chaining
 
-WebdriverIO has some tricks up it's sleeve when it comes to writing clean code.
+WebdriverIO has some tricks up it's sleeve when it comes to writing clean code as it can lazy load the element which allows you to chain your promises.
 
-Lazy load your elements:
 ```js
 // 👎
 const div = await $('div')
--- a/website/docs/BestPractices.md
+++ b/website/docs/BestPractices.md
@@ -95,8 +95,7 @@ await expect(button).toBeDisplayed()
 WebdriverIO has some tricks up it's sleeve when it comes to writing clean code as it can lazy load the element which allows you to chain your promises and reduces the amount of `await`. This also allows you to pass the element as a ChainablePromiseElement instead of an Element and for easier use with page objects.
 
 So when do you have to use `await`?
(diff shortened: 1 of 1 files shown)


christian-bromann (commented):
One comment

website/docs/BestPractices.md
```diff
@@ -90,6 +90,33 @@ It achieves this by automatically retrying the assertion until it passes or time
...
+// or
+await (await (await $('div')).$('>>>button')).click()
+```
+
+```js
+// 👍
+const div = $('div')
+const button = $('>>>button') // shadow dom
+
+await div.button.click()
```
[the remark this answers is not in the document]
erwinheitzman: Good catch, thank you. This has been resolved.

website/docs/BestPractices.md
```diff
@@ -90,6 +90,31 @@ It achieves this by automatically retrying the assertion until it passes or time
 await expect(button).toBeDisplayed()
 ```
 
+## Lazy loading and promise chaining
+
+WebdriverIO has some tricks up it's sleeve when it comes to writing clean code as it can lazy load the element which allows you to chain your promises and reduces the amount of `await`. This also allows you to pass the element as a ChainablePromiseElement instead of an Element and for easier use with page objects.
+
+So when do you have to use `await`?
+- whenever you perform an action using the `browser` global object such as `execute()`, `getUrl()`, etc
+- whenever you perform an action using the `element` global object, such as `click()`, `waitForDisplayed()`, etc
```
christian-bromann, line 99: I would phrase this the other way around: you only don't have to use `await` if fetching elements via `$`. All other cases require it.

erwinheitzman: You're right, I'll update it. Thanks!

My team also had some nice feedback and asked if we can update our types for the `$` and `$$` command because of this as currently it requires us to `await` them according to the typees. I have a feeling we can but am curious about your view on this.

christian-bromann: Actually this is a fantastic idea I didn't think about before. This would help us to resolve this nasty DX bug where the IDE would autocomplete to something like this:

```ts
await (await $('foo')).click()
```

So I am 💯  for this!

christian-bromann (approved):
Nice one 👍

PR #12376, opened 2024-02-27
(webdriverio): polish click command, fix some issues

erwinheitzman:
## Proposed changes

The click command works differently when using it like this `elem.click()` compared to `elem.click({})`.
This meant that the click command did not use the workaround that was implemented ago by default, which in fact it should.
To fix this I have done the following:
 - I have split the logic in three parts as our users found the code confusing to read, we now have little logic in the command which calls two functions `actionClick` and `elementClick`.
 - `elementClick` is now a separate function living in the root of the file which contains very little logic and now has the scroll workaround correctly applied.
 - `actionClick` is now a separate function living in the root of the file which contains quite some logic to calculate the `x` and `y` coördinates.
 - We now merge the actual options and default options instead of only using defaults when nothing is passed
 - The button mapping has been removed as this is already handled by the `action` command, no need to map it twice in a row
 - A useful comment is added to show that when using the `elementClick` we should only apply the workaround when the condition is met, else we should throw so that our middleware can do it's thing (like handle the case of a stale element)

This PR solves all of the above, what does this mean for our users?
- Scroll the element to the center of the page in case a `element click intercepted` error is thrown when using `.click()` just like we do when passing an object to the command
- Easier to read and understand the codeleading to more correct contributions in the future

## Types of changes

## Checklist

## Backport Request


erwinheitzman:
I could use some help fixing the last two tests. It seems like using the previous default `elementClick` a stale element was handled but now we use the `actions` API instead I do not see this behaviour in the tests. No idea what is expected here and what to do next.

packages/webdriverio/src/commands/element/click.ts
```diff
@@ -83,77 +83,62 @@ import type { Button } from '../../utils/actions/index.js'
...
     }
 
     const browser = getBrowserObject(this)
-    if (xOffset || yOffset) {
+    if (x || y) {
         const { width, height } = await browser.getElementRect(this.elementId)
-        if ((xOffset && xOffset < (-Math.floor(width / 2))) || (xOffset && xOffset > Math.floor(width / 2))) {
-            throw new Error('xOffset would cause a out of bounds error as it goes outside of element')
+        if ((x && x < (-Math.floor(width / 2))) || (x && x > Math.floor(width / 2))) {
+            throw Error(`{ x: ${x} } would cause an out of bounds error as it goes outside of element`)
```
christian-bromann, line 118:
```suggestion
            throw new Error(`{ x: ${x} } would cause an out of bounds error as it goes outside of element`)
```

packages/webdriverio/src/commands/element/click.ts
```diff
@@ -83,77 +83,62 @@ import type { Button } from '../../utils/actions/index.js'
...
         }
-        if ((yOffset && yOffset < (-Math.floor(height / 2))) || (yOffset && yOffset > Math.floor(height / 2))) {
-            throw new Error('yOffset would cause a out of bounds error as it goes outside of element')
+        if ((y && y < (-Math.floor(height / 2))) || (y && y > Math.floor(height / 2))) {
+            throw Error(`{ y: ${y} } would cause an out of bounds error as it goes outside of element`)
```
christian-bromann, line 121:
```suggestion
            throw new Error(`{ y: ${y} } would cause an out of bounds error as it goes outside of element`)
```

packages/webdriverio/src/commands/element/click.ts
```diff
@@ -83,77 +83,62 @@ import type { Button } from '../../utils/actions/index.js'
  */
 export async function click(
     this: WebdriverIO.Element,
-    options?: ClickOptions
+    options: Partial<ClickOptions> = {}
 ) {
-    if (typeof options === 'undefined') {
-        return this.elementClick(this.elementId)
```
christian-bromann, line 89: What's the incentive to move away from the basic click command?

christian-bromann: > This meant that the click command did not use the workaround that was implemented ago by default, which in fact it should.

The WebDriver protocol does the "scroll into view" work for us:

> [Scroll into view](https://www.w3.org/TR/webdriver2/#dfn-scrolls-into-view) the element’s [container](https://www.w3.org/TR/webdriver2/#dfn-container).

This has been working reliably.

erwinheitzman: Sure, just run this snippet:
```js
    await browser.url('https://webdriver.io');
    await $('div*=Watch Talks about WebdriverIO').scrollIntoView();
    await $('a[href="https://github.com/webdriverio"]').click();
    await browser.pause(2000);
```
The workaround we have in the click command for this does not work because it works differently by default compared to when passing in an object. I think we should use one method, not two.

erwinheitzman: In order to get our tests with similar behaviour working in our project we have the following workaround which I want to eliminate:
```js
const elemToClick = $('example')
await elemToClick.scrollIntoView({ block: 'center' }) // first scroll the element into view (NOT the top since often times there's a header that is in the way)
await elemToClick.click()
```
Other possible workarounds would be:
```js
const elemToClick = $('example')
await elemToClick.click({}) // since it now sees an element, it will skip the element click and will correctly apply the workaround
```
```js
browser.overwriteCommand('click', async function (originClick, options) {
  originClick(options || {})
}, true);
const elemToClick = $('example')
await elemToClick.click() // can use the click like you normally would but instead of undefined, it passes an empty object so that the workaround is applied
```

erwinheitzman: I have also considered about applying the same workaround to the elementClick but I want to get rid of the difference in behaviour and provide a robust click that behaves the same. Whether or not I pass in an object or not should not matter.
```js
  try {
    return this.elementClick(this.elementId)
  } catch {
    await this.scrollIntoView({ block: 'center', inline: 'center' })
    return this.elementClick(this.elementId)
  }
}      
```

packages/webdriverio/src/commands/element/click.ts
```diff
@@ -84,79 +84,93 @@ const log = logger('webdriver')
...
-                y: yOffset
-            })
+            .move({ origin: element, x, y })
             .down({ button })
             .up({ button })
             .perform(skipRelease)
     }
     try {
-        await clickNested()
+        return clickNested()
```
christian-bromann, line 171: This would never throw because `clickNested` returns a Promise, I'ld suggest to revert this change.

packages/webdriverio/tests/commands/element/__snapshots__/click.test.ts.snap
```diff
@@ -25,7 +25,7 @@ exports[`click test > should allow to left click on an element by passing a butt
     {
       "altitudeAngle": 0,
       "azimuthAngle": 0,
-      "button": 0,
+      "button": "left",
```
christian-bromann, line 28: Can we verify that this is accepted by the WebDriver protocol? From what I see:

> To process a pointer up or pointer down action given action item, and action:
> 
> Let button be the result of getting the property button from action item.

I can't find anywhere that it accepts `left`, `middle` or `right`.

packages/webdriverio/src/commands/element/click.ts
```diff
@@ -84,79 +84,93 @@ const log = logger('webdriver')
...
+* Moreover the action  with 'nearest' behavior by default where element is aligned at the bottom of its ancestor.
+* and could be overlapped. Scroll to center should definitely work even if element was covered with sticky header/footer
+*/
+async function workaround(element: WebdriverIO.Element) {
+    await element.scrollIntoView({ block: 'center', inline: 'center' })
+}
+
+async function elementClick(element: WebdriverIO.Element) {
+    try {
+        return element.elementClick(element.elementId)
```
christian-bromann, line 112: same here
```suggestion
        await element.elementClick(element.elementId)
```

christian-bromann (commented):
Some comments. Can you do me a favor and list the changes that were done to the command and how it impacts the user?

packages/webdriverio/src/commands/element/click.ts
erwinheitzman: We previously also returned the click, we expect it in our tests as sometimes the result is expected to be `null`

packages/webdriverio/src/commands/element/click.ts
erwinheitzman: Could you explain the difference between these two when it comes to throwing please?
```js
async function example() {
  await clickNested() // can throw
}
```
async function example() {
  return clickNested() // can throw
}
```

Comment (erwinheitzman) [reply packages/webdriverio/src/commands/element/click.ts:171]: Both are throwing for me:
Screenshot 2024-03-01 at 10 37 33

Comment (erwinheitzman) [reply packages/webdriverio/src/commands/element/click.ts:171]: So does this example:
Screenshot 2024-03-01 at 10 40 10

File: packages/webdriverio/tests/commands/element/__snapshots__/click.test.ts.snap (hunk 5 above)
Comment (erwinheitzman) [reply packages/webdriverio/tests/commands/element/__snapshots__/click.test.ts.snap:28]: This works because we map it in the `action` command:
Screenshot 2024-03-01 at 10 42 29

Comment (erwinheitzman) [reply packages/webdriverio/tests/commands/element/__snapshots__/click.test.ts.snap:28]: No need to map it twice

Comment (erwinheitzman) [reply packages/webdriverio/tests/commands/element/__snapshots__/click.test.ts.snap:28]: Not sure what happens if you send more params though, could be that we don't map it then in which case it would be a bug in the `action` command and we simply need to fix that.

Comment (erwinheitzman) [reply packages/webdriverio/tests/commands/element/__snapshots__/click.test.ts.snap:28]: comparing the up (top) with the the down (bottom) command, there's already a difference in what we pass. I think it should be the following: 
```js
        function mapParameters(params) {
          const buttons = {
            left: 0,
            middle: 1,
            right: 2
          }
          if (typeof params === 'string') {
            return buttons[params]
          }
          if (typeof params.button === 'string') {
            return { ...params, button: buttons[params.button] }
          }
          return params
        }
        this.sequence.push({
            type: 'pointerUp',
            ...PARAM_DEFAULTS,
            ...mapParameters(params)
        })
``` 
Screenshot 2024-03-01 at 10 46 06

erwinheitzman: I have created a PR to fix and align the mapping: #12390

erwinheitzman:
> Some comments. Can you do me a favor and list the changes that were done to the command and how it impacts the user?

I have updated the description with a list of all the changes, why they were made and what it means for our users.

erwinheitzman:
@christian-bromann 
```js
async function throwsError () {
    throw new Error('bar is missing')
}

async function foo () {
    return throwsError()
}

async function example1 () {
    try {
        await foo()
    } catch (err) {
        console.error('I will be reached!')
    }
}

async function example2 () {
    try {
        return foo()
    } catch (err) {
        console.error('I will never be reached!')
    }
}

await example1()
// the error of example1 stops the execution
// therefore the error for example12 is never thrown
// run them separately and you'll see that both run as expected
await example2()
```

packages/webdriverio/src/commands/element/click.ts
erwinheitzman: You're absolutely right, thanks for catching this. Never knew that this could not be handled like this.

Pushed a fix in which I use `return await`

christian-bromann:
@erwinheitzman can you take at the failing tests?

erwinheitzman:
@christian-bromann do you think we should backport this one?

christian-bromann:
It seems that the changes are mostly improve code quality and don't impact the functionality. I'll leave it up to you. It seems fine to keep it as `v9` change and don't back-port it.

PR #12386, opened 2024-02-29
removal of the chainable promise types

erwinheitzman:
## Proposed changes

Based on a talk I had with Christian and my current team I decided to update our types to reflect that you do not need to `await` the `$` and `$$` commands.

I have replaced the types with TypeScript's `Awaited` utility type: https://www.typescriptlang.org/docs/handbook/utility-types.html#awaitedtype

## Types of changes

## Checklist

### DIFF
--- a/packages/webdriverio/src/types.ts
+++ b/packages/webdriverio/src/types.ts
@@ -18,45 +18,10 @@ type $ElementCommands = typeof ElementCommands
 type ElementQueryCommands = '$' | 'custom$' | 'shadow$' | 'react$'
 type ElementsQueryCommands = '$$' | 'custom$$' | 'shadow$$' | 'react$$'
 type ChainablePrototype = {
-    [K in ElementQueryCommands]: (...args: Parameters<$ElementCommands[K]>) => ChainablePromiseElement<ThenArg<ReturnType<$ElementCommands[K]>>>
+    [K in ElementQueryCommands]: (...args: Parameters<$ElementCommands[K]>) => Awaited<ThenArg<ReturnType<$ElementCommands[K]>>>
 } & {
-    [K in ElementsQueryCommands]: (...args: Parameters<$ElementCommands[K]>) => ChainablePromiseArray<ThenArg<ReturnType<$ElementCommands[K]>>>
+    [K in ElementsQueryCommands]: (...args: Parameters<$ElementCommands[K]>) => Awaited<ThenArg<ReturnType<$ElementCommands[K]>>>
 }
-
-type AsyncElementProto = {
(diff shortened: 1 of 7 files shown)


packages/webdriverio/src/types.ts
```diff
@@ -18,46 +18,11 @@ type $ElementCommands = typeof ElementCommands
...
-    error?: Promise<Error>
-    /**
-     * index of the element if fetched with `$$`
-     */
-    index?: Promise<number>
-}
-export interface ChainablePromiseElement<T> extends
-    ChainablePromiseBaseElement,
-    AsyncElementProto,
-    Promise<T>,
```
christian-bromann, line 58: I was thinking of only removing this. I am not sure if removing `ChainablePromiseElement` would be a good idea.

erwinheitzman: I find that my colleagues have no clue what `ChainablePromiseElement` is and just `await` it and I cannot blame them. I think having `WebdriverIO.Element` as type makes much more sense, what do you think?

christian-bromann: I think the noticeable difference is that:

```ts
const chainablePromiseElement = $('elem')
console.log(chainablePromiseElement.selector) // outputs: "Promise<string>"
```
whereas:

```ts
const wdioElement = await $('elem')
console.log(wdioElement.selector) // outputs: "elem"
```

So I am not sure if we can merge this into one single type since the chainable element is not resolved and accessing properties requires to use `await`. I can see that `ChainablePromiseElement` is confusing and I am in favor to consolidate this. Any suggestion how we can overcome the type difference?

erwinheitzman: Interesting, does this mean that you need to `await` the `$` command when accessing any properties but not when you want to execute an action or chain it?

christian-bromann: Correct. You can do `$('elem').$('elem').$('elem').$('elem')` as our [proxy command wrapper](https://github.com/webdriverio/webdriverio/blob/58eb3c2914a49d2918c4a3ebe568b33280f6db94/packages/wdio-utils/src/shim.ts#L85-L262) allows to chain these promises.

erwinheitzman: I believe you can extend the type to check if it has the elementId and if not you can assign it a Promise like type. I knowledge of Typescript is not where I want it to be so I might get a course on it to learn about more complex types 👍

christian-bromann:
Implemented by https://github.com/webdriverio/webdriverio/pull/13051

Issue #12428, opened 2024-03-04
[🐛 Bug]: CustomService is run twice


tdomzalski:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

8.32.4

### Node.js Version

18

### Mode

Standalone Mode

### Which capabilities are you using?

```typescript
n/a
```

### What happened?

I run the Launcher with custom service from external script (please see the code here: https://github.com/tdomzalski/wdio-customservice-test).
The problem: when I execute `node main.js` the service is being executed twice (see the logs displayed in console: constructor and service's methods are executed twice).

Furthermore I get the following error in `wdio-0-0.log` file and browser is not run:

```
2024-03-04T16:07:24.106Z INFO @wdio/local-runner: Run worker command: run
2024-03-04T16:07:24.135Z ERROR @wdio/local-runner: Failed launching test session: TypeError: Cannot read properties of null (reading 'before')
    at ConfigParser.addService (file:///C:/WORKS/wdio-customservice-test/node_modules/@wdio/config/build/node/ConfigParser.js:207:43)
    at Array.map (<anonymous>)
    at Runner.run (file:///C:/WORKS/wdio-customservice-test/node_modules/@wdio/runner/build/index.js:75:89)
```

### What is your expected behavior?

The service should be run only once.
The browser should be executed and tests executed.

### How to reproduce the bug.

Please see the code here: https://github.com/tdomzalski/wdio-customservice-test 

### Relevant log output

```typescript
C:\WORKS\wdio-customservice-test>node main.js

Execution of 1 workers started at 2024-03-04T16:07:14.828Z

MyService constructor { timestamp: 1709568434789 }
MyService constructor { timestamp: 1709568434789 }
MyService onPrepare
MyService onPrepare
onWorkerStart [ 'file:///C:/WORKS/wdio-customservice-test/test/mytest.spec.js' ]
[0-0] FAILED
onWorkerEnd
MyService onComplete
MyService onComplete

Spec Files:      0 passed, 1 failed, 1 total (100% completed) in 00:00:11
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

christian-bromann:
Thanks to @jemishgopani for fixing this 👏

PR #12430, opened 2024-03-05
[12428] CustomService is run twice

jemishgopani:
## Proposed changes
[12428](https://github.com/webdriverio/webdriverio/issues/12428)

## Types of changes

## Checklist

## Backport Request

- [x] Back-ported PR at [#12751](https://github.com/webdriverio/webdriverio/pull/12751)

## Further comments
- Removed args from the constructor because we are already initializing args in the run method.
- As always I'll add backport PR if this PR changes look good to maintainers.


christian-bromann:
@jemishgopani thanks for the patch, mind taking a look at the failing smoke tests? I think your solution fails for when you pass in command line arguments e.g. `npx wdio run wdio.conf.js --spec foo`

jemishgopani:
@christian-bromann I have tried to fix please take a look and provide your feedback.

packages/wdio-config/src/node/ConfigParser.ts
```diff
@@ -185,6 +187,13 @@ export default class ConfigParser {
             this._config.exclude = object.exclude as string[]
         }
 
+        /**
+         * add this check to avoid duplication of the "services" in the config
+         */
+        if (!addServicesToConfig) {
+            this._config.services = services
```
christian-bromann, line 194: Instead of doing that I suggest to use [`customMerge`](https://www.npmjs.com/package/deepmerge#custommerge) and skip merging for `services` and `reporters`. I think we can generally avoid to merge all arrays here.

christian-bromann (changes requested):

packages/wdio-config/src/node/ConfigParser.ts
jemishgopani: Do you mean we have to remove the flag and implement customMerge?

christian-bromann: Yeah, I would suggest using customMerge option as it makes it easier. Note that we also need this for `reporters` as it is the same use case

packages/wdio-config/src/node/ConfigParser.ts
```diff
@@ -168,8 +169,17 @@ export default class ConfigParser {
...
         const exclude = Array.isArray(object.exclude) ? object.exclude : []
-        this._config = deepmerge(this._config, object) as TestrunnerOptionsWithParameters
-
+        const customDeepMerge = deepmergeCustom({
+            mergeArrays: (values, utils, meta) => {
+                if (meta?.key === 'services') {
+                    return this._config.services
+                } else if (meta?.key === 'reporters') {
+                    return this._config.reporters
+                }
```
christian-bromann, line 178:
```suggestion
                if (meta && ['services', 'reporters'].includes(meta.key)) {
                    return this._config[meta.key as keyof WebdriverIO.Config]
                }
```

packages/wdio-config/src/node/ConfigParser.ts
```diff
@@ -228,6 +238,24 @@ export default class ConfigParser {
         }
     }
 
+    /**
+     * merge only reporters and services to the config object
+     * @param object desired object to merge into the config object
+     * @private
+     */
+    private mergeReportersAndServices(object: MergeConfig) {
```
christian-bromann, line 246: Do we still need this function?

jemishgopani: Yes because any way we have to merge config which is added via launcher

packages/wdio-config/src/node/ConfigParser.ts
jemishgopani: Updated 👍

christian-bromann (commented):
@jemishgopani can we merge what happens in `mergeReportersAndServices` into the custom merge function? I think there is no need to have an extra function to merge reporters and services, instead we can just use the custom merger

jemishgopani:
> @jemishgopani can we merge what happens in `mergeReportersAndServices` into the custom merge function? I think there is no need to have an extra function to merge reporters and services, instead we can just use the custom merger

Applied new changes please have a look.

packages/wdio-config/src/node/ConfigParser.ts
```diff
@@ -177,8 +177,18 @@ export default class ConfigParser {
     private merge(object: MergeConfig = {}, addPathToSpecs = true) {
         const spec = Array.isArray(object.spec) ? object.spec : []
         const exclude = Array.isArray(object.exclude) ? object.exclude : []
-        this._config = deepmerge(this._config, object) as TestrunnerOptionsWithParameters
-
+        const customDeepMerge = deepmergeCustom({
+            mergeArrays: (values, utils, meta) => {
+                if (meta && ['services', 'reporters'].includes(meta.key as string)) {
```
EdmundasD, line 182: Experiencing a similar issue when passing `capabilities` into `Launcher` (capabilities are duped due to `deepmerge`). Here is a reproducible example: https://github.com/EdmundasD/testing-wdio-v8
It appears that adding `capabilities` to the array here resolves that issue, after doing some preliminary testing.

christian-bromann: @EdmundasD can you verify if this is fixed when applying the changes in this PR?

EdmundasD: Yeap, that's exactly what I did. It resolved the issue I was having and passing the capabilities directly in the config also works as expected.

packages/wdio-config/src/node/ConfigParser.ts
```diff
@@ -177,8 +177,18 @@ export default class ConfigParser {
     private merge(object: MergeConfig = {}, addPathToSpecs = true) {
         const spec = Array.isArray(object.spec) ? object.spec : []
         const exclude = Array.isArray(object.exclude) ? object.exclude : []
-        this._config = deepmerge(this._config, object) as TestrunnerOptionsWithParameters
-
+        const customDeepMerge = deepmergeCustom({
```
christian-bromann, line 180: One last ask: can we add a comment describing why we do a custom deepmerge here? It should mention that the default merge would duplicate entries when using the launcher programatically.

packages/wdio-config/src/node/ConfigParser.ts
christian-bromann: Thanks for verifying.

packages/wdio-config/src/node/ConfigParser.ts
jemishgopani: Sure I'll do

packages/wdio-config/src/node/ConfigParser.ts
EdmundasD: @christian-bromann, to be clear, `capabilities` need to be added to the array here in order to resolve the issue I described:
```
...
if (meta && ['services', 'reporters', 'capabilities'].includes(meta.key as string)) {
...
```
cc @jemishgopani

jemishgopani: @christian-bromann Do we need this change in this PR?

christian-bromann: I think we can include it. I doubt it will blow up the scope but I will let you decide.

christian-bromann (commented):
@jemishgopani do you think we can capture this in a unit test?

jemishgopani:
> @jemishgopani do you think we can capture this in a unit test?
we can validate this by checking object length but that type of test doesn't make sense to me.
I think we have to merge these changes.

if all looks good then I'll raise another backport PR to v8.

jemishgopani:
Hi @christian-bromann 
Can you please have a look on changes?

packages/wdio-config/tests/node/configparser.test.ts
```diff
@@ -689,6 +689,35 @@ describe('ConfigParser', () => {
 
             expect(configParser.getCapabilities()).toMatchObject([{ browserName: 'safari' }])
         })
+
+        it('should allow to specifying services', async () => {
+            const configParser = await ConfigParserForTest(FIXTURES_CONF)
+            await configParser.initialize({
+                services: [
+                    ['MyService', { timestamp: 1709633731716 }]
```
christian-bromann, line 697: Can we add a class here too? Let's make sure that the use case from the original issue is covered here.

jemishgopani: Updated this with class.

christian-bromann (commented):
@jemishgopani thanks for adding a test. I reviewed the changes and it doesn't seem to fix the overall issue. Let me explain:

- I tried to verify whether removing your changes in `packages/wdio-config/src/node/ConfigParser.ts` will make the test break, unfortunately it didn't
- then I tried to verify that the issue is reproducible in the unit test, which it is, for that I applied this change:

```diff
diff --git a/packages/wdio-config/tests/node/configparser.test.ts b/packages/wdio-config/tests/node/configparser.test.ts
index d2ff81ce0..33f170fb4 100644
--- a/packages/wdio-config/tests/node/configparser.test.ts
+++ b/packages/wdio-config/tests/node/configparser.test.ts
@@ -13,7 +13,6 @@ import MockFileContentBuilder from '../lib/MockFileContentBuilder.js'
 import type { FilePathsAndContents, MockSystemFilePath, MockSystemFolderPath } from '../lib/MockPathService.js'
 import ConfigParserBuilder from '../lib/ConfigParserBuilder.js'
 import { FileNamed, realReadFilePair, realRequiredFilePair } from '../lib/FileNamed.js'
-import TestCustomService from '../__fixtures__/test-custom-service.js'
 
 const log = logger('')
 const __dirname = url.fileURLToPath(new URL('.', import.meta.url))
@@ -41,6 +40,8 @@ vi.mock('@babel/register', () => ({
     default: vi.fn()
 }))
 
+class CustomService {}
+
 /**
  * Entirely in memory config structure to avoid reading the file system at all
  *
@@ -70,7 +71,8 @@ const TestWdioConfig_AllInMemory = (): MockFileContent => ({
                 path.join(root, '/tests/validateConfig.test.ts'),
                 path.join(root, '/tests/..', 'src/index.ts')
             ]
-        }
+        },
+        services: [[CustomService, { foo: 'bar' }]],
     }
 })
 
@@ -691,26 +693,24 @@ describe('ConfigParser', () => {
             expect(configParser.getCapabilities()).toMatchObject([{ browserName: 'safari' }])
         })
 
-        it('should allow to specifying services', async () => {
+        it.only('should allow to specifying services', async () => {
             const configParser = await ConfigParserForTest(FIXTURES_CONF)
             await configParser.initialize({
                 services: [
-                    [TestCustomService, { timestamp: 1709633731716 }]
+                    [CustomService, { foo: 'bar' }]
                 ]
             })
 
-            const services = (await configParser.getConfig())['services']
+            const services = configParser.getConfig().services
             expect(services).toHaveLength(1)
             expect(services[0][1].timestamp).toBe(1709633731716)
-            expect(typeof services[0][0]).toBe(typeof TestCustomService)
+            expect(typeof services[0][0]).toBe(typeof CustomService)
         })
 
         it('should allow to specifying reporters', async () => {
             const configParser = await ConfigParserForTest(FIXTURES_CONF)
             await configParser.initialize({
-                reporters: [
-                    ['dot', 'spec']
-                ]
+                reporters: ['dot', 'spec']
             })
 
             const reporters = (await configParser.getConfig())['reporters']
```

Note: the unit tests load config files from memory and not from the actual file system. So there isn't a need for `test-custom-service.js`.

I would suggest creating the custom merge as following: merge the service and reporter arrays by removing all entry objects from one side, e.g.:

```diff
diff --git a/packages/wdio-config/src/node/ConfigParser.ts b/packages/wdio-config/src/node/ConfigParser.ts
index f828b3df3..f4f6302d9 100644
--- a/packages/wdio-config/src/node/ConfigParser.ts
+++ b/packages/wdio-config/src/node/ConfigParser.ts
@@ -2,7 +2,7 @@ import path from 'node:path'
 
 import logger from '@wdio/logger'
 import { deepmerge, deepmergeCustom } from 'deepmerge-ts'
-import type { Capabilities, Options, Services } from '@wdio/types'
+import type { Capabilities, Options, Reporters, Services } from '@wdio/types'
 
 import RequireLibrary from './RequireLibrary.js'
 import FileSystemPathService from './FileSystemPathService.js'
@@ -13,6 +13,8 @@ import { SUPPORTED_HOOKS, SUPPORTED_FILE_EXTENSIONS, DEFAULT_CONFIGS, NO_NAMED_C
 import type { PathService, ModuleImportService } from '../types.js'
 
 const log = logger('@wdio/config:ConfigParser')
+const MERGE_DUPLICATION = ['services', 'reporters'] as const
+type KeyWithMergeDuplication = (typeof MERGE_DUPLICATION)[number]
 
 type Spec = string | string[]
 type ESMImport = { config?: TestrunnerOptionsWithParameters }
@@ -182,12 +184,11 @@ export default class ConfigParser {
          * Add deepmergeCustom to remove array('services', 'reporters', 'capabilities') duplication in the config object
          */
         const customDeepMerge = deepmergeCustom({
-            mergeArrays: (values, utils, meta) => {
-                if (meta && ['services', 'reporters', 'capabilities'].includes(meta.key as string)) {
-                    const mergedArr = deepmerge(this._config[meta.key as keyof WebdriverIO.Config], object[meta.key as keyof WebdriverIO.Config]) as []
-                    return mergedArr.filter((item, index, merged) => {
-                        return merged.indexOf(item) === index
-                    })
+            mergeArrays: ([oldValue, newValue], utils, meta) => {
+                const key = meta?.key as KeyWithMergeDuplication
+                if (meta && MERGE_DUPLICATION.includes(key)) {
+                    const origWithoutObjectEntries = oldValue.filter((value: [Services.ServiceClass, WebdriverIO.ServiceOption] | [Reporters.ReporterClass, WebdriverIO.ReporterOption]) => typeof value[0] === 'object')
+                    return new Set(deepmerge(newValue, origWithoutObjectEntries))
                 }
                 return utils.actions.defaultMerge
             }
diff --git a/packages/wdio-config/tests/node/configparser.test.ts b/packages/wdio-config/tests/node/configparser.test.ts
index d2ff81ce0..d22c9ad04 100644
--- a/packages/wdio-config/tests/node/configparser.test.ts
+++ b/packages/wdio-config/tests/node/configparser.test.ts
@@ -13,7 +13,6 @@ import MockFileContentBuilder from '../lib/MockFileContentBuilder.js'
 import type { FilePathsAndContents, MockSystemFilePath, MockSystemFolderPath } from '../lib/MockPathService.js'
 import ConfigParserBuilder from '../lib/ConfigParserBuilder.js'
 import { FileNamed, realReadFilePair, realRequiredFilePair } from '../lib/FileNamed.js'
-import TestCustomService from '../__fixtures__/test-custom-service.js'
 
 const log = logger('')
 const __dirname = url.fileURLToPath(new URL('.', import.meta.url))
@@ -41,6 +40,8 @@ vi.mock('@babel/register', () => ({
     default: vi.fn()
 }))
 
+class CustomServiceOrReporter {}
+
 /**
  * Entirely in memory config structure to avoid reading the file system at all
  *
@@ -70,7 +71,9 @@ const TestWdioConfig_AllInMemory = (): MockFileContent => ({
                 path.join(root, '/tests/validateConfig.test.ts'),
                 path.join(root, '/tests/..', 'src/index.ts')
             ]
-        }
+        },
+        services: ['appium', [CustomServiceOrReporter, { foo: 'bar' }]],
+        reporters: ['spec', [CustomServiceOrReporter, { foo: 'bar' }]],
     }
 })
 
@@ -691,33 +694,48 @@ describe('ConfigParser', () => {
             expect(configParser.getCapabilities()).toMatchObject([{ browserName: 'safari' }])
         })
 
-        it('should allow to specifying services', async () => {
+        it('should ensure there we do not duplicate class entries', async () => {
             const configParser = await ConfigParserForTest(FIXTURES_CONF)
             await configParser.initialize({
                 services: [
-                    [TestCustomService, { timestamp: 1709633731716 }]
-                ]
-            })
-
-            const services = (await configParser.getConfig())['services']
-            expect(services).toHaveLength(1)
-            expect(services[0][1].timestamp).toBe(1709633731716)
-            expect(typeof services[0][0]).toBe(typeof TestCustomService)
-        })
-
-        it('should allow to specifying reporters', async () => {
-            const configParser = await ConfigParserForTest(FIXTURES_CONF)
-            await configParser.initialize({
+                    'appium',
+                    'sauce',
+                    [CustomServiceOrReporter, { foo: 'bar' }]
+                ],
                 reporters: [
-                    ['dot', 'spec']
+                    'dot',
+                    'spec',
+                    [CustomServiceOrReporter, { foo: 'bar' }] as any
                 ]
             })
 
-            const reporters = (await configParser.getConfig())['reporters']
-            expect(reporters).toHaveLength(1)
-            expect(reporters[0]).toHaveLength(2)
-            expect(reporters[0][0]).toBe('dot')
-            expect(reporters[0][1]).toBe('spec')
+            const { services, reporters } = configParser.getConfig()
+            expect(services).toHaveLength(3)
+            expect(services).toMatchInlineSnapshot(`
+              Set {
+                "appium",
+                "sauce",
+                [
+                  [Function],
+                  {
+                    "foo": "bar",
+                  },
+                ],
+              }
+            `)
+            expect(reporters).toHaveLength(3)
+            expect(reporters).toMatchInlineSnapshot(`
+              Set {
+                "dot",
+                "spec",
+                [
+                  [Function],
+                  {
+                    "foo": "bar",
+                  },
+                ],
+              }
+            `)
         })
     })
```

What do you think?

jemishgopani:
@christian-bromann Thanks 🙏

changes seems looking good to me I'll implement this and update this PR soon.

christian-bromann:
@jemishgopani let me know if there is anything I can do to help.

jemishgopani:
@christian-bromann I have made some changes to your suggested code and updated the code. Whenever you have time please review this PR.

christian-bromann (approved):
LGTM 👍

christian-bromann:
@jemishgopani mind raising the same PR for the `v8` branch?

PR #12431, opened 2024-03-05
(webdriverio) polish isClickable and waitForClickable

erwinheitzman:
## Proposed changes

Followup of #12206

## Types of changes

## Checklist

## Backport Request


packages/webdriverio/tests/scripts/isElementClickable.test.ts
```diff
@@ -199,7 +197,6 @@ describe('isElementClickable script', () => {
         global.document = { elementFromPoint: () => null } as any
 
         expect(isElementClickable(elemMock)).toBe(false)
-        expect(elemMock.scrollIntoView).toBeCalledWith({ block: 'nearest', inline: 'nearest' })
```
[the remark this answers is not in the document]
erwinheitzman: Hey Christian, this was no longer needed because of the double scroll that I have changed into one. What we did was first a scroll to the top and if that wasn't enough to the center. Just like the workaround for moveTo and click, I have applied the same behaviour here and scroll once to the center 👍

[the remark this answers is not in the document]
christian-bromann: Sounds good.

erwinheitzman:
> Overall looks good to me. Can we add some information to the command docs telling when people might want to use `withinViewport` to `true` or `false`?

Done 👍

packages/webdriverio/src/commands/element/isClickable.ts
```diff
@@ -3,17 +3,21 @@ import { ELEMENT_KEY } from 'webdriver'
...
- * - its center is not overlapped with another element
+ * - the element exist and is displayed
  * - is not disabled
- *
- * otherwise return false.
+ * - the element is within the viewport
+ * - the element's center is not overlapped with another element
+
+ * If withinViewport is falsy, the command attempts to scroll the element to the center of the viewport and performs the above checks, after which it attempts to scroll back to it's original coordinates
+ * If withinViewport is true, the command does not attempt to scroll the element to the center of the viewport and instead directly runs the above checks
```
christian-bromann, line 20: If the condition for an element is to be _within the viewport_ , it seems by reading this that setting `withinViewport` to `true` it will always return `false` since the browser won't move the element into the viewport.

erwinheitzman: While I was typing out what happens I actually thought of a small bug, will push a change in a bit and explain it all

erwinheitzman: With `withinViewport: true` it will return `true` if the element is within the viewport, not overlapped by another element and not disabled when the command is called else it will return `false.`

With `withinViewport: true` it will be the same, with the exception that it scrolls the element to the center of the viewport then it checks if it is in the viewport and isn't overlapped by another element and stores the result which is then combined with the return expression

christian-bromann: When should one use `withinViewport: true` and when shouldn't the user use that?

erwinheitzman: It should be used when you want to verify that an element is clickable inside the viewport while it should not be used when you wanted to check if the element is clickable even though it is outside of the current viewport

erwinheitzman: I understand, I find it hard to describe but I will try to find some kind of description that makes sense and will update both 👍

erwinheitzman: @christian-bromann I have updated the documentation to this:
```js
* @param {Boolean=} options.withinViewport set this option to true if you want the command to return false when the element is outside of the viewport
```
along with:
```js
 * An element is considered to be clickable when the following conditions are met:
 *
 * - the element exist and is displayed
 * - is not disabled
 * - the element is within the viewport
 * - the element can be scrolled into the viewport (only when withinViewport is falsy)
 * - the element's center is not overlapped with another element
```

christian-bromann:
@erwinheitzman as discussed in Discord, let's remove the `isWithinViewport` option unless there is actual value for it.

Issue #12434, opened 2024-03-05
[🐛 Bug]: Disabling autoCompileConfig.autoCompile does not work (ts-node v8)


johnp:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

latest

### Node.js Version

20.11.1

### Mode

WDIO Testrunner

### What happened?

Neither the wdio config setting `autoCompileOpts.autoCompile`, nor the CLI argument `--autoCompileOpts.autoCompile=false` allow disabling the ts-node loader. Notably, we have ts-node v8 installed as a transitive dependency which does not provide an `esm/transpile-only` loader entry-point, but which is used by WebdriverIO regardless of any setting. 

edit: The wdio config setting `autoCompileOpts.autoCompile` not preventing ts-loader/babel execution for config parsing is apparently expected (Chicken & egg problem).

### What is your expected behavior?

WebdriverIO should work even if there's an older version of ts-node installed; it should then simply not try to use it. 

Furthermore,  `--autoCompileOpts.autoCompile=false` should always be respected if it's passed early as a CLI argument, as it is documented in `wdio --help`.

### How to reproduce the bug.

An example project can be seen [here](https://github.com/johnp/wdio-ts-node-8-bug). Just install the dependencies with `npm` and run the three scripts in the `package.json`:

1. The `wdio` script doesn't work, even though `wdio.conf.js` has `autoCompile` set to `false`. edit: looks like expected behavior (Chicken & egg problem)
2. The `wdio-broken-arg` script also doesn't work, even though we also explicitly set the argument  
`--autoCompileOpts.autoCompile=false` as documented in `wdio --help`.
3. `wdio-working-arg` does work, but is undocumented.

1 and 2 throw the same error. I was able to trace the 2nd case to `autoCompile` being the truthy string `"false"` instead of a boolean here:

https://github.com/webdriverio/webdriverio/blob/46a9637ccdc423e58b583a4f099727f0ac7ce7b7/packages/wdio-config/src/node/utils.ts#L59-L62

Adding `|| autoCompileConfig.autoCompile === "false"` to this if-condition fixes the 2nd case, but it would be even better to fix both, or to just don't try using ts-node if it's below v9.

### Relevant log output

```typescript
INFO @wdio/cli:launcher: Run onPrepare hook
[...]

Issue #11878, opened 2023-12-16
[💡 Feature]: Replace `ts-node` with `tsx`


christian-bromann:
### Is your feature request related to a problem?

We have been running into issues using `ts-node` lately, see:

- https://github.com/webdriverio/webdriverio/issues/10857
- https://github.com/webdriverio/webdriverio/issues/11875

For running TypeScript files on demand we currently use `ts-node` which seems to have difficulties with ESM/CJS as well as sourcemaps.

### Describe the solution you'd like.

Switch to `tsx` which seems to have general better support for it.

### Describe alternatives you've considered.

n/a

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Delagen:
Now I solved this with running
TSX_TSCONFIG_PATH="./tsconfig.json" NODE_OPTIONS="$NODE_OPTIONS --import tsx" wdio run ./wdio.conf.js

Config need to be compiled to JS or need to be written in plain JS in case to prevent loading ts-node in @wdio/config by detection of extention of config

christian-bromann:
@Delagen writing the config to the file system as JS file is not an option. We should provide the ability to just load TS files and execute them.

Delagen:
@christian-bromann I understand, I use config in JS only to prevent loading of ts-node. All tests and etc. written in Typescript, an d not transpiled to JS manually, all compilation in memory via TSX

goosewobbler:
I took a look at this but it seems there is a fair bit of overlap with the removal of transpiler choice / `autoCompileOpts`, so maybe we do them together or at least that one first.

Some things I noted:

- `tsx` has no configuration outside of setting a custom `tsconfig.json` path
- We may need to do some switching on node version to ensure the correct loader syntax is used: https://tsx.is/node#es-modules-only
- The programmatic use, e.g. replacement for `ts-node` register function, may look quite different: https://github.com/privatenumber/tsx/discussions/102

christian-bromann:
> so maybe we do them together or at least the latter one first.

Sounds good to me. I am open to remove `autoCompileOpts` all together and have it "just" work™.

erwinheitzman:
@goosewobbler sounds good. I think it is good to stick as much as possible to how the tools work out of the box (no config using wdio but instead document how wdio can be used with tsx). So if tsx recommends using environment variables to appoint a tsconfig then we just point to the tsx documentation, simple, clean and most of all less maintenance and confusion

[...]

PR #12436, opened 2024-03-06
(wdio-config): check existence of ts-node/esm/transpile-only.mjs & fix a type issue

johnp:
## Proposed changes

Fixes two issues behind https://github.com/webdriverio/webdriverio/issues/12434. The first fix is to check for the existence of the `ts-node/esm/transpile-only.mjs` loader before trying to use it, as it does not exist in ts-node versions <9 (in our case pulled in by a transitive dependency). The 2nd fix is a type issue with the argument `autoCompileConfig.autoCompile=false` which ends up as the string `"false"`. I added a conversion step to take care of this. Note that this CLI argument is already documented in `wdio --help`.

## Types of changes

## Checklist

## Backport Request

- [x] Back-ported PR at [`#12437`](https://github.com/webdriverio/webdriverio/pull/12437)

## Further comments

### DIFF
--- a/packages/wdio-config/src/node/ConfigParser.ts
+++ b/packages/wdio-config/src/node/ConfigParser.ts
@@ -65,6 +65,15 @@ export default class ConfigParser {
         if (_initialConfig.spec) {
             _initialConfig.spec = makeRelativeToCWD(_initialConfig.spec) as string[]
         }
+        /**
+         * the autoCompileOpts.autoCompile CLI argument has to be converted
+         * from type string to type boolean
+         */
+        if (typeof _initialConfig.autoCompileOpts?.autoCompile === 'string') {
+            const strValue = _initialConfig.autoCompileOpts.autoCompile as unknown as string
+            _initialConfig.autoCompileOpts.autoCompile = !!strValue && strValue !== 'false'
+        }
+
         this.merge(_initialConfig, false)
     }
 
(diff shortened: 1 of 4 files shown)


erwinheitzman:
Hi @johnp currently the access is checked but nothing is done with the result which doesn't seem right.

Usually you see something like:
```js
try {
  await access('path')
} catch (error) {
  // do something
}
```

Regarding the autocompileopts, I am planning to remove it and instead recommend use of the environment variables as shown in our docs here: https://webdriver.io/docs/typescript as this is how ts-node expects things to be used. It also prevents us from adding magic/logic that doesn't have to be maintained and isn't needed.

christian-bromann:
I just want to point out that we have https://github.com/webdriverio/webdriverio/issues/11878 planned for v9. I wonder if this might help us resolve some of these issues.

erwinheitzman:
@johnp I tried reproducing this with a new project and using that it works fine. I am currently unsure what the difference is but I think the problem lies in the setup/configuration still

erwinheitzman:
Can you explain why you need this when you can set `autoCompile` to `false` in your config?

erwinheitzman:
@johnp good points. looks good to me. what do you think @christian-bromann ? What I do think is that perhaps we should merge this only to v8

christian-bromann (approved):
LGTM 👍 

Let's move forward and merge this for v9 too in case we decide not to migrate to `tsx`.

Thanks @johnp

PR #12542, opened 2024-03-22
(webdriverio): add execute commands to the element scope/object

erwinheitzman:
## Proposed changes

Adds the `execute` and `executeAsync` commands to the element scope/object.

Why do we do this? Currently when using `await browser.execute(elem => {}, await $('example'))` it won't automatically wait for the element, this PR allows the user to do `await $('example').execute(elem => {})` which automatically waits for the element and feels more intuitive to use.

## Types of changes

## Checklist


packages/webdriverio/src/commands/element/execute.ts
```diff
@@ -0,0 +1,63 @@
...
+ * and the values may be accessed via the arguments object in the order specified.
+ *
+ * Arguments may be any JSON-primitive, array, or JSON object. JSON objects that define a WebElement
+ * reference will be converted to the corresponding DOM element. Likewise, any WebElements in the script
+ * result will be returned to the client as WebElement JSON objects.
+ *
+ * <example>
+    :execute.js
+    it('should wait for the example element to exist, then inject javascript on the page', async () => {
+        const result = await $('example').execute((...args, elem) => {
```
christian-bromann, line 22: This needs to be the other way around. A spread operator has to be last.
```suggestion
        const result = await $('example').execute((elem, ...args) => {
```

packages/webdriverio/src/commands/element/execute.ts
```diff
@@ -0,0 +1,63 @@
...
+            console.log(args) // logs added arguments
+            console.log(elem) // logs the element
+            return a + b + c + d
```
christian-bromann, line 25: I wonder if we could connect the example to an element here, e.g. show an html element like:

```html
<div>Hello World!</div>
```
and the example be like:
```ts
const text = await $('div').execute((elem, a, b, c, d) => {
  return elem.textContent + a + b + c + d
}, 1, 2, 3, 4);
console.log(text); // returns "Hello World1234"
```

packages/webdriverio/src/commands/element/execute.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    /**
+     * instances started as multibrowserinstance can't getting called with
+     * a function parameter, therefore we need to check if it starts with "function () {"
+     */
+    if (typeof script === 'function') {
+        script = `return (${script}).apply(null, arguments)`
+    }
+
+    return this.executeScript(script, verifyArgsAndStripIfElement([...args, this]))
```
christian-bromann, line 60: Having the element first is the better approach here IMHO because it is the fixed variable that always will be there while we never know how many parameters are provided. 
```suggestion
    return this.executeScript(script, verifyArgsAndStripIfElement([this, ...args]))
```

packages/webdriverio/src/commands/element/executeAsync.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * instances started as multibrowserinstance can't getting called with
+     * a function parameter, therefore we need to check if it starts with "function () {"
+     */
+    if (typeof script === 'function') {
+        script = `return (${script}).apply(null, arguments)`
+    }
+
+    return this.executeAsyncScript(script, verifyArgsAndStripIfElement([...args, this]))
```
christian-bromann, line 72: same

packages/webdriverio/tests/commands/element/execute.test.ts
```diff
@@ -0,0 +1,53 @@
...
+describe('execute test', () => {
+    it('should execute the script', async () => {
+        const browser = await remote({
+            baseUrl: 'http://foobar.com',
+            capabilities: {
+                browserName: 'foobar'
+            }
+        })
+
+        await browser.$('#foo').execute((a, b, c) => a + b + c, 1, 2, 3)
```
christian-bromann, line 19: Let's make sure we incorporate the element into the test

christian-bromann (changes requested):
Thanks for raising the PR. This will simplify a lot of use cases.

erwinheitzman:
Hey Christian, I was looking into this and when creating this I didn't realise that `this` is an element reference and not the DOM element. Is there a way that we pass the DOM element reference instead?

So instead of
```js
await $('#foo').execute(elem => {
  console.log(elem) // (wrapped) webdriver element reference
})
```
We have
```js
await $('#foo').execute(elem => {
  console.log(elem) // DOM element reference
```

christian-bromann (approved):
LGTM 👍

Issue #8050, opened 2022-02-23
[💡 Feature]: allow in wdio-devtools-service coverageReporter multiple report types


bkrrrr:
### Is your feature request related to a problem?

Having multiple formats of coverage reports of the same test run is very common
  - HTML for test-writer
  - Text std as fast check for anyone
  - JSON for quality gates 

NYC can do this with multiple `---reporter` arguments [see](https://istanbul.js.org/docs/tutorials/mocha/)

But WDIO allows only one type and is not exposing the raw data, as NYC is doing it; in '.nyc_output'.

So there is no way to create another report.

### Describe the solution you'd like.

``` js
// wdio.conf.js
services: [
    ['devtools', {
        coverageReporter: {
            enable: true,
            type: [ 'html', 'text' ],  // => be able to declare this
            logDir: __dirname + '/coverage'
        }
    }]
]
```

### Describe alternatives you've considered.

Write the raw coverage results to .nyc_output. This would also enable to create multiple coverage reports. 

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

christian-bromann:
Thanks for reporting!

This sounds like a great addition to the service.

Any contributions that helps make this reality are highly appreciated. Please take a look into our [contribution guidelines](https://webdriver.io/docs/contribute) and let us know if you have any questions. Cheers!

cenfun:
I created a WebdriverIO service for coverage reports, see https://github.com/cenfun/wdio-monocart-service

christian-bromann:
Interesting 🤔 it seems like you are using the browser runner which offers istanbul based code coverage reporting as documented in [the docs](https://webdriver.io/docs/component-testing/coverage). This seems to be an interested approach to enable v8 coverage which might be a better alternative.

cenfun:
@christian-bromann Yes, it generates reports directly from native v8 coverage data. Since both Nodejs and Chromium-based browsers Chrome/Edge are native support V8 coverage data, I personally prefer V8 because of its high performance and there is no need to instrument the source code with Istanbul. And we can also revert to using Istanbul by setting `coverageProvider` to `istanbul`.
```js
// wdio.conf.js
{
    services: [
        ['monocart', {
            name: 'My WebdriverIO Coverage Report',
            // v8 (default) or istanbul (requires instrumenting source code)
            // coverageProvider: 'v8',
            entryFilter: '**/src/**',
            reports: [
                'v8',
                'console-details'
            ],
[...]

PR #12581, opened 2024-03-26
feat: v8 coverage

cenfun:
@christian-bromann As we discussed issue https://github.com/webdriverio/webdriverio/issues/8050, I tried to create a PR to implement V8 coverage.

However, the e2e test doesn't pass in my local environment (Windows 10 +Node 20) even though it doesn't make any changes. I can't fix it.

So based on this PR, could you help continue implementing V8 coverage?

- Istanbul (default)
```js
runner: ['browser', {
        coverage: {
          enabled: true
        }
}]
```
- V8 
```js
runner: ['browser', {
        coverage: {
          enabled: true,
          provider: 'v8'
        }
}]
```

### DIFF
--- a/packages/wdio-browser-runner/package.json
+++ b/packages/wdio-browser-runner/package.json
@@ -56,6 +56,7 @@
     "istanbul-reports": "^3.1.5",
     "mlly": "^1.4.0",
     "modern-node-polyfills": "^1.0.0",
+    "monocart-coverage-reports": "^2.7.7",
     "recast": "^0.23.2",
     "safe-stringify": "^1.1.0",
     "source-map-support": "^0.5.21",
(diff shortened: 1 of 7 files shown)


christian-bromann:
@cenfun thanks for raising the issue. I now realize that this requires CDP integration and will only work for Chromium based browser. In this case I would suggest to keep it as is which is: provide a common (instrumented) code coverage that works across all browser and offer a v8 based code coverage for Chromium based browser as a plugin/service.

What do you think? I would suggest to add a note to the docs pointing to your service for better visibility?

cenfun:
Yes, agree. V8 coverage is only for Chromium based browser. Feel free to close this PR.

christian-bromann:
> I would suggest to add a note to the docs pointing to your service for better visibility?

Would you be open raising a PR for this?

PR #12622, opened 2024-04-02
chore: add cli arg info

wswebcreation:
Make providing options through the cli more clear

image

### DIFF
--- a/website/docs/Frameworks.md
+++ b/website/docs/Frameworks.md
@@ -272,6 +272,20 @@ To get up and running quickly with Cucumber, have a look on our [`cucumber-boile
 
 The following options can be applied in your `wdio.conf.js` to configure your Cucumber environment using the `cucumberOpts` property:
 
+:::tip Adjusting options through the command line
+The `cucumberOpts`, such as custom `tags` for filtering tests, can be specified through the command line. This is accomplished by using the `cucumberOpts.{optionName}="value"` format.
+
+For example, if you want to run only the tests that are tagged with `@smoke`, you can use the following command:
+
+```sh
+# When you only want to run tests that hold the tag "@smoke"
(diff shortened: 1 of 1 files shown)


website/docs/Frameworks.md
```diff
@@ -272,6 +272,20 @@ To get up and running quickly with Cucumber, have a look on our [`cucumber-boile
...
 The following options can be applied in your `wdio.conf.js` to configure your Cucumber environment using the `cucumberOpts` property:
 
+:::tip Adjusting options through the command line
+The `cucumberOpts`, such as custom `tags` for filtering tests, can be specified through the command line. This is accomplished by using the `cucumberOpts.{optionName}="value"` format.
+
+For example, if you want to run only the tests that are tagged with `@smoke`, you can use the following command:
+
+```sh
+# When you only want to run tests that hold the tag "@smoke"
+npm run wdio -- --cucumberOpts.tags="@smoke"
```
christian-bromann, line 282: How about:
```suggestion
npx run wdio run wdio.conf.js --cucumberOpts.tags="@smoke"
```

You assume that the user has an NPM script called `wdio` which can be confusing if this is not quite the case, e.g. if the command is `wdio run wdio.conf.js && someOtherCommand`

erwinheitzman: Should be 'npx wdio' instead of 'npx run wdio' then

wswebcreation, line 282:
```suggestion
npx wdio ./wdio.conf.js --cucumberOpts.tags="@smoke"
```

wswebcreation: Thanks for the feedback, I've changed it to

`npx wdio ./wdio.conf.js --cucumberOpts.tags="@smoke"`

wswebcreation:
Assuming you guys have no other feedback and will merge this

PR #12666, opened 2024-04-10
(wdio-config): add outputDir default

erwinheitzman:
## Proposed changes

By default all the webdriver logs are put into the browser. This PR changes this so that the logs are now stored at `{wdio root}/.wdio-artifacts/logs`, making for a smoother experience and less confusion when people see errors that are not important to them (test passes but there is a webdriver error).

Marked as a breaking change because people could be relying on the logs in some ways that we cannot anticipate.

## Types of changes

## Checklist

### DIFF
--- a/packages/wdio-config/src/constants.ts
+++ b/packages/wdio-config/src/constants.ts
@@ -7,7 +7,7 @@ export const DEFAULT_CONFIGS: () => Omit<Options.Testrunner, 'capabilities'> = (
     specs: [],
     suites: {},
     exclude: [],
-    outputDir: './.wdio-artifacts/logs',
+    outputDir: '.wdio-artifacts/logs',
     logLevel: 'info' as const,
     logLevels: {},
     groupLogsByTestSpec: false,
--- a/packages/wdio-config/src/constants.ts
+++ b/packages/wdio-config/src/constants.ts
@@ -7,7 +7,7 @@ export const DEFAULT_CONFIGS: () => Omit<Options.Testrunner, 'capabilities'> = (
     specs: [],
     suites: {},
     exclude: [],
-    outputDir: undefined,
+    outputDir: './.wdio-artifacts/logs',
     logLevel: 'info' as const,
     logLevels: {},
     groupLogsByTestSpec: false,


packages/wdio-config/src/constants.ts
```diff
@@ -7,7 +7,7 @@ export const DEFAULT_CONFIGS: () => Omit<Options.Testrunner, 'capabilities'> = (
     specs: [],
     suites: {},
     exclude: [],
-    outputDir: undefined,
+    outputDir: '.wdio-artifacts/logs',
```
christian-bromann, line 10: I wonder if we should call the directory just `wdio` and maybe start recommending using this for all WebdriverIO related files, e.g. logs, baselines and config files

christian-bromann (commented):
wdyt?

packages/wdio-config/src/constants.ts
christian-bromann: This makes sense to me. I would have one request which would be to allow users to store wdio logs to stdout __and__ in stdout at the same time. AFAIK right now users can only decide between either of them and I think it makes sense to allow the scenario to make both possible.

Background: I personally like to have these WebDriver commands prinited in stdout so I can immediately debug them as my test go on and I print stuff in the console. If we set a default `outputDir` I would need to comment in/out the output dir option.

I think I would be fine if I can disable the `outputDir` option by just passing in a `--no-outputDir` but not sure if yargs supports recognises this and overwrites the hard coded property in the config with `undefined`. Alternatively I can imagine to pass in a `--stdout-logging` or just `--stdout` to enable both.

christian-bromann:
@erwinheitzman we are getting close to finalise `v9`, any updates on this effort? I think we can also bring this in as feature but the default that it always logs to stdout by default and only does both (file and stdout) logging if defined in the config. Wdyt?

erwinheitzman:
I can take a look at this soon 👍

christian-bromann:
I will move this out of the v9 release. I think it would be still great to get this landed some day as part of a minor release.

christian-bromann:
I will go ahead and close this. Let's raise an issue for this and discuss what we want to achieve. I feel like for folks to get started it makes sense to have them see what is going on. But maybe we can find an even more smarter solution. Thanks anyway!

PR #12845, opened 2024-05-08
Add resignApp capability type

dprevost-LMI:
## Proposed changes
Add [resignApp capability](https://www.browserstack.com/docs/app-automate/appium/advanced-features/disable-resigning-of-apps) type

We need to use this capability, but we stumble into typing issues.

## Types of changes

- [x] Polish (an improvement for an existing feature)
- [x] Bugfix (non-breaking change which fixes an issue)

## Checklist

[//]: # (_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._)

- [x] I have read the [CONTRIBUTING](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md) doc
- I have added tests that prove my fix is effective or that my feature works (N/A)

## Backport Request

- [x] Back-ported PR : https://github.com/webdriverio/webdriverio/pull/12851#issue-2288639253

PR #12851, opened 2024-05-09
Add resignApp capabilites v8

dprevost-LMI:
## Proposed changes

Same as https://github.com/webdriverio/webdriverio/pull/12845 but for v8


christian-bromann (approved):
👍

Issue #12781, opened 2024-05-03
[🐛 Bug]: isElementDisplayed returns incorrect result for elements with overflow-x/y: hidden and zero width/height


louisremi:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

latest

### Node.js Version

latest

### Mode

Standalone Mode

### What happened?

In 8.22.2, a regression was introduced in isDisplayed method. Webdriverio is no longer able to detect that an element isn't displayed with the following styles:
- `overflow-x: hidden; width: 0;`
- or `overflow-y: hidden; height: 0;`

It is however correctly able to detect that the element isn't displayed with the following styles:
- `overflow: hidden; width: 0; height: 0;`

### What is your expected behavior?

Webdriver should be consistently able to tell that elements with overflow the following styles are not displayed:
- `overflow: hidden; width: 0; height: 0;`
- `overflow-x: hidden; width: 0;`
- or `overflow-y: hidden; height: 0;`

### How to reproduce the bug.

I've started by forking cucumber-boilerplate and adding a test-suite for isDisplayed with different combinations of `overflow(-x/-y): hidden;` and zero width/height. Here are the steps to reproduce:

1. clone following repo locally https://github.com/louisremi/cucumber-boilerplate/tree/test-case/isElementDisplayed-overflow-bug:
  - `git clone git@github.com:louisremi/cucumber-boilerplate.git`
  - `git checkout test-case/isElementDisplayed-overflow-bug`
2. install dependencies 
  - `npm install`
3. run tests locally
  - `npm run test:features`

The two interesting files in this repo to understand the problem are: 
- https://github.com/louisremi/cucumber-boilerplate/blob/test-case/isElementDisplayed-overflow-bug/demo-app/index.html
- https://github.com/louisremi/cucumber-boilerplate/blob/test-case/isElementDisplayed-overflow-bug/src/features/notDisplayed.feature

### Relevant log output

```typescript
[chrome-headless-shell 124.0.6367.91 linux #0-0] Running: chrome-headless-shell (v124.0.6367.91) on linux
[chrome-headless-shell 124.0.6367.91 linux #0-0] Session ID: d17b349514a984dac88571ec59af8a9d
[chrome-headless-shell 124.0.6367.91 linux #0-0]
[...]

PR #12856, opened 2024-05-10
fix isDisplayed overflow-x/y bug

louisremi:
## Proposed changes

Fix regression inside isDisplay introduced with version 8.2.22: Webdriverio now incorrectly reports elements with style `overflow-x: hidden; width: 0` as displayed.
This fixes https://github.com/webdriverio/webdriverio/issues/12781

## Types of changes

## Checklist

- [ ] ~~I have added the necessary documentation (if appropriate)~~
- [ ] ~~I have added proper type definitions for new commands (if appropriate)~~

## Backport Request

## Further comments

This PR is a Work In Progress:
- [x] Add more tests for `opacity: 0`, `visibility: hidden` and `display: none` to the test suite
- [ ] Decide whether to keep fixture inside test file as Lit component, or to add them to guinea-pig
- [ ] Backport to v8


christian-bromann (commented):
Thanks for taking a stab at this. I have one question on the code changes. Regarding testing, you currently use the component test runner for this which is fine. An alternative is to make a PR to https://github.com/webdriverio/guinea-pig which is a simple website accessible at http://guinea-pig.webdriver.io/. This can be used to then to just keep this as e2e test, use the `url` command instead of `render` and move the test somewhere `e2e/wdio/headless/test.e2e.ts`.

e2e/browser-runner/__fixtures__/is-displayed.html
```diff
@@ -0,0 +1,36 @@
+<head>
```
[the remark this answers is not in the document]
louisremi: Alright, I removed that file

e2e/browser-runner/commands/element/isDisplayed.test.ts
```diff
@@ -0,0 +1,73 @@
...
+// import { screen } from '@testing-library/jest-dom'
+// import userEvent from '@testing-library/user-event'
+import { html, render } from 'lit'
+
+import * as matchers from '@testing-library/jest-dom/matchers'
+expect.extend(matchers)
+
+describe('isDisplayed', () => {
+    beforeEach(async () => {
+        // TODO: this fails with "Failed to load test page […]" error
```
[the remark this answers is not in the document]
louisremi: This is no longer a problem, I got rid of the fixture file. I render a document with Lit for now, and I can submit a PR to guinea pig if you prefer

e2e/package.json
```diff
@@ -5,6 +5,9 @@
   "private": true,
   "scripts": {
     "test:browser": "run-s test:browser:*",
+    "test:browser:isDisplayed": "run-s test:browser:isDisplayed:*",
+    "test:browser:isDisplayed:wdio": "cross-env WDIO_PRESET=isDisplayed node ../packages/wdio-cli/bin/wdio.js ./browser-runner/wdio.conf.js --spec isDisplayed.test.ts",
+    "test:browser:isDisplayed:verifyCoverage": "run-s verifyCoverage",
```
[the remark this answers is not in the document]
louisremi: I got rid of this line

packages/webdriverio/src/scripts/isElementDisplayed.ts
```diff
@@ -266,11 +238,11 @@ export default function isElementDisplayed (element: Element): boolean {
         return false
     }
 
-    if (!elementSubtreeHasNonZeroDimensions(element)) {
+    if (!elementSubtreeHasNonZeroDimension({ element, dimension: 'x' })) {
         return false
     }
 
-    if (isElementSubtreeHiddenByOverflow(element) && !elementHasBoundingBox(element)) {
+    if (!elementSubtreeHasNonZeroDimension({ element, dimension: 'y' })) {
```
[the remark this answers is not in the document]
louisremi: Yes, there are various tests in the test-suite I created that check the result of only using `overflow` instead of `overflow-x/y`

packages/webdriverio/src/scripts/isElementDisplayed.ts
```diff
@@ -114,25 +114,38 @@ export default function isElementDisplayed (element: Element): boolean {
         return cascadedStylePropertyForElement(parentElement, property)
     }
 
-    function elementHasBoundingBox(element: Element): boolean {
+    function elementSubtreeHasNonZeroDimension({
```
erwinheitzman, line 117: It's better to check for Zero dimensions than for non Zero dimensions for the function. Names like these are confusing especially when executed like so !nonZero

packages/webdriverio/src/scripts/isElementDisplayed.ts
```diff
@@ -114,25 +114,38 @@ export default function isElementDisplayed (element: Element): boolean {
...
         }
 
         // Paths can have a zero width or height. Treat them as shown if the stroke width is positive.
-        const boundingBox = element.getBoundingClientRect()
-        if (element.tagName.toUpperCase() === 'PATH' && boundingBox.width + boundingBox.height > 0) {
-            const strokeWidth = cascadedStylePropertyForElement(element, 'stroke-width')
-            return !!strokeWidth && (parseInt(strokeWidth, 10) > 0)
+        if (
+            element.tagName.toUpperCase() === 'PATH' &&
+            (boundingBox.width > 0 || boundingBox.height > 0)
```
erwinheitzman, line 132: Shouldn't this be && instead of ||?

packages/webdriverio/src/scripts/isElementDisplayed.ts
```diff
@@ -144,58 +157,16 @@ export default function isElementDisplayed (element: Element): boolean {
...
-        if (!elementOverflowsContainer(element)) {
-            return false
-        }
-
-        if (!element.childNodes.length) {
-            return false
-        }
-
-        // This element's subtree is hidden by overflow if all child subtrees are as well.
-        return [].every.call(element.childNodes, function (childNode: Element) {
```
erwinheitzman, line 180: We lose this functionality by removing this, the purpose of this is to check if any child elements of an element are not displayed. We need to keep this unless I missed something

erwinheitzman (commented):
Although I like the changes to support the mentioned surface, I do not feel confident in merging this. Especially since changing this code potentially has A LOT of impact on our users, so getting it right is very important. I would like to run this on a big codebase first before merging this. Thank you very much for looking into this and contributing, great find overall 🙂

packages/webdriverio/src/scripts/isElementDisplayed.ts
louisremi: This replaces `boundingBox.width + boundingBox.height > 0` for clarity, so `||` is appropriate

packages/webdriverio/src/scripts/isElementDisplayed.ts
louisremi: The diff makes it seem that this is new code, while it isn't: this is an existing function that I modified to check dimensions one by one instead. 
The function is used recursively. I agree that at the top level it would be better to write `elementSubtreeHasZeroDimension()" instead of "!elementSubtreeHasNonZeroDimension()`. But in the recursive algorithm, the original name makes more sense. We could write a new `const elementSubtreeHasZeroDimension = (...args) => !elemelementSubtreeHasNonZeroDimension(...args)` helper, but I feel that would be a bit overkill

packages/webdriverio/src/scripts/isElementDisplayed.ts
louisremi: After a careful analysis, it seemed that this test `isElementSubtreeHiddenByOverflow` was redundant with the test `elementSubtreeHasNonZeroDimensions`. 
I'm not sure why those two different tests existed in the first place, but `isElementSubtreeHiddenByOverflow` seemed like an inferior version (not testing for path with > 0 stroke, weird recursive implementation). So I got rid of it.
But I agree this change should be carefully tested on a large code-base before being validated.

erwinheitzman:
@louisremi I would really appreciate it if you could do that but I do not expect you to go that far (perhaps nice to make sure the fix works on your end though).

My plan is to run this against all the tests of the client I currently work for. It's scale is huge as well and it should give us a good indication of any impact on existing testcases. I do need to find time for it though

christian-bromann:
@louisremi @erwinheitzman any updates on this one?

christian-bromann:
Closing due to inactivity.

christian-bromann:
@erwinheitzman did you had a chance to test this in a larger code base?

christian-bromann:
Closing in favor of #14075

PR #12929, opened 2024-05-27
Ws/ocr-v9

wswebcreation:
## Proposed changes

This PR merges the V8 OCR docs into main.

There was also an issue regarding generating the V9 docs. I needed to remove the `babel` id from the `Configuration` category because it was missing. Is this intentionally?


christian-bromann (approved):
Awesome job 👍

PR #12957, opened 2024-06-03
Allows reloadSession to connect to remote driver

ccharnkij:
## Proposed changes

This change allows reloadSession to connect to remote driver and vice versa. This is useful for some instances, such as when a test starts in mobile native app and then have to validate data in web.

## Types of changes

## Checklist

### DIFF
--- a/packages/webdriver/src/index.ts
+++ b/packages/webdriver/src/index.ts
@@ -4,7 +4,6 @@ import logger from '@wdio/logger'
 
 import { webdriverMonad, sessionEnvironmentDetector, startWebDriver } from '@wdio/utils'
 import { validateConfig } from '@wdio/config'
-import { deepmerge } from 'deepmerge-ts'
 import type { Options } from '@wdio/types'
 
 import command from './command.js'
@@ -152,11 +151,18 @@ export default class WebDriver {
      * @returns {string}           the new session id of the browser
      */
     static async reloadSession(instance: Client, newCapabilities?: WebdriverIO.Capabilities) {
-        const capabilities = deepmerge(instance.requestedCapabilities, newCapabilities || {})
-        const params: Options.WebDriver = { ...instance.options, capabilities }
+        const capabilities = newCapabilities ? newCapabilities : instance.requestedCapabilities as WebdriverIO.Capabilities
+        let params: Options.WebDriver = { ...instance.options, capabilities }
+
+        for (const prop of ['protocol', 'hostname', 'port', 'path', 'queryParams', 'user', 'key'] as (keyof Options.Connection)[]) {
+            if (prop in capabilities) {
+                params = { ...params, [prop]: capabilities[prop] }
+                delete capabilities[prop]
+            }
+        }
 
         let driverProcess: ChildProcess | undefined
-        if (newCapabilities?.browserName) {
+        if (params.hostname === DEFAULTS.hostname.default && newCapabilities?.browserName) {
             delete params.port
             delete params.hostname
             driverProcess = await startWebDriver(params)
--- a/packages/webdriverio/src/commands/browser/reloadSession.ts
+++ b/packages/webdriverio/src/commands/browser/reloadSession.ts
@@ -11,6 +11,14 @@ const log = logger('webdriverio')
  * Be careful though, this command affects your test time tremendously since spawning
  * new Selenium sessions is very time consuming especially when using cloud services.
  *
+ * Connection parameters such as hostname, port, protocol, etc. can be added along side
+ * browserName when you want to connect to a different remote service. This is useful
+ * in a situation, for example, where you start a test in native app and need to verify
+ * data in web app.
+ *
+ * If you start from remote service, you can pass in 0.0.0.0 for hostname if you want
+ * to switch to local drivers.
+ *
  * <example>
     :reloadSync.js
     it('should reload my session with current capabilities', async () => {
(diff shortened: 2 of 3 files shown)


ccharnkij:
@christian-bromann Yeah, here is a sample one: https://github.com/ccharnkij/v8-reload-example. When you run npm test, you should see `Error: Invalid or unsupported WebDriver capabilities found ("hostname", "port").`

Side note, I'm still not sure why github action would fail on bootstrap part. I have not touched @wdio/types.

ccharnkij:
Hi @christian-bromann Have you a chance to check this out?

packages/webdriver/src/index.ts
```diff
@@ -152,11 +151,11 @@ export default class WebDriver {
      * @returns {string}           the new session id of the browser
      */
     static async reloadSession(instance: Client, newCapabilities?: WebdriverIO.Capabilities) {
-        const capabilities = deepmerge(instance.requestedCapabilities, newCapabilities || {})
-        const params: Options.WebDriver = { ...instance.options, capabilities }
+        const capabilities = { ...sanitizeCaps(newCapabilities || {}, true), capabilities: sanitizeCaps(newCapabilities || instance.requestedCapabilities) }
+        const params: Options.WebDriver = { ...instance.options, ...capabilities }
```
christian-bromann, line 155: This would merge capabilities into the `params` object which should be of type `Options.WebDriver`, can you explain your intention doing this? It feels like this is not what we want, right?

christian-bromann (changes requested):
Sorry for the delayed review, I had a lot of competing priorities and only got a chance now.

packages/webdriver/src/index.ts
ccharnkij: Yeah, the merging is what I aimed for. Because of this error `Error: Invalid or unsupported WebDriver capabilities found ("hostname", "port").`, I need a way to separate out fields like host, port, etc.. from an object that is passed into reloadFunction and use them to overwrite the old values existed in `instance.options`. So, assuming you pass something like this in reloadFunction,
```
{
host: 'xxxx',
port: 1234,
browserName: 'chrome'
}
```
That top line would make `capabilities` field look like this
```
host: 'xxxx',
port: 1234,
capabilities: {
    browserName: 'chrome'
}
```

And I can just use this to overwrite the corresponding value in `instance.options`.

Not sure if I answer your question, but let me know.

christian-bromann: I would suggest to just have some simple if statements and move the connection parameters into the params object, e.g.:

```ts
for (const prop of ['protocol', 'hostname', 'port', 'path'] as Option.Connection) {
    if (prop in capabilities) {
        params[prop] = capabilities[prop];
        delete capabilities[prop];
    }
}
```

ccharnkij: better?

packages/webdriver/src/index.ts
```diff
@@ -154,9 +154,15 @@ export default class WebDriver {
     static async reloadSession(instance: Client, newCapabilities?: WebdriverIO.Capabilities) {
         const capabilities = deepmerge(instance.requestedCapabilities, newCapabilities || {})
         const params: Options.WebDriver = { ...instance.options, capabilities }
+        for (const prop of ['protocol', 'hostname', 'port', 'path'] as string[]) {
```
christian-bromann, line 157: let's better type this so we don't need to use `any`
```suggestion
        for (const prop of ['protocol', 'hostname', 'port', 'path'] as keyof Option.Connection) {
```

christian-bromann (commented):
One more comment, can we also add some docs in the `reloadSession` command for this?

packages/webdriver/src/index.ts
ccharnkij: How about this one?

ccharnkij:
If all is well, let me know if you'd like me to make the same change to main too.

christian-bromann:
Thanks @ccharnkij , on the documentation side, mind adding a sentence to the command docs, mentioning in which situations people may want to reload a session with different capabilities?

ccharnkij:
like this?

packages/webdriverio/src/commands/browser/reloadSession.ts
```diff
@@ -11,6 +11,14 @@ const log = logger('webdriverio')
  * Be careful though, this command affects your test time tremendously since spawning
  * new Selenium sessions is very time consuming especially when using cloud services.
  *
+ * Connection parameters such as hostname, port, protocol, etc. can be added along side
+ * browserName when you want to connect to a different remote service. This is useful
+ * in a situation, for example, where you start a test in native app and need to verify
+ * data in web app.
```
christian-bromann, line 17: Just for understanding: this would completely get rid of the mobile session so you couldn't continue anything on the native app side, correct?

ccharnkij: Yeah, I tested it out locally. This is not meant to be for concurrent session like multi-remote, if that's what you're asking. It's meant to be used within a single session. Btw, I made a small change again, I removed deepmerge because I don't want caps from original session to be mixed with a new one.

christian-bromann (approved):
LGTM 👍

christian-bromann:
Mind raising a PR for `main`?

ccharnkij:
Will do. Thank you

PR #13045, opened 2024-06-17
Add logFilters appium server args type

dprevost-LMI:
## Proposed changes

Add type for log-filters appium server args. See https://appium.io/docs/en/2.9/guides/log-filters/

## Types of changes

- [x] Polish (an improvement for an existing feature)

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md) doc

## Backport Request

- [x] Back-ported PR at https://github.com/webdriverio/webdriverio/pull/13046 (coming)

PR #13046, opened 2024-06-17
Add logFilter appium server arguments type v8

dprevost-LMI:
## Proposed changes

Same as https://github.com/webdriverio/webdriverio/pull/13045, but for v8


christian-bromann (approved):
👍

Issue #13217, opened 2024-07-26
[💡 Feature]: Multiple parallel executions reusing the same node process


alfonso-presa:
**Is your feature request related to a problem?**

Indeed. There's a limitation on how much concurrent sessions a single machine is able to handle because WDIO spawns new nodejs process for every parallel executions that wants to be handled. Each of those processes takes (depending on the services used of course) a minimum of 100MB o memory, which means that executing for example 40 concurrent "hello world" tests can take up to 4GB of memory. When you have TS transpilation and a big code base this situation can be a lot worst (nowadays with 16GB in our systems be can barely afford executing 40 concurrent tests).

**Describe the solution you'd like.**

By taking a look at the code of `@wdio/local-runner`, `@wdio/globals` and the cucumber-framework service (which is the one we are currently using) it looked possible to leverage Proxies and AsyncLocalStorage to keep the global variables while still using a single nodejs process.

As PoC, I've been able to run 10 tests in parallel in the same NodeJS VM by modifying `@wdio/local-runner` (worker.js and run.js) `@wdio/globals` (index.js) and `@wdio/cucumber-framework` (index.js to reuse the supportCodeLibraryBuilder result for steps to be avail). The executions reach correctly to the end but got stucked after the reporting phase for some reason. I guess reporting frameworks are going to be another challenge.

**Describe alternatives you've considered.**

Using worker threads sounded like a possibililty but they still take almost the same amount of memory.

**Additional context**

This was discussed some years ago in [this issue](https://github.com/webdriverio/webdriverio/issues/4235) and some optimizations were done.... but we are still consuming a lot of infrastructure in our CI/CD environments to drive tests in concurrency.

**Code of Conduct**

- []  I agree to follow this project's Code of Conduct

christian-bromann:
Hey @alfonso-presa , thanks for raising the issue!

I am open for optimizations here. Our current setup is probably legacy by now as it hasn't changed for a long time. We definitely can improve a lot here by using modern primitives e.g. using Node.js [`vm`](https://nodejs.org/api/vm.html) primitives instead of spinning up whole processes.

If you can help us lead this initiatives we would love to connect. Note that we can pay out to contributors through our OpenCollective fund.

alfonso-presa:
Hi Christian, 

Thanks for pointing out the VM primitives alternative. Is sounds like a really interesting one.

[...]

PR #13220, opened 2024-07-28
[WIP] Feature/multiple paralle tests in single process

alfonso-presa:
## Proposed changes

Fixes https://github.com/webdriverio/webdriverio/issues/13217 by using AsyncLocalStorage and proxies to enable the use of a single process for all parallel test executions, saving tons of memory and also speeding thigs up (though, it will be single threaded).

Patching the cucumber-framework was needed in order to reuse the step libraries loaded... the solution is not beautifull at all...

I've created a new cucumber-lite-local example demonstating that this change works.

It's still WIP as REPL is not working right now, and I guess some test coverage would be needed for the new files at the local-runner.

## Types of changes

## Checklist

## Backport Request

I would like this to be backported to v8.... and v7, which is the one we currently rely on.

### DIFF
--- a/packages/wdio-local-runner/src/liteWorker.ts
+++ b/packages/wdio-local-runner/src/liteWorker.ts
@@ -1,0 +1,235 @@
+import path from 'node:path'
+import { EventEmitter } from 'node:events'
+import type { WritableStreamBuffer } from 'stream-buffers'
+import type { Options, Workers } from '@wdio/types'
+
+import logger from '@wdio/logger'
+
+import runnerTransformStream from './transformStream.js'
+import ReplQueue from './replQueue.js'
+import RunnerStream from './stdStream.js'
+import type { IsolatedProcess } from './runLite.js'
+import { run } from './runLite.js'
+
+const log = logger('@wdio/lite-runner')
+const replQueue = new ReplQueue()
+const ACCEPTABLE_BUSY_COMMANDS = ['workerRequest', 'endSession']
+
+const stdOutStream = new RunnerStream()
+const stdErrStream = new RunnerStream()
+stdOutStream.pipe(process.stdout)
+stdErrStream.pipe(process.stderr)
+
+/**
+ * WorkerInstance
+ * responsible for spawning a sub process to run the framework in and handle its
+ * session lifetime.
+ */
+export default class LiteWorkerInstance extends EventEmitter implements Workers.Worker {
+    cid: string
+    config: Options.Testrunner
+    configFile: string
+    // requestedCapabilities
+    caps: WebdriverIO.Capabilities
+    // actual capabilities returned by driver
+    capabilities: WebdriverIO.Capabilities
+    specs: string[]
+    retries: number
+    stdout: WritableStreamBuffer
+    stderr: WritableStreamBuffer
+    child?: IsolatedProcess
+    sessionId?: string
+    server?: Record<string, any>
+    logsAggregator: string[] = []
+
+    instances?: Record<string, { sessionId: string }>
+    isMultiremote?: boolean
+
+    isBusy = false
+    isKilled = false
+    isSetup: Promise<boolean>
+    isReadyResolver: (value: boolean | PromiseLike<boolean>) => void = () => {}
+    isSetupResolver: (value: boolean | PromiseLike<boolean>) => void = () => {}
+
+    /**
+     * assigns paramters to scope of instance
+     * @param  {object}   config      parsed configuration object
+     * @param  {string}   cid         capability id (e.g. 0-1)
+     * @param  {string}   configFile  path to config file (for sub process to parse)
+     * @param  {object}   caps        capability object
+     * @param  {string[]} specs       list of paths to test files to run in this worker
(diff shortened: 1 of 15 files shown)


christian-bromann:
@alfonso-presa please note that we are currently working on fixing the build due to some issues we have identified within the last changes. So don't worry about it when working on this branch.

packages/wdio-local-runner/src/liteWorker.ts
```diff
@@ -0,0 +1,235 @@
...
+const stdErrStream = new RunnerStream()
+stdOutStream.pipe(process.stdout)
+stdErrStream.pipe(process.stderr)
+
+/**
+ * WorkerInstance
+ * responsible for spawning a sub process to run the framework in and handle its
+ * session lifetime.
+ */
+export default class LiteWorkerInstance extends EventEmitter implements Workers.Worker {
```
christian-bromann, line 28: Do you think we can extend from the default worker here? It seems like a lot of logic is the same?

packages/wdio-local-runner/src/processProxy.ts
```diff
@@ -0,0 +1,19 @@
+import { AsyncLocalStorage } from 'node:async_hooks'
+
+const originalProcess = process
+const processLocals = new AsyncLocalStorage()
+export const patchedProcess = new Proxy(
```
christian-bromann, line 5: Can you explain your intention with this?

tests/smoke.runner.js
```diff
@@ -140,11 +140,11 @@ const jasmineTestrunner = async () => {
...
-            'expect(object).toHaveTitle(object)',
-            'expect(object).toHaveTitle(object)',
-            'expect(object).toHaveUrl(object)',
-            'expect(object).toHaveUrl(object)',
-            'expect(string).toHaveTitle(object)',
+            'expect(object).toHaveTitle(function)',
+            'expect(object).toHaveTitle(function)',
+            'expect(object).toHaveUrl(function)',
+            'expect(object).toHaveUrl(function)',
+            'expect(string).toHaveTitle(function)',
```
christian-bromann, line 147: Why are these assertions changing?

packages/wdio-local-runner/src/liteWorker.ts
alfonso-presa: I can try :-)

tests/smoke.runner.js
alfonso-presa: To be honest... I'm not sure. I've to investigate it.

packages/wdio-local-runner/src/processProxy.ts
alfonso-presa: The main purpose is to be able to "intercept" some methods and properties of the process object like `env`, `send`, `stdout` ... so that while using the same process each "thread" has it's own values. It's being used in packages/wdio-local-runner/src/runLite.ts:56. This way there are other places in the wdio code base (like in the runner where process.send) which are not impacted by this feature.

christian-bromann:
> Test / E2E Tests / Testrunner Tests (windows-latest) (pull_request) Failing after 2m
Test / E2E Tests / Testrunner Tests (macos-13) (pull_request) Failing after 3m

If these two are failing you are good to go. There is a known issue were we do element lookups using Bidi in iframes which is not supported, we are addressing it in #13223

alfonso-presa:
@christian-bromann . Thanks a lot for the feedback! I guess the TODOs to get this to a mergeable state are:

- [ ] Attemtp to extend the Worker
- [ ] Add REPL debug support
- [ ] Add tests to the added clases
- [ ] Review reason for changing assertions from object to function

I'll try to work on them ASAP.

tests/smoke.runner.js
alfonso-presa: @christian-bromann I now have an explanation for this. I've made it so that `global.browser` and `browser` included by importing/requiring from `@wdio/globals` are consistent. If you take a look at this test in main stream https://github.com/webdriverio/webdriverio/blob/main/packages/wdio-globals/tests/index.test.ts#L6, you'll ralise that browser is a function from the import perspective. I can make it look like and `object` or like a `function`, but not both of them at the same time :-).

alfonso-presa:
@christian-bromann I think this is getting close to a merge candidate... I've added some testing. It only support cucumber. I've tried to add jasime support but even if it sounds possible, reporters get a lot messy, so I hope cucumber is enough to give this feature a first try.

alfonso-presa:
@christian-bromann I've added some extra test coverage. From my perspective this is ready to be merged if you find it convenient. What do you think?

christian-bromann:
@alfonso-presa thank you, please give me a couple of days to get the chance to test run this. Cheers!

christian-bromann:
@alfonso-presa sorry about the delay on this. I haven't been able to properly test your changes but it is on my radar once I get some fixes in for v9.

alfonso-presa:
@christian-bromann No problem at all. Keep it up! Thanks!

kerkeniakov:
@christian-bromann any plans to merge this?

PR #13223, opened 2024-07-29
fix(webdriverio): don't support deep piercing in iframes

christian-bromann:
## Proposed changes

It doesn't seem to be currently supported to locate nodes using Bidi primitives when in an iframe.

## Types of changes

## Checklist

## Backport Request

## Further comments

### DIFF
--- a/e2e/attach.test.ts
+++ b/e2e/attach.test.ts
@@ -10,32 +10,6 @@ import { test, expect } from 'vitest'
 import { remote, attach } from '../packages/webdriverio/build/index.js'
 
 test('allow to attach to an existing session', async () => {
-    const browser = await remote({
-        capabilities: {
-            browserName: 'chrome',
-            'goog:chromeOptions': {
-                args: ['headless', 'disable-gpu']
-            }
-        }
-    })
-
-    await browser.url('http://guinea-pig.webdriver.io')
-    expect(await browser.getTitle()).toBe('WebdriverJS Testpage')
-
-    const otherBrowser = await attach(browser)
-    expect(await otherBrowser.getTitle()).toBe('WebdriverJS Testpage')
-
-    await otherBrowser.deleteSession()
-
-    /**
-     * verify that browser session is deleted
-     */
-    const error = await browser.status().catch((err) => err)
-    expect(error.message).not.toBe('ChromeDriver ready for new sessions.')
-    expect(error.message).toEqual(expect.stringContaining('fetch failed'))
-})
-
-test('can attach to a Bidi session', async () => {
     /**
      * fails in windows due to timeout:
      * > Command browsingContext.navigate with id 1 (with the following parameter: {"context":"BD746B5679530BC3403539C2FEC5A45A","url":"http://guinea-pig.webdriver.io","wait":"interactive"}) timed out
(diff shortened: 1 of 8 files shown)


danielhjacobs:
> It doesn't seem to be currently supported to locate nodes using Bidi primitives when in an iframe.

So how can we locate nodes within an iframe/frame with v9?

christian-bromann:
> So how can we locate nodes within an iframe/frame with v9?

Theoretically yes, the problem is that while we can do look-ups in `iframe`s and we can switch frames before interacting with elements, we have the limitation that this would only work if you interact with a single element, e.g.:

```ts
// works
$('element-within-iframe').getText()
// fails
$$('element-within-iframe').map((el) => el.getText())
```

Hence we decided to not support this yet until we have a solution to this.

danielhjacobs:
Isn't this a regression from v8: see https://github.com/webdriverio/webdriverio/issues/13384 ?

danielhjacobs:
In other words, is there a way I can rewrite https://github.com/ruffle-rs/ruffle/blob/master/web/packages/selfhosted/test/polyfill/classic_frames_injected/test.ts and similar tests?

danielhjacobs:
So for https://github.com/ruffle-rs/ruffle/blob/master/web/packages/selfhosted/test/polyfill/iframes_provided/test.ts, I had an idea to work around this and it seems close to working but I'm having trouble finishing supporting it. Do you know what I'm doing wrong?

```diff
-        await browser.$("ruffle-object").waitForExist();
+        await browser
+            .$(() =>
+                document
+                    .querySelector("#test-frame")
+                    .contentDocument.body.querySelector("ruffle-object"),
+            )
+            .waitForExist();
-        const actual = await browser
-            .$("#test-container")
+        const actual = await browser
+            .$(() =>
+                document
+                    .querySelector("#test-frame")
+                    .contentDocument.body.querySelector(
+                        "#test-container",
+                    ),
+            )
             .getHTML({ includeSelectorTag: false, pierceShadowRoot: false });

```I get an error `TypeError: $3 is not a function` from the `const actual = await browser` line. It's specifically related to the await. If I do the same line as `const promise = browser`, I don't get any error, but of course then `expect(promise).html.to.equal(expected);` says that the promise doesn't equal the string. It doesn't matter where I put the `await` or if I use `resolve`, no matter what I get the TypeError wherever I put the await or resolve.

christian-bromann:
> Isn't this a regression from v8: see #13384 ?

I haven't looked into this bug but no, v8 didn't support iframe lookups, you have to switch into the frame first.

Chances are that there are some regression in regards to using a function as selector, I will have to look into that. Any contributions are appreciated.

danielhjacobs:
I must have misunderstood. We weren't using a function as a selector before though. I switched to that because the CSS selector of ruffle-object within the iframe/frame wasn't working even after switching into the frame in v9, even though it worked in v8.

danielhjacobs:
Maybe I misunderstand, but whether intentional or not, isn't this condition true when using a string CSS selector after switching into an iframe? 

`this.isBidi`: True if there is WebDriver Bidi support

`typeof selector === 'string'`: True for "ruffle-object" as the selector

`selector.startsWith(DEEP_SELECTOR)`: False because "ruffle-object" doesn't start with ">>>"

`shadowRootManager.isWithinFrame(`: True because we are inside a frame so the shadow root is too.

christian-bromann:
Let's continue this conversation in #13384 or create a new issue if this is different.

PR #13523, opened 2024-09-06
Fix  customElement wrapper for custom elements which don't define connectedCallback or disconnectedCallback

swendlandt:
## Proposed changes
Fixes https://github.com/webdriverio/webdriverio/issues/13524

Starting with v9.0.6 webdriver.io has a bug which causes tests to fail when a custom element doesn't define a connectedCallback or a disconnectedCallback.
The reason is with https://github.com/webdriverio/webdriverio/pull/13430 there is no check anymore if the original function is defined before calling it.

## Types of changes

## Checklist

## Backport Request

## Further comments

### DIFF
--- a/packages/webdriverio/src/scripts/customElement.ts
+++ b/packages/webdriverio/src/scripts/customElement.ts
@@ -17,13 +17,13 @@ export default function customElementWrapper () {
                 parentNode = parentNode.parentNode
             }
             console.debug('[WDIO]', 'newShadowRoot', this, parentNode, parentNode === document)
-            return origConnectedCallback && origConnectedCallback.call(this)
+            return origConnectedCallback?.call(this)
         }
 
         const origDisconnectedCallback = Constructor.prototype.disconnectedCallback
         Constructor.prototype.disconnectedCallback = function(this: HTMLElement) {
             console.debug('[WDIO]', 'removeShadowRoot', this)
-            return origDisconnectedCallback && origDisconnectedCallback.call(this)
+            return origDisconnectedCallback?.call(this)
         }
         return origFn(name, Constructor, options)
     }
--- a/packages/webdriverio/src/scripts/customElement.ts
+++ b/packages/webdriverio/src/scripts/customElement.ts
@@ -17,13 +17,13 @@ export default function customElementWrapper () {
                 parentNode = parentNode.parentNode
             }
             console.debug('[WDIO]', 'newShadowRoot', this, parentNode, parentNode === document)
-            return origConnectedCallback.call(this)
+            return origConnectedCallback && origConnectedCallback.call(this)
         }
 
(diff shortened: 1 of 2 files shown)


swendlandt:
@christian-bromann can you please take a look at this PR?

christian-bromann (approved):
Nice catch 👍

christian-bromann:
Mind signing the CLA?

packages/webdriverio/src/scripts/customElement.ts
```diff
@@ -17,13 +17,13 @@ export default function customElementWrapper () {
                 parentNode = parentNode.parentNode
             }
             console.debug('[WDIO]', 'newShadowRoot', this, parentNode, parentNode === document)
-            return origConnectedCallback.call(this)
+            return origConnectedCallback && origConnectedCallback.call(this)
```
erwinheitzman, line 20: let's use modern js syntax
```suggestion
            return origConnectedCallback.call?.(this)
```

packages/webdriverio/src/scripts/customElement.ts
```diff
@@ -17,13 +17,13 @@ export default function customElementWrapper () {
...
         }
 
         const origDisconnectedCallback = Constructor.prototype.disconnectedCallback
         Constructor.prototype.disconnectedCallback = function(this: HTMLElement) {
             console.debug('[WDIO]', 'removeShadowRoot', this)
-            return origDisconnectedCallback.call(this)
+            return origDisconnectedCallback && origDisconnectedCallback.call(this)
```
erwinheitzman, line 26: let's use modern js syntax
```suggestion
            return origDisconnectedCallback.call?.(this)
```

erwinheitzman (changes requested):
Nice change thank you.

Let's use modern js syntax for this though :)

erwinheitzman:
@christian-bromann I just noticed you already approved, what do you think?

packages/webdriverio/src/scripts/customElement.ts
swendlandt: I'm fine with using the `?` operator, but it would have to be after origConnectedCallback in that case:
`return origConnectedCallback?.call(this)`

packages/webdriverio/src/scripts/customElement.ts
swendlandt: Same as above would need to be changed to 
`return origConnectedCallback?.call(this)`

christian-bromann:
What @swendlandt suggests: `return origConnectedCallback?.call(this)` makes sense to me.

christian-bromann:
@swendlandt mind signing the CLA?

swendlandt:
> @swendlandt mind signing the CLA?

Done

Issue #13560, opened 2024-09-13
[🐛 Bug]: All tests are run in all workers


aleks-temp:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

8.40

### Node.js Version

20.15.1

### Mode

WDIO Testrunner

### Which capabilities are you using?

```typescript
maxInstances: instances,
    browserName: 'chrome',
    'goog:chromeOptions': {
      args: [
        '--headless=new',
        '--disable-dev-shm-usage',
        '--disable-gpu',
        '--window-size=1920,1080',
        '--disable-web-security',
        '--no-sandbox',
        '--ignore-certificate-errors',
      ],
    }
```

### What happened?

I am trying to run multiple tests in parallel (all in his own worker) with the following command:

npx wdio wdio.conf.js  --spec=src/features/test1.feature:3 src/features/test2.feature:3

And the result is that I have 2 workers up (which is OK), but both tests are ran in all 2 workers. If there are 3 tests, there will be 3 workers and all 3 tests will ran in the 3 workers e.t.c.

### What is your expected behavior?

I expect (as it was in wdio 7) a test per worker to be ran

### How to reproduce the bug.

Here is the repo with a sample project: https://github.com/aleks-temp/BugSample

Run command npx wdio wdio.conf.js  --spec=src/features/test1.feature:3 src/features/test2.feature:3

### Relevant log output

```typescript
[chrome 128.0.6613.138 windows #0-0] Running: chrome (v128.0.6613.138) on windows
[chrome 128.0.6613.138 windows #0-0] Session ID: c8157f84cadbfdc476e458a0b9a68729
[chrome 128.0.6613.138 windows #0-0]
[chrome 128.0.6613.138 windows #0-0] » \src\features\test1.feature
[chrome 128.0.6613.138 windows #0-0] Dummy feature
[chrome 128.0.6613.138 windows #0-0] Dummy
[chrome 128.0.6613.138 windows #0-0]    ✓ Given some test step
[chrome 128.0.6613.138 windows #0-0]
[chrome 128.0.6613.138 windows #0-0] » \src\features\test2.feature
[chrome 128.0.6613.138 windows #0-0] Dummy feature 2
[chrome 128.0.6613.138 windows #0-0] Dummy2
[...]

PR #13564, opened 2024-09-13
fix(wdio-cucumber-framework): filter cucumberFeaturesWithLineNumbers

johnp:
## Proposed changes

Fixes https://github.com/webdriverio/webdriverio/issues/13560. It looks like the `cucumberFeaturesWithLineNumbers` filtering broke some time ago, possibly with https://github.com/webdriverio/webdriverio/pull/11010.

While `_specs` has already been filtered here, `cucumberFeaturesWithLineNumbers`, which just exists to retain the line numbers, still contains all specs as passed on the cmdline or in the config. Therefore, just passing this to cucumber will inevitably execute all specs/features in every worker. To fix this, we only retain those entries within `cucumberFeaturesWithLineNumbers`, for which we can find a fitting entry in the list of already filtered `_spec`s.

Note that `cucumberFeaturesWithLineNumbers` does not really mean "cucumber features with line numbers". It seems to actually mean: "all passed specs, when at least one of them has a line number, or whatever was placed into the `cucumberFeaturesWithLineNumbers` wdio config option."

## Types of changes

## Checklist

    - [x] Test code which verifies that the correct feature file is executed (fails before the fix) 
    - [x] Test code which verifies that only the correct scenarios are executed

## Backport Request

- [x] Back-ported PR at #13563

## Further comments

Not 100% sure filtering with `startsWith` works 100% of the time (not sure if the paths could, for example, be relative, in one array, and absolute in the other, but they seem to always be absolute in my testing so far). **edit: now with `path.resolve`**

I wish Cucumber's `loadSources` would just take take whole features instead of just `paths`, but apparently it doesn't.

I'm not entirely sure if the `cucumberFeaturesWithLineNumbers` wdio config option is normalized anywhere. I don't know why it even exists, given that you can just put the features into `specs`, where the line numbers will get filtered out automatically.... It feels like `cucumberFeaturesWithLineNumbers` ought to be internal only.

### DIFF
--- a/infra/compiler/src/index.ts
+++ b/infra/compiler/src/index.ts
@@ -1,4 +1,5 @@
 import fs from 'node:fs/promises'
+import { existsSync } from 'node:fs'
 import path from 'node:path'
 import url from 'node:url'
 import { parseArgs } from 'node:util'
@@ -33,7 +34,8 @@ const { values } = parseArgs({ args, options })
  */
 const projects = (await fs.readdir(
     path.resolve(rootDir, 'packages')
-)).filter((dir) => dir !== 'node_modules')
+    // ignore potential stale/empty package directories left over when switching between branches
+)).filter((dir) => dir !== 'node_modules' && existsSync(path.resolve(rootDir, 'packages', dir, 'package.json')))
 
 const packages = (
     await Promise.all(
--- a/tests/cucumber/features/features.given.js
+++ b/tests/cucumber/features/features.given.js
@@ -1,5 +1,5 @@
 import { browser } from '@wdio/globals'
-import { Given } from '../../../packages/wdio-cucumber-framework/build/index.js'
+import { Given } from '@wdio/cucumber-framework'
 
 Given('test1', () => {
     browser.assertExecutingFeatureFileOnce('test1.feature')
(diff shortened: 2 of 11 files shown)


infra/compiler/src/index.ts
```diff
@@ -33,7 +34,7 @@ const { values } = parseArgs({ args, options })
  */
 const projects = (await fs.readdir(
     path.resolve(rootDir, 'packages')
-)).filter((dir) => dir !== 'node_modules')
+)).filter((dir) => dir !== 'node_modules' && existsSync(path.resolve(rootDir, 'packages', dir, 'package.json')))
```
christian-bromann, line 37: What is the origin of this change?

tests/cucumber/features/features.given.js
```diff
@@ -0,0 +1,14 @@
+import { browser } from '@wdio/globals'
+import { Given } from '../../../packages/wdio-cucumber-framework/build/index.js'
```
christian-bromann, line 2: we use pnpm so we can just reference this as normal package
```suggestion
import { Given } from '@wdio/cucumber-framework'
```

christian-bromann (approved):
LGTM 👍 

Mind applying the same changes to the v8 branch?

christian-bromann:
@johannespfrang I've been out on vacation for the last 3 weeks and am planning to release v7, v8 and v9 next week. We have others that could do releases as well but they don't have the context to all the changes which is why we prefer to have me around for them.

aleks-temp:
And beside that, on version 9 when running from my local machine there is no problem with the multiple reruns, but there is the following issue on click:
WebDriver Bidi command "script.callFunction" failed with error: invalid argument - Invalid input in "arguments"/0

Issue #13554, opened 2024-09-11
[🐛 Bug]: Incorrect file type in specs when using cucumber framework


Chubok-bit:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

9.0.8

### Node.js Version

20.9.0

### Mode

Standalone Mode

### Which capabilities are you using?

```typescript
capabilities: [{
        browserName: 'MicrosoftEdge'
    }],
```

### What happened?

When starting a brand new project with the Cucumber framework (? Which framework do you want to use? Cucumber (https://cucumber.io/)) the specs property is not set correctly. When you attempt to run the tests you get the following error:

2024-09-11T13:33:46.462Z WARN @wdio/config:ConfigParser: pattern ./test/specs/**/*.ts did not match any file
2024-09-11T13:33:46.463Z ERROR @wdio/cli:launcher: No specs found to run, exiting with failure

### What is your expected behavior?

    The specs property for the example project needs to change from:
    
    specs: [
        './test/specs/**/*.ts'
    ],
    
    to: 
    
       specs: [
        './test/specs/**/*.featur'
    ],

### How to reproduce the bug.

1. Create a new wdio v9 project with the following settings:
    
? A project named "v9" was detected at "C:\Work\v9", correct? yes
? What type of testing would you like to do? E2E Testing - of Web or Mobile Applications
? Where is your automation backend located? On my local machine
? Which environment you would like to automate? Web - web applications in the browser
? With which browser should we start? Microsoft Edge
? Which framework do you want to use? Cucumber (https://cucumber.io/)
? Do you want to use Typescript to write tests? yes
? Do you want WebdriverIO to autogenerate some test files? yes
? What should be the location of your spec files? C:\Work\v9Testing\test\specs\**\*.ts
? What should be the location of your step definitions? C:\Work\v9Testing\features\step-definitions\steps.ts
? Do you want to use page objects (https://martinfowler.com/bliki/PageObject.html)? yes
? Where are your page objects located? C:\Work\v9Testing\features\pageobjects\**\*.ts
? Which reporter do you want to use? spec
? Do you want to add a plugin to your test setup?
? Would you like to include Visual Testing to your setup? For more information see https://webdriver.io/docs/visual-testing! no
? Do you want to add a service to your test setup?
[...]

PR #13566, opened 2024-09-14
fix(@wdio/cli): cucumber generated config spec path

alcpereira:
## Proposed changes

Solves #13554

After poking around for a few hours, I think that #13348 is to blame.
This comes down to:
```js
{ 
  message: 'What should be the location of your spec files?', // Note: this is NOT the cucumber block
  when: (answers: Questionnair) => answers.generateTestFiles && answers.framework.match(/(mocha|jasmine)/)
  // `answers.framework.match(/(mocha|jasmine)/)` -> returns null
  // `when: () => null` is considered not false, thus true
}
```  
This solution is a quick fix: casting to a boolean, so the `null` becomes `false`.

A few questions:
- Should Dependabot config be reviewed? Automerging + Bumping majors sounds a bit risky to me.
- Should I implement Unit tests? It doesn't seem trivial (no similar test), should I do it in another PR? It also feels that by doing so, we would be testing `Inquirer` instead of the CLI business logic.
- Should we remove the `// @ts-expect-error` [here](https://github.com/webdriverio/webdriverio/blob/1ecf61063a7b16dce5b399eda5a335716149d4af/packages/wdio-cli/src/utils.ts#L568)? I think this bug should not have been caught by a unit test but by the type checking. [Inquirer is expecting a boolean](https://github.com/SBoudrias/Inquirer.js/blob/d2690cc552f6451df30d62b6ca419ead3cea1c9c/packages/inquirer/src/types.mts#L58). This also looks like a big task 😅 , let me know if this is critical for a product pov.

## Types of changes

## Checklist

## Backport Request

## Further comments

### DIFF
--- a/packages/wdio-cli/src/constants.ts
+++ b/packages/wdio-cli/src/constants.ts
@@ -504,7 +504,7 @@ export const QUESTIONNAIRE = [{
         const pattern = isBrowserRunner(answers) ? 'src/**/*.test' : 'test/specs/**/*'
         return getDefaultFiles(answers, pattern)
     },
-    when: /* istanbul ignore next */ (answers: Questionnair) => answers.generateTestFiles && answers.framework.match(/(mocha|jasmine)/)
+    when: /* istanbul ignore next */ (answers: Questionnair) => answers.generateTestFiles && !!answers.framework.match(/(mocha|jasmine)/)
 }, {
     type: 'input',
     name: 'specs',
--- a/packages/wdio-cli/src/constants.ts
+++ b/packages/wdio-cli/src/constants.ts
@@ -504,7 +504,7 @@ export const QUESTIONNAIRE = [{
         const pattern = isBrowserRunner(answers) ? 'src/**/*.test' : 'test/specs/**/*'
         return getDefaultFiles(answers, pattern)
     },
-    when: /* istanbul ignore next */ (answers: Questionnair) => answers.generateTestFiles && !!answers.framework.match(/(mocha|jasmine)/)
+    when: /* istanbul ignore next */ (answers: Questionnair) => answers.generateTestFiles && Boolean(answers.framework.match(/(mocha|jasmine)/))
 }, {
     type: 'input',
     name: 'specs',


christian-bromann (changes requested):
I am not sure if this is the right solution here. We want this question to be asked if the framework contains `mocha` or `jasmine`. Here is the important part:

```ts
{
    type: 'input',
    name: 'specs',
    message: 'What should be the location of your spec files?',
    default: /* istanbul ignore next */ (answers: Questionnair) => {
        const pattern = isBrowserRunner(answers) ? 'src/**/*.test' : 'test/specs/**/*'
        return getDefaultFiles(answers, pattern)
    },
    when: /* istanbul ignore next */ (answers: Questionnair) => answers.generateTestFiles && answers.framework.match(/(mocha|jasmine)/)
}, {
    type: 'input',
    name: 'specs',
    message: 'What should be the location of your feature files?',
    default: (answers: Questionnair) => getDefaultFiles(answers, 'features/**/*.feature'),
    when: /* istanbul ignore next */ (answers: Questionnair) => answers.generateTestFiles && answers.framework.includes('cucumber')
}, {
    type: 'input',
    name: 'stepDefinitions',
    message: 'What should be the location of your step definitions?',
    default: (answers: Questionnair) => getDefaultFiles(answers, 'features/step-definitions/steps'),
    when: /* istanbul ignore next */ (answers: Questionnair) => answers.generateTestFiles && answers.framework.includes('cucumber')
}
```

For some reason, even though we select Cucumber, it asks whether a user wants to select spec file paths instead of feature file paths.

alcpereira:
Sorry, maybe it was not clear in the description. 

> For some reason, even though we select Cucumber, it asks whether a user wants to select spec file paths instead of feature file paths.

The reason is that in Inquirer the `when` logic for prompt has changed in Inquirer `10.2.2`, see this [PR](https://github.com/SBoudrias/Inquirer.js/pull/1541/files).

I've created a minimal reproduction with the different versions:
- [10.1.8 working](https://stackblitz.com/edit/stackblitz-starters-wf7ut5?file=index.js)
- [10.2.2 NOT working](https://stackblitz.com/edit/stackblitz-starters-hhisj3?file=index.js)
- [9.3.6 working](https://stackblitz.com/edit/stackblitz-starters-6kxj1e?file=index.js)

Try running the `pnpm start` with Mocha, Jasmine and Cucumber answers.

> Please ignore the crash at the end of the prompt for versions 10.x, this is an issue with StackBlitz and Inquirer (see [here](https://github.com/SBoudrias/Inquirer.js/issues/1426)).

I don't think this is something we should report in Inquirer, as we are incorrectly returning `null` when the API is requiring a boolean.
Thus my solution to coerce `String.match()` to a boolean and my suggestion to introduce the official types for the question to avoid any other potential bugs.

packages/wdio-cli/src/constants.ts
```diff
@@ -504,7 +504,7 @@ export const QUESTIONNAIRE = [{
         const pattern = isBrowserRunner(answers) ? 'src/**/*.test' : 'test/specs/**/*'
         return getDefaultFiles(answers, pattern)
     },
-    when: /* istanbul ignore next */ (answers: Questionnair) => answers.generateTestFiles && answers.framework.match(/(mocha|jasmine)/)
+    when: /* istanbul ignore next */ (answers: Questionnair) => answers.generateTestFiles && !!answers.framework.match(/(mocha|jasmine)/)
```
christian-bromann, line 507:
```suggestion
    when: /* istanbul ignore next */ (answers: Questionnair) => answers.generateTestFiles && Boolean(answers.framework.match(/(mocha|jasmine)/))
```

christian-bromann (approved):
Let's do it a bit more implicit then. Thanks for the fix, makes total sense!

PR #13728, opened 2024-10-02
fix(webdriverio): implement switchFrame for Bidi

christian-bromann:
## Proposed changes

This is a new approach to handle frame switches with Bidi. I want to introduce a new command `switchFrame` (aligned with other commands like `switchWindow`) that simplifies the way user can switch into frames.

Command docs:

Switches the active context to a frame, e.g. an iframe on the page. There are multiple ways you can query a frame
on the page:

  - If given a string it switches to the frame with a matching context id, url or url that contains that string
    ```ts
    // switch to a frame that has the url "https://the-internet.herokuapp.com/iframe"
    await browser.switchFrame('https://the-internet.herokuapp.com/iframe')
    // switch to a frame that has a certain context id
    await browser.switchFrame('A5734774C41F8C91D483BDD4022B2EF3')
    ```

  - If given a WebdriverIO element that references an `iframe` element it will switch to that frame
    ```ts
    // switch to a frame element queried from current context
    await browser.switchFrame($('iframe'))
    ```

  - If given a function it will loop through all iframes on the page and call the function with the context
    object. The function should return a boolean indicating if the frame should be selected. Within the function
    you can use WebdriverIO commands to e.g. check if a certain frame element exists.
    ```ts
    // switch to first frame that contains an element with id "#frameContent"
    await browser.switchFrame(() => $('#frameContent').isExisting())
    ```

  - If given `null` it will switch to the top level frame
    ```ts
    // first switch into a frame
    await browser.switchFrame($('iframe'))
    // do more automation within that frame, then ...

    // switch to the top level frame
    await browser.switchFrame(null)
    ```

## Types of changes

## Checklist

## Backport Request

## Further comments


packages/webdriverio/src/commands/browser/switchFrame.ts
```diff
@@ -0,0 +1,217 @@
...
+            ...(context.children || []).map(mapContext)
+        ]
+        const sessionContext = getContextManager(this)
+        const currentContext = await sessionContext.getCurrentContext()
+
+        /**
+         * transform context tree into a flat list of context objects with references
+         * to children
+         */
+        const allContexts: Record<string, FlatContextTree> = tree.contexts.map(mapContext).flat(Infinity)
```
erwinheitzman, line 106: what does the .flat(Infinite) do here? haven't seen this before

packages/webdriverio/src/commands/browser/switchFrame.ts
```diff
@@ -0,0 +1,217 @@
...
+    url: string,
+    contexts: local.BrowsingContextInfoList,
+    matcher: typeof byUrl | typeof byUrlContaining
+): string | undefined {
+    for (const context of contexts) {
+        if (matcher(context, url)) {
+            return context.context
+        }
+
+        if (context.children && context.children.length > 0) {
```
erwinheitzman, line 184: not sure if the ? needs to be added between context and children or between children and length or both but it's a bit cleaner
```suggestion
        if (context.children?.length > 0) {
```

packages/webdriverio/src/commands/browser/switchFrame.ts
```diff
@@ -0,0 +1,217 @@
...
+}
+
+/**
+ * Given a certain `BrowsingContextInfoList` and a desired `contextId`, this helper finds
+ * the `BrowsingContextInfo` that matches the `contextId`
+ * @param tree  a context tree returned from `browsingContextGetTree`
+ * @param contextId  a desired context id
+ * @returns the `BrowsingContextInfo` that matches the `contextId` or `undefined` if not found
+ */
+function findContextById (tree: local.BrowsingContextInfoList, contextId: string): local.BrowsingContextInfo | undefined {
```
erwinheitzman, line 202: same function as above, could potentially be merged

erwinheitzman (approved):
Added some suggestions but nothing major, looking great Christian. Well done, love how clean and readable the code is!

packages/webdriverio/src/commands/browser/switchFrame.ts
christian-bromann: `.flat()` by default only flattens with a depth of `1`, so:

```ts
[1, [2, [3]]].flat() === [1, 2, [3]]
```

With infinity we flatten the array infinity times:

```ts
[1, [2, [3]]].flat() === [1, 2, 3]
```

packages/webdriverio/src/commands/browser/switchFrame.ts
christian-bromann: Good call!

danielhjacobs:
This doesn't appear to work for classic frame elements.

PR #13777, opened 2024-10-18
fix(webdriverio): wait for request information to come in

christian-bromann:
## Proposed changes

If the `wait` property for the `url` command is `none` then the returning request object may be undefined if the `browsingContext.navigationStarted` hasn't come in yet. This patch waits until we have the request to come in.

## Types of changes

## Checklist

## Backport Request

## Further comments

n/a

### DIFF
--- a/packages/webdriverio/src/networkManager.ts
+++ b/packages/webdriverio/src/networkManager.ts
@@ -14,6 +14,7 @@ export function getNetworkManager(browser: WebdriverIO.Browser) {
 }
 
 type Context = string
+const UNKNOWN_NAVIGATION_ID = 'UNKNOWN_NAVIGATION_ID'
 
 /**
  * This class is responsible for managing shadow roots and their elements.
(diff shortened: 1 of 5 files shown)


packages/webdriverio/src/networkManager.ts
```diff
@@ -143,8 +147,27 @@ export class NetworkManager {
...
+
+        /**
+         * Chrome v127 and below does not support yet navigation ids, hence we have to
+         * just pick the first request object and assume it is the root request. This
+         * will break if the user operates on multiple windows at the same time.
+         *
+         * @see https://github.com/GoogleChromeLabs/chromium-bidi/issues/1054
+         */
+        const firstRequest = this.#requests.values().next().value
+        return this.#lastNetworkId
```
erwinheitzman, line 164: will this not cause issues on other versions? just double-checking :)

erwinheitzman (approved):
approved but I did have one question

packages/webdriverio/src/networkManager.ts
christian-bromann: I will raise a PR next week to run these bidi tests also in Firefox. Right now they only run in Chrome where the navigation id is not yet available. I am not to what you are referring to when saying "other versions". The way it is implemented right now is broken (uses context id instead of navigation id).

PR #13800, opened 2024-10-25
reorder shadowroot removal

ccharnkij:
## Proposed changes

Fix for #13521. So this happens due to the sequence in which shadowRoots are removed. Currently, shadowRoots that are duplicate can be added to the children set and are removed in a first-in-first-out fashion. In the reproducible, the shadowRoot needed to find the element is the child of the first shadowRoot added in the children set. This is followed by a duplicate and a removeShadowRoot entry, which fetches and removes that first child.

For example,
```
root
root.addShadowElement( child('1') )
root.addShadowElement( '1', child('2') )
root.addShadowElement( '1', child('2') ) <-- now there are 2 child 2
root.addShadowElement( '2', child('3') ) <-- child 3 attached to the first child 2
root.remove( '2' ) <-- remove the first child 2 which has child 3 attached and any element that requires child 3 can't be found

```

I fix this by changing the order when removing and make it remove the last one first. I test this with the reproducible example and it works. However, I'm not experienced in how bidi receives shadowRoots and not sure if the current way is the intended way (maybe should have discuss with you first on discord). Let me know how it goes, and if this is not valid, please close this.

## Types of changes

## Checklist

### DIFF
--- a/packages/webdriverio/src/shadowRoot.ts
+++ b/packages/webdriverio/src/shadowRoot.ts
@@ -339,12 +339,13 @@ export class ShadowRootTree {
     }
 
     remove (element: string): boolean {
-        for (const child of this.children) {
-            if (child.element === element) {
-                return this.children.delete(child)
+        const childArray = Array.from(this.children)
+        for (let i = childArray.length - 1; i >= 0; i--) {
+            if (childArray[i].element === element) {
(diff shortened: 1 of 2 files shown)


christian-bromann (approved):
Love this ❤️ thanks so much!

PR #13870, opened 2024-11-09
feat: enhance scrollIntoView to support native mobile

wswebcreation:
Initial setup to support Native apps

## Proposed changes

This PR enhances the `scrollIntoView` to support native mobile apps by using the native `mobile:execute` methods of the drivers

## Types of changes

## Checklist

## Backport Request

## Further comments


wswebcreation:
Hmm, getting the following error when I try to test the command

```logs
Can't call scrollIntoView on element with selector "//android.widget.TextView[@resource-id="android:id/title" and @text="System"]" because element wasn't found
```

For native the elements are not always available as with web. I might have found a hack I'm not proud of. The [`implicitWait`](https://github.com/webdriverio/webdriverio/blob/main/packages/webdriverio/src/utils/implicitWait.ts#L11) excludes checking for the `*waitUntil*` methods. If I just "hack" the `scrollIntoView` into it, it works, it will not find the element, it will scroll and check if it exists, if not, it will scroll, and so on. 
This is a "hacky" solution, so I would love to get some input/look into a specific direction to make this more stable. I'm open to suggestions 😅

packages/webdriverio/src/commands/element/scrollIntoView.ts
```diff
@@ -2,9 +2,127 @@ import logger from '@wdio/logger'
 import { ELEMENT_KEY } from 'webdriver'
 
 import { getBrowserObject } from '@wdio/utils'
+import type { CustomScrollIntoViewOptions, MobileScrollIntoViewOptions } from 'src/types.js'
```
christian-bromann, line 5:
```suggestion
import type { CustomScrollIntoViewOptions, MobileScrollIntoViewOptions } from '../../types.js'
```

packages/webdriverio/src/commands/element/scrollIntoView.ts
```diff
@@ -2,9 +2,127 @@ import logger from '@wdio/logger'
...
 
 const log = logger('webdriverio')
 
+type MobileScrollUntilVisibleOptions = {
```
christian-bromann, line 9: Can we move the code down, below the command function? Having other types or functions defined before just invites to add comments to them which will ultimately replace the command documentation when generating the docs

packages/webdriverio/src/commands/element/scrollIntoView.ts
```diff
@@ -2,9 +2,127 @@ import logger from '@wdio/logger'
...
+    const scrollableElements = (await browser.$$(
+        selector
+    )) as unknown as WebdriverIO.Element[]
+
+    if (scrollableElements.length > 0) {
+        return scrollableElements[0]
+    }
+
+    throw new Error(
+        `Default scrollable element '${browser.isIOS ? defaultIosSelector : defaultAndroidSelector}' not found.`
```
christian-bromann, line 44: What should the user do in this case? Is this a common scenario? If so, we should add a recommendation on what a user should try next.

packages/webdriverio/src/commands/element/scrollIntoView.ts
```diff
@@ -2,9 +2,127 @@ import logger from '@wdio/logger'
...
+    element,
+    options
+}: {
+    browser: WebdriverIO.Browser,
+    element: WebdriverIO.Element,
+    options: MobileScrollIntoViewOptions
+    }) {
+    const defaultOptions = {
+        maxScrolls: 10,
+        scrollDirection: MobileScrollDirection.Down,
```
christian-bromann, line 100: What if we already scrolled all the way down and now have to do scroll up to find the element?

christian-bromann:
> If I just "hack" the `scrollIntoView` into it, it works, it will not find the element, it will scroll and check if it exists, if not, it will scroll, and so on.

It seem fine to me if we extend the if statement with scrollIntoView if we are in a mobile and native context, e.g.

```ts
const browser = getBrowserObject(currentElement)
const shallSkipForMobileScrollIntoView = browser.isMobile && (await browser.getContext === 'NATIVE')
```

packages/webdriverio/src/commands/element/scrollIntoView.ts
wswebcreation: Thanks, accepted it

packages/webdriverio/src/commands/element/scrollIntoView.ts
wswebcreation: sure, I added it to the top due to coding conventions that we first need to declare before we use

packages/webdriverio/src/commands/element/scrollIntoView.ts
wswebcreation: Good suggestion, I'll update the error

packages/webdriverio/src/commands/element/scrollIntoView.ts
wswebcreation: I'll update the error message if we can't find an element. I believe that users need to know their app in which direction to swipe to find the element. Sadly, native app automation is not that easy in comparison to web 😅

wswebcreation:
> > If I just "hack" the `scrollIntoView` into it, it works, it will not find the element, it will scroll and check if it exists, if not, it will scroll, and so on.
> 
> It seem fine to me if we extend the if statement with scrollIntoView if we are in a mobile and native context, e.g.
> 
> ```ts
> const browser = getBrowserObject(currentElement)
> const shallSkipForMobileScrollIntoView = browser.isMobile && (await browser.getContext === 'NATIVE')
> ```

Thanks, that's a solution. If we have the `isNativeContext` we could even determine better. There's only one downside, and that is that will always do an extra webdriver call for mobile, no matter if it's web or native app

christian-bromann:
> There's only one downside, and that is that will always do an extra webdriver call for mobile, no matter if it's web or native app

How about enhancing the context manager the way we discussed so the extra call is obsolete?

wswebcreation:
> > There's only one downside, and that is that will always do an extra webdriver call for mobile, no matter if it's web or native app
> 
> 
> 
> How about enhancing the context manager the way we discussed so the extra call is obsolete?

That's also what I wanted to do, but in a separate PR which would "revert" the extra call. I wanted to create "smaller" PRs because I don't have a lot of time the coming weeks

wswebcreation:
@christian-bromann 

Not sure why other UTs are failing, I can't see any impact due to my changes (to the fetch Mock). Any ideas?

christian-bromann (approved):
LGTM 👍

Issue #13937, opened 2024-12-01
[💡 Feature]: Proper masking for Reporters and other tools using `onBeforeComand` & `onAfterCommand`


dprevost-LMI:
### Is your feature request related to a problem?

Sometimes, even when using test data, we want to prevent some information from being outputted in logs or reports. Today, in wdio, no configuration allows such behaviour. Therefore, we need a proper masking mechanism. The focus here is mainly on the reporters.

The requests were triggered primarily to mask the test user's password from the report. This password is sent using the `elementSendKeys` command, and the text value can be found in the `body.text` property.

The idea is to implement a mechanism that will prevent reporters from accessing sensitive data. To achieve that, the `onBeforeComand` and `onAfterCommand` are the central points for communicating the text of the sendElementKey command to reporters. Therefore, masking the value at those moments seems a solid solution.

While being on the topic, possible other touchpoints to keep in mind:
- appium have [log filters](https://appium.io/docs/en/2.0/guides/log-filters/) even though it is hard to use.
- BrowserStack has [maskCommand](https://www.browserstack.com/docs/automate/selenium/hide-sensitive-data) even though it's an all-or-nothing. But, it might not mask information in Percy or Insight tools.

### Describe the solution you'd like.

Since the primary focus is masking data in reports, the `onBeforeComand` and `onAfterCommand` commands sound like central places to consider removing sensitive data so that anyone listening there cannot retrieve it.

All reporters go through those hooks, and even the BrowserStack Insight & Percy tools do. So, masking on those hooks sounds like a solid solution.

As pointed out for appium, they have the [log filters](https://appium.io/docs/en/2.0/guides/log-filters/), and since I suspect (like me) have or must use this solution, I inspired mine from that, so it would be easier to configure in both places.

So, the complete idea is to allow the configuration of a Regex similar to the one used in Appium. If the `body.text` of the command matches the RegEx, we replace the content with the string `**SECURE**`, which is the same default value as in the appium solution.

Finally, the solution is scoped to provide a first iteration and is not intended to be a complete configuration or cover every possible need.

### Describe alternatives you've considered.

I searched for alternatives, and nothing seemed aligned or appropriate for my case.

### Additional context

This PR, provide an idea of the solution presented.
I wanted to make it work with capabilities, but I need to figure out how to access them from the code.

I tested it against my project, and we can see the value being masked in the allure reporter.
[...]

PR #13938, opened 2024-12-01
feat(runner+browserstack): Mask sensitive data for Reporters (and more)

dprevost-LMI:
## Proposed changes

To provide easier security, here is a PR which has two main goals:
1. On the `element.setValue`, add a new boolean mask option allowing to mask from everyone besides appium the value set in an input element. This allows with the simple `{mask : true}` to hide the password, for example, of a real production user from wdio logs, any reporters, the performance tool and anyone listening on any `emit`. The cherry on top is the participation of appium's folks, where even in Appium's logs (see [PR](https://github.com/appium/appium/pull/21118)), this value continues to be masked with the simple boolean options. 🎉 
   - Note: Some limitations exist
        1. Appium plugins, which we do not have control over.
        2. Only `setValue` and `addValue` commands are supported. Not the getValue, for example.

3. A regex expression can now be configured in the wdio logs to hide more sensitive information. For example, when using BrowserStack, the user's key is in plain text. Now we can pass a simple regex pattern like `maskingPattern: '--key=([^ ]*)'`, which will mask this value in the logs.

See the [issue](https://github.com/webdriverio/webdriverio/issues/13937) for the history of the discussion that led to the PR decisions. 

Example for case 1:
Logs:
423178061-29c8d9bc-43a3-4187-b699-4dc16f18da0d

Allure-reporter
423178068-e0bd30a2-90b8-4ae3-a2bf-9ae0a0357d0d

Example for case 2:
423178049-d7a20a80-e3eb-4f81-a82e-47befcfeb0bf

## Types of changes

- ~Polish (an improvement for an existing feature)~
-  ~Bugfix (non-breaking change which fixes an issue)~
- ~Breaking change (fix or feature that would cause existing functionality to not work as expected)~
- [X] Documentation update (improvements to the project's docs)
- ~Specification changes (updates to WebDriver command specifications)~
- ~Internal updates (everything related to internal scripts, governance documentation and CI files)~

## Checklist

## Backport Request

- [X] This change is solely for `v9` and doesn't need to be back-ported
- ~Back-ported PR at `#XXXXX`~

## Further comments

Related to issue https://github.com/webdriverio/webdriverio/issues/13937
[Appium-related issue](https://github.com/appium/appium/issues/20844)

### DIFF
--- a/packages/wdio-logger/src/index.ts
+++ b/packages/wdio-logger/src/index.ts
@@ -226,12 +226,12 @@ getLogger.setLogLevelsConfig = (logLevels: Record<string, log.LogLevelDesc> = {}
         loggers[logName].setLevel(logLevel)
     })
 }
-getLogger.setMaskingPatterns = (maskingPatternsStringPerLoggerName: Record<string, string | undefined> = {}, wdioDefault = undefined) => {
+getLogger.setMaskingPatterns = (maskingPatternsStringPerLoggerName: Record<string, string | undefined> = {}, wdioDefault:string|undefined = undefined) => {
 
     /**
      * set the environment variable for masking patterns
      */
-    if (!process.env.WDIO_LOG_MASKING_PATTERNS) {
+    if (process.env.WDIO_LOG_MASKING_PATTERNS === undefined) {
         process.env.WDIO_LOG_MASKING_PATTERNS = wdioDefault
     }
 
@@ -250,7 +250,7 @@ getLogger.setMaskingPatterns = (maskingPatternsStringPerLoggerName: Record<strin
         /**
          * either apply maskingPatterns or use the global one
          */
-        const maskingPatterns = maskingPatternsConfig[logName] ?? parseMaskingPatterns(process.env.WDIO_LOG_LEVEL)
+        const maskingPatterns = maskingPatternsConfig[logName] ?? parseMaskingPatterns(process.env.WDIO_LOG_MASKING_PATTERNS)
 
         loggers[logName].maskingPatterns = maskingPatterns
     })
--- a/packages/wdio-logger/src/index.ts
+++ b/packages/wdio-logger/src/index.ts
@@ -1,3 +1,5 @@
+export { SENSITIVE_DATA_REPLACER } from './utils.js'
+
 import fs from 'node:fs'
 import util from 'node:util'
 
@@ -6,6 +8,7 @@ import type { ColorName } from 'chalk'
 import chalk from 'chalk'
 import prefix from 'loglevel-plugin-prefix'
 import ansiStrip from 'strip-ansi'
+import { mask, parseMaskingPatterns } from './utils.js'
 
 prefix.reg(log)
 
@@ -33,29 +36,31 @@ const SERIALIZERS = [{
     /**
      * display error stack
      */
-    matches: (err: unknown) => err instanceof Error,
+    matches: (err: unknown): err is Error => err instanceof Error,
     serialize: (err: Error) => err.stack
 }, {
     /**
      * color commands blue
      */
-    matches: (log: string) => log === matches.COMMAND || log === matches.BIDICOMMAND,
+    matches: (log: unknown): log is string => log === matches.COMMAND || log === matches.BIDICOMMAND,
     serialize: (log: string) => chalk.magenta(log)
 }, {
     /**
      * color data yellow
      */
-    matches: (log: string) => log === matches.DATA,
+    matches: (log: unknown): log is string => log === matches.DATA,
     serialize: (log: string) => chalk.yellow(log)
 }, {
     /**
      * color result cyan
      */
-    matches: (log: string) => log === matches.RESULT || log === matches.BIDIRESULT,
+    matches: (log: unknown): log is string => log === matches.RESULT || log === matches.BIDIRESULT,
     serialize: (log: string) => chalk.cyan(log)
 }]
 
 interface LoggerInterface extends log.Logger {
+    setMaskingPatternsConfig: () => RegExp[] | undefined
+    maskingPatterns: RegExp[] | undefined
     progress(...msg: string[]): void;
 }
 
@@ -65,13 +70,14 @@ interface Loggers {
 
 const loggers: Loggers = log.getLoggers() as Loggers
 let logLevelsConfig: Record<string, log.LogLevelDesc> = {}
+let maskingPatternsConfig: Record<string, RegExp[] | undefined> = {}
 const logCache = new Set()
 let logFile: fs.WriteStream | null
 
 const originalFactory = log.methodFactory
-const wdioLoggerMethodFactory = function (this: log.Logger, methodName: log.LogLevelNames, logLevel: log.LogLevelNumbers, loggerName: string) {
+const wdioLoggerMethodFactory = (wdioLogger: LoggerInterface) => function (this: log.Logger, methodName: log.LogLevelNames, logLevel: log.LogLevelNumbers, loggerName: string) {
     const rawMethod = originalFactory(methodName, logLevel, loggerName)
-    return (...args: string[]) => {
+    return (...args: [string, string?, ...unknown[]]) => {
         /**
          * create logFile lazily
          */
@@ -80,23 +86,23 @@ const wdioLoggerMethodFactory = function (this: log.Logger, methodName: log.LogL
         }
 
         /**
-         * split `prefixer: value` sting to `prefixer: ` and `value`
+         * split `prefixer: value` string to `prefixer: ` and `value`
          * so that SERIALIZERS can match certain string
          */
         const match = Object.values(matches).filter(x => args[0].endsWith(`: ${x}`))[0]
         if (match) {
-            const prefixStr = args.shift()!.slice(0, -match.length - 1)
+            const prefixStr = (args.shift() as string[]).slice(0, -match.length - 1)
             args.unshift(prefixStr, match)
         }
 
         args = args.map((arg) => {
             for (const s of SERIALIZERS) {
                 if (s.matches(arg)) {
-                    return s.serialize(arg as Error & string) as string
+                    return mask(s.serialize(arg as Error & string), wdioLogger.maskingPatterns)
                 }
             }
-            return arg
-        })
+            return mask(arg, wdioLogger.maskingPatterns)
+        }) as [string, string?, ...unknown[]]
 
         const logText = ansiStrip(`${util.format.apply(this, args as [format: string, ...params: string[]])}\n`)
         if (logFile && logFile.writable) {
@@ -151,16 +157,21 @@ export default function getLogger (name: string) {
     }
 
     loggers[name] = log.getLogger(name) as LoggerInterface
-    loggers[name].setLevel(logLevel)
-    loggers[name].methodFactory = wdioLoggerMethodFactory
-    loggers[name].progress = progress
-    prefix.apply(loggers[name], {
+    const logger = loggers[name]
+
+    logger.setLevel(logLevel)
+    logger.maskingPatterns = maskingPatternsConfig[name] ?? parseMaskingPatterns(process.env.WDIO_LOG_MASKING_PATTERNS)
+    logger.progress = progress
+
+    logger.methodFactory = wdioLoggerMethodFactory(logger)
+
+    prefix.apply(logger, {
         template: '%t %l %n:',
         timestampFormatter: (date) => chalk.gray(date.toISOString()),
         levelFormatter: (level: string) => chalk[COLORS[level]](level.toUpperCase()),
         nameFormatter: (name) => chalk.whiteBright(name)
     })
-    return loggers[name]
+    return logger
 }
 /**
  * Wait for writable stream to be flushed.
@@ -213,6 +224,36 @@ getLogger.setLogLevelsConfig = (logLevels: Record<string, log.LogLevelDesc> = {}
         const logLevel = typeof logLevelsConfig[logLevelName] !== 'undefined' ? logLevelsConfig[logLevelName] : process.env.WDIO_LOG_LEVEL as log.LogLevelDesc
 
         loggers[logName].setLevel(logLevel)
+    })
+}
+getLogger.setMaskingPatterns = (maskingPatternsStringPerLoggerName: Record<string, string | undefined> = {}, wdioDefault = undefined) => {
+
+    /**
+     * set the environment variable for masking patterns
+     */
+    if (process.env.WDIO_LOG_MASKING_PATTERNS === undefined) {
+        process.env.WDIO_LOG_MASKING_PATTERNS = wdioDefault
+    }
+
+    maskingPatternsConfig = {}
+
+    /**
+     * build maskingPatternsConfig object
+     */
+    Object.entries(maskingPatternsStringPerLoggerName).forEach(([logName, maskingPatternsString]) => {
+        maskingPatternsConfig[logName] = parseMaskingPatterns(maskingPatternsString)
+    })
+
+    /**
+     * set masking patterns for each logger
+     */
+    Object.keys(loggers).forEach(logName => {
+        /**
+         * either apply maskingPatterns or use the global one
+         */
+        const maskingPatterns = maskingPatternsConfig[logName] ?? parseMaskingPatterns(process.env.WDIO_LOG_LEVEL)
+
+        loggers[logName].maskingPatterns = maskingPatterns
     })
 }
 const getLogLevelName = (logName: string) => logName.split(':').shift() as log.LogLevelDesc
--- a/packages/wdio-logger/src/index.ts
+++ b/packages/wdio-logger/src/index.ts
@@ -235,13 +235,12 @@ getLogger.setMaskingPatterns = (maskingPatternsStringPerLoggerName: Record<strin
         process.env.WDIO_LOG_MASKING_PATTERNS = wdioDefault
     }
 
-    maskingPatternsConfig = {}
-
     /**
      * build maskingPatternsConfig object
      */
-    Object.entries(maskingPatternsStringPerLoggerName).forEach(([logName, maskingPatternsString]) => {
-        maskingPatternsConfig[logName] = parseMaskingPatterns(maskingPatternsString)
+    maskingPatternsConfig = Object.entries(maskingPatternsStringPerLoggerName).reduce(([logName, maskingPatternsString], prev) => {
+        prev[logName] = parseMaskingPatterns(maskingPatternsString)
+        return prev
     })
 
     /**
--- a/packages/wdio-logger/src/index.ts
+++ b/packages/wdio-logger/src/index.ts
@@ -231,7 +231,7 @@ getLogger.setMaskingPatterns = (maskingPatternsStringPerLoggerName: Record<strin
     /**
      * set the environment variable for masking patterns
      */
-    if (process.env.WDIO_LOG_MASKING_PATTERNS === undefined) {
+    if (!process.env.WDIO_LOG_MASKING_PATTERNS) {
         process.env.WDIO_LOG_MASKING_PATTERNS = wdioDefault
     }
 
--- a/packages/wdio-logger/src/index.ts
+++ b/packages/wdio-logger/src/index.ts
@@ -97,13 +97,15 @@ const wdioLoggerMethodFactory = (wdioLogger: LoggerInterface) => function (this:
         args = args.map((arg) => {
             for (const s of SERIALIZERS) {
                 if (s.matches(arg)) {
-                    return mask(s.serialize(arg as Error & string), wdioLogger.maskingPatterns)
+                    return s.serialize(arg as Error & string)
                 }
             }
-            return mask(arg, wdioLogger.maskingPatterns)
+            return arg
         }) as [string, string?, ...unknown[]]
 
-        const logText = ansiStrip(`${util.format.apply(this, args as [format: string, ...params: string[]])}\n`)
+        const unmaskedLogText = ansiStrip(`${util.format.apply(this, args as [format: string, ...params: string[]])}\n`)
+        const maskedLogText = mask(unmaskedLogText, wdioLogger.maskingPatterns)
+
         if (logFile && logFile.writable) {
             /**
              * empty logging cache if stuff got logged before
@@ -117,15 +119,21 @@ const wdioLoggerMethodFactory = (wdioLogger: LoggerInterface) => function (this:
                 logCache.clear()
             }
 
-            if (!logsContainInitPackageError(logText)) {
-                return logFile.write(logText)
+            if (!logsContainInitPackageError(unmaskedLogText)) {
+                return logFile.write(maskedLogText)
             }
             // If we get Error during init of integration packages, write logs to both "outputDir" and the terminal
-            logFile.write(logText)
+            logFile.write(maskedLogText)
         }
 
-        logCache.add(logText)
-        rawMethod(...args)
+        logCache.add(maskedLogText)
+
+        // To not break console color formatting, we use the masked log text only when needed
+        if (maskedLogText === unmaskedLogText) {
+            rawMethod(...args)
+        } else {
+            rawMethod(maskedLogText.replace(/\n$/, ''))
+        }
     }
 }
 
--- a/packages/wdio-logger/src/index.ts
+++ b/packages/wdio-logger/src/index.ts
@@ -59,7 +59,6 @@ const SERIALIZERS = [{
 }]
 
 interface LoggerInterface extends log.Logger {
-    setMaskingPatternsConfig: () => RegExp[] | undefined
     maskingPatterns: RegExp[] | undefined
     progress(...msg: string[]): void;
 }
--- a/packages/wdio-logger/src/index.ts
+++ b/packages/wdio-logger/src/index.ts
@@ -233,22 +233,27 @@ getLogger.setLogLevelsConfig = (logLevels: Record<string, log.LogLevelDesc> = {}
         loggers[logName].setLevel(logLevel)
     })
 }
-getLogger.setMaskingPatterns = (maskingPatternsStringPerLoggerName: Record<string, string | undefined> = {}, wdioDefault:string|undefined = undefined) => {
 
-    /**
-     * set the environment variable for masking patterns
-     */
-    if (process.env.WDIO_LOG_MASKING_PATTERNS === undefined) {
-        process.env.WDIO_LOG_MASKING_PATTERNS = wdioDefault
+type MaskPattern = string
+type MaskPatternForSpecificLogger = Record<string, MaskPattern>
+getLogger.setMaskingPatterns = (pattern: MaskPattern | MaskPatternForSpecificLogger) => {
+
+    if ( typeof pattern === 'string') {
+        /**
+         * set the environment variable for masking patterns
+         */
+        if (process.env.WDIO_LOG_MASKING_PATTERNS === undefined) {
+            process.env.WDIO_LOG_MASKING_PATTERNS = pattern
+        }
+    } else if ( typeof pattern === 'object') {
+        /**
+         * build maskingPatternsConfig object
+         */
+        maskingPatternsConfig = Object.entries(pattern).reduce((acc, [logName, maskingPatternsString]) => {
+            acc[logName] = parseMaskingPatterns(maskingPatternsString)
+            return acc
+        }, maskingPatternsConfig)
     }
-
-    /**
-     * build maskingPatternsConfig object
-     */
-    maskingPatternsConfig = Object.entries(maskingPatternsStringPerLoggerName).reduce((acc, [logName, maskingPatternsString]) => {
-        acc[logName] = parseMaskingPatterns(maskingPatternsString)
-        return acc
-    }, maskingPatternsConfig)
 
     /**
      * set masking patterns for each logger
--- a/packages/wdio-logger/src/index.ts
+++ b/packages/wdio-logger/src/index.ts
@@ -253,6 +253,8 @@ getLogger.setMaskingPatterns = (pattern: MaskPattern | MaskPatternForSpecificLog
             acc[logName] = parseMaskingPatterns(maskingPatternsString)
             return acc
         }, maskingPatternsConfig)
+    } else {
+        throw new Error(`Invalid pattern property, expected \`string\` or \`Record<string, string>\` but received \`${typeof pattern}\``)
     }
 
     /**
--- a/packages/wdio-logger/src/index.ts
+++ b/packages/wdio-logger/src/index.ts
@@ -238,10 +238,10 @@ getLogger.setMaskingPatterns = (maskingPatternsStringPerLoggerName: Record<strin
     /**
      * build maskingPatternsConfig object
      */
-    maskingPatternsConfig = Object.entries(maskingPatternsStringPerLoggerName).reduce(([logName, maskingPatternsString], prev) => {
-        prev[logName] = parseMaskingPatterns(maskingPatternsString)
-        return prev
-    })
+    maskingPatternsConfig = Object.entries(maskingPatternsStringPerLoggerName).reduce((acc, [logName, maskingPatternsString]) => {
+        acc[logName] = parseMaskingPatterns(maskingPatternsString)
+        return acc
+    }, maskingPatternsConfig)
 
     /**
      * set masking patterns for each logger
--- a/packages/wdio-logger/src/utils.ts
+++ b/packages/wdio-logger/src/utils.ts
@@ -8,6 +8,13 @@ const skipError = (aFunction: Function) => {
     }
 }
 
+/**
+ * Parses a comma-separated string of regular expressions into an array of RegExp objects.
+ * Supports both `/pattern/flags` and plain pattern formats.
+ *
+ * @param {string | undefined} maskingRegexString - The string containing regex patterns.
+ * @returns {(RegExp[] | undefined)} Array of RegExp objects or undefined if input is invalid.
+ */
 export const parseMaskingPatterns = (maskingRegexString: string | undefined) => {
     if (typeof maskingRegexString !== 'string') { return undefined }
     const regexStrings = maskingRegexString?.split(',').filter((regexStr) => regexStr.trim() !== '')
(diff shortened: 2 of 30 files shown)


packages/webdriver/src/command.ts
```diff
@@ -14,10 +14,20 @@ export default function (
...
     commandInfo: CommandEndpoint,
-    doubleEncodeVariables = false
+    doubleEncodeVariables = false,
+    maskingPatterns: string[] = []
 ) {
     const { command, deprecated, ref, parameters, variables = [], isHubCommand = false } = commandInfo
 
+    // TODO dprevost to move to a better place!
+    const maskingRegExps = maskingPatterns.map((regexString) => {
+        if (!regexString.startsWith('/')) {return new RegExp(regexString)}
```
christian-bromann, line 24:
```suggestion
        if (!regexString.startsWith('/')) {
            return new RegExp(regexString)
        }
```

packages/webdriver/src/command.ts
```diff
@@ -14,10 +14,20 @@ export default function (
...
+        const lastSlashIndex = regexString.lastIndexOf('/')
+        const pattern = regexString.slice(1, lastSlashIndex)
```
christian-bromann, line 26: I guess we assume here that the regex starts with `/` should we double check for that?

packages/wdio-types/src/Capabilities.ts
```diff
@@ -370,6 +370,13 @@ export interface WebdriverIOCapabilities {
      * the WebDriver protocol only.
      */
     'wdio:enforceWebDriverClassic'?: boolean
+
+    /**
+     * Regular expression patterns to use when masking user input in commands having a text parameter
+     * Example: `wdio:maskingPatterns: [/@mask@/i.toString()]`
+     * Note: Because RegExp are not serializable, you need to convert them to string
```
christian-bromann, line 377: Can we add some real word examples here to make the description more expressive? Like how would someone mask environment variables?

packages/webdriver/src/command.ts
```diff
@@ -120,10 +130,33 @@ export default function (
...
+                }
+                maskedArgs = args.map((arg) => maskingRegExps.some((pattern) => pattern.test(arg)) ? sensitiveReplacer : arg)
+            }
+        }
+
+        const request = new environment.value.Request(
+            method,
+            endpoint,
+            body,
+            maskedBody,
```
christian-bromann, line 154: Do we actually want to mask the inputs for the WebDriver request? What if someone wants to use actual credentials to log in but don't want to have them exposed in their logs?

christian-bromann (commented):
Some comments

packages/webdriver/src/command.ts
dprevost-LMI: I wanted to use the regExp literally, but it gets lost when starting the runner somewhere. If I recall correctly, something triggered the run on a "message" emitted event, so I had to use string instead.

The first line uses RegExp patterns without the starting slash. This part assumes we have the two slashes. I can add more validation for sure. Also, I think I forgot to test it without flags; I will do so!

packages/wdio-types/src/Capabilities.ts
dprevost-LMI: 🤔 Targeting `environment variables` was not totally within my scope. As a first step, I wanted to target the text of commands. We could increase the scope, but I was expected to test a first iteration and adjust it later. What do you think?

packages/webdriver/src/command.ts
dprevost-LMI: > to mask the inputs for the WebDriver request

I do not mask the request per see that is sent; I only set a masked version and use it in all other places that are not the exact request. 

This is how I masked my actual password. My password values are `@maks@adwegbefb4ty3henttn35,`, and the RegExp is `/@mask@.*/i`. I use the same RegExp to mask it from the appium log with [log filters](https://appium.io/docs/en/2.0/guides/log-filters/). This appium solution is hard to use, but I wanted to align.

zyntax90: If I may - this is the exact use case @christian-bromann is describing that I've stumbled upon. Im using actual credentials, even if it's a test account, to sign-in into our Azure AAD B2C. The value is injected via env-var and simply set (.setValue) on the selected element, but is logged in clear-text on Browserstack. 
Would be nice to also have it masked on Browserstack itself.

dprevost-LMI: 🤔 I can take a look; it is just that the quick fix in before-after hooks now is growing way more than I anticipated.

packages/webdriver/src/command.ts
dprevost-LMI: I covered cases with slashes, without slashes, with flags, and without flags. I also added unit tests, and I moved that conversion into the utils file!
That should do the job; just unresolved the comment if I missed something!

packages/webdriver/src/command.ts
dprevost-LMI: @zyntax90 My approach is to align with the appium approach of [log filters](https://appium.io/docs/en/2.0/guides/log-filters/). If we diverge, you could be masking in wdio logs but not in appium ones.

Maybe I misused the appium strategy, but besides setting a known keyword in my password, I could not mask my values from the appium logs, which output them once with the value surrounded by double quotes and another time with all my password characters separated by a comma (see screenshot below). 

image

My proposed option is to ensure we can be masked in both wdio & appium logs with the same RegExp.

I was not too fond of the appium strategy, but I'm all open to suggestions. Again, if we diverge, people will still need to find some ways to find their proper RegExp using the appium log filter; unless we succeed in also implementing something different in appium... we can dream 😄 

To brainstorm:
Idea #1:
- Instead of a RegExp, as I propose, that masked everything. We could add an option to remove a matching group.
So the config would be like below. `removeOnGroupMatch` could even be optional, and we could just state in the doc that matching groups or removed before sending the value to appium

```
const config = { 
    maskPattern: '/(@mask@).*/i',
    removeOnGroupMatch: true
}
```
With `@mask@myPassword` I have the below
image

Then, I would replace Group 1 with an empty string and send that to appium.

For appium log now, that cannot apply, and we will need to disable the appium logging!

Idea #2 .... thinking!

dprevost-LMI: Also @zyntax90 , for
> but is logged in clear-text on Browserstack.
You need to use [maskCommand](https://www.browserstack.com/docs/automate/selenium/hide-sensitive-data) else what ever we do in wdio it will never mask their `Text Logs`

zyntax90: > Also @zyntax90 , for
> 
> > but is logged in clear-text on Browserstack.
> 
> You need to use [maskCommand](https://www.browserstack.com/docs/automate/selenium/hide-sensitive-data) else what ever we do in wdio it will never mask their `Text Logs` (as far as I know)

Ah, thanks for the clarification and I do think it's a neat idea of using matching group.

dprevost-LMI: @zyntax90 with the appium log disabled in that case, right?

dprevost-LMI:
> WDIO logs like in this screenshot:
Just to clarify, the screenshot is the appium one, not the WDIO one, but they are very similar!

We could choose solution #1, which is mostly as you describe it, @christian-bromann. Additionally, solution # 2 could be nice and require little more effort! 

Solution # 1
The most secure one is where we mask both WDIO and Appium the same way you describe.
   - The burden is to find a proper regEx fitting potentially custom data unless, like me, you can enter a fixed token as in a password
   -  Honestly, the appium way with the log filter, I do not find it "intuitive" since I had to be creative to find a proper RegEx and "inject" my token. 
   - If, for example, you deal with some accessKey that you cannot change, you can go with the length, but we had that, and it was colliding with a random value generated by Faker for test data.
   - Even with the above, I agree with your proposed approach, and if I can filter in the function you mention and inject the RegExp to the appium config, it would be the best and most solid

Solution # 2
To have more flexibility and not have to find a creative Regex or change your secret to contain a token, the alternative would be to inject a token on setValue and remove it with RegEx when you have a matching group before sending it! However, I would have to document the following:
⚠️ Using group matching to remove a given token injected on the setValue makes it vulnerable to be found in Appium Logs; you MUST DISABLED appium logs to have a secured solution

Both solutions could be implemented since they are RegExp-based, and removing a string from a match group is one or two more code lines!

Let me know what you thing!

packages/wdio-types/src/Capabilities.ts
dprevost-LMI: Let's close this since I changed my approach

packages/webdriver/src/command.ts
dprevost-LMI: In the end, I use regex for masking logs and capture, for example, the `--key=mySecertKey` pass when running locally but targeting BrowserStack.

For the input element, I use a boolean on setValue indicating if masking is required; this is the most friendly approach has in `.setValue(userPassword, true);`
 
For appium, I will continue to dig outside of wdio changes since changing their loggers requires drivers of Android and iOS changes, and now they want me to look into a plugin approach. Since it is getting complicated and driving me to stop this development, I'd put it aside and have something working in wdio instead.

So, for now, the appium logs must be disabled for something secure or configured separately as they documented.

Let's close this since I changed my approach.

packages/wdio-logger/src/index.ts
```diff
@@ -215,6 +226,36 @@ getLogger.setLogLevelsConfig = (logLevels: Record<string, log.LogLevelDesc> = {}
         loggers[logName].setLevel(logLevel)
     })
 }
+getLogger.setMaskingPatterns = (maskingPatternsStringPerLoggerName: Record<string, string | undefined> = {}, wdioDefault = undefined) => {
+
+    /**
+     * set the environment variable for masking patterns
+     */
+    if (process.env.WDIO_LOG_MASKING_PATTERNS === undefined) {
```
christian-bromann, line 234: This should be good enough and avoids cases where users may provide an empty string by accident
```suggestion
    if (!process.env.WDIO_LOG_MASKING_PATTERNS) {
```

packages/wdio-logger/src/index.ts
```diff
@@ -215,6 +226,36 @@ getLogger.setLogLevelsConfig = (logLevels: Record<string, log.LogLevelDesc> = {}
...
+    }
+
+    maskingPatternsConfig = {}
+
+    /**
+     * build maskingPatternsConfig object
+     */
+    Object.entries(maskingPatternsStringPerLoggerName).forEach(([logName, maskingPatternsString]) => {
+        maskingPatternsConfig[logName] = parseMaskingPatterns(maskingPatternsString)
+    })
```
christian-bromann, line 245:
```suggestion
    /**
     * build maskingPatternsConfig object
     */
    maskingPatternsConfig = Object.entries(maskingPatternsStringPerLoggerName).reduce(([logName, maskingPatternsString], prev) => {
        prev[logName] = parseMaskingPatterns(maskingPatternsString)
        return prev
    })
```

packages/wdio-protocols/src/protocols/webdriver.ts
```diff
@@ -1007,6 +1007,12 @@ export default {
                     description: 'string to send as keystrokes to the element',
                     required: true,
                 },
+                {
+                    name: 'mask',
+                    type: 'boolean',
+                    description: 'ensure the value is masked outside of the request made to appium server',
+                    required: false,
+                },
```
christian-bromann, line 1015: I think we shouldn't introduce options that aren't defined in the protocol here. As mentioned on the issue I would prefer to mask values globally and allow users to specify when it should not happen.

In case there is an actual reason a user would want to control this on this level I suggest to amend this script to add this option to the type manually.

dprevost-LMI: 🤔. Let's try to cover these points in the issue.

dprevost-LMI: As discussed, if still needed, the above will be reverted and unofficially used only to pass the mask boolean from the test to the HTTP layer and remove that runtime param before sending the official command request to appium.
And we give a look at [transform request](https://webdriver.io/docs/configuration/#transformrequest) if it can replace the need of that mask boolean in the command

packages/webdriver/src/types.ts
```diff
@@ -96,3 +96,11 @@ export interface AttachOptions extends Partial<SessionFlags>, Partial<Options.We
     // original requested capabilities
     requestedCapabilities?: Capabilities.WithRequestedCapabilities['capabilities']
 }
+
+/**
+ * Additional options outside of the WebDriver spec exclusively for WebdriverIO only for runtime and not send to appium
```
dprevost-LMI, line 101:
```suggestion
 * Additional options outside of the WebDriver spec, exclusively for Webdriverio, only for runtime, and not sent to appium
```

packages/webdriverio/src/types.ts
```diff
@@ -677,6 +677,13 @@ export interface SaveScreenshotOptions {
     }
 }
 
+/**
+ * Additional options outside of the WebDriver spec exclusively for WebdriverIO only for runtime and not send to appium
```
dprevost-LMI, line 681:
```suggestion
 * Additional options outside of the WebDriver spec, exclusively for Webdriverio, only for runtime, and not sent to appium
```

packages/wdio-protocols/src/protocols/webdriver.ts
dprevost-LMI: I reverted that change not to alter the protocol. However, note that:
1. I had to use `@ts-ignore` in `setValue.ts` since using a third parameter breaks the type, and I was not able to find another way
```
    if (options) {
        // @ts-ignore bypassing typing until we can find a better solution
        return this.elementSendKeys(this.elementId, value.toString(), options satisfies CommandRuntimeOptions)
    }
    return this.elementSendKeys(this.elementId, value.toString())
```
3. Since I passed a third argument to `protocolCommand` and those are validated, I had to remove it as the first step in that file.
```
        let runtimeOptions = {}
        if (commandInfo.parameters.length < unmaskedArgs.length && typeof unmaskedArgs[unmaskedArgs.length-1] === 'object') {
            // Popping the additional options to not have `Wrong parameters applied` thrown
            runtimeOptions = unmaskedArgs.pop() as CommandRuntimeOptions
        }
```
Also, I have validated the usage of `transformRequest` and:
1. It is made after `request.makeRequest`, which is too late since on top of masking the body, the args need to be masked in the `protocolCommand`
2. I cannot simply transform the body since masking it there will send the masked version to appium
3. Also, because of point 1, it can’t be used as a replacement for the mask parameter.

packages/webdriver/tests/command.test.ts
```diff
@@ -252,6 +253,63 @@ describe('command wrapper', () => {
...
+                type: 'string',
+                description: 'the text to be sent to the element',
+                required: true
+            }]
+        }
+        const elementId = '123'
+        const text = 'mySecretPassword'
+        const runtimeOptionsWithMasking = { mask: true }
+
+        it('should do a the http request with the unmasked text when masking is requested', async () => {
```
dprevost-LMI, line 277:
```suggestion
        it('should do the http request with the unmasked text when masking is requested', async () => {
```

packages/webdriver/tests/command.test.ts
```diff
@@ -252,6 +253,63 @@ describe('command wrapper', () => {
...
+                {
+                    text: 'mySecretPassword',
+                },
+                expect.any(AbortSignal),
+                false,
+                requestHandler
+            )
+        })
+
+        it('should emit result with masked text', async () => {
```
dprevost-LMI, line 294: TODO dprevost Add similar test for the other emit and log.info

packages/webdriverio/src/types.ts
```diff
@@ -734,3 +741,4 @@ declare global {
         interface Dialog extends DialogImport {}
     }
 }
+
```
dprevost-LMI, line 744:
```suggestion
```

packages/webdriver/tests/utils.test.ts
```diff
@@ -6,7 +6,7 @@ import '../src/browser.js'
 
 import {
     isSuccessfulResponse, getPrototype, getSessionError,
-    startWebDriverSession, setupDirectConnect, validateCapabilities
+    startWebDriverSession, setupDirectConnect, validateCapabilities,
```
dprevost-LMI, line 9:
```suggestion
    startWebDriverSession, setupDirectConnect, validateCapabilities
```

packages/webdriver/src/command.ts
```diff
@@ -133,34 +146,41 @@ export default function (
...
-            onRetry: (data) => this.emit('request.retry', data)
+            onRetry: (data) => this.emit('request.retry', data),
+            onLogData: (data) => log.info('DATA', transformCommandLogResult((wasMasked ? maskedBody : data) as Record<string, unknown>))
         })
-        this.emit('command', { command, method, endpoint, body })
-        log.info('COMMAND', commandCallStructure(command, args))
+        this.emit('command', { command, method, endpoint, body: maskedBody })
+        log.info('COMMAND', commandCallStructure(command, maskedArgs))
+
+        const options = maskedBody ? { ...this.options, headers: { ['x-appium-is-sensitive']: 'true' } } : this.options
```
dprevost-LMI, line 165: TODO dprevost add test for this new header

packages/webdriver/tests/command.test.ts
dprevost-LMI: Done

packages/webdriver/src/command.ts
dprevost-LMI: Done

packages/webdriver/src/command.ts
```diff
@@ -133,17 +146,29 @@ export default function (
...
-        this.emit('command', { command, method, endpoint, body })
-        log.info('COMMAND', commandCallStructure(command, args))
+        this.emit('command', { command, method, endpoint, body: maskedBody })
+        log.info('COMMAND', commandCallStructure(command, maskedArgs))
+
+        if (isMasked) {
+            this.options.headers = {
+                ...this.options.headers,
+                ...APPIUM_MASKING_HEADER
+            }
```
dprevost-LMI, line 169: TODO: We should not alter the client options, we should have a const here instead

dprevost-LMI: Fixed

packages/wdio-logger/src/index.ts
```diff
@@ -215,6 +226,35 @@ getLogger.setLogLevelsConfig = (logLevels: Record<string, log.LogLevelDesc> = {}
         loggers[logName].setLevel(logLevel)
     })
 }
+getLogger.setMaskingPatterns = (maskingPatternsStringPerLoggerName: Record<string, string | undefined> = {}, wdioDefault = undefined) => {
```
dprevost-LMI, line 229: TODO: Missing unit tests

dprevost-LMI: Done!

packages/webdriverio/src/commands/element/setValue.ts
```diff
@@ -22,12 +24,14 @@
  *
  * @alias element.setValue
  * @param {string|number}  value  value to be added
+ * @param {WdioOptions}  options additional exclusive to webdriverio
```
dprevost-LMI, line 27:
```suggestion
 * @param {WdioOptions} additional options, exclusive to webdriverio
```

packages/wdio-logger/tests/node.test.ts
```diff
@@ -221,6 +221,126 @@ describe('wdio-logger node', () => {
...
+                        },
+                        expectedPatterns: [expectedPatterns[0], /info/]
+                    },
+                        // TODO dprevost fix this test
+                    /*, {
+                        name: 'should be possible to set logLevel in config for all sub levels',
+                        logger: 'test-setMaskingPatterns-7:foo',
+                        config: { 'test-setMaskingPatterns-7:bar': pattern },
+                        expectPattern
+                    }*/]
```
dprevost-LMI, line 315: TODO. Why does the logger have a `:` in the name, and it is parsed?

packages/wdio-logger/src/index.ts
```diff
@@ -215,6 +226,35 @@ getLogger.setLogLevelsConfig = (logLevels: Record<string, log.LogLevelDesc> = {}
         loggers[logName].setLevel(logLevel)
     })
 }
+getLogger.setMaskingPatterns = (maskingPatternsStringPerLoggerName: Record<string, string | undefined> = {}, wdioDefault:string|undefined = undefined) => {
```
dprevost-LMI, line 229: Verify if, by default, we use `--key=[^ ]*` to mask the user's key from the command line. However, not sure when to apply it, always? Or only when Selenium, Sauce or BrowserStack are used?

christian-bromann:
Thanks @dprevost-LMI , I will take a look asap.

packages/wdio-logger/src/index.ts
christian-bromann, line 229: Can we document this in `packages/wdio-logger/README.md`?

packages/wdio-logger/tests/node.test.ts
christian-bromann: some logger names may contain `:`, e.g. https://github.com/webdriverio/webdriverio/blob/e2250c38e475d39fc8057e5e100540948d91acfb/packages/wdio-cli/src/watcher.ts#L16

packages/webdriver/src/utils.ts
```diff
@@ -455,3 +455,33 @@ export function parseBidiMessage (this: EventEmitter, data: Buffer) {
         log.error(`Failed parse WebDriver Bidi message: ${(err as Error).message}`)
     }
 }
+
+export const APPIUM_MASKING_HEADER = { 'x-appium-is-sensitive': 'true' }
```
christian-bromann, line 459: Can we move this into `packages/webdriver/src/constants.ts`?

packages/webdriver/src/utils.ts
```diff
@@ -455,3 +455,33 @@ export function parseBidiMessage (this: EventEmitter, data: Buffer) {
...
+                const maskedArgs = args.slice(0, textValueArgsIndex).concat(SENSITIVE_DATA_REPLACER).concat(args.slice(textValueArgsIndex + 1))
+                return {
+                    maskedBody: maskedBody,
+                    maskedArgs: maskedArgs,
+                    isMasked: true,
+                }
+            }
+        }
+    }
+    return { maskedBody: body, maskedArgs: args, isMasked: false }
```
christian-bromann, line 486:
```suggestion
    if (!options.mask) {
        return { maskedBody: body, maskedArgs: args, isMasked: false }
    }

    const textValueParamIndex = commandInfo.parameters.findIndex((param) => param.name === 'text')
    if (textValueParamIndex !== -1 ) {
        const textValueIndexInArgs = (commandInfo.variables?.length ?? 0) + textValueParamIndex
        const text = args[textValueIndexInArgs]
        if (text && typeof text === 'string') {
            const maskedBody = {
                ...body,
                text: SENSITIVE_DATA_REPLACER
            } satisfies Record<string, unknown> as Record<string, unknown>
            const textValueArgsIndex = textValueParamIndex + (commandInfo.variables?.length ?? 0)
            const maskedArgs = args.slice(0, textValueArgsIndex).concat(SENSITIVE_DATA_REPLACER).concat(args.slice(textValueArgsIndex + 1))
            return {
                maskedBody: maskedBody,
                maskedArgs: maskedArgs,
                isMasked: true,
            }
        }
    }
```

christian-bromann: Let's return early to save indentation

packages/webdriverio/src/types.ts
```diff
@@ -677,6 +677,13 @@ export interface SaveScreenshotOptions {
     }
 }
 
+/**
+ * Additional options outside of the WebDriver spec, exclusively for Webdriverio, only for runtime, and not sent to appium
+ */
+export interface WdioOptions {
```
christian-bromann, line 683: `WdioOptions` seems to general , maybe we can be a bit more specific here?

```suggestion
export interface InputOptions {
```

christian-bromann: Should we just re-use the `CommandRuntimeOptions` parameter?

packages/webdriver/src/command.ts
```diff
@@ -19,7 +20,14 @@ export default function (
 ) {
     const { command, deprecated, ref, parameters, variables = [], isHubCommand = false } = commandInfo
 
-    return async function protocolCommand (this: BaseClient, ...args: unknown[]): Promise<WebDriverResponse | BidiResponses | void> {
+    return async function protocolCommand (this: BaseClient, ...unmaskedArgs: unknown[]): Promise<WebDriverResponse | BidiResponses | void> {
+
+        let runtimeOptions = {}
+        if (commandInfo.parameters.length < unmaskedArgs.length && typeof unmaskedArgs[unmaskedArgs.length-1] === 'object') {
```
christian-bromann, line 26: Let's find a better way to detect runtime options, maybe we can define a class for them to have `instanceof` checks or require runtime options to always have a `type: "runtimeOptions"` property.

christian-bromann (commented):
Amazing work 👏 

Some minor details but overall LGTM

packages/webdriver/src/utils.ts
dprevost-LMI: 🤔 It was making sense, but turning the code around generates an undefined path:
image

packages/webdriver/src/command.ts
dprevost-LMI: Done by turning into a class `CommandRuntimeOptions`
```ts
if (unmaskedArgs.length > 0 && unmaskedArgs[unmaskedArgs.length - 1] instanceof CommandRuntimeOptions )
```

packages/webdriverio/src/types.ts
dprevost-LMI: Now that `CommandRuntimeOptions` is a class, having `InputOptions` as an interface makes it easier to use in the `setValue` and `addValue`, so I would keep both.

packages/webdriver/src/utils.ts
dprevost-LMI: Since it generates a ts compile issue, I will resolve the comment for now!

packages/webdriver/src/utils.ts
dprevost-LMI: Done!

packages/wdio-logger/src/utils.ts
```diff
@@ -0,0 +1,41 @@
...
+export const mask = <T = unknown>(possibleText: T, maskingPatterns: RegExp[] | undefined) => {
+    if (!maskingPatterns || typeof possibleText !== 'string') { return possibleText }
+
+    let maskedText = possibleText as string
+    maskingPatterns.forEach((maskingRegex) => {
+        // Using the below allows supporting the 'g' flags easier, else using 'g' with text.match(regExp) does not output proper capturing groups
+        const groupCount = (maskingRegex.source.match(/\((?!\?)/g) || []).length
+
+        // Supporting from 0 to 2 capturing groups, where when having two, only the second group is masked to cover cases like `--key=VALUE` for BrowserStack secret key
+        maskedText = maskedText.replace(maskingRegex, groupCount < 2 ? SENSITIVE_DATA_REPLACER: `$1${SENSITIVE_DATA_REPLACER}`)
```
dprevost-LMI, line 38: While documenting, I wonder why I don't just replace all capturing groups with the `**MASKED**` and keep the rest. Maybe it's because of the `g` thing above, but I'll review it since it would ease its logic and usage!

packages/wdio-protocols/src/protocols/webdriver.ts
dprevost-LMI: outadated!

packages/webdriverio/src/types.ts
dprevost-LMI: Resolving since the two are more relevant now, let me know if you disagree!

packages/wdio-logger/src/utils.ts
dprevost-LMI: Fixed

packages/wdio-logger/src/index.ts
dprevost-LMI: Done, I also have:
1. Review to mask all capturing groups to ease the logic and comprehension
2. Provide a way to configure through a capability named `maskingPatterns`

dprevost-LMI: Since I don't know how to check if a cloud provider is used, I suggest letting the user decide to configure their masking patterns. 
Instead, I provide helpful documentation on the `maskingPatterns` capabilities with an example on how to mask your cloud provider key

packages/wdio-logger/tests/node.test.ts
dprevost-LMI: 🤔 Thinking about it, I wonder if I need all this logic now about providing a pattern per logger. 
If we just provide the `maksingPatterns` capabilities, this feature per logger sounds not used and removable 🤔

packages/webdriver/tests/command.test.ts
```diff
@@ -62,17 +66,28 @@ vi.mock('../src/request/request', () => {
...
+    })
+
     return {
         thenMock,
         catchMock,
         finallyMock,
         WebDriverRequest,
+        makeRequestMock,
+        getCapturedRequestHandler: capturedRequestEventHandler
+
```
dprevost-LMI, line 90:
```suggestion
        getCapturedRequestHandler: capturedRequestEventHandler
```

packages/wdio-logger/src/index.ts
```diff
@@ -215,6 +225,35 @@ getLogger.setLogLevelsConfig = (logLevels: Record<string, log.LogLevelDesc> = {}
...
```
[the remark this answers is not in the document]
dprevost-LMI: Done see 200d87283fd07cc07693af97d44e6626cb8b117d

packages/wdio-logger/src/utils.ts
```diff
@@ -0,0 +1,51 @@
...
+        } else if (regexParts?.[1]) {
+            // Case with flag `/(--key=)([^ ]*)/i` or without flag `/(--key=)([^ ]*)/`
+            return skipError(() => regexParts[2] ? new RegExp(regexParts[1], regexParts[2]) : new RegExp(regexParts[1]))
+        }
+        return undefined
+
+    }).filter((regex) => regex !== undefined)
+}
+
+export const mask = <T = unknown>(possibleText: T, maskingPatterns: RegExp[] | undefined) => {
```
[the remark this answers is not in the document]
dprevost-LMI: Done in 73873e657fc0bfeea7613a743b8c47e933eaa101

packages/webdriver/src/utils.ts
```diff
@@ -455,3 +455,35 @@ export function parseBidiMessage (this: EventEmitter, data: Buffer) {
...
+}
```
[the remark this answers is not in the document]
dprevost-LMI: 😮 Oh got it! Thanks for opening my mind 😉 . Done see 6c8cb20751520483dfddf56913070c0af100904b

packages/wdio-logger/tests/node.test.ts
dprevost-LMI: Okay, great, I already removed this comment, so let's keep this as it is for now! Thanks!

packages/webdriver/tests/command.test.ts
```diff
@@ -62,17 +66,27 @@ vi.mock('../src/request/request', () => {
...
 
+    const makeRequestMock = vi.fn().mockReturnValue(promise)
     thenMock.mockReturnValue(promise)
+
+    const capturedRequestEventHandler = vi.fn<() => RequestEventHandler>()
+
+    // Mock the constructor
+    const WebDriverRequest = vi.fn().mockImplementation((
+        _method, _endpoint, _body, _abortSignal, _isHubCommand, requestEventHandler: RequestEventHandler
+    ) =>{
```
christian-bromann, line 78:
```suggestion
    ) => {
```

packages/wdio-logger/src/index.ts
```diff
@@ -215,6 +233,40 @@ getLogger.setLogLevelsConfig = (logLevels: Record<string, log.LogLevelDesc> = {}
...
+        }
+    } else if ( typeof pattern === 'object') {
+        /**
+         * build maskingPatternsConfig object
+         */
+        maskingPatternsConfig = Object.entries(pattern).reduce((acc, [logName, maskingPatternsString]) => {
+            acc[logName] = parseMaskingPatterns(maskingPatternsString)
+            return acc
+        }, maskingPatternsConfig)
+    }
```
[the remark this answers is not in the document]
dprevost-LMI: It's okay, just missed the time to change that + unit tests. It will come in the next days!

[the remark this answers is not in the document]
dprevost-LMI: Finally I took the time and it is now done 🎉

website/docs/Security.md
```diff
@@ -0,0 +1,160 @@
...
+    const passwordInputElement = $('TextField[@name="Password"]');
+
+    // Get focus
+    await passwordInputElement.click();
+
+    await passwordInputElement.setValue('userPassword', { mask: true });
+  }
+```
+
+The above will hide the text value from WDIO logs and also from appium logs.
```
dprevost-LMI, line 31:
```suggestion
The above will hide the text value from WDIO logs and also from Appium logs. 
```

website/docs/Security.md
```diff
@@ -0,0 +1,160 @@
...
+
+### Mask Commands' Text Value
+
+The commands `addValue` and `setValue` support a boolean mask value to mask in WDIO and Appium logs, as well as reporters. Moreover, other tools, such as performance tools and third-party tools, will also receive the mask version, enhancing security.
+
+For example, if you are using a real production user and need to enter a password that you want to mask, then it is now possible with the following:
+
+```ts
+  async enterPassword(userPassword) {
+    const passwordInputElement = $('TextField[@name="Password"]');
```
dprevost-LMI, line 22:
```suggestion
    const passwordInputElement = $('Password');
```

website/docs/Security.md
```diff
@@ -0,0 +1,160 @@
...
+
+The above will hide the text value from WDIO logs and also from appium logs. 
+
+Logs example:
+```text
+2025-05-25T23:02:42. 336Z INFO webdriver: DATA { text: "**MASKED**" }
+```
+
+Limitations:
+  - In Appium, additional plugins could leak even though we ask appium to mask the information.
```
dprevost-LMI, line 39:
```suggestion
  - In Appium, additional plugins could leak even though we ask to mask the information.
```

christian-bromann (approved):
Amazing stuff 🙌

PR #13949, opened 2024-12-06
fix(@wdio/utils): properly detect Appium browser sessions

### LINKED ISSUE #13947: [🐛 Bug]: WDIO 9.4.1 fails to create a session with chromedriver or geckodriver

### WebdriverIO Version

9.4.1

### Node.js Version

22

### Mode

Standalone Mode

```typescript
{
  capabilities: {
    alwaysMatch: {
      browserName: 'MozillaFirefox',
      platformName: 'linux',
      'appium:automationName': 'Gecko',
      webSocketUrl: true,
      unhandledPromptBehavior: 'ignore'
    },
    firstMatch: [ {} ]
  }
}
```

### What happened?

After WDIO upgrade to 9.4.1 session creation with safaridriver or geckdriver@Android always fails. I assume this happens because of the context check call, which is not supported by these drivers.

See 

- https://github.com/appium/appium-geckodriver/actions/runs/12138841083/job/33845133024?pr=123
- https://github.com/appium/appium-safari-driver/actions/runs/12138808640/job/33845033745?pr=130

for more details

It should be possible to create a session same way it was working in previous wdio versions.

### How to reproduce the bug.

Just run the corresponding pipeline

### Relevant log output

```typescript
See the links above
```

### Code of Conduct

christian-bromann:
## Proposed changes

fixes #13947

## Types of changes

## Checklist

## Backport Request

## Further comments

n/a

### DIFF
--- a/packages/wdio-utils/src/envDetector.ts
+++ b/packages/wdio-utils/src/envDetector.ts
@@ -106,6 +106,15 @@ function isMobile(capabilities: WebdriverIO.Capabilities) {
     const browserName = (capabilities.browserName || '').toLowerCase()
     const bsOptions = capabilities['bstack:options'] || {}
     const browserstackBrowserName = (bsOptions.browserName || '').toLowerCase()
+
+    /**
+     * it is _not_ a mobile session if automationName is set to start a mobile session
+     */
+    const automationName = capabilities['appium:options']?.automationName || capabilities['appium:automationName']
+    if (automationName && ['gecko', 'safari', 'chrome', 'chromium'].includes(automationName.toLocaleLowerCase())) {
+        return false
+    }
+
     /**
      * we have mobile capabilities if
      */
--- a/packages/wdio-utils/tests/envDetector.test.ts
+++ b/packages/wdio-utils/tests/envDetector.test.ts
@@ -60,6 +60,13 @@ describe('sessionEnvironmentDetector', () => {
         expect(sessionEnvironmentDetector({ capabilities: tvOSCapsBS, requestedCapabilities }).isMobile).toBe(true)
         const androidCapsBS = { 'bstack:options': { ...chromeCaps, 'platformName': 'Android' } }
         expect(sessionEnvironmentDetector({ capabilities: androidCapsBS, requestedCapabilities }).isMobile).toBe(true)
+
+        const geckoAppiumCaps = { 'appium:automationName': 'Gecko' }
+        expect(sessionEnvironmentDetector({ capabilities: geckoAppiumCaps, requestedCapabilities }).isMobile).toBe(false)
+        const safariAppiumCaps = { 'appium:options': { automationName: 'safari' } }
+        expect(sessionEnvironmentDetector({ capabilities: safariAppiumCaps, requestedCapabilities }).isMobile).toBe(false)
+        const chromiumAppiumCaps = { 'appium:automationName': 'chrome' }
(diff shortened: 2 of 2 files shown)


packages/wdio-utils/src/envDetector.ts
```diff
@@ -106,6 +106,15 @@ function isMobile(capabilities: WebdriverIO.Capabilities) {
     const browserName = (capabilities.browserName || '').toLowerCase()
     const bsOptions = capabilities['bstack:options'] || {}
     const browserstackBrowserName = (bsOptions.browserName || '').toLowerCase()
+
+    /**
+     * it is _not_ a mobile session if automationName is set to start a mobile session
+     */
+    const automationName = capabilities['appium:options']?.automationName || capabilities['appium:automationName']
```
wswebcreation, line 113: @christian-bromann 

this is strange, we are automating a mobile device, but only, if I understand it correctly, with a "browser driver". Shouldn't we introduce a new flag for this and use that to exclude specific mobile commands for non-mobile browser drivers?

christian-bromann: > we are automating a mobile device

No, we are using Appium to automate a browser.

> Shouldn't we introduce a new flag for this and use that to exclude specific mobile commands for non-mobile browser drivers?

I think there is a difference here, users can:

- run a browser session on a mobile device: WebdriverIO properly detects this and sets `isMobile` to true (which I am not sure is right), however Appium in this case supports `getContext` (I think)
- we are using Appium to automate a desktop browser, here `isMobile` should be false but was detected as `true`

This patch fixes the second case. I would need more investigation to better understand the first case.

wswebcreation: @christian-bromann 

This [link/config](https://github.com/appium/appium-geckodriver/blob/master/test/functional/mobile-driver-e2e-specs.js) from @mykola-mokhnach suggests something else. This is FF on Android which would still be a mobile session.

> run a browser session on a mobile device: WebdriverIO properly detects this and sets isMobile to true (which I am not sure is right), however Appium in this case supports getContext (I think)
From what I know the default drivers (appium-xcuitest-driver and appium-uiautomator2-driver) can automate both, meaning web and native. That's the reason why they support the `getContext`.

For now we can proceed with this change, but then we need to add a comment to the code that we need to rethink this logic

packages/wdio-utils/src/envDetector.ts
```diff
@@ -106,6 +106,15 @@ function isMobile(capabilities: WebdriverIO.Capabilities) {
     const browserName = (capabilities.browserName || '').toLowerCase()
     const bsOptions = capabilities['bstack:options'] || {}
     const browserstackBrowserName = (bsOptions.browserName || '').toLowerCase()
+
+    /**
+     * it is _not_ a mobile session if automationName is set to start a mobile session
```
wswebcreation, line 111:
```suggestion
     * There are cases where sessions with `appium:*` prefixed capabilities do not fully support all "native"-mobile commands.  
     * In this case the `appium:automationName` is set with something else than the 
     * `xcuitest|uiautomator2|flutter|espress|..` value. This can be a browser driver or 
     * a "wrapped" appium browser-driver. See also https://github.com/webdriverio/webdriverio/issues/13947
     * Return `isMobile:false` for those cases. There we also accepts that specific mobile browser
     * tests (like the FF one on Android) are not seen as a mobile one
```

PR #13950, opened 2024-12-06
fix(webdriverio): don't fail if getContext is not supported

christian-bromann:
## Proposed changes

We've had folks reporting that they get an error like this:

aaaaa

This patch mitigates this problem by just catching this error.

## Types of changes

## Checklist

## Backport Request

## Further comments

n/a

### DIFF
--- a/packages/webdriverio/src/context.ts
+++ b/packages/webdriverio/src/context.ts
@@ -1,6 +1,8 @@
+import logger from '@wdio/logger'
 import { getMobileContext, getNativeContext } from './utils/mobile.js'
 
 const contextManager = new Map<WebdriverIO.Browser, ContextManager>()
+const log = logger('webdriverio:context')
 
 export function getContextManager(browser: WebdriverIO.Browser) {
     const existingContextManager = contextManager.get(browser)
@@ -122,11 +124,15 @@ export class ContextManager {
             !this.#isNativeContext &&
             !this.#mobileContext
         ) {
-            const context = await this.#browser.getContext()
-            this.#mobileContext = typeof context === 'string' ?
-                context : typeof context === 'object' ?
-                    context.id :
-                    undefined
+            const context = await this.#browser.getContext().catch((err) => {
+                log.warn(`Error getting context: ${err}`)
+                return undefined
+            })
+            this.#mobileContext = typeof context === 'string'
+                ? context
+                : typeof context === 'object'
+                    ? context.id
+                    : undefined
         }
 
         const windowHandle = this.#mobileContext || await this.#browser.getWindowHandle()
--- a/packages/webdriverio/src/context.ts
+++ b/packages/webdriverio/src/context.ts
@@ -125,7 +125,11 @@ export class ContextManager {
             !this.#mobileContext
         ) {
             const context = await this.#browser.getContext().catch((err) => {
-                log.warn(`Error getting context: ${err}`)
+                log.warn(
+                    `Error getting context: ${err}\n\n` +
+                    `WebDriver capabilities: ${JSON.stringify(this.#browser.capabilities)}\n` +
+                    `Requested WebDriver capabilities: ${JSON.stringify(this.#browser.requestedCapabilities)}`
+                )
                 return undefined
             })
             this.#mobileContext = typeof context === 'string'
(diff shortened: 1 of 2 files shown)


packages/webdriverio/src/context.ts
```diff
@@ -122,11 +124,15 @@ export class ContextManager {
             !this.#isNativeContext &&
             !this.#mobileContext
         ) {
-            const context = await this.#browser.getContext()
-            this.#mobileContext = typeof context === 'string' ?
-                context : typeof context === 'object' ?
-                    context.id :
-                    undefined
+            const context = await this.#browser.getContext().catch((err) => {
```
wswebcreation, line 127: @christian-bromann 

Do you also have the context of when this happens, this is not something that should occur when using Appium, unless they are for example 
- using wrong caps/WDIO gets back non W3C ones
- the driver is old as ...

wswebcreation (commented):
Hi @christian-bromann , see my comment

packages/webdriverio/src/context.ts
christian-bromann: I think generally we can't make the assumption that just because a driver is considered as "mobile" they have the `getContext` command implemented. A driver can always fail to send the correct capabilities causing WebdriverIO to consider a session as mobile when it isn't. This patch will continue to log and elevate the problem but won't cause the whole session to fail.

What do you think?

wswebcreation: So my two-cents: 
We then might need to:

- Fix the isMobile check to make it more robust 
- Do this "fallback" for all commands

I also believe we can't "patch" the vendors (because this doesn't seem to be a specific Appium issue, unless a very old driver is used) if they don't return "valid" W3C responses.

christian-bromann: > * Fix the isMobile check to make it more robust

We should totally strive for that but unfortunately can't rely on vendors implementing the specifications correctly.

> * Do this "fallback" for all commands

What do you mean by "all commands"? If some one uses an unsupported command in their test, it should fail. However with these session manager (e.g. context, shadow root manager or the polyfill manager) we are enhancing WebdriverIO in the background and these things shouldn't cause the testrun to fail. Hence this patch which makes the context manager do nothing if `getContext` is not implemented.

> if they don't return "valid" W3C responses.

I think this is not the issue. The issue is that the command is not implemented by the vendor or WebDriver server in which case we should disable the context manager.

wswebcreation: > What do you mean by "all commands"? If some one uses an unsupported command
I mean for mobile commands specific commands. But I agree with what you say

> I think this is not the issue. The issue is that the command is not implemented by the vendor or WebDriver server in which case we should disable the context manager.
I believe you just fixed this "not implemented by WebDriver" with https://github.com/webdriverio/webdriverio/pull/13949, so I wonder if this "fix" is still needed. I would then still let it fail and throw a different error message (maybe throw the capabilities that are used) to determine which drivers are not supporting this. 
Regarding the vendors, from what I now see LT is doing it differently, they default to JSONWP

image

and the W3C caps are not "Appium" proof

image

I'll put checking the vendors on my todo list so we can properly check if it's mobile based on the correct caps

christian-bromann: > I believe you just fixed this "not implemented by WebDriver" with #13949, so I wonder if this "fix" is still needed.

I put up this patch because I am afraid we run into more issues where our mobile capability detection fails and causes a failure with the whole test run. I think this patch makes sense no matter what because we just disable the context manager in case the command fails.

wswebcreation: > > I believe you just fixed this "not implemented by WebDriver" with #13949, so I wonder if this "fix" is still needed.
> 
> I put up this patch because I am afraid we run into more issues where our mobile capability detection fails and causes a failure with the whole test run. I think this patch makes sense no matter what because we just disable the context manager in case the command fails.

Then can you add extra logs for debugging purposes so we can update the logic that is behind this when we get issues?

christian-bromann: What would you suggest aside from the `log.warn(`Error getting context: ${err}`)`?

wswebcreation: Maybe also throw the requested capabilities so we can determine if the mobile checks have made an error

christian-bromann: done

wswebcreation (approved):
Thanks!

PR #14075, opened 2025-01-15
feat(webdriverio): use checkVisibility for display checks

christian-bromann:
## Proposed changes

The web platform now provides a [`checkVisibility`](https://developer.mozilla.org/en-US/docs/Web/API/Element/checkVisibility#visibilityproperty) function to verify if an element is visible or not. It makes sense to rely on that implementation than maintaining one (that we forked from a different project) ourselves.

Furthermore this tweaks the function execution to expose the function name if available to make it easier to understand which WebdriverIO helper functions were called in the logs.

This "should" be a non breaking change but it is hard to say given that it is not much clear how the custom script worked.

## Types of changes

## Checklist

## Backport Request

## Further comments

n/a

### DIFF
--- a/__mocks__/fetch.ts
+++ b/__mocks__/fetch.ts
@@ -292,6 +292,8 @@ const requestMock: any = vi.fn().mockImplementation((uri, params) => {
             result = true
         } else if (body.script.includes('document.URL')) {
             result = 'https://webdriver.io/?foo=bar'
+        } else if (body.script.includes('function checkVisibility')) {
+            result = true
         } else {
             result = script.apply(this, args)
         }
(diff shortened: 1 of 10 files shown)


erwinheitzman:
Oh damn, nice that they standardised this!

I can run some tests with this if we release a "next" version or something like this. What do you think?

christian-bromann:
> I can run some tests with this if we release a "next" version or something like this. What do you think?

@erwinheitzman you can already via:
```sh
npm i https://pkg.pr.new/webdriverio/webdriverio@14075
```

nextlevelbeard:
@christian-bromann Great stuff! But I'm concerned about one thing:
I'm seeing that it is always making the check for the command availability every time `isDisplayed` is called.
It would perhaps be more performative if the check is made once when the session is created / attached / reloaded, and have the commands respect the setting.

christian-bromann:
> I'm seeing that it is always making the check for the command availability every time `isDisplayed` is called.
> It would perhaps be more performative if the check is made once when the session is created / attached / reloaded, and have the commands respect the setting.

@nextlevelbeard can you raise a new issue for this? It seems unrelated to this change given we just change how we find out whether an element is visible.

PR #14154, opened 2025-02-05
for in changed to for of for execArgv

udarrr:
The same like for V8 here https://github.com/webdriverio/webdriverio/pull/14150


christian-bromann (approved):
👍

PR #14166, opened 2025-02-07
[WDIO8] added supressing for runner's deleteSession

udarrr:
## Proposed changes

Until runner close worker, there is attempt to delete session in the end of worker's execution but if the session has been closed already for some reason whole execution is failed even though all tests are passed. I'd suggest to suppress delete session error as it's just finalizing step and don't relate to tests results

I'd suggest to change err.stack to err.message cause err.stack is redundant, there are lots of reference to internal libs no sure how it's helpful but looking like something important happened

### DIFF
--- a/packages/wdio-runner/src/index.ts
+++ b/packages/wdio-runner/src/index.ts
@@ -516,7 +516,16 @@ export default class Runner extends EventEmitter {
             })
         }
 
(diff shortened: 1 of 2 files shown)


christian-bromann (approved):
LGTM 👍

Issue #14167, opened 2025-02-07
[💡 Feature]: wildcards for --spec cli


udarrr:
### Is your feature request related to a problem?

wildcards aren't available in cli

### Describe the solution you'd like.

I'd suggest to change it like i did here in [WDIO8] branch https://github.com/webdriverio/webdriverio/commit/9376e3a02a4f906c07fd06515ebf7d8801633ed0

then anyone can use cli option --spec with wildcards like 

```
wdio --spec ./test/**/* --exclude ./test/login/**/*
```

For supporting groups like specs: [[]] in wdio config i'd suggest to add option --group 

```
wdio --spec ./test/**/* --exclude ./test/login/**/* --group
```

### Additional context

I haven't seen any issue with that so far, it seems working as expected as with single files as with wildcards

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

- [x] [implementation for wdio8](https://github.com/webdriverio/webdriverio/pull/14169) 

- [ ] implementation for wdio9

christian-bromann:
Thanks for raising the issue.

> wildcards aren't available in cli

What happens if you currently run `wdio --spec ./test/**/* --exclude ./test/login/**/*`?

I think this makes totally sense and is something I would expect a test runner to do, so let's make it happen!

udarrr:
> Thanks for raising the issue.
> 
> > wildcards aren't available in cli
> 
> What happens if you currently run `wdio --spec ./test/**/* --exclude ./test/login/**/*`?
> 
> I think this makes totally sense and is something I would expect a test runner to do, so let's make it happen!

So far it's 

![Image](https://github.com/user-attachments/assets/4c37b7a1-8546-4ec6-8ffe-2b5af589defb)

udarrr:
and it's confused because in wdio.conf.ts if the same put in specs: ['./test/**/*'] it will start tests, at this moment only single files for --spec supported no sure why

christian-bromann:
> no sure why

Sometimes there is no reason, sometimes it may have been a regression. So let's make sure we have proper tests for this. I recommend here to create a smoke tests that allows us to initiate the Launcher with proper arguments mimicking CLI args.

christian-bromann:
@udarrr thanks for taking this on and providing a fix.

PR #14169, opened 2025-02-07
[WDIO8] added supporting wildcars for --spec and --exclude, added --group cli…

udarrr:
## Proposed changes

making wildcards available in cli for --spec --exclude options

full description here https://github.com/webdriverio/webdriverio/issues/14167

- [x] change DOC files
- [x] add tests

### DIFF
--- a/packages/wdio-config/src/node/ConfigParser.ts
+++ b/packages/wdio-config/src/node/ConfigParser.ts
@@ -546,11 +546,5 @@ function allKeywordsContainPath(excludedSpecList: string[]) {
 }
 
 function filterEmptyArrayItems(specList: Spec[]) {
-    return specList.filter(item => {
-        if (Array.isArray(item) && item.length){
-            return item
-        } else if (!Array.isArray(item)){
-            return item
-        }
-    })
+    return specList.filter(item=>(Array.isArray(item) && item.length) || !Array.isArray(item))
 }
--- a/packages/wdio-config/src/node/ConfigParser.ts
+++ b/packages/wdio-config/src/node/ConfigParser.ts
@@ -504,7 +504,7 @@ export default class ConfigParser {
     filterSpecs(specs: Spec[], excludeList: string[]) {
         // If 'exclude' is array of paths
         if (allKeywordsContainPath(excludeList)) {
-            return specs.reduce((returnVal: Spec[], currSpec) => {
+            const filteredSpec = specs.reduce((returnVal: Spec[], currSpec) => {
                 if (Array.isArray(currSpec)) {
                     returnVal.push(currSpec.filter(specItem => !excludeList.includes(specItem)))
                 } else if (excludeList.indexOf(currSpec) === -1) {
@@ -512,9 +512,10 @@ export default class ConfigParser {
                 }
                 return returnVal
             }, [])
+            return filterEmptyArrayItems(filteredSpec)
         }
         // If 'exclude' is array of keywords
-        return specs.reduce((returnVal: Spec[], currSpec) => {
+        const filteredSpec = specs.reduce((returnVal: Spec[], currSpec) => {
             if (Array.isArray(currSpec)) {
                 returnVal.push(currSpec.filter(specItem => !excludeList.some(excludeVal => specItem.includes(excludeVal))))
             }
@@ -524,6 +525,7 @@ export default class ConfigParser {
             }
             return returnVal
         }, [])
+        return filterEmptyArrayItems(filteredSpec)
     }
 
     shard(specs: Spec[]) {
@@ -541,4 +543,14 @@ export default class ConfigParser {
 
 function allKeywordsContainPath(excludedSpecList: string[]) {
     return excludedSpecList.every(val => val.includes('/') || val.includes('\\'))
+}
+
+function filterEmptyArrayItems(specList: Spec[]) {
+    return specList.filter(item => {
+        if (Array.isArray(item) && item.length){
+            return item
+        } else if (!Array.isArray(item)){
+            return item
+        }
+    })
 }
--- a/packages/wdio-config/src/node/ConfigParser.ts
+++ b/packages/wdio-config/src/node/ConfigParser.ts
@@ -555,9 +555,9 @@ function filterDublicationArrayItems(specList: Spec[]) {
 
 function isValidRegex(expression: string) {
     try {
-      new RegExp(expression)
-      return true
+        new RegExp(expression)
+        return true
     } catch {
-      return false
+        return false
     }
 }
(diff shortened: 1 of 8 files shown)


packages/wdio-config/src/node/ConfigParser.ts
```diff
@@ -363,15 +364,15 @@ export default class ConfigParser {
...
         const fileList = ConfigParser.getFilePaths(specs, this._config.rootDir, this._pathService)
         cliArgFileList.forEach(filteredFile => {
             filteredFile = removeLineNumbers(filteredFile)
-            // Send single file/file glob to getFilePaths - not supporting hierarchy in spec/exclude
-            // Return value will always be string[]
+            // Send wildcard or single file glob to getFilePaths
+            // Return value will always be string[] or string [][]
             const globMatchedFiles = <string[]>ConfigParser.getFilePaths(
-                this._pathService.glob(filteredFile, path.dirname(this.#configFilePath)),
+                group ? [[filteredFile]] : [filteredFile],
```
erwinheitzman, line 375: Hi thanks for your contribution! There were good reasons why we started using glob, I wouldn't just change away from it and would rather like to see a fix that still uses glob.

udarrr: Hi there is glob as well inside of getFilePaths method, what was the reason to use it twice ?

packages/wdio-config/src/node/ConfigParser.ts
```diff
@@ -541,3 +544,13 @@ export default class ConfigParser {
 function allKeywordsContainPath(excludedSpecList: string[]) {
     return excludedSpecList.every(val => val.includes('/') || val.includes('\\'))
 }
+
+function filterEmptyArrayItems(specList: Spec[]) {
+    return specList.filter(item => {
+        if (Array.isArray(item) && item.length){
```
erwinheitzman, line 550: These return the exact same thing so there's no need to have an if/else

tests/tests-cli-spec-arg/mocha.test011.js
```diff
@@ -0,0 +1,7 @@
+import { expect } from '@wdio/globals'
+
+describe('This test', () => {
+    it('should be passed', async () => {
+        await expect(1).toEqual(1)
```
erwinheitzman, line 5: Why is this added?

erwinheitzman (changes requested):

packages/wdio-config/src/node/ConfigParser.ts
erwinheitzman: Ah I wasn't aware it was used twice. Then this is fine 👍

I did add two more comments but other than that it looks good to me

packages/wdio-config/src/node/ConfigParser.ts
udarrr: yes it could have been clearly, fixed

tests/smoke.runner.js
```diff
@@ -863,6 +864,123 @@ const cliExcludeParamValidationExcludeMultipleSpecsByPath = async () => {
     assert.strictEqual(skippedSpecs, 0)
     assert.strictEqual(failed, 0)
 }
+
+const cliSpecsWithWildCard = async () => {
+    const { passed, skippedSpecs, failed } = await launch(
+        'cliSpecsWithWildCard',
+        path.resolve(severalPassedConfig),
+        {
+            spec: ['./mocha.test*.js']
```
christian-bromann, line 873: The matching specs run a lot of tests, can we create new files that only contain a single hello world test so that we don't rerun the same tests over and over again?

christian-bromann (changes requested):

tests/smoke.runner.js
udarrr: No sure what you mean, cause the tests cases already are simple, all files under wildcard './mocha.test*.js' contain a test case that have a view

```
describe('This test', () => {
    it('should be passed', async () => {
        await expect(1).toEqual(1)
    })
})
```
but anyway i've decreased amount of them to 9 instead of 20 (selected only certain in folder instead of all) hope it'll be not so many so far

udarrr:
Hi @christian-bromann  @erwinheitzman  is there a chance to review it ?

packages/wdio-config/src/node/ConfigParser.ts
```diff
@@ -36,6 +36,7 @@ interface MergeConfig extends Omit<Partial<TestrunnerOptionsWithParameters>, 'sp
     'wdio:specs'?: Spec[]
     exclude?: string[]
     'wdio:exclude'?: string[]
+    group?: boolean
```
christian-bromann, line 39: Can we document this at all places necessary?

website/docs/OrganizingTestSuites.md
```diff
@@ -171,6 +171,12 @@ Or run multiple specs at once:
 wdio wdio.conf.js --spec ./test/specs/signup.js --spec ./test/specs/forgot-password.js
 ```
 
+Or run wildcards:
+
+```sh
+wdio wdio.conf.js --spec ./test/specs/\*\*/\*.js
```
christian-bromann, line 177: Would this work? It seems to look simpler
```suggestion
wdio wdio.conf.js --spec "./test/specs/**/*.js"
```

christian-bromann (commented):
Two last comments. We should make sure to document the new parameter in the CLI and in the docs.

erwinheitzman:
@udarrr ciuld you take a look at Christian's feedback?

website/docs/OrganizingTestSuites.md
udarrr: Done

packages/wdio-config/src/node/ConfigParser.ts
udarrr: Done

udarrr:
> @udarrr could you take a look at Christian's feedback?

yep having been done

christian-bromann (commented):
One last request, can we add the parameter docs also here: https://github.com/webdriverio/webdriverio/blob/5c9daf068d643b0cba23f46d83dcd718ef74a7d6/packages/wdio-cli/src/commands/run.ts#L89

Please feel free to also raise the PR for the `main` branch.

udarrr:
> One last request, can we add the parameter docs also here:
> 
> https://github.com/webdriverio/webdriverio/blob/5c9daf068d643b0cba23f46d83dcd718ef74a7d6/packages/wdio-cli/src/commands/run.ts#L89
> 
> Please feel free to also raise the PR for the `main` branch.

done, regarding main branch let's merge it to wdio8 and release if you don't mind and then i'll do the same in main branch after some time

christian-bromann:
@udarrr can we have the same PR for v9 now? I want to avoid that we forget about this.

udarrr:
> @udarrr can we have the same PR for v9 now? I want to avoid that we forget about this.

hi @erwinheitzman  i did create pr https://github.com/webdriverio/webdriverio/pull/14243 to main branch with all the changes but there are some regression issues that require some efforts, i'll do that shortly when will have capacity, as with wdio8 branch everything is ok i'd like to ask you to merge/release it that then i'll check it in real world

PR #14214, opened 2025-02-21
fix(webdriverio): fix types for the WebdriverIO getValue command and Webdriver getElementProperty command

erwinheitzman:
## Proposed changes

Since the getValue command returns the value property (with the exception of native context in which case it will return the value attribute) it can be of any type rather than only a string. This PR fixes this.

## Types of changes

## Checklist

### DIFF
--- a/packages/webdriverio/src/commands/element/getValue.ts
+++ b/packages/webdriverio/src/commands/element/getValue.ts
@@ -1,3 +1,5 @@
+import { getBrowserObject } from '@wdio/utils'
+
 /**
  *
  * Get the value of a `<textarea>`, `<select>` or text `<input>` found by given selector.
@@ -20,11 +22,13 @@
  * @uses protocol/elements, protocol/elementIdProperty
  *
  */
-export function getValue (this: WebdriverIO.Element) {
+export function getValue<T>(this: WebdriverIO.Element): Promise<T>
+export function getValue(this: WebdriverIO.Element): Promise<string>
+export function getValue<T>(this: WebdriverIO.Element) {
+    const browser = getBrowserObject(this)
     // `!this.isMobile` added to workaround https://github.com/appium/appium/issues/12218
-    if (this.isW3C && !this.isMobile) {
-        return this.getElementProperty(this.elementId, 'value')
+    if (browser.isNativeContext) {
+        return this.getElementAttribute(this.elementId, 'value') as Promise<string>
     }
-
-    return this.getElementAttribute(this.elementId, 'value')
+    return this.getElementProperty(this.elementId, 'value') as Promise<T>
 }
--- a/packages/wdio-protocols/src/protocols/webdriver.ts
+++ b/packages/wdio-protocols/src/protocols/webdriver.ts
@@ -820,7 +820,7 @@ export default {
             ],
             parameters: [],
             returns: {
-                type: 'string',
+                type: 'any',
                 name: 'property',
                 description:
                     'The named property of the element, accessed by calling GetOwnProperty on the element object.',
--- a/packages/webdriverio/src/commands/element/getProperty.ts
+++ b/packages/webdriverio/src/commands/element/getProperty.ts
@@ -15,7 +15,7 @@
  * @param {string} property  name of the element property
  * @return {unknown} the value of the property of the selected element
  */
-export function getProperty (
+export function getProperty(
     this: WebdriverIO.Element,
     property: string
 ): Promise<unknown> {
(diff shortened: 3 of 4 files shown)


packages/webdriverio/src/commands/element/getValue.ts
```diff
@@ -20,11 +22,13 @@
  * @uses protocol/elements, protocol/elementIdProperty
  *
  */
-export function getValue (this: WebdriverIO.Element) {
+export function getValue<T>(this: WebdriverIO.Element): Promise<T>
```
christian-bromann, line 25: The `getValue` command is meant to receive the value for an input element which is, in the browser case, always a string. Can you describe the native mobile case?

packages/webdriverio/src/commands/element/getValue.ts
```diff
@@ -20,11 +22,13 @@
...
+export function getValue(this: WebdriverIO.Element): Promise<string>
+export function getValue<T>(this: WebdriverIO.Element) {
```
christian-bromann, line 27: I think the above overwrites:
```ts
export function getValue<T>(this: WebdriverIO.Element): Promise<T>
export function getValue(this: WebdriverIO.Element): Promise<string>
```
don't make a difference because TypeScript won't be able to resolve the types pre-runtime. I would suggest to default to `string` but allow users to pass in a type when they do mobile native testing.
```suggestion
export function getValue<T = string>(this: WebdriverIO.Element) {
```

christian-bromann (changes requested):

packages/webdriverio/src/commands/element/getValue.ts
erwinheitzman: This is not true, the command returns the value property which can be anything.

You can verify this by going to https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_input_test
Open the devtools, select one on the input elements and then with the element selected, input the following into the console: `$0.value = 2`

The value property will be a number but the value attribute will be a strin representation of it.

Furthermore this function also works on custom elements that have a value property which again can hold any data, even big objects.

packages/webdriverio/src/commands/element/getValue.ts
```diff
@@ -20,11 +22,13 @@
...
+    const browser = getBrowserObject(this)
     // `!this.isMobile` added to workaround https://github.com/appium/appium/issues/12218
-    if (this.isW3C && !this.isMobile) {
-        return this.getElementProperty(this.elementId, 'value')
+    if (browser.isNativeContext) {
```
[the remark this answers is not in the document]
erwinheitzman: No, as @wswebcreation and I discussed in Discord the command works as expected for web and mobile web but not for the native context.

Therefore I made these changes to reflect that.

packages/webdriverio/src/commands/element/getValue.ts
erwinheitzman: You are right, I stashed some code and this accidentally seems to be the old code that I had instead of what I intended. I'll update this thanks!

erwinheitzman: I double checked this and it's actually my intention to create a TypeScript function overload, this doesn't work?

I'm not sure what you mean by pre-runtime but I overwrote the type in the node_modules in one of my projects and that did seem to work.

Reference: https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads

packages/webdriverio/src/commands/element/getValue.ts
christian-bromann: This is what I observe:
Screenshot 2025-02-21 at 12 15 31 PM

So in the browser the `value` will be always a `string`. Even if I change the value to a number:

Screenshot 2025-02-21 at 12 17 15 PM

Note: there are two commands: `getValue` and `getProperty`. The `getValue` command will always return a `string` in the browser while `getProperty` can return any type.

packages/webdriverio/src/commands/element/getValue.ts
christian-bromann: Overloading the function doesn't have an effect here because TypeScript doesn't know if the browser instance variable is running a browser session or a native mobile session. TypeScript can not resolve types at runtime when we know more about these variables. If you could create some type tests you will see that you always get the same return type from TypeScript.

packages/webdriverio/src/commands/element/getValue.ts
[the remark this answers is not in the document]
christian-bromann: I don't understand the reasoning behind this change then. In the browser this command (`getValue`) has to return a string (see convo above). In the mobile case I don't know. Can you provide more context?

PR #14215, opened 2025-02-21
fix(webdriverio): fix failing e2e tests due to language and timezone differences

erwinheitzman:
## Proposed changes

Fix failing e2e tests due to differences in languages and timezones.

## Types of changes

## Checklist

### DIFF
--- a/e2e/wdio/headless/reloadSession.e2e.ts
+++ b/e2e/wdio/headless/reloadSession.e2e.ts
@@ -1,12 +1,7 @@
-import os from 'node:os'
 import { browser, expect } from '@wdio/globals'
 
 describe('reloadSession', () => {
     it('can reload a session with new capabilities', async () => {
-        if (os.platform() !== 'win32') {
-            return
-        }
-
         const sessionId = browser.sessionId
         expect(browser.capabilities.browserName).toBe('chrome')
         await browser.reloadSession({
--- a/e2e/wdio/headless/reloadSession.e2e.ts
+++ b/e2e/wdio/headless/reloadSession.e2e.ts
@@ -1,7 +1,12 @@
+import os from 'node:os'
 import { browser, expect } from '@wdio/globals'
 
 describe('reloadSession', () => {
     it('can reload a session with new capabilities', async () => {
+        if (os.platform() !== 'win32') {
+            return
+        }
+
         const sessionId = browser.sessionId
         expect(browser.capabilities.browserName).toBe('chrome')
         await browser.reloadSession({
--- a/e2e/wdio/wdio.local.conf.ts
+++ b/e2e/wdio/wdio.local.conf.ts
@@ -39,6 +39,12 @@ export const config: WebdriverIO.Config = {
         'moz:firefoxOptions': {
             args: ['-headless']
         }
+    }, {
+        browserName: 'edge',
+        webSocketUrl: true,
+        'ms:edgeOptions': {
+            args: ['headless', 'disable-gpu']
+        }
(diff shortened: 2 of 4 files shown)


e2e/wdio/headless/reloadSession.e2e.ts
```diff
@@ -1,7 +1,12 @@
+import os from 'node:os'
 import { browser, expect } from '@wdio/globals'
 
 describe('reloadSession', () => {
     it('can reload a session with new capabilities', async () => {
+        if (os.platform() !== 'win32') {
```
christian-bromann, line 6: Why do we skip the test for Windows machines?

e2e/wdio/wdio.local.conf.ts
```diff
@@ -70,6 +64,16 @@ if (os.platform() === 'darwin') {
     })
 }
 
+if (os.platform() === 'win32') {
```
christian-bromann, line 67: Why only running tests in Edge on Windows?

christian-bromann (commented):
Can you show me some pipelines that have been failing due to these issues?

erwinheitzman:
Hi Christian, it's not pipelines but running locally is an issue when your browser is in another language and in another time zone. Also using Linux you don't have access to Edge and even though you can use it on MacOS I think the chances are slim anyone has it installed so I limited it to Windows only since our pipeline runs them in Windows anyway

erwinheitzman:
Perhaps I shouldn't have called them flaky as they of course fail consistently on certain operating systems, browsers in other languages and for people in another time zone

erwinheitzman:
Perhaps I shouldn't have called them flaky as they of course fail consistently on certain operating systems, browsers in other languages and for people in another time zone

erwinheitzman:
Perhaps I shouldn't have called them flaky as they of course fail consistently on certain operating systems, browsers in other languages and for people in another time zone

christian-bromann:
Oh, I understand, I don't have a Windows machine and have no idea if these tests pass there with other types of configurations. Maybe we should document requirements to run these tests then in the contribution guidelines. For example `e2e/wdio/wdio.local.conf.ts` is meant to run on all systems as it tests whether WebdriverIO can start a session in all of these browser.

I would assume these tests, except for the ones with the timezome, should also pass for you. What error message are you getting?

erwinheitzman:
Yeah we can make it a requirement that Chrome, Firefox and Edge need to be installed with the exception of Edge on Linux as Linux isn't supported.

What tests are you asking about specifically?

Regarding the Edge browser it complained it wasn't installed.

Regarding the language it showed the title for the Chrome version page in Dutch instead of Engels.

Regarding the time zone it showed a different time.

christian-bromann:
> What tests are you asking about specifically?

The test run via `./wdio/wdio.local.conf.ts` through `test:e2e:webdriver` is a test that should run on all browser. It is meant to verify that WebdriverIO can start/download browser in CI.

> Regarding the Edge browser it complained it wasn't installed.

I think the test should fail then. We should better document in the contribution guidelines that we expect all browser to be installed when someone wants to run that test.

Both other test changes look good to me.

erwinheitzman:
But what if you are on Linux and cannot install Edge (Edge does not support Linux)?

erwinheitzman:
@christian-bromann I've reverted the mentioned changes and have added the requirements in the CONTRIBUTING.md file, thanks for the review!

christian-bromann (approved):
LGTM 👍 

Nice job!

PR #14219, opened 2025-02-23
polish: performance improvements and some general code improvements

erwinheitzman:
## Proposed changes

- Applied Promise concurrency
- Merged .map and .filter into one when possible
- Merged multiple .filter into one by using early returns
- Replaced regex matching for regex testing (returning a match vs returning a boolean)
- Added TODO's for some code that might not be correct

## Types of changes

## Checklist

### DIFF
--- a/packages/wdio-cli/src/utils.ts
+++ b/packages/wdio-cli/src/utils.ts
@@ -446,7 +446,7 @@ async function generateLocalRunnerTestFiles(answers: ParsedAnswers) {
         [(file, stats) => !stats.isDirectory() && !(file.endsWith('.ejs') || file.endsWith('.feature'))]
     )))).reduce((cur, acc) => [...acc, ...(cur)], [])
 
-    for (const file of files) {
+    await Promise.all(files.map(async (file) => {
         const renderedTpl = await renderFile(file, { answers })
         const isJSX = answers.preset && TSX_BASED_FRAMEWORKS.includes(answers.preset)
         const fileEnding = (answers.isUsingTypeScript ? '.ts' : '.js') + (isJSX ? 'x' : '')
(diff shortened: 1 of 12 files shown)


packages/wdio-runner/src/browser.ts
```diff
@@ -103,9 +103,10 @@ export default class BrowserFramework implements Omit<TestFramework, 'init'> {
          * start tests
          */
         let failures = 0
-        for (const spec of this._specs) {
+
+        await Promise.all(this._specs.map(async (spec) => {
```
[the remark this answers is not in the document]
christian-bromann: The outlined code runs the component tests and in cases users group specs, e.g.

```ts
// wdio.conf.ts
export const config: WebdriverIO.Config = {
  // ...
  specs: [["specA", "specB"]]
  //
}
```
We run a set of component test within a single browser session. This means we open the test page, run the component test, collect the results and go to the next page. The reason why the test pass for this PR is because we don't have any test groups in our component test pipeline.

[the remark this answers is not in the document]
erwinheitzman: Thanks I'll revert the change 👍

packages/wdio-cli/src/utils.ts
```diff
@@ -877,19 +877,18 @@ export async function setupTypeScript(parsedAnswers: ParsedAnswers) {
...
-                service.startsWith('@wdio') ||
+                service.package.startsWith('@wdio') ||
                 /**
                  * also include community maintained packages with known
                  * support for TypeScript
                  */
-                COMMUNITY_PACKAGES_WITH_TS_SUPPORT.includes(service)
-            ))
+                COMMUNITY_PACKAGES_WITH_TS_SUPPORT.includes(service.package)
+            )).map(service => service.package)
```
wswebcreation, line 891: I'm not fully understanding why we need to do all the "code changes" here instead of or keeping it like it was or renaming stuff?

packages/wdio-runner/src/browser.ts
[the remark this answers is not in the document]
wswebcreation: @erwinheitzman , maybe add that as a comment in the code so we understand happens there

packages/webdriverio/src/commands/browser/switchWindow.ts
```diff
@@ -63,10 +63,10 @@ export async function switchWindow (
     }
 
     const matchesTarget = (target: string): boolean => {
-        if (typeof matcher ==='string') {
+        if (typeof matcher === 'string') {
             return target.includes(matcher)
         }
-        return !!target.match(matcher)
```
wswebcreation, line 69: Nice, #todayilearned

packages/webdriverio/src/utils/findStrategy.ts
```diff
@@ -44,7 +44,7 @@ const defineStrategy = function (selector: SelectorStrategy) {
 
     const stringSelector = selector as string
     // Check if user has specified locator strategy directly
-    if (stringSelector.match(DIRECT_SELECTOR_REGEXP)) {
```
wswebcreation, line 47: Same, #todayilearned

wswebcreation (commented):
Nice improvements @erwinheitzman

packages/wdio-cli/src/utils.ts
erwinheitzman: Moving .map to after the filter makes for less iterations. In order to achieve this I have to made some small changes

christian-bromann (approved):
#todayilearned 

Great work 👌

PR #14225, opened 2025-02-24
Fix watch files documentation

jbblanchet:
## Proposed changes

Fix documentation to account for change in chokidar not supporting globing anymore.

This was done in this PR: https://github.com/webdriverio/webdriverio/pull/13583

See changelog of chokidar: https://github.com/paulmillr/chokidar/releases/tag/4.0.0

## Types of changes

## Checklist

## Backport Request

## Further comments

There are other options. You could roll back the change to chokidar, or handle the globing before calling chokidar here: https://github.com/webdriverio/webdriverio/blob/main/packages/wdio-cli/src/watcher.ts#L53.

I do not know enough about the project to know if these options are viable, so I did the minimum which was updating documentation to reflect the current state of the code.

Also didn't know if I should include the call to `glob` in the example, but I don't think there was a better way to still keep an example, even if it depends on a 3rd party library now.

### DIFF
--- a/website/docs/WatchFiles.md
+++ b/website/docs/WatchFiles.md
@@ -9,16 +9,13 @@ With the WDIO testrunner you can watch files while you are working on them. They
 wdio wdio.conf.js --watch
 ```
 
-By default it only watches for changes in your `specs` files. However by setting a `filesToWatch` property in your `wdio.conf.js` that contains a list of file paths (globbing supported) it will also watch for these files to be changed in order to rerun the whole suite. This is useful if you want to automatically rerun all your tests if you have changed your application code, e.g.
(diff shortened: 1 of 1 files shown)


christian-bromann:
@jbblanchet thanks for raising the issue!

Since we have `glob` as dependency of `@wdio/config` we could just support this out of the box:

```ts
import { FileSystemPathService } from '@wdio/config'

const service = new FileSystemPathService()
return service.glob(pattern)
```

Would you be open implement that?

jbblanchet:
@christian-bromann Sorry I won't have time to do that. Feel free to close this PR if someone decides to implement the change. I resolved it pretty quickly on my side by just globing the files directly in my config, I'm unclear what the value proposition would be to implement the globing in wdio itself honestly, beside backwards compatibility. And since this was broken months ago and no one reported it, I'm assuming not a lot of people are depending on it.

Issue #13377, opened 2024-08-16
[🐛 Bug]: v9 fails in non-headless, incognito mode


zarko-tg:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

9

### Node.js Version

20.16.0

### Mode

WDIO Testrunner

### Which capabilities are you using?

```typescript
{
    browserName: 'chrome',
    'goog:chromeOptions': {
        args: ['incognito', 'disable-gpu']
    }
}
```

### What happened?

It seems that all tests fail when running non-headless and in incognito mode. The browser window(s) show and disappear quickly, sort of flashing. I used a macOS (14.6.1).

npm run test:local

> jasmine-boilerplate@1.0.0 test:local
> wdio run wdio.local.conf.ts

Execution of 3 workers started at 2024-08-16T14:12:50.600Z

[0-1] RUNNING in chrome - file:///test/specs/dynamic.spec.ts
[0-2] RUNNING in chrome - file:///test/specs/form.spec.ts
[0-0] RUNNING in chrome - file:///test/specs/checkbox.spec.ts
[0-1] FAILED in chrome - file:///test/specs/dynamic.spec.ts
[0-2] FAILED in chrome - file:///test/specs/form.spec.ts
[0-0] FAILED in chrome - file:///test/specs/checkbox.spec.ts

Spec Files:	 0 passed, 3 failed, 3 total (100% completed) in 00:00:05

### What is your expected behavior?

Regular execution and passing tests.

### How to reproduce the bug.

1. Clone https://github.com/webdriverio/jasmine-boilerplate and npm install.

2. Modify
diff --git a/wdio.local.conf.ts b/wdio.local.conf.ts
index 9a5ed83..83528fe 100644
--- a/wdio.local.conf.ts
+++ b/wdio.local.conf.ts
@@ -6,7 +6,7 @@ export const config: WebdriverIO.Config = {
         capabilities: [{
             browserName: 'chrome',
             'goog:chromeOptions': {
-                args: ['headless', 'disable-gpu']
+                args: ['incognito', 'disable-gpu']
             }
         }]
     }

3. npm run test:local

### Relevant log output

```typescript
**wdio.log:**

2024-08-16T14:12:50.606Z INFO @wdio/cli:launcher: Run onPrepare hook
[...]

PR #14233, opened 2025-02-27
Fixed the logic to handle non-Bidi mode.

aswinchembath:


christian-bromann:
@aswinchembath how is this behavior different in Chrome between WebDriver Classic and Bidi

packages/webdriver/src/utils.ts
```diff
@@ -111,7 +111,7 @@ export async function startWebDriverSession (params: RemoteConfig): Promise<{ se
  */
 export function validateCapabilities (capabilities: WebdriverIO.Capabilities) {
     const chromeArgs = capabilities['goog:chromeOptions']?.args || []
-    if (chromeArgs.includes('incognito') || chromeArgs.includes('--incognito')) {
+    if ((chromeArgs.includes("incognito") || chromeArgs.includes("--incognito") )&& !capabilities["wdio:enforceWebDriverClassic"]) {
```
erwinheitzman, line 114:
```suggestion
    if ((chromeArgs.includes('incognito') || chromeArgs.includes('--incognito')) && !capabilities['wdio:enforceWebDriverClassic']) {
```

erwinheitzman:
@christian-bromann could this be related? People shared some intesting things after it had been closed: https://github.com/webdriverio/webdriverio/issues/13377

christian-bromann:
Yes this may be related but again there is no clear value proposition to running in incognito mode vs running with a clear profile. We have disabled passing through the `--incognito` parameter to Chrome because it caused issues with WebDriver.

eyesopen:
My observation is quite the opposite. I had multiple failures since upgrading to WDIO 9. Sessions were unidentified and failing to create. I disabled BiDi and had results more or less back to normal, but the  prompts from chrome password manager, retained cache and similar are still causing a lot of troubles. Certain reported bugs in our system i was able to reproduce in incognito browser yet automation running same steps failed. Incognito is precious and i have not experienced any issues while in incognito before upgrade. Even if this argument is too weak to consider putting incognito back, can this choice be left to the users albeit highly discouraged?

> Yes this may be related but again there is no clear value proposition to running in incognito mode vs running with a clear profile. We have disabled passing through the `--incognito` parameter to Chrome because it caused issues with WebDriver.

christian-bromann:
> Certain reported bugs in our system i was able to reproduce in incognito browser yet automation running same steps failed.

Can you provide more information here? I would like to understand more about the differences between running Chrome in a fresh profile vs in incognito mode.

aswinchembath:
We encountered a failure with an SSO-based application and need to switch to test accounts, which will prompt for basic authentication only when accessed through Incognito mode. However, with BiDi, the browser doesn’t open like Incognito mode, where SSO is already authenticated with my default account. This makes switching to test accounts impossible after updating to v9. This is one of the issue which am facing right now . Not sure someone else opened any defect similar to this

christian-bromann:
I reported this in Chrome: https://github.com/GoogleChromeLabs/chromium-bidi/issues/3189.

In regards to this PR, once the pipeline passes and tests have been added I am happy to merge.

christian-bromann:
@aswinchembath any updates? Do we still want to move forward with this?

christian-bromann:
Closing due to inactivity.

Issue #14039, opened 2025-01-08
[🐛 Bug]: BiDi URL IP 4/6 resolution issue


mykola-mokhnach:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

latest

### Node.js Version

22

### Mode

Standalone Mode

### Which capabilities are you using?

```typescript
{"capabilities":{"acceptInsecureCerts":false,"browserName":"chrome","browserVersion":"131.0.6778.140","chrome":{"chromedriverVersion":"131.0.6778.108 (3b014839fbc4fb688b2f5af512d6ce312ad208b1-refs/branch-heads/6778@{#2393})","userDataDir":"C:\\Windows\\SystemTemp\\scoped_dir3608_1127547173"},"fedcm:accounts":true,"goog:chromeOptions":{"debuggerAddress":"localhost:55664"},"networkConnectionEnabled":false,"pageLoadStrategy":"normal","platformName":"windows","proxy":{},"setWindowRect":true,"strictFileInteractability":false,"timeouts":{"implicit":0,"pageLoad":300000,"script":30000},"unhandledPromptBehavior":"ignore","webSocketUrl":"ws://fv-az1760-623:4780/bidi/d5116166-770a-46f0-a4f9-c3d744df3237","webauthn:extension:credBlob":true,"webauthn:extension:largeBlob":true,"webauthn:extension:minPinLength":true,"webauthn:extension:prf":true,"webauthn:virtualAuthenticators":true}}
```

### What happened?

Check failed tests in https://github.com/appium/appium-chromium-driver/actions/runs/12668118313/job/35302892004?pr=380

When BiDi URL returned by the server contains a hostname (`ws://fv-az1760-623:4780/bidi/d5116166-770a-46f0-a4f9-c3d744df3237`), which could be resolved in both ip4 and ip6 addresses, WDIO only tries one of them (ipv6) and fails instead of trying both.

### What is your expected behavior?

WDIO should fallback to ipv4 resolution if it fails to reach the web socket using ipv6

### How to reproduce the bug.

Run tests from https://github.com/appium/appium-chromium-driver/pull/380

### Relevant log output

```typescript
https://github.com/appium/appium-chromium-driver/actions/runs/12668118313/job/35302892004?pr=380
```

### Code of Conduct

[...]

PR #14259, opened 2025-03-10
fix: Try to resolve ip addresses if no BiDi connection to the host could be established

mykola-mokhnach:
## Proposed changes

Addresses https://github.com/webdriverio/webdriverio/issues/14039

## Types of changes

## Checklist

## Backport Request

## Further comments

### DIFF
--- a/packages/webdriver/src/bidi/core.ts
+++ b/packages/webdriver/src/bidi/core.ts
@@ -90,7 +90,7 @@ export class BidiCore {
             }
             const wsMapping = (await Promise.all(wsConnectPromises))
                 .map((ws, index) => [candidateUrls[index], ws]) as [string, WebSocket | null][]
-            const result = wsMapping.find(([, ws]) => Boolean(ws)) as [string, WebSocket]
+            const result = wsMapping.find(([, ws]) => Boolean(ws)) as [string, WebSocket] | undefined
             // Cleanup extra opened connections
             for (const item of wsMapping) {
                 if (item !== result && item[1]) {
--- a/packages/webdriver/src/bidi/core.ts
+++ b/packages/webdriver/src/bidi/core.ts
@@ -64,8 +64,14 @@ export class BidiCore {
                 dns.promises.resolve4(parsedUrl.hostname),
                 dns.promises.resolve6(parsedUrl.hostname),
             ])
-            const urlMapper = (ip: string) => this.#webSocketUrl.replace(parsedUrl.hostname, ip)
-            candidateUrls.push(...ips4.map(urlMapper), ...ips6.map(urlMapper))
+            const candiadateIps = [...ips4, ...ips6]
+            // If the host resolves to a single IP address
+            // then it does not make sense to try additional candidates
+            // as the web socket DNS resolver would do extactly the same
+            if (candiadateIps.length > 1) {
+                const hostnameMapper = (ip: string) => this.#webSocketUrl.replace(parsedUrl.hostname, ip)
+                candidateUrls.push(...ips4.map(hostnameMapper), ...ips6.map(hostnameMapper))
+            }
         }
 
         this.#waitForConnected = (async () => {
--- a/packages/webdriver/src/bidi/core.ts
+++ b/packages/webdriver/src/bidi/core.ts
@@ -1,5 +1,7 @@
 import logger from '@wdio/logger'
 import type { ClientOptions, RawData, WebSocket } from 'ws'
+import { isIP } from 'node:net'
+import dns from 'node:dns'
 
 import { environment } from '../environment.js'
 import type * as remote from './remoteTypes.js'
@@ -17,9 +19,10 @@ const RESPONSE_TIMEOUT = 1000 * 60
 
 export class BidiCore {
     #id = 0
-    #ws: WebSocket
+    #ws: WebSocket | null = null
     #waitForConnected = Promise.resolve(false)
     #webSocketUrl: string
+    #clientOptions: ClientOptions | undefined
     #pendingCommands: Map<number, (value: CommandResponse) => void> = new Map()
 
     client: Client | undefined
@@ -30,9 +33,7 @@ export class BidiCore {
 
     constructor (webSocketUrl: string, opts?: ClientOptions) {
         this.#webSocketUrl = webSocketUrl
-        log.info(`Connect to webSocketUrl ${this.#webSocketUrl}`)
-        this.#ws = new environment.value.Socket(this.#webSocketUrl, opts) as unknown as WebSocket
-        this.#ws.on('message', this.#handleResponse.bind(this))
+        this.#clientOptions = opts
     }
 
     /**
@@ -54,17 +55,43 @@ export class BidiCore {
             return
         }
 
+        log.info(`Connecting to webSocketUrl ${this.#webSocketUrl}`)
+        const parsedUrl = new URL(this.#webSocketUrl)
+        // https://github.com/webdriverio/webdriverio/issues/14039
+        const candidateUrls: string[] = [this.#webSocketUrl]
+        if (!isIP(parsedUrl.hostname)) {
+            const [ips4, ips6] = await Promise.all([
+                dns.promises.resolve4(parsedUrl.hostname),
+                dns.promises.resolve6(parsedUrl.hostname),
+            ])
+            const candidateIps = [...ips4, ...ips6]
+            candidateUrls.push(
+                ...candidateIps.map((ip) => this.#webSocketUrl.replace(parsedUrl.hostname, ip))
+            )
+        }
+
         this.#waitForConnected = new Promise<boolean>((resolve) => {
-            this.#ws.on('open', () => {
-                log.info('Connected session to Bidi protocol')
-                this._isConnected = true
-                resolve(this._isConnected)
-            })
-            this.#ws.on('error', (err) => {
-                log.warn(`Couldn't connect to Bidi protocol: ${err.message}`)
-                this._isConnected = false
-                resolve(this._isConnected)
-            })
+            for (const candidateUrl of candidateUrls) {
+                const ws = new environment.value.Socket(this.#webSocketUrl, this.#clientOptions) as unknown as WebSocket
+                ws.once('open', () => {
+                    if (this._isConnected) {
+                        return
+                    }
+                    log.info(`Connected session to Bidi protocol at ${candidateUrl}`)
+                    this._isConnected = true
+                    this.#ws = ws
+                    this.#ws.on('message', this.#handleResponse.bind(this))
+                    resolve(this._isConnected)
+                })
+                ws.once('error', (err) => {
+                    if (this._isConnected) {
+                        return
+                    }
+                    log.warn(`Couldn't connect to Bidi protocol: ${err.message}`)
+                    this._isConnected = false
+                    resolve(this._isConnected)
+                })
+            }
         })
         return this.#waitForConnected
     }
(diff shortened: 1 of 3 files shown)


packages/webdriver/src/bidi/core.ts
```diff
@@ -54,18 +55,52 @@ export class BidiCore {
...
+                        }
+                        resolve(null)
+                    })
+                })
+                wsConnectPromises.push(connectPromise)
+            }
+            const wsMapping = (await Promise.all(wsConnectPromises))
+                .map((ws, index) => [candidateUrls[index], ws])
+                .find(([, ws]) => Boolean(ws)) as [string, WebSocket] | undefined
+            if (wsMapping) {
```
mykola-mokhnach, line 94: not sure if we need to close "extra" opened sockets explicitly here or just rely on the garbage collector

packages/webdriver/src/bidi/core.ts
```diff
@@ -54,18 +55,52 @@ export class BidiCore {
...
+        log.info(`Connecting to webSocketUrl ${this.#webSocketUrl}`)
+        const parsedUrl = new URL(this.#webSocketUrl)
+        // https://github.com/webdriverio/webdriverio/issues/14039
+        const candidateUrls: string[] = [this.#webSocketUrl]
+        if (!isIP(parsedUrl.hostname)) {
+            const [ips4, ips6] = await Promise.all([
+                dns.promises.resolve4(parsedUrl.hostname),
+                dns.promises.resolve6(parsedUrl.hostname),
+            ])
+            const candidateIps = [...ips4, ...ips6]
```
erwinheitzman, line 67: ips4 and ips6 are never used
```suggestion
            const candidateIps = await Promise.all([
                dns.promises.resolve4(parsedUrl.hostname),
                dns.promises.resolve6(parsedUrl.hostname),
            ])
```

packages/webdriver/src/bidi/core.ts
```diff
@@ -54,18 +55,52 @@ export class BidiCore {
...
+            candidateUrls.push(
+                ...candidateIps.map((ip) => this.#webSocketUrl.replace(parsedUrl.hostname, ip))
+            )
```
erwinheitzman, line 70: No need to spread everything into the array, concat makes more sense here
```suggestion
            candidateUrls.concat(
                candidateIps.map((ip) => this.#webSocketUrl.replace(parsedUrl.hostname, ip))
            )
```

packages/webdriver/src/bidi/core.ts
```diff
@@ -54,18 +55,52 @@ export class BidiCore {
...
+
+        this.#waitForConnected = (async () => {
+            const wsConnectPromises: Promise<WebSocket | null>[] = []
+            for (const candidateUrl of candidateUrls) {
+                const connectPromise = new Promise<WebSocket | null>((resolve) => {
+                    const ws = new environment.value.Socket(candidateUrl, this.#clientOptions) as unknown as WebSocket
+                    ws.once('open', () => resolve(ws))
+                    ws.once('error', (err) => {
+                        const message = `Couldn't connect to Bidi protocol at ${candidateUrl}: ${err.message}`
+                        if (candidateUrls.length > 1) {
```
erwinheitzman, line 81: why differentiate between the two here?

packages/webdriver/src/bidi/core.ts
```diff
@@ -54,18 +55,52 @@ export class BidiCore {
...
+            candidateUrls.push(
+                ...candidateIps.map((ip) => this.#webSocketUrl.replace(parsedUrl.hostname, ip))
+            )
+        }
+
+        this.#waitForConnected = (async () => {
+            const wsConnectPromises: Promise<WebSocket | null>[] = []
+            for (const candidateUrl of candidateUrls) {
+                const connectPromise = new Promise<WebSocket | null>((resolve) => {
+                    const ws = new environment.value.Socket(candidateUrl, this.#clientOptions) as unknown as WebSocket
```
erwinheitzman, line 77: Could you try getting the correct type without type casting?

packages/webdriver/src/bidi/core.ts
```diff
@@ -54,18 +55,52 @@ export class BidiCore {
...
+                    })
+                })
+                wsConnectPromises.push(connectPromise)
+            }
+            const wsMapping = (await Promise.all(wsConnectPromises))
+                .map((ws, index) => [candidateUrls[index], ws])
+                .find(([, ws]) => Boolean(ws)) as [string, WebSocket] | undefined
+            if (wsMapping) {
+                log.info(`Connected session to Bidi protocol at ${wsMapping[0]}`)
+                this.#ws = wsMapping[1]
```
erwinheitzman, line 96: Could you explain a bit why you need wsMapping[1] instead of wsMapping[0] here? It's not obvious for me right away

erwinheitzman (commented):
Hi, thank you for your contribution!

Looking great, I've added some feedback and questions.

packages/webdriver/src/bidi/core.ts
mykola-mokhnach: ips4 and ips6 are both array of strings. If the above change is made then `candidateIps` type would be `string[][]`, while we want it to be `string[]`. This would work if we could use something like  [flatten](https://www.geeksforgeeks.org/lodash-_-flatten-method/) to the result of Promise.all, although I don't see lodash being present in module dependencies.

packages/webdriver/src/bidi/core.ts
mykola-mokhnach: concat does not mutate arrays, but rather returns catenated one. I prefer to mutate the existing array though to reduce the amount of boilerplace code needed to declare intermediate variables

packages/webdriver/src/bidi/core.ts
mykola-mokhnach: I've changed this logic anyway

packages/webdriver/src/bidi/core.ts
mykola-mokhnach: I am not sure this is possible. It was already done like that before, I did not change the implementation.
Do you have any proposals on how this particular line could be improved?

packages/webdriver/src/bidi/core.ts
mykola-mokhnach: Sure, I've also added explicit typecasts to make it more obvious, and slightly updated the code since the last time:

```js
           // here we map all connected web sockets or nulls if there was a connection error to their particular urls
           // As a result we have array of pairs, where each item looks like [<socket url>, <socket or null>]
           const wsMapping = (await Promise.all(wsConnectPromises))
                .map((ws, index) => [candidateUrls[index], ws]) as [string, WebSocket | null][]
           // here we try to find the very first connected url and web socket pair, and remember it as the method result
            const result = wsMapping.find(([, ws]) => Boolean(ws)) as [string, WebSocket]
            // this is needed to make sure we don't keep any extra opened connections in the memory
            // in case the websocket was listening on multiple interfaces
            // Cleanup extra opened connections
            for (const item of wsMapping) {
                if (item !== result && item[1]) {
                    item[1].close()
                }
            }
            if (result) {
                // if the match was found then use it. The first pair item is the matched url and the second one is the 
                // web socket instance
                log.info(`Connected session to Bidi protocol at ${result[0]}`)
                this.#ws = result[1]
                this.#ws.on('message', this.#handleResponse.bind(this))
                this._isConnected = true
            } else {
```

packages/webdriver/src/bidi/core.ts
```diff
@@ -1,5 +1,7 @@
 import logger from '@wdio/logger'
 import type { ClientOptions, RawData, WebSocket } from 'ws'
+import { isIP } from 'node:net'
+import dns from 'node:dns'
```
christian-bromann, line 4: Since we only use the promise primitives anyway.
```suggestion
import dns from 'node:dns/promises'
```

packages/webdriver/src/bidi/core.ts
```diff
@@ -17,9 +19,10 @@ const RESPONSE_TIMEOUT = 1000 * 60
 
 export class BidiCore {
     #id = 0
-    #ws: WebSocket
+    #ws: WebSocket | null = null
```
christian-bromann, line 22: I am trying to avoid using `null` as much as possible. 
```suggestion
    #ws?: WebSocket
```

packages/webdriver/src/bidi/core.ts
christian-bromann: Luckily these primitives are now [part](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat) of JavaScript
```suggestion
            const candidateIps = (await Promise.all([
                dns.promises.resolve4(parsedUrl.hostname),
                dns.promises.resolve6(parsedUrl.hostname),
            ])).flat()
```

packages/webdriver/src/bidi/core.ts
```diff
@@ -54,18 +55,61 @@ export class BidiCore {
...
+            ])
+            const candidateIps = [...ips4, ...ips6]
+            // If the host resolves to a single IP address
+            // then it does not make sense to try additional candidates
+            // as the web socket DNS resolver would do extactly the same
+            if (candidateIps.length > 1) {
+                const hostnameMapper = (ip: string) => this.#webSocketUrl.replace(parsedUrl.hostname, ip)
+                candidateUrls.push(...candidateIps.map(hostnameMapper))
+            }
+        }
```
christian-bromann, line 75: For testing purposes it makes sense to wrap everything from your comment to here into a function like `getCandidateUrls`

packages/webdriver/src/bidi/core.ts
```diff
@@ -54,18 +55,61 @@ export class BidiCore {
...
+
+        this.#waitForConnected = (async () => {
```
christian-bromann, line 77: Same here, let's move this into a utility function.

packages/webdriver/src/bidi/core.ts
christian-bromann: This cast already exist before this patch, happy to improve it but can remain outside of the PR

packages/webdriver/src/bidi/core.ts
```diff
@@ -54,18 +55,61 @@ export class BidiCore {
...
+            const wsConnectPromises: Promise<WebSocket | null>[] = []
+            const errorMessages: string[] = []
+            for (const candidateUrl of candidateUrls) {
+                const connectPromise = new Promise<WebSocket | null>((resolve) => {
+                    const ws = new environment.value.Socket(candidateUrl, this.#clientOptions) as unknown as WebSocket
```
christian-bromann, line 82: Let's setup a timeout to ensure we are not waiting forever in cases where non of the connection are established or fail for other reasons.

christian-bromann (commented):
Thanks for raising the PR. Added more comments.

packages/webdriver/src/bidi/core.ts
mykola-mokhnach: refactored

packages/webdriver/src/bidi/core.ts
mykola-mokhnach: I think this could also be done in a separate PR since no timeout was present before

packages/webdriver/src/bidi/core.ts
mykola-mokhnach: changed

packages/webdriver/src/bidi/core.ts
mykola-mokhnach: changed

packages/webdriver/src/bidi/core.ts
mykola-mokhnach: done

christian-bromann:
@mykola-mokhnach let me know if you need any help with the unit tests.

mykola-mokhnach:
@christian-bromann Could you please help with unit test updates?

christian-bromann (approved):
Thanks LGTM 👍 

I will push a test after merge.

christian-bromann:
@mykola-mokhnach mind reviewing https://github.com/webdriverio/webdriverio/pull/14272 ?

PR #14272, opened 2025-03-12
chore(webdriver): add unit tests for DNS patch

christian-bromann:
## Proposed changes

This adds unit tests for #14259. It also moves most of the added logic into a separate file as it is Node.js related and we want to keep basic implementations unrelated to the environment where it is executed in.

## Types of changes

## Checklist

## Backport Request

## Further comments

@mykola-mokhnach please review

### DIFF
--- a/packages/webdriver/src/node/bidi.ts
+++ b/packages/webdriver/src/node/bidi.ts
@@ -38,10 +38,15 @@ export async function listWebsocketCandidateUrls(webSocketUrl: string): Promise<
 }
 
 interface ConnectionResult {
-    index: number
     ws: WebSocket
     isConnected: boolean
     errorMessage?: string
+    index: number
+}
+
+interface ConnectionPromise {
+    index: number
+    promise: Promise<ConnectionResult>
 }
 
 /**
@@ -60,19 +65,20 @@ export async function connectWebsocket(candidateUrls: string[], _?: unknown): Pr
         }
     }).filter(Boolean) as WebSocket[]
 
-    const wsConnectPromises: Promise<ConnectionResult>[] = websockets.map((ws, index) => {
-        return new Promise<ConnectionResult>((resolve) => {
+    const wsConnectPromises: ConnectionPromise[] = websockets.map((ws, index) => {
+        const promise = new Promise<ConnectionResult>((resolve) => {
             ws.once('open', () => resolve({ ws, isConnected: true, index }))
             ws.once('error', (err) => {
                 log.debug(`Could not connect to Bidi protocol at ${candidateUrls[index]}: ${err.message}`)
                 resolve({ ws, isConnected: false, errorMessage: err.message, index })
             })
         })
+        return { promise, index }
     })
 
     const connectionTimeoutPromise = new Promise<undefined>((resolve) => {
         setTimeout(() => {
-            log.error(`Could not connect to Bidi protocol of any candidate url: "${candidateUrls.join('", "')}"`)
+            log.error(`Could not connect to Bidi protocol of any candidate url in time: "${candidateUrls.join('", "')}"`)
             return resolve(undefined)
         }, CONNECTION_TIMEOUT)
     })
@@ -101,14 +107,26 @@ export async function connectWebsocket(candidateUrls: string[], _?: unknown): Pr
  * @param promises - list of promises to race
  * @returns the first resolved promise
  */
-function firstResolved (promises: Promise<ConnectionResult>[]): Promise<ConnectionResult> {
+function firstResolved (promises: ConnectionPromise[], errorMessages: string[] = []): Promise<ConnectionResult | undefined> {
+    if (promises.length === 0) {
+        const sep = '\n  - '
+        const errorMessage = errorMessages.length > 0
+            ? sep + errorMessages.join(sep)
+            : ''
+        log.error('Could not connect to Bidi protocol' + errorMessage)
+        return Promise.resolve(undefined)
+    }
+
     // Race the wrapped promises
-    return Promise.race(promises).then(result => {
+    return Promise.race(promises.map(({ promise }) => promise)).then((result) => {
         if (result.isConnected) {
             return result
         }
 
         // If the first result was a rejection, race the remaining promises
-        return firstResolved(promises.filter((_, index) => index !== result.index))
+        return firstResolved(
+            promises.filter(({ index }) => index !== result.index),
+            [...errorMessages, result.errorMessage || 'unknown error']
+        )
     })
 }
--- a/packages/webdriver/src/node/bidi.ts
+++ b/packages/webdriver/src/node/bidi.ts
@@ -6,6 +6,7 @@ import logger from '@wdio/logger'
 import WebSocket from 'ws'
 
 const log = logger('webdriver')
+const CONNECTION_TIMEOUT = 10000
 
 export async function createBidiConnection(webSocketUrl: string, options?: unknown): Promise<WebSocket | undefined> {
     const candidateUrls = await listWebsocketCandidateUrls(webSocketUrl)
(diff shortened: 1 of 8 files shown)


packages/webdriver/src/node/bidi.ts
```diff
@@ -0,0 +1,89 @@
...
+        wsConnectPromises.push(connectPromise)
+    }
+    // We either wait until any web socket is successfully connected
+    // or all of them fail
+    await Promise.race([
+        firstWebsocketConnectedPromise,
+        Promise.all(wsConnectPromises)
+    ])
+
+    const ws = await firstWebsocketConnectedPromise
```
mykola-mokhnach, line 76: this might block forever if all websockets fail to connect. I assume you also wanted to invoke onFirstWebsocketConnected(undefined) in onError handler

packages/webdriver/src/node/bidi.ts
```diff
@@ -0,0 +1,89 @@
...
+    const wsConnectPromises: Promise<WebSocket | undefined>[] = []
+    const errorMessages: string[] = []
+    let onFirstWebsocketConnected = (_ws: WebSocket) => {}
+    const firstWebsocketConnectedPromise = new Promise<WebSocket>((resolve) => {
+        onFirstWebsocketConnected = resolve
+    })
+    const candidateWebsockets: WebSocket[] = []
+    for (const candidateUrl of candidateUrls) {
+        log.debug(`Attempt to connect to webSocketUrl ${candidateUrl}`)
+        const ws = new WebSocket(candidateUrl) as unknown as WebSocket
```
mykola-mokhnach, line 54: I assume the typecast is not really needed anymore

packages/webdriver/tests/node/bidi.test.ts
```diff
@@ -0,0 +1,69 @@
...
+    it('listWebsocketCandidateUrls', async () => {
+        const candidateUrls = await listWebsocketCandidateUrls('ws://foo/bar')
+        expect(candidateUrls).toEqual([
+            'ws://foo/bar',
+            'ws://127.0.0.1/bar',
+            'ws://::1/bar'
+        ])
+    })
+
+    it('createBidiConnection', async () => {
```
mykola-mokhnach, line 45: nice test 👍

packages/webdriver/src/node/bidi.ts
```diff
@@ -0,0 +1,109 @@
...
+    const websockets: WebSocket[] = candidateUrls.map((candidateUrl) => {
+        log.debug(`Attempt to connect to webSocketUrl ${candidateUrl}`)
+        return new WebSocket(candidateUrl)
+    })
+
+    const wsConnectPromises: Promise<ConnectionResult>[] = websockets.map((ws, index) => {
+        return new Promise<ConnectionResult>((resolve) => {
+            ws.once('open', () => resolve({ ws, isConnected: true, index }))
+            ws.once('error', (err) => {
+                log.debug(`Could not connect to Bidi protocol at ${candidateUrls[index]}: ${err.message}`)
```
mykola-mokhnach, line 62: might  there be a situation we see these logs even after the timeout expires?

packages/webdriver/src/node/bidi.ts
```diff
@@ -0,0 +1,109 @@
...
+ */
+function firstResolved (promises: Promise<ConnectionResult>[]): Promise<ConnectionResult> {
+    // Race the wrapped promises
+    return Promise.race(promises).then(result => {
+        if (result.isConnected) {
+            return result
+        }
+
+        // If the first result was a rejection, race the remaining promises
+        return firstResolved(promises.filter((_, index) => index !== result.index))
```
mykola-mokhnach, line 107: not sure if this logic would work as expected if there are more than two promises in the list. I would rather only consider filtering for promises that are in pending state. Check, for example, https://dev.to/devcrafter91/elegant-way-to-check-if-a-promise-is-pending-577g#:~:text=The%20trick,Promise%20%7B%20undefined%20%7D'%20when%20finished.

packages/webdriver/src/node/bidi.ts
christian-bromann: they shouldn't, we call `socket.removeAllListeners()` as part of the cleanup process

packages/webdriver/src/node/bidi.ts
christian-bromann: Since the size of the array never change we can rely on the fact that a certain index can point to a specific promise.

mykola-mokhnach: Lets check an example:

- We have 3 promises in the list
- The second promise has failed and returned undefined
- We now race between promise 1 and 3
- The third promise has failed and returned undefined
- We now race between promise 2 and 1
- Promise 2 is currently fulfilled, so it wins the race with undefined immediately
- We now race between promise 1 and 3
- we continue going to the recursive call infinite number of times if none of promises connects

Another example of undefined behaviour for Promise.race is when we only have one item in the initial list, which fails to connect and then we call the recursion with an empty list

christian-bromann: You are right, I had a logic mistake in there. I pushed a commit with a fix. Good catch 👌

packages/webdriver/src/node/bidi.ts
```diff
@@ -0,0 +1,132 @@
...
+
+/**
+ * Resolve hostnames to IPv4 and IPv6 addresses
+ * @returns list of websocket urls to try
+ * @see https://github.com/webdriverio/webdriverio/issues/14039
+ */
+export async function listWebsocketCandidateUrls(webSocketUrl: string): Promise<string[]> {
+    const parsedUrl = new URL(webSocketUrl)
+    const candidateUrls: string[] = [webSocketUrl]
+    if (!isIP(parsedUrl.hostname)) {
```
mykola-mokhnach, line 24: we may turn this condition into a guard to improve readability:

```
if (isIP(parsedUrl.hostname)) {
  return candidateUrls
}
...
```

packages/webdriver/src/node/bidi.ts
```diff
@@ -0,0 +1,132 @@
...
+ * @param promises - list of promises to race
+ * @returns the first resolved promise
+ */
+function firstResolved (promises: ConnectionPromise[], errorMessages: string[] = []): Promise<ConnectionResult | undefined> {
+    if (promises.length === 0) {
+        const sep = '\n  - '
+        const errorMessage = errorMessages.length > 0
+            ? sep + errorMessages.join(sep)
+            : ''
+        log.error('Could not connect to Bidi protocol' + errorMessage)
```
mykola-mokhnach, line 116: missing space after `protocol `

mykola-mokhnach (approved):

mykola-mokhnach:
It looks like some e2e need to be updated

mato533:
Hi @christian-bromann 

Sorry to jump in.
We did this because we considered it a high priority. as we were concerned that the CI failure would prevent other corrective actions from progressing as quickly as they should.


We have modified some of the code where this PR is concerned to ensure that the E2E test is successful.
Please review the PR #14290 . The PR's base branch is this PR(`cb/bidi-test`).

christian-bromann:
Thanks @mato533 for helping push this over the goal line 🙏

packages/webdriver/src/node/bidi.ts
```diff
@@ -0,0 +1,136 @@
...
+/**
+ * Connect to a websocket
+ * @param candidateUrls - list of websocket urls to try
+ * @returns true if the connection was successful
+ */
+export async function connectWebsocket(candidateUrls: string[], _?: unknown): Promise<WebSocket | undefined> {
+    const websockets: WebSocket[] = candidateUrls.map((candidateUrl) => {
+        log.debug(`Attempt to connect to webSocketUrl ${candidateUrl}`)
+        try {
+            const ws = new WebSocket(candidateUrl)
```
DudaGod, line 61: Can we send `options` here as second argument in order to be able to send headers when connect to browser using ws? These options are already passed to `connectWebsocket` method

mato533:
@DudaGod 
Thank you for comment. It seems to be possible.

@christian-bromann 
Have you any comment? If there is no concern, I will create the PR.

christian-bromann:
Sounds good to me, merged #14303

Issue #14302, opened 2025-03-19
[🐛 Bug]: Defining and using a CustomElement in Firefox breaks `$$`


htho:
### Have you read the Contributing Guidelines on issues?

- [x] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

9.12.0

### Node.js Version

v22.13.1

### Mode

WDIO Testrunner

### Which capabilities are you using?

```typescript
capabilities: [
    {
        browserName: 'chrome',
    },
    {
        browserName: 'firefox',
    },
],
```

### What happened?

I created a repository to reproduce the bug: https://github.com/htho/wdio-repro-collection

* I have a page with a custom element.
* When I use `$$` to count how often a selector matches, this works in Chrome, but not in Firefox.

**In Firefox tests fail, because the selector matched twice as many elements.**

I can share some findings:

1. `console.log(await $$("p").map(e=> e.elementId));` reveals that the elementIds are listed twice.
2. A Custom Element must be defined AND used
   1. A Custom Element that is not defined is ok
   2. A Custom Element that is defined, but never used is also ok

### What is your expected behavior?

The selector should list the correct amount of elements.

### How to reproduce the bug.

```html
<head>
    <script>
        class MyElement extends HTMLElement {}
        window.customElements.define("my-element", MyElement);
    </script>
</head>
<body>
    <div>
        <p>hi</p>
        <p>hi</p>
    </div>
    
    <my-element></my-element>
</body>
```

```ts
import { expect, browser, $$ } from '@wdio/globals'

describe('$$', () => {
    it(`counts`, async () => {
        await browser.url(`https://htho.github.io/wdio-repro-collection/multipleElements.html`)
        console.log(await $$("p").map(e=> e.elementId));
        await expect($$("p")).toBeElementsArrayOfSize(2);
    });
});
```

```cmd
git clone https://github.com/htho/wdio-repro-collection.git
cd wdio-repro-collection
npm ci

npm run test:multipleElements
```

### Relevant log output

```typescript
none
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [x] I have searched the existing issues

erwinheitzman:
Should be fixed in the next release

htho:
Thanks for the quick fix @admc, @erwinheitzman and @christian-bromann

PR #14319, opened 2025-03-25
filter out duplicated nodes

ccharnkij:
## Proposed changes

Fix for #14302. The issue is `browsingContextLocateNodes` for Firefox returns duplicated objects in the result array. I attempted to fix this by doing a simple filtering of any duplicates. I tested it out using the code in the issue and it went fine.

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/packages/webdriverio/src/utils/index.ts
+++ b/packages/webdriverio/src/utils/index.ts
@@ -328,10 +328,12 @@ export async function findDeepElement(
             ? [{ sharedId: (this as WebdriverIO.Element).elementId }]
             : undefined
     const deepElementResult = await browser.browsingContextLocateNodes({ locator, context, startNodes }).then(async (result) => {
-        const nodes: ExtendedElementReference[] = result.nodes.filter((node) => Boolean(node.sharedId)).map((node) => ({
+        let nodes: ExtendedElementReference[] = result.nodes.filter((node) => Boolean(node.sharedId)).map((node) => ({
             [ELEMENT_KEY]: node.sharedId as string,
             locator
         }))
+
+        nodes = returnUniqueNodes(nodes)
 
         if (!(this as WebdriverIO.Element).elementId) {
             return nodes[0]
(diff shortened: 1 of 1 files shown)


ccharnkij:
If that's the case, feel free to close this pr.

christian-bromann:
I don't mind maintaining a fix in WebdriverIO for this if it is well documented. Here is what I suggest:

1. raise an issue for Firefox document this issue
2. move this patch into a function and comment the function with the link to the issue
3. avoid string array duplicates by moving them into a set, e.g. `Array.from(new Set(['1', '2', '1']).values())`
4. verify we don't have other places where we should do this.

Does this make sense?

ccharnkij:
It's up to you guys. I can do what was suggested or let the Firefox guys take care of it.

christian-bromann:
> I can do what was suggested or let the Firefox guys take care of it.

I think it makes sense to have a temporary patch in WDIO but let's make sure we do our due diligence, raise it with the FF team and document this in the WDIO codebase.

packages/webdriverio/src/utils/index.ts
```diff
@@ -427,6 +432,13 @@ export async function findDeepElements(
     return deepElementResult as ElementReference[]
 }
 
+/**
+* Temporary patch for https://github.com/mozilla/geckodriver/issues/2223
+*/
+function returnUniqueNodes(nodes: ExtendedElementReference[]): ExtendedElementReference[] {
+    return Array.from(new Set(nodes.map(node => JSON.stringify(node)))).map(res => JSON.parse(res))
```
erwinheitzman, line 439: I believe we can filter using the element id's only, this way we don't have to do transform to a JSON string and back to an object and we also don't need to map

erwinheitzman:
> How's this?

I like the abstraction but I did add a comment to how the filtering itself is done as I think we can simplify it a bit.

I can take the task of raising an issue for Firefox as I already have an account there (assuming the Firefox and Geckodriver issues are reported in the same place), I might reach out for more details though if needed.

ccharnkij:
I'll change the code the filter with shareId. I've created the issue with Mozilla here https://github.com/mozilla/geckodriver/issues/2223

ccharnkij:
better?

packages/webdriverio/src/utils/index.ts
```diff
@@ -427,6 +432,14 @@ export async function findDeepElements(
     return deepElementResult as ElementReference[]
 }
 
+/**
+* Temporary patch for https://github.com/mozilla/geckodriver/issues/2223
+*/
+function returnUniqueNodes(nodes: ExtendedElementReference[]): ExtendedElementReference[] {
+    const ids = new Set()
+    return nodes.filter((node) => !ids.has(node[ELEMENT_KEY]) && ids.add(node[ELEMENT_KEY]))
```
erwinheitzman, line 440: I actually just saw a small thing at the end, we now filter but don't do anything with it, instead of assigning to the Set, you can filter and what is returned can be put into the Set instead 👍

```suggestion
    return new Set(nodes.filter((node) => !ids.has(node[ELEMENT_KEY]) && ids.add(node[ELEMENT_KEY])))
```

erwinheitzman (changes requested):
Saw one last thing right when I approved, if you update that it would be good to go in my eyes 👌

EDIT: I was mislead, apologies for the confusion

erwinheitzman (approved):

PR #14350, opened 2025-03-29
feat: introduce defineConfig function to create a typed configuration object

erwinheitzman:
## Proposed changes

Introduce defineConfig function (I took the name from Playwrightso it's recognisable but we can opt for any name here) which allows for a typed config setup for both Javascript and Typescript along with showing JsDoc notations/comments.

This also means that there's no more need to assign a type manually.

Example:
```js
import { defineConfig } from '@wdio/config'

export const config = defineConfig({
  capabilities: [{ browserName: 'chrome' }],
  // ...
})
```

IMPORTANT NOTE: I also removed some Babel related code, this might be breaking to our users. I can revert this if we think this is too risky!

Javascript example:
Scherm­afbeelding 2025-03-30 om 14 20 55

TypeScript example (the type can be removed but it's there for showcase purposes):
Scherm­afbeelding 2025-03-30 om 14 20 03

## Types of changes

## Checklist

### DIFF
--- a/__mocks__/@wdio/config.ts
+++ b/__mocks__/@wdio/config.ts
@@ -1,13 +1,10 @@
 import { vi } from 'vitest'
-import type { Options } from '../../packages/wdio-types'
-
-import { DEFAULT_CONFIGS as DEFAULT_CONFIGS_IMPORT } from '../../packages/wdio-config/src/constants.js'
 import {
     isCloudCapability as isCloudCapabilityMock,
     validateConfig as validateConfigMock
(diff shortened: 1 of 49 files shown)


christian-bromann (commented):
I am in favor of this proposal/change. It seems much more intuitive to use a function like `defineConfig` rather than having users know about the `WebdriverIO` namespace.

Do you want to update the way we document configurations in `/website` in a different PR or as part of it?

erwinheitzman:
I got a PR for that here #14352

I'll update that PR with the docs 👍

PR #14361, opened 2025-04-01
fix(webdriverio): fix order of execution in session polyfill

erwinheitzman:
## Proposed changes

Fix possible race condition by fixing the order of execution.

## Types of changes

## Checklist

### DIFF
--- a/packages/webdriverio/src/session/polyfill.ts
+++ b/packages/webdriverio/src/session/polyfill.ts
@@ -48,6 +48,9 @@ export class PolyfillManager extends SessionManager {
             return
         }
 
+        // start listening for browsingContext.contextCreated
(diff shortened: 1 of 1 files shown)


christian-bromann (changes requested):
What bug are you trying to fix with this? Technically this doesn't change anything. Note that you can start listen on Bidi events before you subscribe to them.

erwinheitzman:
Normally you start listening and then trigger it (that's what the subscribe does as far as I can tell). So that's what I believe I changed it to, first we listen to it and then we trigger/subscribe to it

christian-bromann:
That would be usually correct but here we can register the listener first before we subscribe to the Bidi event. This works because the Bidi handler in WebdriverIO emits events separate from browser eventlistener. I think we can go ahead and merge as this may confuse others as well.

christian-bromann (approved):

PR #14370, opened 2025-04-04
feat(webdriverio): introduce page object

christian-bromann:
## Proposed changes

An important concept within WebDriver Bidi is the idea of individual context ids that allows to call a certain action on different browser windows or tabs. In WebDriver Classic a command was only executable on the current active tab/window. WebDriver Bidi supports more freedom and makes handling different tabs much easier.

This patch introduces a new type called `WebdriverIO.Page` which is similar to `WebdriverIO.Browser` with the following differences:

- it will be only available if the session is a Bidi session
- it represents a browsing context rather than the whole browser
- it only contains commands that make sense to run per context, e.g. `execute` or `action` but not `addCommand`
- there is only 1 `WebdriverIO.Browser` per session and there can be endless `WebdriverIO.Page` per session

Here is an example how things currently look like:

```ts
await browser.url('https://google.com')
console.log(await browser.getTitle()) // outputs: "Google"

await browser.newWindow('https://webdriver.io', {
    type:'tab',
    windowName: 'WebdriverIO window',
    windowFeature: 'width=420,height=230,resizable,scrollbars=yes,status=1',
})
console.log(await browser.getTitle()) // outputs: "WebdriverIO · Next-gen browser and mobile automation test framework for Node.js"
console.log(result.type) // outputs: "tab"
await browser.closeWindow()
console.log(await browser.getTitle()) // outputs: "Google"
```

is now much more clear:

```ts
const page = await browser.url('https://google.com')
console.log(await page.getTitle()) // outputs: "Google"

const newWindow = await page.newWindow('https://webdriver.io', {
    type: 'window'
})

console.log(await page.getTitle()) // outputs: "Google"
console.log(page.type) // outputs: "tab"
console.log(await newWindow.getTitle()) // outputs: "WebdriverIO · Next-gen browser and mobile automation test framework for Node.js"
console.log(newWindow.type) // outputs: "window"
```

## Types of changes

It seems like this patch will very likely be a breaking change for all users that were running tests in WebDriver Bidi and potentially using the `WebdriverIO.Request` object when calling `browser.url("..."). I'ld suggest to publish this as v10.

## Checklist

## Backport Request

- [x] This change is solely for `v10` and doesn't need to be back-ported

## Further comments

n/a


packages/webdriverio/src/commands/page/url.ts
```diff
@@ -0,0 +1,153 @@
...
+ * @type protocol
+ *
+ */
+export async function url (
+    this: WebdriverIO.Page,
+    path: string,
+    options: UrlCommandOptions = {}
+): Promise<WebdriverIO.Page> {
+    const request =  await navigateContext.call(this.browser, path, this.contextId, options)
+    this.request = request
```
erwinheitzman, line 151:
```suggestion
    this.request = await navigateContext.call(this.browser, path, this.contextId, options)
```

packages/webdriverio/src/commands/page/setViewport.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * value check
+     */
+    if (options.width < minWindowSize || options.width > maxWindowSize || options.height < minWindowSize || options.height > maxWindowSize) {
+        throw new Error('setViewport expects width and height to be a number in the 0 to 2^31 − 1 range')
+    }
+
+    if (options.devicePixelRatio && (typeof options.devicePixelRatio !== 'number' || options.devicePixelRatio < 0)) {
+        throw new Error('setViewport expects devicePixelRatio to be a number in the 0 to 2^31 − 1 range')
```
erwinheitzman, line 50:
```suggestion
        throw new Error(`setViewport expects devicePixelRatio to be a number in the range of 0 to ${Number.MAX_SAFE_INTEGER}`)
```

packages/webdriverio/src/commands/page/setViewport.ts
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    /**
+     * value check
+     */
+    if (options.width < minWindowSize || options.width > maxWindowSize || options.height < minWindowSize || options.height > maxWindowSize) {
+        throw new Error('setViewport expects width and height to be a number in the 0 to 2^31 − 1 range')
+    }
+
+    if (options.devicePixelRatio && (typeof options.devicePixelRatio !== 'number' || options.devicePixelRatio < 0)) {
```
erwinheitzman, line 49: let's move the type checking to the top where we already do the type checking for the window size

packages/webdriverio/src/commands/page/setViewport.ts
```diff
@@ -0,0 +1,61 @@
...
+ * @type window
+ */
+export async function setViewport(
+    this: WebdriverIO.Page,
+    options: SetViewportOptions
+) {
+    /**
+     * type check
+     */
+    if (typeof options.width !== 'number' || typeof options.height !== 'number') {
```
erwinheitzman, line 38: let's add a Number.isInteger check as well

packages/webdriverio/src/commands/page/keys.ts
```diff
@@ -0,0 +1,62 @@
...
+        const charArray = value as (keyof typeof UNICODE_CHARACTERS)[]
+        for (const charSet of charArray) {
+            keySequence = keySequence.concat(checkUnicode(charSet))
+        }
+    } else {
+        throw new Error('"keys" command requires a string or array of strings as parameter')
+    }
+
+    /**
+     * W3C way of handle it key actions
```
erwinheitzman, line 47: it keys?

packages/webdriverio/src/commands/page/keys.ts
```diff
@@ -0,0 +1,62 @@
...
+     */
+    const keyAction = this.action('key')
+    keySequence.forEach((value) => keyAction.down(value))
```
erwinheitzman, line 50: Shouldn't it be like this? Seems like there's only a pause once now
```suggestion
    keySequence.forEach((value) => {
      keyAction.down(value));
      keyAction.pause(10));
      keyAction.down(value));
    })
```

erwinheitzman (commented):
This is some amazing work!

packages/webdriverio/src/commands/page/keys.ts
christian-bromann: possibly, let's raise a separate issue for it to look into this.

damencho:
I see this is about tabs and windows, but what about iframes? Is it possible to incorporate and iframes, so you don't need to switch constantly between main page and some iframe, same way to avoid changing tabs/windows to execute stuff.
WDYT?

christian-bromann:
@damencho makes total sense and is something I consider to do. Thanks for the feedback.

damencho:
No more plans for adding this?

Issue #14326, opened 2025-03-26
[🐛 Bug]: Opening a new tab crashes webdriver


alechirsch:
### Have you read the Contributing Guidelines on issues?

- [x] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

latest

### Node.js Version

latest

### Mode

Standalone Mode

### Which capabilities are you using?

```typescript
{
        browserName: 'chrome',
        'goog:chromeOptions': {
            args: process.env.CI ? ['headless', 'disable-gpu'] : []
        }
    }
```

### What happened?

When clicking on a button on a page that opens a new tab, webdriver crashes with the following error message:

### What is your expected behavior?

Webdriver should not crash.

### How to reproduce the bug.

```js
import { remote } from 'webdriverio'

const browser = await remote({
    capabilities: {
        browserName: 'chrome',
        'goog:chromeOptions': {
            args: process.env.CI ? ['headless', 'disable-gpu'] : []
        }
    }
})

await browser.url('https://pestpatrol1.com/pay-now/');
let allTabs = await browser.getWindowHandles();
let currentTab = allTabs[0]
await browser.$('#custrefid').setValue('123');
await browser.$('#Submit').click();

// NOTE: Everything below can be commented out and it still crashes, this is for the positive case
// When the the number of tabs change, try to change tabs
let newTabs = await browser.getWindowHandles();

if (newTabs.length !== allTabs.length) {
    let newTab = newTabs.find(tab => tab !== currentTab);
    await browser.switchToWindow(newTab);
}

await new Promise(resolve => setTimeout(resolve, 5000));

await browser.$('#btncancel1').click();
```

Note that the following does work, this seems to only be an issue for standalone:
```js
import { expect, browser } from '@wdio/globals'

describe('My Login application', () => {
    it('should login with valid credentials', async () => {
        await browser.url('https://pestpatrol1.com/pay-now/');
        let allTabs = await browser.getWindowHandles();
        let currentTab = allTabs[0]
        await browser.$('#custrefid').setValue('123');
        await browser.$('#Submit').click();
        

        // When the the number of tabs change, try to change tabs
        let newTabs = await browser.getWindowHandles();
        console.log(allTabs);
        
        if (newTabs.length !== allTabs.length) {
            let newTab = newTabs.find(tab => tab !== currentTab);
[...]

PR #14398, opened 2025-04-10
catching error from addPreloadScript

ccharnkij:
## Proposed changes

Fix for #14326. Seems like the issue is from polyfill -> registerScripts -> scriptAddPreloadScript -> send error not being caught.

## Types of changes

## Checklist

## Further comments

I tried this out with the code from the issue and both standalone and framework went fine. The only part that I was not sure about was the `send` errors were thrown for both standalone and framework, but it only caused standalone to error out from promise uncaught error. The framework (I tried mocha) had `send` error but it kept on going. Maybe the context in which the error was in between the 2 modes were different?

### DIFF
--- a/packages/webdriverio/src/session/polyfill.ts
+++ b/packages/webdriverio/src/session/polyfill.ts
@@ -81,6 +81,15 @@ export class PolyfillManager extends SessionManager {
                 ? this.#browser.scriptAddPreloadScript({
                     functionDeclaration,
                     contexts: [context.context]
+                }).catch(() => {
+                    /**
+                     * In case the context is already destroyed before this promise is finished
+                     * For example:
+                     *   - an unsuspecting window (context) is opened
+                     *   - registerScripts is triggered
+                     *   - that window closes before the bidi call starts
+                     *   - bidi call sends the request and gets something like `call rejected because the connection has been closed` error back
+                     */
                 })
                 : Promise.resolve(),
             this.#browser.scriptCallFunction({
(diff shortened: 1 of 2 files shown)


christian-bromann:
@ccharnkij can we identify why `registerScript` is called in the first place? Your proposal just suppresses the error and I wonder if the actual issue is that we call `registerScript` for a context that it shouldn't be called on or we don't clear up a context id properly.

ccharnkij:
Seems like what happens in this case is a new context (or window) is open AND close before you can blink, but it is enough to trigger `registerScript`. Following the steps from the issue, after a user clicks submit:
 - new tab opens
 - an unsuspecting window pops up
 - triggers `registerScript`
 - that window closes
 - `registerScript` still runs causing `send` to throw error

I notice that that window only pops up in Chromium and not Firefox.

ccharnkij:
I can, but not sure what comment you were referring to though. The code comment? (the // part?)

packages/webdriverio/src/session/polyfill.ts
```diff
@@ -81,6 +81,10 @@ export class PolyfillManager extends SessionManager {
                 ? this.#browser.scriptAddPreloadScript({
                     functionDeclaration,
                     contexts: [context.context]
+                }).catch(() => {
+                    /**
+                     * this may fail if the context is already destroyed
```
christian-bromann, line 86: 👆 this one

erwinheitzman:
I've seen a new tab being added in the handles lately when opening a pdf in Chrome and I wonder if this is related as the tab is not actually visible but it is returned in the list. Not saying that this is causing it but I do find it peculiar

ccharnkij:
@christian-bromann like this?

ccharnkij:
@christian-bromann Have you got a chance to take a look?

christian-bromann (approved):
LGTM 👍

PR #14426, opened 2025-04-23
fix(elemSearch): support for other languages for case-insensitive ele…

sipayRT:
…m search

## Proposed changes

Hi. I'm trying to use case-insensitive element search for Russian text, but it doesn't work. For example,

```html
<span>Моя Страница</span>
```

```
$("span.=моя страница");
```

In this PR I added support for the main languages.

## Types of changes

## Checklist

## Backport Request

## Further comments


packages/webdriverio/src/utils/findStrategy.ts
```diff
@@ -18,7 +18,7 @@ const XPATH_SELECTOR_REGEXP = [
     // optional . or # + class or id
     /(?:(\.|#)(-?[_a-zA-Z]+[_a-zA-Z0-9-]*))?/,
     // optional [attribute-name="attribute-selector"]
-    /(?:\[(-?[_a-zA-Z]+[_a-zA-Z0-9-]*)(?:=(?:"|')([a-zA-z0-9\-_. ]+)(?:"|'))?\])?/,
+    /(?:\[(-?[_a-zA-Z]+[_a-zA-Z0-9-]*)(?:=(?:"|')([a-zA-ZÀÁÂÃÄÅÆÇÈÉÊËÌÍÎÏÐÑÒÓÔÕÖØÙÚÛÜÝÞŸĞİŞа-яА-ЯЁё0-9\-_. ]+)(?:"|'))?\])?/,
```
christian-bromann, line 21: Can we use `\p{Lu}` and `\p{Ll}` instead?

packages/webdriverio/src/utils/findStrategy.ts
```diff
@@ -28,6 +28,9 @@ const IMAGEPATH_MOBILE_SELECTORS_ENDSWITH = [
     '.jpg', '.jpeg', '.gif', '.png', '.bmp', '.svg'
 ]
 
+export const UPPER_CASE_SYMBOLS = 'ABCDEFGHIJKLMNOPQRSTUVWXYZÀÁÂÃÄÅÆÇÈÉÊËÌÍÎÏÐÑÒÓÔÕÖØÙÚÛÜÝÞŸĞİŞАБВГДЕЁЖЗИЙКЛМНОПРСТУФХЦЧШЩЪЫЬЭЮЯ'
+export const LOWER_CASE_SYMBOLS = 'abcdefghijklmnopqrstuvwxyzàáâãäåæçèéêëìíîïðñòóôõöøùúûüýþÿğışабвгдеёжзийклмнопрстуфхцчшщъыьэюя'
```
christian-bromann, line 32:
```suggestion
export const UPPER_CASE_SYMBOLS = '\p{Lu}'
export const LOWER_CASE_SYMBOLS = '\p{Ll}'
```

christian-bromann (changes requested):

packages/webdriverio/src/utils/findStrategy.ts
sipayRT: sorry, but the `translate` function doesn't support such way - https://developer.mozilla.org/en-US/docs/Web/XML/XPath/Reference/Functions/translate (I checked it out). In the browser context we can use only xPath@1x

christian-bromann: Thanks for checking. One last request: can we add a comment above these two variables, documenting why we are including all these characters and that we have evaluated my suggestion and can't use it for this reason? Making sure that people understand the context and don't have to dig through GitHub PRs.

sipayRT: yeah, sure

packages/webdriverio/src/utils/findStrategy.ts
```diff
@@ -17,8 +17,8 @@ const XPATH_SELECTOR_REGEXP = [
     /^([a-z0-9|-]*)/,
     // optional . or # + class or id
     /(?:(\.|#)(-?[_a-zA-Z]+[_a-zA-Z0-9-]*))?/,
-    // optional [attribute-name="attribute-selector"]
-    /(?:\[(-?[_a-zA-Z]+[_a-zA-Z0-9-]*)(?:=(?:"|')([a-zA-z0-9\-_. ]+)(?:"|'))?\])?/,
+    // optional [attribute-name="attribute-selector"]. "\p{L}" - all letters (case-insensitive), "\p{N}" - all numbers
+    /(?:\[(-?[_a-zA-Z]+[_a-zA-Z0-9-]*)(?:=(?:"|')([\p{L}\p{N}\-_. ]+)(?:"|'))?\])?/u,
```
erwinheitzman, line 21: I know this is not directly related to your change but it might be nice to improve this while we are at it.

If the idea of `[_a-zA-Z]+` is to match `[a-zA-Z0-9_]+` this can be simplified using `\w+`.
A similar thing goes for `[_a-zA-Z0-9-]*` with `[\w|\d]*`.

I'm curious if this will work but it's not a blocker for me and we can merge without this change 👍

erwinheitzman, line 21: When using https://regex101.com/ (make sure to select ecmascript) it tells me: `{L} matches a single character in the list {L} (case sensitive)`

erwinheitzman:
Nice improvement 👍

christian-bromann:
@sipayRT anything we can do to help ship this?

Delta456:
We would also need to support CJK character set I believe

christian-bromann:
@sipayRT any chance we can move forward with this? Otherwise we would have to close the PR? Mayeb @Delta456 you are interested re-raising this one and addressing the comments?

christian-bromann:
Closing due to inactivity.

PR #14448, opened 2025-05-02
fix(bidi-dialog): only accept dialogs in active browsing context

### LINKED ISSUE #14295: [🐛 Bug]: Errors interacting with native browser dialogs (Bidi protocol)

### WebdriverIO Version

9.12.0

### Node.js Version

20.17.0

### Mode

WDIO Testrunner

```typescript
{
    browserName: "chrome",
    "goog:chromeOptions": {
            args: [
            "--no-sandbox",
            "--start-maximized",
            "--window-size=1920,1080",
            "--disable-dev-shm-usage",
            //"--headless=new",
            "--lang=en-GB"
        ]
    },
    // "wdio:enforceWebDriverClassic": true,
}
```

### What happened?

I am testing an application that uses native browser dialogs to confirm some user actions (like deletes, ...). With Bidi protocol, I face 2 issues that do not happen with WebdriverClassic.

1. Sometimes, the dialog does not open or opens and closes automatically and almost immediately. This causes the error: "no such alert: WebDriverError: no such alert"
2. When the first error does not happen, and there is an interaction with the dialog, then there is the  following error on the next action after the dialog interaction: "no such frame - Context 9467FF542C3FAB062A8D6E7FF8570DD9 not found". When this error happens, the test breaks and the browser remains open until I manually close it.

To have a similar behavior as we have running the tests with WebdriverClassic.

### How to reproduce the bug.

For the first error, here is a reproducible example:
[browser_native_dialogs.zip](https://github.com/user-attachments/files/19289893/browser_native_dialogs.zip)

And the logs of the error happening: 
[log_1.txt](https://github.com/user-attachments/files/19290056/log_1.txt)

For the second example, I cannot provide a reproducible example (maybe after you first the first one, you will face this one), just the logs:

[log_2.txt](https://github.com/user-attachments/files/19290432/log_2.txt)

### Relevant log output

```typescript
Provided above.
```

### Code of Conduct

Rondleysg:
## Proposed changes

This PR adds a context check in the `accept() `handler so that we only call `dialog.accept() `when the dialog actually belongs to the current browsing context. It uses the `ContextManager.getCurrentContext()` helper and bails out early if `this.#context !== context`, preventing “no such frame” errors when multiple contexts fire native‐dialog events.

## Related issue
Closes #14295

## Types of changes

## Checklist

### DIFF
--- a/packages/webdriverio/src/session/dialog.ts
+++ b/packages/webdriverio/src/session/dialog.ts
@@ -115,7 +115,9 @@ export class Dialog {
         const contextManager = getContextManager(this.#browser)
         const context = await contextManager.getCurrentContext()
 
-        if (this.#context !== context) {return}
+        if (this.#context !== context) {
+            return
+        }
 
         await this.#browser.browsingContextHandleUserPrompt({
             accept: true,
--- a/packages/webdriverio/src/session/dialog.ts
+++ b/packages/webdriverio/src/session/dialog.ts
@@ -1,5 +1,6 @@
 import { type local } from 'webdriver'
 import { SessionManager } from './session.js'
+import { getContextManager } from './context.js'
 
 export function getDialogManager(browser: WebdriverIO.Browser) {
     return SessionManager.getSessionManager(browser, DialogManager)
@@ -111,6 +112,11 @@ export class Dialog {
      * @returns {Promise<void>}
      */
     async accept(userText?: string) {
+        const contextManager = getContextManager(this.#browser)
+        const context = await contextManager.getCurrentContext()
+
+        if (this.#context !== context) {return}
+
         await this.#browser.browsingContextHandleUserPrompt({
             accept: true,
             context: this.#context,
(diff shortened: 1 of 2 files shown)


packages/webdriverio/src/session/dialog.ts
```diff
@@ -111,6 +112,11 @@ export class Dialog {
      * @returns {Promise<void>}
      */
     async accept(userText?: string) {
+        const contextManager = getContextManager(this.#browser)
+        const context = await contextManager.getCurrentContext()
+
+        if (this.#context !== context) {return}
```
christian-bromann, line 118:
```suggestion
        if (this.#context !== context) {
            return
        }
```

christian-bromann (changes requested):

erwinheitzman:
Quick question, shouldn't we add this check to all public methods, like close() etc as well?

erwinheitzman:
Great change, thanks for picking this up 👍

christian-bromann (approved):
LGTM 👍

Rondleysg:
> Quick question, shouldn't we add this check to all public methods, like close() etc as well?

Yes, sorry, I should have added dismiss() too.
I'll do it in a new PR

Issue #14492, opened 2025-05-18
[🐛 Bug]: Webdriver creates a new agent object for each individual request which is expensive


dragosMC91:
### Have you read the Contributing Guidelines on issues?

- [x] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

9.14.0

### Node.js Version

20.10.0

### Mode

WDIO Testrunner

### Which capabilities are you using?

```typescript
{
            browserName: 'chrome',
            'wdio:enforceWebDriverClassic': true,
            'goog:platformName': 'desktop',
            'wdio:exclude': [
                './test/specs/**/*mobile*.js'
            ],
            'goog:loggingPrefs': { browser: 'WARNING' },
            'goog:chromeOptions': {
                args: [
                    '--no-sandbox',
                    '--disable-dev-shm-usage',
                    '--window-position=1050,210',
                    '--window-size=1366,768',
                    ...headlessCapability
                ],
            },
        }
```

### What happened?

It's not a bug per se, just an unnecessary operation from what i can tell.

### What is your expected behavior?

The agent should be reused for the entire duration of the test, i am not sure i see the benefit for creating a new agent for each individual request, even if this operation costs ~0.3-0.5 ms, but this small extra time is 10% for lots of requests

### How to reproduce the bug.

In order to get an idea of how much time this operation adds to each request, i wrote a performance test for 3 very common operations: `fetch element`, `click element` and `is element displayed`.

Clone the issue highlight repo below, and run the same performance test for the original wdio, and patched wdio:
```
ENABLE_HEADLESS=1 DEBUG= yarn test --spec test/specs/suite1/validatePerformance.desktop.test.js
```
`Original` wdio 9.14.0 branch: https://github.com/dragosMC91/wdio-exclude-pattern-bug-highlight/pull/11
`Patched` wdio 9.14.0 branch: https://github.com/dragosMC91/wdio-exclude-pattern-bug-highlight/pull/10 (this just adds a cached dispatcher to the [FetchRequest class](https://github.com/webdriverio/webdriverio/blob/v9.14.0/packages/webdriver/src/request/node.ts#L28))
`Baseline` wdio 8.44.1 branch: https://github.com/dragosMC91/wdio-exclude-pattern-bug-highlight/pull/8

after checking out each branch, don't forget to `yarn install` to make sure you have the applied patch

[...]

PR #14493, opened 2025-05-18
Cache the wd request dispatcher

dragosMC91:
## Proposed changes
I noticed that a new agent is created for each individual wd request. Maybe i am missing something but this seems excessive.

I propose using a cached instance of the dispatcher.

I validated the performance improvement for this change via the performance tests referenced in the issue for this fix - https://github.com/webdriverio/webdriverio/issues/14492

## Types of changes

## Checklist

## Backport Request

### DIFF
--- a/packages/webdriver/tests/request.test.ts
+++ b/packages/webdriver/tests/request.test.ts
@@ -7,8 +7,7 @@ import type { Options } from '@wdio/types'
 
 import '../src/browser.js'
 import { FetchRequest as WebFetchRequest } from '../src/request/web.js'
-import { FetchRequest } from '../src/request/node.js'
-import { SESSION_DISPATCHERS } from '../src/request/node.js'
+import { FetchRequest, SESSION_DISPATCHERS } from '../src/request/node.js'
 
 vi.mock('@wdio/logger', () => import(path.join(process.cwd(), '__mocks__', '@wdio/logger')))
 vi.mock('fetch')
--- a/packages/webdriver/tests/request.test.ts
+++ b/packages/webdriver/tests/request.test.ts
@@ -6,10 +6,20 @@ import logger from '@wdio/logger'
 import type { Options } from '@wdio/types'
 
 import '../src/browser.js'
-import { FetchRequest } from '../src/request/web.js'
+import { FetchRequest as WebFetchRequest } from '../src/request/web.js'
+import { FetchRequest } from '../src/request/node.js'
+import { SESSION_DISPATCHERS } from '../src/request/node.js'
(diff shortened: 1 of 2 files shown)


christian-bromann (changes requested):
I totally agree that we can be more efficient here. However the current implementation would not allow to have multiple WebDriver sessions with different options as we would cache one dispatcher per process. I wonder if we can store them in a map based on a given session id. Wdyt?

dragosMC91:
> I totally agree that we can be more efficient here. However the current implementation would not allow to have multiple WebDriver sessions with different options as we would cache one dispatcher per process. I wonder if we can store them in a map based on a given session id. Wdyt?

i initially wanted to use a map, but thought each process would only have 1 set of options, defined at test start. So i missed this scenario you mentioned.
I will investigate how to implement your suggestion, will probably do so after work tomorrow

christian-bromann (commented):
We should clean up session entries at some point otherwise the map may grow infinitely. One way could be to check if the request is a delete session request and clean up after.

dragosMC91:
> We should clean up session entries at some point otherwise the map may grow infinitely. One way could be to check if the request is a delete session request and clean up after.

good idea, should i also add unit tests for this change ?

christian-bromann:
> should i also add unit tests for this change ?

That would be great, thanks!

dragosMC91:
> > should i also add unit tests for this change ?
> 
> That would be great, thanks!

will do so tomorrow after work

dragosMC91:
@christian-bromann when there are no other change requests, i'd like to rerun the performance tests on the final version with the tests here https://github.com/dragosMC91/wdio-exclude-pattern-bug-highlight/blob/master/test/specs/suite1/validatePerformance.desktop.test.js
i'd appreciate it if we only merged after

packages/webdriver/tests/request.test.ts
```diff
@@ -6,10 +6,20 @@ import logger from '@wdio/logger'
 import type { Options } from '@wdio/types'
 
 import '../src/browser.js'
-import { FetchRequest } from '../src/request/web.js'
+import { FetchRequest as WebFetchRequest } from '../src/request/web.js'
+import { FetchRequest } from '../src/request/node.js'
+import { SESSION_DISPATCHERS } from '../src/request/node.js'
```
christian-bromann, line 11:
```suggestion
import { FetchRequest, SESSION_DISPATCHERS } from '../src/request/node.js'
```

christian-bromann (approved):
LGTM 👍

Issue #14548, opened 2025-06-08
[🐛 Bug]: Overwrite commands original function type problem


dprevost-LMI:
### Have you read the Contributing Guidelines on issues?

- [x] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

9.15.0

### Node.js Version

18

### Mode

Standalone Mode

### Which capabilities are you using?

```typescript
N/A
```

### What happened?

The below code shows the `originalFunction` as `(parameter) originalFunction: (...args: any[]) => (this: WebdriverIO.Element) => Promise<string>` while it should `(parameter) originalFunction: () => Promise<string>`

```
browser.overwriteCommand(
  'getText',
  async function (this: WebdriverIO.Element, originalFunction) {
    // Error with `Type '(this: Element) => Promise<string>' is not assignable to type 'string'.ts(2322)`, while typeof text is string
    const text: string = await originalFunction();
    return text
  },
  true,
);
```

This is caused by the two types below, where `origCommand` is using `(...args: any[]) =>`, and because `$ElementCommands[ElementKey]` is missing `OmitThisParameter`

```
type OverwriteCommandFnScoped<ElementKey extends keyof $ElementCommands, BrowserKey extends keyof $BrowserCommands, IsElement extends boolean = false> = (this: IsElement extends true ? WebdriverIO.Element : WebdriverIO.Browser, origCommand: (...args: any[]) => IsElement extends true ? $ElementCommands[ElementKey] : $BrowserCommands[BrowserKey], ...args: any[]) => Promise<any>;
type OverwriteCommandFn<ElementKey extends keyof $ElementCommands, BrowserKey extends keyof $BrowserCommands, IsElement extends boolean = false> = (origCommand: (...args: any[]) => IsElement extends true ? $ElementCommands[ElementKey] : $BrowserCommands[BrowserKey], ...args: any[]) => Promise<any>;
```

When using the below types, the `originalFunction` is now correctly inferred

```
[...]

PR #14549, opened 2025-06-08
fix(types): Fix overwrite command bad typing

dprevost-LMI:
## Proposed changes

As exposed in [this issue](https://github.com/webdriverio/webdriverio/issues/14548), when using `overwriteCommand`, the typing is not always correctly inferred.  We attempt to fix a couple of them  and more:

Thing done:
- Have the `this` refer to `WebdriverIO.Element` automatically in the `func` of `overwriteCommand`
- Have the `originalCommandFunction` infer the correct return type by adding the return type explicitly in existing browser commands
- Review documentation to be more verbose and also be more type-safe
- Propose `overwriteBrowserCommand` and `overwriteElementCommand`, instead of using the boolean, which makes the API less clear and typing definition even harder

## Types of changes

## Checklist

## Backport Request

## Further comments

Please let me know about the new API proposal and whether you would like to deprecate the former or keep it.

### DIFF
--- a/packages/webdriverio/src/types.ts
+++ b/packages/webdriverio/src/types.ts
@@ -201,7 +201,7 @@ type OverwriteCommandFnScoped<
     IsElement extends boolean = false
 > = (
     this: IsElement extends true ? WebdriverIO.Element : WebdriverIO.Browser,
-    origCommand: (...args: any[]) => IsElement extends true ? $ElementCommands[ElementKey] : $BrowserCommands[BrowserKey],
+    origCommand: IsElement extends true ? OmitThisParameter<$ElementCommands[ElementKey]> : OmitThisParameter<$BrowserCommands[BrowserKey]>,
     ...args: any[]
 ) => Promise<any>
 
@@ -210,7 +210,7 @@ type OverwriteCommandFn<
     BrowserKey extends keyof $BrowserCommands,
     IsElement extends boolean = false
 > = (
-    origCommand: (...args: any[]) => IsElement extends true ? $ElementCommands[ElementKey] : $BrowserCommands[BrowserKey],
+    origCommand: IsElement extends true ? OmitThisParameter<$ElementCommands[ElementKey]> : OmitThisParameter<$BrowserCommands[BrowserKey]>,
     ...args: any[]
 ) => Promise<any>
 
--- a/packages/webdriverio/src/types.ts
+++ b/packages/webdriverio/src/types.ts
@@ -189,7 +189,7 @@ type AddCommandFnScoped<
     InstanceType = WebdriverIO.Browser,
     IsElement extends boolean = false
 > = (
-    this: IsElement extends true ? Element : InstanceType,
+    this: IsElement extends true ? WebdriverIO.Element : InstanceType,
     ...args: any[]
 ) => any
 
@@ -221,7 +221,7 @@ export interface CustomInstanceCommands<T> {
      */
     addCommand<IsElement extends boolean = false>(
         name: string,
-        func: AddCommandFn | AddCommandFnScoped<T, IsElement>,
+        func: IsElement extends true ? AddCommandFnScoped<T, IsElement> : AddCommandFn,
         attachToElement?: IsElement,
         proto?: Record<string, any>,
         instances?: Record<string, WebdriverIO.Browser | WebdriverIO.MultiRemoteBrowser>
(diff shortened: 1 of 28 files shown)


packages/webdriverio/src/types.ts
```diff
@@ -210,32 +210,79 @@ type OverwriteCommandFn<
...
+        func: IsElement extends true ? AddCommandFnScoped<Instance, IsElement> : AddCommandFn,
         attachToElement?: IsElement,
         proto?: Record<string, any>,
-        instances?: Record<string, WebdriverIO.Browser | WebdriverIO.MultiRemoteBrowser>
+        instances?: Record<string, Instances>
+    ): void;
+
+    /**
+     * add command to `browser`
+     * Waiting approval
```
dprevost-LMI: TODO and other

packages/webdriverio/src/types.ts
```diff
@@ -210,32 +210,80 @@ type OverwriteCommandFn<
     BrowserKey extends keyof $BrowserCommands,
     IsElement extends boolean = false
 > = (
-    origCommand: (...args: any[]) => IsElement extends true ? $ElementCommands[ElementKey] : $BrowserCommands[BrowserKey],
+    this: IsElement extends true ? WebdriverIO.Element : WebdriverIO.Browser,
```
dprevost-LMI, line 213: I added this one, but it could be considered overkill as well. Let me know if you wish to remove that!

christian-bromann: Makes total sense to me 👍

christian-bromann:
> * Propose `overwriteBrowserCommand` and `overwriteElementCommand`, instead of using the boolean, to ease the API and not worry about missing the boolean or not. Will also reduce the typing of definitions
> * Propose `addBrowserCommand` and `addElementCommand`, instead of using the boolean, to ease the API and not worry about missing the boolean or not. Will also reduce the typing of definitions

I think this makes sense to me. We should deprecate the `overwriteCommand` then so we can eventually remove it.

dprevost-LMI:
> Great cleanup 👏 let's address above comment and get this merged.

Trying in the following days...

dprevost-LMI:
I will open another PR since I'm getting some challenges, so that we can merge the fixes! I reverted the proposition from the PR!

christian-bromann (approved):
LGTM 👍

dprevost-LMI:
> Propose overwriteBrowserCommand and overwriteElementCommand, instead of using the boolean, to ease the API and not worry about missing the boolean or not. Will also reduce the typing of definitions
Propose addBrowserCommand and addElementCommand, instead of using the boolean, to ease the API and not worry about missing the boolean or not. Will also reduce the typing of definitions
I think this makes sense to me. We should deprecate the overwriteCommand then so we can eventually remove it.

About the information above, I gave it a try, but I have some doubts, unfortunately.
1. `addCommand` is used much more internally than I initially thought.
2. After seeing more code, it made more sense to use `browser.addCommand` when not an element than `browser.AddBrowserCommand`.
3. I found [this code](https://github.com/dprevost-LMI/webdriverio/blob/655a3d6d0273efe02b2bbcc9299db4881b828f3e/packages/webdriverio/src/utils/getElementObject.ts#L103) where `addCommand` is used on an element, which I did not expect and discovered the API also applies to an element. Therefore, having `element.addBrowserCommand` there makes even less sense.

What would make more sense:
1. Separate the `addCommand` API on the browser/driver vs the one on the element
   2. On the element, we do not need the `attachToElement`, `proto` and `instances`
   3. This way, when adding `addElementCommand`, it would also not be added by default to the element API
2. Keep `addCommand` but deprecated only `attachToElement`, which would be replaced by `driver.addElementCommand`

So, as a prototyping start, it would look like the below (utopian thinking since we are currently forgetting for a minute that we need to support a deprecated API)
```ts
type BrowserTypeInstances = WebdriverIO.Browser | WebdriverIO.MultiRemoteBrowser
export interface BrowserInstanceCommands<T extends BrowserTypeInstances> {
    /**
     * Add a command to `browser`
     */
    addCommand (
        name: string,
        func: AddCommandFn,
        proto?: Record<string, any>,
        instances?: Record<string, T>
    ): void;

    /**
     * Add a command to elements
     */
    addElementCommand<Instance extends BrowserTypeInstances = WebdriverIO.Browser>(
        name: string,
        func: AddCommandFnScoped<T | Instance, true>,
        // Unsure if needed or not, maybe just internally needed if so we could separate from public API
        // proto?: Record<string, any>,
        // instances?: Record<string, Instances>
    ): void;

    /**
     * Overwrite a `browser` command
     */
    overwriteCommand<ElementKey extends keyof $ElementCommands, BrowserKey extends keyof $BrowserCommands>(
        name:  BrowserKey,
        func: OverwriteCommandFn<ElementKey, BrowserKey, false>,
        proto?: Record<string, any>,
        instances?: Record<string, BrowserTypeInstances>
    ): void;

    /**
     * Overwrite an element command
     */
    overwriteElementCommand<ElementKey extends keyof $ElementCommands, BrowserKey extends keyof $BrowserCommands>(
        name:  BrowserKey,
        func: OverwriteCommandFnScoped<ElementKey, BrowserKey, true>,
        // Unsure if needed or not, maybe just internally needed if so we could separate from public API
        // proto?: Record<string, any>,
        // instances?: Record<string, BrowserTypeInstances>
    ): void;

    /**
     * Create a custom selector
     */
    addLocatorStrategy(
        name: string,
        func: ((selector: string, root: HTMLElement) => CustomLocatorReturnValue) | ((selector: string) => CustomLocatorReturnValue)
    ): void
}

export interface ElementInstanceCommands<T extends WebdriverIO.Element> {
    /**
     * add command to `browser` or `element` scope
     */
    addCommand<IsElement extends boolean = false, Instance extends BrowserTypeInstances = WebdriverIO.Browser>(
        name: string,
        func: AddCommandFnScoped<T | Instance, IsElement>,
    ): void;

    // Needed?? Probably not...
    // overwriteCommand(): void;
    // addLocatorStrategy(): void
}
```
Seeing the complexity, either I just won't do it, or from now on, it will need its dedicated issue, so we can stamp all this properly instead of hijacking this issue 😅

christian-bromann:
@dprevost-LMI thanks for the explanation.

I agree with you. `addCommand` historically was implemented way before we had different scopes, e.g. `browser` and `element`. I hope soon I will finish my work on the new `page` scope which would mean we have even more scopes.

I have two options in my mind:
- only have `addCommand` and `overwriteCommand` but apply them on the scope based on which prototype you call it on, e.g. `browser.addCommand(...)` adds a browser command while `$(elem).addCommand(...)` adds a command to the element scope - the disadvantage here is that I need an element instance first to modify its prototype which seems less ideal
- have dedicated commands for each scope, e.g. `addBrowserCommand`, `addElementCommand` or `addPageCommand` but that seems verbose and requires to add a new command every time we add a new scope?

Let's create a new issue and get more feedback from the community.

PR #14657, opened 2025-07-22
Create `tsconfig.wdio.json` even when a root tsconfig exists

hammzj:
## Proposed changes
Addresses #14655.

This comes from a run of `npx wdio create`. Originally, if a `tsconfig.json` already exists, it would skip creating it but call reference an incorrect tsconfig file name in the output WDIO config. It would name it either as `tsconfig.wdio.json` or `tsconfig.e2e.json`, even if it does not exist.

After this change, a `tsconfig.wdio.json` will generated and referenced, regardless if a base `tsconfig.json` exists. However, it will not be created if an existing `tsconfig.wdio.json` file already exists. This address the issue with the WDIO config referencing the TSConfig incorrectlyt while also separating the types needed for WDIO testing purposes from the base repository types. 

If the user wants to merge these files together and only use one, they'll need to edit the base tsconfig and WDIO config their selves. 

## Types of changes

## Checklist

## Backport Request

## Further comments

Definitely would like some extra help on others testing the `npx wdio create` command on different machines (Linux, Windows). I don't have access to a good Linux environment.


packages/create-wdio/src/cli/utils.ts
```diff
@@ -93,27 +115,19 @@ export const parseAnswers = async function (yes: boolean): Promise<ParsedAnswers
     /**
      * find relative paths between tests and pages
      */
-    const hasRootTSConfig = await fs.access(path.resolve(projectRootDir, 'tsconfig.json')).then(() => true, () => false)
-    const tsConfigFilePath = !hasRootTSConfig
+    const hasRootTSWdioConfig = await fs.access(path.resolve(projectRootDir, 'tsconfig.wdio.json')).then(() => true, () => false)
```
hammzj: Ah, I need to edit this to use the constant.
```suggestion
    const hasRootTSWdioConfig = await fs.access(path.resolve(projectRootDir, TSCONFIG_FILE_NAME)).then(() => true, () => false)
```

hammzj:
> Thanks for raising the PR!
> 
> I think, having a separate TS config for WebdriverIO makes sense. I just wonder, in a new project, an IDE won't recognize any types without having a default `tsconfig.json` right? In this case should we create two files, a basic `tsconfig.json` and a `tsconfig.wdio.json` that extends from the basic one?

@christian-bromann 
Sadly this is _kind of_ the case, but not quite. I think this is more of an IDE problem than a TypeScript problem. For example, it is possible to run `tsc --project ./tsconfig.wdio.json`; [See here](https://www.typescriptlang.org/docs/handbook/compiler-options.html#compiler-options).

However, in IntelliJ WebStorm for example, I can't get it to recognize a different TSConfig using `--project` as an option. What I can do, though, is point a reference to the other `tsconfig.wdio.json` in the base file, and then the types are recognized.

Therefore, we don't want to extend it from a base config in this case. Rather, it needs to be added to `references` in the base file.

<details>
<summary>Before adding a reference</summary>

```json
{
  "compilerOptions": {
    "target": "es2016",
    "module": "commonjs",
    "esModuleInterop": true,
    "forceConsistentCasingInFileNames": true,
    "strict": true,
    "skipLibCheck": true
  }
}
```

Here, `url` is not recognized on the `browser` object.

Screenshot 2025-07-23 at 10 31 05 AM

</details>

<details>
<summary>After adding a reference</summary>

```json
{
  "compilerOptions": {
    "target": "es2016",
    "module": "commonjs",
    "esModuleInterop": true,
    "forceConsistentCasingInFileNames": true,
    "strict": true,
    "skipLibCheck": true
  },
  "references": [
    {
      "path": "./tsconfig.wdio.json"
    }
  ]
}
```

Now there are no type errors.

Screenshot 2025-07-23 at 10 31 19 AM
</details>

***
### New idea

That above answers that issue, but does bring in additional complication:

If we assume we do **not** want to mutate the base `tsconfig.json` if one already exists, then we're back to square one.
However, if we allow it in this one case, there's a few steps that can  be taken.
1. Check if a base `tsconfig.json` exists. If not, create it empty, and also create a `tsconfig.wdio.json`.
2. Check if `references` already exists in the base `tsconfig.json`.
3. If not, create `"references"` with an empty array in the base `tsconfig.json`.
4. Then push a new `{ "path": "./tsconfig.wdio.json" }` into `references`.
5. Finally, set the `tsConfigPath: "tsconfig.json"` in `wdio.conf.ts`.

YET, we also still need to consider the case if the user wants files autogenerated, which in that case, WDIO just points to a nested `./tests/tsconfig.json`. But in this case, the IDE still won't recognize the types...so let's just remove the need to create a nested `./tests/tsconfig.json` entirely.

It also might be worthwhile to ask the user for acceptance on mutating their original tsconfig before the script does it automatically.

I think this whole process was made way more complex than it needs to be, but the steps above seem to take into consideration everything to get it running. Let me know your thoughts.

Edit: I do understand that, when no base file exists, making a blank `tsconfig.json` with just a reference may look a bit odd. However, this process is declarative, so we know what to expect in either case if one exists or not. I am open to creating a tsconfig if one doesn't exist, though as it was done originally.

christian-bromann:
@hammzj I think it makes sense. Create an "empty" `tsconfig.json` if non exist and add a reference to a create `tsconfig.wdio.json`.

hammzj:
Ok awesome. I'll look again.

hammzj:
Well, this is getting more complicated than I originally thought. The options for these tsconfig files get messy (thanks Microsoft). Here's some additional considerations, which now concern me more.

I've not been able to actually run the initialization command locally (need to figure out how to use `npx` for a linekd package), but I was testing it out manually. With a reference in the `tsconfig.json` to the `tsconfig.wdio.json`, we can't actually run a `tsc` command without additional changes:

<details>
<summary>Error output</summary>

```
hammzj ~/Documents/GitHub/hammzj/webdriver-io-7-bug (dev/tsconfig-alternate-name) $ tsc
tsconfig.json:11:5 - error TS6306: Referenced project '/Users/hammzj/Documents/GitHub/hammzj/webdriver-io-7-bug/test/specs/tsconfig.wdio.json' must have setting "composite": true.

11     {
       ~
12       "path": "./test/specs/tsconfig.wdio.json"
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
13     }
   ~~~~~

tsconfig.json:11:5 - error TS6310: Referenced project '/Users/hammzj/Documents/GitHub/hammzj/webdriver-io-7-bug/test/specs/tsconfig.wdio.json' may not disable emit.

11     {
       ~
12       "path": "./test/specs/tsconfig.wdio.json"
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
13     }
   ~~~~~
```

</details>

So, there's a few things that could work, but this gets hairy as we don't want to break the base config.

I've tried with adding `composite: true` and replacing `noEmit` with `emitDeclarationOnly: true`, but there's still some problems:

1. The config and spec files are still being emitted as JS which is undesired.

<details>
<summary> 2. The global WDIO types are not being registered. </summary>

```
hammzj ~/Documents/GitHub/hammzj/webdriver-io-7-bug (dev/tsconfig-alternate-name) $ tsc  
test/specs/test.e2e.ts:5:23 - error TS2339: Property 'url' does not exist on type 'Browser'.

5         await browser.url(`https://the-internet.herokuapp.com/login`)
                        ~~~

wdio.conf.ts:1:34 - error TS2694: Namespace 'global.WebdriverIO' has no exported member 'Config'.

1 export const config: WebdriverIO.Config = {
                                   ~~~~~~


Found 2 errors in 2 files.

Errors  Files
     1  test/specs/test.e2e.ts:5
     1  wdio.conf.ts:1
```

</details>


***

Really, this makes me think we're back at square one.

Now, I tried running a `tsc -p tsconfig.wdio.json` and things worked as normal. But, running a `tsc` will also create source files for the WebDriverIO, even without references. So, I think there's a lot more that needs figured out in terms of how to separate test types from app types. The base `tsconfig.json` would need to exclude all test and config files used by WebDriverIO in order to work properly, while the `tsconfig.wdio.json` could handle everything else. I don't think we want to further modify a base configuration to do this.

PR #14658, opened 2025-07-22
polish: simplify deprecation warning to prevent it causing issues

erwinheitzman:
## Proposed changes

I found a case where the warning was still shown, therefore I simplified the approach so that it cannot accidentally show anymore.

## Types of changes

## Checklist

### DIFF
--- a/packages/webdriver/src/command.ts
+++ b/packages/webdriver/src/command.ts
@@ -48,12 +48,10 @@ export default function (
          *
          * Note: there are situations where we have to use deprecated commands, e.g. `switchToFrame`
          * internally where we don't want to have the message shown to the user. In these cases we
-         * use the `DISABLE_WEBDRIVERIO_DEPRECATION_WARNINGS` env variable to suppress the message.
+         * use the suppress the message.
          */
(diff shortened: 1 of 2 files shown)


christian-bromann:
> I found a case where the warning was still shown

Can you describe that case?

erwinheitzman:
> > I found a case where the warning was still shown
> 
> Can you describe that case?

Not specifically, with the v9 upgrade at work we saw multiple warnings when running the tests while there's no reason for the warnings to show. I think it might be related to running tests in parallel where the environment variable is shared causing it to be toggled globally for a short time for all tests.

christian-bromann:
@erwinheitzman we had an issue reported recently around this which we fixed, mind trying with latest version of WebdriverIO?

erwinheitzman:
Sure but this PR simplifies things regardless right?

christian-bromann:
It simplifies the code but misses the intention of actually printing the deprecation command. I actually want to have the message printed if a user uses `switchToFrame`. I _only_ want to _not_ print it if we use it internally.

christian-bromann:
Closing due to inactivity.

Issue #14216, opened 2025-02-21
[🐛 Bug]: Thrown Error In Test Stops Spec Reporter From Giving A Summary


pegnovi:
### Have you read the Contributing Guidelines on issues?

- [x] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

v9.9.3

### Node.js Version

v20.11.0

### Mode

WDIO Testrunner

### Which capabilities are you using?

```typescript
Not sure if the bug is related to capabilities but here is what's in my wdio.conf.js

{
    browserName: "chrome",
}
```

### What happened?

When a test is running and it throws an error (regardless of whether it's caught in a catch block or not)  the spec reporter sometimes does not display the summary of test results.
It is inconsistent behaviour.

### What is your expected behavior?

In the terminal where the wdio test runner is run, I should see this summary towards the end:

 "spec" Reporter:
------------------------------------------------------------------
[chrome 133.0.6943.127 mac #0-0] Running: chrome (v133.0.6943.127) on mac
[chrome 133.0.6943.127 mac #0-0] Session ID: 3f778a02b48c2c263c2ab9c6ed38eea6
[chrome 133.0.6943.127 mac #0-0]
[chrome 133.0.6943.127 mac #0-0] » file:///test/test.js
[chrome 133.0.6943.127 mac #0-0] Dummy Test
[chrome 133.0.6943.127 mac #0-0]    ✓ Just Passes
[chrome 133.0.6943.127 mac #0-0]    ✖ Dummy Test
[chrome 133.0.6943.127 mac #0-0]    ✓ Just Passes Again
[chrome 133.0.6943.127 mac #0-0]
[chrome 133.0.6943.127 mac #0-0] 2 passing (925ms)
[chrome 133.0.6943.127 mac #0-0] 1 failing
[chrome 133.0.6943.127 mac #0-0]
[chrome 133.0.6943.127 mac #0-0] 1) Dummy Test Dummy Test
[chrome 133.0.6943.127 mac #0-0] Failed to fetch
[chrome 133.0.6943.127 mac #0-0] TypeError: Failed to fetch
[chrome 133.0.6943.127 mac #0-0]     at sendPostRequestNoCatch 

### How to reproduce the bug.

I have a mini sample project here: https://github.com/HassHz/sampleworkspace/tree/errorNoSummary
All it does is make wdio run a test.js.
[...]

PR #14667, opened 2025-07-25
Moved browser runner's checkForTestError

pegnovi:
## Proposed changes
This PR aims to fix an issue with the browser runner where when an error is thrown or an error is logged, it will be caught by the `checkForTestError` function and the test runner will end without running any succeeding tests.
This also prevents tests from purposely simulating a scenario where an error may be thrown and to properly handle it within the test.

As discussed in this (issue)[https://github.com/webdriverio/webdriverio/issues/14216], the suggested solution is to move the error logic to be near the addInitScript and to just check for errors at the beginning (instead of running it at a set interval while the tests are running).

List of changes:
- Moved browser runner's checkForTestError to be run right after url loading and removed logic to call it at a regular interval.
- Added mocha-browser test to run mocha tests with the browser runner.
  - This can be tested by going into the `webdriverio/examples/wdio` folder and running the following commands:
    - `pnpm test:mocha-browser` - This will run mocha tests that contain failure scenarios with the browser runner.
    - `pnpm test:mocha-browser --startFailure=vite` - This will simulate a Vite error which will be caught at the beginning and end the test runner.
    - `pnpm test:mocha-browser --startFailure=browserLoad` - This will simulate a browser load error which will be caught at the beginning and end the test runner.

The user can provide their own error handling logic either through a catch block in the test itself or through addInitScript with browser object or BiDi usage.

## Types of changes

## Checklist

## Backport Request

## Further comments


packages/wdio-runner/src/browser.ts
```diff
@@ -131,6 +130,9 @@ export default class BrowserFramework implements Omit<TestFramework, 'init'> {
          */
         if (!this._config.sessionId) {
             await browser.url(`/?cid=${this._cid}&spec=${(new URL(spec)).pathname}`)
+            // The above triggers the addInitScript if any was provided.
+            // We check for initial errors right after it runs.
+            this.#checkForTestError()
```
christian-bromann, line 135: We can't just run `#checkForTestError` once , it would only check for errors at the beginning and that's it. The better approach is to also update the `checkForTestError` function to use `addInitScript` to allow the browser to push error events to the worker.

pegnovi: Do you mean I would alter the logic for `checkForTestError` such that
- It still doesn't get run at a set interval (as in this PR)
- It gets run once but listens for browser errors via the Browser Object (https://webdriver.io/docs/api/browser/)
- It passes whatever error it finds to `addInitScript` where the user determines what to do with with the error.

examples/wdio/mocha-browser/.eslintrc
```diff
@@ -0,0 +1,5 @@
+{
+  "env": {
+    "mocha": true
+  }
+}
```
christian-bromann, line 5: We are using Eslint v9 which doesn't support this anymore afaik

packages/wdio-runner/src/browser.ts
christian-bromann: Yes, instead of using `setInterval` to check for errors, we should use `addInitScript` which can notify the Node.js environment (e.g. `packages/wdio-runner/src/browser.ts`) that an error has been thrown.

PR #14668, opened 2025-07-27
feat(wdio-cli): add duration to CLI summary

NaamuKim:
## Proposed changes
Added execution duration tracking to WebDriverIO CLI across 4 phases: setup, prepare, execute, and complete.

#14495

## Types of changes

## Checklist

## Backport Request

## Further comments

I've added duration logging as a proof of concept with minimal modifications to the existing codebase.
If you have any additional feedback on WebDriverIO requirements or suggestions for the 4-stage process I've implemented, please feel free to share them extensively. I'll incorporate the feedback as quickly as possible.

### DIFF
--- a/packages/wdio-cli/src/duration.ts
+++ b/packages/wdio-cli/src/duration.ts
@@ -28,7 +28,7 @@ class DurationTracker {
             `complete ${formatDuration(complete)}`
         ].join(', ')
 
-        return `Duration: ${chalk.green(formatDuration(total))} (${breakdown})`
+        return `${chalk.green(formatDuration(total))} (${breakdown})`
     }
 
     reset(): void {
@@ -37,18 +37,10 @@ class DurationTracker {
     }
 }
 
-function formatDuration(ms: number): string {
+export function formatDuration(ms: number): string {
     if (ms >= 60000) {return `${(ms / 60000).toFixed(1)}m`}
     if (ms >= 1000) {return `${(ms / 1000).toFixed(1)}s`}
     return `${ms}ms`
 }
 
-const defaultTracker = new DurationTracker()
-
-export default {
-    start: defaultTracker.start.bind(defaultTracker),
-    end: defaultTracker.end.bind(defaultTracker),
-    getSummary: defaultTracker.getSummary.bind(defaultTracker),
-    reset: defaultTracker.reset.bind(defaultTracker),
-    formatDuration
-}
+export default new DurationTracker()
--- a/packages/wdio-cli/src/duration.ts
+++ b/packages/wdio-cli/src/duration.ts
@@ -1,0 +1,54 @@
+import chalk from 'chalk'
+
+class DurationTracker {
+    private timers: Record<string, number> = {}
+    private durations: Record<string, number> = {}
+
+    start(phase: string): void {
+        this.timers[phase] = performance.now()
+    }
+
+    end(phase: string): number {
+        if (!this.timers[phase]) {return 0}
+        const duration = Math.round(performance.now() - this.timers[phase])
+        delete this.timers[phase]
+
+        this.durations[phase] = duration
+        return duration
+    }
+
+    getSummary(): string {
+        const { setup = 0, prepare = 0, execute = 0, complete = 0 } = this.durations
+        const total = setup + prepare + execute + complete
+
+        const breakdown = [
+            `setup ${formatDuration(setup)}`,
+            `prepare ${formatDuration(prepare)}`,
+            `execute ${formatDuration(execute)}`,
+            `complete ${formatDuration(complete)}`
+        ].join(', ')
+
+        return `Duration: ${chalk.green(formatDuration(total))} (${breakdown})`
+    }
+
+    reset(): void {
+        this.timers = {}
+        this.durations = {}
+    }
+}
+
+function formatDuration(ms: number): string {
+    if (ms >= 60000) {return `${(ms / 60000).toFixed(1)}m`}
+    if (ms >= 1000) {return `${(ms / 1000).toFixed(1)}s`}
+    return `${ms}ms`
+}
+
+const defaultTracker = new DurationTracker()
+
+export default {
+    start: defaultTracker.start.bind(defaultTracker),
+    end: defaultTracker.end.bind(defaultTracker),
+    getSummary: defaultTracker.getSummary.bind(defaultTracker),
+    reset: defaultTracker.reset.bind(defaultTracker),
+    formatDuration
+}
--- a/packages/wdio-cli/src/interface.ts
+++ b/packages/wdio-cli/src/interface.ts
@@ -357,11 +357,11 @@ export default class WDIOCLInterface extends EventEmitter {
         }
 
         return this.log(
-            '\nSpec Files:\t', chalk.green(this.result.passed, 'passed') + ', ' + retries + failed + skipped + totalJobs, 'total', `(${percentCompleted}% completed)`, 'in',
+            '\nSpec Files:\t', chalk.green(this.result.passed, 'passed') + ', ' + retries + failed + skipped + totalJobs, 'total', `(${percentCompleted}% completed)`,
             this.#hasShard()
                 ? `\nShard:\t\t ${this._config.shard!.current} / ${this._config.shard!.total}`
                 : '',
-            `\n${duration.getSummary()}`
+            `\nDuration:\t ${duration.getSummary()}`
         )
     }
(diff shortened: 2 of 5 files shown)


packages/wdio-cli/src/duration.ts
```diff
@@ -0,0 +1,46 @@
...
+class DurationTracker {
+    private timers: Record<string, number> = {}
+    private durations: Record<string, number> = {}
+
+    start(phase: string): void {
+        this.timers[phase] = performance.now()
+    }
+
+    end(phase: string): number {
+        if (!this.timers[phase]) {return 0}
```
christian-bromann:
```suggestion
        if (!this.timers[phase]) { return 0 }
```

packages/wdio-cli/src/interface.ts
```diff
@@ -357,11 +357,11 @@ export default class WDIOCLInterface extends EventEmitter {
         }
 
         return this.log(
-            '\nSpec Files:\t', chalk.green(this.result.passed, 'passed') + ', ' + retries + failed + skipped + totalJobs, 'total', `(${percentCompleted}% completed)`, 'in', elapsed,
+            '\nSpec Files:\t', chalk.green(this.result.passed, 'passed') + ', ' + retries + failed + skipped + totalJobs, 'total', `(${percentCompleted}% completed)`,
             this.#hasShard()
                 ? `\nShard:\t\t ${this._config.shard!.current} / ${this._config.shard!.total}`
                 : '',
-            '\n'
+            `\nDuration:\t ${duration.getSummary()}`
```
christian-bromann: I am not sure if users always want to see this. I think we should just log it with the WDIO logger so people who are interested can find it.

christian-bromann (commented):
I think we should move the `DurationTracker` to the `@wdio/utils` package so we can also log the performance within the worker.

packages/wdio-cli/src/interface.ts
NaamuKim: Sounds possible!
I've changed it to use the WDIO logger with info level logging.

NaamuKim:
> I think we should move the `DurationTracker` to the `@wdio/utils` package so we can also log the performance within the worker.

I've moved the DurationTracker to the @wdio/utils package as suggested.
I think updating the logging in the worker should be straightforward too, but it might be clearer to split that into a separate PR. 
What do you think?
I was thinking something like this for the logging:.
```
[INFO] @wdio/runner: [0-0] Worker completed in 2.1s (worker-lifecycle 2.1s)
[INFO] @wdio/runner: [0-1] Worker completed in 1.9s (worker-lifecycle 1.9s)
[INFO] @wdio/runner: [1-0] Worker completed in 2.3s (worker-lifecycle 2.3s)
```
If any extra Duration details are needed in the worker, those can be included during the update.

packages/wdio-cli/src/interface.ts
NaamuKim: Sounds possibble! I’ve updated it to log with the WDIO logger at the info level.

packages/wdio-cli/src/launcher.ts
```diff
@@ -65,10 +67,12 @@ class Launcher {
     }
 
     /**
-     * run sequence
-     * @return  {Promise}  that only gets resolved with either an exitCode or an error
-     */
+    * run sequence
+    * @return  {Promise}  that only gets resolved with either an exitCode or an error
+    */
```
christian-bromann:
```suggestion
     * run sequence
     * @return  {Promise}  that only gets resolved with either an exitCode or an error
     */
```

packages/wdio-cli/src/launcher.ts
```diff
@@ -140,12 +145,20 @@ class Launcher {
                 setupBrowser(config, caps)
             ])
 
+            duration.end('prepare')
+            duration.start('execute')
+
             exitCode = await this._runMode(config, caps)
+
+            duration.end('execute')
+            log.info('TEST INFO - should be visible')
```
christian-bromann:
```suggestion
```

packages/wdio-cli/src/interface.ts
```diff
@@ -356,6 +357,7 @@ export default class WDIOCLInterface extends EventEmitter {
             snapshotNotes.forEach((note) => this.log(note))
         }
 
+        log.info('Duration:\t', chalk.green(duration.getSummary()))
```
christian-bromann: Can you provide an example output of the `wdio.log` file?

christian-bromann:
> I was thinking something like this for the logging:.

I am not quite sure what the `worker-lifecycle` is referring to 🤔  it also seems that the duration is identical. Is this worth adding then?

Maybe we can add the total duration to the worker result log, e.g.:

```
[0-0] PASSED in chrome - file:///test/specs/test.e2e.ts
```

becomes

```
[0-0] PASSED in chrome in 2.3s - file:///test/specs/test.e2e.ts
```

Thoughts?

packages/wdio-cli/src/interface.ts
NaamuKim:
```
2025-08-16T12:02:24.021Z INFO @wdio/cli: Duration:	 12.6s (setup 5ms, prepare 392ms, execute 12.2s, complete 0ms)
```
It looks like this. It would be better to remove the `\t`. I'll remove it.

NaamuKim:
```
2025-08-16T12:16:45.594Z INFO @wdio/cli: Duration: 12.2s (setup 4ms, prepare 243ms, execute 11.9s, complete 0ms)
```
How about this one

NaamuKim:
> ```
> [0-0] PASSED in chrome in 2.3s - file:///test/specs/test.e2e.ts
> ```
> 
> Thoughts?

I really like your opinion.
Then I'll add just the total duration measured in the same way.

NaamuKim:
```
[safari (unknown) #0-0] ✖ Failed to create a session:
[safari (unknown) #0-0] Unable to connect to "http://localhost:49273/", make sure browser driver is running on that address.
```
this failure is not for this request

packages/wdio-utils/src/duration.ts
```diff
@@ -0,0 +1,44 @@
+class DurationTracker {
+    private timers: Record<string, number> = {}
+    private durations: Record<string, number> = {}
```
christian-bromann: Since the `getSummary` function explicitly expects certain types of duration I think it makes sense to apply some stricter typing here.

packages/wdio-utils/src/duration.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    start(phase: string): void {
```
christian-bromann: Similar here and in `end`.

christian-bromann: I think it makes sense to check if a phase has been started before and if so throw an error. We don't want to start a phase multiple times and get skewed results.

christian-bromann (changes requested):

packages/wdio-utils/src/duration.ts
NaamuKim: I agree! I've added validation to prevent duplicate phase starts, which could corrupt timing data.

tests/smoke.runner.js
```diff
@@ -19,6 +19,31 @@ const allPassedWildCardConfig = path.resolve(__dirname, 'tests-cli-spec-arg/wdio
...
+    }
+}
+duration.end = function(phase) {
+    try {
+        return originalEnd(phase)
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    } catch (e) {
+        return 0
+    }
+}
```
christian-bromann, line 46: This should not be required. It seems like the current way the duration tracker is implemented will not allow users to run multiple WDIO sessions at the same time using the launcher programatically. I think this is a limitation we can't put on users given the value this tracker provides.

packages/wdio-utils/src/duration.ts
```diff
@@ -0,0 +1,60 @@
...
+    }
+}
+
+export function formatDuration(ms: number): string {
+    if (ms >= 60000) {return `${(ms / 60000).toFixed(1)}m`}
+    if (ms >= 1000) {return `${(ms / 1000).toFixed(1)}s`}
+    return `${ms}ms`
+}
+
+export default new DurationTracker()
```
christian-bromann, line 60: Why do we export an instance of the `DurationTracker` here?

christian-bromann (changes requested):

Issue #14588, opened 2025-06-25
[🐛 Bug]: switchWindow() throws WebDriverError: no such window: target window already closed


PratikKumarSdet:
### Have you read the Contributing Guidelines on issues?

- [x] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

9.16

### Node.js Version

22.10

### Mode

WDIO Testrunner

### Which capabilities are you using?

```typescript

```

### What happened?

This is happening with specific scenarios of switchWindow() case. Suppose there is a parent page which has a hyperlink that opens a following/child url in a new browser window(not tab). The child window has a button that closes the child window. Here if we are using switchWindow() to switch back control to parent window after closing the child window, it is throwing the **WebDriverError: no such window: target window already closed; from unknown error: web view not found**. Although the execution continues, the console is full of this errors.

### How to reproduce the bug.

Create 2 html files with the code given below:

**index.html**

```
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8" />
  <title>Modern Web Development</title>
  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
</head>
<body>
  <h1>Welcome to Modern Web Development</h1>
  <p>
    Web development today involves design, performance, and accessibility. Explore
    <a href="#" onclick="openInNewWindow('https://developer.mozilla.org/en-US/docs/Web/JavaScript')">
      vanilla JavaScript</a> documentation, or review
    <a href="#" onclick="openChildWindow()" id="hyperlinkTexted">best practices on web.dev</a> to stay ahead in the field.
  </p>

  <script>
    function openInNewWindow(url) {
      window.open(url, '_blank', 'width=900,height=700,resizable=yes');
    }

    function openChildWindow() {
      window.open(
        'child.html',
        'ChildWindow',
        'width=700,height=600,resizable=yes'
      );
    }
  </script>
</body>
</html>

```

**child.html:**

```
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8" />
  <title>Web Development Best Practices</title>
  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
</head>
<body>
  <h1>Web Development Best Practices</h1>
  <p>
    Staying ahead in web development means continuously evolving your skills and understanding the core concepts of the web. 
[...]

PR #14712, opened 2025-08-19
fix(webdriverio): don't print error when swithcing to an existing window

phantomwolf:
…dow (#14588)

## Proposed changes

### Why?

- In `switchWindow`, `getWindleHandle()` is always called.
- When current window is closed by JavaScript `window.close()`, `getWindowHandle()` will fail and log error messages(retry 3 times and take 1.6 secs on my side).
- #14588 : although the test is switching to an available window, `getWindleHandle()` failed and printed errors, causing sub-optimal user experience.

### What?
Introduced a cache for the current window handle, so that we can get current window handle without calling `getWindleHandle()`.

### How?

- Use ContextManager's `onCommandResultBidiAndClassic` hook to cache results of `getWindowHandle` and `switchToWindow` on success.
- Add `ContextManager.getCurrentWindowHandle()` to return the cache.
- Always initialize ContextManager when remote is created, so that it can start capturing `getWindowHandle` and `switchToWindow` commands from the beginning.

### Risk: out of sync
When JavaScript `window.close()` is called, the cache will still return the previous window handle, which is out of sync with reality.

### Why is the fix still valid?
Only `window.close()` can invalidate the cache. However, if the cache value is still included in the results of `getWindowHandles()`, it means the window is not closed and the cache value is still valid.

In switchWindow() command, first call `tabs = getWindowHandles()` to get all open window handles, and only compare `matcher` with `getCurrentWindowHandle()` if `matcher` is included in `tabs`. This guarantees that the value of the cache is valid.

## Types of changes

## Checklist

## Backport Request

## Further comments

### DIFF
--- a/packages/webdriverio/src/session/context.ts
+++ b/packages/webdriverio/src/session/context.ts
@@ -25,6 +25,7 @@ export class ContextManager extends SessionManager {
     #mobileContext?: string
     #isNativeContext: boolean
     #getContextSupport = true
+    #currentWindowHandle?: string
 
     constructor(browser: WebdriverIO.Browser) {
         super(browser, ContextManager.name)
@@ -109,18 +110,43 @@ export class ContextManager extends SessionManager {
         }
     }
 
-    #onCommandResultBidiAndClassic(event: { command: string, result: unknown }) {
+    #onCommandResultBidiAndClassic(event: { command: string, result: unknown, body: unknown }) {
         /**
          * the `closeWindow` command returns:
          *   > the result of running the remote end steps for the Get Window Handles command, with session, URL variables and parameters.
          */
         if (event.command === 'closeWindow') {
+            // Clear cached window handle
+            this.#currentWindowHandle = undefined
+
             const windowHandles = (event.result as { value?: string[] }).value || []
             if (windowHandles.length === 0) {
                 throw new Error('All window handles were removed, causing WebdriverIO to close the session.')
             }
             this.#currentContext = windowHandles[0]
             return this.#browser.switchToWindow(this.#currentContext)
+        }
+
+        /**
+         * Update current window handle when 'getWindowHandle' succeeds
+         */
+        if (event.command === 'getWindowHandle') {
+            const windowHandle = (event.result as { value?: string }).value || undefined
+            this.#currentWindowHandle = windowHandle
+        }
+
+        /**
+         * the `closeWindow` command returns:
+         *  > the body of the Close Window command request, including the window handle it's switching to.
+         *  > the result of the Close Window command, either an error object on failure, or null data on success.
+         */
+        if (event.command === 'switchToWindow') {
+            const err = (event.result as { error?: unknown }).error || undefined
+            // Only update current window handle when 'switchToWindow' has no error
+            if (!err) {
+                const windowHandle = (event.body as { handle?: string }).handle || undefined
+                this.#currentWindowHandle = windowHandle
+            }
         }
     }
 
@@ -242,6 +268,23 @@ export class ContextManager extends SessionManager {
             return this.initialize()
         }
         return this.#currentContext
+    }
+
+    /**
+     * Sets the cached current window handle value.
+     * @param handle current window handle to set
+     */
+    setCurrentWindowHandle (handle?: string) {
+        this.#currentWindowHandle = handle
+    }
+
+    /**
+     * Returns the cached window handle.
+     *
+     * @returns the current window handle, or undefined if the current window is closed.
+     */
+    getCurrentWindowHandle() {
+        return this.#currentWindowHandle
     }
 
     get isNativeContext() {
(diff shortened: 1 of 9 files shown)


erwinheitzman:
Hi @phantomwolf could you add some tests to verify the new behaviour and sign the CLA? Your code looks good to me so with the above it would be good to merge imo 👍

phantomwolf:
Hi @erwinheitzman, thanks for reviewing. I've signed the CLA. I’ll proceed with the tests next, both existing and new ones.

packages/webdriverio/src/session/context.ts
```diff
@@ -122,6 +123,28 @@ export class ContextManager extends SessionManager {
             this.#currentContext = windowHandles[0]
             return this.#browser.switchToWindow(this.#currentContext)
         }
+
+        /**
+         * Update current window handle when 'getWindowHandle' succeeds
+         */
+        if (event.command === 'getWindowHandle') {
+            const windowHandle = (event.result as { value?: string }).value || undefined
```
erwinheitzman, line 134: Can you explain why you explicitly assign undefined here when the value is falsy? Same for line 141

erwinheitzman:
@phantomwolf I've been giving the caching some more thought and was wondering, the cache can get out of sync with reality right?

If that is true, what do we then do? Considering this might be a posential risk, does caching really help much here?

Let me know if you feel like I am overlooking something but I think if the handles update and a handle no longer exists, it might still be in the cache, potentially causing side effects. I'm curious to hear your thoughts on this.

phantomwolf:
@erwinheitzman Yes, the cache can be outdated in some situations(e.g. window closed by javascript inside webpage). The main reason for introducing this cache is for fixing [bug 14588](https://github.com/webdriverio/webdriverio/issues/14588).

**Context:** in this bug, the current window was closed by javascript. When switching to another window, switchWindow() will first call `getWindowHandle()` to get current window handle, which will fail and print error.

**How does the cache fix it:** in this situation, `getWindowHandles()` still works. If we compare the cached value with the results of `getWindowHandles()`, we can know whether the current window is still open without calling `getWindowHandle()`. That's why I did the `matcher === await contextManager.getCurrentWindowHandle()` check after calling `getWindowHandles()`.

I'm not sure if this is the best solution to fix it. Feedback appreciated.

erwinheitzman:
@phantomwolf thank you for the clear explanation. I like the changes you made but I feel that the original issue might not be fixed with this, especially when `const currentWindow = await this.getWindowHandle().catch(() => undefined)` is already supposed to ignore the error mentioned in the original post to begin with. Have you been able to verify that your changes fix the original issue locally?

phantomwolf:
@erwinheitzman It's true that `await this.getWindowHandle().catch(() => undefined)` catches the exception, but `getWindowHandle()` will keep failing(retry 3 times, took 1.6 secs on my laptop) and printing the following error messages:

```
[0-0] from unknown error: web view not found
[0-0]   (Session info: chrome=137.0.7151.120) when running "window" with method "GET"
[0-0] 2025-06-25T09:44:39.499Z WARN webdriver: WebDriverError: no such window: target window already closed
```

This is because `switchWindow` calls `getWindowHandle()` no matter which window it's switching to. The [bug 14588](https://github.com/webdriverio/webdriverio/issues/14588) is about "The test shouldn't print errors because it's switching another window which is still open".

My idea is to find a way to get the current window handle without calling `getWindowHandle()`. Currently not sure if this cache can be used for other use cases.

packages/webdriverio/src/session/context.ts
phantomwolf: When `getWindowHandle` fail, it indicates there's no current window(probably closed outside webdriverio). So I'm setting the cache to `undefined` to indicate there's no current window handle.

phantomwolf:
Hi @erwinheitzman @christian-bromann I've added unit tests and updated the pull request description for better clarity.

ccharnkij:
Hi all, it looks fine to me. However, I wonder if it's more beneficial do this at the source of this logging issue so that it will affect all webdriver commands. So I found that the log comes from here: https://github.com/webdriverio/webdriverio/blob/main/packages/webdriver/src/request/request.ts#L188

Perhaps, addressing that logging line would be more impactful?

phantomwolf:
This approach does look more straightforward, but I’m a bit concerned that completely removing this error log may make debugging harder. Alternatively, silencing it conditionally for `getWindowHandle` would:

- hide error logs for other `getWindowHandle` use cases.
- introduce tighter coupling between wdio-protocols and webdriver.

IMHO, the `retry` method might not be the best place to handle this.

@ccharnkij Could you share more on how we can address this logging line?

ccharnkij:
I'm not sure, tbh. I was thinking, the error is already threw, so maybe that line can be deleted and just let that propagate up on its own? If it's a legit error, the test would fail and the error will be displayed. Otherwise, like in this case, nothing would appear. @christian-bromann ?

phantomwolf:
Hi @christian-bromann @erwinheitzman @ccharnkij Any updates on this PR?

christian-bromann:
Thank you for the contribution 👍

Issue #14713, opened 2025-08-19
[💡 Feature]: Allow element’s custom command on non-visible element


dprevost-LMI:
### Is your feature request related to a problem?

I wanted to implement a method called `swipeToAndPress`, which combines swiping up or down in a list to locate an element and then pressing it once found. However, I discovered that to use an element’s custom command, the element must exist and be visible first unless the command has a specific name, such as `isClickable` or `isDisplayed`.

For now, I've added a patch to include the prefix swipe [in this line](https://github.com/dprevost-LMI/webdriverio/blob/b6c4be917a8d7f4e8822a525dd38fc3df842f05a/packages/webdriverio/src/utils/implicitWait.ts#L15), which allows me to move forward. However, configuring the custom command to remove this requirement could be a better option.

### Describe the solution you'd like.

In the next week, I’ll try to tackle a PR so that we can pass some configuration option on the element custom command to disable the `implicitWait` on the element.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

dprevost-LMI:
In progress there: https://github.com/webdriverio/webdriverio/pull/14718

PR #14718, opened 2025-08-24
feat: Draft solution to exclude element custom command to implicitWait

dprevost-LMI:
## Proposed changes

For [this issue](https://github.com/webdriverio/webdriverio/issues/14713), where we want to be able to run an element custom command even when the element does not exist/is not displayed.

## Types of changes

- [X] Polish (an improvement for an existing feature)

## Checklist

## Backport Request

- [X] This change is solely for `v9` and doesn't need to be back-ported

## Further comments

Missing UTs

### DIFF
--- a/packages/webdriverio/src/index.ts
+++ b/packages/webdriverio/src/index.ts
@@ -18,6 +18,7 @@ import { environment } from './environment.js'
 
 import type { AttachOptions } from './types.js'
 import type * as elementCommands from './commands/element.js'
+import { IMPLICIT_WAIT_EXCLUSION_LIST } from './middlewares.js'
 
 export * from './types.js'
 export const Key = KeyConstant
@@ -63,7 +64,7 @@ export const remote = async function(
 
     const { Driver, options } = await getProtocolDriver({ ...params, ...config })
     const prototype = getPrototype('browser')
-    const instance = await Driver.newSession(options, modifier, prototype, wrapCommand) as WebdriverIO.Browser
+    const instance = await Driver.newSession(options, modifier, prototype, wrapCommand, IMPLICIT_WAIT_EXCLUSION_LIST) as WebdriverIO.Browser
 
     /**
      * we need to overwrite the original addCommand and overwriteCommand
@@ -172,6 +173,7 @@ export const multiremote = async function (
      */
     if (!isStub(automationProtocol)) {
         const origAddCommand = driver.addCommand.bind(driver)
+        // TODO dprevost need to add disableElementImplicitWait here ?
         driver.addCommand = (name: string, fn, attachToElement) => {
             driver.instances.forEach(instanceName =>
                 driver.getInstance(instanceName).addCommand(name, fn, attachToElement)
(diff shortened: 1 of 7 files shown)


dprevost-LMI:
Back to draft to investigate `✖ supports custom command registration in before hook`

packages/webdriverio/src/index.ts
```diff
@@ -172,9 +173,9 @@ export const multiremote = async function (
      */
     if (!isStub(automationProtocol)) {
         const origAddCommand = driver.addCommand.bind(driver)
-        driver.addCommand = (name: string, fn, attachToElement) => {
+        driver.addCommand = (name: string, fn, attachToElement, _proto, _instance, disableElementImplicitWait) => {
             driver.instances.forEach(instanceName =>
-                driver.getInstance(instanceName).addCommand(name, fn, attachToElement)
+                driver.getInstance(instanceName).addCommand(name, fn, attachToElement, undefined, undefined, disableElementImplicitWait)
```
christian-bromann: This doesn't seem like the right step forward: having to pass in `undefined` twice. How about we make the 3rd parameter an option object that holds the remaining information. I think this can be implemented in a backwards compatible way.

christian-bromann (changes requested):
I would be against adding another parameter to the `addCommand` function and instead transform the last parameters into an configuration object.

packages/webdriverio/src/index.ts
dprevost-LMI: I agree; I didn't like it, but it was the easiest. I will look to support two object types and start using an object and deprecate the former parameter

dprevost-LMI: I was able to deprecated the former function and now have this one with an `options` object:
```ts
  /**
   * add command to `browser` or to an `element` when using options.attachToElement to true
   */
  addCommand<IsElement extends boolean = false, Instance extends Instances = WebdriverIO.Browser>(
      name: string,
      func: IsElement extends true ? AddCommandFnScoped<T | Instance, IsElement> : AddCommandFn,
      options?: CustomCommands.CustomCommandOptions<IsElement>
  ): void;

...

export type CustomCommandOptions<IsElement extends boolean = false> = {
    attachToElement?: IsElement,
    proto?: Record<string, unknown>,
    instances?: Record<string, Instances>,
    disableElementImplicitWait?: boolean
}
```
Did this fulfill your request?

packages/webdriverio/src/types.ts
```diff
@@ -217,18 +217,31 @@ type OverwriteCommandFn<
...
 
-type Instances = WebdriverIO.Browser | WebdriverIO.MultiRemoteBrowser
+export type Instances = CustomCommands.Instances
+export type CustomCommandOptions<IsElement extends boolean> = CustomCommands.CustomCommandOptions<IsElement>
+export type AddCommandFunction<IsElement extends boolean, T = any, Instance = WebdriverIO.Browser> = IsElement extends true ? AddCommandFnScoped<T | Instance, IsElement> : AddCommandFn
 
 export interface CustomInstanceCommands<T> {
+
     /**
+     * @deprecated use the below overload with options object instead
```
christian-bromann: The user won't see any other overload docs, only the one the overload applies to
```suggestion
     * @deprecated use option object as 3rd parameter
```

christian-bromann (approved):
LGTM 👍

PR #14914, opened 2025-12-04
fix(protocols): update WebDriver & Appium protocols

eglitise:
## Proposed changes

Recently I noticed a few minor inconsistencies between some WebDriver & Appium protocol endpoints and their WDIO implementations. This PR aims to fix most of them, while also updating relevant documentation:
* `setTimeouts`: `null` values should be allowed ([ref](https://w3c.github.io/webdriver/#dfn-timeouts-configuration))
* `addVirtualAuthenticator`: `protocol` and `transport` should be required since they do not have a default value ([ref](https://www.w3.org/TR/webauthn-2/#sctn-automation-add-virtual-authenticator))
  * This could be argued as a breaking change, but I am not sure this endpoint would have worked without providing these properties anyway.
* `addCredential`: `userHandle` should be optional and `signCount` should allow `null` values ([ref](https://www.w3.org/TR/webauthn-2/#sctn-automation-add-credential))
  * `signCount` has also been changed to optional to fix TS issues, though it is required according to protocol. I believe this is still better than the other option of incorrectly keeping `userHandle` as required.
* `setUserVerified`: missing `isUserVerified` property ([ref](https://www.w3.org/TR/webauthn-2/#sctn-automation-set-user-verified))
  * Similarly to `addVirtualAuthenticator`, might seem like a breaking change, but in practice it is unlikely to have worked without this property.

* `installApp`/`activateApp`/`removeApp`/`terminateApp`/`isAppInstalled`/`queryAppState`: should support `bundleId` and/or `options` parameters
* `backgroundApp`: restoration can be prevented with a `-1` value, not `null` (docs issue)
* `getEvents`: `type` can also be `string` and should be optional

I've also added deprecation marks for all Appium/MJSONWP commands removed in Appium 3, and updated the documentation links for all remaining Appium commands.

There are 2 other Appium endpoints that I have not changed here, as they could be considered breaking:
* `getDeviceTime` uses the `GET` endpoint, but Appium also supports the `POST` endpoint with an optional `format` property
* `stopRecordingScreen` should be aligned with `startRecordingScreen` and only accept a single optional `options` property

## Types of changes

- [x] Specification changes (updates to WebDriver command specifications)

## Checklist

## Backport Request

## Further comments


wswebcreation (approved):
LGTM, thanks 

@christian-bromann , WDYT?

christian-bromann (approved):

Issue #13663, opened 2024-09-26
[💡 Feature]: V9 Browser Logs


DanLambe:
### Is your feature request related to a problem?

With previous versions of WebdriverIO there was always a browser.log file with the context from the browser console of what was happening with the different sessions and runs. With WebdriverIO V9 this feature was removed making it difficult to debug within a CI/CD pipeline or catch issues within the browser during test runs.

### Describe the solution you'd like.

I'd like a flag within the config to enable this behavior or a reporter that brings feature parity to WebdriverIO V9 for gathering context around the browser and its behavior.

### Describe alternatives you've considered.

A service, or reporter would work. As long as there is a nice way of seeing what was happening within the browser per session basis.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

christian-bromann:
What do you think of elevating browser logs as part of the CLI output?

> As long as there is a nice way of seeing what was happening within the browser per session basis.

How should that look like? I assume it would be helpful to know which test is currently running and which browser logs have been reported as part of it, right?

I would prefer to rather integrate this into the reporter instead of creating a new reporter for it.

sebastian-sauer:
After we've added the addWorkerLogs option to junit-reporter @christian-bromann  already mentioned to include console logs, too. It's currently on my list of things i'd like to add soon (as with bidi this should be a very small thing to do).

erwinheitzman:
Great suggestion.

I'd prefer it if we could configure whether we want to output to a file and/or the terminal.

christian-bromann:
Please keep in mind that browser logs can become confusing when you run multiple browser tests in parallel.

One thing we could do is always logging the browser logs via `log.debug(...)` in the worker so everyone gets them in the terminal output or in the file depending on whether `outputDir` is set or not. I like this option because it doesn't require introducing any new options.

If we want to go the route of additionally printing browser logs somewhere else, e.g. in the terminal output, let's make sure to connect the log to the worker id at least, ideally to the test that is being executed.

DanLambe:
> How should that look like? I assume it would be helpful to know which test is currently running and which browser logs have been reported as part of it, right?

[...]

PR #15049, opened 2026-01-25
feat(runner): add logBrowserConsole option to capture browser logs

mccmrunal:
This feature enables capturing browser logs (via BiDi or classic) and printing them to the worker stdout or log file. Includes safety cleanup for log streams.

## Proposed changes
This PR introduces a new configuration option logBrowserConsole (default: false) to the 
Options.ts
 definition.

When enabled:

Runner: Listens for log.entryAdded events (BiDi support) and emits a client:logEntry event to all reporters.
SpecReporter: Implements 
onClientLogEntry
 to print browser logs directly in the terminal, interleaved with test results for better context (Display).
BaseReporter: Handles file logging. If outputDir is configured, logs are written to a dedicated file wdio-{workerId}-browserconsole.log for debugging (Archival).
WDIOReporter: Defines the event interface allowing any reporter to subscribe to browser logs.
This architecture ensures separation of concerns: the Runner handles data capture, concrete reporters handle display, and the base reporter handles archival. This addresses the need for better visibility into browser-side events during test execution in V9, as discussed in #13663.

## Types of changes

## Checklist

## Backport Request

## Further comments
This implementation aligns with the discussion in #13663 regarding opting-in to browser logs to avoid terminal noise, while providing a dedicated file for debugging when needed. The generated log file is created only if outputDir is specified, ensuring no simplified setups are cluttered. Values for logBrowserConsole are strictly validated to prevent crashes on unknown log levels from the browser.

### DIFF
--- a/packages/wdio-runner/src/reporter.ts
+++ b/packages/wdio-runner/src/reporter.ts
@@ -1,5 +1,8 @@
 import path from 'node:path'
+import fs from 'node:fs'
+import os from 'node:os'
 import logger from '@wdio/logger'
+
 import DotReporter from '@wdio/dot-reporter'
 import { initializePlugin } from '@wdio/utils'
 import type { Options, Capabilities, Reporters } from '@wdio/types'
@@ -15,6 +18,7 @@ const mochaAllHooks = ['"before all" hook', '"after all" hook']
 export default class BaseReporter {
     private _reporters: Reporters.ReporterInstance[] = []
     private listeners: ((ev: unknown) => void)[] = []
+    private _browserConsoleLogStream?: fs.WriteStream
 
     constructor(
         private _config: Options.Testrunner,
@@ -28,6 +32,13 @@ export default class BaseReporter {
         this._config.reporters = this._config.reporters || []
         if (this._config.reporters.length === 0) {
             this._config.reporters.push([DotReporter, {}])
+        }
+
+        if (this._config.logBrowserConsole && this._config.outputDir) {
+            const logFileName = `wdio-${this._cid}-browserconsole.log`
+            const logFile = path.resolve(this._config.outputDir, logFileName)
+            fs.mkdirSync(path.dirname(logFile), { recursive: true })
+            this._browserConsoleLogStream = fs.createWriteStream(logFile, { flags: 'w' })
         }
     }
 
@@ -58,6 +69,11 @@ export default class BaseReporter {
         retry?: number,
     }) {
         payload.cid = this._cid
+
+        if (e === 'client:logEntry') {
+            const logEntry = payload as unknown as { level: string, text: string }
+            this._browserConsoleLogStream?.write(logEntry.text + os.EOL)
+        }
 
         /**
          * Send failure message (only once) in case of test or hook failure
@@ -196,6 +212,15 @@ export default class BaseReporter {
                 // wait otherwise
             }, this._config.reporterSyncInterval)
         })
+    }
+
+    /**
+     * close log stream
+     */
+    closeStream () {
+        if (this._browserConsoleLogStream) {
+            this._browserConsoleLogStream.end()
+        }
     }
 
     /**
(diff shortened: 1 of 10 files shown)


BorisOsipov:
IMO, it must be the reporter, not part of the runner.

mccmrunal:
@BorisOsipov 
Thanks for the feedback!

I initially placed it in the Runner because it has direct access to the browser instance and the BiDi log.entryAdded event, which seemed like the simplest path for capturing the data.

However, I agree that 
BaseReporter
 is the correct place for output handling to ensure proper separation of concerns. I have refactored the PR so that the Runner now only emits a client:logEntry event, and 
BaseReporter
 handles the actual logging and file stream management.

BorisOsipov:
> BaseReporter
> handles the actual logging and file stream management.

The base reporter should only provide a way to access client:logEntry events, while concrete reporters should handle printing them to the console or sending them to a server or smth else.

For example, this feature currently exists in the[spec reporter](https://webdriver.io/docs/spec-reporter#addconsolelogs). Perhaps it would be better to enhance it.

mccmrunal:
@BorisOsipov  please review the changes again and please let me know your feedback! Thanks!

packages/wdio-types/src/Options.ts
```diff
@@ -275,6 +275,14 @@ export interface Testrunner extends Hooks, WebdriverIO, WebdriverIO.HookFunction
...
      */
     groupLogsByTestSpec?: boolean,
+    /**
+     * Add console.logs from browser to your local log output.
+     * These logs will be in the browserconsole namespace and loglevel is according to
+     * the browsers log level.
+     *
+     * @default false
+     */
+    logBrowserConsole?: boolean
```
christian-bromann: Do we still want this, given it should be handled by the reporter (e.g. spec reporter)

packages/wdio-runner/src/reporter.ts
```diff
@@ -59,6 +70,11 @@ export default class BaseReporter {
     }) {
         payload.cid = this._cid
 
+        if (e === 'client:logEntry') {
+            const logEntry = payload as unknown as { level: string, text: string }
+            this._browserConsoleLogStream?.write(logEntry.text + os.EOL)
+        }
```
christian-bromann: As @BorisOsipov also mentioned I suggest we leave this up to the reporter to handle it.

packages/wdio-runner/src/reporter.ts
```diff
@@ -29,6 +33,13 @@ export default class BaseReporter {
         if (this._config.reporters.length === 0) {
             this._config.reporters.push([DotReporter, {}])
         }
+
+        if (this._config.logBrowserConsole && this._config.outputDir) {
+            const logFileName = `wdio-${this._cid}-browserconsole.log`
+            const logFile = path.resolve(this._config.outputDir, logFileName)
+            fs.mkdirSync(path.dirname(logFile), { recursive: true })
+            this._browserConsoleLogStream = fs.createWriteStream(logFile, { flags: 'w' })
+        }
```
christian-bromann: Having a separate log file with only browser logs makes it rather difficult to connect that to a certain place within a test. Wdyt?

christian-bromann (changes requested):
Following what @BorisOsipov said: let's simplify elevating console.logs within reporter like spec reporter.

packages/wdio-types/src/Options.ts
mccmrunal: I agree. I removed the separate browser log file creation in wdio-runner. Browser logs are now emitted as events and handled directly by reporters

packages/wdio-runner/src/reporter.ts
mccmrunal: I have removed the global logBrowserConsole option from wdio-types. Log display is now controlled at the reporter level. For example, I have added an addBrowserConsoleLogs option to wdio-spec-reporter.

packages/wdio-runner/src/reporter.ts
mccmrunal: Done, removed the log writing logic from wdio-runner. The runner now simply emits the client:logEntry event, and I implemented the handler in wdio-spec-reporter to display these logs with appropriate color formatting

christian-bromann:
Will go ahead and close this since the process has been staled and it is unclear to me what the value of this is.

Issue #15026, opened 2026-01-19
Website footer shows outdated copyright year (2025)


Tilakmahajan:
### Pre-check

- [x] I know I can edit the docs but prefer to file this issue

### Describe the improvement

Other

### Description of the improvement / report

The footer on https://webdriver.io currently displays:

Copyright © 2025 OpenJS Foundation

Since it is now 2026, this appears to be caused by the documentation
site not being rebuilt since last year.

As the docs are statically generated and deployed via a GitHub Actions
pipeline to S3, triggering a rebuild/redeploy should update the year
automatically if it is generated dynamically.

<img width="377" height="319" alt="Image" src="https://github.com/user-attachments/assets/3bd45d24-da3c-46bc-97f4-7a3f7a04f10d" />

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

christian-bromann:
Thanks for flagging, want to raise a PR for this?

Tilakmahajan:
Yes, I can take this up 👍
I’ll open a PR to update the footer so the copyright year is generated dynamically.

Tilakmahajan:
I think the footer logic is already correct — it uses new Date().getFullYear() in website/docusaurus.config.ts.

I noticed there are also recent commits in January 2026 (including docs-related ones), so the source is active. That said, I think it’s still possible the production docs site wasn’t fully rebuilt or is serving cached output.

A manual docs rebuild or cache refresh might update the year automatically.

christian-bromann:
Thanks @Tilakmahajan and @mccmrunal for the contributions.

PR #15054, opened 2026-01-27
fix(docs): update footer copyright year to 2026

Tilakmahajan:
Fixes -#15026

### What this PR does

- Updates the footer copyright year shown on https://webdriver.io
- Fixes a hardcoded year in the i18n footer translation file

### Context

The footer was still displaying **2025** even though the site and docs are active in 2026.  
The year was coming from a hardcoded value in the footer translation JSON rather than being updated automatically.


mccmrunal:
Thanks for the PR!
While this fixes the immediate issue, it means we would have to manually update this file again next year repeated chore.
I have raised a PR , I think its a better approach and think it will fix this issue once and for all

PR #15065, opened 2026-01-29
fix(junit-reporter): handle Jasmine filename-only paths

### LINKED ISSUE #13052: [🐛 Bug]: Jasmine and the JUnit reporter output empty spec files since v8.15.9 due to suite file property not including path

### WebdriverIO Version

v8.39.0

### Node.js Version

v18.17.1

### Mode

WDIO Testrunner

_No response_

### What happened?

I recently bumped our WDIO dependencies for latest Chrome compatibility, and noticed that our JUnit reports started coming out empty:

```
<?xml version="1.0" encoding="UTF-8"?>
<testsuites/>
```

I pinned this down to the changes in https://github.com/webdriverio/webdriverio/releases/tag/v8.15.9 , and in particular https://github.com/webdriverio/webdriverio/pull/10977 (fixing report test duplication). In order to address the duplicate reports, "same file" detection was added, comparing the spec file name to the suite file path; [on main this is done here](https://github.com/webdriverio/webdriverio/blob/main/packages/wdio-junit-reporter/src/index.ts#L274).

When running with the Jasmine framework, the `suite.file` is a single filename (eg `concurrency.spec.js`) while the `specFileName` is a full file path (eg `/Users/rowan/code/.../__test__/tests/specs/headless/concurrency.spec.js`). So the duplicate detection rejects the log, and no suite results get output.

Digging through the various pieces involved, I can see that [wdio-reporter has some code to try to provide](https://github.com/webdriverio/webdriverio/blob/c38781b567a344b47a59c9401d30d398dd7799d1/packages/wdio-reporter/src/index.ts#L76-L85) a `file` if the jasmine reporter doesn't provide it, but this doesn't appear to get invoked. Instead, [the jasmine-reporter does appear to supply a file property](https://github.com/webdriverio/webdriverio/blob/main/packages/wdio-jasmine-framework/src/reporter.ts#L204) - but it's using only the filename, not the filepath.

This is somewhat of a post-and-run - I don't have time to do a full repro case today, so I'm creating this bug as a placeholder so I will remember to update it later. I'm fairly confident that the jasmine-reporter needs to supply a full filepath but I don't yet see how best to do that :grin

### How to reproduce the bug.

(Use jasmine framework + junit reporter. Repro project to follow)

### Relevant log output

```typescript
"junit" Reporter:
<?xml version="1.0" encoding="UTF-8"?>
<testsuites/><?xml version="1.0" encoding="UTF-8"?>
<testsuites/>

Spec Files:	 2 passed, 2 total (100% completed) in 00:00:28  
```
```


### Code of Conduct

lancecasey:
## Summary
- Fix JUnit reporter producing empty XML files when used with Jasmine framework
- Jasmine provides only filenames (e.g., `happyPath.spec.js`) while `runner.specs` contains full paths
- Updated `_sameFileName()` to compare basenames when either path is just a filename (no directory separator)

## Root Cause
The `_sameFileName()` method compared full paths, but Jasmine's `suite.file` only contains the filename without the path, causing the comparison to always fail and reject all test results.

## Test Plan
- [x] Added unit test `_sameFileName - compares basenames when one path is just a filename (issue #13052)`
- [x] Verified all existing junit-reporter tests pass (42 tests)

Fixes #13052

### DIFF
--- a/packages/wdio-junit-reporter/src/index.ts
+++ b/packages/wdio-junit-reporter/src/index.ts
@@ -1,4 +1,5 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
+import path from 'node:path'
 import url from 'node:url'
 import os from 'node:os'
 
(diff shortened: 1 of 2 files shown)


christian-bromann (approved):
LGTM 👍

christian-bromann:
Closing in favor of #15616

PR #15096, opened 2026-02-13
fix(webdriverio): handle dynamic function payloads in mock.respond

mccmrunal:
## Proposed changes
  fixes #14926
  Fixed an issue where mock.respond() would fail when provided with a function payload because it eagerly serialized the function to undefined, creating an invalid response body.

  ## Types of changes
  
  [//]: # (What types of changes does your code introduce to WebdriverIO?)
  [//]: # (_Put an `x` in the boxes that apply_)
  
  - [ ] Polish (an improvement for an existing feature)
  - [x] Bugfix (non-breaking change which fixes an issue)
  - [ ] New feature (non-breaking change which adds functionality)
  - [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
  - [ ] Documentation update (improvements to the project's docs)
  - [ ] Specification changes (updates to WebDriver command specifications)
  - [ ] Internal updates (everything related to internal scripts, governance documentation and CI files)
  
  ## Checklist
  
  [//]: # (_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._)
  
  - [x] I have read the [CONTRIBUTING](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md) doc
  - [x] I have added tests that prove my fix is effective or that my feature works
  - [ ] I have added the necessary documentation (if appropriate)
  - [x] I have added proper type definitions for new commands (if appropriate)
  
  ## Backport Request
  
  [//]: # (The current `main` branch is the development branch for WebdriverIO v9. If your change should be released to the current major version of WebdriverIO (v8), please raise another PR with the same changes against the `v8` branch.)
  
  - [x] This change is solely for `v9` and doesn't need to be back-ported
  - [ ] Back-ported PR at `#XXXXX`
  
  ## Further comments
  The initial issue was caused by JSON.stringify returning undefined for functions. The fix ensures that the function is preserved and executed when the network event occurs, allowing dynamic modification of the response. A regression test case was added to index.test.ts to verify this behavior.
  
  ### Reviewers: @webdriverio/project-committers

PR #15120, opened 2026-02-28
fix: add afterEach cleanup to context management E2E tests

elaichenkov:
## Problem

The `context management` E2E tests in `test.e2e.ts` had no cleanup between tests. When one test timed out (e.g. due to a slow external website), it left extra browser windows open, corrupting the browser state for every subsequent test and causing a **cascade of timeouts** through the `context management`, `switchFrame`, and `open resources` test groups.

This has been causing intermittent CI failures across multiple PRs (e.g. #15119, #15096).

## Root Cause

The `closeAllWindowsButFirst` helper existed but was only called at the **start** of individual tests, not after failures. When a test timed out mid-window-switch:
1. Extra windows remained open
2. The next test inherited a polluted browser state
3. Window count assertions (`length === 3`) would never match because the expected count was based on leftover windows from prior tests

## Fix

1. **Add `afterEach` hook** that calls `closeAllWindowsButFirst()` — ensures each test starts with a clean single-window state, even after failures
2. **Fix window count expectations** in the last test — with cleanup, each test starts from 1 window:
   - After clicking a link that opens a new tab: expect **2** windows (was 3)
   - After closing one window: expect **1** remaining (was 2)

### DIFF
--- a/e2e/wdio/headless/test.e2e.ts
+++ b/e2e/wdio/headless/test.e2e.ts
@@ -521,6 +521,10 @@ describe('main suite 1', () => {
             await browser.switchToWindow(allHandles[0])
         }
 
+        afterEach(async () => {
+            await closeAllWindowsButFirst()
+        })
+
         it('should allow user to switch between contexts', async () => {
             await browser.url('https://guinea-pig.webdriver.io/')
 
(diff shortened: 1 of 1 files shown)


christian-bromann:
Thanks a ton!

PR #15141, opened 2026-03-08
fix(appium): rename deprecated Appium protocol commands for Appium 3 compatibility

### LINKED ISSUE #15124: [🐛 Bug]: Appium 3 deprecated/removed a lot of endpoints

### WebdriverIO Version

latest

### Node.js Version

latest

### Mode

WDIO Testrunner

```typescript

```

### What happened?

Appum 3 deprecated a lot of endpoints, see https://appium.io/docs/en/3.0/guides/migrating-2-to-3/, we need to be aware of this and update WDIO

_No response_

### How to reproduce the bug.

-

### Relevant log output

```typescript
-
```

### Code of Conduct

wswebcreation:
> [!IMPORTANT] 
> After merge and release we should also release the docs

## Related issue

Closes #15124

## Summary

Appium 3 removed ~90 JSONWP-era endpoints from `base-driver`. This PR contains multiple steps/fixes to make WebdriverIO compatible with Appium 3 while keeping backward compatibility with Appium 2 users on older drivers.

## Research

Most of the research is in #15124. Here's a summary

### Backward Compatibility & `mobile:` Command Availability

After further research into the driver changelogs, a clearer picture emerged on how to handle backward compatibility with Appium 2.

#### When were the `mobile:` replacements introduced?

For **UIAutomator2 (Android)**, all relevant `mobile:` replacement commands (`mobile: lock`, `mobile: unlock`, `mobile: isLocked`, `mobile: getCurrentPackage`, `mobile: getCurrentActivity`, `mobile: pressKey`, `mobile: gsmCall`, `mobile: getPerformanceData`, `mobile: getSystemBars`, `mobile: openNotifications`, etc.) were available before January 2024 — added progressively in UIAutomator2 v2.17–v2.24 (April–May 2023).

For **XCUITest (iOS)**, the same applies for most commands. The one exception is `mobile: startXCTestScreenRecording` / `mobile: stopXCTestScreenRecording`, which landed in XCUITest **v7.3.0 (March 2024)**.

#### What this means

There are three user groups

| Group | What works | Risk |
|---|---|---|
| Appium 2 + drivers from 2023 or earlier | Old endpoint only | Calling `mobile:` would fail |
| Appium 2 + drivers from 2024+ | Old endpoint **and** `mobile:` | None, both paths work |
| Appium 3 + drivers from 2024+ | `mobile:` only | Old endpoint still works via driver grace period, but shows deprecation warning |

The group that would break if WDIO switched unconditionally to `mobile:` is **Appium 2 users on drivers from 2023 or earlier**. This is a small and shrinking group, but worth protecting.

#### Chosen approach: try/catch with specific error handling

WDIO will attempt the `mobile:` execute path first. If the driver responds with an "unknown command" error (meaning the driver is too old to support it), WDIO falls back to the old endpoint and logs a clear deprecation warning. Any other error (device disconnected, wrong parameters, etc.) is re-thrown immediately and not swallowed by the catch.

This means:
- **Appium 3 + 2024+ drivers**: `mobile:` succeeds, no warning, no double call
- **Appium 2 + 2024+ drivers**: `mobile:` also succeeds, same as above
- **Appium 2 + 2023 or earlier drivers**: `mobile:` fails with unknown command → caught → falls back to old endpoint → deprecation warning logged once

The deprecation warning will tell users which driver they need to upgrade to to remove the old endpoint dependency.

### What this PR does (protocol layer)

- **Renames 37 protocol commands** that have a `mobile:` replacement by prefixing them with `appium` (e.g. `lock` → `appiumLock`, `getCurrentPackage` → `appiumGetCurrentPackage`). The original clean names are reserved for upcoming mobile command wrappers with try/catch fallback logic.
- **Adds missing `deprecated` notices** to `startRecordingScreen`, `stopRecordingScreen`, `rotateDevice`, and all five IME commands.
- **Commands without a known `mobile:` replacement** (`rotateDevice`, IME commands, JSONWP touch commands) keep their original names with a generic deprecation message.
- **Updates internal call sites** (`getContexts`, `switchContext`, `relaunchActiveApp`, `saveRecordingScreen`) that used old protocol names, with TODO comments pointing to the upcoming mobile command implementations.

### Architecture decision

Two-layer strategy:
1. Protocol commands, raw HTTP endpoints, prefixed with `appium`, marked deprecated
2. Mobile commands, smart wrappers that try `mobile:` execute first, catch "unknown command" errors, fall back to the `appium*` protocol endpoint for Appium 2 backward compat, and log a deprecation warning

This means users calling e.g. `driver.lock()` will transparently get the Appium 3-compatible implementation via the mobile command layer, with no breaking changes.

### Commands renamed (37 total)

| Old name | New protocol name | Future mobile command |
|---|---|---|
| `shake` | `appiumShake` | `shake` |
| `lock` | `appiumLock` | `lock` |
| `unlock` | `appiumUnlock` | `unlock` |
| `isLocked` | `appiumIsLocked` | `isLocked` |
| `startRecordingScreen` | `appiumStartRecordingScreen` | `startRecordingScreen` |
| `stopRecordingScreen` | `appiumStopRecordingScreen` | `stopRecordingScreen` |
| `getCurrentPackage` | `appiumGetCurrentPackage` | `getCurrentPackage` |
| `getCurrentActivity` | `appiumGetCurrentActivity` | `getCurrentActivity` |
| `pressKeyCode` | `appiumPressKeyCode` | `pressKeyCode` |
| `longPressKeyCode` | `appiumLongPressKeyCode` | `longPressKeyCode` |
| `background` | `appiumBackground` | `background` |
| `launchApp` | `appiumLaunchApp` | `launchApp` |
| `closeApp` | `appiumCloseApp` | `closeApp` |
| `getStrings` | `appiumGetStrings` | `getStrings` |
| `getSystemBars` | `appiumGetSystemBars` | `getSystemBars` |
| `getDisplayDensity` | `appiumGetDisplayDensity` | `getDisplayDensity` |
| `openNotifications` | `appiumOpenNotifications` | `openNotifications` |
| `startActivity` | `appiumStartActivity` | `startActivity` |
| `touchId` | `appiumTouchId` | `touchId` |
| `toggleEnrollTouchId` | `appiumToggleEnrollTouchId` | `toggleEnrollTouchId` |
| `toggleAirplaneMode` | `appiumToggleAirplaneMode` | `toggleAirplaneMode` |
| `toggleData` | `appiumToggleData` | `toggleData` |
| `toggleWiFi` | `appiumToggleWiFi` | `toggleWiFi` |
| `toggleLocationServices` | `appiumToggleLocationServices` | `toggleLocationServices` |
| `toggleNetworkSpeed` | `appiumToggleNetworkSpeed` | `toggleNetworkSpeed` |
| `gsmCall` | `appiumGsmCall` | `gsmCall` |
| `gsmSignal` | `appiumGsmSignal` | `gsmSignal` |
| `gsmVoice` | `appiumGsmVoice` | `gsmVoice` |
| `powerCapacity` | `appiumPowerCapacity` | `powerCapacity` |
| `powerAC` | `appiumPowerAC` | `powerAC` |
| `sendSms` | `appiumSendSms` | `sendSms` |
| `fingerPrint` | `appiumFingerPrint` | `fingerPrint` |
| `setClipboard` | `appiumSetClipboard` | `setClipboard` |
| `getClipboard` | `appiumGetClipboard` | `getClipboard` |
| `getPerformanceDataTypes` | `appiumGetPerformanceDataTypes` | `getPerformanceDataTypes` |
| `getPerformanceData` | `appiumGetPerformanceData` | `getPerformanceData` |
| `sendKeyEvent` | `appiumSendKeyEvent` | `pressKeyCode` |

## Test plan

- [x] All existing tests pass
- [x] Mobile command wrappers with try/catch fallback
- [x] TypeScript interface in `packages/wdio-protocols/src/commands/appium.ts` updated to new names (follow-up)
- [x] Unit tests for each new mobile command

### DIFF
--- a/packages/webdriver/tests/index.test.ts
+++ b/packages/webdriver/tests/index.test.ts
@@ -295,7 +295,6 @@ describe('WebDriver', () => {
                 }
             }) as unknown as TestClient
             expect(client.isMobile).toBe(true)
-            // TODO: update to use new mobile command names once `isLocked` and `shake` mobile commands are implemented
             expect(client.appiumIsLocked).toBeTruthy()
             expect(client.appiumShake).toBeTruthy()
         })


packages/webdriverio/tests/utils/mobile.test.ts
```diff
@@ -1,7 +1,55 @@
...
+import path from 'node:path'
+import { describe, it, expect, vi, beforeEach } from 'vitest'
+import logger from '@wdio/logger'
+import { calculateAndroidPinchAndZoomSpeed, getMobileContext, getNativeContext, isUnknownMethodError, logAppiumDeprecationWarning, validatePinchAndZoomOptions } from '../../src/utils/mobile.js'
+
+vi.mock('@wdio/logger', () => import(path.join(process.cwd(), '__mocks__', '@wdio/logger')))
+const log = logger('test')
 import type { PinchAndZoomOptions } from '../../src/types.js'
 
+describe('isUnknownMethodError', () => {
```
dprevost-LMI: FYI: Using the function itself allows renaming to follow
```suggestion
describe(isUnknownMethodError, () => {
```

wswebcreation: ahh, never knew this, thanks. I'll push a change for all tests

dprevost-LMI:
From the below, does that mean that until people migrate to Appium 3, there will be a non-working call followed by a working one that falls back on Appium 2?
>Mobile commands, smart wrappers that try mobile: execute first, catch "unknown command" errors, fall back to the appium* protocol endpoint for Appium 2 backward compat, and log a deprecation warning

If so, is there any better way to do this so it is not constantly doing "failing request"?
I did ask AI yesterday, but nothing bullet-proof was really found, but we could have maybe some alternate ways, like
- Query an endpoint that ensures it is Appium 3 vs 2
- Use some kind of config or capability that could be easily provided so that we target the right version on the first try.

Note: 
- I'm not a big open source maintainer, so maybe your approach is more robust and should be the way to go! 
- Also, maybe it is easier to just upgrade to Appium 3 instead of trying the above options in the code!

wswebcreation:
Thanks for the feedback. In theory you are right, but...


- appium (2 and 3) have a server package and separate drivers
- The `mobile:..` command is supported by the server and the drivers since a long time
- only user group 3, appium 2 with very old drivers will have issues. These issues will already be with the drivers because they are "too old" and will give issues with the OS version of the phone. They would be from around 2023/2024. For appium drivers very old
- there is a way to check if people have appium 3, but not a way to check if they have the right drivers (very old drivers won't even work with Appium 3), so that could be an option and I've considered this, but then we would loose user group 2 (appium 2 with more recent drivers, 2024 and up)


Hope this makes sense

eglitise:
Hi, I'd love to assist with this from Appium side, especially since Appium Inspector uses WDIO under-the-hood and entirely depends on its supported endpoints and commands.

The described implementation approach generally seems good to me. I suppose it may also make sense to drop support for group 3 users in WDIO v10, but that's up to you.

One thing I would like to point out is that Claude seems to have made a few mistakes regarding removed commands:
* IME related commands have not been removed and are still available in Appium 3: https://appium.io/docs/en/latest/reference/api/jsonwp/ They are however deprecated. The UiAutomator2 driver does also support them separately from the Appium server, but before we remove them from the server, I think it would first make sense to add their `mobile:` equivalents, so that WDIO users have a clear migration path once WDIO eventually removes the deprecated endpoint. For now I would suggest leaving them unchanged here.
* `rotateDevice` is also still available in Appium 3, so no changes needed: https://appium.io/docs/en/latest/reference/api/mjsonwp/
* `startRecordingScreen`/`stopRecordingScreen` do not actually have a direct `mobile:` equivalent, however, the commands themselves are still supported in all the main drivers, so I would probably suggest leaving its WDIO implementation as-is.
* The replacement for `toggleNetworkSpeed` is `mobile: networkSpeed`, not `mobile: setNetworkSpeed`

wswebcreation:
Hi @eglitise 

Thanks for looking into this and for the feedback. This wasn't all Claude, it's just so helpful when fixing a lot of commands in the same way 🤣 , but in the analysis part Claude and I missed a few things you pointed out.

Here's my feedback


> The described implementation approach generally seems good to me. I suppose it may also make sense to drop support for group 3 users in WDIO v10, but that's up to you.

Thanks, the challenge here is that a lot of cloudvendors don't have updated drivers, or they have the old drivers connected to old OS versions. So we want to still be "backwards compatible"

> * IME related commands have not been removed and are still available in Appium 3: https://appium.io/docs/en/latest/reference/api/jsonwp/ They are however deprecated. The UiAutomator2 driver does also support them separately from the Appium server, but before we remove them from the server, I think it would first make sense to add their `mobile:` equivalents, so that WDIO users have a clear migration path once WDIO eventually removes the deprecated endpoint. For now I would suggest leaving them unchanged here.

We need to soften the message, but we will proced with introducing the same command names with the `mobile: equivelant so we are ready

> * `rotateDevice` is also still available in Appium 3, so no changes needed: https://appium.io/docs/en/latest/reference/api/mjsonwp/

will revert this

> * `startRecordingScreen`/`stopRecordingScreen` do not actually have a direct `mobile:` equivalent, however, the commands themselves are still supported in all the main drivers, so I would probably suggest leaving its WDIO implementation as-is.

This is what is in the docs
image

I see that the new command implementation isn't platform specific, so I need to fix that


> * The replacement for `toggleNetworkSpeed` is `mobile: networkSpeed`, not `mobile: setNetworkSpeed`

This was my bad, will fix this

wswebcreation:
@eglitise ,

See the fixes. This should fix all points there were mentioned by you

eglitise:
> We need to soften the message, but we will proced with introducing the same command names with the `mobile: equivelant so we are ready

To clarify, these commands are only deprecated in the main Appium server (since they are platform-specific), but they are _not_ deprecated in the UiAutomator2/Espresso drivers (which are the only drivers where these commands are relevant), so the WDIO command will still work the same way once they are removed from the Appium server.

That's why I suggest leaving them unchanged for now. If the Appium team decides to later deprecate them in the drivers and add `mobile:` equivalents, then yes, it would make sense to add the deprecation warning and wrapper in WDIO as well.

> This is what is in the docs

Yes, it says that the endpoints have been moved to drivers, which includes all the main ones, so they are effectively still supported, and there's no need to change the WDIO implementation.
(I do agree that this is confusing; I feel like the endpoints are universal enough to still be used in the main server)

Also, `XCTestScreenRecording`/`MediaProjectionRecording` use different functionality under the hood, and have a different set of accepted input parameters, so they would not work as drop-in replacements.

wswebcreation:
@eglitise 

> To clarify, these commands are only deprecated in the main Appium server (since they are platform-specific), but they are not deprecated in the UiAutomator2/Espresso drivers (which are the only drivers where these commands are relevant), so the WDIO command will still work the same way once they are removed from the Appium server.

This means the endpoint is passed directlty from the server to the driver?

> Yes, it says that the endpoints have been moved to drivers, which includes all the main ones, so they are effectively still supported, and there's no need to change the WDIO implementation.
(I do agree that this is confusing; I feel like the endpoints are universal enough to still be used in the main server)
> Also, XCTestScreenRecording/MediaProjectionRecording use different functionality under the hood, and have a different set of accepted input parameters, so they would not work as drop-in replacements.

Ok, I'll revert it, thanks

eglitise:
> This means the endpoint is passed directlty from the server to the driver?

Essentially, yes. The endpoints supported by the server are actually defined in the Appium base driver, which is a core dependency for all other drivers. As a result, drivers have access to the full set of these base driver endpoints, and can extend it with endpoints with their own. And when a session is started, the supported endpoints are retrieved from the driver directly.

wswebcreation:
> Yes, it says that the endpoints have been moved to drivers, which includes all the main ones, so they are effectively still supported, and there's no need to change the WDIO implementation.
> (I do agree that this is confusing; I feel like the endpoints are universal enough to still be used in the main server)
> Also, XCTestScreenRecording/MediaProjectionRecording use different functionality under the hood, and have a different set of accepted input parameters, so they would not work as drop-in replacements.

Thanks! That makes it more clear. I've reverted all commands we discussed and will push it to this branch. If you then don't have any remarks then I'll merge it

eglitise (approved):

eglitise:
@wswebcreation sorry, I had missed one more issue - the `queryAppState` endpoint only had its `GET` version removed in Appium 3, and its `POST` version still remains. WDIO only ever exposed the `POST` endpoint, so there's no need to add a wrapper for it, either.

wswebcreation:
ok, I'll remove that one, thanks

PR #15214, opened 2026-04-23
fix(webdriver): avoid enumerating string request bodies

### LINKED ISSUE #15213: [🐛 Bug]: browser.uploadFile() can crash with RangeError: Too many properties to enumerate on large request bodies

### WebdriverIO Version

9.27.0

### Node.js Version

v20.19.4

### Mode

WDIO Testrunner

```typescript
{
    browserName: 'chrome',
    browserVersion: 'latest',
    'goog:chromeOptions': {
      args: ['start-maximized']
    },
    'bstack:options': {
      os: 'OS X',
      osVersion: 'Sequoia',
      resolution: '1920x1080',
      projectName: 'Repro',
      buildName: `repro-${Date.now()}`,
      debug: true,
      seleniumVersion: '4.8.0',
      seleniumCdp: true
    }
  }
```

### What happened?

When a test uses `browser.uploadFile()` with a sufficiently large file, WebdriverIO can throw:
```
  RangeError: Too many properties to enumerate
      at Function.keys (<anonymous>)
```
  In my real-world case this happened during media upload flows in Selenium tests, but after tracing it down, the failure appears to come from the webdriver request layer itself.

  The problem seems to be:

  1. createOptions() stringifies the request body:
```
  requestOptions.body = JSON.stringify(this.body)
```
  2. _request() later does:
```
  if (fullRequestOptions.body && Object.keys(fullRequestOptions.body).length) {
      this.eventHandler.onLogData?.(fullRequestOptions.body)
  }
```
  At that point fullRequestOptions.body is a string, not an object. For large upload payloads, `Object.keys(hugeString)` throws `RangeError: Too many properties to enumerate.`

  I bisected this to:

  - likely regression commit: a4cc13e92ee1c8c9cf99a1914556352eb700b8c7
  - title: fix(webdriver): pass all options to fetch (#14799)

  That appears to be the point where `requestOptions.body` became a JSON string while `_request()` still treated it like an object.

browser.uploadFile() should not crash in the webdriver request logger path.

  If fullRequestOptions.body is a string, it should either:

  - be logged directly, or
  - be skipped for key-enumeration purposes

### How to reproduce the bug.

I don’t yet have a standalone public repo, but I do have a minimal internal repro and a proposed upstream fix.

  Minimal logic repro from the webdriver layer:

  // pseudo-repro of the failing shape
  const fullRequestOptions = {
    method: 'POST',
    body: JSON.stringify({ file: 'x'.repeat(10_000_000) })
  }

  Object.keys(fullRequestOptions.body).length
  // RangeError: Too many properties to enumerate

  In actual WDIO flow, this happens after browser.uploadFile() when the request body for /se/file is large.

  The issue can be reproduced in a real test by:

  1. starting a remote webdriver session
  2. calling browser.uploadFile() with a sufficiently large local file
  3. continuing through the file upload flow until WDIO logs the request body

  I also verified that patching packages/webdriver/src/request/request.ts to avoid Object.keys() on string bodies fixes the failing test.

  Proposed fix:
```
  function hasLoggableBodyContent (body: BodyInit | null | undefined): body is BodyInit {
      if (!body) {
          return false
      }

      if (typeof body === 'string') {
          return body.length > 0
      }

      if (body instanceof URLSearchParams || body instanceof FormData) {
          return Array.from(body.keys()).length > 0
      }

      if (body instanceof Blob || body instanceof ArrayBuffer || ArrayBuffer.isView(body)) {
          return (body.size ?? body.byteLength) > 0
      }

      return typeof body === 'object' && Object.keys(body).length > 0
  }
```
  and then:
```
  if (hasLoggableBodyContent(fullRequestOptions.body)) {
      this.eventHandler.onLogData?.(fullRequestOptions.body)
  }
```

### Relevant log output

```typescript
Application-level stack I captured:

  RangeError: Too many properties to enumerate
      at Function.keys (<anonymous>)
      at FetchRequest._request (.../packages/webdriver/src/request/request.ts:1968:43)
      at FetchRequest.makeRequest (.../packages/webdriver/src/request/request.ts:1893:17)

  The last successful command before the crash in my run was the file upload request:

  COMMAND: POST ... /se/file - {"file":"[base64] 21118844 bytes"}
  RESULT: {"file":"[base64] 21118844 bytes"}

  After patching the webdriver request logging logic locally, the previously failing upload-based test passed.
```

### Code of Conduct

kbrooks:
## Summary
- avoid calling `Object.keys()` on stringified request bodies in `FetchRequest._request()`
- keep request body logging for non-empty strings and structured body types
- add a regression test for string request bodies

## Problem
Large `/se/file` upload payloads can throw `RangeError: Too many properties to enumerate` because `_request()` calls `Object.keys(fullRequestOptions.body)` after `createOptions()` has already converted the body to a JSON string.

Fixes #15213.

### DIFF
--- a/packages/webdriver/src/request/request.ts
+++ b/packages/webdriver/src/request/request.ts
@@ -26,6 +26,26 @@ const DEFAULT_HEADERS = {
 
 const log = logger('webdriver')
 
+function hasLoggableBodyContent (body: BodyInit | null | undefined): body is BodyInit {
+    if (!body) {
+        return false
+    }
+
+    if (typeof body === 'string') {
+        return body.length > 0
+    }
+
+    if (body instanceof URLSearchParams || body instanceof FormData) {
+        return Array.from(body.keys()).length > 0
+    }
+
+    if (body instanceof Blob || body instanceof ArrayBuffer || ArrayBuffer.isView(body)) {
+        return body.size ?? body.byteLength > 0
+    }
+
+    return typeof body === 'object' && Object.keys(body).length > 0
+}
+
 export abstract class WebDriverRequest {
     protected abstract fetch(url: URL, opts: RequestInit): Promise<Response>
 
@@ -165,7 +185,7 @@ export abstract class WebDriverRequest {
     ): Promise<WebDriverResponse> {
         log.info(`[${fullRequestOptions.method}] ${(url as URL).href}`)
 
-        if (fullRequestOptions.body && Object.keys(fullRequestOptions.body).length) {
+        if (hasLoggableBodyContent(fullRequestOptions.body)) {
             this.eventHandler.onLogData?.(fullRequestOptions.body)
         }
 
--- a/packages/webdriver/tests/request.test.ts
+++ b/packages/webdriver/tests/request.test.ts
@@ -262,6 +262,19 @@ describe('webdriver request', () => {
             }))
         })
 
+        it('should log string request bodies without enumerating them', async () => {
+            const onLogData = vi.fn()
+            const req = new WebFetchRequest('POST', webdriverPath, {}, undefined, false, {
+                onLogData
+            })
+
+            const url = new URL('/session/foobar-123/element', baseUrl)
+            const opts = { body: JSON.stringify({ file: 'x'.repeat(10_000) }) }
+            await req['_request'](url, opts)
+
+            expect(onLogData).toHaveBeenCalledWith(opts.body)
+        })
+
         it('should short circuit if request throws a stale element exception', async () => {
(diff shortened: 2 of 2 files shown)


mccmrunal:
@kbrooks there are build failures please check

dprevost-LMI:
That sounds already fixed in https://github.com/webdriverio/webdriverio/pull/15221, right?
@greptileai, can you check if this PR is a duplicate of https://github.com/webdriverio/webdriverio/pull/15221?

christian-bromann:
Closing due to inactivity.

PR #15216, opened 2026-04-23
fix(browserstack-service): route browserstack_executor commands via HTTP/S in BiDi sessions

### LINKED ISSUE #15207: [🐛 Bug]: @wdio/browserstack-service testObservability crashes with BiDi — browserstack_executor is not valid JavaScript

### WebdriverIO Version

9.27.0

### Node.js Version

22.21.1

### Mode

WDIO Testrunner

```typescript
{
  "bstack:options": {
    os: "Windows",
    osVersion: "11",
    seleniumVersion: "4.40.0",
    seleniumBidi: "true",
    // ...other bstack options
  },
  browserName: "chrome",
  browserVersion: "latest",
}
```

### What happened?

This is a follow-up to #[14922](https://github.com/webdriverio/webdriverio/issues/14922), which was closed after a fix was merged in v9.22.0 but subsequently reverted in commit 6edc9c7 (Jan 17, 2026) because the fix caused issues elsewhere. The issue was never reopened, and the bug persists in v9.27.0.

When seleniumBidi: "true" is set in BrowserStack capabilities and testObservability is enabled (the default), @wdio/browserstack-service calls:

`browser.execute(`browserstack_executor: ${JSON.stringify({
    action: 'annotate',
    arguments: { data: `ObservabilitySync:${Date.now()}`, level: 'debug' }
})}`)`

With classic WebDriver, browser.execute() sends this string via HTTP to BrowserStack's hub, which intercepts browserstack_executor: as a special command before it ever reaches the browser.  

With BiDi enabled, browser.execute() uses script.evaluate over WebSocket directly to the browser, which tries to parse the string as JavaScript. browserstack_executor: {...} is not valid JavaScript, resulting in:

> SyntaxError: Unexpected token ':'
>     at new Function (<anonymous>)
>     at Browser.execute (node_modules/webdriverio/build/node.js:3685:53)
>     at async performO11ySync (node_modules/@wdio/browserstack-service/build/index.js:4088:9)

This affects both Chrome and Edge with BiDi enabled.

@wdio/browserstack-service should detect when a session is using BiDi and route browserstack_executor commands through the classic WebDriver HTTP endpoint instead of browser.execute(), or use an alternative mechanism that doesn't rely on executing non-JavaScript strings in the browser context.  The only current workaround is testObservability: false in the browserstack service config, which disables observability features entirely.

### How to reproduce the bug.

- Configure any test with seleniumBidi: "true" in bstack:options 
- Use @wdio/browserstack-service with default settings (testObservability is true by default) 
- Run any test spec

### Relevant log output

```typescript
[0-0] <anonymous_script>:3
[0-0] browserstack_executor: {"action":"annotate","arguments":{"data":"ObservabilitySync:1776843280286","level":"debug"}}
[0-0]                                 ^
[0-0] SyntaxError: Unexpected token ':'
[0-0]     at new Function (<anonymous>)
[0-0]     at Browser.execute (node_modules/webdriverio/build/node.js:3685:53)
[0-0]     at async performO11ySync (node_modules/@wdio/browserstack-service/build/index.js:4088:9)
```

### Code of Conduct

RohanImmanuel:
## Proposed changes

Fixes #15207

In BiDi sessions `browser.execute()` routes via `script.callFunction` over WebSocket directly to the browser by bypassing BrowserStack's hub. This causes `browserstack_executor:` commands to fail

This fix overwrites `execute` inside the `before()` hook of `@wdio/browserstack-service` to intercept any `browserstack_executor:` string and redirect it through `executeScript` which is HTTP/S only so it will be intercepted by the BrowserStack hub. All other scripts pass through to the original `execute` path unchanged. The fix covers both single browser and multiremote sessions and is entirely contained within `@wdio/browserstack-service`

## Types of changes

## Checklist

## Backport Request

- [ ] Back-ported PR at

## Further comments

The root cause is that `browser.execute()` in BiDi sessions calls `script.callFunction` over the WebSocket connection, which goes directly to the browser and bypasses BrowserStack's HTTP hub entirely. `executeScript` raw protocol command has no BiDi implementation and is absent from `webdriverBidi.ts` and `BidiHandler` so it always goes via `POST /session/{id}/execute/sync` over HTTP/S which BrowserStack's hub can intercept as expected

### DIFF
--- a/packages/wdio-browserstack-service/src/service.ts
+++ b/packages/wdio-browserstack-service/src/service.ts
@@ -225,26 +225,17 @@ export default class BrowserstackService implements Services.ServiceInstance {
         PerformanceTester.scenarioThatRan = this._scenariosThatRan
 
         if (this._browser) {
-            // redirect browserstack_executor calls from BiDi(execute) WebSocket to HTTP/S(executeScript)
-            if (this._browser.isMultiremote) {
-                const multiRemoteBrowser = this._browser as unknown as WebdriverIO.MultiRemoteBrowser
-                Object.keys(this._caps).forEach((browserName) => {
-                    const instance = multiRemoteBrowser.getInstance(browserName)
-                    instance.overwriteCommand('execute', async (originalExecute, script, ...args) => {
-                        if (typeof script === 'string' && script.startsWith('browserstack_executor:')) {
-                            return instance.executeScript(script, [])
-                        }
-                        return originalExecute(script, ...args)
+            try {
+                if (this._browser.isMultiremote) {
+                    const multiRemoteBrowser = this._browser as unknown as WebdriverIO.MultiRemoteBrowser
+                    Object.keys(this._caps).forEach((browserName) => {
+                        this._routeBidiExecutorToHttp(multiRemoteBrowser.getInstance(browserName))
                     })
-                })
-            } else {
-                const browser = this._browser as WebdriverIO.Browser
-                browser.overwriteCommand('execute', async (originalExecute, script, ...args) => {
-                    if (typeof script === 'string' && script.startsWith('browserstack_executor:')) {
-                        return browser.executeScript(script, [])
-                    }
-                    return originalExecute(script, ...args)
-                })
+                } else {
+                    this._routeBidiExecutorToHttp(this._browser as WebdriverIO.Browser)
+                }
+            } catch (err) {
+                BStackLogger.warn(`Failed to patch execute for BiDi browserstack_executor routing; executor commands may not work in BiDi sessions: ${err}`)
             }
 
             try {
@@ -737,6 +728,19 @@ export default class BrowserstackService implements Services.ServiceInstance {
                 : `Update job with sessionId ${sessionId}`
             )
             return this._update(sessionId, requestBody)
+        })
+    }
+
+    _routeBidiExecutorToHttp (browser: WebdriverIO.Browser) {
+        if (!browser.isBidi) {
+            return
+        }
+
+        browser.overwriteCommand('execute', async (originalExecute, script, ...args) => {
+            if (typeof script === 'string' && script.startsWith('browserstack_executor:')) {
+                return browser.executeScript(script, args)
+            }
+            return originalExecute(script, ...args)
         })
     }
 
--- a/packages/wdio-browserstack-service/src/service.ts
+++ b/packages/wdio-browserstack-service/src/service.ts
@@ -225,6 +225,28 @@ export default class BrowserstackService implements Services.ServiceInstance {
         PerformanceTester.scenarioThatRan = this._scenariosThatRan
 
         if (this._browser) {
+            // redirect browserstack_executor calls from BiDi(execute) WebSocket to HTTP/S(executeScript)
+            if (this._browser.isMultiremote) {
+                const multiRemoteBrowser = this._browser as unknown as WebdriverIO.MultiRemoteBrowser
+                Object.keys(this._caps).forEach((browserName) => {
+                    const instance = multiRemoteBrowser.getInstance(browserName)
+                    instance.overwriteCommand('execute', async (originalExecute, script, ...args) => {
+                        if (typeof script === 'string' && script.startsWith('browserstack_executor:')) {
+                            return instance.executeScript(script, [])
+                        }
+                        return originalExecute(script, ...args)
+                    })
+                })
+            } else {
+                const browser = this._browser as WebdriverIO.Browser
+                browser.overwriteCommand('execute', async (originalExecute, script, ...args) => {
+                    if (typeof script === 'string' && script.startsWith('browserstack_executor:')) {
+                        return browser.executeScript(script, [])
+                    }
+                    return originalExecute(script, ...args)
+                })
+            }
+
             try {
                 const sessionId = this._browser.sessionId
 
--- a/packages/wdio-browserstack-service/tests/service.test.ts
+++ b/packages/wdio-browserstack-service/tests/service.test.ts
@@ -112,6 +112,7 @@ beforeEach(() => {
     browser = {
         execute: vi.fn(),
         executeScript: vi.fn(),
+        overwriteCommand: vi.fn(),
         on: vi.fn(),
         sessionId: sessionId,
         config: {},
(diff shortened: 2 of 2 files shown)


packages/wdio-browserstack-service/src/service.ts
```diff
@@ -209,6 +209,28 @@ export default class BrowserstackService implements Services.ServiceInstance {
...
+                })
+            } else {
+                const browser = this._browser as WebdriverIO.Browser
+                browser.overwriteCommand('execute', async (originalExecute, script, ...args) => {
+                    if (typeof script === 'string' && script.startsWith('browserstack_executor:')) {
+                        return browser.executeScript(script, [])
+                    }
+                    return originalExecute(script, ...args)
+                })
+            }
```
amaanbs: ## 🔴 Critical — [GRACEFUL DEGRADATION] `overwriteCommand` block has no error boundary

### Problem

`overwriteCommand` can throw if the browser instance doesn't expose the method — older WDIO versions, custom browser implementations, or edge-case multiremote instances that don't conform to the standard API. With no `try/catch` around this block, any such failure propagates through `before()` and aborts session setup entirely, breaking **all user tests** with no useful diagnostic.

The SDK's own convention is explicit on this: instrumentation failures must not block test execution. If patching fails, the correct behavior is to log a warning and continue — not to throw.

### Suggested Fix

Wrap the entire block (both the `isMultiremote` branch and the single-browser branch) in a `try/catch` that logs and continues:

```typescript
try {
    if (this._browser.isMultiremote) {
        const multiRemoteBrowser = this._browser as unknown as WebdriverIO.MultiRemoteBrowser
        Object.keys(this._caps).forEach((browserName) => {
            const instance = multiRemoteBrowser.getInstance(browserName)
            instance.overwriteCommand('execute', async (originalExecute, script, ...args) => {
                if (typeof script === 'string' && script.startsWith('browserstack_executor:')) {
                    return instance.executeScript(script, args)
                }
                return originalExecute(script, ...args)
            })
        })
    } else {
        const browser = this._browser as WebdriverIO.Browser
        browser.overwriteCommand('execute', async (originalExecute, script, ...args) => {
            if (typeof script === 'string' && script.startsWith('browserstack_executor:')) {
                return browser.executeScript(script, args)
            }
            return originalExecute(script, ...args)
        })
    }
} catch (err) {
    log.warn('BrowserstackService: failed to patch execute for BiDi routing, executor commands may not work in BiDi sessions:', err)
}
```

**Confidence:** 🟢 Matches SDK anti-pattern: instrumentation must not throw into the test lifecycle.

amaanbs (commented):
## Automated SDK PR Review

**Verdict:** ⚠️ Fix 4 issues before merging

**Summary:** 1 critical · 3 warnings · 1 suggestion across 2 files reviewed.

The routing mechanism itself is technically sound — `executeScript` has no BiDi implementation so the HTTP/S fallthrough is guaranteed. The issues are around hardening (missing error boundary), scoping (intercept runs on every session not just BiDi), test completeness (spread not exercised), and an implicit protocol assumption (`args: []`). All are addressable in a follow-up commit on this branch.

See inline comments below for full **Problem** and **Suggested Fix** detail on each finding.

_Generated by Automated SDK PR review._

packages/wdio-browserstack-service/src/service.ts
```diff
@@ -209,6 +209,28 @@ export default class BrowserstackService implements Services.ServiceInstance {
         PerformanceTester.scenarioThatRan = this._scenariosThatRan
 
         if (this._browser) {
+            // redirect browserstack_executor calls from BiDi(execute) WebSocket to HTTP/S(executeScript)
+            if (this._browser.isMultiremote) {
```
amaanbs: ## ⚠️ Warning — [FEATURE CORRECTNESS] `overwriteCommand` applied unconditionally — no BiDi guard

### Problem

The intercept registers on **every session**, including non-BiDi sessions where `browser.execute()` already routes over HTTP/S correctly. In those sessions the intercept is harmless for non-executor scripts (falls through to `originalExecute`), but for `browserstack_executor:` strings it now routes through `executeScript` instead of `execute` — a different command path.

This is functionally equivalent today, but:
- It adds an unnecessary wrapper on the ~95% of sessions that don't need it
- It makes the fix invisible to future maintainers (why is `executeScript` being called here when the session isn't BiDi?)
- It widens the blast radius if `executeScript` ever behaves differently from `execute` in a specific context

### Suggested Fix

Gate the entire `overwriteCommand` block on a BiDi capability check:

```typescript
const isBidi = !!(
    (this._caps as WebDriver.Capabilities)['webSocketUrl'] ||
    (this._caps as WebDriver.Capabilities)['goog:webSocketUrl']
)
if (isBidi) {
    // overwriteCommand block here
}
```

This scopes the fix to exactly the sessions it's needed for and makes the intent self-documenting.

**Confidence:** 🟢 Objectively verifiable: the capability is set only for BiDi sessions; the intercept is only needed there.

packages/wdio-browserstack-service/src/service.ts
```diff
@@ -209,6 +209,28 @@ export default class BrowserstackService implements Services.ServiceInstance {
...
+                            return instance.executeScript(script, [])
+                        }
+                        return originalExecute(script, ...args)
+                    })
+                })
+            } else {
+                const browser = this._browser as WebdriverIO.Browser
+                browser.overwriteCommand('execute', async (originalExecute, script, ...args) => {
+                    if (typeof script === 'string' && script.startsWith('browserstack_executor:')) {
+                        return browser.executeScript(script, [])
```
amaanbs: ## ⚠️ Warning — [PROTOCOL ADHERENCE] `executeScript` called with hardcoded empty `args` array

### Problem

Both `instance.executeScript(script, [])` (line 219) and `browser.executeScript(script, [])` (line 228) hardcode `args: []`, dropping whatever arguments the original `execute()` caller passed alongside the `browserstack_executor:` script.

The WebDriver wire protocol body for `POST /session/{id}/execute/sync` is `{"script": ..., "args": [...]}`. All current `browserstack_executor:` convention encodes its payload entirely in the JSON string — not as WebDriver-level args — so this is safe today. But it's an implicit contract assumption rather than an explicit design choice, and it silently drops args if any caller ever passes them.

### Suggested Fix

Pass the original `args` through rather than hardcoding `[]`:

```typescript
// line 219 — multiremote
return instance.executeScript(script, args)

// line 228 — single browser
return browser.executeScript(script, args)
```

If the current contract is intentionally "executor commands never use WebDriver args", add an inline comment to make it explicit rather than relying on the empty array as silent documentation.

**Confidence:** 🟢 Objectively verifiable: `args` are in scope but not forwarded.

packages/wdio-browserstack-service/tests/service.test.ts
```diff
@@ -578,6 +579,46 @@ describe('before', () => {
...
+
+        const overwrite = vi.mocked(browser.overwriteCommand).mock.calls[0][1] as Function
+        const originalExecute = vi.fn()
+
+        await overwrite(originalExecute, 'browserstack_executor: {"action":"annotate"}')
+        expect(browser.executeScript).toHaveBeenCalledWith('browserstack_executor: {"action":"annotate"}', [])
+        expect(originalExecute).not.toHaveBeenCalled()
+
+        await overwrite(originalExecute, 'return document.title')
+        expect(originalExecute).toHaveBeenCalledWith('return document.title')
```
amaanbs, line 634: ## ⚠️ Warning — [FEATURE CORRECTNESS] Single-browser test doesn't exercise `...args` spread

### Problem

The non-executor passthrough assertion:

```typescript
await overwrite(originalExecute, 'return document.title')
expect(originalExecute).toHaveBeenCalledWith('return document.title')
```

passes because `args` is empty here — `originalExecute(script, ...args)` with `args = []` is identical to `originalExecute(script)`. If the spread were accidentally dropped in the implementation (i.e., `return originalExecute(script)` instead of `return originalExecute(script, ...args)`), this test would **not** catch the regression.

### Suggested Fix

Add a variant that passes extra arguments to verify the spread is preserved:

```typescript
// existing assertions above...

// also verify that extra args are forwarded
const extraArg = { key: 'value' }
await overwrite(originalExecute, 'return document.title', extraArg)
expect(originalExecute).toHaveBeenCalledWith('return document.title', extraArg)
```

The same gap exists in the multiremote test — add an equivalent check there too.

**Confidence:** 🟢 Objectively verifiable: spread is not exercised by the current call; a spread-drop bug passes undetected.

packages/wdio-browserstack-service/src/service.ts
```diff
@@ -209,6 +209,28 @@ export default class BrowserstackService implements Services.ServiceInstance {
...
+                    instance.overwriteCommand('execute', async (originalExecute, script, ...args) => {
+                        if (typeof script === 'string' && script.startsWith('browserstack_executor:')) {
+                            return instance.executeScript(script, [])
+                        }
+                        return originalExecute(script, ...args)
+                    })
+                })
+            } else {
+                const browser = this._browser as WebdriverIO.Browser
+                browser.overwriteCommand('execute', async (originalExecute, script, ...args) => {
```
amaanbs: ## 💡 Suggestion — [ROBUSTNESS] `overwriteCommand` double-wrap guard

### Suggestion

`overwriteCommand` accumulates wrappers if `before()` is called more than once on the same browser instance. WDIO normally prevents this at the service level, but multiremote retry edge cases could trigger it. Adding an idempotency flag costs nothing and prevents executor commands from being double-intercepted:

```typescript
// multiremote variant
const instance = multiRemoteBrowser.getInstance(browserName)
if (!(instance as any)._bstackExecutorPatched) {
    instance.overwriteCommand('execute', async (originalExecute, script, ...args) => { ... })
    ;(instance as any)._bstackExecutorPatched = true
}

// single-browser variant
if (!(browser as any)._bstackExecutorPatched) {
    browser.overwriteCommand('execute', async (originalExecute, script, ...args) => { ... })
    ;(browser as any)._bstackExecutorPatched = true
}
```

amaanbs:
@christian-bromann Reviewed the PR ^ & shared my findings.

RohanImmanuel:
thanks for the review @amaanbs

pushed the follow-up commit and rebased

- added try/catch and fail logs using BStackLogger.warn
- moved the duplicated overwrite logic into a single _routeBidiExecutorToHttp helper
- agreed scoping should not run on every session. used the browser.isBidi flag instead of checking the caps for this
- now forwards the original args to executeScript
- added assertions in both the single-browser and multiremote tests for extra args pass through ...args spread
- fixed the multiremote test passing capabilities in the wrong constructor argument.

left out the double wrap guard for now as before() runs only once per session and adding it would just recheck the prefix with no effect, so the flag did not seem worth the extra state
happy to add it if you think the multiremote retry case needs it

Bhargavi-BS (approved):

dprevost-LMI:
`wdio-browserstack-service` unfortunately moved to https://github.com/browserstack/wdio-browserstack-service. I suggest moving the PR there too

dprevost-LMI:
`@wdio/browserstack-service` moved to https://github.com/browserstack/wdio-browserstack-service. I also suggest moving this PR there.

osho-20 (approved):

dprevost-LMI:
Closing since ported there: https://github.com/browserstack/wdio-browserstack-service/commit/e6dc8d2dba1902ea3f25ad15da6f2e327dcb16ff

Issue #1755, opened 2016-12-07
<Span> Element is not clickable on mobile web.


varap:
## The problem
<button class="button-email btn btn-default" type="button"><span>Sign up with Email</span></button>

* WebdriverIO version: 3.4.0
* Node.js version: 6.7
* [Standalone mode or wdio testrunner](http://webdriver.io/guide/getstarted/modes.html):test runner
* if wdio testrunner, running synchronous or asynchronous tests: synchronous
* Additional wdio packages used (if applicable):

## Details

Describe in more detail the problem you have been experiencing, if necessary.

the react JS form having the signup Button with SPAN element is not working when we try to click through command(.click()). interestingly, on the web, this is not working for chrome Versions after 48 and on the mobile web checked with iphone 6s model . 

im using browserstack for the configuartion.

## Code To Reproduce Issue [ Good To Have ]

let signupModalUserName = '.login-modal-container input[type="email"]';
let signupModalPassword ='.login-modal-container input[type="password"]';
let signupModalSubmitButton ='.login-modal-container.col-md-5.col-sm-12.col-xs-12 > div > div > div.row > div > button > span';

                .setValue(signupModalUserName, validEmail)
                .setValue(signupModalPassword, password)
                .click(submitButton)

The most confused part is it works for chrome version 48 and firefox version 42 . mobile web not working . pls help me in the analysis.
Please remember that, with sample code; it's easier to reproduce bug and much faster to fix it.

christian-bromann:
@varap can you provide logs please?

varap:
@christian-bromann 
i'm using browserstack to do the runs and so i got the raw logs from their run  , let me know if you need more data on it .

2016-12-8 2:31:56:245 REQUEST [2016-12-8 2:31:56:245] POST /session/7e17a5b824a10acba1b371047ecd93cd25f29b67/element/5121/value {"value":["v","a","r","a","+","4","8","0","9","0","d","9","0","-","b","c","e","e","-","1","1","e","6","-","8","9","0","b","-","e","d","b","0","3","d","8","1","5","9","9","7","@","s","h","o","p","b","a","c","k",".","c","o","m"]}
2016-12-8 2:31:56:503 DEBUG https://s3.amazonaws.com/testautomation/7e17a5b824a10acba1b371047ecd93cd25f29b67/screenshot-5535cf3692.jpeg
[...]

PR #910, opened 2015-11-30
Update elements.js

blindedByCode:
Adding additional information to the command, which eliminates the need for the user to print the entire array and understanding its structure before using it. Makes the command easier to use and implement.

PR #15219, opened 2026-04-27
fix:handles binary file busy error at execution(spawn)

shivam5643:
## Proposed changes

Fixes a binary file busy (ETXTBSY/EBUSY) error that occurred when the BrowserStack CLI binary was being executed while a new version was being written to the same file path during spawn.

## Types of changes

## Checklist

## Backport Request

## Further comments

### DIFF
--- a/packages/wdio-browserstack-service/src/cli/cliUtils.ts
+++ b/packages/wdio-browserstack-service/src/cli/cliUtils.ts
@@ -643,8 +643,6 @@ export class CLIUtils {
             const downloadedFileStream = fs.createWriteStream(zipFilePath)
 
             return new Promise<string | null>((resolve, reject) => {
-                const binaryName = null
-
                 const processDownload = async () => {
                     const abortController = new AbortController()
                     const timeout = setTimeout(
@@ -683,7 +681,6 @@ export class CLIUtils {
                         // Set up the downloadFileStream handler before pipeline
                         CLIUtils.downloadFileStream(
                             downloadedFileStream,
-                            binaryName,
                             zipFilePath,
                             cliDir,
                             (result: string) => {
@@ -721,7 +718,6 @@ export class CLIUtils {
 
     static downloadFileStream(
         downloadedFileStream: fs.WriteStream,
-        binaryName: string | null,
         zipFilePath: string,
         cliDir: string,
         resolve: (path: string) => void,
@@ -736,18 +732,16 @@ export class CLIUtils {
                 const zipfile = await yauzlOpenPromise(zipFilePath, {
                     lazyEntries: true,
                 })
+                let resolvedBinaryPath: string | null = null
+
                 zipfile.readEntry()
                 zipfile.on('entry', async (entry) => {
                     if (/\/$/.test(entry.fileName)) {
-                        // Directory entry — skip, no temp file needed
                         zipfile.readEntry()
                         return
                     }
 
                     const isBinaryEntry = path.basename(entry.fileName).startsWith('binary-')
-                    if (!binaryName && isBinaryEntry) {
-                        binaryName = entry.fileName
-                    }
 
                     if (!isBinaryEntry) {
                         const directStream = fs.createWriteStream(
@@ -770,40 +764,43 @@ export class CLIUtils {
                         return
                     }
 
-                    // Binary entry: extract to PID-scoped temp file, then atomic rename.
-                    // This prevents ETXTBSY/EBUSY(text file busy): the file being executed is never the file being written.
+                    // Binary entry: extract to PID-scoped temp file, chmod, atomic rename inline.
+                    // Prevents ETXTBSY/EBUSY: the file being executed is never the file being written.
                     const finalPath = path.join(cliDir, entry.fileName)
                     const tempPath = path.join(cliDir, `${entry.fileName}.tmp.${process.pid}`)
 
-                    let writeStream: fs.WriteStream
-                    try {
-                        writeStream = fs.createWriteStream(tempPath)
-                    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-                    } catch (createErr: any) {
-                        if (BINARY_BUSY_ERROR_CODES.includes(createErr.code) && fs.existsSync(finalPath)) {
-                            logger.warn(`Temp file busy, falling back to existing binary: ${entry.fileName}`)
-                            fsp.unlink(zipFilePath).catch(() => {})
-                            zipfile.close()
-                            resolve(finalPath)
-                            return
-                        }
-                        zipfile.close()
-                        reject(createErr)
-                        return
-                    }
+                    const writeStream = fs.createWriteStream(tempPath)
 
-                    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-                    writeStream.on('error', (writeErr: any) => {
-                        if (BINARY_BUSY_ERROR_CODES.includes(writeErr.code) && fs.existsSync(finalPath)) {
-                            logger.warn(`Temp write busy, falling back to existing binary: ${writeErr.message}`)
-                            writeStream.close()
-                            fsp.unlink(tempPath).catch(() => {})
-                            fsp.unlink(zipFilePath).catch(() => {})
+                    writeStream.on('error', (writeErr) => {
+                        fsp.unlink(tempPath).catch(() => {})
+                        zipfile.close()
+                        reject(writeErr as Error)
+                    })
+
+                    // 'finish' fires on successful flush only; 'error' path won't reach here.
+                    writeStream.on('finish', async () => {
+                        try {
+                            await fsp.chmod(tempPath, '0755')
+                            try {
+                                await fsp.rename(tempPath, finalPath)
+                            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+                            } catch (renameErr: any) {
+                                // Narrow fallback to cross-device (EXDEV) only
+                                if (renameErr.code !== 'EXDEV') {
+                                    throw renameErr
+                                }
+                                logger.warn(`Atomic rename failed (cross-device), falling back to copy: ${renameErr.message}`)
+                                await fsp.copyFile(tempPath, finalPath)
+                                await fsp.unlink(tempPath).catch(() => {})
+                            }
+                            if (!resolvedBinaryPath) {
+                                resolvedBinaryPath = finalPath
+                            }
+                            zipfile.readEntry()
+                        } catch (err) {
+                            await fsp.unlink(tempPath).catch(() => {})
                             zipfile.close()
-                            resolve(finalPath)
-                        } else {
-                            zipfile.close()
-                            reject(writeErr)
+                            reject(err as Error)
                         }
                     })
 
@@ -814,7 +811,6 @@ export class CLIUtils {
                         const readStream = await openReadStreamPromise(entry)
                         readStream.on('end', function () {
                             writeStream.end()
-                            writeStream.on('close', () => zipfile.readEntry())
                         })
                         readStream.pipe(writeStream)
                     } catch (zipErr) {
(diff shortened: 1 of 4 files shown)


xxshubhamxx:
## Code Review — WDIO BrowserStack Service PR #15219: Binary File Busy Fix (SDK-5786)

This is a parallel fix to the C# SDK change — same root cause, same atomic-rename approach. The refactor to `_spawnAndAwaitReady` and the `settled` flag are genuine improvements. Two warnings before merge.

---

### ⚠️ Warning 1: `firstEntryName` fallback in `'end'` handler will fail with ENOENT if triggered — `cliUtils.ts`

In `zipfile.once('end', ...)`:
```typescript
if (!binaryName) {
    if (firstEntryName) {
        binaryName = firstEntryName  // ← fallback to first non-dir entry
    } else {
        reject(new Error('No binary entry found in zip'))
        return
    }
}

const tempPath = path.join(cliDir, `${binaryName}.tmp.${process.pid}`)
const finalPath = path.join(cliDir, binaryName)

fsp.chmod(tempPath, '0755')   // ← will fail: tempPath doesn't exist
    .then(() => fsp.rename(tempPath, finalPath))
    ...
```

The issue: non-binary entries (those not matching `binary-*`) are written directly to their final path — no temp file is created for them. But the `'end'` handler unconditionally constructs `tempPath = ${binaryName}.tmp.${pid}` and tries to `chmod` it.

If `firstEntryName` is a non-binary entry (e.g., a config file), `tempPath` was never created. `fsp.chmod` fails with `ENOENT`, the `.catch` branch checks for `renameErr.code === 'EXDEV'` (doesn't match), and the promise rejects with `ENOENT: no such file or directory` — a confusing error that hides the real problem ("no `binary-*` entry in zip").

The `'No binary entry found in zip'` error path lower in the same block is the correct signal, but it's bypassed by the `firstEntryName` fallback.

**Fix:** Either document that this fallback is intentionally unreachable (the zip always contains `binary-*`), or restructure:
```typescript
if (!binaryName) {
    zipfile.close()
    reject(new Error('No binary entry found in zip'))
    return
}
```
...and remove `firstEntryName` tracking if the `binary-` prefix convention is reliable.

---

### ⚠️ Warning 2: `_spawnAndAwaitReady` has no timeout — `index.ts`

```typescript
private _spawnAndAwaitReady(cmd: Array<string>): Promise<void> {
    return new Promise<void>((resolve, reject) => {
        ...
        this.process.on('close', (code: number) => {
            if (code !== 0) {
                settle(reject, new Error(`CLI process exited with code ${code}`))
            }
            // code === 0 with no "ready" → settle() is never called → promise hangs
        })
    })
}
```

If the binary exits with code 0 before emitting `"ready"` (unexpected clean shutdown), `settle()` is never called. The retry loop uses `await this._spawnAndAwaitReady(cmd)`, so a hang here blocks all subsequent retry attempts — the process never proceeds to attempt 2 or 3.

This is a pre-existing issue from the original `start()` (the old code had the same pattern), but extracting it into `_spawnAndAwaitReady` makes it more opaque: the retry loop appears to handle all failure modes when it actually only handles rejection, not hangs.

**Fix:** Add a timeout inside `_spawnAndAwaitReady`:
```typescript
const timeout = setTimeout(() => {
    settle(reject, new Error('CLI process did not emit ready within timeout'))
}, CLI_SPAWN_TIMEOUT_MS)

this.process.on('close', (code: number) => {
    clearTimeout(timeout)
    if (code !== 0) {
        settle(reject, new Error(`CLI process exited with code ${code}`))
    } else {
        settle(reject, new Error('CLI process exited cleanly without emitting ready'))
    }
})
```

---

### 💡 Suggestion: Document the `binary-` prefix convention — `cliUtils.ts`

```typescript
const isBinaryEntry = path.basename(entry.fileName).startsWith('binary-')
```

This embeds an implicit contract about the zip's naming convention. A future maintainer changing the binary naming (e.g., `cli-linux-x64`) would silently break the atomic-rename path — all entries would take the direct write path and ETXTBSY would resurface.

Add a one-liner:
```typescript
// BrowserStack binary zips name the executable 'binary-{platform}' (e.g. binary-linuxstatic-x64, binary-macos-arm64).
// Only binary entries need atomic rename; other entries (configs, etc.) can be written directly.
const isBinaryEntry = path.basename(entry.fileName).startsWith('binary-')
```

---

### ✅ What's Good

- **`settled` flag in `_spawnAndAwaitReady`** is a real improvement over the original code — prevents double-resolve/reject from concurrent stdout data + process close events.
- **`BINARY_BUSY_ERROR_CODES` constant** exported from `constants.ts` and shared between `cliUtils.ts` and `index.ts` — no string duplication.
- **Binary vs non-binary split in `extractZip`** is correct: only executables need atomic rename, config files can be written directly.
- **`EXDEV` narrowing** in the rename fallback is precise — cross-device rename is the only legitimate case for falling back to `copyFile`. Broad `catch` would hide real errors.
- **Retry cleanup** calls `removeAllListeners()` on `stdout`/`stderr` before nulling `this.process` — prevents stale event handlers from firing on a dead reference.
- **`null` check for `SDK_CLI_BIN_PATH`** before spawn turns a confusing `TypeError: Cannot read properties of undefined` into a clear error message.
- **`macOS` short-circuit in `isBinaryBusy`** is correct — macOS doesn't enforce ETXTBSY on write-while-executing the same way Linux does.

packages/wdio-browserstack-service/src/cli/cliUtils.ts
```diff
@@ -705,35 +738,89 @@ export class CLIUtils {
...
+                    const tempPath = path.join(cliDir, `${entry.fileName}.tmp.${process.pid}`)
 
-                        if (entry.fileName === binaryName) {
+                    let writeStream: fs.WriteStream
+                    try {
+                        writeStream = fs.createWriteStream(tempPath)
+                    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+                    } catch (createErr: any) {
+                        if (BINARY_BUSY_ERROR_CODES.includes(createErr.code) && fs.existsSync(finalPath)) {
+                            logger.warn(`Temp file busy, falling back to existing binary: ${entry.fileName}`)
```
kamal-kaur04: The "fallback to existing binary on temp-write busy" branches (both the `createWriteStream` catch above and this `writeStream.on('error')` handler) are effectively dead code. The temp path is `${entry.fileName}.tmp.${process.pid}` — unique per process, never executed, never opened by anyone else. ETXTBSY can't fire (not an exec path); EBUSY needs another open handle.

The actual race window is the `fsp.rename` in the `once('end')` handler, not the temp write. C# in the sibling PR #789 correctly omits this branch.

**Fix:** drop both temp-write busy fallbacks; keep busy handling on `isBinaryBusy` and the version-check path.

packages/wdio-browserstack-service/src/cli/cliUtils.ts
```diff
@@ -743,18 +830,39 @@ export class CLIUtils {
...
-                            `Failed to delete zip file: ${zipFilePath}`,
-                        )
+                        logger.warn(`Failed to delete zip file: ${zipFilePath}`)
                     })
-                    fsp.chmod(`${cliDir}/${binaryName}`, '0755')
+
+                    // The BrowserStack zip always contains a binary-* entry by convention,
+                    // so binaryName is non-null here.
+                    const tempPath = path.join(cliDir, `${binaryName}.tmp.${process.pid}`)
+                    const finalPath = path.join(cliDir, binaryName!)
```
kamal-kaur04: The `binaryName!` non-null assertion + chmod-then-rename only runs once at the end of all entries. Two edge cases:

1. If the zip ever contains 2+ `binary-*` entries (sidecar binaries, etc.), `binaryName` is set on the first only — subsequent ones are written to `.tmp.<pid>` and orphaned (no rename, no cleanup).
2. If the zip contains 0 `binary-*` entries (broken upstream contract), `binaryName!` is null at runtime. `${null}.tmp.${pid}` → `fsp.chmod` fails with `ENOENT`, falling through to the non-`EXDEV` reject branch with a confusing error.

Java/Python/C# rename inline per entry, eliminating both edge cases.

**Fix:** either (a) rename inline within the binary-entry block (matches the other 3 SDKs), or (b) explicitly fail with a clear error at the start of the `once('end')` handler when `binaryName` is null.

kamal-kaur04 (commented):
SDK-5786 review (wdio side of a 4-PR series spanning Java/Python/C#/wdio). The atomic-rename + `isBinaryBusy` probe is the right core fix, and wdio is one layer ahead of its siblings — `acquireLock` and `cleanupTemporaryDownloads` (from #15066) supersede most of the additional defenses. Inline comments call out: (1) the temp-write busy fallback in the entry handler is dead code (temp path is `${entry.fileName}.tmp.${process.pid}`, can't be busy); (2) the `binaryName!` non-null assertion + single chmod-then-rename in the `once('end')` handler is fragile — fails confusingly with 0 or 2+ binary-* entries.

Verdict: APPROVE WITH SUGGESTIONS.

packages/wdio-browserstack-service/src/cli/cliUtils.ts
shivam5643: yes, the temp file write will never goes to race, as its unique per process.Have Fixed that.

packages/wdio-browserstack-service/src/cli/cliUtils.ts
shivam5643: Refactored to per-entry chmod+rename inline.

xxshubhamxx (approved):

packages/wdio-browserstack-service/src/cli/cliUtils.ts
```diff
@@ -703,37 +732,91 @@ export class CLIUtils {
...
+                            zipfile.close()
                             reject(zipErr as Error)
                         }
+                        return
+                    }
+
+                    // Binary entry: extract to PID-scoped temp file, chmod, atomic rename inline.
+                    // Prevents ETXTBSY/EBUSY: the file being executed is never the file being written.
+                    const finalPath = path.join(cliDir, entry.fileName)
+                    const tempPath = path.join(cliDir, `${entry.fileName}.tmp.${process.pid}`)
```
kamal-kaur04, line 774: Follow-up on the cross-cutting item flagged in the review-level summary — the orphan `.tmp.<pid>` cleanup. Could you address this in this PR rather than deferring it?

**Note:** The existing `cleanupTemporaryDownloads` (cliUtils.ts:555) only sweeps `downloaded_file_*.zip` files (matched by `CLI_DOWNLOAD_TMP_PREFIX` + `CLI_DOWNLOAD_TMP_SUFFIX`). It does **not** match the new `${entry.fileName}.tmp.${process.pid}` extracted binaries this PR added — those files have a completely different naming scheme.

**Concern:** When a worker crashes mid-extract (panic, OOM, `SIGKILL`), the `.tmp.<pid>` it was writing remains in `cliDir` forever. Over a long-running CI fleet running thousands of test sessions, these accumulate — each is binary-sized (tens of MB).

**Suggested fix:** extend `cleanupTemporaryDownloads` (or add a parallel helper) to also match `*.tmp.*`:

```ts
const cleanupStaleTempFiles = (maxAgeMs = CLI_LOCK_TIMEOUT_MS) => {
    try {
        const now = Date.now()
        for (const entry of fs.readdirSync(cliDir)) {
            if (!/\.tmp\.\d+$/.test(entry)) {continue}
            const filePath = path.join(cliDir, entry)
            try {
                const stats = fs.statSync(filePath)
                if (now - stats.mtimeMs < maxAgeMs) {continue}
                fs.unlinkSync(filePath)
            } catch (err) {
                logger.debug(`Failed to delete stale temp file ${filePath}: ${util.format(err)}`)
            }
        }
    } catch (err) {
        logger.debug(`Failed to scan stale temp files: ${util.format(err)}`)
    }
}
```

Please coordinate with the sibling PRs (#1755 Java, #910 Python, #789 C#) — none of them have this cleanup either, and the same pattern should land in all four. Worth aligning so a future port doesn't have to chase 4 separate semantics.

shivam5643: cleanup is already there, just added .tmp.<pid> there as well.

kamal-kaur04 (approved):
LGTM — all review feedback addressed across two iterations.

**Round 1 (2 findings):** dead-code temp-write fallback removed; `binaryName!` non-null assertion replaced with a clean per-entry inline `chmod`+atomic-rename refactor (matches the structure in sibling Java/Python/C# PRs).

**Round 2 (1 cross-cutting):** existing `cleanupTemporaryDownloads` extended (DRY) to also match `*.tmp.<pid>` files — single regex `\.tmp\.\d+$`, reuses `CLI_LOCK_TIMEOUT_MS`. Cleaner than the parallel-helper approach taken by the sibling PRs #1755/#910/#789.

xxshubhamxx (approved):

kamal-kaur04:
Closing in favour of https://github.com/webdriverio/webdriverio/pull/15246

PR #15241, opened 2026-05-11
feat(webdriverio): extend dialog commands to support mobile permission dialogs (iOS/Android)

christian-bromann:
This PR extends the WebdriverIO dialog commands to support mobile permission dialogs on iOS and Android, in addition to the existing browser dialog support.

### Core Implementation (`packages/webdriverio/src/session/dialog.ts`)
- Added `SPRINGBOARD_BUNDLE_ID` constant for iOS system UI access
- Updated `Dialog.accept()` to handle mobile dialogs via new `#handleMobileDialog()` private method
- Updated `Dialog.dismiss()` to handle mobile dialogs
- Implemented iOS-specific flow: get current app bundle ID → activate SpringBoard → accept/dismiss alert → reactivate original app
- Implemented Android flow: directly accept/dismiss alert using WebDriver protocol
- Silently handles "no such alert" errors (permission already granted or not triggered)

### Documentation Updates
- Updated `accept.ts` and `dismiss.ts` with mobile examples and `:::info` blocks explaining dual browser/mobile support
- Fixed file path reference in comment (was `src/dialog.ts`, now `src/session/dialog.ts`)

### Tests (`packages/webdriverio/tests/session/dialog.test.ts`)
- Added `Dialog - Browser` test suite for BiDi dialog handling
- Added `Dialog - iOS Mobile` test suite covering accept, dismiss, and prompt text scenarios
- Added `Dialog - Android Mobile` test suite covering accept, dismiss, and prompt text scenarios
- Tests verify the exact command sequence for iOS (SpringBoard activation flow)

### iOS
When a permission dialog appears on iOS, the dialog is owned by SpringBoard (system UI) rather than the app itself. The implementation:
1. Gets the current app's bundle ID via `mobile: activeAppInfo`
2. Activates SpringBoard (`com.apple.springboard`) to make the dialog accessible
3. Accepts or dismisses the alert (with optional text for prompts)
4. Reactivates the original app

### Android
Permission dialogs on Android are handled directly via the WebDriver `acceptAlert`/`dismissAlert` commands without app switching.

## Usage Example

```javascript
// Listen for dialogs (works on both browser and mobile)
browser.on('dialog', async (dialog) => {
  // Accept a permission dialog
  await dialog.accept()
  
  // Or dismiss it
  // await dialog.dismiss()
  
  // For prompt dialogs, pass text
  // await dialog.accept('some input')
})
```


packages/webdriverio/src/session/dialog.ts
```diff
@@ -125,29 +130,107 @@ export class Dialog {
...
+            // Activate SpringBoard so the permission dialog is accessible
+            await browser.execute('mobile: activateApp', { bundleId: SPRINGBOARD_BUNDLE_ID })
+
+            try {
+                // Accept or dismiss the alert (may throw if no alert is present)
+                if (action === 'accept') {
+                    if (userText) {
+                        await browser.sendAlertText(userText)
+                    }
+                    await browser.acceptAlert()
```
wswebcreation:
```suggestion
                    await (userText ? $(`~${userText}`).click() : browser.acceptAlert())
```
- Added the correct element find, the elerttext doesn't work for native
- There was not else, so this runs twice

packages/webdriverio/src/session/dialog.ts
```diff
@@ -125,29 +130,107 @@ export class Dialog {
...
+            return
+        }
+
+        // Android: directly accept/dismiss the alert
+        try {
+            if (action === 'accept') {
+                if (userText) {
+                    await browser.sendAlertText(userText)
+                }
+                await browser.acceptAlert()
```
wswebcreation:
```suggestion
                    await (userText ? $(`//android.widget.Button[@text='${userText}']`).click() : browser.acceptAlert())
```
- Added the correct element find, the elerttext doesn't work for native
- There was not else, so this runs twice

packages/webdriverio/src/session/dialog.ts
```diff
@@ -125,29 +130,107 @@ export class Dialog {
      * @returns {Promise<void>}
      */
     async accept(userText?: string) {
-        const contextManager = getContextManager(this.#browser)
+        const browser = this.#browser
+
+        /**
+         * Handle mobile permission dialogs (iOS/Android)
+         */
+        if (browser.isMobile) {
```
wswebcreation:
```suggestion
        if (browser.isMobile && browser.isNativeContext) {
```

Only do this when we are in the native context

wswebcreation (changes requested):
Thanks for the enhancement. Can you maybe also add this command to 

- https://github.com/webdriverio/webdriverio/blob/main/packages/webdriverio/src/commands/mobile.ts
- and https://github.com/webdriverio/webdriverio/blob/main/packages/webdriverio/src/commands/mobile/dialog.ts

This can be the same as for https://github.com/webdriverio/webdriverio/blob/main/packages/webdriverio/src/commands/mobile/dragAndDrop.ts,  but then with some specific mobile docs. Then we have it in one place and the same as for the rest

Can you also look at the suggestions? The code wasn't working but with this it does

packages/webdriverio/src/session/dialog.ts
christian-bromann: > `$(`//android.widget.Button[@text='${userText}']`).click()`

What's the iOS version of this?

> * There was not else, so this runs twice

I assumed in the browser these were two separate functions, one setting the alert text, the other submitting it.

christian-bromann:
@wswebcreation thanks for the review. I realised that on mobile we don't have an event handler (or do we?) to provide a dialog instance. So on mobile use need to acquire the dialog instance. I will follow what you suggested and add a mobile dialog command:

```ts
const dialog = browser.dialog()
await dialog.accept()
```

packages/webdriverio/src/commands/mobile/dialog.ts
```diff
@@ -0,0 +1,167 @@
...
+    await dialog.dismiss();
+ * </example>
+ *
+ * @support ["ios","android"]
+ * @returns {MobileDialog} A MobileDialog instance for interacting with native mobile alerts.
+ */
+export function dialog(this: WebdriverIO.Browser): MobileDialog {
+    const browser = this
+
+    if (!browser.isMobile) {
```
wswebcreation, line 30:
```suggestion
    if (!browser.isMobile && browser.isNativeContext) {
```

wswebcreation (commented):
Thanks for adjusting the PR. I only have one remark, I have some "issues" with the two-step `browser.dialog()` + `dialog.accept()` pattern. On mobile with Appium, permission dialogs and alerts are already on screen when the test reaches them. IMHO the instantiation step adds "noise" without adding value for the most common mobile use case and it's also not not as we have on all other mobile commands.

Can we do something like this

```ts
// Accept a dialog, optionally by button label
await browser.acceptDialog('Allow');       // iOS permission
await browser.acceptDialog('OK');          // Android alert
await browser.acceptDialog();                 // accept whatever the default/first button is

// Dismiss a dialog
await browser.dismissDialog();
```

Or we could do something like `browser.handleDialog()` or `browser.handleMobileDialog()`

dprevost-LMI:
Draft: Some comments still pending, some P1 of Greptile.

christian-bromann:
Superseded by #15614

PR #15259, opened 2026-05-22
[browserstack-service] Add Load Testing Service (LTS) support

MihirR-BS:
## Summary

Adds support for BrowserStack's Load Testing Service in `@wdio/browserstack-service` — a runtime mode for users running their WebdriverIO suites against virtual users on BrowserStack's load-testing infrastructure. Under this mode, the pod runs the user's tests against a local Selenium hub but still needs to land per-test rows in TestHub with the correct origin classification.

## What changes

Six files in `packages/wdio-browserstack-service/src/`:

| File | Change |
|---|---|
| `util.ts` | New `isLoadTestingSession()` / `getLtsSessionId()` helpers (env-driven via `BROWSERSTACK_LTS_SESSION_ID`); `getCloudProvider()` returns `'browserstack'` under LTS |
| `testHub/utils.ts` | `getProductMap()` / `getProductMapForBuildStartCall()` set `automate=false` + `lts=true` under LTS |
| `cli/modules/testHubModule.ts` | `sendTestSessionEvent` forces `provider='browserstack'` + `product='loadTesting'` + `frameworkSessionId=<LTS env>` on the AutomationSession |
| `cli/grpcClient.ts` | Pass-through optional `product` field in `AutomationSessionConstructor.create()` so the LTS override actually reaches the binary over the wire |
| `insights-handler.ts` | `getIntegrationsObject()` session_id override + `product='loadTesting'` |
| `reporter.ts` | TestRun / Hook event integrations: session_id override + `product='loadTesting'` |

## Gating

All LTS code paths are gated on `isLoadTestingSession()`. Returns `true` only when `BROWSERSTACK_LTS_SESSION_ID` is set in the environment (the LTS runtime exports it) or `BROWSERSTACK_LTS=true`/`1` for local testing. **Non-LTS runs see zero behavior change.**

## Type of change

Adds a new feature behind a runtime-only env-var gate. No public API or default-config changes.


packages/wdio-browserstack-service/src/util.ts
```diff
@@ -1028,7 +1028,36 @@ export function getUniqueIdentifierForCucumber(world: ITestCaseHookParameter): s
...
+    // LTS pod-iterations run against a local hub (localhost:4444), so the
+    // hostname check below would resolve to 'unknown_grid' and TestHub's
+    // o11y classifier would land the row with origin=UnknownGrid and
+    // session_hashed_id=NULL. Force 'browserstack' under LTS so events flow
+    // through the integrations.browserstack.* path. Mirror of py-sdk
+    // 3af3bba6 (LTS pytest: override AutomationSession.provider to
+    // 'browserstack' so TestHub records origin=LoadTesting).
+    if (isLoadTestingSession()) {
+        return 'browserstack'
+    }
```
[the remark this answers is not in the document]
MihirR-BS: Addressed in 3f5b669. Reverted the LTS branch in `getCloudProvider()` so it goes back to the hostname-only check — Automate guards (`automateModule.onBefore/onAfterTest`, `webdriverIOModule.KEY_IS_BROWSERSTACK_HUB`, `service.ts` skip-guards, `accessibility-handler`) keep their original semantics under LTS and won't fire authenticated PUTs to Automate REST with local-Selenium session ids.

Moved the LTS-aware integration-key resolution into the two reporter paths only — `reporter.ts` (1 callsite) and `insights-handler.ts` (5 callsites). Each derives locally:

```ts
const provider = isLoadTestingSession() ? 'browserstack' : getCloudProvider(browser)
```

This is exactly your suggested shape. Same effective behavior for the TestHub-reporting path under LTS, zero behavior change for Automate-side callers under LTS.

packages/wdio-browserstack-service/src/cli/modules/testHubModule.ts
```diff
@@ -184,20 +185,35 @@ export default class TestHubModule extends BaseModule {
...
+                // LTS: BLU runner pods route through a local Selenium hub
+                // so KEY_IS_BROWSERSTACK_HUB resolves to false — without
+                // this gate the AutomationSession lands with provider=
+                // unknown_grid and TestHub's o11y classifier sets
+                // test_run.origin=UnknownGrid. Mirrors py-sdk 3af3bba6
+                // (force provider='browserstack' + product='loadTesting'
+                // under LTS) and 0efca1ae (override frameworkSessionId
+                // with the LTS pod-iteration env id).
+                const ltsActive = isLoadTestingSession()
+                const ltsSessionId = ltsActive ? getLtsSessionId() : ''
```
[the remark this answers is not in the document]
MihirR-BS: Done in 3f5b669. Hoisted both `isLoadTestingSession()` and `getLtsSessionId()` above the `for (const autoInstance of autoInstances)` loop. They're now read once per `sendTestSessionEvent` invocation instead of N times.

packages/wdio-browserstack-service/src/util.ts
```diff
@@ -1028,7 +1028,36 @@ export function getUniqueIdentifierForCucumber(world: ITestCaseHookParameter): s
...
+    if (process.env.BROWSERSTACK_LTS_SESSION_ID) {
+        return true
+    }
+    const flag = process.env.BROWSERSTACK_LTS
+    return flag === 'true' || flag === '1'
+}
+
+export function getLtsSessionId(): string {
+    return process.env.BROWSERSTACK_LTS_SESSION_ID || ''
+}
```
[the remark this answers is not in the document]
MihirR-BS: Done in 3f5b669. Exported both env var names as named constants in `constants.ts` alongside the existing `BROWSERSTACK_OBSERVABILITY` / `BROWSERSTACK_PERCY` / `BROWSERSTACK_ACCESSIBILITY` group:

```ts
export const BROWSERSTACK_LTS_SESSION_ID = 'BROWSERSTACK_LTS_SESSION_ID'
export const BROWSERSTACK_LTS = 'BROWSERSTACK_LTS'
```

`isLoadTestingSession()` and `getLtsSessionId()` now access them via `process.env[BROWSERSTACK_LTS_SESSION_ID]` / `process.env[BROWSERSTACK_LTS]` instead of bare string literals.

packages/wdio-browserstack-service/src/testHub/utils.ts
```diff
@@ -14,12 +14,19 @@ export interface Errors {
...
     return {
         observability: config.testObservability.enabled,
         accessibility: config.accessibility,
         percy: config.percy,
-        automate: config.automate,
-        app_automate: config.appAutomate
+        automate: lts ? false : config.automate,
+        app_automate: config.appAutomate,
+        lts
     }
```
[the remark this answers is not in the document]
MihirR-BS: Good catch — addressed in 3f5b669 (and reconciled with the entries-based refactor from #15267 in the merge commit 8ecd931).

`getProductMap()` and `getProductMapForBuildStartCall()` now only emit the `lts` key when `isLoadTestingSession()` returns true. For non-LTS runs the product_map payload is byte-identical to pre-LTS-PR shape — no new `lts: false` field, so backends with strict unknown-key validation aren't surprised.

```ts
const entries: [string, boolean | undefined][] = [
    ['observability', config.testObservability.enabled],
    ['accessibility', config.accessibility],
    ['percy', config.percy],
    ['automate', lts ? false : config.automate],
    ['app_automate', config.appAutomate],
]
if (lts) {
    entries.push(['lts', true])
}
```

xxshubhamxx (commented):
**BrowserStack LTS support review** — 3 actionable findings (2 warnings, 1 suggestion) against the LTS gate logic.

Non-LTS paths are unaffected; the changes are well-structured. Findings are LTS-path-specific.

packages/wdio-browserstack-service/src/cli/modules/testHubModule.ts
```diff
@@ -182,22 +183,39 @@ export default class TestHubModule extends BaseModule {
...
             for (const autoInstance of autoInstances) {
-                const sessionProvider = AutomationFramework.getState(autoInstance, AutomationFrameworkConstants.KEY_IS_BROWSERSTACK_HUB) as boolean
+                const sessionProvider = ltsActive
                     ? 'browserstack'
-                    : 'unknown_grid'
+                    : (AutomationFramework.getState(autoInstance, AutomationFrameworkConstants.KEY_IS_BROWSERSTACK_HUB) as boolean
+                        ? 'browserstack'
+                        : 'unknown_grid')
+
+                const driverFrameworkSessionId = AutomationFramework.getState(
```
xxshubhamxx: **⚠️ Warning — potential `TypeError` when `KEY_FRAMEWORK_SESSION_ID` is absent under LTS**

`driverFrameworkSessionId` is computed unconditionally before the `(ltsActive && ltsSessionId)` ternary. If the LTS local-Selenium session has never had `KEY_FRAMEWORK_SESSION_ID` set on this `autoInstance`, `AutomationFramework.getState(...)` returns `undefined` and `.toString()` throws:

```
TypeError: Cannot read properties of undefined (reading 'toString')
```

The value is only *used* when the ternary falls through to the non-LTS path, but the crash happens before the ternary is ever reached.

**Fix**: Guard the call or evaluate it lazily:
```typescript
const driverFrameworkSessionId = (ltsActive && ltsSessionId)
    ? ''
    : (AutomationFramework.getState(
          autoInstance,
          AutomationFrameworkConstants.KEY_FRAMEWORK_SESSION_ID,
      )?.toString() ?? '')
```
Or simply add a null-safe `?.toString() ?? ''` to the existing assignment so it can't crash even if the state is absent.

packages/wdio-browserstack-service/src/testHub/utils.ts
```diff
@@ -14,13 +14,23 @@ export interface Errors {
...
     const entries: [string, boolean | undefined][] = [
         ['observability', config.testObservability.enabled],
         ['accessibility', config.accessibility],
         ['percy', config.percy],
-        ['automate', config.automate],
+        ['automate', lts ? false : config.automate],
         ['app_automate', config.appAutomate],
     ]
+    if (lts) {
+        entries.push(['lts', true])
```
xxshubhamxx, line 32: **⚠️ Warning — `shouldProcessEventForTesthub` not updated; LTS-only sessions silently drop all TestHub events**

`getProductMap` (this hunk) correctly emits `lts: true` under LTS. But `shouldProcessEventForTesthub` (line 37, same file) only gates on `BROWSERSTACK_OBSERVABILITY`, `BROWSERSTACK_ACCESSIBILITY`, and `BROWSERSTACK_PERCY`. It has no awareness of LTS.

A session started with only `BROWSERSTACK_LTS=true` (the documented "local repro" flag, no observability/accessibility/percy env vars) will have `shouldProcessEventForTesthub` return `false` for every event type — meaning every `onTestStart`, `onTestEnd`, `onHookStart`, etc. in `service.ts` silently no-ops. The product map says LTS is active, but no events reach TestHub.

In production LTS runs `BROWSERSTACK_OBSERVABILITY` is presumably always set by the pod infra, but the `BROWSERSTACK_LTS=true` local-repro path doesn't guarantee that. **Add an LTS short-circuit to `shouldProcessEventForTesthub`:**

```typescript
export const shouldProcessEventForTesthub = (eventType: string): boolean => {
    if (isLoadTestingSession()) {
        return true  // LTS: always emit testhub events
    }
    if (isTrue(process.env[BROWSERSTACK_OBSERVABILITY])) {
        return true
    }
    // ... rest unchanged
}
```

packages/wdio-browserstack-service/src/insights-handler.ts
```diff
@@ -939,13 +946,21 @@ class _InsightsHandler {
...
+        // test_run.origin=LoadTesting. Mirrors py-sdk 0efca1ae (session_id
+        // override at event-emission layer) + a245a814 (product=loadTesting
+        // tag on AutomationSession). Non-LTS runs see zero behavior change.
+        const ltsActive = isLoadTestingSession()
+        const ltsSessionId = ltsActive ? getLtsSessionId() : ''
+
         return {
             capabilities: caps,
-            session_id: sessionId,
+            session_id: (ltsActive && ltsSessionId) ? ltsSessionId : sessionId,
```
xxshubhamxx: **⚠️ Warning — `BROWSERSTACK_LTS=true` without `BROWSERSTACK_LTS_SESSION_ID` produces incoherent payload**

When only the opt-in flag is set (`BROWSERSTACK_LTS=true|1`) but `BROWSERSTACK_LTS_SESSION_ID` is absent, `getLtsSessionId()` returns `''`. The guard `(ltsActive && ltsSessionId)` is `(true && '')` → falsy, so `session_id` falls back to the WebDriver driver session. But `product` is unconditionally `'loadTesting'` because `ltsActive` is `true`.

Result: the backend receives `product: 'loadTesting'` paired with a WebDriver session ID it won't recognize as an LTS session. The same pattern exists in `reporter.ts` (line 301) and `testHubModule.ts` (line 215).

**Options:**
1. Emit a warning when `ltsActive=true` but `ltsSessionId=''` so the misconfiguration is visible:
```typescript
if (ltsActive && !ltsSessionId) {
    BStackLogger.warn('LTS active but BROWSERSTACK_LTS_SESSION_ID is not set — session_id will use WebDriver session')
}
```
2. Or treat `ltsActive && !ltsSessionId` as non-LTS for *all* fields (product, session_id) consistently, not just session_id.

packages/wdio-browserstack-service/src/insights-handler.ts
```diff
@@ -373,7 +375,8 @@ class _InsightsHandler {
         if (eventType === 'HookRunStarted') {
             testData.integrations = {}
             if (this._browser && this._platformMeta) {
-                const provider = getCloudProvider(this._browser)
+                // LTS reporter override (see getCloudProvider note in util.ts)
+                const provider = isLoadTestingSession() ? 'browserstack' : getCloudProvider(this._browser)
```
xxshubhamxx: **💡 Suggestion — cache `isLoadTestingSession()` result at class construction; it's called 5× per event dispatch**

`isLoadTestingSession()` reads `process.env` twice on every call. It's now invoked at 5 independent sites in `_InsightsHandler` (lines 379, 719, 745, 856, 924) plus once inside `getIntegrationsObject()` (line 954). That's 6 `process.env` reads per test event, all producing the same answer for the process lifetime.

Since `isLoadTestingSession()` is stable (env vars don't change mid-run), compute it once in the constructor:

```typescript
class _InsightsHandler {
    private readonly _ltsActive: boolean
    private readonly _ltsSessionId: string

    constructor(...) {
        this._ltsActive = isLoadTestingSession()
        this._ltsSessionId = this._ltsActive ? getLtsSessionId() : ''
        // ...
    }
}
```

Then replace the 5 inline ternaries with `this._ltsActive`. Also eliminates the `const ltsActive = isLoadTestingSession()` re-binding in `getIntegrationsObject()`.

packages/wdio-browserstack-service/src/cli/modules/testHubModule.ts
MihirR-BS: Fixed in 0df48bd. Switched to `?.toString() ?? ''` null-safe chain.

```ts
const driverFrameworkSessionId = (
    AutomationFramework.getState(
        autoInstance,
        AutomationFrameworkConstants.KEY_FRAMEWORK_SESSION_ID,
    )?.toString() ?? ''
)
```

Pure defensive change — non-LTS path strictly safer (before: TypeError crashed the whole batch; now: empty `frameworkSessionId` for that one instance, batch continues). In practice `KEY_FRAMEWORK_SESSION_ID` is always set so this is a guard, not a behavior change.

packages/wdio-browserstack-service/src/testHub/utils.ts
MihirR-BS: Fixed in 0df48bd. Added early-exit at the top of `shouldProcessEventForTesthub`:

```ts
if (isLoadTestingSession()) {
    return true   // LTS: always emit testhub events
}
```

Mirrors py-sdk's [`should_send_event_for_testhub`](https://github.com/browserstack/browserstack-python-sdk/pull/993). Also added 2 unit tests covering the LTS short-circuit path (`BROWSERSTACK_LTS=true` alone, `BROWSERSTACK_LTS_SESSION_ID` alone) and extended `beforeEach` to clear those env vars for test isolation. Non-LTS test cases unaffected.

packages/wdio-browserstack-service/src/insights-handler.ts
MihirR-BS: Went with **option 1** (warning) in 0df48bd. Reasoning: the local-repro path is explicitly documented as `BROWSERSTACK_LTS=true` without a session id, and we don't want option 2 to silently drop those rows. Warning at SDK init makes the misconfig visible without changing the payload semantics.

Emitted once in `InsightsHandler` constructor (Fix 4's caching gives us a natural single-fire site):

```ts
if (this._ltsActive && !this._ltsSessionId) {
    BStackLogger.warn('LTS active (BROWSERSTACK_LTS=true) but BROWSERSTACK_LTS_SESSION_ID is not set — event payload will pair product=loadTesting with WebDriver-assigned session_id rather than the pod-iteration id. Set BROWSERSTACK_LTS_SESSION_ID to silence this warning.')
}
```

If backend ends up rejecting mixed payloads, we can flip to option 2 in a follow-up — but that's a behavior change that'd silently drop the local-repro path, so going conservative for now.

packages/wdio-browserstack-service/src/insights-handler.ts
MihirR-BS: Done in 0df48bd. Added two `readonly` fields to `_InsightsHandler` and computed once at construction:

```ts
private readonly _ltsActive: boolean
private readonly _ltsSessionId: string

constructor(...) {
    this._ltsActive = isLoadTestingSession()
    this._ltsSessionId = this._ltsActive ? getLtsSessionId() : ''
    // ...
}
```

Replaced all 5 inline `isLoadTestingSession() ? 'browserstack' : getCloudProvider(this._browser)` ternaries with `this._ltsActive ? ... : ...`. Also dropped the `const ltsActive = isLoadTestingSession()` re-binding inside `getIntegrationsObject` since it now reads from the cached field. Net: 6 process.env reads per event → 0.

harshit-browserstack (approved):

PR #15262, opened 2026-05-23
fix(browserstack-service): restore boolean product map types

kbrooks:
PR #15215 loosened getProductMap's return to `{ [key: string]: boolean | undefined }`, which broke the public `Capabilities.buildProductMap` and `EventProperties.productMap` contracts (both `{ [key: string]: boolean }`). It also left launchTestSession declaring `accessibilityAutomation: boolean | null` while the launcher now passes `boolean | undefined`, producing 9 TS errors during `npm run setup`.

Coerce `config.accessibility` (`boolean | undefined`) to `boolean` via `=== true` in getProductMap, and align launchTestSession's parameter to `?: boolean`. Preserves the PR's "omit unset accessibility" intent for the build-start payload without leaking `undefined`/`null` into typed boolean capabilities.

## Proposed changes

[//]: # (Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue.)

## Types of changes

[//]: # (What types of changes does your code introduce to WebdriverIO?)
[//]: # (_Put an `x` in the boxes that apply_)

- [ ] Polish (an improvement for an existing feature)
- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update (improvements to the project's docs)
- [ ] Specification changes (updates to WebDriver command specifications)
- [ ] Internal updates (everything related to internal scripts, governance documentation and CI files)

## Checklist

[//]: # (_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._)

- [x] I have read the [CONTRIBUTING](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md) doc
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added the necessary documentation (if appropriate)
- [ ] I have added proper type definitions for new commands (if appropriate)

## Backport Request

[//]: # (The current `main` branch is the development branch for WebdriverIO v9. If your change should be released to the current major version of WebdriverIO `v8`, please raise another PR with the same changes against the `v8` branch.)

- [ ] This change is solely for `v9` and doesn't need to be back-ported
- [ ] Back-ported PR at `#XXXXX`

## Further comments

[...]

PR #15223, opened 2026-04-30
chroe:Regenerate BiDi-related files from the latest CDDL artifact

mccmrunal:
## Proposed changes
Regenerate BiDi-related files from the latest CDDL artifact 
[//]: # (Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue.)

## Types of changes

[//]: # (What types of changes does your code introduce to WebdriverIO?)
[//]: # (_Put an `x` in the boxes that apply_)

- [ ] Polish (an improvement for an existing feature)
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update (improvements to the project's docs)
- [x] Specification changes (updates to WebDriver command specifications)
- [ ] Internal updates (everything related to internal scripts, governance documentation and CI files)

## Checklist

[//]: # (_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._)

- [x] I have read the [CONTRIBUTING](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md) doc
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added the necessary documentation (if appropriate)
- [ ] I have added proper type definitions for new commands (if appropriate)

## Backport Request

[//]: # (The current `main` branch is the development branch for WebdriverIO v9. If your change should be released to the current major version of WebdriverIO `v8`, please raise another PR with the same changes against the `v8` branch.)

- [x] This change is solely for `v9` and doesn't need to be back-ported
- [ ] Back-ported PR at `#XXXXX`

## Further comments

[//]: # (If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...)

PR #15263, opened 2026-05-23
fix(bidi): make generate:bidi safe to run against latest spec

kbrooks:
## Summary

`pnpm run generate:bidi` against the current W3C CDDL artifact produces a `remoteTypes.ts` that fails `tsc --emitDeclarationOnly`, breaking `pnpm run setup`. Even after fixing that, the unit test `bidi.test-d.ts` fails because a manually-preserved deprecation alias is wiped on every regeneration. Both problems are addressed here, and the regenerated sources are committed so Unit Tests can see them.

### 1. Dangling cross-file CDDL references

The spec splits types across `local.cddl` and `remote.cddl`, but some shared aliases (e.g. `browsingContext.Screencast = text`) are declared in only one file even though both reference them. The generator transforms each file in isolation, so the output has dangling references:

```
src/bidi/remoteTypes.ts(503,17): error TS2552:
  Cannot find name 'BrowsingContextScreencast'. Did you mean 'BrowsingContextStopScreencast'?
```

**Fix:** before transforming, walk each AST for unresolved `group`/`group_array` references and copy missing assignments from the other AST. Symmetric so drift in either direction is handled. Against the current artifact this backfills exactly one ref (`browsingContext.Screencast`).

### 2. Lost manual deprecation aliases

[#15158](https://github.com/webdriverio/webdriverio/pull/15158) deprecated `SessionSubscriptionRequest` in v9 (renamed to `SessionSubscribeParameters`) and kept the old name as a re-export so existing consumers wouldn't break until v10. The alias was added manually to the auto-generated `remoteTypes.ts`, so the next regeneration wipes it — which is exactly why [#15223](https://github.com/webdriverio/webdriverio/pull/15223) ("Regenerate BiDi-related files") was reverted via [#15224](https://github.com/webdriverio/webdriverio/pull/15224): regenerating broke `packages/webdriver/tests/bidi.test-d.ts`.

**Fix:** add a `REMOTE_TYPE_DEPRECATIONS` constant in `scripts/bidi/constants.ts` that the generator appends to `remoteTypes.ts` after CDDL transformation. Single source of truth — to graduate a deprecation to a breaking change in v10, delete its entry.

### 3. Regenerated sources committed

The Build job (`pnpm run setup`) regenerates `remoteTypes.ts` in its working tree and uploads `packages/*/build` as an artifact. The Unit Tests job downloads that artifact but runs vitest against the **checked-in source** — so without committed regenerated sources, the generator fixes above only help the Build job, and Unit Tests still see stale types.

Regenerated against the current CDDL artifact (`2026-05-22T14:37:09Z`) with both generator fixes applied. The regenerated files also pick up unrelated upstream spec additions (`BrowsingContextStartScreencastResult`, `BrowsingContextStopScreencastResult`, new `ErrorCode` entries) — picking up the latest spec is the point of regenerating.

## Stacking

Depends on [#15262](https://github.com/webdriverio/webdriverio/pull/15262) — until the browserstack-service compile errors land, CI never reaches the BiDi compile step.

## Test plan

- [x] `pnpm run generate:bidi` produces a clean `remoteTypes.ts` with both `BrowsingContextScreencast` and `SessionSubscriptionRequest` defined
- [x] `pnpm run setup` completes with exit 0
- [x] `vitest packages/webdriver/tests/` — 170/170 pass, including the previously-failing `bidi.test-d.ts`
- [x] ESLint clean on `scripts/bidi/`

## SOX compliance

Changes to a build script and to auto-generated WebDriver BiDi type definitions. No runtime business logic.

### DIFF
--- a/scripts/bidi/index.ts
+++ b/scripts/bidi/index.ts
@@ -11,7 +11,7 @@ import { parse as parseCDDL, type PropertyReference, type Property } from 'cddl'
 import downloadSpec from './downloadSpec.js'
 import type { CddlType } from './utils.js'
 import { backfillCrossCddlRefs, findGroupByName, writeFile } from './utils.js'
-import { BASE_PROTOCOL_SPEC, CDDL_PARSE_ERROR_MESSAGE } from './constants.js'
+import { BASE_PROTOCOL_SPEC, CDDL_PARSE_ERROR_MESSAGE, REMOTE_TYPE_DEPRECATIONS } from './constants.js'
 
 const b = types.builders
 const __dirname = url.fileURLToPath(new URL('.', import.meta.url))
@@ -52,7 +52,10 @@ backfillCrossCddlRefs(astRemote, astLocal)
 backfillCrossCddlRefs(astLocal, astRemote)
 
 await Promise.all(cddlTypes.map(async (type, i) => {
-    const cddl = transform(asts[i], { useUnknown: true })
+    let cddl = transform(asts[i], { useUnknown: true })
+    if (type === 'remote') {
+        cddl += REMOTE_TYPE_DEPRECATIONS
+    }
     await writeFile(
         path.resolve(__dirname, '..', '..', 'packages', 'webdriver', 'src', 'bidi', `${type}Types.ts`),
         cddl
--- a/scripts/bidi/index.ts
+++ b/scripts/bidi/index.ts
@@ -10,7 +10,7 @@ import { parse as parseCDDL, type PropertyReference, type Property } from 'cddl'
 
 import downloadSpec from './downloadSpec.js'
 import type { CddlType } from './utils.js'
-import { findGroupByName, writeFile } from './utils.js'
+import { backfillCrossCddlRefs, findGroupByName, writeFile } from './utils.js'
 import { BASE_PROTOCOL_SPEC, CDDL_PARSE_ERROR_MESSAGE } from './constants.js'
 
 const b = types.builders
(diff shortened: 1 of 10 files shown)


dprevost-LMI:
>The spec splits types across local.cddl and remote.cddl, but some shared aliases (e.g. browsingContext.Screencast = text) are declared in only one file even though both reference them

In fact, the declaration is missing in `remoteType.ts`. If not mistaken, local and remote do not share types. I encountered similar issues in the past, and it is often the generated CDDL from [https://github.com/w3c/webdriver-bidi](https://github.com/w3c/webdriver-bidi) that has some trouble.

I use this validator to verify whether the CDDL is in error.
image

I entered an issue to fix this missing reference and have CDDL generated correctly. See https://github.com/w3c/webdriver-bidi/issues/1125

While wiating on their fix, the simple and quickest solution is to force the code manually as below
```
export type BrowsingContextScreencast = string

export interface BrowsingContextStopScreencastParameters {
    screencast: BrowsingContextScreencast;
}
```

Until the next CDDL generation, I expect webdriver-bidi will merge my upcoming PR that will fix the issue in their repo

scripts/bidi/index.ts
```diff
@@ -31,23 +31,35 @@ if (!hasNewSpec) {
...
 
-    const cddl = transform(ast, { useUnknown: true })
+/**
+ * The CDDL spec sometimes references shared type aliases (e.g.
+ * `browsingContext.Screencast = text`) only from `local.cddl` even though
+ * `remote.cddl` consumes them. Backfill missing references in each direction
+ * from the other AST before transforming so neither output has dangling types.
+ */
+backfillCrossCddlRefs(astRemote, astLocal)
+backfillCrossCddlRefs(astLocal, astRemote)
```
dprevost-LMI, line 52: As commented directly in the PR, I strongly suspect this is a CDDL file reference problem that will be fixed in [here](https://github.com/w3c/webdriver-bidi/issues/1125).

Thereforce, having a backfill mechanism does not really seem needed and we could just fix it manually until  `webdriver-bidi` push the fix which is fairly quickly usually.

dprevost-LMI:
FYI, https://github.com/w3c/webdriver-bidi/issues/1125 is merged but CDDL is not yet up to date and do not have the merged PR information

dprevost-LMI:
FYI: Main is patched while waiting on https://github.com/w3c/webdriver-bidi/issues/1125, but it also breaks for `SessionSubscriptionRequest` since bidi was regenerated without your deprecation patches.

dprevost-LMI:
Not required anymore

PR #15268, opened 2026-05-26
fix(@wdio/browserstack-service): surface BrowserStack credential errors at bootstrap

kamal-kaur04:
## Proposed changes

When BrowserStack's binary cannot authenticate with the testhub backend (HTTP 401 on the build-creation endpoint), it populates `StartBinSessionResponse.testhub.errors` with a JSON-encoded `{ ERROR_ACCESS_DENIED | ERROR_INVALID_CREDENTIALS: { message, type } }` object. Today `@wdio/browserstack-service` consumes the response in `BrowserstackCLI.startMain()` without reading this field, so on invalid credentials the user sees:

1. `bootstrap: failed to bootstrap TypeError: Cannot read properties of undefined (reading 'automate')` from `APIUtils.updateURLSForGRR` (because the auth-failed response has no `config.apis` block to dereference)
2. A 2-minute hub `POST /session` timeout
3. The only `401 Unauthorized` mention buried in a "funnel data" side-channel error at the very end of the log

— with no actionable indication that the actual cause is invalid credentials.

This PR adds a small `logCredentialErrors(response)` method to `BrowserstackCLI` in `packages/wdio-browserstack-service/src/cli/index.ts`. It runs between `startBinSession()` and `loadModules()` in `startMain()`, parses `response.testhub?.errors` (a `Uint8Array` / `Buffer` of JSON bytes), and on the first match of `ERROR_INVALID_CREDENTIALS` / `ERROR_ACCESS_DENIED` emits the message via `this.logger.error(...)`. Parse failures are silently ignored — `testhub.errors` is best-effort and we never want this log line to break bootstrap.

The credential message now appears in the WDIO log *before* the existing downstream error, so the user sees the actionable cause first. No behavior change when authentication succeeds or when the binary doesn't populate `testhub.errors`.

A companion change in `browserstack-binary` ensures the `testhub.errors` field is reliably populated and survives the gRPC response — see https://github.com/browserstack/browserstack-binary/pull/1259 (BrowserStack-internal).

Related BrowserStack tracking: https://browserstack.atlassian.net/browse/SDK-2038

## Types of changes

## Checklist

- [x] I have added tests that prove my fix is effective or that my feature works — end-to-end reproduction was executed and confirms the bug, but adding a unit test for `logCredentialErrors` is in progress; happy to add it on review feedback
- [x] I have added the necessary documentation (if appropriate) — JSDoc on the new method explains the purpose
- [x] I have added proper type definitions for new commands (if appropriate) — uses the existing `StartBinSessionResponse` type

## Backport Request

## Further comments

The implementation deliberately reuses the existing `TestHub.errors` proto field rather than introducing a new field on `StartBinSessionResponse`, because the binary's testhub module already has the `ACCESS_DENIED` / `INVALID_CREDENTIALS` constants and the `logBuildError` pipeline that populates the same field for other error keys. Keeping the wire surface unchanged means future BrowserStack consumers (other thin SDKs) can adopt the same pattern without coordinating a proto bump.


rounak610 (commented):
## Automated SDK PR Review

**Verdict:** Approve with comments — code is correct and a genuine improvement, but needs a unit test before merge.

**Summary:** 0 critical · 2 warnings · 1 suggestion across 1 file reviewed.

Moving error logging ahead of `updateURLSForGRR` plus the `!config.apis` guard is the right fix — the old inline `testhub.errors` block sat *after* the line that threw on auth failure, so it was unreachable. The wire contract checks out against the companion `browserstack-binary#1259`: `testhub.errors` is a `Buffer` of a JSON map `{ ERROR_KEY: { message, type } }`, exactly what `Object.entries(JSON.parse(...))` consumes here. Note this PR is paired with that binary change — the credential message only surfaces once it ships in the bundled binary.

See inline comments below for full Problem and Suggested Fix detail on each finding.

_Generated by Automated SDK PR review._

packages/wdio-browserstack-service/src/cli/index.ts
```diff
@@ -184,6 +188,29 @@ export class BrowserstackCLI {
...
     }
 
+    /**
+     * Log any build errors the binary reported via testhub.errors. The
+     * field is a JSON-encoded { [errorKey]: { message, type } } map. Called
+     * early in loadModules so the user sees the actionable cause (e.g.
+     * invalid credentials) before any downstream bootstrap step that
+     * depends on a fully populated config.
+     */
+    private logBuildErrors(startBinResponse: StartBinSessionResponse) {
```
rounak610, line 190: ## ⚠️ Warning — [TESTING] No unit test for the new error-surfacing behavior

### Problem

This change has no accompanying test (the checklist notes one is "in progress"). The project's contribution gate requires tests, so this will block review. The behavior worth covering is concrete and easy to fake.

### Suggested Fix

Add a unit test that constructs a `StartBinSessionResponse` with `testhub.errors` set to a `Buffer` of `JSON.stringify({ ERROR_ACCESS_DENIED: { message: 'Access to BrowserStack denied due to incorrect credentials.', type: 'info' } })`, spies on `BStackLogger.error`, and asserts: (1) `logBuildErrors` logs `[Build] ERROR_ACCESS_DENIED: ...`; (2) with `config.apis` absent, `loadModules` throws the incomplete-config error; (3) a malformed/empty `testhub.errors` neither logs nor throws (best-effort path).

**Confidence:** 🟢 Grounded — no test is present in the diff; verifiable.

packages/wdio-browserstack-service/src/cli/index.ts
```diff
@@ -130,6 +130,22 @@ export class BrowserstackCLI {
         this.logger.info(`loadModules: binSessionId=${this.binSessionId}`)
 
         this.setConfig(startBinResponse)
+
+        // Surface any build errors the binary populated on testhub.errors
+        // BEFORE the config.apis dereference below. On auth failure the
+        // binary returns an empty/degenerate config, so reporting errors
+        // here ensures the user sees the actionable cause (e.g. invalid
+        // credentials) instead of only a downstream TypeError.
+        this.logBuildErrors(startBinResponse)
```
rounak610, line 139: ## ⚠️ Warning — [DOCS] PR description doesn't match the implementation

### Problem

The description (and checklist) refer to a method `logCredentialErrors(response)` that logs only the *first* `ERROR_INVALID_CREDENTIALS` / `ERROR_ACCESS_DENIED` match and runs in `startMain()` between `startBinSession()` and `loadModules()`. The actual method is `logBuildErrors(startBinResponse)`, it logs **all** error codes the binary reports, and it runs **inside** `loadModules()` right after `setConfig`. The code is fine; the description describes an earlier iteration.

### Suggested Fix

Update the PR description and checklist to match `logBuildErrors` (logs all reported build errors, called early in `loadModules`), so reviewers evaluate the change that's actually here.

**Confidence:** 🟢 Grounded — description vs. diff mismatch is directly verifiable.

packages/wdio-browserstack-service/src/cli/index.ts
```diff
@@ -130,6 +130,22 @@ export class BrowserstackCLI {
...
+
+        // If the binary couldn't authenticate enough to populate apis,
+        // short-circuit with a clean error rather than letting
+        // updateURLSForGRR crash with a confusing TypeError. The build
+        // errors logged above already explained the cause.
+        if (!this.config?.apis) {
+            throw new Error('BrowserStack binary returned an incomplete config; see preceding build errors.')
```
rounak610: ## 💡 Suggestion — [CORRECTNESS] "see preceding build errors" can dangle on an older binary

### Problem

This message assumes `logBuildErrors` just logged something. Against a BrowserStack binary that doesn't yet include the companion fix (browserstack-binary#1259), `testhub.errors` is empty, nothing gets logged, and the thrown error points to non-existent "preceding build errors." This PR can release independently of the bundled binary version, so that skew is reachable in the field. (It's still strictly better than today's `TypeError` + 2-min hub timeout.)

### Suggested Fix

Make the message stand on its own, e.g. `throw new Error('BrowserStack binary returned an incomplete config — this usually indicates invalid BrowserStack credentials. Check any preceding [Build] errors for details.')`

rounak610 (approved):

rounak610 (approved):

07souravkunda (approved):

dandonarahul2002 (approved):

Issue #2, opened 2011-11-02
Not working with grid2 hub yet?


baechul:
I ran the hub like this and registered several browsers from different nodes.
java -jar selenium-server-standalone-2.0rc1.jar -role hub -port 9000

The following failed.
var webdriverjs = require("webdriverjs");
var client = webdriverjs.remote({host: "localhost", port: 9000});
client.init().url("http://some.host.com/");

And here is the console output.
[13:36:20]:  COMMAND    POST     "/wd/hub/session"
[...]

Issue #1, opened 2011-09-27
LICENSE file


dhanik:
Hi Camilo,

Would it be possible for you to add a LICENSE file to repository
in the case your code is allowed to be used by others?

Unfortunately without clear license statement no one is allowed to use or copy it...

Thanks,
Denis

PR #15310, opened 2026-06-08
fix(wdio-utils): finish the correct test/hook when a mocha test or hook times out

xxshubhamxx:
## Proposed changes

### Issue
When a mocha test or a `before`/`after` hook exceeds its timeout, WebdriverIO's own race timer rejects it just before mocha. Mocha records the failure and **synchronously advances the active runnable to the next one**, so by the time the awaited wrapper continuation resumes it reads a stale `context.test` and emits `afterTest`/`afterHook` with the **wrong identity**. The timed-out entity never receives a matching finish event, so its `TestRunStarted`/`HookRunStarted` is left open — downstream reporters/services keep it "in progress", and (e.g. on BrowserStack Test Observability) a ~60-min inactivity watchdog eventually force-closes the build, inflating the reported duration.

### Fix
1. **`@wdio/utils` › `testFnWrapper`** (root cause) — snapshot the runnable identity *before* `await`-ing the spec/hook and reuse it for the after-hook args; result/error/duration and the live context are still read post-await. The correct identity now reaches `afterTest`/`afterHook` for every service.
2. **`@wdio/utils` › `shim`** — floor a disabled/zero hook timeout to a finite value so the internal `Promise.race` timer always fires (a `0` timeout otherwise arms the race at a negative, immediately-clamped delay, so no finish is ever emitted). Explicit non-zero hook timeouts are untouched; tests are never floored.
3. **`@wdio/browserstack-service`** — a worker-teardown sweep in `after()` that emits a synthetic finish for any started-but-unfinished test/hook before the request queue closes — a safety net for entities that never reach `afterX`.

### Justification
- (1) is the minimal root-cause fix: the identity was simply read one microtask too late. Snapshotting it is framework-agnostic and fixes tests **and** hooks for every downstream service in a single place.
- (2) closes the residual case where a hook's race timer never fires (no finish is emitted at all); flooring only the `0`/undefined case avoids regressing legitimate long timeouts.
- (3) is defense-in-depth for never-returning entities where `after()` still runs.
- **Verified on a live grid:** the unpatched control reproduces the orphan (`passed 1 / in_progress 1`, hung test missing); patched → `failed 1 / in_progress 0` (the hung test is recorded). 136 unit tests pass (new coverage for the identity snapshot, the hook-timeout floor, and the teardown sweep); `tsc` clean on both packages.

## Types of changes

## Checklist

## Backport Request

## Further comments

The root-cause fix lives in `@wdio/utils` (`testFnWrapper` + `shim`) and benefits every reporter/service. The third change is `@wdio/browserstack-service`-specific (a teardown safety net) and is independent of the framework fix.


packages/wdio-utils/src/test-framework/testFnWrapper.ts
```diff
@@ -121,7 +144,18 @@ export const testFrameworkFnWrapper = async function (
...
+     *
+     * The snapshot is `[identityObject, context]`. We keep the snapshotted identity (index 0,
+     * the `{...context.test, parent}` object that may have gone stale post-await) but re-read the
+     * live `context` (index 1) so anything the after-hook derives from the live framework context
+     * stays current. If `afterFnArgs` was undefined we fall back to an empty args array.
+     */
+    const liveAfterArgs = typeof afterFnArgs === 'function' ? afterFnArgs(this) : []
+    const afterArgs = (identitySnapshot && identitySnapshot.length > 0)
+        ? [identitySnapshot[0], ...liveAfterArgs.slice(1)]
+        : liveAfterArgs
```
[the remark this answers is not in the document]
xxshubhamxx: Thanks for the suggestion — I looked into it, but it doesn't end up achieving the intended optimization, so I'll keep the current form. Two reasons:

1. **It doesn't remove the double `afterFnArgs(this)` call.** The pre-await `identitySnapshot` and the post-await `liveAfterArgs` are both untouched by the edit — the `identityChanged` guard only gates the array rebuild (`.slice(1)` + spread), not the two `afterFnArgs(this)` invocations. So the redundant-call overhead it's meant to address is still there.

2. **`identitySnapshot[0] !== liveAfterArgs[0]` isn't a reliable "identity changed" signal**, because `afterFnArgs[0]` isn't a stable reference in the same way across frameworks:
   - **mocha** (`[{ ...context.test, parent }, context]`) and **jasmine** (`[{ ...self._lastTest }, context]`) build a *fresh* object literal at index 0 on every call, so the comparison is **always true** regardless of timeout — the guard never short-circuits.
   - **cucumber** (`[hookParams?.step, context]`) reads the *same* `hookParams.step` reference both times, so it's **always false**.

   In all three cases the resulting `afterArgs` is identical to the current code, so the change is behavior-equivalent — it just swaps an explicit, framework-uniform expression for a comparison that means different things per framework, with no runtime saving.

Net: no overhead removed and no behavior change, so I'd prefer to keep the current explicit form. Appreciate the review!

packages/wdio-browserstack-service/src/insights-handler.ts
```diff
@@ -425,6 +444,99 @@ class _InsightsHandler {
...
+            }
+            // Never emit a finish with a missing uuid — the backend cannot match it to a start,
+            // which would orphan the entity rather than close it.
+            if (!meta.uuid) {
+                BStackLogger.warn('Skipping synthetic finish for ' + fullTitle + ' — stashed uuid is missing.')
+                continue
+            }
+
+            try {
+                meta.finishedAt = (new Date()).toISOString()
```
harshit-browserstack: `meta.finishedAt` is stamped before the listener call and isn't cleared when the listener throws. If `hookFinished`/`testFinished` raises here, the entry stays marked as finished while the event never reached the backend — any later pass that checks `finishedAt` will skip it, silently re-orphaning the exact entity this sweep exists to close, and the failure is only visible at debug level.

Suggestion: stamp into a local, build the payload from it, and only assign `meta.finishedAt` after the listener call succeeds (or clear it in the `catch`) — and consider logging the sweep failure at `warn` rather than `debug`. There's currently no test covering a listener throw during the sweep.

packages/wdio-browserstack-service/src/insights-handler.ts
```diff
@@ -425,6 +444,99 @@ class _InsightsHandler {
...
+            scopes: meta.scopes,
+            identifier: fullTitle,
+            file_name: filename ? path.relative(process.cwd(), filename) : undefined,
+            location: filename ? path.relative(process.cwd(), filename) : undefined,
+            vc_filepath: (this._gitConfigPath && filename) ? path.relative(this._gitConfigPath, filename) : undefined,
+            started_at: meta.startedAt,
+            finished_at: meta.finishedAt,
+            framework: this._framework,
+            result: 'failed',
+            ...getFailureObject(new Error('Test/hook did not finish before the session ended (incomplete).'))
```
harshit-browserstack: Minor: the `...getFailureObject(...)` spread comes after `result: 'failed'`, so if that helper ever returns a `result` key it would silently override the synthetic failure status. Setting `result` after the spread (or a short comment noting the helper never returns `result`) would make the invariant explicit.

packages/wdio-browserstack-service/src/insights-handler.ts
```diff
@@ -681,7 +793,10 @@ class _InsightsHandler {
 
     private getRunData (test: Frameworks.Test, eventType: string, results?: Frameworks.TestResult) {
         const fullTitle = getUniqueIdentifier(test, this._framework)
-        const testMetaData = this._tests[fullTitle]
+        // never let a missing meta entry crash the finish (which would silently drop the
+        // event and orphan the test). A start always seeds this._tests[fullTitle]; if it is absent
+        // here the identity was mislabelled upstream — degrade gracefully with an empty meta.
+        const testMetaData = this._tests[fullTitle] || ({} as TestMeta)
```
harshit-browserstack, line 829: With this fallback, when the identity is missing from `_tests`, `testMetaData.uuid` is `undefined` and `getRunData` returns a payload with `uuid: undefined`. `afterTest` now has the null-uuid guard so that path is covered — but `afterHook` calls `getRunData` on the same path and has no equivalent guard, so a hook whose entry is missing would still emit an unmatched `HookRunFinished` with no uuid. Should `afterHook` get the same guard as `afterTest`?

packages/wdio-browserstack-service/src/service.ts
```diff
@@ -503,6 +582,17 @@ export default class BrowserstackService implements Services.ServiceInstance {
                 }
             })()
 
+            // Teardown safety net: close any started-but-unfinished test/hook with a synthetic
+            // finish so the backend does not orphan them to the build watchdog (which would inflate
+            // the reported duration). This MUST run before onWorkerEnd, which shuts the event
+            // batcher down inside teardown; events enqueued after that are dropped. Wrapped so a
+            // sweep failure never breaks the session teardown.
+            try {
+                await this._insightsHandler?.sweepUnfinished()
```
harshit-browserstack, line 581: `sweepUnfinished` closes the `_tests` entries, but `_openMochaTests` and `_cliTestUuids` are only drained in `afterTest` — exactly the callback that was skipped for the tests this sweep handles. The entries for swept tests therefore stay behind; stale FIFO entries from a timeout cycle could make a later `_resolveFinishedMochaTest` in the same worker dequeue the wrong test if lifecycle ordering is ever unexpected. Low severity given this runs at per-worker teardown, but worth clearing `_openMochaTests` and the corresponding `_cliTestUuids` entries here (or at the end of `after()`) to close that class of future bugs.

packages/wdio-browserstack-service/src/service.ts
```diff
@@ -412,6 +473,24 @@ export default class BrowserstackService implements Services.ServiceInstance {
...
+                    const trackedInstance = TestFramework.getTrackedInstance()
+                    if (trackedInstance) {
+                        TestFramework.setState(trackedInstance, TestFrameworkConstants.KEY_TEST_UUID, resolvedUuid)
+                    }
+                    // Clean up so the per-worker map does not grow across the run.
+                    this._cliTestUuids.delete(identifier)
+                }
+            }
             await BrowserstackCLI.getInstance().getTestFramework()!.trackEvent(TestFrameworkState.LOG_REPORT, HookState.POST, { test, result: results })
             await BrowserstackCLI.getInstance().getTestFramework()!.trackEvent(TestFrameworkState.TEST, HookState.POST, { test, result: results, suiteTite: this._suiteTitle })
```
harshit-browserstack: Pre-existing, but while touching this area: `suiteTite` → `suiteTitle`.

harshit-browserstack (commented):
Reviewed the BrowserStack-service layers of this fix — the overall approach (identity snapshot in `@wdio/utils` + FIFO correlation + teardown sweep as defense-in-depth) looks sound, and the test coverage for the FIFO/sweep happy paths is solid. A few inline findings below: two correctness gaps in the sweep/finish paths (error handling around the listener call, and a missing null-uuid guard on the hook path) and one cleanup gap in the new per-worker bookkeeping.

packages/wdio-browserstack-service/src/insights-handler.ts
xxshubhamxx: Fixed. sweepUnfinished now stamps the finish time into a local `finishedAt` and only assigns `meta.finishedAt` after the listener emit returns; on a throw the entry stays unstamped so a later pass cannot skip and re-orphan it. The sweep is single-pass and `buildSweepFinishData` is the only reader of the timestamp (consumed synchronously to build the payload), so there is no second consumer that could double-emit — stamping-on-success is strictly more accurate here. The catch now logs at warn (a failed emit is a persisted orphan), and the timestamp is passed into `buildSweepFinishData` rather than mutating meta first. Added a test: listener.testFinished throws for one of two unfinished entries — asserts the throwing entry is left unfinished, the other is still swept (per-item isolation), and the warn is logged.

packages/wdio-browserstack-service/src/insights-handler.ts
xxshubhamxx: Confirmed getFailureObject only returns failure/failure_reason/failure_type (never a result key), so this is latent today — but moved `result: 'failed'` to after the spread so the synthetic failure status stays authoritative if that ever changes, plus a one-line comment stating the invariant.

packages/wdio-browserstack-service/src/insights-handler.ts
xxshubhamxx: Yes — added the same guard to afterHook. beforeHook always seeds a uuid, so the only path to an undefined uuid is a missing/mislabelled entry degrading getRunData to an empty meta, which is exactly the unmatched HookRunFinished you describe. The guard wraps only the hookFinished emit (skip + warn when uuid is absent); the skip-propagation block stays outside it since it mints its own fresh per-test uuids and is independent of the hook uuid. Covered by two afterHook tests (emit-with-uuid, skip+warn-without-uuid).

packages/wdio-browserstack-service/src/service.ts
xxshubhamxx: Renamed to suiteTitle. Worth flagging this is not purely cosmetic: onAfterTest in automateModule reads `args.suiteTitle`, so the typo meant the TEST POST event was passing the suite title under a key the consumer never reads (it received undefined for the after-test session-name computation). The fix aligns this POST with the PRE event, which already sends suiteTitle correctly.

packages/wdio-browserstack-service/src/service.ts
xxshubhamxx: Cleared both right after sweepUnfinished in after(): `_openMochaTests.length = 0` and `_cliTestUuids.clear()`. Both are per-worker FIFOs drained only in afterTest, so swept tests left stale entries that `_resolveFinishedMochaTest` could later mis-dequeue. Clearing is safe at this point — it runs after the sweep at per-worker teardown, no further afterTest consumes them. Added a test seeding stale FIFO/uuid state and asserting both are empty after after().

packages/wdio-browserstack-service/src/service.ts
```diff
@@ -399,8 +435,33 @@ export default class BrowserstackService implements Services.ServiceInstance {
...
+     * FIFO is the one whose afterTest is firing — even when a hang past the mocha timeout has
+     * advanced mocha's internal pointer so the `test` argument names the NEXT test. We dequeue the
+     * FIFO front and use it; if its identity matches the passed `test` (the common, healthy case)
+     * nothing changes. The FIFO is only consulted for mocha; other frameworks pass through.
+     */
+    private _resolveFinishedMochaTest(test: Frameworks.Test): Frameworks.Test {
+        if (this._config.framework !== 'mocha' || this._openMochaTests.length === 0) {
+            return test
+        }
+        const oldestOpen = this._openMochaTests.shift() as Frameworks.Test
```
AakashHotchandani: The FIFO correlation here is the load-bearing part of the fix, and it depends on `beforeTest` pushes and `afterTest` shifts staying perfectly paired in start order. My concern is the mid-run desync case: if a single test's `afterTest` is genuinely never delivered mid-run (not merely mislabelled — e.g. the worker tears a test down without an `afterTest` for some other reason), the FIFO is permanently off-by-one for the rest of the worker, and every subsequent `afterTest` will `shift()` the wrong (older) test and mis-correlate the finish. In other words this converts a single orphan into a cascading mislabel for all later tests in the spec.

The `after()` drain handles the end-of-worker case, but not mid-run. Could you confirm mocha guarantees exactly one `afterTest` per `beforeTest` for every terminal state (passed/failed/timed-out/skipped) within a worker? If there is any path where `beforeTest` fires without a matching `afterTest`, the FIFO assumption breaks. An identity-keyed lookup (matching the resolved test against the FIFO and removing that specific entry, rather than blindly shifting the front) would be more robust against desync — though it's a bigger change.

packages/wdio-browserstack-service/src/service.ts
```diff
@@ -372,6 +397,11 @@ export default class BrowserstackService implements Services.ServiceInstance {
     @PerformanceTester.Measure(PERFORMANCE_SDK_EVENTS.EVENTS.SDK_HOOK, { hookType: 'beforeTest' })
     async beforeTest (test: Frameworks.Test) {
         this._currentTest = test
+        // record the start of this test so a later, possibly mislabelled afterTest
+        // can be correlated back to it (mocha finishes tests in start order within a worker).
+        if (this._config.framework === 'mocha') {
+            this._openMochaTests.push(test)
```
AakashHotchandani: Related to the FIFO question: how does this interact with mocha test retries? When a test is retried, `beforeTest` fires once per attempt, so a test with N retries would push N entries onto `_openMochaTests`. If `afterTest` only fires for the final attempt (or fires a different number of times than `beforeTest`), the FIFO depth would not match the number of finishes and would desync. Worth a test that exercises a retried-then-passing test through `beforeTest`/`afterTest` to confirm the push/shift counts stay balanced.

packages/wdio-utils/src/shim.ts
```diff
@@ -22,6 +22,17 @@ const ELEMENT_RETURN_COMMANDS = ['getElement', 'getElements']
...
+/**
+ * Floor applied to a mocha hook's effective timeout when the hook timeout is disabled or unset
+ * (0/undefined). With a disabled hook timeout the runnable timeout resolves to 0, the race timer
+ * below would be armed at `0 - TIME_BUFFER` (a non-positive delay) and never act as a real guard,
+ * so a never-returning hook would hang without ever settling and never emit its after-hook. The
+ * floor guarantees the race always has a finite, firing timer. It is intentionally large so a
+ * legitimately slow setup hook is not cut short; an EXPLICIT non-zero hook timeout is always
+ * honoured unchanged and is never lowered to this value.
+ */
+export const DEFAULT_HOOK_TIMEOUT = 20000
```
AakashHotchandani: This effectively caps a hook whose timeout was deliberately disabled (`this.timeout(0)` in a `before`/`after` hook, which in mocha means "no timeout") at 20s. That is a subtle backward-compatibility change: a user who intentionally disabled the timeout for a legitimately long-running setup hook (e.g. provisioning, large downloads, DB seeding) would now see that hook aborted at 20s where it previously ran to completion. Since this lives in `@wdio/utils` it affects every reporter/framework consumer, not just one service.

Two thoughts: (1) the value is a magic constant — could it derive from the configured `mochaOpts.timeout` / framework default instead of a fixed 20000, so it tracks user config? (2) at minimum this behavior change deserves a mention in the PR description / changelog, since disabling a hook timeout is a documented mocha feature. Capping a never-returning hook is reasonable, but 20s may be too aggressive as a silent default for intentionally-disabled timeouts.

packages/wdio-browserstack-service/src/service.ts
```diff
@@ -412,8 +473,26 @@ export default class BrowserstackService implements Services.ServiceInstance {
...
+                    if (trackedInstance) {
+                        TestFramework.setState(trackedInstance, TestFrameworkConstants.KEY_TEST_UUID, resolvedUuid)
+                    }
+                    // Clean up so the per-worker map does not grow across the run.
+                    this._cliTestUuids.delete(identifier)
+                }
+            }
             await BrowserstackCLI.getInstance().getTestFramework()!.trackEvent(TestFrameworkState.LOG_REPORT, HookState.POST, { test, result: results })
-            await BrowserstackCLI.getInstance().getTestFramework()!.trackEvent(TestFrameworkState.TEST, HookState.POST, { test, result: results, suiteTite: this._suiteTitle })
+            await BrowserstackCLI.getInstance().getTestFramework()!.trackEvent(TestFrameworkState.TEST, HookState.POST, { test, result: results, suiteTitle: this._suiteTitle })
```
AakashHotchandani, line 485: Good catch fixing the `suiteTite` -> `suiteTitle` typo — the misspelled key meant the suite title never actually reached the `TEST`/`POST` event payload. One thing to double-check: is anything on the receiving end (the CLI/event consumer) currently reading the old misspelled `suiteTite` key? If a consumer was coded against the typo, renaming it here could change what that consumer sees (from "always undefined" to "now populated", or vice-versa). Assuming the consumer reads `suiteTitle`, this is a clean fix — just worth confirming the two sides agree on the corrected spelling.

packages/wdio-utils/src/test-framework/testFnWrapper.ts
```diff
@@ -121,7 +144,18 @@ export const testFrameworkFnWrapper = async function (
...
+     * Reuse the pre-await identity snapshot so a timed-out runnable reports its OWN identity.
+     *
+     * The snapshot is `[identityObject, context]`. We keep the snapshotted identity (index 0,
+     * the `{...context.test, parent}` object that may have gone stale post-await) but re-read the
+     * live `context` (index 1) so anything the after-hook derives from the live framework context
+     * stays current. If `afterFnArgs` was undefined we fall back to an empty args array.
+     */
+    const liveAfterArgs = typeof afterFnArgs === 'function' ? afterFnArgs(this) : []
+    const afterArgs = (identitySnapshot && identitySnapshot.length > 0)
+        ? [identitySnapshot[0], ...liveAfterArgs.slice(1)]
```
AakashHotchandani, line 155: Minor robustness point: when `identitySnapshot` is non-empty but the post-await `liveAfterArgs` comes back shorter (e.g. an `afterFnArgs` that returns `[]` post-await), `liveAfterArgs.slice(1)` yields `[]`, so the live `context` (index 1) is silently dropped and only the snapshotted identity is forwarded. For the mocha/jasmine `hookArgsFn` this won't happen (it always returns a 2-tuple), so this is theoretical — but since the whole point of re-reading `liveAfterArgs` is to forward the *live* context, it might read more clearly to take the context from the pre-await `identitySnapshot[1]` as a fallback when the live one is absent. Not blocking; just noting the asymmetry between "snapshot identity" and "live everything-else".

AakashHotchandani (commented):
Thanks for the detailed write-up and the runtime verification — the root-cause analysis (mocha advancing its runnable pointer synchronously on timeout, so the awaited continuation reads a stale identity) is clear and well argued, and I traced the `testFnWrapper` premise to confirm it holds: the mocha `hookArgsFn` returns `[{ ...context.test, parent }, context]`, a fresh spread each call, so snapshotting `afterFnArgs(this)` before the `await` does genuinely freeze the identity. The layering — framework-agnostic fix in `@wdio/utils` plus a service-side safety net — is a sensible structure, and the test coverage is good (identity-snapshot, hook-timeout floor, teardown sweep, listener-throw isolation, FIFO drain).

Overall this looks like a solid fix. I left inline comments; the items I'd most like a response on before merge:

1. **FIFO desync robustness** (`service.ts` `_resolveFinishedMochaTest`): the blind `shift()` assumes `beforeTest`/`afterTest` stay perfectly paired in start order for the life of the worker. A single genuinely-missing `afterTest` mid-run would desync the FIFO permanently and mis-correlate every later test. An identity-matched removal would be more defensive than dequeuing the front. (See inline.)

2. **Retries interaction** (`service.ts` `beforeTest` push): a retried test fires `beforeTest` per attempt — does `afterTest` fire symmetrically? If not, the FIFO depth won't match the finish count. A retried-test unit test would lock this down.

3. **`DEFAULT_HOOK_TIMEOUT = 20000` backward-compat** (`shim.ts`): this caps a deliberately-disabled hook timeout (`this.timeout(0)`) at 20s for every consumer of `@wdio/utils`, not just this service. That's a behavior change for users with intentionally long setup hooks. Consider deriving from configured framework timeout rather than a fixed magic number, and at minimum calling it out in the changelog.

Minor / optional:
- `service.ts:495` — nice catch on the `suiteTite` -> `suiteTitle` typo fix; worth confirming the consuming side reads the corrected key.
- `testFnWrapper.ts:157` — `liveAfterArgs.slice(1)` silently drops the live context if the post-await args array is unexpectedly short; theoretical for mocha/jasmine but slightly asymmetric.

Things I checked that look good: the `!frameworkTimeout` floor correctly catches `0`/`undefined`/`NaN` and leaves explicit non-zero hook timeouts untouched; the null-uuid guards in `afterTest`/`afterHook`/`sweepUnfinished` are consistent and the "leave unfinished on emit-throw so a later pass can retry" reasoning in the sweep is sound; `getRunData`'s degrade-to-empty-meta avoids a crash-on-finish; the sweep is correctly ordered before `onWorkerEnd`/batcher teardown; and the cucumber exclusion is explicitly documented as a known follow-up gap rather than silently skipped.

(Sharing as comments only — I'm an external contributor, so I'll leave the approve/request-changes call to the maintainers.)

packages/wdio-browserstack-service/src/service.ts
xxshubhamxx: No, nothing in receiving end for binary here

packages/wdio-utils/src/test-framework/testFnWrapper.ts
xxshubhamxx: Dropping context here is structurally impossible. `identitySnapshot` and `liveAfterArgs` are both produced by the SAME `afterFnArgs` callback, so when it's a function they share the identical 2-tuple `[identityObject, context]` shape — `slice(1)` therefore re-attaches the live `context` that was always present and can never drop one. When `afterFnArgs` is NOT a function, `liveAfterArgs` is `[]` AND `identitySnapshot` is `undefined`, so the `identitySnapshot && identitySnapshot.length > 0` branch isn't taken and we fall through to `liveAfterArgs` unchanged. All frameworks (mocha/jasmine/cucumber) return the 2-tuple, so there's no mixed-arity case, and the suggested defensive fallback is a no-op for every reachable input. I added a one-line comment to the `afterArgs` block making the shared-`afterFnArgs` invariant explicit so this reads clearly.

packages/wdio-utils/src/shim.ts
xxshubhamxx: Agreed on both points: the floor is orthogonal scope to the orphan fix, and the bare `20000` was a magic number. I've also corrected the rationale, which was factually wrong. The old comment claimed an untimed hook "would hang forever and never emit its after-hook" — that's not what happens. With `this.timeout(0)` the runnable timeout is `0`, so `_timeout` becomes `0 - TIME_BUFFER` = `-3`ms; Node clamps a non-positive `setTimeout` delay to ~1ms and fires it, so the after-hook does fire. The floor's real purpose is to prevent that premature ~1ms kill of a hook the author deliberately marked untimed (e.g. a healthy slow provisioning/download setup hook). It is scoped to `type === 'Hook'` AND a falsy resolved timeout only; an explicit non-zero hook timeout is untouched.

On the magic number: I kept a single named constant (`DEFAULT_DISABLED_HOOK_TIMEOUT`) with a doc comment rather than deriving from config, because no larger timeout config is cleanly reachable at this layer — the only mocha-supplied timeout is `runnable._timeout`, which in the disabled case is exactly the `0` being floored, and `mochaOpts.timeout` isn't threaded down to `executeAsync` (the `timeout` param resolves to the same default for a mocha hook). A `max(config, default)` would just derive one constant from another, and threading the test timeout down would actually be more aggressive (it would cap an explicitly-untimed hook at the test default). Renamed the constant and corrected the comment at all sites (constant doc, `executeAsync`, `testFnWrapper`, and the shim test comment).

packages/wdio-browserstack-service/src/service.ts
xxshubhamxx: The FIFO stays balanced under retries. `beforeTest`/`afterTest` here are driven by `testFnWrapper` (one `beforeTest` + one `afterTest` per wrapper invocation), not by mocha reporter events — so push and shift are coupled 1:1 per invocation. A wdio-level retry re-runs the body inside a single wrapper call (still one push/shift pair). A mocha-native retry clones the test and emits a fresh `EVENT_TEST_BEGIN` (with no `EVENT_TEST_END` for the failed attempt), which re-invokes `wrappedFn` → a fresh self-contained `testFnWrapper` (its own push + shift). Either way the open depth returns to baseline after every attempt and never accumulates residue, so a later unrelated test can't dequeue a leftover from a retried predecessor. Added a test (`keeps the FIFO balanced across retry attempts`) that drives a fail/fail/pass attempt sequence and asserts the depth rises by one mid-attempt and returns to baseline after each, with finishes correlated to the right test.

packages/wdio-browserstack-service/src/service.ts
xxshubhamxx: Oldest-open rather than an identity-keyed lookup is deliberate. The passed `test`'s identity is unreliable precisely in the timeout case this FIFO exists to fix: on a timeout mocha advances `context.test` to the next test before the late `afterTest` fires, so the `test` argument names the WRONG (next) runnable. An identity-keyed lookup would then match and remove the wrong queued entry and defeat the correction. The FIFO can't desync mid-run because push (`beforeTest`) and shift (`afterTest`) are coupled 1:1 within a single `testFnWrapper` execution per test, and the race timer always settles a test so `afterTest` always fires. The per-worker `after()` clear of `_openMochaTests`/`_cliTestUuids` bounds any residue from a timeout cycle. Added a comment in `_resolveFinishedMochaTest` documenting this so the oldest-open shift isn't later "fixed" into an identity match.

packages/wdio-utils/src/shim.ts
rounak610: Reinforcing this thread with what the code actually does on `main`, because it changes the framing of the backward-compat risk (in a way that mostly *supports* the floor, but leaves one real residual concern):

Verified in `executeAsync`: with `this.timeout(0)`, `runnableTimeout === 0`, and `0 ?? …` stays `0` (nullish coalescing doesn't fall through on `0`), so `_timeout = 0 - TIME_BUFFER = -3`. Node clamps the non-positive `setTimeout` delay to ~1 ms and the race timer **fires**, rejecting with `Timeout`. So today a `this.timeout(0)` hook is killed at ~1 ms — it does **not** run to completion. The floor therefore doesn't *cap* something that previously ran unbounded; it raises a ~1 ms kill to a 20 s window, which is strictly an improvement. So the "previously ran to completion → now aborted at 20s" regression doesn't exist on this path.

The residual concern that *does* stand: `this.timeout(0)` in mocha is an explicit "no timeout" request, and this silently substitutes an **unconfigurable 20 s ceiling** for it, in core `@wdio/utils`, for every framework/reporter consumer — not just browserstack-service. A genuinely-healthy untimed setup hook that needs >20 s (device provisioning, large download, DB seed) is now killed at 20 s. Two concrete asks:
1. Derive the floor from `mochaOpts.timeout` / the framework default rather than a fixed `20000`, or expose it as config, so it tracks user intent instead of a magic number.
2. Document this as an intentional behavior change (untimed hooks are now bounded at the floor) in the changelog — it's a semantic change to `this.timeout(0)` that lives below the service layer.

rounak610 (commented):
Reviewed the full diff against the local source on the PR head (commit 07be218). The layered approach is sound and the test coverage is genuinely good. I verified the root-cause analysis end-to-end (`hookArgsFn` returns `[{...context.test, parent}, context]`, mocha advances `context.test` on timeout, snapshot-before-await fixes it). Three points below — one is a design question about overlap between fix #1 and fix #2, one a coverage gap on the CLI path, and I've also replied in-thread on the `shim.ts` floor discussion with the verified `executeAsync` mechanics. Nothing here is a hard blocker; the first two are worth a deliberate answer before merge.

_(Secrets scan: clean. Trailing-newline check: clean.)_

packages/wdio-browserstack-service/src/service.ts
```diff
@@ -399,8 +435,40 @@ export default class BrowserstackService implements Services.ServiceInstance {
...
+            )
+        }
+        return oldestOpen
+    }
+
     @PerformanceTester.Measure(PERFORMANCE_SDK_EVENTS.EVENTS.SDK_HOOK, { hookType: 'afterTest' })
-    async afterTest(test: Frameworks.Test, context: never, results: Frameworks.TestResult) {
+    async afterTest(originalTest: Frameworks.Test, context: never, results: Frameworks.TestResult) {
+        // correct a possibly-stale test identity before it reaches either reporting path.
+        const test = this._resolveFinishedMochaTest(originalTest)
```
rounak610: **fix #1 and fix #2 overlap — and the blind-shift can now *override* a correct identity.**

With the `testFnWrapper` snapshot (fix #1) in place, `afterTest`'s `originalTest` arg is *already* the correct timed-out identity even on a timeout: `testFnWrapper` builds `afterArgs = [identitySnapshot[0], liveContext, results]` and that `identitySnapshot[0]` (`{...context.test, parent}` captured pre-await) is exactly the runnable that timed out. So by the time we reach here, `originalTest` is trustworthy.

`_resolveFinishedMochaTest` then *discards* that corrected arg and `shift()`s the FIFO front instead. The stated rationale for blind-shift-over-identity ("the passed `test`'s identity is unreliable in the timeout case") is precisely what fix #1 eliminates — once #1 lands, the arg is reliable. The net effect for the legacy path: the FIFO can only *degrade* the identity relative to trusting `originalTest`, because if the FIFO ever desyncs (the open concern on L458) it will throw away a correct `originalTest` in favor of a wrong positional guess.

Question: with #1 in place, why not trust `originalTest` directly (or use the FIFO front only to *validate* it, logging on mismatch and preferring the arg)? That removes the cascade risk entirely while keeping the diagnostic. For the CLI/gRPC `_cliTestUuids` restore below, the lookup key (`getUniqueIdentifier(test)`) is also reliable straight from the corrected `originalTest`, so it doesn't need the FIFO either.

packages/wdio-browserstack-service/src/insights-handler.ts
```diff
@@ -425,6 +454,110 @@ class _InsightsHandler {
...
+     */
+    public async sweepUnfinished() {
+        if (this._framework !== 'mocha') {
+            return
+        }
+
+        for (const fullTitle of Object.keys(this._tests)) {
+            const meta = this._tests[fullTitle]
+            // Already finished, or no kind tag (cucumber/legacy entry) — nothing to sweep.
+            if (!meta || meta.finishedAt || (meta.kind !== 'hook' && meta.kind !== 'test')) {
```
rounak610, line 489: **The teardown sweep silently skips the CLI/gRPC path — the same orphan scenario it exists to cover.**

This guard requires `meta.kind` to be `'hook'`/`'test'`, which is only stamped in `beforeTest`/`beforeHook`. On the CLI/gRPC path, `beforeTest` returns early after calling `setTestData(test, uuid)` (service.ts), and `setTestData` populates `_tests[fullTitle]` as `{ uuid, startedAt }` with **no `kind`**. So every CLI-path entry hits `meta.kind !== 'hook' && meta.kind !== 'test'` → `continue`, and is never swept — even though it has a `startedAt` and could be a real orphan.

The method docstring calls out cucumber as an intentional gap but says nothing about the CLI path. Two possibilities: (a) CLI-path orphans are closed by the binary's own test_run lifecycle, in which case the JS sweep correctly stays out and this is just worth a one-line comment; or (b) it's an unnoticed hole and the sweep's safety-net doesn't actually fire when running under the CLI. Could you confirm which, and document it? If (a), a quick note here would prevent a future reader from "fixing" `setTestData` to tag `kind` and accidentally double-emitting finishes.

packages/wdio-utils/src/shim.ts
xxshubhamxx: You're right the residual stands. Removed the floor entirely — shim.ts is now a clean revert to main, and testFnWrapper.ts keeps only the pre-await identity snapshot. The floor was orthogonal to the orphan fix, which doesn't depend on it: afterHook fires regardless via the snapshot. The pre-existing ~1ms premature kill of deliberately-untimed (this.timeout(0)) hooks — _timeout becomes 0 - TIME_BUFFER = -3ms, which Node clamps to a positive minimum and fires the race timer — is a separate bug, left for its own ticket.

packages/wdio-browserstack-service/src/service.ts
xxshubhamxx: Agreed. With the testFnWrapper snapshot, originalTest is already the correct (timed-out) identity by the time it reaches afterTest, so the FIFO blind-shift was redundant and could only degrade it on a desync. Removed _openMochaTests and _resolveFinishedMochaTest; afterTest now trusts originalTest directly for both reporting paths, and the CLI _cliTestUuids restore keys off getUniqueIdentifier(originalTest) — the same key beforeTest set. One note: duplicate forEach-generated it-titles collapse on getUniqueIdentifier so they share a _cliTestUuids key. That's a pre-existing limitation on all paths (the FIFO didn't address it either — the next getUniqueIdentifier call collapsed the objects to one key regardless), so it's out of scope here.

packages/wdio-browserstack-service/src/insights-handler.ts
xxshubhamxx: Confirmed it's (a): the CLI/gRPC-path tests' test_run lifecycle is owned by the binary (trackEvent → gRPC → stopBinSession), while the sweep emits on the JS-listener pipeline. Tagging kind in setTestData would make the sweep treat them as sweepable and double-emit a finish on a transport that never opened them. Added a clarifying comment at the guard so it isn't later "fixed" into a double-emit. Any genuine CLI-path orphan is a binary-side concern.

packages/wdio-browserstack-service/src/service.ts
xxshubhamxx: Update — reversed in the latest push. With the testFnWrapper identity snapshot, originalTest is reliable, so the oldest-open FIFO was redundant and is removed in favor of the identity-keyed lookup. Superseded by the discussion on the FIFO/correlation thread.

packages/wdio-browserstack-service/src/service.ts
xxshubhamxx: Update — the FIFO is removed as of the latest push; afterTest now trusts the testFnWrapper identity snapshot (originalTest) directly, so this balance concern no longer applies. Superseded by the FIFO/correlation thread.

xxshubhamxx (commented):
## Code Review

**Verdict: Approve with minor notes**

The root-cause fix in `testFnWrapper` is clean, minimal, and framework-agnostic — it benefits all downstream services, not just BrowserStack. The defense-in-depth layers are well-engineered. Test coverage is strong.

---

### 🚨 Critical: Described Fix #2 (shim) Is Absent From the Diff

The PR description states:

> **`@wdio/utils › shim`** — floor a disabled/zero hook timeout to a finite value so the internal `Promise.race` timer always fires

No `shim.ts` change appears in the actual diff. The described residual case — a hook whose race timer never fires because timeout is `0`/undefined — is **documented but not fixed**. The teardown sweep in `insights-handler.ts` partly compensates for BrowserStack-specific reporting, but other services/reporters remain exposed to this edge case.

Please clarify: is this intentionally deferred (in which case a follow-up issue reference in the description would help), or a missing commit?

---

### Fix #1 — `testFnWrapper.ts` Identity Snapshot ✅

The design is sound: freeze the identity (index 0) pre-await, re-read the live context (index 1) post-await. Minimal, correct, framework-agnostic.

**One subtlety worth documenting:** `identitySnapshot[0]` is a reference, not a deep copy. The fix is safe because Mocha *replaces its runnable pointer* on timeout (so the snapshot keeps referencing the abandoned object), not because the object is cloned. If another framework mutates the identity object in place rather than replacing the pointer, the snapshot would still go stale. A single-line comment noting "reference is safe because Mocha replaces its pointer, not mutates the object" would prevent future confusion.

---

### Fix #2 — `service.ts` CLI Path (`_cliTestUuids`) ✅

The snapshot-at-`beforeTest`, restore-and-delete-at-`afterTest`, drain-in-`after()` lifecycle is clean and correct. Normal tests drain on `afterTest`; only timed-out tests accumulate entries until `after()`.

Also a good catch on the pre-existing typo: `suiteTite` → `suiteTitle` in the `trackEvent` call — this was silently affecting suite-title reporting.

---

### Fix #3 — `insights-handler.ts` Teardown Sweep + Guards ✅

Per-item isolation in `sweepUnfinished` (try/catch inside the loop, failed entries left unstamped) is the right pattern — a sweep failure on one entity must not silently re-orphan the others. The `getRunData` fallback to `{} as TestMeta` degrades gracefully: empty meta → no uuid → null-uuid guard warns and skips. The Cucumber exclusion is correctly documented.

One minor note on `buildSweepFinishData`:
```ts
...getFailureObject(new Error('...')),
result: 'failed'   // overrides any result key from the spread
```
The comment explains this; it's correct. ✅

---

### Test Coverage

Strong overall. One untested path: when `_cliTestUuids` has a snapshot but `TestFramework.getTrackedInstance()` returns `null` (the `if (trackedInstance)` guard at `service.ts:449`). The guard exists correctly but has no test.

| Scenario | Covered |
|---|---|
| Identity snapshot emits original (pre-await) runnable | ✅ |
| Fallback when `afterFnArgs` is undefined | ✅ |
| `afterHook` null-uuid guard (emits + warns correctly) | ✅ |
| Sweep: unfinished hook → synthetic `HookRunFinished` | ✅ |
| Sweep: unfinished test → synthetic `TestRunFinished` | ✅ |
| Already-finished entry not re-swept | ✅ |
| Sweep ordering: before `onWorkerEnd` | ✅ |
| Per-item isolation: throw doesn't kill subsequent entries | ✅ |
| CLI uuid restore via `_cliTestUuids` | ✅ |
| `_cliTestUuids` drained in `after()` | ✅ |
| `getTrackedInstance()` returns null during uuid restore | ❌ |

---

### Nits

- Comment verbosity is high (multi-paragraph blocks on several changes). Appropriate given the non-obvious race condition, but upstream reviewers may ask for trimming.
- The `TestMeta` interface additions (`kind`, `name`, `scopes`, `fileName`) are minimal and correctly optional. ✅

AakashHotchandani (approved):

harshit-browserstack (approved):

PR #15311, opened 2026-06-08
fix(browserstack-service): don't mark external-grid runs as Automate

AakashHotchandani:
### Proposed changes

When a run executes against an **external (non-BrowserStack) Selenium Grid**, the service still classified the run's origin as `Automate` because `config.automate` ignored `isBrowserstackInfra`. This mis-attributed external-grid runs.

**Change:** thread `isBrowserstackInfra` through to `BrowserStackConfig` so the origin is classified correctly and external-grid runs are no longer reported as Automate.

### Types of changes

### DIFF
--- a/packages/wdio-browserstack-service/src/config.ts
+++ b/packages/wdio-browserstack-service/src/config.ts
@@ -65,9 +65,10 @@ class BrowserStackConfig {
         options?: BrowserstackConfig & Options.Testrunner,
         config?: Options.Testrunner,
         capabilities?: Capabilities.TestrunnerCapabilities,
+        isBrowserStackInfra?: boolean,
     ): BrowserStackConfig {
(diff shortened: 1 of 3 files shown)


AakashHotchandani:
**Internal review:** reviewed and approved internally as `AakashHotchandani/webdriverio#8` (https://github.com/AakashHotchandani/webdriverio/pull/8) before upstreaming — approvals are visible on that PR.

**Re: Greptile (`launcher.ts` — `caps` not forwarded):** good catch — fixed in beb42acbd. `isBrowserstackInfra` now receives `capabilities`, so per-capability `hostname` overrides (multi-remote configs where only some capabilities target an external grid) are honored, consistent with `shouldAddServiceVersion` in the same package.

vishnuv688 (approved):

PR #15312, opened 2026-06-08
fix(browserstack-service): raise CLI download connect timeout to avoid 10s undici abort

AakashHotchandani:
### Proposed changes

Node's global `fetch` hard-codes the connection-establishment timeout to **10s**, and it is **not** overridable via `AbortSignal` (the signal bounds the overall request, not the socket connect). In constrained / containerised CI networks (e.g. Kubernetes pods) a TLS/TCP handshake to the CLI download endpoint can legitimately take longer than 10s, so the download aborts with `UND_ERR_CONNECT_TIMEOUT` even though the endpoint is healthy.

**Change:** add an internal `connectTimeoutMs` option to the `_fetch` wrapper. When set, the request is routed through a custom undici dispatcher (`Agent`, or `ProxyAgent` when an `HTTP(S)_PROXY` is configured) with a larger connect timeout; the CLI binary download now uses it. Dispatchers are cached per `(proxy, connectTimeout)` to avoid rebuilding a connection pool on every request. The default `fetch` path is untouched, so behaviour is unchanged when the option isn't used.

### Types of changes

---
ℹ️ This edits `fetchWrapper.ts`, which is also touched by a companion `proxyCaCertificate` PR. They have been validated together; whichever lands second needs only a trivial rebase of the shared dispatcher helper.

### DIFF
--- a/packages/wdio-browserstack-service/src/cli/cliUtils.ts
+++ b/packages/wdio-browserstack-service/src/cli/cliUtils.ts
@@ -43,6 +43,15 @@ const CLI_LOCK_POLL_MS = 1000
 const CLI_DOWNLOAD_TIMEOUT_MS = 5 * 60 * 1000
 const CLI_DOWNLOAD_TMP_PREFIX = 'downloaded_file_'
 const CLI_DOWNLOAD_TMP_SUFFIX = '.zip'
+// Connection-establishment timeout for CLI network calls (update_cli + binary
+// download). Node's global fetch hard-codes this at 10s and it is NOT overridable
+// via AbortSignal, so a slow TLS/TCP handshake in a constrained/containerised CI
+// network (e.g. Kubernetes pods) aborts with UND_ERR_CONNECT_TIMEOUT even though
+// the endpoint is healthy (SDK-6152). Routed through a custom undici dispatcher.
+// Overridable via env for tuning.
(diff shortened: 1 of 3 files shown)


AakashHotchandani:
**Internal review:** reviewed and approved internally as `AakashHotchandani/webdriverio#6` (https://github.com/AakashHotchandani/webdriverio/pull/6) before upstreaming — approvals are visible on that PR.

**Re: Greptile** — all three are P2; validated, no change for now (happy to follow up if maintainers prefer):

1. **Binary download lacks retry** — the connect-timeout *is* the fix for this PR's issue (10s `UND_ERR_CONNECT_TIMEOUT`) and it is applied to the binary download. Wrapping the streaming download in `fetchWithRetry` would need a fresh `AbortController` per attempt — the current one carries the overall `CLI_DOWNLOAD_TIMEOUT_MS` and, once aborted, stays aborted — so retrying the download is a separate hardening rather than part of this timeout fix.
2. **Dispatcher cache never closed** — benign as the comment notes: at most 1–2 `(proxy, timeout)` agents are created and reused for the life of a single CLI/runner process; their pools are released on process exit.
3. **Retry tests use real timers** — the delays are small (~1.5s total) and deterministic; a fake-timer refactor of the async retry loop risks promise-flush flakiness for marginal CI savings, so leaving them as real timers.

vishnuv688 (approved):

PR #15313, opened 2026-06-08
feat(browserstack-service): support proxyCaCertificate for SSL-inspecting proxies

AakashHotchandani:
### Proposed changes

Adds a `proxyCaCertificate` service option (and a `BROWSERSTACK_EXTRA_CA_CERTS` env equivalent) so users behind SSL-inspecting corporate proxies (Zscaler, Netskope, Forcepoint, …) can have the service trust their proxy's root CA for outbound HTTPS.

- **New `caCert.ts`** — reads the customer CA (PEM single cert or bundle, **or** DER, any extension), merges it with Node's default roots (`tls.rootCertificates`), installs a global undici dispatcher trusting the merged CA, and exports `NODE_EXTRA_CA_CERTS` so detached child Node processes trust it at startup. Idempotent and **never throws** — a missing/misconfigured cert falls back to the system trust store.
- **`fetchWrapper.ts`** — the proxy path attaches the merged CA to the `ProxyAgent`'s `requestTls`, so the tunnel's upstream TLS also validates. The `ProxyAgent` is cached per `(proxyUrl, hasCa)` to avoid re-parsing the CA on every request.
- **Security** — when converting a DER cert to a PEM trust-anchor file, it is written into a fresh `mkdtemp` dir with mode `0600` + `O_EXCL`/`O_NOFOLLOW`, avoiding a symlink / pre-plant race on the file the process is about to trust as a CA.

**Scope:** covers the service's own egress and detached Node children. The BrowserStack Local (Go) binary has its own proxy flags and is intentionally out of scope (consistent with the other BrowserStack SDKs).

### Types of changes

---
ℹ️ This edits `fetchWrapper.ts`, also touched by a companion CLI connect-timeout PR. Validated together; whichever lands second needs only a trivial rebase of the shared dispatcher helper.

### DIFF
--- a/packages/wdio-browserstack-service/src/caCert.ts
+++ b/packages/wdio-browserstack-service/src/caCert.ts
@@ -92,6 +92,13 @@ export function configureCaCertificate(options?: { proxyCaCertificate?: string }
         // Trade-off: if something had already installed a custom global dispatcher (tuned
         // timeouts/pools, interceptors), that config is discarded — the idiomatic undici approach.
         setGlobalDispatcher(new Agent({ connect: { ca: mergedCa } }))
+        // The custom CA is now trusted for this process's undici fetch — the primary effect is
+        // done, so mark configured BEFORE the secondary NODE_EXTRA_CA_CERTS export below. This
+        // stops a failure in that best-effort export from (a) being logged as a total "setup
+        // failed / falling back to system trust store" when the dispatcher is in fact active, and
+        // (b) leaving `configured` false so a later call re-installs the global dispatcher.
+        configured = true
+        BStackLogger.info(`proxyCaCertificate: trusting custom CA from ${certPath} (merged with system roots).`)
         // Child Node processes (e.g. the detached cleanup spawn) inherit NODE_EXTRA_CA_CERTS and
         // trust it at startup. It must be a PEM file: reuse the customer's path when already PEM,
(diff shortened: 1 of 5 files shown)


AakashHotchandani:
**Internal review:** reviewed and approved internally as `AakashHotchandani/webdriverio#7` (https://github.com/AakashHotchandani/webdriverio/pull/7) before upstreaming — approvals are visible on that PR.

**Re: Greptile:**

- **P1 (broad catch / `configured` left unset)** — fixed in a68dd5f42. `configured = true` now runs immediately after `setGlobalDispatcher` succeeds, and the `NODE_EXTRA_CA_CERTS` export is moved into its own scoped `try/catch` with an accurate message. No more "setup failed, falling back to system trust store" when the CA is in fact active, and a child-process-trust failure is now reported distinctly (and doesn't leave `configured` false / risk re-installing the global dispatcher on a later call).
- **P2 (mkdtemp dir not cleaned up)** — intentional. `NODE_EXTRA_CA_CERTS` must outlive this process so detached child Node processes (e.g. the cleanup spawn) trust the CA at their startup; removing it on `exit` would break them. The file is a CA certificate (not a private key), written `0600` in a random `mkdtemp` dir, and reaped by the OS tmp policy. Cleaning it up safely would require tracking child lifetimes, which isn't worth the regression risk here.

PR #15468, opened 2026-08-05
chore(pnpm): fix failing pnpm tests

Delta456:
## Proposed changes

## Types of changes

`pnpm test` was failing in `test:component` on two independent, unrelated bugs surfaced by dependency drift within already-declared semver ranges (no `package.json` changes needed — `pnpm install` alone picks up the newer versions):

**1. Lit marker comments leaking into `getHTML()` snapshots** (`lit.test.js`)

`getHTML()`'s `removeCommentNodes` option (default `true`) is supposed to strip HTML comments, including the marker comments Lit inserts in dev mode. For shadow-DOM piercing, comments are stripped via `$('*').contents().filter(comment).remove()` on a Cheerio tree. That selector never descends into the content of `<template>` elements — the declarative shadow roots `getHTML()` itself builds up in `populateHTML()` — because parse5/Cheerio store `<template>` children in a separate document fragment. So any comment inside a captured shadow root silently survived `removeCommentNodes: true`. This was latent for a while; it started failing tests once lit-html 3.3.0 began emitting marker comments in more places.

Fix: replace the flat selector with a recursive walk (`removeCommentNodesDeep`) that also descends into template content, at any nesting depth.

**2. Stencil style getter throwing `AppProfileStyle0 is not a function`** (`stencil.test.tsx`)

Since Stencil 4.39, the compiler generates `static get style() { return XStyle0(); }` — calling the style import as a function — instead of the previous `return XStyle0;`. `@stencil/core`'s runtime (`internal/client`) only requires the *result* to be a string. Our Vite plugin resolves that CSS import via Vite's native `?inline` mode, which yields the raw CSS text directly (not a function), so the generated call threw.

Fix: import the inlined CSS text under a private name and rebind the original name to a zero-arg function returning it, matching the calling convention the compiler now expects.

Both were caught by running the full `pnpm test` suite locally; all steps (eslint, typings, unit tests, smoke tests, and every browser-runner preset — react/vue/lit/stencil) pass end-to-end now.

## Checklist

## Backport Request

## Further comments

### DIFF
--- a/e2e/browser-runner/__snapshots__/lit.test.js.snap
+++ b/e2e/browser-runner/__snapshots__/lit.test.js.snap
@@ -41,22 +41,3 @@ exports[`snapshot testing > of objects 3`] = `
   "foo": "bar",
 }
 `;
-
-exports[`snapshot testing > of objects 5`] = `
-{
-  "parsed": {
-    "alpha": 0,
-    "hex": "#000000",
-    "rgba": "rgba(0,0,0,0)",
-    "type": "color",
-  },
-  "property": "background-color",
-  "value": "rgba(0,0,0,0)",
-}
-`;
-
-exports[`snapshot testing > of objects 7`] = `
-{
-  "foo": "bar",
-}
-`;
--- a/packages/wdio-browser-runner/src/vite/frameworks/stencil.ts
+++ b/packages/wdio-browser-runner/src/vite/frameworks/stencil.ts
@@ -121,26 +121,30 @@ async function stencilVitePlugin(rootDir: string): Promise<Plugin> {
              * Ensure that CSS imports by Stencil have an `&inline` query parameter.
              *
              * Since Stencil 4.39, the compiler emits `static get style() { return
-             * ${styleVarName}(); }`, i.e. it calls the imported style as a function
-             * (see https://github.com/ionic-team/stencil/ ... the runtime only
-             * accepts the *result* of `Cstr.style` to be a string, see
-             * `@stencil/core/internal/client`'s `typeof Cstr.style === 'string'`
-             * check). We bypass Stencil's own CSS bundling and resolve the import
-             * via Vite's native `?inline` CSS import instead, which yields the raw
-             * CSS text as the default export rather than a function. To satisfy
-             * the generated `${styleVarName}()` call, we import the CSS text under
-             * a private name and re-export the original name as a function
-             * returning that text.
+             * ${styleVarName}(); }`, i.e. it calls the imported style as a function,
+             * whereas older versions emit `return ${styleVarName};` and use the
+             * import directly (the runtime only requires the *result* of `Cstr.style`
+             * to be a string, see `@stencil/core/internal/client`'s
+             * `typeof Cstr.style === 'string'` check). We bypass Stencil's own CSS
+             * bundling and resolve the import via Vite's native `?inline` CSS import
+             * instead, which yields the raw CSS text as the default export rather
+             * than a function. We inspect the generated code to see which calling
+             * convention was actually emitted for this style var and only wrap the
+             * import in a function when the compiler calls it as one, so this keeps
+             * working across the whole declared `@stencil/core` `^4.20.0` range.
              */
             findStaticImports(transformedCode)
                 .filter((imp) => imp.specifier.includes('&encapsulation=shadow'))
                 .forEach((imp) => {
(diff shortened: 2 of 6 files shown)


packages/wdio-browser-runner/src/vite/frameworks/stencil.ts
```diff
@@ -118,15 +118,29 @@ async function stencilVitePlugin(rootDir: string): Promise<Plugin> {
...
              */
             findStaticImports(transformedCode)
                 .filter((imp) => imp.specifier.includes('&encapsulation=shadow'))
                 .forEach((imp) => {
                     const cssPath = path.resolve(path.dirname(id), imp.specifier)
+                    const styleVarName = imp.imports.trim()
                     transformedCode = transformedCode.replace(
                         imp.code,
-                        `import ${imp.imports.trim()} from '/@fs/${cssPath}&inline';\n`
+                        `import __${styleVarName}Css from '/@fs/${cssPath}&inline';\n` +
```
harsha509: This wrapping is unconditional, so it breaks any project on @stencil/core below 4.39, and it breaks quietly.

`@stencil/core` is declared here as optionalDependencies: "^4.20.0". I ran transpileSync with this plugin's exact options across that range:


`4.20.0  → static get style() { return AppProfileStyle0; }`
`4.30.0  → static get style() { return AppProfileStyle0; }`
`4.38.3  → static get style() { return AppProfileStyle0; }`
`4.39.0  → static get style() { return AppProfileStyle0(); }`
`4.43.5  → static get style() { return AppProfileStyle0(); }`

On anything under `4.39` the getter returns the identifier itself, so after this change Cstr.style is a function. The runtime's `typeof Cstr.style === "string"` check then fails, style ends up undefined, and registerStyle calls new CSSStyleSheet().replaceSync(undefined) — which does not throw (verified in Chrome: no error, 0 rules). The user just gets unstyled components with nothing to debug, which is worse than the is not a function error this fixes.

Since the generated code is right here, we can check what was actually emitted instead of assuming a version:

```
const stylesIsCalled = new RegExp(`return\\s+${styleVarName}\\s*\\(\\)`).test(transformedCode)
transformedCode = transformedCode.replace(
    imp.code,
    stylesIsCalled
        ? `import __${styleVarName}Css from '/@fs/${cssPath}&inline';\n` +
          `const ${styleVarName} = () => __${styleVarName}Css;\n`
        : `import ${styleVarName} from '/@fs/${cssPath}&inline';\n`
)
```
This matches the 4.39+ emit and not the older one, so it works on both sides with no version detection.

e2e/browser-runner/__snapshots__/lit.test.js.snap
```diff
@@ -41,22 +41,3 @@ exports[`snapshot testing > of objects 3`] = `
   "foo": "bar",
 }
 `;
-
-exports[`snapshot testing > of objects 5`] = `
```
harsha509, line 45: is this deletion required ? i think its unrelated change to this fix

packages/webdriverio/src/commands/element/getHTML.ts
```diff
@@ -195,6 +195,33 @@ function populateHTML (
...
+ *
+ * `$('*').contents()` only walks the "regular" DOM tree and does not descend
+ * into the content of `<template>` elements (e.g. the declarative shadow
+ * roots built up in `populateHTML`), since their children live in a separate
+ * document fragment. Walking the tree manually via `.contents()` on each
+ * node ensures we also reach comments nested inside (potentially multiple
+ * levels of) shadow roots, e.g. the marker comments Lit adds to track
+ * dynamic template parts.
+ */
+function removeCommentNodesDeep ($: CheerioAPI, node: CheerioNode) {
```
harsha509, line 215: Need a test may be

e2e/browser-runner/__snapshots__/lit.test.js.snap
Delta456: This was cleanup-ed as a part of running `--updateSnapshots` and was old junk code

packages/wdio-browser-runner/src/vite/frameworks/stencil.ts
Delta456: Done

packages/webdriverio/src/commands/element/getHTML.ts
Delta456: Added

harsha509 (approved):
LGTM!

Issue #15467, opened 2026-08-05
[🐛 Bug]: Stale BiDi shadow root cache produces "SharedId belongs to different document" warnings after page-initiated navigations


Badisi:
## WebdriverIO Version

9.30.1

## Node.js Version

v24.18.0

## Mode

WDIO Testrunner

## Which capabilities are you using?

```typescript
{
    browserName: 'chrome',
    browserVersion: 'stable', // Chrome 151.0.7922.72
    'goog:chromeOptions': {
        args: ['--headless', '--no-sandbox', '--disable-gpu', '--disable-dev-shm-usage']
    }
}
```

ChromeDriver 151.0.7922.76, macOS (headless and headed).

## What happened?

The `ShadowRootManager` (packages/webdriverio/src/session/shadowRoot.ts) caches the shadow roots of the current document per browsing context and passes them as `startNodes` to `browsingContext.locateNodes` on every element lookup (see `findDeepElement` in packages/webdriverio/src/commands/browser/$.ts). After a navigation, those cached SharedIds reference the *old* document, so the driver rejects them:

```
WARN webdriverio: Failed to execute browser.browsingContextLocateNodes({ ... }) due to Error: WebDriver Bidi command "browsingContext.locateNodes" failed with error: no such node - SharedId "f.….d.<old-document-id>.e.…" belongs to different document. Current document is <new-document-id>., falling back to regular WebDriver Classic command
```

WDIO only invalidates that cache in two situations, both of which miss common navigation flows:

1. `#handleBidiCommand` clears the cache when it sees a `browsingContext.navigate` **command** sent through the protocol. **Page-initiated** navigations (e.g. a button click that sets `location.href` and redirects to an IdP, then a redirect back) never send such a command.
2. `handleLogEntry` purges the cache heuristically when the *new* document registers a custom element with a shadow root (`newShadowRoot` log + document id change). If the destination page uses light DOM (no custom elements — e.g. most IdP login pages), that purge never fires.

Additionally, the manager subscribes to `browsingContext.navigationStarted` in its constructor but never registers a handler for it, so that event is effectively unused.

The warning is benign (WDIO falls back to the classic WebDriver command and still finds the element), but it produces a flood of `WARN` log noise after every page-initiated navigation in a shadow-DOM-heavy application, and it only gets worse the more shadow roots are cached.

## What is your expected behavior?

[...]

PR #15477, opened 2026-08-07
fix(webdriverio): reset current context when browsing context is destroyed

### LINKED ISSUE #15476: [Bug]: ContextManager keeps a stale #currentContext after a browsing context is destroyed (e.g. window.close()) — commands fail with "Cannot find context"
## WebdriverIO Version

9.30.1 (present on `main`)

## Node.js Version

v24.18.0

## Mode

WDIO Testrunner

## Which capabilities are you using?

```json
{
    "browserName": "chrome",
    "browserVersion": "stable"
}
```

## What happened?

`ContextManager` (packages/webdriverio/src/session/context.ts) caches the current browsing context in `#currentContext` and only ever revalidates it:

- on `browsingContext.navigationStarted`, and
- on WDIO-initiated `closeWindow` / `switchToWindow` commands.

When a browsing context is destroyed outside those paths — e.g. a popup self-closes via `window.close()` after an OAuth redirect, or an SPA removes a context — no event is handled, so `#currentContext` keeps referencing a dead context id. `browsingContext.navigationStarted` never fires for a destroyed context, so the stale context persists. Every subsequent BiDi command (`browsingContext.locateNodes`, etc.) fails with `no such frame - Context ... not found`, and the Classic fallback then throws `no such window: target window already closed`.

## What is your expected behavior?

Subscribe to `browsingContext.contextDestroyed` and, when the destroyed context is the current one, reset the cached context/window handle and switch to a live window.

## How to reproduce the bug.

Serve `main.html` and `popup.html` locally (e.g. `python3 -m http.server 8081`), then run the WDIO spec below.

`main.html`:

```html
<!doctype html>
<html>
<body>
    <button id="open">Open popup</button>
    <script>
        document.getElementById('open').onclick = () => window.open('popup.html', 'popup', 'width=400,height=400');
    </script>
</body>
</html>
```

`popup.html` (self-closes shortly after opening):

```html
<!doctype html>
<html>
<body>
    <div id="popup-only">popup</div>
    <script>setTimeout(() => window.close(), 500);</script>
</body>
</html>
```

`test.js`:

```js
const { browser, expect } = require('@wdio/globals')

describe('current context after popup self-close', () => {
    it('command runs while #currentContext still references the destroyed popup', async () => {
        await browser.url('http://localhost:8081/main.html')
        const mainHandle = await browser.getWindowHandle()
        await $('#open').click()
        const handles = await browser.getWindowHandles()
        const popupHandle = handles.find((h) => h !== mainHandle)
        expect(popupHandle).toBeDefined()
        await browser.switchToWindow(popupHandle)
        await browser.pause(1500)   // popup self-closes
        await expect($('#popup-only')).not.toBeExisting()   // runs in the stale, destroyed context
    })
})
```

## Relevant log output

The last assertion **fails** the test:

```
[0-0] WARN webdriverio: Failed to execute browser.browsingContextLocateNodes({ ... }) due to Error: WebDriver Bidi command "browsingContext.locateNodes" failed with error: no such frame - Context D4E4C439E0D7D2971195AA465599BDA5 not found, falling back to regular WebDriver Classic command
[0-0] ERROR webdriver: WebDriverError: no such window: target window already closed
no such window: WebDriverError: no such window: target window already closed
1 failing (3.9s)
```

Environment: Chrome 151.0.7922.72 + ChromeDriver 151.0.7922.76 on macOS.

## patch-package fix

Save as `patches/webdriverio+9.30.1.patch` and run `npx patch-package webdriverio`. (Source-level equivalent in `packages/webdriverio/src/session/context.ts`: add `browsingContext.contextDestroyed` to the subscription and handle it.)

```diff
diff --git a/node_modules/webdriverio/build/node.js b/node_modules/webdriverio/build/node.js
--- a/node_modules/webdriverio/build/node.js
+++ b/node_modules/webdriverio/build/node.js
@@ -309,6 +309,7 @@
   #onCommandListener;
   #onCommandResultMobileListener;
   #navigationStartedListener;
+  #destroyedListener;
   constructor(browser) {
     super(browser, _ContextManager.name);
     this.#browser = browser;
@@ -323,6 +324,7 @@
     this.#onCommandListener = this.#onCommand.bind(this);
     this.#onCommandResultMobileListener = this.#onCommandResultMobile.bind(this);
     this.#navigationStartedListener = this.#navigationStarted.bind(this);
+    this.#destroyedListener = this.#contextDestroyed.bind(this);
     this.#browser.on("result", this.#onCommandResultBidiAndClassicListener);
     if (!this.isEnabled() && !this.#browser.isMobile) {
       return;
@@ -332,9 +334,10 @@
       this.#browser.on("result", this.#onCommandResultMobileListener);
     } else {
       this.#browser.sessionSubscribe({
-        events: ["browsingContext.navigationStarted"]
+        events: ["browsingContext.navigationStarted", "browsingContext.contextDestroyed"]
       });
       this.#browser.on("browsingContext.navigationStarted", this.#navigationStartedListener);
+      this.#browser.on("browsingContext.contextDestroyed", this.#destroyedListener);
     }
   }
   removeListeners() {
@@ -345,6 +348,7 @@
       this.#browser.off("result", this.#onCommandResultMobileListener);
     } else {
       this.#browser.off("browsingContext.navigationStarted", this.#navigationStartedListener);
+      this.#browser.off("browsingContext.contextDestroyed", this.#destroyedListener);
     }
   }
   async #navigationStarted(nav) {
@@ -360,6 +364,19 @@
       return;
     }
   }
+  async #contextDestroyed(destroyed) {
+    if (!this.#currentContext || destroyed.context !== this.#currentContext) {
+      return;
+    }
+    this.#currentWindowHandle = void 0;
+    this.#currentContext = void 0;
+    const windowHandles = await this.#browser.getWindowHandles();
+    const handle = windowHandles.find((windowHandle) => windowHandle !== destroyed.context) ?? windowHandles[0];
+    if (handle) {
+      this.setCurrentContext(handle);
+      await this.#browser.switchToWindow(handle);
+    }
+  }
   #onCommandResultBidiAndClassic(event) {
```

## Related issues

- #15467 — sibling bug in the same BiDi session-manager layer (stale shadow-root cache after page-initiated navigations). Different root cause and file.

Badisi:
## What's changed

When the current browsing context is destroyed (e.g. the user closes a tab or a popup window), `ContextManager` keeps its stale `#currentContext`. Subsequent commands are then executed against a context that no longer exists, which surfaces as:

```
WARN webdriverio: Failed to find the context "…" in the current browsing context tree.
… WebDriverError: no such window: target window already closed
```

This PR makes `ContextManager` listen for the `browsingContext.contextDestroyed` event. When the destroyed context is the **current** one, it clears the cached context and window handle and switches to a remaining window handle (falling back to the first handle if none other is reported). Contexts that are not the current one are ignored.

## Implementation

- Subscribe to `browsingContext.contextDestroyed` alongside `browsingContext.navigationStarted` (`packages/webdriverio/src/session/context.ts`)
- Add `#contextDestroyed` handler that resets `#currentContext` / `#currentWindowHandle` and switches to a live window
- Register/unregister the listener in the constructor and `removeListeners()`
- Add unit tests in `packages/webdriverio/tests/session/context.test.ts` covering: listener registration, current-context destroyed → reset + switch, non-current context destroyed → ignored, and no remaining handles → no switch

## Notes

This complements the same-context navigation handling already done via `browsingContext.navigationStarted` (see #15467). It specifically covers the multi-context case where the previous context disappears entirely.

## Repro

1. Open `main.html` (opens a popup window via `window.open`)
2. Run the spec below — the popup context becomes the current context, then it is closed

```html
<!-- main.html -->
<!doctype html>
<html>
<body>
    <button id="open">Open popup</button>
    <button id="close">Close popup</button>
    <script>
        let win
        document.getElementById('open').onclick = () => { win = window.open('popup.html', 'popup', 'width=400,height=400') }
        document.getElementById('close').onclick = () => { win.close() }
    </script>
</body>
</html>
```

```html
<!-- popup.html -->
<!doctype html>
<html>
<body><p>popup</p></body>
</html>
```

```js
const { browser } = require('@wdio/globals')

describe('context destroyed', () => {
    it('recovers after the current context is destroyed', async () => {
        await browser.url('http://localhost:8081/main.html')
        await $('#open').click()
        await browser.switchWindow('popup.html')
        await browser.pause(300)
        await $('#close').click()
        await browser.pause(300)
        await browser.switchWindow('main.html') // works after the fix; fails with "no such window: target window already closed" before
    })
})
```

## Checklist

- [x] Existing unit tests still pass
- [x] New unit tests cover the regression
- [x] `eslint` passes on the changed files

Closes #15476

### DIFF
--- a/packages/webdriverio/src/session/context.ts
+++ b/packages/webdriverio/src/session/context.ts
@@ -30,6 +30,7 @@ export class ContextManager extends SessionManager {
     #onCommandListener: (event: { command: string, body: unknown }) => void
     #onCommandResultMobileListener: (event: { command: string, result: unknown }) => void
     #navigationStartedListener: (nav: local.BrowsingContextNavigationInfo) => void
+    #destroyedListener: (destroyed: local.BrowsingContextInfo) => void
 
     constructor(browser: WebdriverIO.Browser) {
         super(browser, ContextManager.name)
@@ -46,6 +47,7 @@ export class ContextManager extends SessionManager {
         this.#onCommandListener = this.#onCommand.bind(this)
         this.#onCommandResultMobileListener = this.#onCommandResultMobile.bind(this)
         this.#navigationStartedListener = this.#navigationStarted.bind(this)
+        this.#destroyedListener = this.#contextDestroyed.bind(this)
 
         /**
          * Listens for the 'closeWindow' browser command to handle context changes.
@@ -77,9 +79,10 @@ export class ContextManager extends SessionManager {
              * through navigation within e.g. frames.
              */
             this.#browser.sessionSubscribe({
-                events: ['browsingContext.navigationStarted']
+                events: ['browsingContext.navigationStarted', 'browsingContext.contextDestroyed']
             })
             this.#browser.on('browsingContext.navigationStarted', this.#navigationStartedListener)
+            this.#browser.on('browsingContext.contextDestroyed', this.#destroyedListener)
         }
     }
 
@@ -91,6 +94,7 @@ export class ContextManager extends SessionManager {
             this.#browser.off('result', this.#onCommandResultMobileListener)
         } else {
             this.#browser.off('browsingContext.navigationStarted', this.#navigationStartedListener)
+            this.#browser.off('browsingContext.contextDestroyed', this.#destroyedListener)
         }
     }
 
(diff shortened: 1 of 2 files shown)


Badisi:
Addressed both Greptile findings (confidence 3/5):

1. **Child frames vs top-level windows**: `#contextDestroyed` now checks `destroyed.parent`. If the destroyed current context is a child frame, it switches back to its parent context instead of falling through to `getWindowHandles()` and potentially selecting an unrelated top-level window. Added regression test 'switches to the parent context when a destroyed current context is a child frame'.

2. **Failed recovery consistency**: the recovery logic is now wrapped in try/catch — on a failed `switchToWindow`, a warning is logged and the cached `#currentContext`/`#currentWindowHandle` remain reset, so the next `getCurrentContext()` re-initializes from a fresh handle instead of pointing at a window we never switched to. `setCurrentContext()` is only called after the awaited switch succeeds. Added regression test 'logs a warning and does not throw if the recovery switch fails after the current context is destroyed'.

Badisi:
Re-triggering review.

Badisi:
Addressed the second round of Greptile findings (confidence 2/5):

1. **Retain failed handle**: the recovery now clears the cached context on any failed `switchToWindow` (catch block resets both `#currentContext` and `#currentWindowHandle`), so even with the production `switchToWindow` command-event side-effect caching the target handle before success, a failure no longer leaves stale state. Regression test simulates the command-event side-effect (emits the `command` event before rejecting) and asserts the next `getCurrentContext()` re-initializes.

2. **Nested frames**: `#getTopLevelContext()` walks the `browsingContextGetTree` result up the parent chain from `destroyed.parent`, resolving the top-level context before calling `switchToWindow` — no more passing immediate nested-frame parents to a top-level window command. Regression test covers a frame destroyed inside a nested child frame.

3. **Unguarded state writes**: recovery now checks `#currentContext` after resolving the target handle and bails out if a newer context transition already happened, preventing the recovery from clobbering it. Regression test simulates a concurrent `switchToWindow` during `getWindowHandles()` and asserts recovery does not overwrite it.

Badisi:
Addressed the remaining race (confidence 4/5):

- The recovery now also guards the cached context *after* the awaited `switchToWindow` resolves: if a newer transition took over while the switch was pending (`#currentContext` set and `!== handle`), the recovery does not overwrite it.
- On switch failure, the catch block only clears the cached context if no newer transition took over (i.e. `#currentContext` is still undefined or still the recovery's own handle).
- Added regression tests: 'does not overwrite a newer context transition that happens during the recovery switch' and 'does not clear a newer context transition when the recovery switch fails'.

mccmrunal:
@Badisi thanks for the PR

PR #15479, opened 2026-08-07
fix(webdriverio): record matching mock responses even when not blocked

### LINKED ISSUE #15465: [🐛 Bug]: browser.mock().waitForResponse() times out for matching requests reported with isBlocked: false at network.responseStarted
## WebdriverIO Version

9.30.1

## Node.js Version

v24.18.0

## Mode

WDIO Testrunner

## Which capabilities are you using?

```typescript
{
    browserName: 'chrome',
    browserVersion: 'stable', // Chrome 151.0.7922.72
    'goog:chromeOptions': {
        args: ['--headless', '--no-sandbox', '--disable-gpu', '--disable-dev-shm-usage']
    }
}
```

ChromeDriver 151.0.7922.76, macOS (headless and headed).

## What happened?

`browser.mock(url)` installs a BiDi network intercept on the phases `['beforeRequestSent', 'responseStarted']`. Requests that a page starts while a navigation is still in progress — e.g. fonts, stylesheets or XHRs fired on page load — are emitted by the driver at `network.responseStarted` with `isBlocked: false`, even though they match the mock's URL pattern.

`WebDriverInterception.#handleResponseStarted` in `packages/webdriverio/src/utils/interception/index.ts` gates *all* matching logic behind `#isRequestMatching()`, which is defined as:

```ts
#isRequestMatching<T>(request: T) {
    const matches = this.#pattern && this.#pattern.test(request.request.url)
    return request.isBlocked && matches
}
```

Because the driver reports these requests with `isBlocked: false`, they are treated as "not matching" and are never pushed into `#calls`. Consequences:

- `mock.calls` stays empty (`calls.length === 0`)
- `mock.waitForResponse()` never resolves and always rejects with `waitForResponse timed out after Nms`

The same request *is* reported as `isBlocked: true` at `beforeRequestSent` and is handled fine — the problem is specific to the `responseStarted` event. It also reproduces with a pure observation spy (`browser.mock()` without `respond()`/`request()`), so the bug is about *collecting* the call, not about intercepting/overwriting a response.

## What is your expected behavior?

`waitForResponse()` should resolve as soon as a response matching the mock's pattern has been observed, regardless of whether the driver marked the request as `isBlocked`. `networkProvideResponse()` (which requires the request to actually be paused) should only be issued for requests that are blocked.

## How to reproduce the bug.

Minimal example — serve `index.html` locally, then run the WDIO spec below.

`index.html`:

```html
<!doctype html>
<html>
<head>
    <link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Roboto">
</head>
<body>Hello</body>
</html>
```

`test.js`:

```js
const { browser, expect } = require('@wdio/globals')

describe('mock spy', () => {
    it('observes a subresource loaded during navigation', async () => {
        const spy = await browser.mock('https://fonts.googleapis.com/css2?family=Roboto')
        await browser.url('http://localhost:8080/')
        await browser.refresh() // triggers the font request while navigating
        await spy.waitForResponse()
        expect(spy.calls.length).toBe(1)
    })
})
```

The `refresh()` request hangs: `waitForResponse()` times out and `spy.calls.length` stays `0`.

Environment: Chrome 151.0.7922.72 + ChromeDriver 151.0.7922.76 on macOS.

## Relevant log output

```typescript
[0-0] Error: waitForResponse timed out after 10000ms
    at WebDriverInterception.waitForResponse (webdriverio/build/utils/interception/index.js:...)
```

## patch-package fix

I worked around it via patch-package. The fix records matching requests even when the driver reports `isBlocked: false`, and only issues `networkProvideResponse()` for requests that are actually blocked:

```diff
     #handleResponseStarted(request) {
-    if (!this.#isRequestMatching(request)) {
-      if (request.intercepts?.includes(this.#mockId)) {
+    const isHandledByThisMock = request.intercepts?.includes(this.#mockId);
+    const urlMatches = this.#pattern && this.#pattern.test(request.request.url);
+    if (!urlMatches) {
+      if (isHandledByThisMock) {
         return this.#browser.networkProvideResponse({
           request: request.request.request
         }).catch(this.#handleNetworkProvideResponseError);
       }
       return;
     }
     this.#attachPostData(request);
-    if (!this.#matchesFilterOptions(request)) {
+    const filterMatches = this.#matchesFilterOptions(request);
+    if (filterMatches) {
+      // record matching requests even when the driver did not mark them as "blocked"
+      // (e.g. subresources loaded while navigating) so calls[]/waitForResponse() resolve
+      this.#calls.push(request);
+    }
+    if (!request.isBlocked) {
+      if (isHandledByThisMock) {
+        return this.#browser.networkProvideResponse({
+          request: request.request.request
+        }).catch(this.#handleNetworkProvideResponseError);
+      }
+      return;
+    }
+    if (!filterMatches) {
       this.#emit("continue", request.request.request);
       return this.#browser.networkProvideResponse({
         request: request.request.request
       }).catch(this.#handleNetworkProvideResponseError);
     }
-    this.#calls.push(request);
     if (this.#respondOverwrites.length === 0 || !this.#respondOverwrites[0].overwrite) {
```

Apply with:

```sh
npx patch-package webdriverio
```

Badisi:
## What's changed

`browser.mock(url)` installs a BiDi network intercept on the phases `['beforeRequestSent', 'responseStarted']`. Requests that a page starts while a navigation is still in progress — e.g. fonts, stylesheets or XHRs fired on page load — are emitted by the driver at `network.responseStarted` with `isBlocked: false`, even though they match the mock's URL pattern.

`WebDriverInterception.#handleResponseStarted` gated *all* matching logic behind `#isRequestMatching()`, which is defined as:

```ts
#isRequestMatching<T>(request: T) {
    const matches = this.#pattern && this.#pattern.test(request.request.url)
    return request.isBlocked && matches
}
```

Because the driver reports these requests with `isBlocked: false`, they were treated as "not matching" and never pushed into `#calls`. Consequences:

- `mock.calls` stays empty (`calls.length === 0`)
- `mock.waitForResponse()` never resolves and rejects with `waitForResponse timed out after Nms`

The same request *is* reported as `isBlocked: true` at `beforeRequestSent` and handled fine — the problem is specific to the `responseStarted` event. It also reproduces with a pure observation spy (`browser.mock()` without `respond()`/`request()`), so the bug is about *collecting* the call, not about intercepting/overwriting a response.

## Implementation

`packages/webdriverio/src/utils/interception/index.ts`:

- Match on the URL pattern **independent** of `isBlocked` in `#handleResponseStarted`
- Push matching requests into `#calls` as soon as they're observed, so `mock.calls` and `waitForResponse()` work for observation-only mocks
- Only send `networkProvideResponse()` for requests the driver actually paused (`isBlocked: true`); non-blocked requests are recorded but never "provided" a response

`#isRequestMatching()` remains unchanged and is still used by `#handleBeforeRequestSent`, where requiring a paused request is the correct behavior.

## Tests

Added to `packages/webdriverio/tests/utils/interception/index.test.ts`:

- matching `responseStarted` with `isBlocked: false` is recorded and does not trigger `networkProvideResponse`
- non-matching URLs are not recorded
- `waitForResponse()` resolves for a matching response reported with `isBlocked: false` (full repro of the issue)

## Repro

Serve `index.html` locally, then run the spec below. `waitForResponse()` times out and `spy.calls.length` stays `0` before this fix.

```html
<!-- index.html -->
<!doctype html>
<html>
<head>
    <link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Roboto">
</head>
<body>Hello</body>
</html>
```

```js
const { browser, expect } = require('@wdio/globals')

describe('mock spy', () => {
    it('observes a subresource loaded during navigation', async () => {
        const spy = await browser.mock('https://fonts.googleapis.com/css2?family=Roboto')
        await browser.url('http://localhost:8080/')
        await browser.refresh() // triggers the font request while navigating
        await spy.waitForResponse()
        expect(spy.calls.length).toBe(1)
    })
})
```

## Checklist

- [x] Existing unit tests still pass
- [x] New unit tests cover the regression
- [x] `eslint` passes on the changed files

Closes #15465


Badisi:
Resolved the Greptile finding (confidence 4/5): non-blocked requests that carry this mock's intercept id are no longer sent through `networkProvideResponse`. In `#handleResponseStarted`, unblocked requests are now recorded but never continued/provided, and the catch-all branch additionally guards on `request.isBlocked` before calling `networkProvideResponse`. Added a regression test asserting `networkProvideResponse` is not called for a non-blocked request carrying `intercepts: ['mock-id']`.

Badisi:
Re-triggering review.

PR #15480, opened 2026-08-08
fix(@wdio/utils): keep an undefined element that passes filter

luantaraschi:
## Proposed changes

`filter` in `@wdio/utils` marks a rejected element with `undefined`, then strips every `undefined` from the result:

```ts
promiseArray.map(async (p, i) => {
    if (await p) {
        return await array[i]
    }
    return undefined
})).then((result) => result.filter((val) => typeof val !== 'undefined'))
```

So an element that is itself `undefined` and passes the test is stripped along with the rejected ones:

```js
await filter([undefined, 1, undefined, 2], async () => true)
// [1, 2]
```

Both references disagree with that. The docblock says the function implements `Array#filter`, and the native method keeps them. `filterSeries`, a few lines below in the same file, keeps them too, because it decides from the callback rather than from the value:

```js
[undefined, 1, undefined, 2].filter(() => true)                    // [undefined, 1, undefined, 2]
await filterSeries([undefined, 1, undefined, 2], async () => true) // [undefined, 1, undefined, 2]
await filter([undefined, 1, undefined, 2], async () => true)       // [1, 2]
```

A step returning `undefined` is ordinary, so an array of them run through the concurrent `filter` silently comes back short.

### How it is fixed

A module-level symbol marks the rejected elements, so no user value can be mistaken for one.

Holes need care, because today they are dropped by the same `typeof val !== 'undefined'` pass. The result is now assembled in an explicit loop that skips holes up front, which is where the decision belongs anyway: a hole never ran the callback. `filter should skip holes in arrays` still passes untouched, callback count included.

## Types of changes

## Checklist

The last two are left unticked because neither applies: no public API changed and no command was added. The behaviour now matches what the existing docblock already promises.

## Backport Request

Leaving both boxes unticked, because neither is true yet and I would rather you decide.

`packages/wdio-utils/src/pIteration.ts` on `v8` carries the same `filter` byte for byte, so the bug is there as well. I have not raised a `v8` PR: the branch last moved on 2026-07-01 and I do not know whether you still want fixes of this size on it. Say the word and I will open it against `v8` with the same diff.

## Further comments

Two cases were added next to the existing `filter` ones: an `undefined` that passes the test survives, cross-checked against both `Array#filter` and `filterSeries` for the same input, and an `undefined` that fails the test is still dropped.

The first fails on `main` and passes here. The second passes on both sides, so a harness that stopped reaching the function would not read as a fix.

`vitest run packages/wdio-utils/tests/pIteration.test.ts`: 76 passed. `eslint` is clean on both files.

One caveat worth stating: I installed with `--ignore-scripts`, so the workspace type packages are unbuilt and `tsc -p packages/wdio-utils` reports 82 pre-existing errors about missing `@wdio/types` declarations. None of them are in `pIteration.ts`.

While reading the file I noticed something adjacent that I left alone: `find` and `findIndex` resolve with whichever callback settles first rather than the lowest matching index, which reads differently from `Array#find`. The docblocks do say callbacks run concurrently and point at `findSeries`, so I could not tell whether that is intended. Happy to open an issue if it is not.

### DIFF
--- a/packages/wdio-utils/src/pIteration.ts
+++ b/packages/wdio-utils/src/pIteration.ts
@@ -1,3 +1,8 @@
+/* Marks an element the callback rejected. A plain `undefined` cannot do it:
+ * an element that is itself `undefined` and passes the test has to survive.
+ */
+const REJECTED = Symbol('rejected')
+
 /**
  * Implements ES5 [`Array#forEach()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach) method.<br><br>
  * Executes the provided callback once for each element.<br>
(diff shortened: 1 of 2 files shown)


christian-bromann (approved):
👍

PR #15490, opened 2026-08-11
docs: make generated API command pages machine-readable

AutomatedTester:
## Proposed changes

The API command pages under `docs/api/**` are generated from `scripts/templates/template.eta`. The markup it emits renders correctly in a browser, but loses most of its structure the moment the page is consumed as text — by an LLM, a search indexer, or anything reading the markdown rather than the rendered React tree. This is the generator half of a wider docs review; it's self-contained and touches one template plus one tag in the formatter.

Four changes, all in the generator:

**Heading levels.** Sections were emitted as `#####` (h5) directly under the page `h1`. That isn't a valid outline, and Docusaurus only collects h2–h3 into a table of contents — so every generated command page shipped an **empty TOC and no anchors** for exactly the sections you'd want to link someone to (`Parameters`, `Returns`, `Support`). They now sit one level below the command heading: `h2` on standalone command pages, `h3` on protocol pages where each command is already under an `h2` (`hasHeader` in `protocolDocs.ts`).

**Parameters table.** Was a raw `<table>` using the JSX-only `className="label labelWarning"`. The signature data is the most structured content on the site and was in the least extractable form. Now a GFM table, which also drops **318 occurrences of `className`** from the generated output — those would leak verbatim if we ever serve the markdown directly.

**Awaited usage snippets.** Snippets read `browser.url(...)` and `$(selector).click(...)` with no `await`, while the `dialog` and `clock` branches did await. WebdriverIO has been async-only since v8, and these snippets are the most-copied text on the site. Commands are now awaited unless tagged with a new `@skipAwait`, which marks the ones that genuinely don't return a promise:

- `browser.action` (returns a `KeyAction`/`PointerAction`/`WheelAction` builder)
- `browser.addCommand`, `browser.overwriteCommand`
- the eight synchronous `mock` commands — `abort`, `abortOnce`, `clear`, `redirect`, `request`, `requestOnce`, `respond`, `respondOnce`

`mock.restore()` and `mock.waitForResponse()` are async and correctly pick up the `await`, so all ten mock pages are now right rather than the previous eight-of-ten.

**Platform support as text.** `@support` was rendered only as image alt text (`Support for android`), so anything stripping images lost it. The platforms are now also listed as a text line beside the icons.

## Types of changes

## Checklist

## Backport Request

### How this was verified

All 172 generated pages were rebuilt and diffed against the same pages generated from unmodified `main`:

| Check | `main` | This PR |
|---|---|---|
| `className=` in generated markdown | 318 | **0** |
| Orphan table separators | 0 | 0 |
| Unclosed code fences | 0 | 0 |
| Table rows not followed by a blank line | 0 | 0 |
| Headings with no preceding blank line | 295 | 295 (unchanged) |

The last row is pre-existing and benign — protocol pages put `## <command>` straight after the frontmatter and the description straight after the heading. ATX headings interrupt paragraphs in CommonMark, so it parses fine; I left it alone rather than widen this PR.

Spot-checked cases worth calling out, all confirmed in the generated output:

- `@rowInfo` tags without a `::` separator are section notes rather than parameters, so they now break the table and start a fresh one below (`mobile/swipe`, `mobile/tap`, `element/scrollIntoView`)
- `|` inside parameter descriptions is escaped, so `NEVER swipe from the exact top|bottom|left|right` no longer breaks the row
- `browser.action()` renders without `await`, `await mock.restore()` with it

### Two things a reviewer may want to weigh in on

**I have not run a full `pnpm docs:build`.** The static checks above cover markdown structure, but only a real build proves MDX parses every page. I'm leaning on CI for that — happy to iterate if it surfaces anything.

**The template's dead `@support` branch.** `formatter.ts` normalises `@support` to a platform-name array (`Array.isArray(parsed) ? parsed : Object.keys(parsed)`), so the template's `{ platform: { driver: version } }` branch had become unreachable. Resolving a rebase conflict there meant reasoning about it either way, so I collapsed the two branches into one that handles both shapes. Shout if you'd rather I left the dead branch in place.

### Found while in here, deliberately not fixed

`element/waitForDisplayed` generates a 318-byte page with **no Parameters and no Examples**, and a description that is a stray fragment (`` `true` to check if the element is within the viewport. false by default. ``) appearing nowhere in the source file. Something in `formatter.ts` is selecting the wrong JSDoc block. This reproduces on unmodified `main`, so it predates this PR — I can open a separate issue if that's useful.

### DIFF
--- a/__mocks__/fetch.ts
+++ b/__mocks__/fetch.ts
@@ -311,6 +311,13 @@ const requestMock: any = vi.fn().mockImplementation((uri, params) => {
             result = body.args[0][ELEMENT_KEY] === genericElementId
                 ? { [ELEMENT_KEY]: 'some-next-elem' }
                 : {}
+        } else if (body.script.includes('getBoundingClientRect') && body.script.includes('scrollX')) {
+            // scrollIntoView: combined rect + viewport + scroll metrics
+            result = {
+                elemRect: { x: 15, y: 20, height: 30, width: 50 },
+                viewport: { width: 600, height: 800 },
+                scroll: { x: 0, y: 0 }
+            }
         } else if (body.script.includes('scrollX')) {
             result = [0, 0]
         } else if (body.script.includes('function isFocused')) {
@@ -405,6 +412,19 @@ const requestMock: any = vi.fn().mockImplementation((uri, params) => {
         ++requestMock.retryCnt
 
         return Promise.reject(timeoutError)
+    }
+
+    if (uri.pathname.endsWith('abortTimeout')) {
+        if (params?.signal?.aborted) {
+            return Promise.reject(params.signal.reason)
+        }
+
+        if (requestMock.retryCnt < 5) {
+            ++requestMock.retryCnt
+            return Promise.reject(
+                new DOMException('The operation was aborted due to timeout', 'TimeoutError')
+            )
+        }
     }
 
     if (uri.pathname.startsWith(`/session/${sessionId}/element/`) && uri.pathname.includes('/attribute/')) {
--- a/e2e/browser-runner/__snapshots__/lit.test.js.snap
+++ b/e2e/browser-runner/__snapshots__/lit.test.js.snap
@@ -41,22 +41,3 @@ exports[`snapshot testing > of objects 3`] = `
   "foo": "bar",
 }
 `;
-
-exports[`snapshot testing > of objects 5`] = `
-{
-  "parsed": {
-    "alpha": 0,
-    "hex": "#000000",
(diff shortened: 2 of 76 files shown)


christian-bromann:
Addressed the Greptile P1 on nested protocol support: fd4be7e00 restores driver/version ranges (e.g. `XCUITest (9.3+)`) for `{ platform: { driver: version } }` objects passed in from `protocolDocs.ts`, in both icon alt text and the **Supported platforms** line. Array-shaped `@support` tags are unchanged.

christian-bromann (approved):
👍

PR #15497, opened 2026-08-14
fix(@wdio/allure-reporter): don't cut the package label at a Windows drive letter

### LINKED ISSUE #15496: [🐛 Bug]: @wdio/allure-reporter toPackageLabel() collapses every test into a bogus "C" package on Windows

### WebdriverIO Version

9.30.1

### Node.js Version

v24.15.0

### Mode

Standalone Mode

```typescript
{
  browserName: 'chrome',
  platformName: 'Windows'
}
```

### What happened?

When running a WDIO suite with the `allure` reporter (`@wdio/allure-reporter`) on a Windows machine or CI agent, every single test in the generated Allure report gets grouped under one bogus top-level package literally named "C" instead of the expected package tree derived from the spec file's directory structure.

Root cause is in `toPackageLabel()` (compiled to `packages/wdio-allure-reporter/build/index.js`):

    var toPackageLabel = (p) => {
      if (!p) return "";
      const fsPath = fromUrlish(p);
      const noPos = fsPath.split(":")[0];   // <-- bug
      const rel = relNoSlash(noPos);
      return rel.replace(/\//g, ".");
    };

`fsPath.split(":")[0]` is meant to strip a trailing ":line:col" position suffix (e.g. "file.js:12:5"). That works on POSIX paths, which never contain a colon. On Windows, though, the spec's path is stored as an absolute path like:

    C:/Users/me/project/test/specs/mySuite.e2e.js

Splitting on the *first* colon here hits the drive-letter colon instead of a position suffix, so `noPos` becomes just "C" the rest of the path is discarded. Every spec ends up mapped to the same package, "C".

The file already has a correct helper for this exact purpose, used elsewhere in the same module (`fromUrlish`):

    var dropPosSuffix = (p) => p.replace(/:(\d+)(?::\d+)?$/, "");

`toPackageLabel` just doesn't use it.

Fix:

    -  const noPos = fsPath.split(":")[0];
    +  const noPos = dropPosSuffix(fsPath);

We've verified this locally with a patch-package diff against `@wdio/allure-reporter@9.30.1` and confirmed running the identical suite/config on a Linux CI agent produces the correct package tree — confirming this is Windows-path-specific, not a config issue.

Tests should be grouped in the Allure report's "Packages" view by their spec file's directory structure relative to the project root (e.g. "myTests.suiteA", "myTests.suiteB", ...), the same as they are on Linux/macOS  not all collapsed under a single package named "C".

### How to reproduce the bug.

1. Run any WDIO suite with the `allure` reporter on Windows (or a Windows CI agent).
2. Generate the Allure HTML report.
3. Open the "Packages" view.

Expected: package tree matching spec directories.
Actual: everything under one package named "C".

### Relevant log output

```typescript
NA
```

### Code of Conduct

MFA-G:
### Proposed changes

Fixes #15496.

`toPackageLabel()` in `@wdio/allure-reporter` strips a trailing `:line[:col]` position suffix off the spec path before turning it into a dotted Allure package label. It did so with `fsPath.split(':')[0]`, which is only safe on POSIX paths — those never contain a colon.

A Windows spec path starts with one:

```
C:/Users/me/project/test/specs/mySuite.e2e.js
        ^ first colon is the drive letter, not a position suffix
```

so `noPos` became `"C"` and every spec in the run collapsed into a single bogus top-level package named `C` in Allure's **Packages** view, instead of the package tree derived from the spec directories.

The module already has the correct helper for this, used by `fromUrlish()` and `toPackageLabelCucumber()`:

```ts
const dropPosSuffix = (p: string) => p.replace(/:(\d+)(?::\d+)?$/, '')
```

It anchors the suffix to the end of the string, so the drive letter is left alone. `toPackageLabel()` simply wasn't using it. One-line change:

```diff
-    const noPos = fsPath.split(':')[0]
+    const noPos = dropPosSuffix(fsPath)
```

### Types of changes

### Checklist

Added `toPackageLabel` unit tests covering the empty path, the normal directory → dotted-package mapping, `:12` / `:12:5` position suffixes (so the behaviour the old `split(':')` provided stays covered), and the Windows drive-letter regression.

The Windows case fails on `main` and passes with the fix:

```
 × toPackageLabel > keeps a Windows drive letter out of the position suffix
   → expected 'C' to not deeply equal 'C'
```

```
$ npx vitest run packages/wdio-allure-reporter/tests/utils.test.ts
 ✓ packages/wdio-allure-reporter/tests/utils.test.ts (28 tests) 15ms
 Test Files  1 passed (1)
      Tests  28 passed (28)
Type Errors  no errors

$ npx eslint packages/wdio-allure-reporter/src/utils.ts packages/wdio-allure-reporter/tests/utils.test.ts
(clean)
```

The test builds the expected paths with `path.join(process.cwd(), ...)` so it stays correct on any platform; the Windows case asserts on the suffix rather than the full label, since `path.relative` resolves the foreign absolute path differently on POSIX.

### DIFF
--- a/packages/wdio-allure-reporter/src/utils.ts
+++ b/packages/wdio-allure-reporter/src/utils.ts
@@ -282,7 +282,10 @@ export function isEmptyObject(value: unknown): boolean {
 export const toPackageLabel = (p?: string) => {
     if (!p) {return ''}
     const fsPath = fromUrlish(p)
-    const noPos = fsPath.split(':')[0]
+    // Only a trailing `:line[:col]` is a position suffix. Splitting on the first
+    // colon instead would cut a Windows path at its drive letter, collapsing every
+    // spec into a single `C` package.
+    const noPos = dropPosSuffix(fsPath)
     const rel = relNoSlash(noPos)
     return rel.replace(/\//g, '.')
 }
--- a/packages/wdio-allure-reporter/tests/utils.test.ts
+++ b/packages/wdio-allure-reporter/tests/utils.test.ts
@@ -1,6 +1,7 @@
 import { afterEach, beforeAll, describe, expect, it, vi } from 'vitest'
 import type { CommandArgs } from '@wdio/type'
 import process from 'node:process'
+import path from 'node:path'
 import { Status } from 'allure-js-commons'
 import CompoundError from '../src/compoundError.js'
 import {
@@ -14,6 +15,7 @@ import {
     isEachTypeHooks,
     isEmpty,
     isScreenshotCommand,
+    toPackageLabel,
 } from '../src/utils.js'
 import { linkPlaceholder } from '../src/constants.js'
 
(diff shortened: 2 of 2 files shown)


mccmrunal:
@MFA-G  can you sign the CLA

mccmrunal:
@MFA-G closing this PR for now its resolved

PR #15506, opened 2026-08-17
fix(webdriverio): use classic navigateTo for plain url() on macOS

### LINKED ISSUE #15481: [🐛 Bug]: browser.url() on BiDi is much slower than classic navigateTo for a simple page load

### WebdriverIO Version

Latest

### Node.js Version

v24.12.0

### Mode

WDIO Testrunner

```typescript
{
  browserName: 'chrome',
  // Chrome 151.0.7922.76, macOS arm64, headless
  // session is BiDi (browser.isBidi === true)
}
```

### What happened?

I was comparing navigation cost for a local static page (http://localhost:4173/index.html). Same browser, same URL, reloads in one session.
When the session is on BiDi, `browser.url()` is consistently in the ~450–700ms range.
If I call classic `navigateTo` on the same URL in that same session, it lands around ~65ms.

So most of the time is not “waiting for the page to load”, the page can be ready in ~65ms. The rest looks like BiDi path overhead: context lookup, `browsingContext.navigate`, and then the `url()` wrapper still waiting on network request payload so it can return the rich request object.

Rough split from my measurements:

```
browser.url() (BiDi)     ≈ 450–700ms
  ├── navigateTo (classic) ≈ 65ms     ← actual navigation / load
  └── BiDi overhead        ≈ 400–600ms
```

I also timed the raw BiDi call:

- `browsingContextNavigate` with `wait: 'complete'` → ~600ms
- `browsingContextNavigate` with `wait: 'none'` → still ~430ms

So even `wait: 'none'` does not get close to classic. The `wait` flag is not the whole story, a lot of cost is already in the BiDi navigate path / surrounding setup.

I’m not saying BiDi `url()` is wrong, the richer options (headers, auth, `onBeforeLoad`, `networkIdle`, request metadata) are useful. The issue is that the default path pays for that setup even when I only do browser.url('/index.html') and never use those options.

For a plain `browser.url(url)` with no BiDi-only options, I’d expect cost much closer to classic navigateTo / a normal page-load wait, not hundreds of ms of overhead on a local static page.

### How to reproduce the bug.

```js
// repro-url-bidi.mjs

const URL = 'http://localhost:4173/index.html // or any cheap local static page
const N = 10

  logLevel: 'error',
  capabilities: {
    browserName: 'chrome',
    'goog:chromeOptions': { args: ['--headless=new'] },
  },
})

console.log('isBidi', browser.isBidi)
await browser.url(URL) // warm-up

const urlTimes = []
const classicTimes = []
const biDiComplete = []
const biDiNone = []

const tree = await browser.browsingContextGetTree({})
const context = tree.contexts[0].context

for (let i = 0; i < N; i++) {
  let t0 = performance.now()
  await browser.url(URL)
  urlTimes.push(Math.round(performance.now() - t0))

  t0 = performance.now()
  await browser.navigateTo(URL)
  classicTimes.push(Math.round(performance.now() - t0))

  t0 = performance.now()
  await browser.browsingContextNavigate({ context, url: URL, wait: 'complete' })
  biDiComplete.push(Math.round(performance.now() - t0))

  t0 = performance.now()
  await browser.browsingContextNavigate({ context, url: URL, wait: 'none' })
  biDiNone.push(Math.round(performance.now() - t0))
}

const med = (a) => [...a].sort((x, y) => x - y)[Math.floor(a.length / 2)]
console.log({
  urlMed: med(urlTimes),
  classicNavigateToMed: med(classicTimes),
  browsingContextCompleteMed: med(biDiComplete),
  browsingContextNoneMed: med(biDiNone),
  urlSamples: urlTimes,
  classicSamples: classicTimes,
})

await browser.deleteSession()
```

```
node repro-url-bidi.mjs
```

On my machine (Chrome headless, BiDi session, warm reloads) I see url and browsingContextNavigate hundreds of ms above navigateTo.

### Relevant log output

```typescript
Based on my local http://localhost:4173/index.html


isBidi true
{
  urlMed: 334,
  classicNavigateToMed: 35,
  browsingContextCompleteMed: 694,
  browsingContextNoneMed: 336,
  urlSamples: [
    131, 356, 326, 332,
    331, 330, 335, 420,
    334, 422
  ],
  classicSamples: [
    56, 30, 29, 34, 36,
    35, 30, 36, 28, 39
  ]
}
```

### Code of Conduct

wswebcreation:
## Proposed changes

Fixes [#15481](https://github.com/webdriverio/webdriverio/issues/15481).

On **macOS**, Chrome’s BiDi `browsingContext.navigate` is much slower than classic `navigateTo` for a simple page load (same URL, same session). CI reproduced this on the gear shop fixture: classic ~100ms vs BiDi navigate ~700ms–1.7s, while Linux showed no meaningful gap (same for local Windows).

This PR fast-paths a plain `browser.url(url)` (no BiDi-only options) to classic `navigateTo` **on darwin only**. Linux/Windows keep the BiDi path and the `Request` return value. BiDi-only options (`headers`, `auth`, `onBeforeLoad`, `wait` other than `'complete'`, or any future option) still use BiDi on all platforms.

Local Mac micro-bench (localhost static page, N=10): plain `url()` ~61ms after vs ~700ms+ on the BiDi path.

**macOS note:** plain `url()` no longer returns `WebdriverIO.Request` on darwin (returns `undefined`). Pass a BiDi-only option if you need request/response metadata there. Documented in the `url` command docs.

## Types of changes

> Breaking only on **macOS** for callers that use the return value of plain `browser.url(url)` without BiDi-only options.

## Checklist

## Backport Request

## Further comments

- Root cause of the slowness is Chrome BiDi navigate on macOS (even `wait: 'none'` is expensive); WDIO can’t make that RPC cheap.
- Product fix: don’t pay for BiDi navigate (and mandatory request-metadata assembly) when the user didn’t ask for BiDi-only features — and only on the platform where that tax shows up.
- `requiresBidiNavigation()` uses a **classic-safe option allowlist** (`wait`, `timeout`); any other / future option forces BiDi so we don’t silently drop features.

Question is also if we need to merge this to `main` or to `performance-improvements`. Secondly if this is a breaking change (under certain conditions where the fix is small). I've now pointed it to `main`


packages/webdriverio/src/commands/browser/url.ts
```diff
@@ -282,13 +313,59 @@ export async function url (
...
+
+    return false
+}
+
+/**
+ * macOS Chrome pays a large BiDi `browsingContext.navigate` cost vs classic
+ * `navigateTo` for simple loads; other platforms do not (webdriverio#15481).
+ */
+export function preferClassicUrlFastPath (): boolean {
+    return process.platform === 'darwin'
```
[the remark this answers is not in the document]
wswebcreation: Good one, will check this. Running on cloudservices will definitely slow this down if the runner is Linux. Will create a fix for this

packages/webdriverio/src/commands/browser/url.ts
```diff
@@ -282,13 +313,59 @@ export async function url (
         return request
     }
 
-    if (Object.keys(options).length > 0) {
+    if (Object.keys(options).length > 0 && !this.isBidi) {
         throw new Error('Setting url options is only supported when automating browser using WebDriver Bidi protocol')
```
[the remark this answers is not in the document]
wswebcreation: Feels like a edge case

dprevost-LMI:
@greptileai, can you analyze in depth whether the fix will impact `WebDriverInterception` (see https://github.com/webdriverio/webdriverio/blob/cac60ec3afd7d8ad240eab239c2d735ea4ae69e7/packages/webdriverio/src/utils/interception/index.ts#L35)?

packages/webdriverio/src/commands/browser/url.ts
```diff
@@ -169,11 +191,23 @@ export async function url (
...
 
-    if (this.isBidi && path.startsWith('http')) {
+    /**
+     * Prefer BiDi navigation whenever we need BiDi-only features, or when the
+     * browser platform does not pay a large BiDi navigate tax. On macOS Chrome,
+     * `browsingContext.navigate` is much slower than classic `navigateTo` for a
+     * simple load (webdriverio#15481), so plain `url()` uses classic there.
+     * Use session `platformName` (remote browser), not the Node runner OS.
+     */
+    const useBidiNavigation = (
```
erwinheitzman: I would flip the check as it would simplify the conditional check and reasoning.

When MacOS and We don't require BiDi, then don't use it

packages/webdriverio/src/commands/browser/url.ts
```diff
@@ -282,13 +316,60 @@ export async function url (
...
+             */
+            return true
+        }
+
+        if (key === 'wait' && value !== 'complete') {
+            return true
+        }
+
+        /**
+         * `timeout` alone is a no-op without `networkIdle` (already BiDi above).
```
erwinheitzman: I don't understand why this comment is here

packages/webdriverio/src/commands/browser/url.ts
```diff
@@ -282,13 +316,60 @@ export async function url (
...
+ */
+export function requiresBidiNavigation (options: UrlCommandOptions = {}): boolean {
+    for (const [key, value] of Object.entries(options)) {
+        if (value === undefined) {
+            continue
+        }
+
+        if (!CLASSIC_SAFE_OPTION_KEYS.has(key)) {
+            /**
+             * headers / auth / onBeforeLoad today; any future option tomorrow.
```
erwinheitzman: Not sure what this comment is supposed to add

erwinheitzman (commented):
Added some feedback to check but nothing major. If you prefer to keep the conditional check as it is, that is fine as it's just a possible improvement to reasoning/readability 👍

packages/webdriverio/src/commands/browser/url.ts
wswebcreation: Explaining the last condition we would have in comments, but not in code. But will remove it 😆

packages/webdriverio/src/commands/browser/url.ts
wswebcreation: it's double, you're right, will remove it.

wswebcreation:
> Added some feedback to check but nothing major. If you prefer to keep the conditional check as it is, that is fine as it's just a possible improvement to reasoning/readability 👍

@erwinheitzman , tnx, wdyt regarding:

- merge to `main` or not
- see this, under these conditions, as a breaking change regarding the output? (undefined)

packages/webdriverio/src/commands/browser/url.ts
wswebcreation: Not sure what you mean here?

erwinheitzman:
@wswebcreation maybe we can return the Request when it is BiDi and return an object that would satisfy the any code that might rely on it instead? That would mean that any code should still work right?

wswebcreation:
> @wswebcreation maybe we can return the Request when it is BiDi and return an object that would satisfy the any code that might rely on it instead? That would mean that any code should still work right?

Hmm, we would then need to return an empty object @erwinheitzman , not sure if that is what would "not" break it. 

Just my take, and everyone can (dis)agree. This would only be breaking under the following conditions 
- Session is a BiDi session **AND**
- Platform is a Mac **AND** 
- they did **NOT** provide (valid) arguments **AND**
- The result of the URL is used for expectations or other use cases where they want to use the return data

I believe the last case is already rare, then in combination with not providing the arguments and the platform makes this in my opinion the risk of breaking for a lot of people very small. The fix would be to provide args (simple search)

wswebcreation:
..., seems to be driver-related.

PR #15492, opened 2026-08-11
docs(website): publish a machine-readable surface and un-hide the flowcharts

AutomatedTester:
## Proposed changes

A growing share of WebdriverIO's audience never reaches webdriver.io directly — they ask an assistant, and the assistant answers from whatever it can extract from our site. Today that extraction is lossier than it needs to be: everything has to be scraped out of rendered React HTML, with the navbar, announcement bar and OpenJS footer stripped off every page.

This came out of an audit of the site's build/publish pipeline. Worth saying up front: **the content is in good shape** — the hand-written corpus has zero raw HTML tables and zero JSX leakage. Every problem this PR addresses sits in the publish layer, not in the prose.

### Publish layer

- **`website/static/robots.txt`** — the site currently has none, so `webdriver.io/robots.txt` 404s. Its value here is mainly carrying the `Sitemap:` directive and pointing at the new machine-readable entry points; an absent `robots.txt` already means "allowed", so this is not about unblocking anything.
- **`@signalwire/docusaurus-plugin-llms-txt`** — emits `/llms.txt`, `/llms-full.txt`, and a clean Markdown twin for every page (`/docs/api/browser.md` and friends). There is no official Docusaurus plugin for this; alternatives are listed under *Further comments*.
- **Explicit `sitemap.lastmod`** — the sitemap currently uses preset defaults with no `lastmod`, so freshness signals are absent.

### Flowcharts

The four `docs/flowcharts/` pages were one sentence plus a `BrowserOnly` component, with the actual graphs living as template literals inside `CreateFlowcharts.tsx`. Four pages documenting the worker-process lifecycle, test execution and CLI command flow were invisible to anything without a JS runtime — and the graph source was unreviewable in a diff.

- All six mermaid graphs moved into ` ```mermaid ` fences, **verbatim**.
- `markdown.mermaid` enabled via `@docusaurus/theme-mermaid`, with light/dark themes configured to replace what `flowcharts.css` was doing with CSS variables.
- Deletes the `mermaid@8.5.1` `<script>` loaded from unpkg (2020-era, no SRI), `static/js/flowchart.js`, `CreateFlowcharts.tsx`, and the now-dead `flowcharts.css`. Side benefit: one less third-party CDN dependency on every page load.

## Types of changes

- [x] Polish (an improvement for an existing feature)
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
[...]

PR #15517, opened 2026-08-22
fix(docs): unbreak non-English docs builds after flowchart component removal

AutomatedTester:
### Proposed changes

Fixes the **Docs Deploy** failure introduced by #15492: [run 32444583896](https://github.com/webdriverio/webdriverio/actions/runs/32444583896/job/96661680704).

```
[ERROR] Error: Unable to build website for locale ar.
  Docusaurus static site generation failed for 4 paths:
  - "/ar/docs/flowcharts/createlocalworkerprocess"
  - "/ar/docs/flowcharts/highleveloverview"
  - "/ar/docs/flowcharts/testexecution"
  - "/ar/docs/flowcharts/wdiocommands"
    Error: Expected component `CreateFlowcharts` to be defined
```

#### Root cause

#15492 replaced `<CreateFlowcharts />` with inline ` ```mermaid ` fences in the English
flowchart docs and deleted the component plus its entry in `website/src/theme/MDXComponents.js`.

Translated docs are not in this repo. `scripts/docs-generation/downloadDocsTranslations.ts`
downloads [`webdriverio/i18n`](https://github.com/webdriverio/i18n) as a zipball at generate
time, and those pages are snapshots of older English source. **All 15 non-English locales
still ship the four flowchart pages containing the literal JSX tag** (60 files), so every one
of them now fails MDX compilation. `ar` is simply first alphabetically — the build aborts there.

The English build passes, which is why nothing showed up on #15492.

#### Impact

The last successful docs deploy was **2026-08-12**, before #15492 merged on 08-18. Production
still serves that build, so `/llms.txt` and `/robots.txt` currently 404 and no docs change
merged since 08-12 is live. The site is stale, not broken.

#### Fix 1 — `downloadDocsTranslations.ts`

Extends `applyTranslationFixes()`, whose documented purpose is exactly this ("*Add entries here
whenever a doc restructure breaks translated pages*", alongside two existing precedents). It
splices the mermaid body from the corresponding English page in place of the JSX tag, keyed on
frontmatter `id`.

- The English markdown stays the **single source of truth** — no duplicated graph text.
- The diagrams were never translated (they were rendered from a shared React component), so
  nothing is lost.
- Translated frontmatter and intro prose are preserved verbatim.
- An unmapped `id` **throws** rather than silently leaving the tag in place.

#### Fix 2 — `test-docs.yml`, a docs build on PR CI

The site was previously only ever built by `deploy.yml`, which is `workflow_dispatch` only.
Config, plugin and MDX regressions therefore could not surface on a PR — which is precisely how
#15492 went green and still broke the deploy.

Two details worth reviewing:

- It builds **`ar` as well as `en`**, deliberately. An `en`-only build cannot catch this class
  of failure.
- Both locales go to a **single** `docusaurus build` invocation. Docusaurus disables baseUrl
  localization when exactly one `--locale` is passed (`isAutomaticBaseUrlLocalizationDisabled`),
  so two separate build steps would each emit at the site root and the second would silently
  overwrite the first.

It also asserts `robots.txt`, `llms.txt`, `llms-full.txt` and sitemap `<lastmod>` are present,
so #15492's acceptance criteria are now enforced somewhere.

#### Verification

The changed function was run verbatim against **real** translated files fetched from
`webdriverio/i18n` (`ar`, `zh`, `de` — 12 genuinely stale files). 74/74 checks passed: JSX tag
removed, mermaid fence injected, fences balanced, translated frontmatter and intro
byte-preserved, all 4 English diagrams resolved by id (including `wdiocommands` carrying all
three), second run a no-op, and a descriptive throw on an unknown id.

**Not verified locally:** the full 16-locale build, and `tsc`/`eslint`. This PR's own new
`Docs Build` job is the real test of both halves — it should self-test here.

#### Follow-up (not in this PR)

The proper fix is to update those 60 files in `webdriverio/i18n` to use mermaid fences directly,
so translators own the section headings and node labels. Once that lands, the patch added here
can be deleted.

### Types of changes

### Checklist

### Reviewers

@webdriverio/project-committers

### DIFF
--- a/.github/workflows/test-docs.yml
+++ b/.github/workflows/test-docs.yml
@@ -1,0 +1,51 @@
+# Builds the documentation website on pull requests.
+#
+# The docs site is otherwise only ever built by `deploy.yml`, which is manually
+# dispatched — so a broken Docusaurus config, plugin or MDX page could only be
+# discovered at deploy time.
+#
+# This deliberately builds a non-English locale as well as `en`. Translated pages are
+# downloaded from the `webdriverio/i18n` repo at generate time and lag behind the
+# English source, so an `en`-only build cannot catch a translation referencing an MDX
+# component that no longer exists.
+name: Docs Build
+
+on:
+  workflow_call:
+
+jobs:
+  docs_build:
+    name: build
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - name: ⬇️ Checkout Code
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+
+      - name: 🧑‍🔧 Get Core Dependencies
+        uses: ./.github/workflows/actions/get-core-dependencies
+
+      - name: 📖 Generate Docs
+        run: pnpm run docs:setup && pnpm run docs:generate
+        env:
+          # required by scripts/docs-generation/downloadDocsTranslations.ts to fetch
+          # translations from the public webdriverio/i18n repo; the default read-only
+          # token is sufficient, including for pull requests from forks
+          GITHUB_AUTH: ${{ secrets.GITHUB_TOKEN }}
+
+      # Both locales must be passed to a single invocation: Docusaurus disables baseUrl
+      # localization when exactly one --locale is given, so two separate builds would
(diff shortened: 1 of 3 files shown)


christian-bromann (approved):
👍

Issue #15472, opened 2026-08-07
[🐛 Bug]: element.scrollIntoView() wheel Action puts scroll deltas in x/y (not deltaX/deltaY) resulting in "move target out of bounds"


wswebcreation:
### Have you read the Contributing Guidelines on issues?

- [x] I have read the [Contributing Guidelines on issues](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md#reporting-new-issues).

### WebdriverIO Version

latest

### Node.js Version

24.12.0

### Mode

WDIO Testrunner

### Which capabilities are you using?

```typescript
{
  browserName: 'chrome',
  // Chrome 151.0.7922.76 on macOS Darwin 25.5.0 arm64
}
```

### What happened?

During a performance review of a view commands, I saw that on desktop, `element.scrollIntoView()` uses the WebDriver wheel Actions API, then falls back to in-page Element.scrollIntoView on failure (after 3 retries.
For off-screen (or far) elements, the Actions call consistently fails with `move target out of bounds`. WDIO then logs a warn and falls back to the Web API. Scrolling still “works” for the user, but every call pays, layout prep + failed performActions + fallback execute.

**Root cause (in packages/webdriverio/src/commands/element/scrollIntoView.ts):**
Computed scroll amounts are passed as wheel aim coordinates x/y, while deltaX/deltaY stay at the defaults (0):

```ts
.await browser.action('wheel')
  .scroll({ duration: 0, x: deltaX, y: deltaY, origin: this })
  .perform()
```

Per WDIO’s own ScrollParams (packages/webdriverio/src/utils/actions/wheel.ts):

- x / y — starting coordinates relative to origin
- deltaX / deltaY — scroll amount

In my case Chrome is asked to aim the wheel thousands of px away from the element (often outside the viewport)  which results in `move target out of bounds`. Even if aim were valid, scroll amount would still be 0.

**Related (secondary) issues in the same function:**
- Before calculating deltas, current `window.scrollX/Y` are overwritten with element/viewport values when the element is outside the viewport.
- Preparations uses calls (getElementRect + getWindowSize + execute for scroll position) before Actions, can be one execute returning { elemRect, viewport, scroll } (optimization).

### What is your expected behavior?

Wheel Actions should receive scroll amounts in deltaX/deltaY, with aim at the element (x: 0, y: 0 with origin: element), e.g.:

```ts
await browser.action('wheel')
  .scroll({
    duration: 0,
[...]

PR #15533, opened 2026-08-27
fix(webdriverio): restore nested/Shadow DOM scrolling in scrollIntoView

### LINKED ISSUE #15521: element click intercepted after 9.31.x scrollIntoView wheel-origin change (#15473) with nested/Shadow DOM scroll containers
## Environment

- webdriverio / @wdio/cli / @wdio/local-runner: 9.31.1 (also reproduced on 9.31.2)
- Previously working version: webdriverio/@wdio/* 9.30.1
- OS: Windows 11
- Browser: Chrome 151.0.7922.138
- Framework: @wdio/cucumber-framework, runner: local

Relevant wdio config (trimmed to what matters):
- runner: local
- hostname: 127.0.0.1, port: 4444, path: /wd/hub (local Selenium Grid)
- framework: cucumber
- waitforTimeout: 10000 (59000 in CI), waitforInterval: 1200
- connectionRetryTimeout: 120000, connectionRetryCount: 3
- capabilities: chrome, 'wdio:enforceWebDriverClassic': true (classic WebDriver protocol, not BiDi), acceptInsecureCerts: true
- goog:chromeOptions args: --window-size=1980,1080 --disable-gpu --disable-dev-shm-usage --no-sandbox --disable-backgrounding-occluded-windows --disable-search-engine-choice-screen (no --headless)
- cucumberOpts: timeout 99999999, retry 2, retryTagFilter @retry

## What happened

After upgrading from webdriverio/@wdio/* 9.30.1 to 9.31.1, a previously 100%-passing step in our real app (a Lit/Shadow-DOM based data table with checkboxes) started failing 100% reproducibly with:

element click intercepted: WebDriverError: element click intercepted: Element is not clickable at point (1142, 1198)

The click target is a checkbox inside a shadow-DOM table row, reached via element.scrollIntoView({ block: 'center', inline: 'center' }) followed by element.click() (our own custom waitAndClick command, which already has a scroll-and-retry fallback for click interception - it retries with more aggressive scrolling, including walking up parent containers, but still fails at the exact same coordinate every time).

## A/B isolation (confirmed the version bump is the cause)

- Reverting ONLY webdriverio/@wdio/* back to 9.30.1 (nothing else changed - same app build, same test code, same DOM): the exact same scenario passes 100%, no click-intercepted error at all.
- Re-applying 9.31.1: fails 100% reproducibly again.
- Toggled back and forth twice to confirm.

## Suspected root cause

Comparing packages/webdriverio/src/commands/element/scrollIntoView.ts before/after PR #15473 (https://github.com/webdriverio/webdriverio/pull/15473/files):

Before (9.30.1):
browser.action('wheel').scroll({ duration: 0, x: deltaX, y: deltaY, origin: this }).perform()

After (9.31.x):
browser.action('wheel').scroll({ duration: 0, x: 0, y: 0, deltaX, deltaY }).perform()

Before, the wheel-scroll action originated at the target element itself (origin: this), so the synthetic wheel event was always dispatched over whatever scrollable ancestor actually contains the element - correctly scrolling nested/shadow-DOM scroll containers. After the change, the wheel action no longer has an origin tied to the element and is dispatched at a fixed point (x: 0, y: 0); if that point isn't over the specific scrollable container that holds the target element (very likely in apps with nested/shadow-DOM scroll regions, sticky headers, or multiple independently-scrollable panels), the scroll may not move the intended container at all, or may scroll an unrelated one - leaving the element under a fixed/sticky overlay (a bottom action toolbar, in our case) at click time.

## Attempt at a minimal reproduction (not fully conclusive - asking for help)

I tried to build a minimal public HTML reproduction (Shadow DOM host with an internal overflow-y: auto container + a position: fixed bottom bar, mimicking our real table + sticky toolbar), comparing element.scrollIntoView({block:'center'}) + element.click() on 9.30.1 vs 9.31.1/9.31.2. Findings:

- For an element that is literally the last item in its scrollable container (no more room to scroll further), the click fails on both 9.30.1 and 9.31.x identically - that's expected/correct behavior, not a regression.
- For every other row I tried (with genuine room to scroll clear of the fixed overlay), Chrome's own native "scroll into view before click" step (mandated by the WebDriver spec at the Element Click endpoint) ended up compensating for whatever our scrollIntoView() call did or didn't do, so the click still succeeded in my simplified page on both versions - meaning I have not yet isolated the exact extra condition (probably a specific combination of nested scrollable ancestors / sticky headers / element sizing) that lets the bug surface in our real app but not in my simplified reproduction.

I'm continuing to try to narrow this down, but wanted to open this now given we have a 100% reproducible real-world A/B regression tied directly to the scrollIntoView change in #15473. Happy to share more details from our real app's DOM structure (can't share the app itself, it's a private product) or try any repro shape you suspect might trigger it - e.g. specifically targeting the "skip native auto-scroll because bounding box is nominally within viewport" heuristic mentioned in the WebDriver spec for Element Click, which is my current best theory for why our real case differs from my simplified attempts.

mccmrunal:
PR #15473 fixed the "move target out of bounds" bug (#15472) by moving scroll amounts from x/y into deltaX/deltaY, but as a side effect also dropped the wheel action's `origin: this`, replacing it with a fixed viewport-relative dispatch point. That broke scrolling through nested scroll containers and Shadow DOM: a wheel event dispatched at a static point only affects whatever scrollable ancestor happens to be under that point, not the element's actual one.

Restores `origin: this` for the wheel action, but as a two-phase attempt (zero-delta probe + real delta) to avoid re-introducing the double-scroll overshoot that motivated removing it. When the element is already visible, skips straight to a single wheel action with the real delta.

Not every WebDriver implementation resolves the origin-based pre-scroll the same way (Chromium does; geckodriver does not), so the outcome is verified afterward rather than trusted: `isPainted`, a real hit-test at the element's own center point via `elementsFromPoint()` (recursing into open shadow roots, which it doesn't pierce on its own), replaces window-bounds arithmetic as the ground truth for both the fast-path decision and the final success check - geometry alone can't tell "visible" from "clipped by a nested container" or "covered by an unrelated overlapping element". Falls back to `Element.scrollIntoView()` when it isn't.

Fixes #15521

## Proposed changes

## Types of changes

## Checklist

## Backport Request

## Further comments


mccmrunal:
@wswebcreation  can you review this  ? Actually the issue it tries to resolve is side effect of your earlier PR

wswebcreation (approved):
LGTM, but the same with my PR, a lot of code to fix this 🤣

dprevost-LMI:
@mccmrunal, sounds like we can merge?

mccmrunal:
> @mccmrunal, sounds like we can merge?

merged

PR #15541, opened 2026-08-28
fix(webdriverio): avoid expensive findElements call in isExisting on mobile

### LINKED ISSUE #7580: [🐛 Bug]: isExisting calls findElements (multiple), a potentially expensive op

zynp-KC:
## Proposed changes

`isExisting` (and therefore `waitForExist`, which reuses it) always used `$$` / `findElements`. On mobile this can be very expensive — as reported in #7580, large scrollable lists on iOS can take 30+ seconds, because `findElements` materializes the whole list.

On mobile this PR switches `isExisting` to a single `$` / `findElement` lookup and checks for a resolved `elementId`, following the `isMobile`-conditional direction discussed in #7580.

Web keeps using `findElements` on purpose. As @BorisOsipov pointed out in the issue, switching web to `findElement` would make every lookup of a non-existent element pay the full `implicitWait` timeout. Scoping the change to `isMobile` targets exactly the case where `findElements` is the expensive path and `implicitWait` is rarely used, without changing web behavior.

I also reviewed `waitForExist`'s post-wait refetch: its non-indexed branch already uses `$` (single), and the `this.index !== undefined` branch keeps `$$` because resolving a positional selector (`$$(...)[n]`) inherently needs the full list. That `findElements` call is intentional and left as-is.

Fixes #7580

## Types of changes

## Checklist

## Backport Request

### DIFF
--- a/packages/webdriverio/tests/commands/element/isExisting.test.ts
+++ b/packages/webdriverio/tests/commands/element/isExisting.test.ts
@@ -42,6 +42,20 @@ describe('isExisting test', () => {
 
     })
 
+    it('should use findElement (single) on mobile', async () => {
+        const mobileBrowser = await remote({
+            baseUrl: 'http://foobar.com',
+            capabilities: {
+                platformName: 'iOS'
+            }
+        })
+        const elem = await mobileBrowser.$('#foo')
+        await elem.isExisting()
+        // @ts-expect-error mock implementation
+        expect(vi.mocked(fetch).mock.calls[2][0]!.pathname)
+            .toBe('/session/foobar-123/element')
+    })
+
     afterEach(() => {
         vi.mocked(fetch).mockClear()
     })
--- a/packages/webdriverio/tests/commands/element/isExisting.test.ts
+++ b/packages/webdriverio/tests/commands/element/isExisting.test.ts
@@ -50,10 +50,15 @@ describe('isExisting test', () => {
             }
         })
         const elem = await mobileBrowser.$('#foo')
+
+        // clear the mock so we only assert isExisting's own request,
+        // not the findElement from $('#foo') above
+        vi.mocked(fetch).mockClear()
+
         await elem.isExisting()
-        // @ts-expect-error mock implementation
-        expect(vi.mocked(fetch).mock.calls[2][0]!.pathname)
-            .toBe('/session/foobar-123/element')
+        expect(vi.mocked(fetch)).toHaveBeenCalledTimes(1)
+        const url = vi.mocked(fetch).mock.calls[0][0] as URL | string
+        expect(url.toString().endsWith('/element')).toBe(true)
     })
 
     afterEach(() => {
--- a/packages/webdriverio/tests/commands/element/isExisting.test.ts
+++ b/packages/webdriverio/tests/commands/element/isExisting.test.ts
@@ -57,8 +57,10 @@ describe('isExisting test', () => {
 
         await elem.isExisting()
         expect(vi.mocked(fetch)).toHaveBeenCalledTimes(1)
-        const url = vi.mocked(fetch).mock.calls[0][0] as URL | string
-        expect(url.toString().endsWith('/element')).toBe(true)
+        expect(vi.mocked(fetch)).toHaveBeenCalledWith(
+            expect.toSatisfy((url: URL | string) => url.toString().endsWith('/element')),
+            expect.any(Object)
+        )
     })
 
     afterEach(() => {
--- a/packages/webdriverio/src/commands/element/isExisting.ts
+++ b/packages/webdriverio/src/commands/element/isExisting.ts
@@ -61,6 +61,22 @@ export async function isExisting (this: WebdriverIO.Element) {
         )
     }
 
+    /**
+     * on mobile, `findElements` can be very expensive (e.g. large scrollable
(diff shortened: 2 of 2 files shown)


dprevost-LMI:
🤔 I wonder if having a per `findElement` configuration which would disable `implicitWait` would make a more cross-platform compatible fix. 
Unfortunately, after a bit of digging, that sounds like going against the W3C WebDriver specification 😒.

packages/webdriverio/tests/commands/element/isExisting.test.ts
```diff
@@ -42,6 +42,20 @@ describe('isExisting test', () => {
...
+            baseUrl: 'http://foobar.com',
+            capabilities: {
+                platformName: 'iOS'
+            }
+        })
+        const elem = await mobileBrowser.$('#foo')
+        await elem.isExisting()
+        // @ts-expect-error mock implementation
+        expect(vi.mocked(fetch).mock.calls[2][0]!.pathname)
+            .toBe('/session/foobar-123/element')
```
dprevost-LMI: @greptileai, can we use toHaveBeenCalledWith here instead of `.mock.calls[2][0]!.pathname`

zynp-KC: @dprevost-LMI you're right, the indexed lookup was brittle. I've switched it to clear the mock right before `isExisting()` and check the call count and endpoint instead

I kept the `mockClear()` rather than a plain `toHaveBeenCalledWith` because, as @greptileai noted, the `$('#foo')` above already does a `findElement` without clearing first, the assertion would pass even if `isExisting` regressed to `findElements`. Also confirmed it fails when I revert the fix

dprevost-LMI: @greptileai, instead of 
```ts
     const url = vi.mocked(fetch).mock.calls[0][0] as URL | string
        expect(url.toString().endsWith('/element')).toBe(true)
```        
Is there a better way to do something similar but leveraging `toHaveBeenCalledWith`

dprevost-LMI:
@greptile could we implement something like https://github.com/webdriverio/webdriverio/pull/15541#issuecomment-5462730517?

zynp-KC:
@dprevost-LMI switched it to toHaveBeenCalledWith. Sharing what I ran into in case it's useful.

Matching the exact fetch options object (expect.any(Object) or objectContaining({ method: 'POST' })) kept failing, because the second arg carries an AbortSignal and a Headers instance that the matcher wouldn't accept cleanly. What worked was asserting the URL with a toSatisfy matcher:

```
expect(vi.mocked(fetch)).toHaveBeenCalledWith(
    expect.toSatisfy((url: URL | string) => url.toString().endsWith('/element')),
    expect.any(Object)
)
```
Kept the mockClear() and toHaveBeenCalledTimes(1). Confirmed it still fails when I revert the fix.

dprevost-LMI:
@greptileai, can you add a new command to `findElementNoWait` and add it to the `IMPLICIT_WAIT_EXCLUSION_LIST` exclusion in https://github.com/webdriverio/webdriverio/blob/main/packages/webdriverio/src/middlewares.ts#L8 so it works on mobile and web platforms?

PR #15544, opened 2026-08-29
fix: encode file URLs for absolute plugin paths

### LINKED ISSUE #15543: [🐛 Bug]: Absolute plugin paths containing URL-reserved characters fail to load

### WebdriverIO Version

v9.31.4

### Node.js Version

v18.20.0

### Mode

WDIO Testrunner

```typescript

```

### What happened?

initializePlugin() accepts absolute filesystem paths for custom services, reporters, and frameworks and converts them into file:// URLs before importing them.

The current implementation builds the file URL manually. As a result, URL-reserved characters in file or directory names are not percent-encoded.

For example, this valid Windows path:

C:\work\plugins\plugin%fixture.mjs

is converted to:

file:///C:/work/plugins/plugin%fixture.mjs

The raw % is interpreted as the beginning of a URL percent-escape sequence. Because %fi is not a valid escape sequence, module resolution fails with an error such as:

URI malformed

and WebdriverIO cannot load the plugin.

The same problem can occur with other URL-reserved characters such as #, ?, and spaces.

Windows drive paths recognized by `isAbsolute()` should preserve their Windows file URL representation independently of the host OS, while URL-reserved characters are encoded correctly.

C:\path\plugin%fixture.js
→ file:///C:/path/plugin%25fixture.js

### How to reproduce the bug.

Create a custom service, reporter, or framework whose filename contains a URL-reserved character, for example:
C:\work\plugins\plugin%fixture.mjs
Configure WebdriverIO to load the plugin using its absolute filesystem path.
The path is passed through initializePlugin().
WebdriverIO converts it to:
file:///C:/work/plugins/plugin%fixture.mjs
Module resolution fails because the % is not percent-encoded.

A minimal unit-level reproduction would be:

const pluginPath = path.resolve(
    __dirname,
    '__fixtures__',
    'plugins',
    'plugin%fixture.mjs'
)

await initializePlugin(pluginPath, 'service')

### Relevant log output

```typescript
raw URL: file:///C:/Users/gitfo/Desktop/open source_proj/webdriverio/packages/wdio-utils/tests/__fixtures__/plugins/plugin%fixture.mjs
URIError: URI malformed
```

### Code of Conduct

CAOShurong:
## Proposed changes

- build native absolute plugin URLs with Node's pathToFileURL() so reserved characters are encoded correctly
- preserve Windows drive-path support when WebdriverIO resolves those paths on non-Windows hosts
- add a regression fixture whose filename contains a literal percent sign

Fixes #15543.

## Verification

- corepack pnpm@10.34.5 -r --filter=@wdio/compiler run build -p @wdio/logger -p @wdio/types -p @wdio/globals -p @wdio/protocols -p @wdio/repl`n- corepack pnpm@10.34.5 exec vitest --config vitest.config.ts --run packages/wdio-utils/tests (14 files, 281 tests, no type errors)
- corepack pnpm@10.34.5 exec eslint packages/wdio-utils/src packages/wdio-utils/tests`n- git diff --check`n
AI assistance was used to help implement and validate this change; the behavior was reproduced against a real percent-sign fixture and the resulting patch was reviewed and tested locally.

### DIFF
--- a/packages/wdio-utils/src/initializePlugin.ts
+++ b/packages/wdio-utils/src/initializePlugin.ts
@@ -1,4 +1,6 @@
 import type { Services } from '@wdio/types'
+import { platform } from 'node:os'
+import { pathToFileURL } from 'node:url'
 
 import { safeImport, isAbsolute, REG_EXP_WINDOWS_ABS_PATH, SLASH } from './utils.js'
 
@@ -57,12 +59,17 @@ function ensureFileURL(path:string) {
 
     // Windows drive path
     if (REG_EXP_WINDOWS_ABS_PATH.test(path)) {
-        return `${FILE_PROTOCOL}/${path.replace(/\\/g, '/')}`
+        if (platform() === 'win32') {
+            return pathToFileURL(path).href
+        }
+
+        const pathname = path.replace(/\\/g, '/').split('/').map(encodeURIComponent).join('/')
+        return `${FILE_PROTOCOL}/${pathname}`
     }
 
     // Unix absolute path
     if (path.startsWith(SLASH)) {
-        return `${FILE_PROTOCOL}${path}`
+        return pathToFileURL(path).href
     }
 
     return path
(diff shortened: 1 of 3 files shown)


dprevost-LMI:
@CAOShurong do you mind signing the CLA?

dprevost-LMI:
Duplicate of https://github.com/webdriverio/webdriverio/pull/15596

PR #15551, opened 2026-08-31
fix(wdio-allure-reporter): write results incrementally instead of only at runner end

### LINKED ISSUE #15529: [🐛 Bug]: wdio-allure-reporter only save result at the end of the runner

### WebdriverIO Version

latest

### Node.js Version

latest

### Mode

WDIO Testrunner

```typescript

```

### What happened?

Hello,
Since september 2025 and Allure 3, the allure files inside the allure-results folder (attachment.txt, result.json, container.json) are only written at the runner end with `await state.processRuntimeMessage()` ➡️ [here](https://github.com/webdriverio/webdriverio/blob/2b9721d052ac24207e8628c65a63c3a0de122a2a/packages/wdio-allure-reporter/src/reporter.ts#L529)

```
    async onRunnerEnd(_runner: RunnerStats): Promise<void> {
        this._isFlushing = true
        try {
            for (const [cid, state] of this.allureStatesByCid) {
                await state.processRuntimeMessage()
                this.allureStatesByCid.delete(cid)
            }
        } finally {
            if (this._options.addConsoleLogs) {
                process.stdout.write = this._originalStdoutWrite
            }
            this._isFlushing = false
        }
        this._allureRuntime.writeEnvironmentInfo()
    }
```
The problem is, if the runner crash during the test execution, we loose 100% of the logs.

🔗[Discord message](https://discord.com/channels/1097401827202445382/1097674067601010709/1541833611592990720)

In the best case scenario, I think we need a setting to allow the user to choose when the files are created.

**Basic solution:**

* In the reporter entry:
```
saveAllureResultOnTestEnd: true
```

* In reporter.ts add:
```
  async onTestEnd(test) {
    if (this._options?.saveAllureResultOnTestEnd === true) {
      this._isFlushing = true;
      try {
        for (const [cid, state] of this.allureStatesByCid) {
          await state.processRuntimeMessage();
          this.allureStatesByCid.delete(cid);
        }
      } finally {
        if (this._options.addConsoleLogs) {
          process3.stdout.write = this._originalStdoutWrite;
        }
        this._isFlushing = false;
      }
    }
  }
```
**Advance solution:**
Maybe it's a better idea to have an option with mode to save:
- "after each test",
- after each suite",
- "after each..."

### How to reproduce the bug.

a) Add a stoppoint at the beginning of "onRunnerEnd"
b) Run your test
c) At the stoppoint, check the content of allure-results
➡️ The folder is empty
d) Continue after the stoppoint
➡️ Files are created

### Relevant log output

```typescript
N/A
```

### Code of Conduct

devangpratap:
## Proposed changes

Fixes #15529.

`@wdio/allure-reporter` buffers every runtime message and only calls `AllureReportState.processRuntimeMessage()` once, in `onRunnerEnd`. Nothing reaches the results directory until the whole run is over, so if the runner dies mid-run you lose every result, not just the test that was in flight.

The blocker was that `processRuntimeMessage()` was not safe to call more than once: it always replayed `this.messages` from index 0, and it finalised the run at the end by writing the current test and unwinding the scope stack.

This makes it resumable and lets the reporter flush as the run goes:

- `AllureReportState` keeps a cursor (`_processedIndex`) and starts each pass from it, so a message is never applied twice. The cursor points at the message being handled, so a pass that throws leaves it there and the next pass retries that message instead of replaying the batch.
- `processRuntimeMessage(final = false)` applies buffered messages without the finalisation tail. The default stays `true`, so the runner-end call still finalises the run exactly as it does today.
- `AllureReporter` flushes once a new `allure:test:start` has been buffered. That is the point at which the previous test gets written, because a test's own scope has to stay open past its `test:end` for `"after each"` hooks to attach to it.
- Flushes are chained so they cannot overlap, and `onRunnerEnd` is counted as a flush itself, so an incremental flush completing while `onRunnerEnd` awaits the chain cannot flip `isSynchronised` back to `true` while the final pass is still writing.

The order in which messages are applied is unchanged, and so is the final content of the results directory. The only difference is when the loop runs.

One boundary worth naming: because a test is written when the *next* one starts, a crash costs you the test that was in flight, and a spec with a single test still only produces its result at runner end. Everything before the in-flight test is on disk.

I did not add the `saveAllureResultOnTestEnd` option suggested in the issue. Flushing unconditionally leaves the reporter's output identical to what it produces today, so there is nothing to opt out of.

## Types of changes

## Checklist

## Backport Request

## Further comments

Three tests were added, and the two regression tests both fail if the `src` changes are reverted:

- `tests/state.test.ts`: two mid-run flushes over a suite with one finished and one started test, asserting the finished test is on disk before the run is finalised, that a flush with nothing new is a no-op, and that the finished test is not written twice by the final pass.
- `tests/suite.test.ts`: the same at reporter level, also asserting the reporter reports `isSynchronised === false` while a flush is in flight.
- `tests/suite.test.ts`: an equivalence check that a run flushed per test and a run flushed only at the end produce the same results.

`npx vitest run ./packages/wdio-allure-reporter/tests` passes (200 tests, no type errors) and eslint is clean on the package.

### DIFF
--- a/packages/wdio-allure-reporter/src/reporter.ts
+++ b/packages/wdio-allure-reporter/src/reporter.ts
@@ -126,6 +126,8 @@ export default class AllureReporter extends WDIOReporter {
     private _consoleOutput = ''
     private _originalStdoutWrite: typeof process.stdout.write
     private _isFlushing = false
+    private _flushChain: Promise<void> = Promise.resolve()
+    private _pendingFlushes = 0
     private _cid?: string
 
     private _testPlan: TestPlanV1 | undefined
@@ -305,9 +307,44 @@ export default class AllureReporter extends WDIOReporter {
         this._pushRuntimeMessage({ type: 'allure:suite:end', data: {} })
     }
 
+    /**
+     * Write out everything buffered so far without finalising the run, so results of
+     * completed tests reach the results directory while the run is still going. Flushes are
+     * chained so they cannot overlap.
+     */
+    private _flushIncrementally(): void {
+        const state = this._ensureState(this._currentCid())
+
+        this._beginFlush()
+        this._flushChain = this._flushChain
+            .then(() => state.processRuntimeMessage(false))
+            // the state's cursor stays on the message that threw, so the runner end pass
+            // retries it and rethrows if it still fails
+            .catch(() => {})
+            .then(() => this._endFlush())
+    }
+
+    private _beginFlush(): void {
+        this._pendingFlushes++
+        this._isFlushing = true
+    }
+
+    private _endFlush(): void {
+        this._pendingFlushes--
+        if (this._pendingFlushes === 0) {
+            this._isFlushing = false
+        }
+    }
+
     private _startTest(payload: { name: string; start: number; uuid?: string }): void {
         this._pushRuntimeMessage({ type: 'allure:test:start', data: payload })
         this._setTestParameters()
+        /**
+         * flush once the new test:start is buffered: a test is only written when the
+         * following test starts, because its own scope has to stay open past its test:end
+         * for "after each" hooks to attach to
+         */
+        this._flushIncrementally()
     }
 
     private _endTest(payload: {
@@ -523,8 +560,13 @@ export default class AllureReporter extends WDIOReporter {
     }
 
     async onRunnerEnd(_runner: RunnerStats): Promise<void> {
-        this._isFlushing = true
+        /**
+         * counted like any other flush, so an incremental flush finishing while we await it
+         * cannot report the reporter as synchronised while this pass is still writing
+         */
+        this._beginFlush()
         try {
+            await this._flushChain
             for (const [cid, state] of this.allureStatesByCid) {
                 await state.processRuntimeMessage()
                 this.allureStatesByCid.delete(cid)
@@ -533,7 +575,7 @@ export default class AllureReporter extends WDIOReporter {
             if (this._options.addConsoleLogs) {
                 process.stdout.write = this._originalStdoutWrite
             }
-            this._isFlushing = false
+            this._endFlush()
         }
         this._allureRuntime.writeEnvironmentInfo()
     }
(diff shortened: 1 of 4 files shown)


devangpratap:
The one red check, `Testrunner E2E Tests / Testrunner Tests (windows-latest)`, is failing on main too and is unrelated to this PR.

main run 33394782074 (today, 13:02 UTC, before this PR was opened):

```
Error in "main suite 1.moveTo tests.moves to position x,y outside of iframe when passing the arguments undefined"
Expected: 494
Received: 80
Spec Files: 5 passed, 1 failed, 6 total
```

This PR:

```
Error in "main suite 1.moveTo tests.moves to position x,y outside of iframe when passing the arguments {"xOffset":10}"
Expected: 494
Received: 90
Spec Files: 5 passed, 1 failed, 6 total
```

Same job, same test at `e2e/wdio/headless/test.e2e.ts:261`, same expected value, same 5/1 split. The received value moves around because the assertion races the pointer position, but it does not reach 494. Testrunner E2E has failed on main in 4 of the last 5 non-passing runs of the Test workflow, windows-latest twice and macos-15 twice.

Nothing here touches `e2e/`, and every e2e config uses `reporters: ['spec']`, so the allure reporter is not loaded in that suite. The other 70 checks are green, including the full XVFB matrix.

@dprevost-LMI @mccmrunal would one of you be able to take a look?

devangpratap:
@P-Courteille validated this against his original failure mode — [issue comment](https://github.com/webdriverio/webdriverio/issues/15529#issuecomment-5494626471). He installed the PR build from pkg.pr.new, ran a spec with several tests, and killed the Appium server before the spec finished. Result files were written progressively as the run went, and the report generated from them was correct.

The one red check is `Testrunner Tests (windows-latest)`, failing on `moveTo` "moves to position x,y outside of iframe" with `Expected: 494 / Received: 90` at `e2e/wdio/headless/test.e2e.ts:261`. Unrelated to this change — it's a pointer-position assertion and this PR only touches when the allure reporter flushes buffered messages.

devangpratap:
@dprevost-LMI could you take a look when you get a chance? I merged main in and all checks are green now, including the Windows testrunner job that was failing before for an unrelated reason. The allure reporter tests still pass (200). P-Courteille already validated the fix against the original crash in #15529.

christian-bromann (approved):
👍  Thanks a lot!

PR #15559, opened 2026-09-04
fix(webdriver): keep new driver PID after reloadSession

harsha509:
## Proposed changes

When `reloadSession` is called with a new `browserName` on a local session, a new driver is spawned and its PID is written to `wdio:driverPID`. The code then restored the previous session's PID over it. The new driver was never killed on `deleteSession` and was left running.

This change restores the previous PID only when no new driver was spawned. A unit test with distinct PIDs for the original and reloaded driver is added.

## Types of changes

## Checklist

## Backport Request


dprevost-LMI:
@greptile is there a e2e we could add in `e2e/wdio/headless/test.e2e.ts` to reproduce the issue?

PR #15562, opened 2026-09-04
fix(wdio-utils): track dialog listener lifecycle

### LINKED ISSUE #15558: [🐛 Bug]: `browser.once('dialog', ...)` does not disable default auto dismiss

### WebdriverIO Version

9.30.0

### Node.js Version

v24.19.0

### Mode

WDIO Testrunner

(probably irrelevant)

```typescript
{
  browserName: 'firefox',
  browserVersion: 'stable_151.0.4',
  'wdio:geckodriverOptions': {
    geckoDriverVersion: '0.36.0',
  },
}
```

### What happened?

As mentioned in the [Dialog documentation](https://webdriver.io/docs/api/dialog/) by default there is an auto dismiss of dialogs unless you use `browser.on('dialog')`.

However, when using `browser.once('dialog', ...)` to only handle a dialog once (for example to prevent the listener from affecting other tests), then that does not disable the auto dismiss, and the listener is never called.

This seems related to this code:
https://github.com/webdriverio/webdriverio/blob/f9b0721f520ee61a62d811a13f350fb315f0d5c1/packages/wdio-utils/src/monad.ts#L304-L315

It only handles `on` and `off` but not `once` (and also neither `removeListener` nor `removeAllListeners` for enabling auto dismiss again).

Ideally `browser.once('dialog', ...)` should disable the auto dismiss as well, and enable it again once the listener was called (and if there are no other custom listeners?).

Alternatively, if this is the expected behavior, then maybe it would be good to mention this in the [Dialog documentation](https://webdriver.io/docs/api/dialog/).

### How to reproduce the bug.

Use the following test:
```typescript
import { browser } from '@wdio/globals'
import { describe } from 'mocha'

describe('dialog test', () => {
  it('test', async () => {
    browser.once('dialog', async (dialog) => {
      console.info('DEBUG dialog handler was called')
      return dialog.dismiss()
    })

    await browser.url(
      'https://developer.mozilla.org/en-US/docs/Web/API/Window/confirm',
    )
    await browser.switchFrame($('iframe[title="runner"]'))
    await $('button=Open new tab').click()
  })
})
```

:x: Bug: You will notice that no message was logged, and the custom dialog listener was not called.

If you change `once` to `on`, or if you additionally register a second listener with `on`, then the listener will be called.

### Relevant log output

_none_

### Code of Conduct

Fnine59:
## Proposed changes

Fixes #15558.

Track the number of registered `dialog` handlers in the WebDriver monad and emit dialog lifecycle events only when the count crosses between zero and non-zero. This keeps automatic dialog handling disabled while any listener remains, and restores it after the last listener is removed or a one-time listener fires.

The regression test covers `once`, `off`, `removeListener`, scoped and global `removeAllListeners`, and multiple concurrent dialog listeners.

## Types of changes

## Checklist

## Backport Request

## Further comments

Focused verification: `pnpm exec vitest --config vitest.config.ts --run packages/wdio-utils/tests/monad.test.ts`.

### DIFF
--- a/packages/wdio-utils/src/monad.ts
+++ b/packages/wdio-utils/src/monad.ts
@@ -61,25 +61,111 @@ export default function WebDriver(options: object, modifier?: Function, properti
 
     const mittInstance = mitt()
 
+    type EventHandler = (event: unknown) => void
+    const dialogListeners: Array<{ listener: EventHandler, registered: EventHandler }> = []
+
+    const emitDialogListenerTransition = (previousCount: number) => {
+        if (previousCount === 0 && dialogListeners.length > 0) {
+            mittInstance.emit('_dialogListenerRegistered')
+        } else if (previousCount > 0 && dialogListeners.length === 0) {
+            mittInstance.emit('_dialogListenerRemoved')
+        }
+    }
+
+    const addDialogListener = (listener: EventHandler, registered = listener) => {
+        const previousCount = dialogListeners.length
+        mittInstance.on('dialog', registered)
+        dialogListeners.push({ listener, registered })
+        emitDialogListenerTransition(previousCount)
+    }
+
+    const removeDialogListener = (listener?: EventHandler, registered?: EventHandler) => {
+        const previousCount = dialogListeners.length
+        let index = -1
+
+        for (let i = dialogListeners.length - 1; i >= 0; i--) {
+            const entry = dialogListeners[i]
+            if (registered ? entry.registered === registered : entry.listener === listener) {
+                index = i
+                break
+            }
+        }
+
+        if (index === -1) {
+            return
+        }
+
+        const [entry] = dialogListeners.splice(index, 1)
+        mittInstance.off('dialog', entry.registered)
+        emitDialogListenerTransition(previousCount)
+    }
+
+    const removeAllDialogListeners = () => {
+        const previousCount = dialogListeners.length
+        mittInstance.off('dialog')
+        dialogListeners.length = 0
+        emitDialogListenerTransition(previousCount)
+    }
+
     // Create EventEmitter-compatible interface
     const eventHandler = {
-        on: mittInstance.on.bind(mittInstance),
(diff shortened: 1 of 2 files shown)


dprevost-LMI:
@greptile, could you look into `website/docs` and see whether any documentation is worth upgrading? If so, can you propose an upgraded version?

dprevost-LMI:
@greptile, is the dialog autoDismiss not managed in DialogManager from `packages/webdriverio/src/session/dialog.ts`? Would it be more scoped and potentially less intrusive to disable and enable the flag `autoHandleDialog` (see https://github.com/webdriverio/webdriverio/blob/b60474af916440534d5ec016d6eef278630969b3/packages/webdriverio/src/session/dialog.ts#L89) when a `once` event is registered? If yes, can you propose a solution?

dprevost-LMI:
@greptile, following https://github.com/webdriverio/webdriverio/pull/15562#issuecomment-5552664228, does the current PR implement this correctly, or are changes needed? If changes are needed, provide the entire solution!

dprevost-LMI:
@greptile, if we change `autoHandleDialog` to be a count and wrap the `once` of the dialogue like below, would that make a less invasive solution (and assume we also handle also `removeListener` and `removeAllListeners` if needed? 
```
            if (eventCommand === 'once' && args[0] === 'dialog') {
                eventHandler.emit('_dialogListenerRemoved')
                wrappedListener = () => {
                    (args[1] as (event: unknown) => void)(args[0])
                    eventHandler.emit('_dialogListenerRemoved')
                }

            }
```

dprevost-LMI:
@Fnine59 Great job, I challenge Greptile to try to shift the job into `DialogManager`, but your solution remains the best one!

dprevost-LMI:
@greptile, should we update https://webdriver.io/docs/api/dialog/ to include something like 
> Dialogs are dismissed automatically, unless there is a `browser.on('dialog')` or `browser.once('dialog')` listener

Issue #15570, opened 2026-09-07
[🐛 Bug]: BiDi cached browsing context is never invalidated on contextDestroyed; classic commands mask it and switchToParentFrame() cannot recover


mccmrunal:
### Environment
- WebdriverIO: **9.30.1** (`webdriverio` standalone, no CLI/framework needed)
- Node.js: 24.15.0
- Chrome / chromedriver: 152.0.7977.76 / matching
- OS: Windows 11
- Protocol: WebDriver BiDi (`webSocketUrl: true`)

### What happens

`ContextManager` caches the current browsing-context id and passes it as `target.context` to
`script.callFunction` (`packages/webdriverio/src/session/context.ts`). When that context is
**destroyed by the page** rather than by a WebdriverIO command, the cache is never invalidated,
so every subsequent `browser.execute()` fails permanently:

```
WebDriver Bidi command "script.callFunction" failed with error:
  no such frame - Context 5D50FE58A74D41D74F22CEF6031213AF not found
```

Three things make this hard to diagnose and impossible to recover from in user land:

1. **The browser announces the destruction and WebdriverIO ignores it.** The constructor
   subscribes only to `browsingContext.navigationStarted`. I verified that
   `browsingContext.contextDestroyed` fires with exactly the id that is about to go stale —
   the information is already on the wire.

2. **Classic commands keep succeeding and mask the broken state.** `getTitle()` / `getUrl()`
   go over classic WebDriver against "whatever context is current", so they return normally
   while every BiDi script call fails. A health-check like
   `try { await browser.getTitle() } catch { /* re-anchor */ }` can therefore never detect this.

3. **`switchToParentFrame()` does not recover.** It is listed in `COMMANDS_REQUIRING_RESET`,
   but `#onCommand` returns early in its own branch:

   ```js
   #onCommand(event) {
     if (event.command === 'switchToParentFrame') {
       if (!this.#currentContext) return
       return this.#browser.browsingContextGetTree({}).then(...)   // <-- returns here
     }
     ...
     if (COMMANDS_REQUIRING_RESET.includes(event.command)) {
       this.#currentContext = undefined                            // <-- never reached
     }
   }
   ```

   And `findParentContext(#currentContext, contexts)` cannot find a parent for a context that
   no longer exists in the tree, so it returns without setting anything. The cache stays stale.

### Reproducible example

`npm i webdriverio@9.30.1 && node repro.mjs` — deterministic, asserts each claim, ~10 lines of output.

[...]

PR #15573, opened 2026-09-07
fix(webdriverio): recover owning window after nested frame destruction

lucasmariano003-wq:
## Proposed changes

When the current iframe is removed from inside another iframe, context recovery calls `switchToWindow()` with the surviving parent iframe ID. Chrome rejects that command with `no such window`. This still reproduces on the `main` baseline (`ba488fa`, v9.31.7), following the recovery work in #15477; it is related to the destroyed-frame scenario in #15570.

BiDi `browsingContext.getTree` deliberately omits `parent` on child entries. Find the top-level tree containing the surviving parent instead of following absent `parent` fields. The regression fixture now reflects the real response and includes an unrelated window before the owning window.

CI follow-up: the affected window/frame tests use a local HTTP fixture with Lighthouse enabled and preserve URL/title, handle and nested-frame checks. The branch is reconciled with upstream main, which includes EdgeDriver 6.3.1, and uses its exact valid lockfile. Frozen installation, the full build and full lint checks pass; 1,606 WebdriverIO/utils tests and both navigation cases pass on the final tree.

## Types of changes

## Checklist

- [x] I have added the necessary documentation (if appropriate; no public API changes)
- [x] I have added proper type definitions for new commands (if appropriate; no new commands)

## Backport Request

## Validation

- Corrected regression test fails before the fix: expected the owning window, received `frame-parent`.
- All WebdriverIO package tests: **1,248 passed, 1 skipped** across **181 files**; no type errors reported by the runner.
- ESLint on both changed files and `git diff --check` pass.
- WebdriverIO package compilation passes.
- Real headless Chrome **153.0.8010.12**: removing a nested iframe causes `switchToWindow(iframe-id)` / `no such window` before the fix; after the fix it targets the owning top-level window and `browser.execute()` returns the top-level document content. This used synthetic local `data:` / `srcdoc` pages, without external services.
- The complete published tree at `752bf31` was fetched back and matches the tested local tree.

The [BiDi serialization algorithm](https://w3c.github.io/webdriver-bidi/#get-the-navigable-info) serializes child entries with `include parent id` set to false, matching the observed Chrome response.

## Further comments

AI disclosure: OpenAI Codex performed the investigation, implementation, and local validation for this GitHub account. Maintainer review is pending.

Could a TSC member confirm whether this contribution qualifies under the [non-project-member contribution expense policy](https://github.com/webdriverio/webdriverio/blob/main/GOVERNANCE.md#expenses-from-non-project-members), and the approved amount and claim requirements if accepted?

### DIFF
--- a/packages/webdriverio/src/session/context.ts
+++ b/packages/webdriverio/src/session/context.ts
@@ -207,11 +207,10 @@ export class ContextManager extends SessionManager {
      */
     async #getTopLevelContext(contextId: string) {
         const { contexts } = await this.#browser.browsingContextGetTree({})
-        let context = this.findContext(contextId, contexts, 'byContextId')
-        while (context?.parent) {
-            context = this.findContext(context.parent, contexts, 'byContextId')
-        }
-        return context?.context
+        // Child entries in getTree omit parent, so find the root containing the context.
+        return contexts.find((context) =>
+            this.findContext(contextId, [context], 'byContextId')
+        )?.context
     }
 
     #onCommandResultBidiAndClassic(event: { command: string, result: unknown, body: unknown }) {
--- a/packages/webdriverio/tests/session/context.test.ts
+++ b/packages/webdriverio/tests/session/context.test.ts
@@ -192,17 +192,20 @@ describe('ContextManager', () => {
         expect(manager.getCurrentWindowHandle()).toBeUndefined()
     })
 
-    it('switches to the top-level context when a nested child frame is destroyed (regression test)', async () => {
+    it('switches to the top-level context when a nested child frame is destroyed and child nodes omit parent', async () => {
         const wid = process.env.WDIO_UNIT_TESTS
         delete process.env.WDIO_UNIT_TESTS
         const stub = createBrowserStub({ isBidi: true } as any)
         const browser = stub.browser
         ;(browser as any).browsingContextGetTree.mockResolvedValue({
             contexts: [{
+                context: 'other-window', parent: null, children: [], url: '',
+                clientWindow: 'window-2', originalOpener: null, userContext: 'default'
+            }, {
                 context: 'context-1', parent: null, url: '', clientWindow: 'window-1',
                 originalOpener: null, userContext: 'default',
                 children: [{
-                    context: 'frame-parent', parent: 'context-1', url: '', clientWindow: 'window-1',
+                    context: 'frame-parent', url: '', clientWindow: 'window-1',
                     originalOpener: null, userContext: 'default', children: null
                 }]
             }]


lucasmariano003-wq:
Five fork workflows are awaiting maintainer approval; EasyCLA has passed. Could a maintainer approve those runs? Here is the standalone reproduction used to verify the nested-frame case. Save it as `repro-nested-frame.mjs` at the repository root, then run:

```sh
pnpm install --frozen-lockfile
pnpm run compile:all:core
pnpm run compile:all:main
node repro-nested-frame.mjs
```

It uses only synthetic `data:` / `srcdoc` pages in a fresh headless Chrome session.

```js
import assert from 'node:assert/strict';
import { remote } from './packages/webdriverio/build/node.js';

  logLevel: 'warn',
  connectionRetryCount: 0,
  capabilities: {
    browserName: 'chrome',
    webSocketUrl: true,
    'goog:chromeOptions': { args: ['--headless=new'] }
  }
});
try {
  const inner = '<h2>inner</h2>';
  const outer = '<h1>outer</h1><iframe id="nested" srcdoc="' + inner + '"></iframe>';
  const page = '<h1>top-level</h1><iframe id="frame" srcdoc="' +
    outer.replaceAll('&', '&amp;').replaceAll('"', '&quot;') + '"></iframe>';
  await browser.url('data:text/html,' + encodeURIComponent(page));
  await browser.switchFrame(await browser.$('#frame'));
  await browser.switchFrame(await browser.$('#nested'));
  assert.equal(await browser.execute(() => 1), 1);

  const { contexts } = await browser.browsingContextGetTree({});
  const top = contexts[0], parent = top.children[0];
  assert.equal(Object.hasOwn(parent, 'parent'), false);
  const attempts = [];
  let onResult;
  const completed = new Promise(resolve => {
    onResult = event => { if (event.command === 'switchToWindow') resolve(event); };
    browser.on('result', onResult);
  });
  const onCommand = event => {
    if (event.command === 'switchToWindow') attempts.push(event.body.handle);
  };
  browser.on('command', onCommand);
  await browser.scriptCallFunction({
    functionDeclaration: 'function(){document.getElementById("nested").remove()}',
    awaitPromise: true, target: { context: parent.context }
  });
  let timer;
  const result = await Promise.race([completed, new Promise((_, reject) => {
    timer = setTimeout(() => reject(new Error('Recovery did not complete')), 5000);
  })]).finally(() => clearTimeout(timer));
  browser.off('command', onCommand);
  browser.off('result', onResult);
  assert.deepEqual(attempts, [top.context], 'Recovery must target the owning window');
  assert.equal(result.result.error, undefined);
  assert.equal(await browser.execute(() => document.querySelector('h1')?.textContent), 'top-level');
  console.log('PASS: nested iframe removal recovers to its owning top-level window');
} finally {
  await browser.deleteSession();
}
```

On macOS / Node 22.23.2 / Chrome 153.0.8010.12, base `ba488fa` targets the parent iframe ID, returns `no such window`, and fails the owning-window assertion. This patch targets the root window and passes the final document assertion. The portable version using the relative import and Chrome auto-detection was also run successfully against the patched build.

lucasmariano003-wq:
Addressed the remaining CI failures in `718a3f3`:

- Updated the EdgeDriver minimum/lock to 6.3.1, which contains the [upstream fallback fix](https://github.com/webdriverio-community/node-edgedriver/pull/565). With the same missing full browser-version download, 6.3.0 reproduces the exact `Blobs` error; 6.3.1 successfully selects the platform-compatible version, downloads it, and runs `--version`.
- The CI artifact shows the correct new window was selected, followed by a roughly 300-second renderer stall in Lighthouse's `getUrl()` against the external page. The window/frame timeouts then cascade. The two affected navigation tests now use a local HTTP fixture while retaining Lighthouse, URL/title switching, positive/negative content checks, and nested-frame behavior. They additionally verify the selected handles and Classic frame text. Their retries were removed; no timeout was increased.

Validation: all packages compile; **298 utility/context tests pass**; the two navigation cases pass with Lighthouse in four consecutive local Chrome runs; frozen-lockfile install, ESLint, and diff checks pass. The final published tree matches the tested local tree.

Please approve/run the updated fork workflows so the final head can be checked in the CI matrix.

lucasmariano003-wq:
@dprevost-LMI The remaining two failures on `752bf31` occur before the relevant tests run:

- [Windows E2E Launch](https://github.com/webdriverio/webdriverio/actions/runs/34915262127/job/104222690712) stops in Download Build Archive because ListArtifacts returns HTTP 403. The build artifact exists and is not expired; the test step is skipped.
- [macOS E2E Launch](https://github.com/webdriverio/webdriverio/actions/runs/34915262127/job/104222690755) reuses cached ChromeDriver 152.0.7977.83, which immediately fails to spawn with Darwin EBADMACHO (`-88`), before session creation. The other four browsers pass. The spec retries reuse the same driver cache. I verified this in the downloaded log artifact and checked its SHA-256.

Could you rerun just these two jobs on fresh runners? The remaining checks, including navigation and unit tests, have passed.

The final-head workflows for #15574 (`4f1976a`), #15575 (`199afde`, now restricted to the shadow-host fix) and the separate CI-test PR #15603 (`a13f52f`) are also awaiting maintainer approval. Could you approve those runs when reviewing them?

PR #15579, opened 2026-09-09
feat(utils): support conditional worker service activation

### LINKED ISSUE #10039: [💡 Feature]: Allow services to export a static method that decides if an instance should get created or not
### Is your feature request related to a problem?

Looking at services like `@wdio/sauce-service` or `@wdio/browserstack-service` they all have to manually manage whether the hook should be run or not based on whether the session is running on these services or not. For example: if no `user` and `key` is provided there is no reason for the service to be even imported within the worker.

### Describe the solution you'd like.

I'ld suggest to allow services to export a function that receives the config object and returns either `true` or `false` which determines whether the service should be ignored in the worker. This would allow service author to determine if the service should be even loaded in the worker or not based on e.g. all defined capabilities.

At the same time we should allow a function to be run in the worker process that determines if the service should be initiated or not. Different from the solution from above, here the service author can determine the initiation of the service based on the used capability in the worker.

### Describe alternatives you've considered.

Allow service to extend from a base class that could do this.

### Additional context

I am reviewing some additions to the BS service and discovered the need for it.

### Code of Conduct

jackvansickle1:
## Proposed changes

Fixes #10039.

Services currently have to decide inside their hooks whether they are relevant to a run. Add two optional activation checks so service authors can avoid unnecessary worker imports and construction:

- A named `shouldLoad(config, capabilities)` export runs in the launcher. Returning `false` excludes the module from worker imports while preserving its launcher service. The decision is shared by entries and aliases that resolve to the same module.
- A worker service's static `shouldRun(options, capabilities, config)` method runs before its constructor. Returning `false` skips that instance for the worker's capability. This also supports custom service classes.

Both checks support synchronous or asynchronous decisions. Only explicit `false` disables a service; services without these checks and preconstructed hook objects keep their existing behavior. Predicate errors retain service context, including custom class names. The public types and custom-service guide describe both stages.

## Types of changes

## Checklist

## Backport Request

## Validation

- Complete `@wdio/utils` Vitest suite: 15 files, 326 tests passed, including three compile-time cases; no type errors.
- Service initialization suite: 56 tests passed, including 42 new runtime cases. The original implementation fails 32 of these cases; restoring the exact fix passes all 56.
- Repository compiler: affected `@wdio/types` and `@wdio/utils` builds, including declaration generation, passed. Core dependency builds also passed.
- Standalone TypeScript check of `@wdio/types`: passed.
- ESLint on the changed TypeScript files and `git diff --check`: passed.
- Additional temporary ESM/CommonJS fixtures ran through the built package in separate launcher and worker Node processes. They confirmed disabled packages are not imported in workers, launcher constructors still run, and worker-specific decisions select the correct constructors.

Dependencies were restored with the pinned pnpm version and frozen lockfile. No browser, device, or live cloud-service test was run, and existing cloud services are not changed to opt in automatically.

## Further comments

Prepared and tested with OpenAI Codex, including independent code review. Please consider this contribution under the project's contributor compensation program, subject to maintainer review and eligibility.

### DIFF
--- a/packages/wdio-utils/src/initializeServices.ts
+++ b/packages/wdio-utils/src/initializeServices.ts
@@ -98,7 +98,8 @@ export async function initializeLauncherService (
     ignoredWorkerServices: string[];
     launcherServices: Services.ServiceInstance[];
 }> {
-    const ignoredWorkerServices = []
+    const ignoredWorkerServices: string[] = []
+    const workerServiceLoadDecisions = new Map<Services.ServicePlugin, boolean>()
     const launcherServices: Services.ServiceInstance[] = []
     let serviceLabelToBeInitialised = 'unknown'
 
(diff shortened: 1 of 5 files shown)


dprevost-LMI:
@greptile, is the naming `shouldLoad` & `shouldRun` clear and accurate? Since the feature is named `service activation`, should we expect to see it as enabled/disabled or is activated / deactivate?

packages/wdio-utils/src/initializeServices.ts
```diff
@@ -131,14 +132,27 @@ export async function initializeLauncherService (
                 launcherServices.push(new service(serviceConfig, caps as Capabilities.ResolvedTestrunnerCapabilities, config))
             }
 
+            /**
+             * Package-wide activation decisions are made once in the launcher, so disabled
+             * packages need not be imported into any workers. Launcher hooks are unaffected.
+             */
+            const servicePlugin = service as Services.ServicePlugin
+            if (serviceName && typeof servicePlugin.shouldLoad === 'function' && !workerServiceLoadDecisions.has(servicePlugin)) {
+                serviceLabelToBeInitialised = `"${serviceName}"`
```
dprevost-LMI, line 141: @greptileai, why does this need to be wrapped in backticks and double quotes?

PR #15597, opened 2026-09-14
fix(@wdio/browser-runner): import custom Vite and Stencil config files via file URLs

kaanisthatyou:
## Proposed changes

`@wdio/browser-runner` imports two user config files with a plain absolute path:

- `viteConfig` given as a string (`src/vite/server.ts`): `await import(path.resolve(rootDir, viteConfig))`
- the Stencil config (`src/vite/frameworks/stencil.ts`): `await import(path.join(rootDir, 'stencil.config.ts'))`

In Node's ESM loader an absolute Windows path such as `C:\project\vite.config.ts` is not a valid specifier, so the import throws `ERR_UNSUPPORTED_ESM_URL_SCHEME`:

```
Only URLs with a scheme in: file, data, and node are supported by the default ESM loader.
On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'
```

For `viteConfig` this makes the runner fail to start on Windows. For Stencil the error is swallowed by the `.catch(() => ({ config: {} }))`, so the Stencil config is silently ignored. Both paths are now converted with `url.pathToFileURL(...).href`, as `mochaOpts.require` and the Tailwind CSS optimization already do. Behavior on macOS and Linux is unchanged.

The unit tests did not catch this because Vitest resolves dynamic imports through its own module runner, which accepts Windows paths. The Stencil test now mocks the config through its file URL, so it checks the specifier the runner actually imports. On Windows it fails without this change and passes with it.

## Types of changes

## Checklist

## Backport Request

## Further comments

Tested on Windows 11 / Node 24.14.0: `vitest run packages/wdio-browser-runner/tests/vite` passes; `tests/runner.test.ts` needs `@wdio/local-runner` to be built and was failing in my local setup before this change as well. I did not run a full component test with a custom `viteConfig` end to end on Windows.

https://claude.ai/code/session_016Wuh9RTwTUHB7wxpNxrSiw


dprevost-LMI:
@kaanisthatyou, thanks for your contribution. Would you mind singing the CLA?

packages/wdio-browser-runner/src/vite/frameworks/stencil.ts
```diff
@@ -219,7 +219,7 @@ function getCompilerOptions(ts: any, rootDir: string): CompilerOptions | null {
  */
 export async function importStencilConfig(rootDir: string) {
     const configPath = path.join(rootDir, 'stencil.config.ts')
-    const config = await import(configPath).catch(() => ({ config: {} }))
+    const config = await import(url.pathToFileURL(configPath).href).catch(() => ({ config: {} }))
```
dprevost-LMI: @greptileai, would it be better to report the error somehow instead of swallowing it?

kaanisthatyou: Agreed, I pushed eece268. `importStencilConfig` now checks whether `stencil.config.ts` exists (reusing `hasFileByExtensions`) and returns the empty config only when it doesn't. If the file exists, import and evaluation errors propagate, so a broken config is no longer treated as "no config". I added a test for the missing-config case; it fails against the previous catch-all. `vitest run packages/wdio-browser-runner/tests/vite`: 32 passed.

dprevost-LMI: @greptileai, is the error better processed now?

dprevost-LMI:
@kaanisthatyou, just missing the CLA signature, and I'll run the workflow and merge!

kaanisthatyou:
The CLA is signed now — thanks for your patience, and sorry for the wait.

CI came back red, but the failure is pre-existing on `main` and unrelated to this PR:

- **This PR** — `Testrunner E2E Tests / Testrunner Tests (windows-latest)`: 104 passing, 1 failing → `main suite 1 > moveTo tests > moveTo without iframe with 0 offsets`, `Expect $('#text') to have property value: - center + out` ([log](https://github.com/webdriverio/webdriverio/actions/runs/35597888451/job/106330351323))
- **`main` @ 2d85a53** (push, 2026-09-21 06:46 UTC, ~5 h before the run above): the same job fails with the same test and the same diff ([log](https://github.com/webdriverio/webdriverio/actions/runs/35569886256/job/106242619677))

This PR only changes how `@wdio/browser-runner` resolves a *custom* Vite/Stencil config file — `import(path)` → `import(pathToFileURL(path).href)` — plus the unit tests for it. It doesn't touch pointer input, the e2e fixture page, or any runtime the `moveTo` spec goes through. Every other job on this PR is green, including the Windows unit, smoke, multiremote and standalone suites.

Happy to rebase onto the latest `main` if you'd prefer a fresh run before merging.

PR #15619, opened 2026-09-17
fix(wdio-utils): settle some and every on all-hole arrays

devangpratap:
### Motivation

some() and every() skip holes with `counter++; continue`, and apart from the
early return for an empty array, the only code that calls resolve is the `check`
callback. When every index of the array is a hole, no callback ever runs, so
neither promise ever settles:

    await some(new Array(3), async () => true)    // never settles
    await every(new Array(3), async () => false)  // never settles

Array#some returns false and Array#every returns true for that input, and
someSeries/everySeries in this file already behave that way. filter and map
still settle on the same input, since they await Promise.all over an empty list.

This is not reachable through $$(), which always builds a dense array. The fix
is for parity with someSeries/everySeries and with Array#some/Array#every, in
the same spirit as #15577.

### Fix

Count the callbacks the loop actually schedules, and resolve immediately when
that count is zero. Arrays with at least one real element take the existing path
unchanged.

### Testing

Two test.each cases covering some/someSeries and every/everySeries, following the
shape used for map/mapSeries in #15577. Against main the two parallel variants
fail while the Series variants pass, which isolates the bug to the parallel
implementations. packages/wdio-utils/tests/pIteration.test.ts passes in full with
the change, and eslint is clean.


christian-bromann (approved):
LGTM

PR #15624, opened 2026-09-19
fix(webdriverio): let switchToParentFrame recover a destroyed context

PloneMraz:
Refs #15570.

Two of the three points in that issue are already fixed on `main` by #15573 — `browsingContext.contextDestroyed` is subscribed and the owning window is recovered. This covers the remaining one: **`switchToParentFrame()` cannot recover a context the page destroyed.**

## Why it cannot

`switchToParentFrame` is listed in `COMMANDS_REQUIRING_RESET`, but `#onCommand` handles it in its own branch and returns from every path in that branch, so the reset further down is never reached for it:

```js
if (event.command === 'switchToParentFrame') {
    if (!this.#currentContext) {
        return
    }
    return this.#browser.browsingContextGetTree({}).then(({ contexts }) => {
        const parentContext = this.findParentContext(this.#currentContext!, contexts)
        if (!parentContext) {
            return          // <- cached id stays, even though it is gone from the tree
        }
        this.setCurrentContext(parentContext.context)
    })
}
...
if (COMMANDS_REQUIRING_RESET.includes(event.command)) {
    this.#currentContext = undefined    // <- unreachable for switchToParentFrame
}
```

`findParentContext()` walks the tree looking for a node whose `children` contain the cached id. A context destroyed by the page is not in the tree at all, so there is no parent to find and the early `return` leaves the dead id cached. From there every BiDi command fails with `no such frame`, and there is no user-land way to clear it — which is what makes this worth fixing rather than documenting.

## The change

In that branch, when no parent is found, clear `#currentContext` and `#currentWindowHandle`. The next command then resolves the active context again instead of addressing a frame that no longer exists. A context that genuinely has a parent still switches to it, unchanged.

I deliberately did **not** remove `switchToParentFrame` from `COMMANDS_REQUIRING_RESET` even though that entry is now provably dead — that felt like your call rather than mine, and it is a separate change.

## Tests

Two cases in `packages/webdriverio/tests/session/context.test.ts`:

- a cached context that is gone from the tree → cache is dropped, the next `getCurrentContext()` no longer returns the dead id
- a cached context that does have a parent → still switches to the parent (guards against over-clearing)

- `vitest --config vitest.config.ts --run packages/webdriverio/tests/session/context.test.ts`: **19 passed**
- Reverting only the `if (!parentContext)` body: **1 failed, 18 passed** — the first case, so the test is pinned to the change

One note for reviewers writing tests in this file: `SessionManager`'s base constructor registers a `command` listener before `ContextManager` does, so the manager's handler is the **last** entry in `getListeners().command`, not the first. I lost a run to that and left a comment in the test so the next person does not.


christian-bromann (approved):
LGTM 👍

PR #15626, opened 2026-09-19
fix(webdriverio): validate cached BiDi context against the context tree before reuse

### LINKED ISSUE #15570: [🐛 Bug]: BiDi cached browsing context is never invalidated on contextDestroyed; classic commands mask it and switchToParentFrame() cannot recover

nikolas-sapa:
## Description

Fixes #15570

`ContextManager.getCurrentContext()` returned the cached `#currentContext` id unconditionally. If the page itself destroyed a frame (not via a WebdriverIO command — e.g. an iframe removed by site code between test commands), every subsequent BiDi-scoped command (`browser.execute()`, etc.) failed with:

```
no such frame - Context <id> not found
```

while classic commands (`getTitle`, `getUrl`) kept succeeding, masking the breakage. The `browsingContext.contextDestroyed` handler does reset the cache, but recovery can race with command execution, and any path that misses the event leaves a permanently stale id cached.

## Changes

- `packages/webdriverio/src/session/context.ts`: `getCurrentContext()` now validates the cached context id against `browsingContextGetTree` before returning it. If the id no longer exists, the cache is cleared and `initialize()` re-resolves a valid context (existing switch/recovery logic takes over from there). Mobile sessions short-circuit — context validation is a BiDi-desktop concern there. If the validation call itself fails, we fall back to the cached id rather than introducing a new failure mode on every command.
- `packages/webdriverio/tests/session/context.test.ts`: regression tests for the `switchToParentFrame` reset path from the issue report.

## How has this been tested?

- [x] Unit tests added and passing locally
- [x] Manual repro from the issue (iframe destroyed by page code → `browser.execute()` recovers instead of failing permanently)

**Note:** this adds one `browsingContextGetTree` round-trip per `getCurrentContext()` call when a context is cached. If maintainers are concerned about the extra traffic on hot paths, a cheaper variant is to only validate after a command error containing `no such frame`, but that requires plumbing error context back into the manager — happy to rework in that direction if preferred.


christian-bromann:
@nikolas-sapa thanks for the PR, I think we should address the Greptil comments. Happy to merge after.

nikolas-sapa:
Closing this — on re-review it's invalid. #15570 is already fixed on `main`: the `switchToParentFrame` branch in `ContextManager.#onCommand` now returns early, so it no longer falls through to the `COMMANDS_REQUIRING_RESET` reset that used to clobber the async parent-context resolution. My branch inadvertently reverted that upstream fix and layered a per-call `browsingContextGetTree` validation on top that regressed frame switching (the `switchToParentFrame` e2e failed with `Expect $('h1') not to exist`). Apologies for the noise — no change is needed here.

PR #15628, opened 2026-09-19
fix(webdriverio): make throttleNetwork reach service-worker initiated requests

### LINKED ISSUE #7452: `browser.throttle` does not apply to fetch requests initiated in service worker scripts
**Environment:**
 - **WebdriverIO version:** 7.12.6
 - **Mode:**  WDIO Testrunner
 - **Node.js version:** 12.16.3
 - **NPM version:** 6.14.4
 - **Browser name and version:** Chrome 93
 - **Platform name and version:** Manjaro 5.10.63
 - **Additional wdio packages used:** @wdio/cli @wdio/local-runner @wdio/mocha-framework @wdio/spec-reporter

**Config of WebdriverIO**
```javascript
{
  runner: 'local',
  specs: [ 'test/*.spec.js',],
  maxInstances: 1,
  capabilities: [{
    browserName: 'chrome',
  }],
  baseUrl: `http://localhost:${port}`,
  logLevel: 'info',
  outputDir: __dirname,
  waitForTimeout: 1000000,
  framework: 'mocha',
  reporters: ['spec'],
  mochaOpts: {
    ui: 'bdd',
    timeout: 70000,
  }
}
```

**Describe the bug**
`browser.throttle` does not apply to fetch requests initiated in service worker scripts. Even if the browser is throttled with the `offline` preset, service worker registration is still successful. 

**To Reproduce**
Please ckeckout this repo and run tests: https://github.com/dianabarsan/wdio-throttle-sw

**Expected behavior**
Fetch requests initiated by a service worker script should be throttled. 
According to the documentation:
> Throttle the network capabilities of the browser. This can help to emulate certain scenarios where a user loses their internet connection and your app needs to address that.

Where an offline-first app, that pre-caches all resources cannot be tested. 

**Additional context**
We're developing an offline-first application that pre-caches all resources, using a service worker. It's important to test service worker behavior, registration  and updating success with different network capabilities.

nikolas-sapa:
## Description

Fixes #7452

`browser.throttleNetwork()` had no effect on requests initiated from service workers: even the `offline` preset left service-worker registration succeeding, making offline-first apps untestable.

### Root causes

1. The `Network` domain was never enabled on the CDP session before calling `Network.emulateNetworkConditions` — emulation only applies to domains that are enabled, so the conditions could be silently ignored for the target page and any service workers attached to it.
2. Service workers that are **already registered** when throttling is applied do not re-evaluate their network stack; the conditions only take effect for workers (re)started under them.

## Changes

- `packages/webdriverio/src/commands/browser/throttleNetwork.ts`:
  - send `Network.enable` on the CDP session before applying emulation
  - reload the page (`Page.reload`) after applying the conditions so already-registered service workers are re-evaluated under the new network constraints

## How has this been tested?

- [x] Manual verification against the repro from #7452: with `offline` throttling, service-worker registration now fails as expected on a page whose worker performs its own fetches

**Note to maintainers:** the reload makes throttling stateful-by-side-effect (one page reload per `throttleNetwork` call). If that is too invasive as default behavior, an alternative is gating the reload behind an option (e.g. `throttleNetwork(params, { reload: true })`) — happy to rework. Also worth noting this fix targets the CDP path; BiDi network emulation support would be a separate piece.


nikolas-sapa:
Closing in favor of #15629, which covers the same fix for #7452 plus regression tests (falsified by reverting the source change) and a docs note about service workers registered after the call. Apologies for the duplicate — both came from parallel work in progress on the same issue; #15629 is the one to review.

PR #15629, opened 2026-09-19
fix(webdriverio): apply throttleNetwork conditions to service workers

### LINKED ISSUE #7452: `browser.throttle` does not apply to fetch requests initiated in service worker scripts

nikolas-sapa:
## Proposed changes

Fixes #7452.

`Network.emulateNetworkConditions` is scoped to the CDP session it is sent on. `throttleNetwork` sends it on the page's session only, but a service worker runs in its own CDP target with its own session — so every fetch initiated inside a registered service worker bypassed the page-level emulation entirely. That is why `browser.throttleNetwork('offline')` left service worker registrations succeeding in the reproduction from the issue, and why offline-first apps could not be tested.

This PR applies the same conditions to every service worker target that is registered at the time of the call (`target.type() === 'service_worker'`), on a dedicated CDP session per target. `Network.enable` is sent on those sessions first, mirroring what DevTools does for worker targets.

Two things worth flagging:

- **Service workers registered *after* the command ran are not covered** — their target does not exist yet when the conditions are applied. The command docs now note that `throttleNetwork` should be called again if a worker registers later. Covering late registrations would mean holding conditions in state and subscribing to target creation, which is a bigger change than this fix; happy to follow up if wanted.
- **The page does not reload.** A reload after applying conditions would reset page state mid-test; attaching to the service worker target fixes the reported repro directly.

The deprecation shim `browser.throttle` delegates to `throttleNetwork` and picks this up automatically. Sauce sessions keep using `sauceThrottleNetwork` as before.

## Types of changes

## Checklist

## Backport Request

## Further comments

Two tests were added to `packages/webdriverio/tests/commands/browser/throttleNetwork.test.ts`: one asserting the conditions (and `Network.enable`) reach a registered service worker's session, one asserting non-worker targets are left untouched. Reverting the source change turns the first test red, so it pins the behaviour rather than just describing it. All 6 tests in the file pass.

For sessions without service workers, `targets()` returns nothing matching the filter and no extra CDP traffic is produced — the page-session behaviour is byte-identical to before.


nikolas-sapa:
One robustness follow-up from a pre-merge self-review, pushed as ec0eea0: a service worker can unregister between enumerating the targets and attaching a CDP session to it. That would previously reject `throttleNetwork` *after* the page session was already throttled; the command now skips a worker that dies during attach (debug-logged) and keeps the page throttling result. All 6 tests in the file still pass.

christian-bromann (approved):
LGTM 👍

PR #15630, opened 2026-09-19
feat(webdriver): add isAppium session flag and expose Appium commands to non-mobile Appium sessions

### LINKED ISSUE #15600: [🐛 Bug]: Appium protocol commands are not available in non-mobile Appium sessions

### WebdriverIO Version

9.31.7 (plain `webdriver`, not `webdriverio`)

### Node.js Version

24.20.0

### Mode

WDIO Testrunner

```typescript
{
  "platformName": "mac",
  "appium:automationName": "safari"
}
```

### What happened?

When creating an Appium-based non-mobile session (for example, using one of Appium's browser drivers like chromium/gecko/safari), no Appium protocol commands are available, even though many commands in this protocol are still supported in a session like this (for example, retrieval of settings/capabilities/commands/extension methods/etc.).

Direct `curl` calls to Appium endpoints work as expected - the respective commands are simply not available on the WDIO level.

I expect all Appium protocol methods to be accessible in any Appium session, regardless of whether it is mobile or not.

With the context of previous discussions (https://github.com/webdriverio/webdriverio/issues/13947, https://github.com/webdriverio/webdriverio/issues/14140), while it is true that not all Appium protocol methods are supported by all drivers, I would not expect WDIO to know which driver supports which commands, especially considering the existence of multiple third-party Appium drivers. Therefore I think it's better for WDIO to allow all protocol commands, and if the driver does not support a command, simply return the error coming from the driver.

The current implementation appears to only include the Appium protocol commands if the session is determined to be a mobile session (`driver.isMobile`), in [`webdriver/src/utils.ts`](https://github.com/webdriverio/webdriverio/blob/main/packages/webdriver/src/utils.ts#L303). In turn, `isMobile` returns `false` for browser-based Appium sessions, in [`wdio-utils/src/envDetector.ts`](https://github.com/webdriverio/webdriverio/blob/main/packages/wdio-utils/src/envDetector.ts#L135).

To remedy this, I would probably recommend adding a new `isAppium` top-level property, which would return `true` if the `appium:automationName` capability is present and populated, regardless of its value. Then update the protocol inclusion check to use `isAppium` instead of `isMobile`.

### How to reproduce the bug.

1. Install Appium and the `safari` driver
2. Start a WDIO session using the above capabilities
3. Observe that Appium protocol methods such as `getAppiumCommands` are not available

### Relevant log output

```typescript
N/A
```

### Code of Conduct

nikolas-sapa:
## Description

Fixes #15600

Appium protocol commands were only attached to the session prototype when `isMobile` was true. But Appium can also wrap **desktop** browser drivers (e.g. its `safari` or `gecko` driver on macOS): those sessions report `browserName` with a non-mobile `platformName`, so `isMobile` is `false` — yet the session is still served by an Appium server and its `/session/:id/appium/*` endpoints exist. Result: commands like `executeDriverScript`, `startScreenRecording`/`stopScreenRecording` or the settings API were missing from the client even though the server supports them.

## Changes

- `packages/wdio-utils/src/envDetector.ts`: new `isAppium(capabilities)` detector — true when the session response carries an `automationName`, `appium:options`, W3C-namespaced `appium:*` capabilities, or legacy non-namespaced Appium keys (`appiumVersion` etc.). Exposed as `isAppium` in both `sessionEnvironmentDetector` and `capabilitiesEnvironmentDetector`.
- `packages/webdriver/src/types.ts`: add `isAppium` to `SessionFlags`.
- `packages/webdriver/src/utils.ts`:
  - `getPrototype` now merges `AppiumProtocol` for non-mobile Appium sessions (`!isMobile && isAppium`). MJSONWP legacy commands stay mobile-only.
  - `getEnvironmentVars` exposes the `isAppium` flag on the instance.

## How has this been tested?

- [x] Unit tests: `packages/wdio-utils/tests/envDetector.test.ts` (new `isAppium` cases incl. desktop-Appium and non-Appium fixtures) and `packages/webdriver/tests/utils.test.ts` (prototype includes `appiumLock` for `isMobile: false, isAppium: true`, excludes MJSONWP `getNetworkConnection`)
- [x] Full `wdio-utils` + `webdriver` suites: 25 files / 532 tests passing
- [ ] Manual smoke test against a local Appium server with a wrapped desktop driver (needs Appium installed — happy to do it if reviewers want it)

**Naming note:** I kept the `is<Thing>` convention of the existing flags. If maintainers would rather gate Appium commands behind a broader condition (e.g. also folding the `appium:automationName: 'gecko'|'safari'|...:` wrapped-driver cases from #13947 into `isMobile`), I'm happy to rework.


dprevost-LMI:
This is a duplicate of https://github.com/webdriverio/webdriverio/pull/15618. Can you sync with @scs0209 PR?

nikolas-sapa:
Apologies — I missed #15618 when researching #15600 (my issue scan didn't surface it). @scs0209's PR predates mine and takes the same approach (`isAppium` session flag + merging `AppiumProtocol` for non-mobile Appium sessions), so theirs should be the one reviewed. Closing this as a duplicate; happy to help test or review #15618 instead.

PR #15634, opened 2026-09-19
fix(@wdio/local-runner): keep --enable-source-maps and de-dupe worker NODE_OPTIONS

### LINKED ISSUE #15633: [🐛 Bug]: worker NODE_OPTIONS drops --enable-source-maps and duplicates parent flags when NODE_OPTIONS is set
### Environment

`@wdio/local-runner` on `main` (`packages/wdio-local-runner/src/worker.ts`, `startProcess()`).

### What happened?

`startProcess()` builds the worker's `NODE_OPTIONS` like this:

```ts
const runnerEnv = Object.assign({
    NODE_OPTIONS: '--enable-source-maps',
}, process.env, this.config.runnerEnv, { /* WDIO_WORKER_ID, NODE_ENV */ })
// ...
runnerEnv.NODE_OPTIONS = process.env.NODE_OPTIONS + ' ' + (runnerEnv.NODE_OPTIONS || '')
```

The `Object.assign(..., process.env, ...)` overwrites the `--enable-source-maps` default with the parent's `NODE_OPTIONS` **before** the concat line runs. So when the parent process has `NODE_OPTIONS` set (very common — `--import tsx`, `--experimental-vm-modules`, `--max-old-space-size`, etc.):

- **`--enable-source-maps` is silently dropped** — worker stack traces lose source-map support, the exact thing this code is meant to guarantee; and
- **the parent flags are duplicated**, e.g. `--import tsx --import tsx`.

When the parent has **no** `NODE_OPTIONS`, the string becomes the literal `"undefined --enable-source-maps"` (Node happens to tolerate the bare `undefined` token, but it is clearly unintended).

### Reproduction (the exact expression)

```js
const parentEnv = { NODE_OPTIONS: '--import tsx' }
const runnerEnv = Object.assign({ NODE_OPTIONS: '--enable-source-maps' }, parentEnv, {})
runnerEnv.NODE_OPTIONS = parentEnv.NODE_OPTIONS + ' ' + (runnerEnv.NODE_OPTIONS || '')
// => "--import tsx --import tsx"  (--enable-source-maps lost, --import tsx duplicated)
```

### Expected

The worker's `NODE_OPTIONS` should always contain `--enable-source-maps`, preserve the parent's flags, not duplicate them, and never contain a literal `undefined`.

### Fix

I have a fix + regression tests ready (merge the parent tokens with `--enable-source-maps` and de-duplicate, dropping empty tokens) and will open a PR referencing this issue.

nikolas-sapa:
Fixes #15633.

## The bug

`startProcess()` builds the worker's `NODE_OPTIONS` as:

```ts
const runnerEnv = Object.assign({ NODE_OPTIONS: '--enable-source-maps' }, process.env, this.config.runnerEnv, { /* ... */ })
// ...
runnerEnv.NODE_OPTIONS = process.env.NODE_OPTIONS + ' ' + (runnerEnv.NODE_OPTIONS || '')
```

`Object.assign(..., process.env, ...)` overwrites the `--enable-source-maps` default with the parent's `NODE_OPTIONS` before the concat runs. So when the parent process has `NODE_OPTIONS` set (common: `--import tsx`, `--experimental-vm-modules`, …):

- **`--enable-source-maps` is silently dropped** — worker stack traces lose source-map support (the thing this code exists to guarantee);
- **parent flags are duplicated** (e.g. `--import tsx --import tsx`).

With no parent `NODE_OPTIONS`, the value became the literal `"undefined --enable-source-maps"`.

## The fix

Build `NODE_OPTIONS` by merging the parent's tokens with `--enable-source-maps` and de-duplicating, dropping empty tokens:

```ts
runnerEnv.NODE_OPTIONS = [
    ...(process.env.NODE_OPTIONS ?? '').split(' '),
    '--enable-source-maps',
].filter((flag, index, flags) => flag !== '' && flags.indexOf(flag) === index).join(' ')
```

Results: unset parent → `--enable-source-maps`; `--import tsx` → `--import tsx --enable-source-maps`; already-present → not duplicated; never a literal `undefined`.

## Tests

Adds three regression tests in `worker.test.ts` (unset parent, parent flags preserved, no duplication) that spy on `ProcessFactory.prototype.createWorkerProcess` and assert the worker `env.NODE_OPTIONS`. All three fail against the previous code (`'undefined --enable-source-maps'`, dropped source-maps, duplication) and pass with the fix. Full `worker.test.ts` is green (28/28), eslint clean.


christian-bromann (approved):
LGTM 👍

christian-bromann:
> 1. P1 **Worker options are discarded** [▶](https://github.com/webdriverio/webdriverio/pull/15634#discussion_r4054429623)
> 2. P1 **Repeated options become malformed** [▶](https://github.com/webdriverio/webdriverio/pull/15634#discussion_r4054429636)

This seems worth addressing.

christian-bromann (changes requested):
Still remaining Greptile comments

nikolas-sapa:
Pushed e0e35f5 to address both remaining review comments:

1. **Worker options were discarded** — the merge now uses the resolved `runnerEnv.NODE_OPTIONS` as its base (where `config.runnerEnv` already takes precedence over the parent `process.env`), so worker-specific flags injected through `config.runnerEnv.NODE_OPTIONS` are preserved instead of being overwritten by flags built from `process.env` alone.

2. **Repeated options became malformed** — instead of splitting on whitespace and de-duplicating every token (which detached operands, e.g. `--require a.js --require b.js` → `--require a.js b.js`), it now appends `--enable-source-maps` as a single token only when it isn't already present. Repeated option/value pairs stay intact.

Added two regression tests: one asserting `config.runnerEnv.NODE_OPTIONS` survives, and one asserting `--require a.js --require b.js` keeps both operands attached. Full `worker.test.ts` suite: 30/30 passing, no type errors, eslint clean.

mccmrunal:
@greptileai all comments resovled?

mccmrunal:
@greptileai safe to merge?

PR #15637, opened 2026-09-19
fix(webdriverio): keep a valid top-level context on switchToParentFrame

PloneMraz:
Follow-up to #15624, fixing a case the review bot flagged there and I did not cover.

## What I got wrong

#15624 clears the cached context whenever `findParentContext()` returns nothing. That condition is broader than the bug it was meant to fix: **a valid top-level context also has no parent.** So calling `switchToParentFrame()` while already at the top level now throws away a perfectly good context and window handle instead of being a no-op. The next context-dependent command has to reinitialise through `getWindowHandle()`, and the cleared window handle also bypasses `switchWindow`'s same-window shortcut.

## The distinction that was missing

"No parent" means two different things, and the tree already tells them apart:

- **still in the tree** → top-level context, nowhere to step up to → do nothing, as before #15624
- **absent from the tree** → the page destroyed it → drop the cache, which is what #15624 was for

One `findContext(..., 'byContextId')` call separates them.

## Tests

Adds the top-level case that was missing, alongside the two from #15624:

- top-level context, still in the tree → context survives the call
- destroyed context, absent from the tree → cache dropped
- context with a real parent → still switches up

- `vitest --config vitest.config.ts --run packages/webdriverio/tests/session/context.test.ts`: **20 passed**
- Removing the new `stillInTree` guard, i.e. running current `main`: **1 failed, 19 passed** — the new top-level case, so it is pinned to this change

Credit to the greptile review on #15624 for catching it.


christian-bromann (approved):
👍

PR #15638, opened 2026-09-19
fix(webdriverio): release the session-manager registration on removeListeners

PloneMraz:
Found while working on #15624, not from an issue report.

## The inconsistency

`SessionManager` guards its base `command` listener with a module-level set:

```ts
const registrationId = `${this.#browser.sessionId}-${this.#scope}`
if (!listenerRegisteredSession.has(registrationId)) {
    this.#browser.on('command', this.#onCommandListener)
    listenerRegisteredSession.add(registrationId)
}
```

`removeListeners()` detaches the listener but never removes the marker, so from that point the set claims a listener exists where none does.

## Why it matters

That listener is what makes a manager clean itself up: it watches for `deleteSession` and calls `removeListeners()` plus deletes the instance from `sessionManager`. So:

- a manager created for the same session and scope **after** a previous one was removed registers **no** listener, and therefore can never clean itself up in turn — its subclass listeners (`result`, `browsingContext.*`) stay attached for good
- the set grows for the life of the process, one entry per session and scope, and nothing ever removes an entry

It is also visible in the existing tests: `should listener registered` and `should listener registered only once...` have to reassign `browser.sessionId` between cases specifically to dodge the leftover markers.

## The change

Delete the marker in `removeListeners()`, next to the `off()` that makes it stale. One line, and the set then tracks exactly the listeners that are actually attached.

## Tests

- new case: create a manager, `removeListeners()`, create another for the same session and scope → expects two registrations. On current `main` it reports **1 instead of 2** (verified by reverting just the delete)
- `vitest --config vitest.config.ts --run packages/webdriverio/tests/session/`: **36 passed** across all 5 files in that folder

I kept this to the one line rather than also giving the set a proper lifecycle, since anything larger there is a design call rather than a bug fix.


PloneMraz:
Good catch — fixed in the push just now.

You are right that the unconditional delete was unsafe: an instance that found the key already taken registers no listener, so its `off()` detaches nothing, and clearing the marker from there would let the next instance add a second listener beside the owner's. Both would then handle the same `deleteSession`.

Each instance now records the marker it actually added (`#ownedRegistrationId`) and releases only that one. I also made it store the id used at registration rather than rebuilding it from `sessionId` in `removeListeners()`, since `sessionId` can change in between — rebuilding it could delete someone else's key and leave its own behind.

New test covers the case you described: two managers on one key, the non-owning one must not free it, the owning one must. Reverting to the unconditional delete makes it report 2 registrations where 1 is expected.

`tests/session/`: 37 passed. eslint clean.

christian-bromann (approved):
👍

PR #15073, opened 2026-02-02
fix: resolve mock state error and race condition (#15072)

mccmrunal:
## Proposed changes

[//]: # (Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue.)
Fixes #15072 - This PR resolves two critical bugs in WebdriverIO's mock functionality:

Bug 1: Duplicate Mock State Error
When mocking the same URL twice in a test, the second attempt fails with:

Root Cause: WebDriver BiDi maintains state for network intercepts. Creating a second mock without properly cleaning up the first causes a state conflict.

Solution: Implemented a cleanup-before-create pattern that detects existing mocks with the same URL pattern and filter options, calls restore() to remove the BiDi intercept, and then creates the new mock.

Bug 2: Race Condition with In-Flight Requests

When restore() is called while requests are in-flight (blocked and waiting to be continued), those requests could hang indefinitely.
Root Cause: The restoration flow set the #restored flag before cleaning up in-flight requests and removing the BiDi intercept, creating a window where requests could be ignored.

Solution:
Added #blockedRequests Set to track all blocked request IDs throughout their lifecycle Reordered restore()
 to: (1) continue all in-flight blocked requests, (2) remove BiDi intercept, (3) set #restored = true
Enhanced early return guards to handle late-arriving requests gracefully
## Types of changes

[//]: # (What types of changes does your code introduce to WebdriverIO?)
[//]: # (_Put an `x` in the boxes that apply_)

- [ ] Polish (an improvement for an existing feature)
- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update (improvements to the project's docs)
- [ ] Specification changes (updates to WebDriver command specifications)
- [ ] Internal updates (everything related to internal scripts, governance documentation and CI files)

## Checklist

[//]: # (_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._)

- [x] I have read the [CONTRIBUTING](https://github.com/webdriverio/webdriverio/blob/main/CONTRIBUTING.md) doc
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added the necessary documentation (if appropriate)
[...]

PR #15643, opened 2026-09-20
fix(webdriverio): release an intercepted request once per phase

### LINKED ISSUE #15072: [🐛 Bug]: Error happens if same url is mocked 2x in the same scenario

### WebdriverIO Version

9.23.2

### Node.js Version

24.13.0

### Mode

WDIO Testrunner

```typescript
{
        browserName: "chrome",
        "goog:chromeOptions": {
            args: [
                "--start-maximized",
                "--lang=en-GB"
            ]
        },
    }
```

### What happened?

I need to mock the same URL multiple times within a single test scenario to verify different responses. However, on the 2nd attempt, the run fails with the following error:

`Error: WebDriver Bidi command "network.continueRequest" failed with error: unknown error - Invalid state for continueInterceptedRequest`

Subsequent mock definitions for the same URL should successfully override previous ones without causing state errors (as it was work with WDIO v8)

### How to reproduce the bug.

You have here a reproducible example:

[wdio.zip](https://github.com/user-attachments/files/24994339/wdio.zip)

### Relevant log output

```typescript
Error: WebDriver Bidi command "network.continueRequest" failed with error: unknown error - Invalid state for continueInterceptedRequest
```

### Code of Conduct

PloneMraz:
### Proposed changes

Fixes #15072 — mocking the same URL twice in one scenario failed with
`WebDriver Bidi command "network.continueRequest" failed with error: unknown error - Invalid state for continueInterceptedRequest`.

Each mock registers its own `network.addIntercept` and its own listener on
`network.beforeRequestSent` / `network.responseStarted`. When two mocks match the
same URL, the driver reports both intercepts on the event and **both** handlers
try to release the same paused request. The first call succeeds, the second is
rejected. Nothing awaits an event handler, so that rejection escapes as an
unhandled rejection — which is why the same code sometimes fails a test loudly
and sometimes appears to work.

Reproduced on `main` (`2e38958`) with Chrome 153.0.8010.52, running the sequence
from the issue (mock the URL, navigate and search, `mockClearAll()`, mock the same
URL again, navigate and search again). The BiDi log shows the double release:

```
BIDI COMMAND network.continueRequest {"request":"23672.143"}
BIDI COMMAND network.continueRequest {"request":"23672.143"}
BIDI RESULT   {"id":23,"result":{},"type":"success"}
BIDI RESULT   {"id":24,"error":"unknown error","message":"Invalid state for continueInterceptedRequest"}

BIDI COMMAND network.provideResponse {"request":"23672.143", ...second mock's body}
BIDI COMMAND network.provideResponse {"request":"23672.143", ...first mock's body}
BIDI RESULT   {"id":25,"result":{},"type":"success"}
BIDI RESULT   {"id":26,"error":"unknown error","message":"Invalid InterceptionId."}
```

and the event itself names both intercepts:

```
beforeRequestSent req=27104.143 blocked=true intercepts=["a9d45f6c-…","10fd3d44-…"]
```

The fix routes every release through `#continueRequest` / `#provideResponse` /
`#failRequest`. All listeners of one BiDi event receive the same parameter
object, so when `intercepts` holds more than one id the first mock to release the
request marks the event and the others skip a call that can only fail. A request
blocked by a single intercept takes no mark and goes through untouched, so the
common case is unchanged.

`beforeRequestSent` and `responseStarted` arrive as separate event objects, which
is what keeps a request releasable once in each phase.

Whichever mock releases the request first still wins, exactly as today — this
changes nothing about which response is delivered, only that the losing mocks no
longer send a call the driver rejects. If maintainers would rather define an
explicit precedence between overlapping mocks (e.g. "the most recently registered
one wins"), that is a separate change and I am happy to follow up.

One incidental change: `networkFailRequest` in the `mock.respond()` error path
kept its `.catch(#handleNetworkProvideResponseError)`, and the `abort()` path now
shares it. It only swallows "no such request", which is the same harmless race.

### Verification

- `npx vitest run packages/webdriverio` — 1334 passed, 1 skipped, no type errors.
- Four new unit tests in `tests/utils/interception/index.test.ts`: a doubly
  blocked request is continued once, responded to once and failed once, plus a
  control asserting a single mock still releases the same request in both phases.
- Mutation check: making `claimRelease()` always return true turns the three
  multi-mock tests red and leaves the control green.
- e2e: `wdio/headless/mocking.e2e.ts` — 9 passing. Full headless e2e run —
  105 passing. (`puppeteer.e2e.ts` crashes under parallel load on this Windows
  machine with `Assertion failed: !(handle->flags & UV_HANDLE_CLOSING)`; it does
  the same on an unmodified `main` and passes in isolation, so it is unrelated.)
- Re-running the issue's scenario against the built package: two
  `network.continueRequest` and two `network.provideResponse` for two requests,
  and no protocol error.

### Types of changes

### Checklist


christian-bromann:
Thank you! Can we update the branch?

PloneMraz:
Updated — the branch is now rebased onto `main` (`2ce0cf7`).

While rebasing I saw that #15073 landed for the same issue, so I re-measured before pushing rather than assume this was still needed. #15073 makes `mock()` replace a mock that has an *identical* definition, which removes the exact case in the issue report. The overlap it does not cover is the one the `mockClearAll` docs already show — mocks whose **definitions differ** while their patterns match the same request:

```ts
const broad  = await browser.mock('https://cdn.jsdelivr.net/npm/**')
const narrow = await browser.mock('https://cdn.jsdelivr.net/npm/hammerjs@1.1.3/hammer.min.js', { method: 'get' })
```

On current `main`, the request both of them match arrives with two intercepts and is released twice:

```
beforeRequestSent intercepts=2
network.continueRequest  {"request":"3892.5"}   x2
network.provideResponse  {"request":"3892.5"}   x2
-> {"error":"unknown error","message":"Invalid InterceptionId."}
```

With this branch: one `continueRequest` and one `provideResponse` per request per phase, and no protocol error.

One thing changed since your first look. Guarding the release naively would have let a mock that is only releasing because the request **does not** match it take the release away from the mock it belongs to — which would break two mocks separated by `method`. So a declining mock now defers to the end of the event dispatch and releases only if nobody claimed the request. Verified in the browser:

```ts
const getMock  = await browser.mock(url, { method: 'GET'  }); getMock.respond({ who: 'get' })
const postMock = await browser.mock(url, { method: 'POST' }); postMock.respond({ who: 'post' })
// GET  -> {"who":"get"}     getMock.calls  === 1
// POST -> {"who":"post"}    postMock.calls === 1
```

There are two unit tests covering that precedence, plus one for the case where every mock declines and the request must still be released.

Verification: `npx vitest run packages/webdriverio` — 1347 passed, 1 skipped, no type errors. `mocking.e2e.ts` — 9 passing. Mutation check: making the claim always succeed turns 5 of the new tests red, and letting a declining mock release immediately turns exactly the precedence test red.

One note in case it is useful to you: `remocking.e2e.ts` hangs on this machine, and it does so on an unmodified `main` as well, so it is not something this branch introduces. It stalls during the first page load — `browser.mock('**/api/test', …)` is sent as `network.addIntercept {"urlPatterns":[{"type":"pattern"}]}`, i.e. with no pathname, so it intercepts every request on the page. I did not chase it further since it is outside this change, but it may be worth a look.

christian-bromann (approved):
LGTM 👍

PR #15672, opened 2026-09-21
breaking(@wdio/runner): remove the LegacyCustomStubCommand and setupExpect shims

PloneMraz:
### Proposed changes

Fixes #15651 — removes the two `@wdio/runner` shims that were marked *to remove in v10*.

**`LegacyCustomStubCommand`.** It existed because `addCommand` still accepts the deprecated positional form and the protocol stub recorded whatever arguments it was handed, so the runner had to branch on the shape when replaying commands onto the real session:

```ts
for (const [name, func, thirdArg, proto, instances] of customStubCommands) {
    if (typeof thirdArg === 'object' && thirdArg !== null) { … }
    else if (typeof thirdArg === 'boolean') { … }
    else { … }
}
```

The stub now folds the positional form into the options object as it records it, the same way `monad.ts` and the multiremote `addCommand` already do. The transfer therefore has one shape and the runner replays `[name, fn, options]` unconditionally.

The **public** positional overloads of `addCommand` are deliberately untouched — the issue calls those out as separate work — and they keep working through exactly the path they always did, since `CustomCommandOptions` already carries `proto` and `instances`.

**`setupExpect`.** The `Map` overload and the runtime shim that made both `Object.entries(matchers)` and `matchers.entries()` work are gone; the signature is now single. No adapter in this repository used the `Map` form (`@wdio/jasmine-framework` already takes `typeof wdioCustomMatchers`), so this only affects third-party adapters, which need to iterate with `Object.entries(wdioMatchers)`.

### Acceptance criteria

- [x] `LegacyCustomStubCommand` type is deleted
- [x] Protocol stub / session-start transfer only accepts `[name, fn, CustomCommandOptions]`
- [x] `setupExpect` has a single signature; the `Map` overload and runtime shim are gone
- [x] Mocha / Jasmine / Cucumber adapters compile against the new signature
- [x] Runner unit tests no longer exercise the positional tuple or `Map` form

### Verification

- `npx vitest run` across the monorepo — 4245 passed, 8 skipped. One unrelated failure, `wdio-browser-runner`'s `mockHoisting` snapshot, which mismatches on a Windows drive prefix (`/E:/some/other/…` vs `/some/other/…`) and fails the same way on an unmodified `v10`.
- `npx vitest run packages/wdio-runner packages/webdriverio/tests/protocol-stub.test.ts` — 74 passed, no type errors.
- New tests in `packages/webdriverio/tests/protocol-stub.test.ts` cover where the behaviour moved: the options object is recorded as given, the positional form is folded into it, and an options object is recorded even when only a name and a function are passed.
- Mutation check: dropping the positional fold (`return {}`) turns the positional test red; recording `[name, fn]` without the options element turns three of the stub tests red.
- `npx tsc --noEmit` on `packages/webdriverio` is clean. `packages/wdio-runner` reports three errors about `TimingTracker`, `cpuProf` and `heapProf` — those are present on an unmodified `v10` here too and look like a stale local `@wdio/utils` build rather than anything in this change.

### Types of changes

### Checklist


PloneMraz:
The `Testrunner Tests (windows-latest)` job is red, and I do not think this change is the cause — but since it touches the session-start path every e2e test goes through, here is the evidence rather than an assertion.

**All six failures are tests that fetch `the-internet.herokuapp.com`, and only those:**

| test | URL |
| --- | --- |
| can query shadow elements | `/shadowdom` |
| supports basic auth | `/basic_auth` |
| should see that content is no longer displayed when window is closed | `/iframe` |
| can switch to a frame via element | `/nested_frames` |
| can switch to a frame via function | `/nested_frames` |
| should reset the frame when the page is reloaded | `/iframe` |

Every test in that spec that does *not* reach that host passed, 99 of them. The errors are all "still not displayed after 5000ms" / "doesn't exist" rather than anything protocol-shaped.

**The tests that do exercise this change passed in that same CI run:**

```
✓ should support custom element command on non-existing elements
✓ should support legacy custom element command on existing elements
✓ should support browser custom command
```

The middle one is `browser.addCommand('myElementLegacyCustomCommand', fn, true)` at module scope in `test.e2e.ts` — the deprecated positional form, registered before the session starts, so it goes through exactly the protocol-stub normalization this PR adds.

**Locally on this branch, the whole spec is green**, including all six of the CI failures:

```
105 passing (1m 29s)

✓ can query shadow elements
✓ supports basic auth
✓ should see that content is no longer displayed when window is closed
✓ can switch to a frame via element
✓ can switch to a frame via function
✓ should reset the frame when the page is reloaded
```

`the-internet.herokuapp.com` answers in ~1.5 s now, so it looks like it was unavailable or slow while that job ran. I cannot re-run the job myself — could someone with write access retry it?

**One correction to the PR description.** I reported three `tsc` errors in `packages/wdio-runner` (`TimingTracker`, `cpuProf`, `heapProf`) and said they were present on an unmodified `v10`. They were, but the cause was a stale local build of the workspace dependencies, not anything in the repository: after `npm run compile:all:core && npm run compile:all:main` they are gone and `@wdio/runner` builds cleanly. Sorry for the noise.

mccmrunal:
LGTM

PR #15674, opened 2026-09-21
breaking(webdriverio): return a MultiRemoteElementArray from multiremote $$

PloneMraz:
### Proposed changes

Fixes #15650 — `multiRemote.$$()` now always returns an ElementArray-like object, and `WDIO_ENABLE_MULTI_REMOTE_ELEMENT_ARRAY` is gone.

**The flag.** `multiremote.ts` returned the plain zipped array unless the env variable was set, so on multiremote `$$` had no `map` / `filter` / `selector` / `foundWith` and expect matchers that iterate an ElementArray did not work. v9 shipped the array behind the flag in #15532; this makes it the default.

**The type.** `MultiRemoteElementArray` is now a real exported type instead of a cast. `enhanceElementsArray` gained an overload for multiremote elements, which removes both `as unknown as` casts at the call site:

```diff
-const elementArray = enhanceElementsArray(
-    wrappedResult as unknown as WebdriverIO.Element[],
-    this as unknown as WebdriverIO.Browser,
-    selector,
-    commandName
-)
-Object.assign(elementArray, { isMultiremote: true })
+const elementArray = enhanceElementsArray(wrappedResult, this, selector, commandName)
+elementArray.isMultiremote = true
```

`isMultiremote: true` is part of `MultiRemoteElementArray`, so it is declared rather than stitched on, and `MultiRemoteBrowserCommandsType` returns it for `$$` / `custom$$` / `react$$` instead of `MultiRemoteElement[]`.

### Acceptance criteria

- [x] `multiRemote.$$()` always returns an ElementArray-like object
- [x] `process.env.WDIO_ENABLE_MULTI_REMOTE_ELEMENT_ARRAY` is deleted
- [x] A real `MultiRemoteElementArray` type is exported (no `as unknown as WebdriverIO.Element[]` cast)
- [x] `isMultiremote: true` is part of that type
- [x] Unit tests and e2e no longer toggle the env flag
- [x] Docs for MultiRemote `$$` describe the new return value

### Breaking change

Index access is unchanged, so code that only reads `elements[0]` keeps working. Code that relied on the result being a plain `MultiRemoteElement[]` — `Array.isArray` still returns `true`, but the prototype methods are the async ones — needs to treat it as an ElementArray. The docs section says so, and #15664 can pick it up for the migration guide.

### Verification

- `npx vitest run packages/webdriverio` — 1346 passed, 1 skipped, no type errors.
- Three new unit tests in `packages/webdriverio/tests/multiremote.test.ts`: the array shape without any opt-in, the instances carried by each entry, and the async array helpers.
- Mutation check: returning `wrappedResult` early again (the old flagged-off path) turns two of those three red.
- e2e: `wdio/wdio-multiRemote.conf.ts` — 30 passing on Chrome ×2 + Firefox, including the rewritten "should return a MultiRemoteElementArray at runtime" that used to sit behind the env flag.
- `npm run test:oxlint` — clean.

One note on the e2e: I removed the `describe('when enabling process.env.…')` wrapper and folded its assertions into a plain test, since the behaviour it opted into is now the default. It also asserts `isMultiremote` and no longer needs the `ElementArray` casts.

### Types of changes

### Checklist


PloneMraz:
The red checks here are not from this change, as far as I can tell — evidence rather than an assertion:

**`E2E Launch Test (ubuntu-latest)`** fails while installing chromedriver, before any test runs:

```
Error: Failed downloading chromedriver v152.0.7977.82 using
  {"cacheDir":"/tmp","buildId":"152.0.7977.82","platform":"linux",…}:
  The browser folder (/tmp/chromedriver/linux-152.0.7977.82) exist…
    at async setupChromedriver (packages/wdio-utils/build/node.js:262:7)
    at async Promise.all (index 3)
```

Four workers install into the same `cacheDir`, and one of them finds the folder already there. In the **same run**, the other two matrix entries passed:

```
E2E Launch Test (ubuntu-latest) -> failure
E2E Launch Test (macos-15)      -> success
E2E Launch Test (windows-latest)-> success
```

If this PR broke launching, all three would fail rather than one.

**`check-labels`** ran at 05:03 while the PR had no labels. It now carries `PR: Breaking Change 💥`, so it should pass on a re-run.

**`CI`** is the aggregator job for the two above.

I cannot re-run these myself — could someone with write access retry the run?

Locally on this branch: `npx vitest run packages/webdriverio` is 1346 passed / 1 skipped with no type errors, and the multiremote e2e suite (`wdio/wdio-multiRemote.conf.ts`) is 30 passing on Chrome ×2 + Firefox, including the rewritten `should return a MultiRemoteElementArray at runtime`.

PloneMraz:
Both review points were right — fixed in `77daa49`.

**Return types mismatch runtime.** Confirmed: `commandWrapper` only zips and enhances for `commandName === '$$'`; `custom$$` and `react$$` fall through to `return result`, one result per instance. So only `$$` now claims a `MultiRemoteElementArray`, and the other two keep the element-list type they already had:

```ts
} & {
    [K in keyof Pick<BrowserCommandsType, '$$'>]: … => ThenArg<WebdriverIO.MultiRemoteElementArray>
} & {
    [K in keyof Pick<BrowserCommandsType, Exclude<MultiElementCommandNames, '$$'>>]: … => ThenArg<WebdriverIO.MultiRemoteElement[]>
}
```

I did not extend the runtime zipping to `custom$$` / `react$$` — that would be a second breaking change beyond what #15650 asks for. Happy to open it separately if you want them consistent.

**Iterators erase multiremote types.** Also confirmed, and worse than it looked: `AsyncIterators<T>` took a `T` but hard-coded every element position to `WebdriverIO.Element`, and `forEach`, `find`, `some`, `every`, `filter`, `mapSeries` and `reduce` each declared their own `<T>` that shadowed it — so the interface parameter was reaching only `array: T[]` and `thisArg`. The element positions now use the interface's `T`, `filter` / `filterSeries` return `Promise<T[]>`, `entries()` yields `[number, T]`, and `thisArg` is `unknown` instead of a stray generic.

Covered in `tests/typings/webdriverio/async.ts`:

```ts
expectType<string[][]>(await mrElems.map((el) => el.instances))
expectType<WebdriverIO.MultiRemoteElement[]>(await mrElems.filter(async () => true))
// and the single-browser side still resolves to Element
expectType<WebdriverIO.Element[]>(await elems.filter(async () => true))
```

Mutation check on that: putting `WebdriverIO.Element` back into `map` and `filter` makes exactly those assertions fail —

```
async.ts(66,57): error TS2339: Property 'instances' does not exist on type 'Element'.
async.ts(67,50): error TS2345: Argument of type 'Element[]' is not assignable to parameter of type 'MultiRemoteElement[]'.
```

Re-verified after the change: `packages/webdriverio` typechecks clean, 1346 unit tests pass, `tests/typings/webdriverio` compiles with no errors in `async.ts`, and `npm run test:oxlint` is clean. (`config.ts` and `globalImport.ts` report two typing errors, but they do so on an unmodified `v10` as well.)

Issue #14959, opened 2025-12-25
WebdriverIO v10 Wish List


christian-bromann:
Let's collect a list of (possible breaking) changes we want to land in v10 of WebdriverIO.

- [ ] require Node.js v22 and up (#15136)
- [ ] throw an error if `$` matches multiple elements (#14703)
- ...

Please comment in this thread if there is anything you would like us to include.

dprevost-LMI:
From https://github.com/webdriverio/expect-webdriverio/issues/1893

1. Review the augmentation to have better compatibility and alignment, or get rid of them!
    - For example, provide a wdio "extended" lib for Jest and Jasmine, like [this library](https://github.com/jest-community/jest-extended)
2. Review Jasmine augmentation/framework so that it leverages their `expectAsync` API and potentially fixes [this issue](https://github.com/webdriverio/expect-webdriverio/issues/1407#issuecomment-3194562421) and others where merging the async in the expect seems to break some tsc
3. Try to fix most of the other issues raised in the above ticket!

dprevost-LMI:
- [ ] Remove the global `multiremotebrowser` in favour of `multiRemoteBrowser` that will be soon deprecated in v9

christian-bromann:
@dprevost-LMI what are your thoughts on moving to Vitests `expect` primitive?

dprevost-LMI:
FYI: I'm a big Jest user. For vitest I'm starting to get more used to it

At first glance (and on paper), Vitest test has everything that expect-wdio already offers

Dirty quick analysis:
From [“expect” API doc](https://vitest.dev/api/expect.html#expect)
It would keep the “Jest” like matcher name approach “toBe” (and not by default the “chai” approach) 
> and also Jest compatible assertions built on top of chai

Mentions of a custom message approach here that we could align (I wonder if using "Jest" like this is still supported)
> Unlike Jest, Vitest supports a message as the second argument - if the assertion fails, the error message will be equal to it.

They have a [documented soft](https://vitest.dev/api/expect.html#soft) approach that is good

Support of the asymmetric matcher too
  - Note I would potentially deprecated all wierd option using a duplicate asymetrics matcher approach, like containing StringOptions. Other options without an equivalent, such as trim, should be retained, though.

Support of
- [Snapshot](https://vitest.dev/guide/snapshot.html#snapshot)
- [Request mocking](https://vitest.dev/guide/mocking/requests.html#mocking-requests)
[...]

Issue #14703, opened 2025-08-13
[💡 Feature]: Strict element count validation in `browser.$()` API


just-boris:
### Is your feature request related to a problem?

Currently the `await browser.$('button').click()` command clicks on the first matching button even if there are multiple buttons. This may be unexpected.

### Describe the solution you'd like.

I expect there is a built-in validation that the element under interaction is a single instance.

Such validation is available by default in other popular tools

* Playwright: https://playwright.dev/docs/locators#strictness
* Cypress: https://docs.cypress.io/api/commands/click#Click-all-elements-with-id-starting-with-btn

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

erwinheitzman:
Hi @just-boris I can certainly see the value of this but I believe that for this to work we'd have to query all elements for each command, adding a significant delay for each command being executed.

I'm open to suggestions on how to achieve this without an impact in performance and like the idea in general 👍

mccmrunal:
I have started working on this and I have few questions

Playwright and Cypress throw by default when $() matches multiple elements. Should we:

A) Default to 'warn'
B) Default to true (throws, matches Playwright/Cypress)
C) Keep false 


I have implemented accessibility/<name>[role=<role>] syntax (e.g., accessibility/Submit[role=button]). Is this syntax acceptable, or should we use a shorter prefix like a11y/?

would really appreciate a quick reply from someone🤞

christian-bromann:
> Currently the `await browser.$('button').click()` command clicks on the first matching button even if there are multiple buttons. This may be unexpected.

This is how `document.querySelector` works too and never felt it to be unintuitive. However I can see the value of making sure selecting elements is more strict.

I've added this feature to the v10 wish list. I wonder tho what the best resolution for this is 🤔 should we make selector strictness configurable?

mccmrunal:
we ship the capability now for those who want robust strict selectors, and  v10 we can simply flip the default to true.

christian-bromann:
Awesome, amazing contribution @mccmrunal

PR #15677, opened 2026-09-21
breaking(webdriverio): make `$` strict and throw when the selector matches multiple elements

mccmrunal:
### Proposed changes

Closes #15666 (refs #14703, second item on the v10 wishlist #14959).

`$` now represents **exactly one** element. When a selector resolves to more than one element the command throws a `StrictSelectorError` instead of silently using the first match:

```
strict mode violation: `$("button")` resolved to 12 elements, expected 1.
Use `$$("button")` to work with all matches, `$$("button")[0]` if you explicitly want the first one, or narrow down the selector so it matches a single element.
Opt out for a single call with `$("button", { strict: false })` or globally by setting `strictSelectors: false` in your WebdriverIO config.
```

Behavior:

- **Strict by default** via a new `strictSelectors` config option (`true`).
- **Per-call opt out**: `$(selector, { strict: false })` — and `{ strict: true }` to opt back in when disabled globally.
- `$$` is unchanged (zero or many).
- A selector matching nothing still returns a lazily-resolved element, so `waitForExist` and auto-waiting behave exactly as before.
- Strict at **every step of a chain**, and for mobile matcher objects, shadow DOM and custom locator strategies. Element references (`$(await browser.getActiveElement())`) are exempt — they can only be one node.
- Elements **remember their strictness**, so re-fetching them (stale-element refetch, implicit wait, `waitForExist`) keeps the semantics of the original call. Elements from `$$` are never strict.

Implementation note: a strict `$` goes through the same lookup path `findElement` would have picked, just in its "find elements" flavor (`findDeepElements` on BiDi, `findElements` otherwise) — counting the matches is the only way to enforce the rule. It is still a single round trip, but it is visible to services and WebDriver mocks that key off the `findElement` command.

Two supporting fixes came out of this:

- `@wdio/config`: `validateConfig` dropped falsy option defaults (`false`, `0`, `''`) because it tested the default for truthiness. That is what made `strictSelectors: false` inexpressible as a default. The only caller with falsy defaults (`TESTRUNNER_DEFAULTS` in the launcher) discards the return value, so nothing else changes.
- `@wdio/runner`: the `$` global was registered as `(selector) => browser.$(selector)`, which silently dropped the new options argument under the testrunner (it worked in standalone). Now forwards all arguments.

Internal `$` callers that intentionally want the first of several matches (`selectByVisibleText`'s XPath union, the native dialog button tap) opt out explicitly.

### Types of changes

- [x] Documentation

### Tests

- `packages/webdriverio/tests/strictSelectors.test.ts` — 12 unit tests: single match, zero match, multiple match, chaining, `$$` unaffected, per-call and global opt-in/opt-out, element references, strictness persistence.
- `e2e/wdio/headless/strictSelectors.e2e.ts` + fixture — 7 tests against real Chrome covering single/zero (with `waitForExist`)/multiple match, chained scope, shadow DOM and the per-call opt-out. **Verified passing locally.**
- `tests/typings/webdriverio/{async,config}.ts` — type coverage for the option and the per-call argument.
- Full unit suite (4288 tests), all typings suites, `test:smoke` and the headless e2e suite pass locally.

Two test-infra switches keep the pre-existing suites meaningful rather than rewriting them wholesale:

- `tests/setup/strictSelectors.ts` (new vitest setup file) flips the default off for unit tests. The fetch mock answers every "find elements" request with three references, so otherwise nearly every `$` in the existing unit tests would be a strict-mode violation — those tests are about the commands they cover, not about strictness.
- `tests/helpers/launch.js` does the same for the smoke suites, whose `@wdio/webdriver-mock-service` scenarios stub `findElement` with exact `nock` call counts.

The e2e suite deliberately runs **with** strict mode on; `e2e/wdio/headless/test.e2e.ts` was migrated to `$$(...)[0]` in the seven places that relied on `$` picking the first of many — the same migration users will do.

### Docs

- `website/docs/Selectors.md` — new "Strict Mode" section: what changed, what is not affected, how to audit (`$$(sel).length`), how to migrate, how to opt out.
- `website/docs/v10Migration.md` — new "From v9 to v10" page (added to the sidebar) leading with this change, structured so the remaining v10 items can be appended.
- `website/docs/Configuration.md` — `strictSelectors`.
- JSDoc on both `$` commands (drives the API docs), plus `StrictSelectorError` is exported from `webdriverio` so it can be caught by class.

As called out in the issue, `@wdio/codemod` cannot auto-fix this — whether a second match is a bug or intentional is a judgement call — so the docs cover auditing instead.

### Checklist

### Open questions for reviewers

1. **Sibling commands.** `custom$`, `react$` and `shadow$` have their own `$$` counterparts and are left non-strict here. Happy to extend the rule to them in this PR if you'd rather land it all at once.
2. **`validateConfig` fix.** It is technically unrelated, but without it a `false` default is silently dropped. Fine here, or split into its own PR?


mccmrunal:
Closing — opened in error, not intended for review.

PR #15678, opened 2026-09-21
fix(@wdio/utils): share a driver install between concurrent setups

### LINKED ISSUE #15676: [🐛 Bug]: concurrent workers racing on a shared cacheDir fail chromedriver setup
### What happened?

With several workers starting at once and sharing one `cacheDir`, chromedriver setup fails on a cold cache:

```
Error: Failed downloading chromedriver v152.0.7977.82 using
  {"cacheDir":"/tmp","buildId":"152.0.7977.82","platform":"linux",
   "browser":"chromedriver","unpack":true}:
  The browser folder (/tmp/chromedriver/linux-152.0.7977.82) exists but the executable
  (/tmp/chromedriver/linux-152.0.7977.82/chromedriver-linux64/chromedriver) is missing
    at async setupChromedriver (packages/wdio-utils/build/node.js:262:7)
    at async Promise.all (index 3)
    at async Launcher.run (packages/wdio-cli/build/index.js:888:7)
```

The message appears **twice** per run — once for the first attempt and once after `_install`'s retry — so the cleanup-and-retry path added in #15608 does not recover it.

`setupChromedriver` checks `fsp.access(executablePath)` and installs when it is missing. With N workers that check races: each sees no executable, each calls `install()` into the same folder, and whichever arrives second finds a directory that exists but is not finished being extracted. `_install`'s retry then removes that directory — which, if a sibling worker is still extracting into it, can take the other worker's install down as well. I have not instrumented the deletion race, so that last step is a hypothesis; the double failure message is measured.

### Where it shows up

This repository's own `E2E Tests / E2E Launch Test` job on `ubuntu-latest`, with `Execution of 4 workers`. Observed on three consecutive pull requests — #15672, #15674 and #15675 — and in each run the `macos-15` and `windows-latest` entries of the same matrix passed, which is what makes it look like a race rather than a broken download:

```
E2E Launch Test (ubuntu-latest)  -> failure
E2E Launch Test (macos-15)       -> success
E2E Launch Test (windows-latest) -> success
```

It reads as a red check on unrelated pull requests, so it costs review time as well as CI time.

Concurrent workers on a cold shared cache should end up with one chromedriver install and no failures — for example by serialising the install per `cacheDir`/`buildId` (a lock or an in-process promise cache), extracting to a temporary directory and renaming it into place, or treating "folder exists but executable missing" as "another install is in flight" and waiting rather than deleting.

### How to reproduce the bug.

1. Empty the cache directory (`rm -rf /tmp/chromedriver`).
2. Run a WDIO suite with `maxInstances` > 1 and several spec files, so more than one worker starts before any install finishes.
3. On a cold cache the second worker onwards can hit the error above.

It is timing dependent — the same job passes on other runners in the same run.

### Relevant log output

See the quoted error above; the same line appears twice, before and after the retry.

### Code of Conduct

PloneMraz:
### Proposed changes

Fixes #15676.

**The cause.** `SUPPORTED_BROWSERNAMES.chrome` is `['chrome', 'googlechrome', 'chromium', 'chromium-browser']`, so `isChrome()` is true for both `chrome` and `chromium`. `mapCapabilities` groups the setup work by **browser name**, which keeps them apart, and both then call `setupChromedriver(cacheDir, undefined)` at the same time. The two installs race on one cache directory and the loser sees a folder that exists but has no executable in it yet.

This repository's own `wdio.local.conf.ts` has exactly that shape — `chrome`, `firefox`, `edge`, `chromium` — and `chromium` is capability index 3, which is what the failing stack trace pointed at:

```
at async setupChromedriver (packages/wdio-utils/build/node.js:262:7)
at async Promise.all (index 3)
at async Launcher.run (packages/wdio-cli/build/index.js:888:7)
```

`_install`'s retry does not recover it: the same message appears twice per run, once for the attempt and once for the retry, because the other install is still in flight.

**The fix.** Concurrent setups of the same driver, cache directory and version share the install that is already running. The map entry is dropped once the promise settles, so a later call can still retry a failed download, and after a successful one the executable is simply found in the cache.

Applied to geckodriver and edgedriver too, since the same grouping holds for them — `edge`, `microsoftedge` and `msedge` are one family, as are the Firefox spellings.

### Verification

- `npx vitest run packages/wdio-utils` — 369 passed, no type errors.
- Two new tests: two concurrent `setupChromedriver` calls for the same cache directory install once, and two calls for *different* cache directories still install twice.
- Mutation check: dropping the in-flight lookup makes the first test fail with `expected "spy" to be called 1 times, but got 2 times`, and leaves the second green.
- `npx eslint` on both changed files — clean.

Two notes on the wider suite, both unrelated and both reproduced on an unmodified `main` here: `wdio-lighthouse-service` fails to import `lighthouse/lighthouse-core/...` because my local `node_modules` had Lighthouse 13 from another branch, and `wdio-browser-runner`'s `mockHoisting` snapshot mismatches on a Windows drive prefix.

### Types of changes

### Checklist


PloneMraz:
Both findings were right — fixed in `7418066`.

**Equivalent versions bypass sharing.** Correct: the key used the raw `driverVersion` while `installChromedriver` resolved it afterwards, so `undefined` and `'stable'` got different keys for the same download. `setupChromedriver` now resolves the platform and build id first and keys on those:

```ts
const version = driverVersion || getBuildIdByChromePath(await locateChromeSafely()) || ChromeReleaseChannel.STABLE
const buildId = await resolveBuildId(Browser.CHROMEDRIVER, platform, version)

return shareDriverSetup(
    `chromedriver:${cacheDir}:${platform}:${buildId}`,
    () => installChromedriver(cacheDir, platform, version, buildId)
)
```

Resolving reads no state and writes nothing, so doing it once per caller before the sharing point costs nothing.

**Failure eviction remains untested.** Also correct — the PR text claimed retryability that nothing checked. There is now a test where the shared install rejects and a later call starts a fresh one.

Two new tests, and each mutation lands on the one that covers it:

| mutation | red test |
| --- | --- |
| key by the raw version again | *shares one install between version requests that resolve to the same build* |
| keep the promise in the map after it settles | *lets a later call retry after a shared install failed* (and the sharing test, since the map then leaks across cases) |

`npx vitest run packages/wdio-utils` — 371 passed, no type errors; `npx eslint` on both files is clean.

I left geckodriver and edgedriver keyed on the raw version. They delegate resolution to the `geckodriver` and `edgedriver` packages, so there is no resolved build id to key on without duplicating their logic here, and the family-name collision that motivated this (`edge` / `microsoftedge` / `msedge`) produces the same raw version anyway. Say the word if you would rather they were normalised too.

PR #15680, opened 2026-09-21
breaking(webdriverio): remove deprecated throttle and touchAction commands

devangpratap:
## Proposed changes

Part of #15654. Removes two of the four deprecated commands on the v10 line: `browser.throttle` and `browser.touchAction` / `element.touchAction`.

- `throttle` was a thin alias that warned and called `throttleNetwork`. Deleted the command and its export.
- `touchAction` is deleted from both scopes, along with its helpers in `commands/constant.ts` (`formatArgs`, `validateParameters`, the action lists) and the public `TouchAction`, `TouchActions`, `TouchActionParameter` and `ActionTypes` types. The raw `touchPerform` / `multiTouchPerform` protocol commands are not touched.
- `examples/appium/appium-draws-its-logo.js` drew the logo with `touchAction`. It now uses touch pointer chains through `browser.action` / `browser.actions`. The old example used `moveTo` as relative offsets, so the helpers now return the absolute points those offsets added up to, and the strokes that used to go through `multiTouchPerform` (several arcs at once) run as parallel pointer chains, one per finger. Also swapped the old `driver.click(selector)` call in the same file for `$(selector).click()`.
- Dropped the `touchAction` block from `tests/typings/webdriverio/async.ts` and replaced the `touchAction` example in CONTRIBUTING's "Testing Type Definitions" section with the `dragAndDrop` one. The old example also pointed at `tests/typings/sync/sync.ts`, which no longer exists.

Not in this PR, so the issue should stay open:
- `executeAsync`: `isStable` and `scrollIntoView` still call it internally, so it needs its own PR that rewrites those paths to `execute` first.
- `switchToFrame` public docs/typings.
- Codemod transforms and the migration guide sections (#15664).

Happy to take `executeAsync` next.

Side note: I tried adding `@ts-expect-error` checks for the removed commands to the typings tests, but they don't fire. `expect-webdriverio@6.0.9` pulls in the published `webdriverio@9.31.2`, whose global `WebdriverIO.Browser` augmentation still declares `touchAction` and `throttle`, so the typings tests see both declarations until that dependency moves to v10.

## How you tested

- `npx vitest run packages/webdriverio/tests`: 184 files, 1325 passed, 1 skipped.
- `pnpm run test:typings:webdriverio`, plus the `webdriver`, `mocha`, `jasmine` and `cucumber` typings suites: all pass.
- `npx oxlint` on the touched paths: clean.
- After `pnpm run setup`, the built `webdriverio` package has no `touchAction` or `throttle` command, export or type left.

- [x] Unit tests for the touched package(s)
- [x] Type definition tests (if the public API / command types changed)
- [ ] Smoke or e2e (only if the testrunner, launcher, or a real browser path changed)

## Types of changes

## Checklist

- [ ] I have added the necessary documentation (if appropriate). Migration docs are tracked in #15664.

## Backport Request

- [x] This change is solely for `v10` and doesn't need to be back-ported


christian-bromann (approved):
👍

